Issue #4777, opened 2018-06-23
Datafile suggested citation doesn't include file DOI


qqmyers:
The suggested citation that appears on a 4.9 datafile page is currently just the suggested citation for the dataset with the file name and the text'[filename]' postpended. With file DOIs now being generated, it seems like the suggested citation should be updated, e.g. to postpend the file name and file DOI for example. An alternative would be to change the whole citation to just reference the file (which would require changing the text 'This file is part of "<dataset name>". If you use this file, please cite the data project:' that is directly above the citation block). I personally think that keeping the citation focused on the dataset and just adding the filename and DOI is a good approach, but I don't know if that would work for everyone/if this needs to be a config option, etc.

qqmyers:
Per discussion in  https://github.com/IQSS/dataverse/issues/2438 and on the community group (https://groups.google.com/forum/#!msg/dataverse-community/ZY53hidy-MU/MaU2YtJUBwAJ), it sounds like the idea of including both the dataset and file DOI in the citation was not a desired solution. 

It may still be an open question whether a direct data file citation is desirable as an alternative to citing the dataset and mentioning the file...

djbrooke:
@scolapasta you mentioned yesterday in backlog grooming that there were a few options for moving forward here. Assigning to you for the specifics.

@qqmyers - do you have any availability to pick this up once we get feedback from the Community + @mercecrosas  and others about what we want to do here?

qqmyers:
@djbrooke - I think so. Looks like the latest discussion is along the lines of the image below. In testing how to do this, I realized that the code for the endnote, ris, and bibtex downloads has to be modified as well... (and the help link to data citation practices might need to move so it is not repeated.)  

![image](https://user-images.githubusercontent.com/6731983/42059204-a9895ca2-7af0-11e8-899a-67e06ce33afa.png)

pdurbin:
@qqmyers thanks for making pull request #4802! I just dragged this issue to code review at https://waffle.io/IQSS/dataverse

pdurbin:
It was driving me crazy that DataCitation.java had zero code coverage among it's 623 lines so I pushed a test in 4c67e61 to increase it to 24%. I don't want to hold up QA but in the future I hope we can add more tests.

## Before

![screen shot 2018-07-09 at 3 12 41 pm](https://user-images.githubusercontent.com/21006/42471409-910bf66e-838b-11e8-85fc-ab1da769fdb3.png)

## After

[...]

PR #4802, opened 2018-07-03
4777 datafile suggested citation with file doi

qqmyers:
## New Contributors

## Related Issues

- connects to #4777 : Datafile suggested citation doesn't include file DOI

## Pull Request Checklist


qqmyers:
I consider this PR to be in-progress, awaiting further discussion of the on-screen citation display and how to handle the exportable citation formats. 
The citation display has been updated to be in-sync with the community forum discussion, adding a new 'direct' file citation option that uses the file's DOI, rather than using the dataset DOI followed by the file name. The PR shows both options and has Bundle text strings to explain them.

The more significant code changes are related to my refactoring to bring the various export formats into one class and to run everything through the DataCitation class. The existing code had slight variations in how, for example, harvested datasets were handled, and whether the configured citation date was used or if the publication date was hardcoded. My sense was just that the code had been written over time, so I tried to use the 'best of' option in the DataCitation class and was able to consolidate the retrieval/parsing of info for the exported citations into one class with two constructors - one for datasets starting with a DatasetVersion, and the other for files starting from FileMetadata. I added one parameter - boolean direct - that, if true for file citations, retrieves the file DOI rather than the dataset one.

With these changes, I was able to bring the code for exporting Bibtex, RIS, and Endnote into the class, and reconfigured them to directly write to a steam (which is what's really needed in FileDownloadServiceBean). 

I expect further discussion might adjust the citation display and probably will change the specifics of what's written in the bibtex, RIS, and Endnote outputs, but I wanted to share the code and perhaps get opinions on the general refactoring - if there's some reason not to consolidate the parsing code, or there's a better approach, I can probably make that type of change in parallel with ongoing discussion of the specifics of the display and file contents.

adam3smith:
Thinking more about this, I'm not convinced what we're doing with this is optimal. As of now we have three different citations:
1. Dataset citation
2. Dataset with file citation (equivalent of current data file citation)
3. Data file citation.

1. is on the dataset landing page, 2 and 3 are on the data file landing page.

Logically these also correspond to three different sets of metadata. I think that's very confusing. With two different versions of a file citation on the file landing page, it's also unclear which of them should be encoded in schema.org metadata on that page.

I would suggest, instead, to make 1. and 2. identical, so we have 
1. Dataset citation 
2. Data file citation

In other words, get rid of the <filename>[file name] at the end of the dataset citation on file landing pages.

If people want to cite a specific file, they should just use 2. In the more common case, where they're looking at a file but really mean to cite the whole dataset, we should have 1. available on the file landing page (as per the discussion in the community group). Thoughts?

adam3smith:
To illustrate: Currently implemented
image

Suggested change:
image

(recall that QDR uses "data project" instead of "dataset")

scolapasta:
@adam3smith I had been wondering the same thing. Would you be ok to copy these comments into the issue itself?

adam3smith:
Sorry, not clear which issue itself you're referring to?

scolapasta:
#4777 
(you put the comments int the PR, rather than the issue)

adam3smith:
oops, sorry; this was meant for 4777 in the first place

qqmyers:
I think this PR may be ready again for review. It should now contain @adam3smith 's updates to the RIS and endnote exports and the consensus to provide a dataset and file citations on the file page (with the latter using the file DOI if published and appending the filename to the dataset citation if not).

src/main/webapp/file.xhtml
```diff
@@ -222,51 +223,95 @@
...
                                         </div>
+                                    </div>
+                                </div>
+                            </div>
+                            <div class="col-xs-12">
+                                <div class="alert alert-warning" style="margin:0;padding:0 .5em;background:none;border-radius:0;border:0;color:#333;">
+                                  <h:outputFormat value="#{bundle['file.citation.datafile']}" escape="false" />
+                                </div>
+                            </div>
+                            <div id="citation-block-2" class="col-xs-12">
```
mheppler: While re-using the ID and adding a "-2" to the end solves the duplicate ID warnings when validating, it breaks the CSS being applied to the original ID. I am working on a commit to resolve these minor layout issues, and maybe refactor some of the original CSS code to make it more scalable.

src/main/webapp/file.xhtml
```diff
@@ -109,6 +109,7 @@
                                         </button>
                                         <ul class="dropdown-menu pull-right text-left" role="menu">
                                             <ui:fragment rendered="#{systemConfig.provCollectionEnabled}">
+                                            <!--  Should be FilePage.isLockedFromEdits() ? -->
```
mheppler: Good question. Is this a question? I will look into this more.

src/main/java/Bundle.properties
```diff
@@ -1648,8 +1648,11 @@ file.share.fileShare.tip=Share this file on your favorite social media networks.
 file.share.fileShare.shareText=View this file.
 file.title.label=Title
 file.citation.label=Citation
-file.citation.notice=This file is part of "{0}". If you use this file, please cite the dataset:
+file.citation.notice=This file is part of "{0}".
```
mheppler: What originally was one message, on one line, is now two messages, on two lines. It unnecessarily takes up more real estate. I'm going to try to clean this up.

src/main/webapp/file.xhtml
qqmyers: ~yes. Seems like a cut/paste error but separate from what I was working on.

pdurbin:
#4971 is the new version of the pull request so I'm closing this one.

Issue #4219, opened 2017-10-19
UI - Bootstrap 3.3.7 Upgrade


mheppler:
Upgrade our Bootstrap library from v3.3.2 to v4.0. Read more here: https://getbootstrap.com/docs/4.0/migration/

As of this posting, it is still in beta, but that [beta release](https://blog.getbootstrap.com/2017/08/10/bootstrap-4-beta/) was two years in the making, with "6 major alpha releases", and has been out for two months now.

They do warn that they've dropped older browser support. "Need support for IE8/IE9, Safari 8-, iOS 8-, etc? Keep using Bootstrap 3." So we should take that into consideration before upgrading.

mheppler:
There was some discussion with @pameyer and @matthew-a-dunlap in Slack about the PrimeFaces upgrade, specifically the jQuery libraries included since those need to be updated because of security concerns. It was determined that as long as we upgrade to Bootstrap 4 and PrimeFaces 6.2, we'll be good.

>PrimeFaces 6.2 is out now, and if we upgrade to Bootstrap 4, we would need no lower than 6.2 due to the jQuery versions.

It appears that [PrimeFaces 6.2 uses jQuery 3.2.1](https://github.com/primefaces/primefaces/issues/3062) which is the version that [Bootstrap 4 recommends](https://getbootstrap.com/docs/4.0/getting-started/introduction/).

mheppler:
This appears to be the migration guide to follow for 3.x to 4.x.

https://v4-alpha.getbootstrap.com/migration/

mheppler:
It appears that I overlooked the fact that there was a 3.3.7 release, which would provide a lot of the improvements we were aiming for in upgrading (read: jQuery 3 support).

http://blog.getbootstrap.com/2016/07/25/bootstrap-3-3-7-released/

I have retitled this issue to upgrade to that release, as the 4.0 upgrade would require far more effort than we intended for this issue. If and when we redesign the UI as a whole, it would be worth considering an upgrade to Bootstrap 4.0. But for our proposes now with this current UI, 3.3.7 is sufficient and will get done much sooner.

mheppler:
Initial commit for upgrading Bootstrap versions in the app and guides. Also cleaned up or upgraded a few other random issues.

DATAVERSE
- Removed local version of Bootstrap 3.3.2 and updated `dataverse_template.xhtml` to point to CDN copies of v. 3.3.7
[...]

Issue #4784, opened 2018-06-26
Update Chinese bundle properties file: Bundle_zh_CN.properties


pengchengluo:
The English bundle properties file has about 1800 key-value pairs. However, the Chinese bundle properties file updated two years ago and only has about 500 key-value pairs.  When viewing the web pages from Chinese web browser, there are only a few Chinese Characters. So, we want to translate all the key-value pairs and update some error translation, such as 
dataverse.permissions.roles.description.title=Describe the role (1000 characters max) has been erroneously translate into "描述中的作用（1000个字符）" which can't been understood by Chinese. We have translated 1000 key-value pairs based on dataverse 4.9 and plan to complete it in the following few days. We hope our updated Bundle_zh_CN.properties file could be added into dataverse code repository.

Besides, we also hope to continually update the Bundle_zh_CN.properties file in future releases of dataverse.

![image](https://user-images.githubusercontent.com/7343585/41883430-2ac78c16-7921-11e8-9a29-b4f4ef5c6375.png)

oscardssmith:
While you are doing this, it probably makes sense to reduce duplicate bundle entries if you see them. A script I ran last summer found that ~20% are duplicates.

djbrooke:
Tagging @scolapasta for his thoughts as he's coordinating internationalization efforts with other groups (see the list of child issues in #209)

scolapasta:
@pengchengluo that's great! We'll review when it's ready.

Be aware we are planning on updating / streamlining the bundles as part of both internationalization and the Dataverse 5.0 release, i.e. the 1800 will hopefully be reduced in scope. 

(Generally, we're always in favor of "bringing back the forks", so let is know how we can help with that)

pengchengluo:
@oscardssmith，I extract the keys from Bundle.properties and use Excel to find the following 14 duplicate keys:
advanced.search.datasets.persistentId.tip
advanced.search.files.persistentId.tip
authenticationProvider.name.builtin
authenticationProvider.name.github
authenticationProvider.name.google
authenticationProvider.name.null
authenticationProvider.name.orcid
authenticationProvider.name.orcid-sandbox
authenticationProvider.name.shib
citationFrame.banner.countdownMessage.seconds
ingest.csv.invalidHeader
ingest.csv.lineMismatch
ingest.csv.nullStream
ingest.csv.recordMismatch
[...]

PR #4806, opened 2018-07-03
Upgrade Bootstrap v. 3.3.7  [ref #4219]

mheppler:
## Related Issues

- connects to #4219 UI - Bootstrap 3.3.7 Upgrade 

## Pull Request Checklist


src/main/webapp/dataverse_template.xhtml
```diff
@@ -23,8 +23,8 @@
         </f:facet>
         <link type="image/png" rel="icon" href="#{resource['images/favicondataverse.png']}" />
         <link type="image/png" rel="image_src" href="#{resource['images/dataverseproject.png']}" />
-        <link type="text/css" rel="stylesheet" href="#{resource['bs/css/bootstrap.min.css']}?version=#{systemConfig.getVersion()}" />
-        <link type="text/css" rel="stylesheet" href="#{resource['bs/css/bootstrap-theme.min.css']}?version=#{systemConfig.getVersion()}" />
+        <link type="text/css" rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous"/>
```
pameyer: Summarizing some discussion w\ @mheppler and @kcondon Tue, using an external CDN brings up policy and technical questions.

Technical:
- Should the CDN be user-configurable without forking?
- Should there be a fallback / failover to a local copy if the CDN is down (aka - this is introducing a dependency on an additional external service)?
- Resource integrity was discussed (aka - knowing the CDN is providing the expected resource without modification); @mheppler already solved this with the `integrity` attribute.
- Performance implications were also discussed (without measuring, my intuition is that there would be minimal improvement from this change); but performance improvements were explicitly out of scope for this issue.

Policy:
- Use of an external CDN means that Dataverse users are potentially leaking information to the CDN (IP addresses at least, potentially other browser-specific information).  More privacy concious users are more likely to be impacted (eg - private browsing session / incognito window starting with an empty cache, or users manually clearing their cache).

oscardssmith: This seems like a good thing to make configurable especially if easy. It would allow installations to choose CDN's near their likely consumers for a small perf increase.

pdurbin: Meh. It's more likely that someone's Dataverse installation will be down that one of these popular CDNs. @pameyer are you concerned enough about privacy that you wouldn't want to run this code? If you're fine with it, let's let someone else come along and say no to CDNs. I'm excited about the performance boost.

oscardssmith: The other thing is that if you were worried about the privacy, an installation could theoretically mock its own cdn or something.

pameyer: @pdurbin If I measure and it doesn't show significant (aka - order of magnitude) performance improvements from the CDN, I'd definitely roll the non-CDN resources into our fork.

Whether or not that's me being too concerned about user privacy or not is a different question.

@oscardssmith mocking the CDN would require forking the code (DNS resolution would be happening on the end user / browser side).

oscardssmith: Don't we want to support using different CDN's without forking? this seems important for non US installations, and it would be a silly reason for people to fork.

pdurbin: @pameyer ok. This reminds me of the conversation I was having with the Google Fonts guy. I was asking if it's possible to host the fonts yourself rather than getting them from a CDN and he kind of looked at me like I had three heads. For me it's about control more than privacy, I guess, but it's good to consider various angles.

@oscardssmith yeah. In case you weren't already aware, @pameyer (and many other people) already run a fork: https://github.com/IQSS/dataverse/issues/4784#issuecomment-400515577

PR #4807, opened 2018-07-03
4800-better_error_meesage_on_pid_provider_failure

qqmyers:
when DataCite/the PID provider is down (non-201 response), the original code was just
ignoring the issue. This change triggers the configured error message
from Bundle.properties that occurs if an exception is thrown from the
service call.
Also updated error to end with 'if the problem persists, please contact
<contact.support>' rather than 'Does issue continue to persist?'

## Related Issues

- connects to #4819

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -170,12 +171,14 @@ private void publicizeExternalIdentifier(CommandContext ctxt) throws CommandExce
         String protocol = theDataset.getProtocol();
         IdServiceBean idServiceBean = IdServiceBean.getBean(protocol, ctxt);
         try {
-            idServiceBean.publicizeIdentifier(theDataset);
+        	//A false return value indicates a failure in calling the service
+            if(!idServiceBean.publicizeIdentifier(theDataset)) throw new Throwable();
```
oscardssmith: Shouldn't you be throwing something more specific? At the very least, and `Exception`

qqmyers: The try/catch block catches Throwable and never looks at it's contents - it just creates a CommandException which is what is passed upward. I can change this to Exception (and the one on line 181), but I don't know if there is really a Throwable but non-Exception that can be thrown from something else to know if the catch can be changed.

oscardssmith: I wouldn't change the catch as you never know why it was put there.

Issue #4763, opened 2018-06-20
Separate secrets in openshift.json


danmcp:
openshift.json today uses hardcoded values for usernames and passwords.  Ex:

https://github.com/IQSS/dataverse/blob/develop/conf/openshift/openshift.json#L182

This is obviously a bad security practice.  OpenShift templates have the ability to generate values.  Ex: 

https://github.com/openshift/origin/blob/master/examples/sample-app/application-template-dockerbuild.json#L20

https://github.com/openshift/origin/blob/master/examples/sample-app/application-template-dockerbuild.json#L417

That's the easy part.  What is also necessary is to make the usernames/passwords and other secret information configurable to be passed (through the secrets) to the install script(s).

pdurbin:
@tkmonson thanks for making pull request #4809

I moved this issue to code review at https://waffle.io/IQSS/dataverse

pdurbin:
@tkmonson hi! We discussed this issue and pull request #4809 at standup this morning. It looks like @danmcp asked you to update `doc/sphinx-guides/source/developers/containers.rst` over at https://github.com/IQSS/dataverse/pull/4809#issuecomment-403124206 ("update the openshift section of the readme about how this affects the experience. Ex: How would I login to postgres on the command line from the glassfish container.") and I'm happy to help you if you have questions. At http://guides.dataverse.org/en/4.9.1/developers/documentation.html you can read about how we created the Dataverse guides using Sphinx. I guess I would suggest adding to the "Troubleshooting" section:

![screen shot 2018-07-09 at 11 43 31 am](https://user-images.githubusercontent.com/21006/42461018-7e407370-836d-11e8-85ef-aafebceb8a0c.png)

pdurbin:
@tkmonson thanks for adding those doc changes. I move this issue to QA.

pdurbin:
Oh, while doing QA we should bear in mind that @pameyer noted at https://github.com/IQSS/dataverse/pull/4809#issuecomment-403124241 that Solr wasn't working for him:

> "attempted to execute; new-app failed with solr pod deployment error. checked develop (cb98419); still seeing solr pod deployment error there as well."

pdurbin:
Pull request #4809 was merged prematurely and the change was backed out from the "develop" branch with pull request #4820.

New pull request #4827 was just made and contains the same commit I approved for QA in the old pull request so I'm moving this issue to QA.

@tkmonson if you could write a bit here about how to test, it would be much appreciated.

[...]

Issue #3799, opened 2017-04-26
install script should support non-interactive configuration of options


pameyer:
The existing installation script doesn't support non-interactive configuration of installation options (mail server, postgresql configuration, default contact email address, etc).  This causes friction with automated provisioning tools (or requires them to duplicate logic in the installation script).

Options:
 - installation script as a TODO for providing default values in a text file (possibly with hard-coded defaults if that files is missing); this could be implemented
 - command line arguments for these options could be implemented
 - values from environmental variables could be used, with default values being used if they aren't set

Potentially related to #3390

pameyer:
Bringing previous work on this in sync (mix of option 1 and 2): https://github.com/IQSS/dataverse/tree/3799-noninteractive_installer)

pdurbin:
@pameyer thanks for the pull request at #3803! I took a quick look and here's what I would suggest:

- Invoke the new functionality from Vagrant. Here's the line that you'd change: https://github.com/IQSS/dataverse/blob/v4.6.1/scripts/vagrant/install-dataverse.sh#L13 . This will build confidence in the fix before we move this issue to QA.
- Move the content of `scripts/installer/README-noninteractive.md` to the "develop" branch equivalent of https://github.com/IQSS/dataverse/blob/v4.6.1/doc/sphinx-guides/source/installation/installation-main.rst#running-the-dataverse-installer . This will make the improvements much more visible to the community.

Thanks!!

pdurbin:
I fixed the backticks in 03c0dd0 to be rst style instead of markdown style. This is in pull request #3803

@pameyer you're using `-f` ... are you piggypacking off `--force`? Here's `GetOptions`:

    "f|force" => \$force,

pameyer:
@pdurbin - thanks for the correction.  And it is using the same `--force` / `-f` that was already there.

pdurbin:
@pameyer ok. Thanks. Moving to QA at https://waffle.io/IQSS/dataverse . The pull request is #3803.

pdurbin:
Oh, wait. Not moving to QA yet because I think @landreev wants to take a look. I'm happy with the changes. @kcondon I didn't actually test `vagrant up` but that would be a good thing to test since it exercises the changes in the pull request.

pdurbin:
@landreev and I just talked about this and we're both happy with the pull request (#3803) so I'm moving this to QA. Again, I'd recommend testing `vagrant up`.

kcondon:
[...]

PR #4809, opened 2018-07-03
4763 separate secrets

tkmonson:
connects to #4763

### DIFF
--- a/conf/openshift/openshift.json
+++ b/conf/openshift/openshift.json
@@ -13,6 +13,59 @@
   },
   "objects": [
     {
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-admin-secret"
+      },
+      "stringData" : {
+        "admin-password" : "${ADMIN_PASSWORD}"
+      }
+    },
+  	{
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-postgres-secret"
+      },
+      "stringData" : {
+        "postgres-user" : "${POSTGRES_USER}",
+        "postgres-password" : "${POSTGRES_PASSWORD}"
+      }
+    },
+    {
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-postgresql-secret"
+      },
+      "stringData" : {
+        "postgresql-user" : "${POSTGRESQL_USER}",
+        "postgresql-password" : "${POSTGRESQL_PASSWORD}"
+      }
+    },
+    {
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-postgresql-master-secret"
+      },
+      "stringData" : {
+        "postgresql-master-user" : "${POSTGRESQL_MASTER_USER}",
+        "postgresql-master-password" : "${POSTGRESQL_MASTER_PASSWORD}"
+      }
+    },
+    {
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-postgresql-admin-secret"
+      },
+      "stringData" : {
+        "postgresql-admin-password" : "${POSTGRESQL_ADMIN_PASSWORD}"
+      }
+    },
+    {
       "kind": "Service",
       "apiVersion": "v1",
       "metadata": {
@@ -153,7 +206,7 @@
                     "protocol": "TCP"
                   }
                 ],
-                "resources": {"limits": {"memory": "1024Mi"
+                "resources": {"limits": {"memory": "2048Mi"
                   }
                   
                   },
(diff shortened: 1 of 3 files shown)


conf/openshift/openshift.json
```diff
@@ -12,6 +12,59 @@
...
+      }
+    },
+  	{
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "dataverse-postgres-secret"
+      },
+      "stringData" : {
+        "postgres-user" : "${POSTGRES_USER}",
```
pameyer: My initial thinking was that database user names might not be sensitive enough to be treated as secret; but considering it's password auth vs pub key auth (and low-implementation complexity, and already done) it seems reasonable to go with it as is.

scripts/installer/install
```diff
@@ -180,8 +180,14 @@ if ( -e $config_default_file )
 		if ( 2 == @xs )
 		{
 			my $k = $xs[0];
-			my $v = trim($xs[1]);
-			$CONFIG_DEFAULTS{$k}=$v;
+                        my $v = trim($xs[1]);
+                        if (defined $ENV{$k} && ($k eq "POSTGRES_USER" || $k eq "POSTGRES_PASSWORD")) {
```
pameyer: If I'm reading this correctly, this changes the behavior of the install script to read from environmental variables instead of a configuration file and command line arguments.  Part of the thinking for using a configuration file was to defer the discussion of environmental variables vs command line arguments vs configuration files.

Options:
- accept additional tech debt for current installer (aka - config file, command line options and now environmental variables)
- discuss technical options, and decide one option for passing non-interactive options to the installer, and rework existing options to that choice.

For this PR, I think it makes sense to go with the first option.  Other thoughts?

pdurbin: @pameyer this is exactly why I requested a code review from you. Is this going to break the ability to provide the config using a file?

pameyer: will need to test to confirm, but I don't believe it it will.

danmcp:
@tkmonson Please squash the commits and update the openshift section of the readme about how this affects the experience.  Ex: How would I login to postgres on the command line from the glassfish container.

pameyer:
attempted to execute; `new-app` failed with solr pod deployment error.  checked develop (cb984193d0daa6b915ed7ca8575cae3990083e8f); still seeing solr pod deployment error there as well.

scripts/installer/install
pdurbin: @pameyer I just assigned this issue to myself and will probably test it in Vagrant.

@tkmonson is it clear what we're talking about? We don't want to lose the feature that @pameyer added in #3799 which is a way to run the installer non-interactively. The pull request was #3803.

pdurbin: @pameyer I added a custom hostname in default.config (by temporarily editing install-dataverse.sh and I won't be checking in the change) and it seems to work fine. Please see below.

```
==> standalone: You should now have a running DVN instance at
==> standalone:   http://issue4763:8080
==> standalone: If vagrant up was successful (check output above) Dataverse is running on port 8080 of the Linux machine running within Vagrant, but this port has been forwarded to port 8888 of the computer you ran vagrant up on. For this reason you should go to http://localhost:8888 to see the Dataverse app running.
murphy:dataverse pdurbin$ 
murphy:dataverse pdurbin$ git diff
diff --git a/scripts/vagrant/install-dataverse.sh b/scripts/vagrant/install-dataverse.sh
index ac48217..0d4027c 100644
--- a/scripts/vagrant/install-dataverse.sh
+++ b/scripts/vagrant/install-dataverse.sh
@@ -19,7 +19,7 @@ if [ -e default.config ]; then
        mv default.config tmp-${pid}-default.config
 fi
 
-echo "HOST_DNS_ADDRESS localhost" > default.config
+echo "HOST_DNS_ADDRESS issue4763" > default.config
 echo "GLASSFISH_DIRECTORY      /home/glassfish/glassfish4" >> default.config
 
 if [ ! -z "$MAILSERVER" ]; then
murphy:dataverse pdurbin$ vagrant ssh
MAIL_SERVER environment variable not specified. Using localhost by default.
To specify it in bash: export MAIL_SERVER=localhost
[vagrant@standalone ~]$ sudo -i
[root@standalone ~]# grep issue4763 /home/glassfish/glassfish4/glassfish/domains/domain1/config/domain.xml
    <mail-resource host="localhost" from="do-not-reply@issue4763" user="dataversenotify" jndi-name="mail/notifyMailSession"></mail-resource>
        <jvm-options>-Ddataverse.fqdn=issue4763</jvm-options>
[root@standalone ~]# 
```

pdurbin:
> update the openshift section of the readme about how this affects the experience

@tkmonson the file that @danmcp is talking about is doc/sphinx-guides/source/developers/containers.rst

scripts/installer/install
tkmonson: @pdurbin This code still reads from the configuration file and writes the keys and values into the CONFIG_DEFAULTS hash. It just overrides the value in the config file if a corresponding environment variable has been defined (by openshift.json). So if a randomized user/password exists, its value is put into CONFIG_DEFAULTS instead of the default value.

The installer should be able to run non-interactively. That's how I used it in testing.

djbrooke:
Hey @tkmonson, this was merged prematurely. We still need to take this change through QA before merging to develop. We'll need to revert this. Are you available tomorrow morning to have someone from our team take you through the process?

Issue #4843, opened 2018-07-13
Regression: datasets can no longer be destroyed when using DataCite (rather than EZID)


pdurbin:
Pull request #4606 was just merged an a regression was introduced such that datasets can no longer be destroyed if you are using DataCite rather than EZID as your DOI provider.

I'm on commit 06b5e92

The dataset I created has a file:

![screen shot 2018-07-13 at 2 49 16 pm](https://user-images.githubusercontent.com/21006/42708666-f38f3c2e-86ab-11e8-89f5-80088570ee1b.png)

Here's the command to destroy a dataset (copied and pasted from #2593 but I wasted a little time checking the API Guide first):

`curl -H 'X-Dataverse-key: 461a6cf3-a91e-42ed-ab57-cb0f872d4801' -X DELETE http://localhost:8080/api/datasets/:persistentId/destroy?persistentId=doi:10.5072/FK2/JNADXB`

The resulting error:

`{"status":"ERROR","message":"Command edu.harvard.iq.dataverse.engine.command.impl.DestroyDatasetCommand@35765ee6 failed: null"}`

Note especially from the stacktrace below the error `A system exception occurred during an invocation on EJB DOIDataCiteRegisterService, method: public java.lang.String edu.harvard.iq.dataverse.DOIDataCiteRegisterService.registerIdentifier(java.lang.String,java.util.Map,edu.harvard.iq.dataverse.DvObject) throws java.io.IOException]]`

Unfortunately, this error is preventing the API test suite from passing on the phoenix server because it is using DataCite: #4840 

```
[2018-07-13T14:46:28.590-0400] [glassfish 4.1] [WARNING] [AS-EJB-00056] [javax.enterprise.ejb.container] [tid: _ThreadID=105 _ThreadName=http-listener-1(3)] [timeMillis: 1531507588590] [levelValue: 900] [[
  A system exception occurred during an invocation on EJB DOIDataCiteRegisterService, method: public java.lang.String edu.harvard.iq.dataverse.DOIDataCiteRegisterService.registerIdentifier(java.lang.String,java.util.Map,edu.harvard.iq.dataverse.DvObject) throws java.io.IOException]]

[2018-07-13T14:46:28.590-0400] [glassfish 4.1] [WARNING] [] [javax.enterprise.ejb.container] [tid: _ThreadID=105 _ThreadName=http-listener-1(3)] [timeMillis: 1531507588590] [levelValue: 900] [[
  
[...]

PR #4845, opened 2018-07-13
prevent NPE when destroying datasets using DataCite #4843

pdurbin:
connects to #4843


oscardssmith:
I can't find a better way of doing this, so approved.

Issue #4811, opened 2018-07-03
filter file pids with special characters


matthew-a-dunlap:
As part of #4761 we have been discussing pid structure and the challenges of external systems using these values. There is concern that we will start importing pids with a lot of weird characters, for example this valid DOI:

https://doi.org/10.1002/(sici)1099-1409(199908/10)3:6/7<672::aid-jpp192>3.0.co;2-8

These edge case pids could lead to systems breaking, injection attacks, etc. This is of more concern as we are trying to have all external systems use pids to reference files and datasets, and they are unlikely to expect such a variety of characters. Especially legacy systems that only expect the alphanumeric pids generated in Dataverse.

A solution that was discussed was to enforce limits on the characters we import. The general idea is to be pretty strict but simple and then add in edge cases as requested.

See here for more info on DataCite's pid. I did not read into Handle's or ezid, tho my limited understanding is that those allow a different/weird assortment of characters. https://support.datacite.org/docs/doi-basics https://www.crossref.org/blog/dois-and-matching-regular-expressions/

This was originally discussed in #3083 but we decided to break it out into its own story. We really shouldn't release #3083 without it tho. https://github.com/IQSS/dataverse/pull/4606#pullrequestreview-134118650

matthew-a-dunlap:
This was originally discussed in #3083 but we decided to break it out into its own story. We really shouldn't release #3083 without it tho. https://github.com/IQSS/dataverse/pull/4606#pullrequestreview-134118650

djbrooke:
Thanks @scolapasta and @matthew-a-dunlap for the discussion post standup. Moved to backlog for discussion.  Before we pick this up, we should determine the criteria for filtering.

pdurbin:
I just made pull request #4852 and moved this issue to code review.

scolapasta:
This is a good start, but only blocks certain characters and would need to be updated for others. Let's take the opposite approach and define what we allow and make sure PIDs parse that way. (and when some group says they need us to allow something else, we can consider adding it to what is allowable).

PR #4852, opened 2018-07-16
disallow some special characters in dataset DOI import #4811

pdurbin:
connects to #4811


src/main/java/edu/harvard/iq/dataverse/GlobalId.java
```diff
@@ -226,6 +229,11 @@ private boolean checkDOIAuthority(String doiAuthority){
         
         return true;
     }
-    
-    
+
+    public static boolean hasDisallowedImportCharacters(String pidParam) {
+        Pattern p = Pattern.compile(DISALLOWED_CHARACTERS);
```
pameyer: This brings up the "enumerating badness" vs "enumerating goodness" debate.

Would it be a big complexity increase to flip the regex?

matthew-a-dunlap: I think I agree that we may be better served with a whitelist instead of a blacklist, at least to start. But I can see potential pitfalls, especially with supporting other languages.

Specifically, I wonder if there should be some restriction on the number of "/"s in the identifier

src/main/java/edu/harvard/iq/dataverse/GlobalId.java
```diff
@@ -25,6 +27,8 @@
     public static final String HDL_PROTOCOL = "hdl";
     public static final String HDL_RESOLVER_URL = "https://hdl.handle.net/";
     public static final String DOI_RESOLVER_URL = "https://doi.org/";
+    public static final String DISALLOWED_CHARACTERS = ".*[<>()].*";
```
matthew-a-dunlap: This can safely be removed as we are just whitelisting

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -252,10 +252,13 @@ public Response importDataset( String jsonBody, @PathParam("identifier") String
             if ( version.getVersionState() == null ) {
                 version.setVersionState(DatasetVersion.VersionState.DRAFT);
             }
-            
-            if ( nonEmpty(pidParam) ) {
+
+            if (nonEmpty(pidParam)) {
+                if (GlobalId.verifyImportCharacters(pidParam)) {
+                    return badRequest("PID parameter contains characters not allowed. They must match this regex on import: " + GlobalId.ALLOWED_CHARACTERS);
```
matthew-a-dunlap: May be good to say in this "Not allowed by the Dataverse application" for clarity, as they are allowed by the provider.

pameyer: Other than that, this looks like an informative error message :)

src/main/java/edu/harvard/iq/dataverse/GlobalId.java
benjamin-martinez: Thought I took that out... but that's what code review is for, right?

matthew-a-dunlap: Totally! One more thing, we should probably put the blacklist string into the bundle.properties file. It helps a bit with later maintenance and customization

src/main/java/edu/harvard/iq/dataverse/GlobalId.java
```diff
@@ -25,6 +27,7 @@
     public static final String HDL_PROTOCOL = "hdl";
     public static final String HDL_RESOLVER_URL = "https://hdl.handle.net/";
     public static final String DOI_RESOLVER_URL = "https://doi.org/";
+    public static final String ALLOWED_CHARACTERS = "^[A-Za-z0-9._/:\\-]*$";
```
matthew-a-dunlap: Looking at this more, I think this should actually be moved to our bundle.properties file. It'll help with later maintenance to have this config-y attribute stored there. Sorry I didn't catch this the first time through!

Issue #4718, opened 2018-05-25
Validate setting for Authority


sekmiller:
After #3583 the installation setting for Authority in the setting table should not include a "Shoulder" ("/FK2" or "/DVN", etc.) or a "/".  There isn't any validation for this requirement. Failing to abide by it may lead to an invalid identifier error on dataset publication. The workaround is simply to make sure that every installation abides by this requirement. We could put a validation (via Bean Validation - or other method) to catch this before there are invalid identifier errors.

qqmyers:
Validating the setting would be good. FWIW: If it is set incorrectly now, I think a dataset could be created with a bad authority entry, but it should be caught at the first attempt to access the dataset page - the parsing logic will look for the first '/' and fail to find the dataset. A work-around if this happens would be to reset the :Authority and to run the same type of sql being used to update legacy entries:
UPDATE dvobjectSET identifier=substring(authority, strpos(authority,"/")+1) || "/" || identifier WHERE strpos(authority,"/")>0;
UPDATE dvobject SET authority=substring(authority from 0 for strpos(authority,"/")) WHERE strpos(authority,"/")>0;

sekmiller:
Right. The not finding the dataset page was the error we were seeing yesterday in testing before we updated the setting table. The hope here would be to try to avoid that.  (I have since updated the sql script to add the update of the Authority setting and the adding the new field of DoiShoulder.) Thanks for the update data scripts.

djbrooke:
We'll address this through a documentation change.

pdurbin:
I just made pull request #4853 and will move this to code review.

PR #4853, opened 2018-07-16
no slash ("/") allowed in :Authority #4718

pdurbin:
connects to #4718


matthew-a-dunlap:
I ran this by @dlmurphy and he was ok with it as well!

Issue #4822, opened 2018-07-10
Target URLs for deleted files do not resolve properly


sekmiller:
Every month we receive a list from EZID of target urls that do not resolve. We just got the first one after implementing file DOIs. It includes a bunch of File DOIs. A spot check reveals that the files had been part of published datasets but have been removed from the current version of the dataset.  They still exist as files and  we want the DOI to remain. At first glance it seems that the solution would be to make sure that the original target url (https://dataverse.harvard.edu/file.xhtml?persistentId=doi:10.7910/DVN/Y5CXM8/YZZYPZ) resolves or update the target url so that it includes the version of the last version that the file was active in (https://dataverse.harvard.edu/file.xhtml?persistentId=doi:10.7910/DVN/Y5CXM8/YZZYPZ&version=1.4)

sekmiller:
By consensus, we want to leave the target url alone and have it resolve to the latest version the file is active in.

pdurbin:
I just created pull request #4858 and moved it to code review at https://waffle.io/IQSS/dataverse

PR #4858, opened 2018-07-17
redirect PID for deleted file to latest version it existed in #4822

pdurbin:
connects to #4822


sekmiller:
Looks good, generally. Would it make sense to put the FindMostRecentVersionFileIsIn in the DataFileServiceBean where it might be of use elsewhere?

pdurbin:
@sekmiller good idea. Fixed in c896079

Issue #4656, opened 2018-05-09
Dataset Files DataTable - Paginator, Selection, Download, Counter


mheppler:
Based on feedback from @scolapasta @TaniaSchlatter and @djbrooke on the issue _Spike: Files DataTable, Selection, Paginator vs Scroll #4437_, creating a new issue for development.

Revise the dataset files table to improve UX of the table with improvements to paginator, selection, counter, download and potentially sort features of the dataset pg files table. Bug and feature request issues have been consolidated and included in _Dataset - File UI Improvements #3404_.

If there is anything we learn about performance of this files table, we should add it to this issue, _Performance degradation investigation #3631_.

Defined scope:
- Swap out dataTable liveScroll attribute and instead use the Paginator (out of box paginator will need CSS hacking to look more Bootstrapy)
- Dynamic file counter that shows how many files are being displayed in the table, in addition to how many files total, and how many files are selected
- Text links for selecting all and clearing selections displayed along with the dynamic count of how many files are currently selected
- Review how the current Download button in the dataTable header works with relation to the "select all" hack (#4051) added to the checkbox in the dataTable header

![files - dataset - depositor jpeg](https://user-images.githubusercontent.com/687227/39823277-be3ac73c-537a-11e8-90ba-0351b9f00c48.png)

mheppler:
Regarding performance, there can be front-end refactoring improvements to optimize the XHTML, towards small gains in render and load time.

For example, in `filesFragment.xhtml` include file for the dataTable, there are 16 references to the `DatasetPage.editMode` render logic, but in `dataset.xhtml` there is render logic for `empty DatasetPage.editMode` that does not display the tab containing the datatable in the include.

```
<p:tab id="dataFilesTab" title="#{bundle.files}" rendered="#{(!DatasetPage.workingVersion.deaccessioned or 
                                                              (DatasetPage.workingVersion.deaccessioned and DatasetPage.canUpdateDataset()))  and
                                                              (empty DatasetPage.editMode)}">
                                <ui:include src="filesFragment.xhtml">
                                    <ui:param name="fileDownloadHelper" value="#{DatasetPage.fileDownloadHelper}"/>
                                </ui:include>
</p:tab>
```

djbrooke:
[...]

PR #4881, opened 2018-07-23
4656 filetable paginator

pdurbin:
connects to #4656


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -4283,4 +4296,29 @@ public String getJsonLd() {
         }
         return "";
     }
+
+    public void selectAllFiles() {
+        logger.info("selectAllFiles called");
+        List<FileMetadata> allFiles = new ArrayList<>();
+        for (FileMetadata fm : workingVersion.getFileMetadatas()) {
```
oscardssmith: why not just `selectedFiles = workingVersion.getFileMetadatas();`

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -4283,4 +4296,29 @@ public String getJsonLd() {
...
+        selectedFiles = Collections.EMPTY_LIST;
+    }
+    
+    public void fileListingPaginatorListener(PageEvent event) {
+        setFilePaginatorPage(event.getPage());      
+    }
+    
+    public void refreshPaginator() {
+        FacesContext facesContext = FacesContext.getCurrentInstance();
+        org.primefaces.component.datatable.DataTable dt = (org.primefaces.component.datatable.DataTable) facesContext.getViewRoot().findComponent("datasetForm:tabView:filesTable");
```
oscardssmith: importing `DataTable` would make this line much clearer. ie. `DataTable dt = (DataTable) facesContext.getViewRoot().findComponent("datasetForm:tabView:filesTable");`

sekmiller: The problem here is that there's already a DataTable object defined in dataverse so the ugliness is in line is related to making the distinction with the native DataTable object. If there's a cleaner way to make the distinction...

oscardssmith: Oh yeah, that sucks. You could import up to `component`. It's not ideal, but I think better.

oscardssmith: how I wish for python's `import x as y`

pdurbin: This isn't worth changing.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
pdurbin: @oscardssmith good idea. Fixed in c592236.

Issue #4882, opened 2018-07-23
Better error handling 


pameyer:
Having the dataset persistent ID as a GET parameter to the dataset page can result in 500 errors when $pid is interpretable by postgres.

pdurbin:
@pameyer if you're curious about how we remove `.xhtml` from URLs, see dfe0488 for an example from src/main/webapp/WEB-INF/pretty-config.xml

PIDs are especially challenging because they have slashes in them. When we added the ability to target a dataset using a PID rather than a database id in #1837 we used a (GET) query parameter.

pdurbin:
I just created pull request #4898 and moved this issue to code review at https://waffle.io/IQSS/dataverse

kcondon:
Under the test scenario I was given, this issue still occurs. Passing back to Dev. See me or pete for test scenario.

PR #4898, opened 2018-07-26
better error handling #4882

pdurbin:
connects to #4882


sekmiller:
Looks good. Moved to QA.

PR #4907, opened 2018-07-30
4703 merge dual

matthew-a-dunlap:
- connects to

## Pull Request Checklist

### DIFF
--- a/conf/docker-aio/install.bash
+++ b/conf/docker-aio/install.bash
@@ -3,6 +3,7 @@ sudo -u postgres createuser --superuser dvnapp
 #./entrypoint.bash &
 unzip dvinstall.zip
 cd dvinstall/
+echo "beginning installer"
 ./install -admin_email=dvAdmin@mailinator.com -y -f > install.out 2> install.err
 
 echo "installer complete"
--- a/conf/docker-dcm/readme.txt
+++ b/conf/docker-dcm/readme.txt
@@ -2,7 +2,7 @@ This docker-compose setup is intended for use in development, small scale evalua
 
 Setup:
 
-- build docker-aio image with name dv0 as described in ../docker-aio` (don't run setupIT.bash)
+- build docker-aio image with name dv0 as described in ../docker-aio` (don't start up the docker image or run setupIT.bash)
 - work in the `conf/docker-dcm` directory for below commands
 - download/prepare dependencies: `./0prep.sh`
 - build dcm/dv0dcm images with docker-compose: `docker-compose -f docker-compose.yml build`


src/main/java/edu/harvard/iq/dataverse/S3PackageImporter.java
```diff
@@ -0,0 +1,216 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
oscardssmith: Are we sure this code shouldn't go in StorageIO? This feels like the type of communication that we should be using it for.

src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
```diff
@@ -150,6 +150,10 @@ public boolean isRsyncDownload() {
         return systemConfig.isRsyncDownload();
     }
     
+    public boolean isHTTPUpload(){
```
oscardssmith: Should you use the `SettingsWrapper` cache for this?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1161,6 +1231,17 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
             return wr.getResponse();           
         }
         
+        //------------------------------------
+        // (2a) Make sure dataset does not have package file
+        //
+        // --------------------------------------
+        
+        if (dataset.getEditVersion().isHasPackageFile()){
```
oscardssmith: isHas?

src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
matthew-a-dunlap: That bit of code is from dual mode. I closed this pr because the comparison isn't accurate.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
matthew-a-dunlap: That bit of code is from dual mode. I closed this pr because the comparison isn't accurate.

src/main/java/edu/harvard/iq/dataverse/S3PackageImporter.java
matthew-a-dunlap: I kept it separate because the file mode code also keeps it separate. I thought that was also the better approach from the bits of discussion I had with others on the team. I would be up for moving it though if it seems right. I'm curious what @landreev @scolapasta think.

fyi this pr isn't representative of the actual changes, because it includes the dual mode code that has yet to be merged.

Issue #4865, opened 2018-07-17
File Reingest API


djbrooke:
We've added support for new versions of Stata in #2301. Following the release of the code for #2301, we should reingest the files in production that had previously failed ingest due to being an unsupported Stata version. An API endpoint instead of a db script seems the best way to go for community and reusability purposes. 

This should not be a version change and it's OK if the citation gets a UNF added or changed as a result of the reingest. 

@landreev you mentioned you wanted to put in a few notes as well

landreev:
Status update: 
I checked in the initial implementation from last week. 
It's working but I realized I need to add a little more before I consider it done: 
- I'm adding API tests; 
- I need to add some extra code for re-checking the mime type of the data file; for example, the uningested Stata v.14 files in the prod. db are listed as application/x-stata; but they need to be re-identified as application/x-stata-1[345] in order to be associated with the "new stata" ingest plugin. 
- The current system of running tab. ingest asynchronously hides all the status info from the caller. For example, if you try to ingest a file, but we skip it because of the size, this status info will only be available in the server log. Since this is the decision we make early on, before we attempt anything else, it should be easy to communicate this back to the API (or page user), and inform the user. 

Also, I didn't like using the term "reingest" for this. Strictly speaking, "reingest" would be attempting to ingest something as tabular data AGAIN; i.e. running ingest on something that's not yet a tabular file. (I see a use case for that too - for example, if an updated ingest plugin fixes some ingest issues that need to be corrected...). So I would save the word "reingest" for that.  I called this api call /api/files/{id}/ingestAsTabular
If anyone has a better name in mind/has strong opinions on this, please let me know.

oscardssmith:
Why wouldn't both of these cases be the same API? Maybe it should be called `tryIngest` or something with the behavior that if it is ingested, it will drop the data and re-ingest, and if not ingested it will try to complete an ingest.

landreev:
Yeah, I decided to use the same API for both. And to call it "reingest" after all. 
It will require a "force" option, in order to run it on a datafile that's already been ingested as tabular.

landreev:
For QA: An easy way to upload stata, etc. files without ingesting is tabular is to set the limit (:TabularIngestSizeLimit) to something very low. 
Then you can reset the limit, and test the API: 
```
/api/files/{FILEID}/reingest
```
You have to be superuser.

oscardssmith:
[...]

PR #4914, opened 2018-08-01
4865 file reingest api

landreev:
## New Contributors

## Related Issues

- connects to #4865: File Reingest API 

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -503,6 +503,89 @@ public int compare(DataFile d1, DataFile d2) {
         }
     }
     
+    public String startIngestJobForSingleFile(DataFile dataFile, AuthenticatedUser user) {
```
oscardssmith: Now that we have this function, should we simplify `startIngestJobs` to be

    public void startIngestJobs(Dataset dataset, AuthenticatedUser user) {
        for (DataFile dataFile : dataset.getFiles()) {
            if (dataFile.isIngestScheduled()) {
               startIngestJobForSingleFile(dataFile, user);
            }
        }
    }

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -86,6 +86,16 @@
...
     private static final String[] TABULAR_DATA_FORMAT_SET = {"POR", "SAV", "DTA", "RDA"};
+    // The list of formats for which we need to re-test, if we want to attempt 
+    // to ingest the file again:
+    // Example: We have added support for Stata-13+; the new ingest plugin is called
+    // when the file type is detected as "application/x-stata-1[345]". But the 
+    // previously uploaded, but not ingested stata files are in the database
+    // typed as "application/x-stata". So we want to run the new-and-improved
+    // DTA check that will re-identify the specific DTA flavor. 
+    // If similar cases are introduced in the future, the affected formats will
+    // need to be added to the list.
```
oscardssmith: Why not just retest all? it's not expensive (I think).

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
landreev: Actually, no - that would result in a different behavior. When a user uploads 10 ingestable files, startIngestJobs will schedule the ingest as one queued JMS job, for the entire batch of 10 files. We want to keep the dataset locked for as long as the entire batch is being processed. With 10 individual queued jobs this would be harder to control. The UNF recalculation would be slightly trickier too; there may be other potential issues. 

At the same time, these 2 methods do have more duplicated code than strictly necessary. I could definitely get away with using the old one (startIngestJobs) for  both ingest and reingest, with just slight modifications. I just felt like writing a method from scratch, for some reason.

As a reviewer, if you feel this is ugly/potentially confusing to a future developer, I'm willing to quickly do that and merge these 2 methods into one.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
landreev: OK, I think you're right on this. It could actually be useful to retest everything; in case the type detection code has been improved since the last ingest attempt.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -328,6 +332,91 @@ public Response uningestDatafile(@PathParam("id") String id) {
...
+            }
+        } catch (WrappedResponse wr) {
+            return wr.getResponse();
+        }
+        
+        DataFile dataFile;
+        try {
+            dataFile = findDataFileOrDie(id);
+        } catch (WrappedResponse ex) {
+            dataFile = null;
```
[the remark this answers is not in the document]
landreev: That was in case findDataFileOrDie() returns null, without throwing an exception. 
Which I guess it should never be the case. As the name of the method suggests. :)

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
oscardssmith: If they were both nice small functions, I'd say don't bother, but this adds another ~100 lines of duplicate complicated code, so I think it's probably a good idea.

landreev: OK, done. There's now a startIngestJobs(); and an extra small startIngestJobsForDataset(), that doesn't duplicate any code.

Issue #4944, opened 2018-08-10
PermissionServiceBean Cleanup 


michbarsinai:
`PermissionServiceBean` has many deprecated methods, especially those that calculate permissions without taking IP groups into account (a.k.a "static permissions").
This issue will:

- [x] Remove the static permission code
- [x] Update code that uses static permissions to use the dynamic permission code
- [x] Update the developer guide, so it's easier for developers to understand how to query permissions in Dataverse
- [x] Perform general cleanup on the bean itself, and the related RoleAssignee area

⚠️ This issue was spun off from #2122.

djbrooke:
@michbarsinai - I moved this back to Community Dev for now. When you have time, please respond to @oscardssmith's feedback and move it back to code review.

michbarsinai:
@djbrooke done

pameyer:
Took a quick look at 406fb41a81c2ac42fb2cb4613235a9b367a3647b after hearing @kcondon and @landreev discussing deployment issues (of course, they were talking about 2122-querying-root-dataverse-contents).  No issues building or deploying to docker-aio; but integration tests failed (`[ERROR] Tests run: 67, Failures: 24, Errors: 17, Skipped: 0`).

djbrooke:
Hi @michbarsinai - I moved this back to Community Dev for now, as there are still a few items to work on. If you have the time to work on this, great! If not, let me know and we can get together some resources on this side.

pdurbin:
@michbarsinai are you still interested in this?

michbarsinai:
Not in any effective way, I'm afraid. Plus my knowledge is probably outdated on this by now. Happy to consult anyone implementing this, though.

pdurbin:
@michbarsinai no worries, we'll cook up a fresh issue whenever you want one. Closing! And yes let's consult about something! We miss you! ❤️

michbarsinai:
I Miss you too, good people!❤️

PR #4998, opened 2018-08-24
4944 permission service bean cleanup

michbarsinai:
## Related Issues

- connects to #4944 PermissionServiceBean Cleanup

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1521,9 +1517,9 @@ private String init(boolean initFull) {
                             setHasRsyncScript(false);
                         }
                     } catch (RuntimeException ex) {
-                        logger.warning("Problem getting rsync script: " + ex.getLocalizedMessage());
+                        logger.warning(()->"Problem getting rsync script: " + ex.getLocalizedMessage());
                     } catch (CommandException cex) {
-                        logger.warning("Problem getting rsync script (Command Exception): " + cex.getLocalizedMessage());
+                        logger.warning(()->"Problem getting rsync script (Command Exception): " + cex.getLocalizedMessage());
```
oscardssmith: What do these changes do? It looks like it just creates clutter by adding anonymous functions. Is there a benefit?

src/main/java/edu/harvard/iq/dataverse/MapLayerMetadataServiceBean.java
```diff
@@ -109,14 +106,14 @@ public MapLayerMetadata findMetadataByDatafile(DataFile datafile){
         First check if the given user has permission to edit this data.
         
     */
-    public boolean deleteMapLayerMetadataObject(MapLayerMetadata mapLayerMetadata, User user){
+    public boolean deleteMapLayerMetadataObject(MapLayerMetadata mapLayerMetadata, DataverseRequest req){
```
oscardssmith: Should this use `@notnull`? It will create compile time warnings if someone tries to pass null in.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -50,347 +53,392 @@
...
+
+    /**
+     * A set of permissions that con only be granted to {@link AuthentictedUser}s.
+     */
     private static final Set<Permission> PERMISSIONS_FOR_AUTHENTICATED_USERS_ONLY = 
-           EnumSet.copyOf(Arrays.asList(Permission.values()).stream()
-                    .filter( Permission::requiresAuthenticatedUser )
-                    .collect( Collectors.toList() ));
-    
+        EnumSet.copyOf(Arrays.asList(Permission.values()).stream()
```
oscardssmith: Why copy?

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -50,347 +53,392 @@
...
+     * create detailed reports on which permissions are missing. Probably not
+     * useful beyond that area.
+     */
+    interface InsufficientPermissionCallback {
+        void insufficientPermissions( DvObject dvo, Set<Permission> required, Set<Permission> granted);
+    }
+    
+    /**
+     * A request-level permission query (e.g includes IP groups). These queries
+     * form a DSL for querying permissions, in a way that's hopefully more
```
oscardssmith: What's a DSL?

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -50,347 +53,392 @@
...
      */
-    public class StaticPermissionQuery {
-
-        final DvObject subject;
-        final RoleAssignee user;
-
-        private StaticPermissionQuery(RoleAssignee user, DvObject subject) {
-            this.subject = subject;
-            this.user = user;
+    public RequestPermissionQuery on(DvObject d) {
```
oscardssmith: Use `@notnull` here

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -50,347 +53,392 @@
...
+    
+    /**
+     * Internal call; checks permissions, allows a callback for fine-grained
+     * response if case there are not enough permissions.
+     * 
+     * @param aCommand The command to be tested/
+     * @param ipc Callback, called if there are insufficient permissions for the passed command.
+     * @return {@code true} iff the command will not be blocked due to permission issues.
+     */
+    boolean isPermitted(Command aCommand, InsufficientPermissionCallback ipc) {
```
oscardssmith: Does anything use this anymore? If not, we should remove.

src/main/java/edu/harvard/iq/dataverse/authorization/groups/GroupProvider.java
```diff
@@ -46,6 +38,18 @@
      */
     public Set<T> groupsFor( RoleAssignee ra, DvObject dvo );
```
oscardssmith: Can This go away now?

src/main/java/edu/harvard/iq/dataverse/authorization/groups/GroupProvider.java
```diff
@@ -46,6 +38,18 @@
...
     
+    /**
+     * Looks up the groups this provider has for a dataverse request, in the context of a {@link DvObject}.
+     * @param req The request whose group memberships we evaluate.
+     * @param dvo the DvObject which is the context for the groups. May be {@code null}.
+     * @return The set of groups the user is member of.
+     */
+    public Set<T> groupsFor( DataverseRequest req, DvObject dvo );
+    
+    public Set<T> groupsFor( RoleAssignee ra);
```
oscardssmith: And This?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
michbarsinai: There's a huge benefit in not creating the string unless it's actually going to be logged. Note that creating strings requires O(n) time and contiguous space, which means that the memory management and garbage collector have to work harder.
NetBeans actually displays a warning and offers an auto-fix when you use the previous version.

src/main/java/edu/harvard/iq/dataverse/MapLayerMetadataServiceBean.java
michbarsinai: The change I made in this method was replacing the user with the request, so that it uses the updated permission check system.

Adding `@notNull` would change the method behavior, so I suggest not adding it now, as it is not in scope, and would require additional testing (Also, the `null` -> `false` semantics might make sense). 

There are other problems in this method (e.g. ignored result on line 121), but, again, not in scope for issue.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
michbarsinai: Because we want the actual `Set` object to be an `EnumSet`, not just any set. This means set operations are done over bit of a `long` value, rather than over objects.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
michbarsinai: Domain Specific Language. Updated the docs.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
michbarsinai: added

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
michbarsinai: Yes, used twice (engine and internally in the bean).

src/main/java/edu/harvard/iq/dataverse/authorization/groups/GroupProvider.java
michbarsinai: No. It's in use by the GroupServiceBean.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -184,7 +187,7 @@ public RequestPermissionQuery on(DvObject dvo) {
      * @param d the object on which the permissions will be queried
      * @return A permission query
      */
-    public RequestPermissionQuery on(DvObject d) {
+    public RequestPermissionQuery on( @NotNull DvObject d) {
         if (d == null) {
```
oscardssmith: `d` isn't null. that's why to `NotNull` it

michbarsinai: Ideally, yes. But as far as I can tell, you still need to do runtime validation in regular Java (i.e. not using Lombok et al). See https://blogs.oracle.com/java-platform-group/java-8s-new-type-annotations (scroll down to the section "Optional Type Annotations are not a substitute for runtime validation").

It may generate compiler warnings (https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html), but as long as there are no errors from standard `javac`, we have to do the check.

That said, I'll be very happy to see an official document saying we can remove that check :-)

oscardssmith: It does say that "Your code should still perform runtime validation", but on the other hand, maven and any ide will not allow it to compile, at which point I feel like we should not have runtime checking for something we already check at compile time. It would be like testing isinstance every time we use generics because java's generics are broken

michbarsinai: Dunno. This compiles on my machine with warning only:

    public static void main(String[] args) {
        new PermissionServiceBean().on(null);
    }

(leaving the Java generics part for the next time we meet 😄)

pdurbin:
@michbarsinai there are a fair number of merge conflicts.

pdurbin:
This pull request is over a year old (currently our oldest pull request), has lots of merge conflicts, and hasn't seen any activity lately. Closing.

Issue #5044, opened 2018-09-11
Workflow Enhancements


qqmyers:
As part of #4706, I needed the workflow mechanism to allow read-access to some settings and to enable internal workflows to act on behalf of a user (i.e. via access to their apiToken). In addition, I ran into some issue in which post-publication workflows only had access to the state of a dataset that did not include the publication updates (e.g. a change of the latest version to 'released') and, when resolving that, to some inconsistencies in adding and removing locks.

pdurbin:
@qqmyers I moved this over to code review based on what you wrote at https://github.com/IQSS/dataverse/pull/5048#issuecomment-426732955 but if you don't mind I'm curious about what sort of workflows you're imagining. "As a curator, I want to..."

qqmyers:
@pdurbin - this is part of the DPN submission of bagit bags #4706  - I needed to update the workflow mechanism itself to be able to get additional settings, for a post-publication workflow to be able to access the state of the dataset being published and further update it, etc.  As with the sha256/512 PR before, the changes here are used for/were motivated by DPN submission but are stand-alone technically.

pdurbin:
@qqmyers perfect. And #4706 has tons of detail, especially the wiki page your wrote, on what DPN submission is about. Thanks!

kcondon:
@qqmyers Do you mind updating this branch from /develop? I can't build/deploy due to old version. I've just verified rsync workflow/engine with existing code and am ready to verify/merge this branch. Thanks!

kcondon:
@qqmyers Thanks!

qqmyers:
@kcondon - After an error in which eclipse thought it merged with develop but left 100+ files out of synch, I've gone ahead and added a couple 'repair' commits to 'recover'. At this point, it should be merged except that it appears a few files have permission differences (no difference in source).  If you can fix those, this can go ahead. If you prefer, I can create a new branch.

pdurbin:
@qqmyers I just took at look at pull request #5048 after @kcondon mentioned your comment above to me. It does seem to have weird "chmod" stuff going on, so we'd lose the ability to execute scripts. If it's not too much trouble, a fresh pull request would be appreciated!

qqmyers:
https://github.com/IQSS/dataverse/pull/5224 is the cleaned up PR

kcondon:
[...]

PR #5048, opened 2018-09-11
Workflow enhancements

qqmyers:
## New Contributors

## Related Issues

- connects to #5044- Workflow Enhancements

## Pull Request Checklist


michbarsinai:
All in all looks good. I didn't get two things:

* Why the direct lock usage? I know JPA is a pain this setting (it was when writing the original version, at least), but the solution of directly playing around with locks will fire back once the lock mechanism change. Is it possible to use the system? If the problems are JPA-related, maybe even instantiate the involved service beans and pass the entity manager directly?
* Why the removal of the DOI provider?

qqmyers:
The direct lock usage is the way it was. The only thing I changed was to the way the lock is removed since the dataset is no longer find()-ed from the database during the refresh() call and hence the prior remove operation was failing. Overall, the EE/JPA aspects of this are not something I'm overly familiar with and I didn't want to change more than I needed to that might also affect the original workflow design (each step in its own transaction, etc.). However, I don't think I tried reverting to use the addlock command (that was already commented out in the code and perhaps wasn't working due to the way the dataset was being passed and find()-ed before), so maybe that's worth a test - to do both lock and unlock through the commands.

As for removing the doiprovider - to start, I could find anywhere that it was actually being used. Beyond that, once I added the means to get any setting, I figured that would be a better mechanism for any workflow that did require the doiprovider to get its value.

scolapasta:
@qqmyers I'm going to assign back to you and move back to Community Dev. If you could test the using of commands for the locks, that would be great. If you are don't have the time, no worries, we can live with it as is (since it was already that way), but I think it makes sense to try now if you can.

michbarsinai:
Sorry for that late reply. I’m agree with Gustavo - that would be a big plus. Other than that, the PR is good to go.

qqmyers:
@michbarsinai, @scolapasta - I've tried flipping back to using the add and remove lock commands and I get org.eclipse.persistence.exceptions.OptimisticLockExceptions when I do. My potentially incorrect understanding here is that, since we're in the middle of a workflow transaction, calling the lock commands, which require new transactions, cause an issue when they try to update the dataset object (I've tried with the in-memory object as well as doing a find by id to get a new dataset object from the db). By only writing a lock to the database and not updating the dataset object involved, and/or not creating new transactions, the local lock/unlock commands get around this while still correctly setting/removing the lock. (I have not run an external workflow, but did test commenting out the unlock command and verifying that when my local workflow finishes, the lock is in the database, which is what should happen when an external workflow happens. (For a local workflow, the lock is removed when the workflow completes - external ones only get to that point once the callback has occurred.)
So - I don't have any good ideas about how the lock commands might be adapted at this point. Unless the above description gives one of you some ideas, I think the current PR should work for all the workflow types.

pdurbin:
Closing in favor of #5224 which doesn't change the permissions (chmod) of some files.

PR #5047, opened 2018-09-11
IQSS/5043 streaming and new exporters

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Thanks!

## Related Issues

- connects to #5043 - Streaming and new exporters

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: https://github.com/IQSS/dataverse/tree/develop/scripts/database/upgrades
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #5035, opened 2018-09-07
SHA-256 and SHA-512 support

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Thanks!

## Related Issues

- connects to #5033 - Support for SHA-256 and SHA-512

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: https://github.com/IQSS/dataverse/tree/develop/scripts/database/upgrades
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #5049, opened 2018-09-11
Iqss/4706 dpn submission of archival copies

qqmyers:
## New Contributors

## Related Issues

- connects to #4706 - DPN Submission of archival Bags

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -88,6 +88,7 @@ import edu.harvard.iq.dataverse.engine.command.impl.RequestRsyncScriptCommand;
 import edu.harvard.iq.dataverse.engine.command.impl.PublishDatasetResult;
 import edu.harvard.iq.dataverse.engine.command.impl.RestrictFileCommand;
 import edu.harvard.iq.dataverse.engine.command.impl.ReturnDatasetToAuthorCommand;
+import edu.harvard.iq.dataverse.engine.command.impl.SubmitToArchiveCommand;
 import edu.harvard.iq.dataverse.engine.command.impl.SubmitDatasetForReviewCommand;
 import edu.harvard.iq.dataverse.externaltools.ExternalTool;
 import edu.harvard.iq.dataverse.externaltools.ExternalToolServiceBean;
@@ -4308,7 +4309,7 @@ public class DatasetPage implements java.io.Serializable {
 
     public void clearSelection() {
         logger.info("clearSelection called");
-        selectedFiles = Collections.EMPTY_LIST;
+        selectedFiles = Collections.emptyList();
     }
     
     public void fileListingPaginatorListener(PageEvent event) {       
@@ -4321,4 +4322,34 @@ public class DatasetPage implements java.io.Serializable {
         setFilePaginatorPage(dt.getPage());      
         setRowsPerPage(dt.getRowsToRender());
     }  
+    
+    public void archiveVersion(Long id) {
+        if (session.getUser() instanceof AuthenticatedUser) {
+            AuthenticatedUser au = ((AuthenticatedUser) session.getUser());
+            if (au.isSuperuser()) {
+                DatasetVersion dv = datasetVersionService.retrieveDatasetVersionByVersionId(id).getDatasetVersion();
+                SubmitToArchiveCommand cmd = new SubmitToArchiveCommand(dvRequestService.getDataverseRequest(), dv);
+                try {
+                    DatasetVersion version = commandEngine.submit(cmd);
+                    logger.info("Archived to " + version.getArchivalCopyLocation());
+                    if(version.getArchivalCopyLocation()!=null) {
+                    	resetVersionTabList();
+                    	this.setVersionTabListForPostLoad(getVersionTabList());
+                    JsfHelper.addSuccessMessage(BundleUtil.getStringFromBundle("datasetversion.archive.success"));
+                    } else {
+                        JsfHelper.addErrorMessage(BundleUtil.getStringFromBundle("datasetversion.archive.failure"));
+                    }
+                } catch (CommandException ex) {
+                    logger.log(Level.SEVERE, "Unexpected Exception calling  submit archive command", ex);
+                    JsfHelper.addErrorMessage(BundleUtil.getStringFromBundle("datasetversion.archive.failure"));
+                }
+            }
+        } else {
+            logger.warning("Non-superuser calling archiveVersion()");
+            // Shouldn't happen since button only shows for superuser
+            FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "Authentication error",
+                    "Contact an administrator");
+            FacesContext.getCurrentInstance().addMessage(null, message);
+        }
+    }
 }
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
@@ -1,0 +1,249 @@
+package edu.harvard.iq.dataverse.engine.command.impl;
+
+import edu.harvard.iq.dataverse.DOIDataCiteRegisterService;
+import edu.harvard.iq.dataverse.DataCitation;
+import edu.harvard.iq.dataverse.Dataset;
+import edu.harvard.iq.dataverse.DatasetVersion;
+import edu.harvard.iq.dataverse.DvObject;
+import edu.harvard.iq.dataverse.DatasetLock.Reason;
+import edu.harvard.iq.dataverse.authorization.Permission;
+import edu.harvard.iq.dataverse.authorization.users.ApiToken;
+import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
+import edu.harvard.iq.dataverse.engine.command.Command;
+import edu.harvard.iq.dataverse.engine.command.CommandContext;
+import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
+import edu.harvard.iq.dataverse.engine.command.RequiredPermissions;
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
+import edu.harvard.iq.dataverse.util.bagit.BagGenerator;
+import edu.harvard.iq.dataverse.util.bagit.OREMap;
+import edu.harvard.iq.dataverse.workflow.step.Failure;
+import edu.harvard.iq.dataverse.workflow.step.WorkflowStepResult;
+
+import static edu.harvard.iq.dataverse.engine.command.CommandHelper.CH;
+
+import java.io.IOException;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.nio.charset.Charset;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.io.IOUtils;
+import org.duracloud.client.ContentStore;
+import org.duracloud.client.ContentStoreManager;
+import org.duracloud.client.ContentStoreManagerImpl;
+import org.duracloud.common.model.Credential;
+import org.duracloud.error.ContentStoreException;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+@RequiredPermissions(Permission.ArchiveDatasetVersion)
+public class SubmitToArchiveCommand implements Command<DatasetVersion> {
+
+    private final DatasetVersion version;
+    private final DataverseRequest request;
+    private final Map<String, DvObject> affectedDvObjects;
+    private static final Logger logger = Logger.getLogger(SubmitToArchiveCommand.class.getName());
+    private static final String DEFAULT_PORT = "443";
+    private static final String DEFAULT_CONTEXT = "durastore";
+
+    public SubmitToArchiveCommand(DataverseRequest aRequest, DatasetVersion version) {
+        this.request = aRequest;
+        this.version = version;
+        affectedDvObjects = new HashMap<>();
+        affectedDvObjects.put("", version.getDataset());
+    }
+
+    @Override
+    public DatasetVersion execute(CommandContext ctxt) throws CommandException {
+        String host = ctxt.settings().getValueForKey(SettingsServiceBean.Key.DuraCloudHost);
+        String port = ctxt.settings().getValueForKey(SettingsServiceBean.Key.DuraCloudPort);
+        String dpnContext = ctxt.settings().getValueForKey(SettingsServiceBean.Key.DuraCloudContext);
+        AuthenticatedUser user = request.getAuthenticatedUser();
+        ApiToken token = ctxt.authentication().findApiTokenByUser(user);
+        if ((token == null) || (token.getExpireTime().before(new Date()))) {
+            token = ctxt.authentication().generateApiTokenForUser(user);
+        }
+        performDPNSubmission(version, user, host, port, dpnContext, token);
+        return ctxt.em().merge(version);
+    }
+
+    @Override
+    public Map<String, DvObject> getAffectedDvObjects() {
+
+        return affectedDvObjects;
+    }
+
+    @Override
+    public DataverseRequest getRequest() {
+        return request;
+    }
+
+    @Override
+    public Map<String, Set<Permission>> getRequiredPermissions() {
+        return CH.permissionsRequired(getClass());
+    }
+
+    @Override
+    public String describe() {
+        // Is this ever used?
+        return "DatasetVersion: " + version.getId() + " " + version.getDataset().getDisplayName() + ".v"
+                + version.getFriendlyVersionNumber();
+    }
+
+    public static WorkflowStepResult performDPNSubmission(DatasetVersion dv, AuthenticatedUser user, String host,
+            String aPort, String aDpnContext, ApiToken token) {
+
+        String port = aPort != null ? aPort : DEFAULT_PORT;
+        String dpnContext = aDpnContext != null ? aDpnContext : DEFAULT_CONTEXT;
+        if (host != null) {
+            Dataset dataset = dv.getDataset();
+            if (dataset.getLockFor(Reason.pidRegister) == null) {
+                //Use Duracloud client classes to login
+                ContentStoreManager storeManager = new ContentStoreManagerImpl(host, port, dpnContext);
+                Credential credential = new Credential(System.getProperty("duracloud.username"),
+                        System.getProperty("duracloud.password"));
+                storeManager.login(credential);
+
+                String spaceName = dataset.getGlobalId().asString().replace(':', '-').replace('/', '-')
+                        .replace('.', '-').toLowerCase();
+
+                ContentStore store;
+                try {
+                    /* If there is a failure in creating a space, it is likely that a prior version has not been fully processed 
+                     * (snapshot created, archiving completed and files and space deleted - currently manual operations 
+                     * done at the project's duracloud website) 
+                     */
+                    store = storeManager.getPrimaryContentStore();
+                    //Create space to copy archival files to
+                    store.createSpace(spaceName);
+                    DataCitation dc = new DataCitation(dv);
+                    Map<String, String> metadata = dc.getDataCiteMetadata();
+                    String dataciteXml = DOIDataCiteRegisterService.getMetadataFromDvObject(
+                            dv.getDataset().getGlobalId().asString(), metadata, dv.getDataset());
+                 
+                    try {
+                 // Add datacite.xml file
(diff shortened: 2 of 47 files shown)


qqmyers:
This PR includes PRs #5035, #5047, and #5048. Only the last 2 commits are unique.

qqmyers:
merging changes from other branches and fixed a bug in creating the bag info file when the contact and description fields have been given different URIs.

doc/sphinx-guides/source/admin/integrations.rst
```diff
@@ -85,6 +85,65 @@ SHARE
...
+
+The minimal configuration to support DPN integration involves adding one Dataverse Key and two Glassfish jvm options\:
+
+\:DuraCloudHost - the URL for your organization's DPN Duracloud site. For example: 
+
+    `curl http://localhost:8080/api/admin/settings/:DuraCloudHost -X PUT -d "qdr.duracloud.org"`
+
+:DuraCloudPort and :DuraCloudContext are also defined if you are not using the defaults ("443" and "duracloud" respectively).
+
+The username and password associated with your organization's DPN account should be configured in Glassfish:
```
scolapasta: Is there a particular reason to have some as settings and some as jvm options? I'm assuming we possibly do some of this ourselves, but we thinking of generally trying to get away from JVM options. 

(this doesn't mean we have to change thens definitively, just trying to understand the reasoning)

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -282,3 +282,25 @@ Available variables are:
 * ``minorVersion``
 * ``majorVersion``
 * ``releaseStatus``
+
```
scolapasta: this doc is called big data support. Should we create something new for workflows generally and have this doc point to that?

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
```diff
@@ -163,59 +162,31 @@ public DatasetVersion findByFriendlyVersionNumber(Long datasetId, String friendl
...
-                // DO nothing, just return null.
-            }
-            return foundDatasetVersion;
-
-        }
-        
-        if (majorVersionNumber == null && minorVersionNumber == null) {
-
+        if (majorVersionNumber == null) {
+            // Must always have a major version
```
scolapasta: not sure I understand this change - can you clarify?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -4349,4 +4350,34 @@ public void refreshPaginator() {
         setFilePaginatorPage(dt.getPage());      
         setRowsPerPage(dt.getRowsToRender());
     }  
+    
+    public void archiveVersion(Long id) {
```
scolapasta: Since we removed the UI does it make sense to leave this here? At the very list, we should have a comment explaining that there currently is no UI and this is here for future use.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
```diff
@@ -0,0 +1,247 @@
...
+import org.duracloud.client.ContentStore;
+import org.duracloud.client.ContentStoreManager;
+import org.duracloud.client.ContentStoreManagerImpl;
+import org.duracloud.common.model.Credential;
+import org.duracloud.error.ContentStoreException;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+@RequiredPermissions(Permission.ArchiveDatasetVersion)
```
scolapasta: I get the idea here behind this, but I'd like us to think a little more on permissions before adding a new one. As an example, we currently keep move dataverse and link dataverse as superuser only, because depending on what / how you move things it could lead to major re indexing. One idea we had was to have a new permission that is effectively a pseudo-superuser at a dataverse level (I call it that, because superuser means outside of the permission system), that would let you do the advanced things. Archive Dataset maybe falls under that?

So, I'd rather us maybe keep this superuser only for now until we can fully decide on what permission revamping we want (similarly, we're considering removing DeleteDatasetDraft, as we allow anyone who can edit to delete drafts). 

If you feel strongly about this, please let me/us know and we can maybe schedule a call to discuss. To be clear, I'm not saying this should permanently be superuser only, just for now, while we figure out things.

doc/sphinx-guides/source/admin/integrations.rst
qqmyers: @scolapasta - not really. I was trying to roughly follow the conventions used with PID providers with username/password in options. That said, it seems a little more secure to have password info only available to privileged accounts (versus an API call restricted to localhost). I doubt that's too significant since access to production servers would be limited anyway, so if you want everything to be settings, I can make an update.

doc/sphinx-guides/source/developers/big-data-support.rst
qqmyers: Sounds reasonable. I didn't want to start rearranging your docs, but I agree that, while there are big data uses for workflows, there's really nothing big-data specific about workflows or the included steps...

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
qqmyers: I'll add a comment. QDR is keeping the UI part for internal use, so it's helpful to have the underlying java code included. (Since the only other change required is in the xhtml, it's also nominally something that others could add by editing an xhtml file after deploying.) Also, fwiw - it was significant work to design so that the code could be called from a workflow and the DataPage - keeping this here might  help avoid future edits that would break one or the other...

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
qqmyers: @scolapasta  ...I'm not sure any more. I think I was trying to intuit what the FIXME was about and perhaps tried using the submitToDPN api call with a "1" or "2" and didn't get what I expected. All I can see now is that the old code would allow a query with the major version as null since the first query comes before the check at old line 184. That said, I don't see now that the change does much except simplify at the cost of changing what's returned for a major version number only - now you'd get 1.0 versus 1.x with the largest x.
Looking at the code, I don't see any use of this method except in my submit to DPN call, so I can restore the old code, leave it as is, or perhaps split it into find by friendly version and findLatestByMajorVersion(). What would you like?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
qqmyers: @scolapasta - I think my expectation was that people may want to allow curators (or other role) to submit versions, particularly from the GUI rather than someone who already had API access). Since the UI option is not going into the base version now, I don't think there's that much to gain from the permission. I'll go ahead and remove it.

qqmyers:
@scolapasta - I think an Archiver interface would be useful, but I'm not sure I know enough to abstract at this point. 

I could imagine a minor extension that would allow a setting to specify any subclass of an ArchiveCommand to plug into the API and perhaps generalizing the performDPNSubmission method to pass an array of settings rather than specific ones (both the command and the workflow step call this method after retrieving the DPN specific settings). Potentially a generic workflow step could use the same subclass setting to use whatever archive was installed (since it just gathers params that have to be specified in the json workflow specification and passes them to performDPNSubmission). 

The harder thing for me to think about generalizing is the DB entry to capture the result. For DPN v1, that's just a URL where the submission went, but I'd be surprised if a single URL will be enough for other archives or for DPN if/when we try to track whether the DPN submission has moved from their ingest space to long-term storage. Handling this probably involves having an Archiver manage a custom db table linked to datasetversion, but differences in what 'status' means would have to propagate to any GUI and perhaps involve more methods/api calls (for any status change beyond the initial submit/fail). And none of this addresses what happens for archives that would use an external workflow/asynchronous handling (a generic returnedFromProcessing method and second API call that could be null for simpler Archivers?). (One more thought - not sure if an Archiver class could/should automate plugging in as a post-publication workflow and powering an API? Right now, the API works if you add the settings, but turning on the workflow for new pubs is a separate step.)

That said, if you think there's value in doing the first few steps above, I can do that (the only oddity I can think of is that the workflow json would have to always include the ArchiverClass setting along with whatever settings that class needs (since workflows can only access the settings specified in the json).

The last comment also brings up that I did make changes to the workflow mechanism that help here, such as allowing you to specify settings that should be passed in and working through the transaction issues that allow the workflow to make updates (e.g. to write the submitted Url). Hopefully those make it easier to design other Archivers and start getting a better idea of what's common...

doc/sphinx-guides/source/admin/integrations.rst
scolapasta: Ok, I think we can leave for now, but we may want to change this at some future point.

doc/sphinx-guides/source/developers/big-data-support.rst
scolapasta: Rearranging is fine. You can always ping @dlmurphy if you want some help sorting through the docs.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
scolapasta: Gotcha. Adding a comment will be enough, then.

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
scolapasta: I do think if you only have a major version, it should return the latest (i.e. 1.x) and not 1.0. The example being that we if you have a version 1.x you still have version 1 in the citation. Minor revisions are really meant to represent the latest of that version #.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
scolapasta: Can you explain again, in this case, what you mean by the curator? In other words, why wouldn't checking if they have the publishDataset permission be enough?

scolapasta:
@qqmyers I do think the changes suggested above (about generalizing) would be useful - it may not fully scale, but at least it would allow for a use to something that works similar to, but is not, DPN.

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
qqmyers: OK - I'll revert this.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitToArchiveCommand.java
qqmyers: Just the curator role. So it would work to reuse the publishDataset permission / that would be better than superuser. (It may be that there are still use cases where separate permissions make sense, but just not restricting to superuser only should address QDR's interests for now.) So I'll remove the permission but switch to requiring the publishDataset permission rather than restricting to superuser.

qqmyers:
OK - I'll give an Archiver class a shot and see if any issues come up...

qqmyers:
@scolapasta - I think this PR is ready for review again. I created an abstract base class for archiver integration and a setting used to specify the archive specific class to use. I've tested the basics at QDR from the workflow and API/our custom GUI angles...

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
```diff
@@ -0,0 +1,99 @@
...
+import static edu.harvard.iq.dataverse.engine.command.CommandHelper.CH;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+@RequiredPermissions(Permission.PublishDataset)
+public abstract class AbstractSubmitToArchiveCommand implements Command<DatasetVersion> {
```
scolapasta: @qqmyers  Any reason not to extend AbstractCommand (and then you wouldn't have to override so many methods?)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
```diff
@@ -0,0 +1,99 @@
...
+        return request;
+    }
+
+    @Override
+    public Map<String, Set<Permission>> getRequiredPermissions() {
+        return CH.permissionsRequired(getClass());
+    }
+
+    @Override
+    public String describe() {
```
scolapasta: @qqmyers this is used by the ActionLogRecord; there is a generic implementation in AbstractCommand.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
```diff
@@ -0,0 +1,99 @@
...
+        // Is this ever used?
+        return "DatasetVersion: " + version.getId() + " " + version.getDataset().getDisplayName() + ".v"
+                + version.getFriendlyVersionNumber();
+    }
+
+    /**
+     * This method is the only one that should be overwritten by other classes. Note
+     * that this method may be called from the execute method above OR from a
+     * workflow in which execute() is never called and therefore in which all
```
scolapasta: I haven't looked in workflows in detail.Is there a reason they can't just call execute?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
qqmyers: DatasetVersion is not a DvObject so AbstractCommand doesn't quite fit.

scolapasta: Hmm, good point. I wonder if you could do something like this then (pass version number):
https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetSpecificPublishedDatasetVersionCommand.java

It could even have several constructors (one with version numbers, one with version ids).

And let the command get the version.

Alternatively, it could make sense to modify AbstractCommand to be more flexible (not necessarily in scope for this issue)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
qqmyers: OK - how's my update to more closely mirror the format for DvObjects in AbstractCommand (latest commit)?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
qqmyers: Possibly. That wouldn't log the version number without modifying AbstractCommand though. And since the api call and future GUI start from the version, they'd have to find the dataset to send just so the version could be looked up again. Just brainstorming - some other redesign that let you put any/all params into a Map (whether DvObject or not) with AbstractCommand printing simple params like longs and calling callbacks on other objects to have them print themselves might be useful...

scolapasta: OK, so I found an example of a command that does take a DataseVersion:
https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeaccessionDatasetVersionCommand.java

Could you mimic that?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
scolapasta: It's ok to make this similar to the abstractCommand (though I'm still hoping you won't need to) or to modify it to put info that is important to record in the ActionLogRecord, but not handled by the generic describe. So, I'm ok with this.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractSubmitToArchiveCommand.java
scolapasta: Definitely some room for redesign (I also consulted with @michbarsinai who had worked on both these sections - " we could refactor the Commands and WorkFlow packages to be a bit closer at some point. Or maybe add a workflow step wrapper for commands or something". But we'll leave that for maintenance and out of scope for this issue.

pdurbin:
@qqmyers heads up that this pull request now has merge conflicts. If you could resolve them it would be appreciated!

qqmyers:
@pdurbin - done. Happy Holidays!

kcondon:
@qqmyers I was unable to complete this before break but am working on this now. Do you mind one last update from /develop, since we've just cut 4.10.1? Fingers crossed I will complete testing on this asap.

qqmyers:
@kcondon - try that. No conflicts on the merge so I didn't retest...

kcondon:
@qqmyers Thanks, Jim but I'm getting errors on deployment relating to loggers:
Caused by: java.lang.IllegalStateException: Detected both log4j-over-slf4j.jar AND slf4j-log4j12.jar on the class path, preempting StackOverflowError. See also http://www.slf4j.org/codes.html#log4jDelegationLoop for more details.
        at org.apache.log4j.Log4jLoggerFactory.<clinit>(Log4jLoggerFactory.java:51)
        ... 84 more
]]

qqmyers:
@kcondon - I'm not sure I've found anything - there were some logger.info statements added during testing v4.10.1 that shouldn't have been here (and aren't in v4.10.1) and are removed with the latest commit but I don't see how that could change anything about the logger classes used. Could something else have changed in your setup?

kcondon:
@qqmyers /develop builds and deploys ok on the same box but this branch does not.

kcondon:
@qqmyers Now getting build failure due to test failure:
[ERROR] Failures: 
[ERROR]   MarkupCheckerTest.testSanitizeBasicHTML:95
[INFO] 
[ERROR] Tests run: 586, Failures: 1, Errors: 0, Skipped: 5
[INFO] 
[ERROR] There are test failures.

kcondon:
@qqmyers That fixed the test failure, back to the deployment issue/logger conflict. I have not goggled the error much but one post showed it may result from some libraries choosing which logger to use which may cause a conflict. Solution was to explicitly exclude logger from pom.xml. Also followed the link in error, then sub link to read more about bridge logging for legacy loggers:

Caused by: java.lang.IllegalStateException: Detected both log4j-over-slf4j.jar A
ND slf4j-log4j12.jar on the class path, preempting StackOverflowError. See also 
http://www.slf4j.org/codes.html#log4jDelegationLoop for more details.
        at org.apache.log4j.Log4jLoggerFactory.<clinit>(Log4jLoggerFactory.java:
51)
:

qqmyers:
@kcondon - it's odd: the two libraries in your error are both in the branch I'm running here. I did find that the BagGenerator class imports log4j rather than java.util.logging, so I'm switching that now in hopes it works. If not, I agree that trying to exclude the library is a next step to try.

qqmyers:
@kcondon - this update doesn't break my build/deploy - does it fix anything for you?

kcondon:
@qqmyers Same logger conflict, sorry. Leonid suggested that maybe we should clear the maven libraries dir, m2, on Jenkins build machine in case there is something cached there. What do you think?

pameyer:
For what it's worth, I took at very quick look at this branch last night.  As of 9ffe806ef ; I didn't have any problems either building or deploying.

qqmyers:
@kcondon - both sound reasonable. I've uploaded a war to http://phoenix.dataverse.org/dataset.xhtml?persistentId=doi%3A10.5072%2FFK2%2FZ71GCR&version=DRAFT (let me know if somewhere else is better). The only change in that was changing the version in the pom.xml.

I still can't spot any code changes that should affect this so the idea that there have been pom changes in both the dev branch and this one that are somehow conflicting / have led to an extra/different jar getting built in seems like the best guess.

kcondon:
@qqmyers I've tried both and neither makes a difference. I'm going to touch base with our dev team to see whether someone else might assist.

pdurbin:
My first thought is that someone could try the fancy new ec2-create-instance.sh script on @qqmyers 's branch. The syntax would be 

`./ec2-create-instance.sh -b IQSS/4706-DPN_Submission_of_archival_copies -r https://github.com/QualitativeDataRepository.git`

See also http://guides.dataverse.org/en/4.10.1/developers/deployment.html

kcondon:
I'm putting this back into dev for now.

qqmyers:
@kcondon - while I still don't understand why several of us can build and you can't, I did find the dependency for the log4j-over-slf4j.jar in the duracloud libraries and I've updated the pom to exclude it (latest commit). I've verified that the excludes remove that jar and don't affect submission to Duracloud. So - if this works for you, I think we're good (despite the mystery if why it worked, it seems clear that both jars in the error message should not be in there.)

(Also FWIW: I noticed that the version 1.7.7 of the SLF4J jars was hardcoded in the pom.xml as of Dataverse 4.3 and they're up to v1.7.25 now...)

Issue #5078, opened 2018-09-21
java.util.MissingResourceException: Can't find bundle for base name Bundle, locale en


pdurbin:
I spent quite a bit of time today struggling with my dev environment on Mac after pulling the latest from the "develop" branch, commit d0cb402, which was the merge for pull request #5023 for #4892.

I was unable to deploy to Glassfish and saw this error in Glassfish's server.log:

    java.util.MissingResourceException: Can't find bundle for base name Bundle, locale en

This was coming from line 50 of BundleUtil:

   bundle = ResourceBundle.getBundle(defaultBundleFile, bundle_locale, loader);

Here's a handy link: https://github.com/IQSS/dataverse/blob/d0cb4021a9396e6ac47aef97910abe028fa49510/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java#L50

In desperation, I almost completely rebuilt my dev environment, which the exception of Solr:

- new installation of Glassfish (unzipped it fresh)
- re-ran the installer
- re-cloned the repo

What seemed to fix it was the following:

```
mkdir /tmp/lang
cp src/main/java/Bundle.properties /tmp/lang
```

Why `/tmp/lang`? Because that's what the code defaults to if you don't have the new `dataverse.lang.directory` JVM option defined (the installer doesn't set it):

https://github.com/IQSS/dataverse/blob/d0cb4021a9396e6ac47aef97910abe028fa49510/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java#L29-L32

The extra weird thing is that now that I can deploy to Glassfish again, I don't seem to be able to get back to seeing MissingResourceException. For example, I tried moving the lang directory aside, stopping Glassfish, and removing the "generated" folder. I expected the error to come back but it didn't. I can still deploy. Very strange.

Also very strange is the fact that this problem only seems to happen on Mac rather than Linux.

I'm concerned that new developers will have a bad experience trying to get their dev environments set up, so I think we should do something, but I'm not sure what.

If anyone out there runs into this problem too, please let us know.

pameyer:
This sounds similar to an issue that was fixed in the PR.  It is odd that I'm not seeing this on linux systems, but that it shows up on os x.

pdurbin:
Yep. It's a weird one. @sekmiller also mentioned that when he changes Bundle.properties like we always have the changes aren't reflected so he updated the new JVM option to point to the traditional location of Bundle.properties as a workaround.

[...]

PR #5079, opened 2018-09-21
Bundle trouble #5078

pdurbin:
connect to #5078


JayanthyChengan:
@pdurbin : when we deploy the app from develop branch ( if the JVM option is not set for 
-Ddataverse.lang.directory and if the /tmp/lang directory doesn't exists ) ,  then the app should not look for files in file system and it should not throw exception while deploying.

pdurbin:
@JayanthyChengan ok, thanks. I'm not sure why it's throwing an exception then. Very strange.

JayanthyChengan:
@pdurbin : 
The changes in BundleUtil.java is fine.
Do we still need the quick fix comments in doc/sphinx-guides/source/developers/dev-environment.rst ?

pdurbin:
@JayanthyChengan I hope not. I just reverted it in 9ce1b3a

JayanthyChengan:
The logic in src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java looks good.

We have to follow the same logic in src/main/java/edu/harvard/iq/dataverse/util/LocalBundle.java ( which is used in faces-config.xml) for resourcebundle.

JayanthyChengan:
@sekmiller:
After reading the comment at https://github.com/IQSS/dataverse/issues/5078#issuecomment-424365753
    
Noticed that the if-condition at https://github.com/IQSS/dataverse/blob/26bb4280f0cb8e4d97e26dc769c120ff901c31a1/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java#L31 , checks only if the **length of string** is 0 ,

Not checking whether the directory set from JVM **exists** and whether the directory contains files.

I feel, it's better to check that the directory mentioned in the JVM option exists and is not empty.

sekmiller:
Hi  @JayanthyChengan, we did make a change to the logic so that if the jvm option is present we want to throw a missing resource error if the directory isn't present - just in case there's a typo in the jvm option, say and the installation would silently go about its business using the default bundles. 

If you disagree with this approach, please let us know why.

Thanks!

JayanthyChengan:
@sekmiller:  It's clear now, and I accept your suggestion.  Thanks for the explanation

Issue #4261, opened 2017-11-06
Dataverse discovery in Google - Sitemaps and schema.org


eugene-barsky:
Hello:

AS per Philip's request, we sat in a Google session about data discoverability, and Dr. Natasha Noy @Google was mentioning Google preference to include sitemaps for better data indexing. 

They also talked about Google preference for schema.org coding for the landing pages.

Thanks,

Eugene
<img width="1199" alt="32423887-4f77b876-c277-11e7-91c1-a4894e05420f" src="https://user-images.githubusercontent.com/10885112/32454470-280f6e98-c2d4-11e7-9c3b-281031abf876.png">

pdurbin:
Yep, as I mentioned at https://github.com/IQSS/dataverse/issues/2717#issuecomment-342036473 it was about 24 minutes into the video at https://www.rd-alliance.org/making-data-discoverable-web-search-engines that sitemaps were mentioned and the slide above appeared. Thanks for opening this issue @eugene-barsky

mheppler:
In issue #4555 @pameyer commented:

>"Why don't my datasets show up in google?" seems like a question that comes up relatively commonly (but completely out of scope for this issue).

This would be the appropriately scoped issue for that question.

djbrooke:
We'll estimate this and work on this. I'm removing the schema.org part of the title, since that was delivered in 4.8.4. We still need to add a sitemap. This is more important since Google Dataset Search is now a thing. :)

pdurbin:
I'm playing around with Documentation Driven Development (DDD), if that's a thing, by making pull request #5084 which for now is only a stub of the direction I think we're going. See d2ccf59

At standup I mentioned that for my family site I recently switched from Jekyll to Hugo, which creates a sitemap at http://thedurbins.com/sitemap.xml . Hugo creates an XML file but at https://support.google.com/webmasters/answer/183668 Google indicates they support multiple formats:

- XML
- RSS, mRSS, and Atom 1.0
- Text

My assumption is that we want to create an XML file. I assume we'll be using an EJB timer to control how often the XML file is updated. If I'm misunderstanding any requirements, please advise.

pdurbin:
As I mentioned at standup, I did a little code and docs clean up and now I'm ready for code review for pull request #5084.

kcondon:
Sending back, here's your requested "Punch List" @pdurbin, mostly what we had discussed last evening:

[...]

PR #5084, opened 2018-09-24
add doc stub for sitemaps #4261

pdurbin:
connects to #4261

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -418,6 +418,9 @@ Out of the box, Dataverse attempts to block search engines from crawling your in
 Letting Search Engines Crawl Your Installation
 ++++++++++++++++++++++++++++++++++++++++++++++
 
+Ensure robots.txt Is Not Blocking Search Engines
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 For a public production Dataverse installation, it is probably desired that search agents be able to index published pages (AKA - pages that are visible to an unauthenticated user).
 Polite crawlers usually respect the `Robots Exclusion Standard <https://en.wikipedia.org/wiki/Robots_exclusion_standard>`_; we have provided an example of a production robots.txt :download:`here </_static/util/robots.txt>`).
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/sitemap/SiteMapUtil.java
```diff
@@ -0,0 +1,157 @@
...
+        String sitemapPathAndFile;
+        // i.e. /usr/local/glassfish4/glassfish/domains/domain1
+        String domainRoot = System.getProperty("com.sun.aas.instanceRoot");
+        if (domainRoot != null) {
+            // Note that we write to a directory called "sitemap" but we serve just "/sitemap.xml" using PrettyFaces.
+            sitemapPath = domainRoot + File.separator + "docroot" + File.separator + "sitemap";
+        }
+        sitemapPathAndFile = sitemapPath + File.separator + SITEMAP_FILENAME;
+
+        DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
```
matthew-a-dunlap: It seems like this generation could benefit from schema validation via DocumentBuilderFactory.setSchema(schema) ?

src/test/java/edu/harvard/iq/dataverse/sitemap/SiteMapUtilTest.java
```diff
@@ -0,0 +1,109 @@
...
+            assertTrue(XmlValidator.validateXmlWellFormed("/tmp/sitemap.xml"));
+        } catch (Exception ex) {
+            System.out.println("Exception caught checking that XML is well formed: " + ex);
+            wellFormedXmlException = ex;
+        }
+        assertNull(wellFormedXmlException);
+
+        Exception notValidAgainstSchemaException = null;
+        try {
+            assertTrue(XmlValidator.validateXmlSchema("/tmp/sitemap.xml", new URL("https://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd")));
```
matthew-a-dunlap: Ahh I see we validate here. Do we know if-when google updates there schema whether they will support old versions? I could see value in general of validating against the schema on each run if it lets us know more quickly that google won't stop indexing us.

Issue #5027, opened 2018-09-06
Admin Guide - Dashboard


mheppler:
There is no page in the Admin Guide for the Dashboard. There are references to the Dashboard in the Managing Harvesting Clients, Managing Harvesting Server and Sets, and User Administration pages, but there should also be a more detailed page, which those references should link to.

mheppler:
Added placeholder page for Dashboard docs in Admin Guide. Fixed original references to the Dashboard link location, which was recently changed as part of the UI changes required for the new language toggle #4685.

mheppler:
The most detailed reference to the Dashboard is on the Admin Guide index.

>These “superuser” tasks are managed via the new page called the Dashboard. A user logged in as a Dataverse Admin will see the Dashboard link rendered in the upper right corner of every Dataverse page.

This is something that should probably be moved to the new Dashboard page, and the index can link to that page for more detail.

pdurbin:
I just created pull request #5114 and will move it to code review.

PR #5114, opened 2018-09-28
add page for superuser dashboard #5027

pdurbin:
connects to #5027

### DIFF
--- a/doc/sphinx-guides/source/admin/dashboard.rst
+++ b/doc/sphinx-guides/source/admin/dashboard.rst
@@ -1,0 +1,31 @@
+Dashboard
+=========
+
+A dashboard of administrative tools is available to superusers by clicking "Log In", then their name, and then "Dashboard" in the dropdown.
+
+.. contents:: Contents:
+	:local:
+
+Harvesting
+----------
+
+Harvesting Clients
+~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


doc/sphinx-guides/source/admin/dashboard.rst
```diff
@@ -0,0 +1,31 @@
+Dashboard
+=========
+
+Dataverse offers a dashboard of administrative tools for superusers only. If you are a logged-in superuser, you will see a "Dashboard" link in the navbar, next to your name. You can verify that you are a superuser by checking your name in the navbar. If it's red, you have the right permissions to use the Dashboard.
```
mheppler: With the language toggle added to the navbar, the dashboard link has moved from the navbar to the account name dropdown menu. Also, would it be worth linking to how to make a user a superuser so that they could see this page? (Note: the User Administration pg of the Admin Guide is the appropriate page to link to, but that page only mentions "whether or not they have Superuser status" as opposed to clearly pointing out "this is where you grant superuser status". See: http://guides.dataverse.org/en/4.9.2/admin/user-administration.html)

dlmurphy: Thanks for letting me know about the dropdown menu change, I'll update this! I'll link to the User Admin page too, and look into editing that page to add a little more detail about granting superuser status.

Issue #5126, opened 2018-10-01
Internationalization:  Need to create key=value property file for Built-in Roles(JSON files)


JayanthyChengan:
Noticed that there are English terms in the built-in roles json files, which are loaded to database table(dataverserole) and displays in roles-assign.xhtml, roles-edit.xhtml and permission-manage.xhtml.  Planning to create a property files for Internationalization.

scolapasta:
Hi,
I was looking for my review comment about Custom Roles, but couldn't find it. The issue here is that if you do have a custom role it would look for it in the bundle and not find it (rather than not display info from what is in db). Though if we remove the columns from the db, then that's a separate problem.

JayanthyChengan:
commented here: https://github.com/IQSS/dataverse/pull/5158#discussion_r226311372

pdurbin:
@JayanthyChengan I see @scolapasta was asking about custom roles. Are custom roles still supported? If so, what do installations need to do to continue using them and can you add something in the Admin Guide about custom roles? You can see the custom roles we use in Harvard Dataverse at https://dataverse.harvard.edu/role_permission_helper.xhtml

JayanthyChengan:
Hi @pdurbin:  I am attaching the screenshot of permissions-manage.xhtml, where I concentrated for Roles/Permission Internationalization in this issue. 

In my test instance, custom roles are displayed under Roles tab in permission-manage.xhtml without any issue and there is no need for any additional installation steps.  

Created https://github.com/IQSS/dataverse/blob/28d05a43b68bf5be2a1abedf3ba935e22bc1216c/src/main/java/BuiltInRoles.properties -  the property file with roles name/description pulled from JSON files to handle multiple languages.

Also few keys related to permission were added to bundle.properties.

Thanks for pointing me to this page: role_permission_helper.xhtml, I was not aware of it( I will create a separate issue for this).  


![screen shot 2018-10-23 at 9 51 17 am](https://user-images.githubusercontent.com/12862735/47366249-50e90400-d6ab-11e8-875c-51fe4a5fdafe.png)
![screen shot 2018-10-23 at 9 51 26 am](https://user-images.githubusercontent.com/12862735/47366250-50e90400-d6ab-11e8-89cc-473d4919c475.png)
![screen shot 2018-10-23 at 9 52 03 am](https://user-images.githubusercontent.com/12862735/47366251-50e90400-d6ab-11e8-996a-f264814c6d89.png)

[...]

PR #5158, opened 2018-10-09
5126 json built in roles

JayanthyChengan:
connects to #5126


src/main/java/Bundle.properties
```diff
@@ -2069,3 +2069,32 @@ admin.api.migrateHDL.failure.must.be.hdl.dataset=Dataset was not registered as a
 admin.api.migrateHDL.success=Dataset migrate HDL registration complete. Dataset re-registered successfully.
 admin.api.migrateHDL.failure=Failed to migrate Dataset Handle id: {0}
 admin.api.migrateHDL.failureWithException=Failed to migrate Dataset Handle id: {0} Unexpected exception: {1}
+
+#Roles
```
scolapasta: These are actually permissions; let's rename them that way.

src/main/java/Bundle.properties
```diff
@@ -2069,3 +2069,32 @@ admin.api.migrateHDL.failure.must.be.hdl.dataset=Dataset was not registered as a
...
+roles.EditDataverse=EditDataverse
+roles.EditDataset=EditDataset
+roles.ManageDataversePermissions=ManageDataversePermissions
+roles.ManageDatasetPermissions=ManageDatasetPermissions
+roles.PublishDataverse=PublishDataverse
+roles.PublishDataset=PublishDataset
+roles.DeleteDataverse=DeleteDataverse
+roles.DeleteDatasetDraft=DeleteDatasetDraft
+
+permission.AddDataverse=Add a dataverse within another dataverse
```
scolapasta: and this is their descriptions.

src/main/java/Bundle.properties
scolapasta: I'm not convinced we want to translate these - though yes to their descriptions. (it's probably ok to translate, it's just these are "reserved words" in the code.

src/main/java/edu/harvard/iq/dataverse/authorization/DataverseRole.java
```diff
@@ -117,15 +119,29 @@ public void setId(Long id) {
 	}
 
 	public String getName() {
-		return name;
+		if (alias != null )
+		{
+			String key = "role." + alias.toLowerCase() +".name";
+			return BundleUtil.getStringFromPropertyFile(key,"BuiltInRoles" );
```
scolapasta: How will this work with Custom roles?

JayanthyChengan: @scolapasta:  I didn't handle custom role in this pull request.  Is custom role also comes from JSON files?  Haven't investigated on the logic behind custom role.  Please add some information related to custom roles,  can we handle it in separate issue?

src/main/java/Bundle.properties
JayanthyChengan: I will discuss with @amberleahey and @mhvezina to clarify whether those "reserved words" from line number 2074 to 2086 to be translated.

mhvezina: Since those strings are displayed in the user interface, I would like to have a way to be able to translate them.

src/main/java/edu/harvard/iq/dataverse/authorization/DataverseRole.java
amberleahey: After testing the fix @JayanthyChengan made today, custom roles in our local dvdev is now working as expected, can create/edit/apply custom roles. 
customroles
applyingmytestrole

src/main/java/Bundle.properties
amberleahey: I too agree we should translate these: they may be role types that are stored but they are readable/meaningful in the user interface.  This is working as per the updated pull request
customroles

src/main/java/edu/harvard/iq/dataverse/authorization/DataverseRole.java
JayanthyChengan: @scolapasta: the code got outdated after I committed new code( which will handle custom roles).  Hope I understood correctly and updated the code accordingly.  Please review

djbrooke:
Thanks @JayanthyChengan. Is this ready for code review?

JayanthyChengan:
Yes, it's ready for code review

Issue #4371, opened 2017-12-08
As a researcher, I want more dataset metadata in schema.org exports so that my data is more discoverable


jggautier:
In issue #2243, some metadata fields important for dataset discovery were excluded from mapping to Schema.org. We said we'd include them in a later issue. This is that issue, and these are those fields ([dun dun](https://www.youtube.com/watch?v=gP3MuUTmXNk)):

- Creator types (person or organization) for dataset authors (a separate ticket has been opened for this, https://github.com/IQSS/dataverse/issues/5029, since the missing creator types are making Google's Dataset Search engine not display creator names and there are UI implications)
- Dataset identifier (DOI or HDL as a URL) using the `@id` property
- Dataset identifier (DOI or HDL as a URL) using the `url` property
- Name of funding source (that's all schema.org supports for now; how to include other funding source details (like grant numbers) are discussed in [this github issue in schema.org's repo](https://github.com/schemaorg/schemaorg/issues/383))
- Author identifiers
- Geographic coverage
- Multiple dataset descriptions
- File metadata (see schema.org [from Zenodo](https://zenodo.org/record/1419226/export/schemaorg_jsonld#.W5wh1f5KhTY) and [from ICPSR](https://www.icpsr.umich.edu/icpsrweb/ICPSR/studies/23980/export) for an example of how these fields are used)
  - File PID 
  - File download URL (when there is one - excludes restricted files and files in datasets with guestbooks)
  - File name
  - File description
  - File format

We'll also need to fix:
- The property used for dataset authors. Dataverse is using `author`, but I think Google Dataset Search is ignoring author and prefers `creator`. (See [this comment](https://github.com/IQSS/dataverse/issues/5029#issuecomment-424835378) on #5029)
- Keywords and Topic Classifications (Dataverse concatenates each keyword into one value, and each topic classification into one value; see [this dataset's schema.org metadata](https://dataverse.harvard.edu/api/datasets/export?exporter=schema.org&persistentId=doi%3A10.7910/DVN/24034))
- Provider: Change the value to use the installation name (instead of hardcoding "Dataverse")
  - For the `provider` property we hardcode "Dataverse" and put the installation name in the `DataCatalog` `name` property, but Dataset Search is displaying a "Data provided by" field and is using what's in the `provider` property.

[...]

PR #5169, opened 2018-10-11
improved Schema.org JSON-LD output

pdurbin:
connects to #4371


qqmyers:
W.r.t. the discussion about reloading the tsv file. From some IRC discussion with @pdurbin, I see that adding identifiers will still allow the code to match the old entries since it tries matching on name before identifier. So my concerns above are moot/don't apply.

Issue #5183, opened 2018-10-13
restrict/unrestrict performance for large number of files


qqmyers:
In QDR, trying to restrict/unrestrict  ~1900 files (all but a few files in the dataset), can take hours. Looking at the code, it appears that there are unnecessary nested loops. Will submit a PR...

PR #5184, opened 2018-10-13
IQSS-5183 - improve RestrictFileCommand performance

qqmyers:
## New Contributors

## Related Issues

- connects to #5183 restrict/unrestrict large numbers of files

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -2376,10 +2376,11 @@ public class DatasetPage implements java.io.Serializable {
                 requestContext.execute("PF('selectFilesForRestrict').show()");           
             return "";
         } else {           
-            boolean validSelection = false;
+            boolean validSelection = true;
             for (FileMetadata fmd : selectedFiles) {
-                if (!fmd.isRestricted() ){
-                    validSelection = true;
+                if (fmd.isRestricted() == true) {
+                    validSelection = false;
+                    break;
                 }
             }
             if (!validSelection) {
@@ -2406,8 +2407,9 @@ public class DatasetPage implements java.io.Serializable {
         } else {
             boolean validSelection = true;
             for (FileMetadata fmd : selectedFiles) {
-                if ((fmd.isRestricted() && restricted) || (!fmd.isRestricted() && !restricted)) {
+                if (fmd.isRestricted() == restricted) {
                     validSelection = false;
+                    break;
                 }
             }
             if (!validSelection) {
@@ -2433,61 +2435,16 @@ public class DatasetPage implements java.io.Serializable {
         return  returnToDraftVersion();
     }
 
-    public void restrictFiles(boolean restricted) throws CommandException {
-   
-        //if (previouslyRestrictedFiles == null) {
-        // we don't need to buther with this "previously restricted" business 
-        // when in Create mode... because all the files are new, so none could 
-        // have been restricted previously;
-        // (well, it looks like the code below should never be called in the 
-        // CREATE mode in the first place... the edit files fragment uses
-        // its own restrictFiles() method there; also, the fmd.getDataFile().equals(fmw.getDataFile()))
-        // line is not going to work on a new file... so be mindful of all this
-        // when the code between the 2 beans is merged in 4.3.
-        if (editMode != EditMode.CREATE) {
-            previouslyRestrictedFiles = new ArrayList<>();
-            for (FileMetadata fmd : workingVersion.getFileMetadatas()) {
-                if (fmd.isRestricted()) {
-                    previouslyRestrictedFiles.add(fmd);
-                }
+    private void restrictFiles(boolean restricted) throws CommandException {
+        Command<Void> cmd;
+        previouslyRestrictedFiles = new ArrayList<>();
+        for (FileMetadata fmd : this.getSelectedFiles()) {
+            if(fmd.isRestricted()) {
+                previouslyRestrictedFiles.add(fmd);
             }
-            
-            Command cmd;
-            String fileNames = null;
-            for (FileMetadata fmw : workingVersion.getFileMetadatas()) {
-                for (FileMetadata fmd : this.getSelectedFiles()) {
-                    if (restricted && !fmw.isRestricted()) {
-                    // collect the names of the newly-restrticted files, 
-                        // to show in the success message:
-                        // I don't think this does the right thing: 
-                        // (adds too many files to the message; good thing this 
-                        // message isn't used, normally)
-                        if (fileNames == null) {
-                            fileNames = fmd.getLabel();
-                        } else {
-                            fileNames = fileNames.concat(fmd.getLabel());
-                        }
-                    }
-                    if (fmd.getDataFile().equals(fmw.getDataFile())) {
-                        cmd = new RestrictFileCommand(fmw.getDataFile(), dvRequestService.getDataverseRequest(), restricted);
-                        commandEngine.submit(cmd);
-                        
-                        
-//                        fmw.setRestricted(restricted);
-//                        if (workingVersion.isDraft() && !fmw.getDataFile().isReleased()) {
-//                            // We do not really need to check that the working version is 
-//                            // a draft here - it must be a draft, if we've gotten this
-//                            // far. But just in case. -- L.A. 4.2.1
-//                            fmw.getDataFile().setRestricted(restricted);
-//                        }
-                    }
-                }
-            }
-            if (fileNames != null) {
-                String successMessage = JH.localize("file.restricted.success");
-                logger.fine(successMessage);
-                successMessage = successMessage.replace("{0}", fileNames);
-                JsfHelper.addFlashMessage(successMessage);
+            if (restricted  != fmd.isRestricted()) {
+                cmd = new RestrictFileCommand(fmd.getDataFile(), dvRequestService.getDataverseRequest(), restricted);
+                commandEngine.submit(cmd);
             }
         }
     }
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommand.java
@@ -50,56 +50,35 @@ public class RestrictFileCommand extends AbstractVoidCommand {
         if (publicInstall) {
             throw new CommandExecutionException("Restricting files is not permitted on a public installation.", this);
         }
-        if (file.getOwner() == null){
+        // check if this file is already restricted or already unrestricted
+        if (restrict == file.getFileMetadata().isRestricted()) {
+            String text = restrict ? "restricted" : "unrestricted";
+            throw new CommandExecutionException("File " + file.getDisplayName() + " is already " + text, this);
+        }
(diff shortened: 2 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2376,10 +2376,11 @@ public String testSelectedFilesForRestrict(){
...
             return "";
         } else {           
-            boolean validSelection = false;
+            boolean validSelection = true;
             for (FileMetadata fmd : selectedFiles) {
-                if (!fmd.isRestricted() ){
-                    validSelection = true;
+                if (fmd.isRestricted() == true) {
+                    validSelection = false;
+                    break;
```
qqmyers: These changes just match the logic update in the next method that was done previously - to assure the whole set of files is valid rather than that there is at least one that would be changed.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2406,8 +2407,9 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
         } else {
             boolean validSelection = true;
             for (FileMetadata fmd : selectedFiles) {
-                if ((fmd.isRestricted() && restricted) || (!fmd.isRestricted() && !restricted)) {
+                if (fmd.isRestricted() == restricted) {
```
qqmyers: simpler logic, same result - files shouldn't already be in the state that this method is changing them to.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2406,8 +2407,9 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
...
                     validSelection = false;
+                    break;
```
qqmyers: just more efficient

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2433,61 +2435,16 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
...
-        // its own restrictFiles() method there; also, the fmd.getDataFile().equals(fmw.getDataFile()))
-        // line is not going to work on a new file... so be mindful of all this
-        // when the code between the 2 beans is merged in 4.3.
-        if (editMode != EditMode.CREATE) {
-            previouslyRestrictedFiles = new ArrayList<>();
-            for (FileMetadata fmd : workingVersion.getFileMetadatas()) {
-                if (fmd.isRestricted()) {
-                    previouslyRestrictedFiles.add(fmd);
-                }
+    private void restrictFiles(boolean restricted) throws CommandException {
```
qqmyers: private since it is not called externally.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2433,61 +2435,16 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
...
+            if(fmd.isRestricted()) {
+                previouslyRestrictedFiles.add(fmd);
             }
-            
-            Command cmd;
-            String fileNames = null;
-            for (FileMetadata fmw : workingVersion.getFileMetadatas()) {
-                for (FileMetadata fmd : this.getSelectedFiles()) {
-                    if (restricted && !fmw.isRestricted()) {
-                    // collect the names of the newly-restrticted files,
```
qqmyers: because this is called from restrictSelectedFiles above that calls save() which already sets a success message, this list of filenames is never used/never shown because the attempt to add a flash message below fails (an error in the log).

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2433,61 +2435,16 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
...
+        for (FileMetadata fmd : this.getSelectedFiles()) {
+            if(fmd.isRestricted()) {
+                previouslyRestrictedFiles.add(fmd);
             }
-            
-            Command cmd;
-            String fileNames = null;
-            for (FileMetadata fmw : workingVersion.getFileMetadatas()) {
-                for (FileMetadata fmd : this.getSelectedFiles()) {
-                    if (restricted && !fmw.isRestricted()) {
```
qqmyers: What this loop does that the replacement above doesn't is to handle the case where a file is, for example, restricted in the released version and already unrestricted in the draft and the user wants to unrestrict it. I don't think there's any real effect of letting this happen (an unrestricted file being set to unrestricted), but if this case needs to still be caught, it can be caught more efficiently in the command itself (which has to find the FileMetadata for this file that is associated with the draft version anyway), and/or could be caught when deciding if the selected files are all valid (it would be inefficient to do it there, but if the user needs to be alerted (versus the current mode of silently doing a noop), it might be better to completely remove the selection and edit items from a display of released versions when a draft version exists ("Edits can be made to the existing draft(link) version"). (Similar issue if the restrict request is from a published version and the file has already been removed from an existing draft. In the existing code and this PR (after second commit), there is silently no change made for this case.)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2433,61 +2435,16 @@ public String restrictSelectedFiles(boolean restricted) throws CommandException{
...
-                    // collect the names of the newly-restrticted files, 
-                        // to show in the success message:
-                        // I don't think this does the right thing:
```
qqmyers: Not too relevant now but it's true - this loop listed all files that could possibly be updated because it is just testing the whole filemetadata list (repeatedly). Moving it inside the test for fmd.getDataFile().equals(fmw.getDataFile()) would make it work.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommand.java
```diff
@@ -50,56 +50,41 @@ protected void executeImpl(CommandContext ctxt) throws CommandException {
...
-                    if (file.getFileMetadata().isRestricted() != restrict) {
-                        throw new CommandExecutionException("Failed to update the file metadata", this);
+            // We need the FileMetadata for the file in the draft dataset version and the
+            // file we have may still reference the fmd from the prior released version
+            FileMetadata draftFmd = file.getFileMetadata();
+            if (dataset.isReleased()) {
+                // We want to update the draft version, which may not exist (if the file has
+                // been deleted from an existing draft, so we want null unless this file's
+                // metadata can be found in the current version
+                draftFmd=null;
```
qqmyers: This check here handles the dropped file case that was one reason for the double loop in DatasetPage

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommand.java
```diff
@@ -50,56 +50,41 @@ protected void executeImpl(CommandContext ctxt) throws CommandException {
...
+                    if (file.equals(fmw.getDataFile())) {
+                        draftFmd = fmw;
+                        break;
                     }
                 }
             }
-            else{
-                throw new CommandExecutionException("Working version must be a draft", this);
+            if (draftFmd != null) {
+                draftFmd.setRestricted(restrict);
```
qqmyers: Could check here to make sure the current file metadata is not already in this state.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommandTest.java
```diff
@@ -91,6 +92,7 @@ public void testRestrictUnpublishedFile() throws CommandException{
     public void testRestrictPublishedFile() throws Exception{
         file.setOwner(dataset);
         dataset.setPublicationDate(new Timestamp(new Date().getTime()));
+        file.setPublicationDate(dataset.getPublicationDate());
```
qqmyers: This test was passing with the original code, but I don't think it should have - the file is not released unless the date is set. What I think was really happening was that the call to dataset.getEditVersion was not finding a version that had this file in its list of metadatas, so no change was being made to the file.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommandTest.java
```diff
@@ -131,6 +133,99 @@ public void testRestrictRestrictedFile() throws Exception {
         
     }
     
+    @Test
```
qqmyers: Basically mirror images of the existing tests - just verified that the logic is symmetric.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommandTest.java
```diff
@@ -89,19 +91,33 @@ public void testRestrictUnpublishedFile() throws CommandException{
...
+        // Restrict on a published file will cause the creation of a new draft dataset version
+        // and should update only the FileMetadata in the draft version for the test file.
+        // So we need to make sure that we use one of the files in the dataset for the test 
+        DataFile file = dataset.getFiles().get(0);
+        // And make sure is is file.isReleased() == true
+        file.setPublicationDate(dataset.getPublicationDate());
+        // And set its owner, which is usually done automatically, but not in the test setup
+        file.setOwner(dataset);
+        //And set the version state to released so that the RestrictFileCommand will create a draft version
+        dataset.getLatestVersion().setVersionState(VersionState.RELEASED);
```
qqmyers: All these updates do a better job of putting the dataset/it's initial version into the released state and having the file being restricted as one of the ones in the list of files in the dataset (and its metadata in the list of file metadatas).

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/RestrictFileCommandTest.java
```diff
@@ -89,19 +91,33 @@ public void testRestrictUnpublishedFile() throws CommandException{
...
             if (file.equals(fmw.getDataFile())) {
-                assertEquals(fmw, file.getFileMetadata());
+                fileFound=true;
+                //If it worked fmw is for the draft version and file.getFileMetadata() is for the published version
                 assertTrue(fmw.isRestricted());
+                assertTrue(!file.getFileMetadata().isRestricted());
+                break;
             }
         }
+        assertTrue(fileFound);
```
qqmyers: The updated asserts make sure that the file is found in the dataset and verify that only the file metadata for the new draft version is updated to restricted.

oscardssmith:
One other thing that will speed this up a bit is cherypicking one line of my permissions pr that got bloated and never merged. Specifically, there is a line in the permissions code that causes a double recursion over groups.

pdurbin:
@oscardssmith which line was that?

oscardssmith:
It's in `permissionServiceBean` in  `permissionsForSingleRoleAssignee`. The recursive loop over groups is non necessary since the transitive closure is found outside.

pdurbin:
@oscardssmith ok, so https://github.com/IQSS/dataverse/blob/v4.9.4/src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java#L298 . Thanks.

Issue #5185, opened 2018-10-14
:ExcludeEmailFromExport setting no longer has an effect


pameyer:
in docker-aio:
```
[ERROR] Failures: 
[ERROR]   DatasetsIT.testExcludeEmail:565 expected:<[[]]> but was:<[sammi@sample.com]>
[INFO] 
[ERROR] Tests run: 76, Failures: 1, Errors: 0, Skipped: 1
```
which would appear to break https://github.com/IQSS/dataverse/issues/3443.

Not fully clear on where the 1 skip is coming from, but that's a problem for another day.

IQSS jenkins appears to be showing same behavior https://build.hmdc.harvard.edu:8443/job/phoenix.dataverse.org-apitest-develop/.

pdurbin:
@pameyer thanks for opening this issue. I tweaked the title a bit and I'll paste a screenshot of http://guides.dataverse.org/en/4.9.4/installation/config.html#excludeemailfromexport to explain what the :ExcludeEmailFromExport setting does:

![screen shot 2018-10-15 at 10 08 59 am](https://user-images.githubusercontent.com/21006/46956139-8b76ef00-d062-11e8-81a8-93e1e3f389c7.png)

Based on the timing I suspect pull request #5047 may have introduced this regression from a quick look at the code, I can't tell. I can definitely reproduce the regression on my laptop. Here's how it looks in Netbeans:

![screen shot 2018-10-15 at 10 12 49 am](https://user-images.githubusercontent.com/21006/46956268-e27cc400-d062-11e8-95ac-49703483e0d2.png)

pameyer:
@pdurbin thanks - looks good to me (Tests run: 76, Failures: 0, Errors: 0, Skipped: 1).  Still not sure where the skip is coming from, but that's a different problem.

pdurbin:
@qqmyers thanks for pull request #5192 which is a variation on my pull request #5191. I'm going to let someone with more Java experience than me decide which approach is better. The main thing is that we've identified a fix for the regression.

scolapasta:
It seems a little strange to have a static reference to an EJB. (but it also looks as if that was there before and you are just setting it*).  Not necessarily for this issue, but as tech debt cleanup, I wonder if Json Printer should be a service bean as well (and then could just inject the setting service).

[...]

PR #5191, opened 2018-10-15
get :ExcludeEmailFromExport working again #5185

pdurbin:
connects to #5185

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
@@ -154,7 +154,9 @@ public class ExportService {
             if (releasedVersion == null) {
                 throw new ExportException("No released version for dataset " + dataset.getGlobalId().toString());
             }
-            final JsonObjectBuilder datasetAsJsonBuilder = JsonPrinter.jsonAsDatasetDto(releasedVersion);
+            // We pass settingsService into the JsonPrinter constructor to check the :ExcludeEmailFromExport setting.
+            JsonPrinter jsonPrinter = new JsonPrinter(settingsService);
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -154,7 +154,9 @@ public void exportAllFormats(Dataset dataset) throws ExportException {
             if (releasedVersion == null) {
                 throw new ExportException("No released version for dataset " + dataset.getGlobalId().toString());
             }
-            final JsonObjectBuilder datasetAsJsonBuilder = JsonPrinter.jsonAsDatasetDto(releasedVersion);
+            // We pass settingsService into the JsonPrinter constructor to check the :ExcludeEmailFromExport setting.
+            JsonPrinter jsonPrinter = new JsonPrinter(settingsService);
+            final JsonObjectBuilder datasetAsJsonBuilder = jsonPrinter.jsonAsDatasetDto(releasedVersion);
```
qqmyers: Could we keep this as a static reference (JsonPrinter.jsonAsDatasetDto(releasedVersion); and perhaps change the constructor line to just be a static set method and use that here instead? (I suspect it was the warning that 'static methods should be accessed in a static manner' from my IDE that caused me to make this change in the first place...) (I can make a PR to your branch if you want.)

pdurbin: @qqmyers sure, please make a PR against my branch. Thanks!

scolapasta:
closing. Will push https://github.com/IQSS/dataverse/pull/5192 along.

Issue #5218, opened 2018-10-19
Support arbitrary filetypes in index


matthew-a-dunlap:
If Dataverse hits a filetype that is not in its known list (MimeTypeFacets), it throws an exception on indexing. This causes the filetype to not be indexed (the file itself still indexed).

There was already code in place to catch this exception when there was no match, but that exception was being caught by a _different_ piece of code before it could be used to add a filetype.

I noticed this updating the index #5053. Note that in the fix I made sure that the exception is still being caught by the lower level code for all cases other than filetype indexing.

matthew-a-dunlap:
Here is how my image results changed (268 files).

Previous:
![screen shot 2018-10-19 at 3 38 01 pm](https://user-images.githubusercontent.com/8901015/47240059-23ebd700-d3b5-11e8-8bdb-85630f5a7d70.png)

After fix:
![screen shot 2018-10-19 at 3 42 04 pm](https://user-images.githubusercontent.com/8901015/47240176-8fce3f80-d3b5-11e8-8789-af6fd2c0dde4.png)

pdurbin:
I just left a review on pull request #5219 with some discussion points.

PR #5219, opened 2018-10-19
5218 arbitrary filetype index

matthew-a-dunlap:
## Related Issues

- connects to #5218 

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
@@ -29,7 +29,7 @@ public class BundleUtil {
         return getStringFromBundleCatchMissing(key, arguments, bundle);
     }
 
-    public static String getStringFromBundle(String key, List<String> arguments, ResourceBundle bundle) {
+    public static String getStringFromBundle(String key, List<String> arguments, ResourceBundle bundle) throws MissingResourceException {
         if (key == null || key.isEmpty()) {
             return null;
         }
@@ -47,7 +47,7 @@ public class BundleUtil {
         }
     }
     
-    //This call was added to allow bypassing the exception catch, for filetype indexing
+    //This call was added to allow bypassing the exception catch, for filetype indexing which also catches it
     public static String getStringFromBundleCatchMissing(String key, List<String> arguments, ResourceBundle bundle) {
         try {
             return getStringFromBundle(key, arguments, bundle);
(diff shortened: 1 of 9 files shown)


src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
```diff
@@ -28,18 +29,15 @@ public static String getStringFromBundle(String key, List<String> arguments) {
         return getStringFromBundle(key, arguments, bundle);
     }
 
-    public static String getStringFromBundle(String key, List<String> arguments, ResourceBundle bundle) {
+    public static String getStringFromBundleNoMissingCheck(String key, List<String> arguments, ResourceBundle bundle) throws MissingResourceException {
```
pdurbin:
```suggestion
    private static String getStringFromBundleNoMissingCheck(String key, List<String> arguments, ResourceBundle bundle) throws MissingResourceException {
```

In addition to making the new method above private, we could do the same for `getResourceBundle` below.

src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
```diff
@@ -48,11 +46,30 @@ public static String getStringFromBundle(String key, List<String> arguments, Res
...
+    }
 
     public static String getStringFromPropertyFile(String key, String propertyFileName  ) {
         ResourceBundle bundle = getResourceBundle(propertyFileName);
         return getStringFromBundle(key, null, bundle);
     }
+    
+    //Added this function to fix indexing issue where it was expecting the exception
+    // to bubble up. See FileUtil.getFacetFileType --MAD 4.9.4
+    public static String getStringFromPropertyFileAllowMissing(String key, String propertyFileName  ) {
```
pdurbin: Can this comment be converted to javadoc?

pdurbin: In general, I feel like the change could be a focused a little more on how getStringFromPropertyFile works. That method is now only used twice? Instead of having it and getStringFromPropertyFileAllowMissing can we just have a single method?

Before (develop):

screen shot 2018-10-22 at 10 52 35 am


After (this pull request):

screen shot 2018-10-22 at 10 53 11 am

matthew-a-dunlap: After the latest commit!

screen shot 2018-10-22 at 11 43 50 am

Thanks for the feedback, its even better now

matthew-a-dunlap: I decided to switch getResourceBundle back to public so it can be used in the future alongsize getStringFromBundle passing in an arbitrary bundle, even though right now it doesn't need the scope. Allows better testing as well.

Issue #5220, opened 2018-10-20
Slow and duplicate DataCite registration 


qqmyers:
I was just looking through the DataCite code (hoping to borrow it for another project) and noticed a) that the code creates/closes an httpClient for each call, which repeats all of the ssl setup, etc., and b) that, at least for a 'normal' dataset, the publicizeExternalIdentifier() call is made twice. 
For a) it looks like it is a reasonable practice to just have a singleton httpClient for DataCite calls. For b) I'm not sure how/when the two calls were introduced. One is now inside an
```
if (!datasetExternallyReleased) 
```
while the second is inside

```
        if ( theDataset.getLatestVersion().getVersionState() != RELEASED ) {
            // some imported datasets may already be released.
            if (!datasetExternallyReleased){
```
My guess is that the former should removed, but I wonder if the datasetExternallyReleased flag was needed here because the first copy above was there. (In the externally released case, is the latest version RELEASED?).

I'll submit a PR...

pameyer:
@qqmyers Have you tested the import API with this branch?  It might be possible that some of this is related to importing a dataset with an existing PID.

qqmyers:
@pameyer - I haven't. Both copies appear to have been there for a while and the check if(!datasetExternallyReleased) was applied to both copies, presumably as part of supporting the import with existing PIDs...

matthew-a-dunlap:
For testing this branch we should ensure our test cases for publishing with datacite still work. It looks like `publicizeExternalIdentifier` was being called twice accidentally (one has been removed in this branch), but its possible that removal has opened up some case. To me it looks like the removed call was publishing released datasets when it shouldn't.

qqmyers:
@matthew-a-dunlap  - I'm not sure there is/was a case where the publish command was getting triggered on released datasets, unless it was the externally released case. So a) hopefully it was just publishing twice on drafts, and b) a question: was the datasetExternallyReleased flag needed because the first call was publishing them even though they were 'released' / should datasetExternallyReleased==true datasets also be released when they hit this code and therefore the second and now only call will skip them and would do so without the extra flag? (hopefully that makes sense). I don't know enough about the import code to know if that's the case...

PR #5221, opened 2018-10-20
5220 slow data cite registration

qqmyers:
## New Contributors

## Related Issues

- connects to #5220 Slow and duplicate DataCite registration

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteRegisterService.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteRegisterService.java
@@ -42,11 +42,19 @@ public class DOIDataCiteRegisterService {
 
     @EJB
     DataverseServiceBean dataverseService;
+
     @EJB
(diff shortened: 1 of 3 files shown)


qqmyers:
I tested this with an 80 file dataset. The original code tool 114 seconds to run publicizeExternalIdentifier once (with file DOIs turned on), and 68 seconds to run it again a second time. The changes above dropped that to ~48 seconds. Doing a second publication (major or minor version) is taking 36-39 seconds. (Other than the title being one version back, which I believe is a separate issue, it looks like publishing new versions is updating the metadata of the dataset and all datafiles, regardless of whether they have changed...).

I'm not quite sure how to interpret the difference between the first and second runs - if the internal processing is ~40 seconds, meaning the new code is spending almost no time on the https calls themselves, and the second run in the old code used cached data, so the 68 seconds was mostly https setup, I think the results make sense. In any case, it looks like both removing the duplicate call and keeping a singleton httpClient both help...

Issue #5288, opened 2018-11-07
Dependency Housekeeping and how developers will benefit from it


poikilotherm:
This issue is a story from or a blocker to #5292.
This issue is related to #4260, #5274, #3921, #4172 and more.
Please feel addressed @matthew-a-dunlap and @scolapasta :wink: 

While working on #4172 I discovered the problems noted down in #5274, which are in turn a left-over of #3921. This lead me to the impression, that the current `pom.xml` could use some love in terms of dependency management, hopefully avoiding more problems down the road. More and more deps keep kicking in while functionality is added to Dataverse. (As @pdurbin said on IRC: Dataverse 4.0 was around 45MB, 4.9.4 is at 146 MB)

I made some first steps primarily to address #5274, but this might be a could starting point for more work in this. Reducing WAR size, find wrong imports, update dependencies, address convergence issues and more...

### Done so far ~~(mostly for the resolution of #5274)~~

- [x] Add initial dependency management for this via BOMs
- [x] Add direct dependencies to Jackson as it is in use within the code.
- [x] Remove JaCoCo compile scope dependency and fix resulting import errors from transitive deps not present anymore.
- [x] Add checksums to the `local_lib` JARs to make Maven happier
- [x] Write usefull docs (and rules?) about dependency management for old and new Dataverse developers (placed in a new section within the dev docs)

### Things open (and will stay open till we are on Glassfish 5+, see #5274)
- [ ] Remove AWS SDK bundle and replace with "S3 only"
- [ ] Test the S3 stuff against some server to be sure it works, as no integration tests are present (see #5068 and #4690)

### Things to consider

- [ ] Think and talk about moving `local_lib` into a GitHub based "poor mans" Maven repo.
- [ ] Cleanup direct and transitive deps for the SWORDv2 code
- [ ] Think and talk about adding [Maven Enforcer](https://maven.apache.org/enforcer/maven-enforcer-plugin/) as a plugin permanently, maybe add it to the build cycle and fail builds on convergence issues.
- [ ] Address more/all conflicts from `mvn enforcer:enforce`
- [ ] Address issues from `mvn dependency:analyze`
- [ ] Think about replacing the retired [Apache Abdera](https://abdera.apache.org/) library with sth. like [ROME](https://rometools.github.io/rome/) - Note: ROME is used by Tika and included anyway.

poikilotherm:
Hi @pdurbin,

I just finished writing my doc draft. 

[...]

Issue #5274, opened 2018-11-05
Stripping AWS Mail component with TrueZIP blocks packaging for docker images


poikilotherm:
Dear devs,
Dear @bsilverstein as the one who wrote the code for #3921,

while working on #4172, I am crafting a new Maven profile to generate smallest possible Docker images based on the [payara 5 server-full image](https://hub.docker.com/r/payara/server-full) using https://dmp.fabric8.io.

As I want to make things really small, I don't want to bundle the dataverse dependencies within the WAR file, but create a small WAR _without_ all deps, containing only classes and ressources. The deps get moved in by copying them the `/lib` of the Glassfish/Payara domain (so they are kind of "provided" scoped in Maven terminology). (For those interested in Docker: this should provide us with smaller layer changes and smaller hub transfers!)

The stripping of the mailing stuff from the AWS SDK makes this hard to achieve in two ways: a) I would to fiddle around with this, too and b) the TrueZIP plugin simply fails as I don't build the WAR with deps inside and it is not easy to skip it in a profile without making a mess in the `pom.xml`...

Back in #3921 where all of this was introduced, there has been a pointer to the [AWS dev docs](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-project-maven.html#configuring-maven-individual-components) about using individual components.

Did you guys ever try to get that rolling? If not: maybe this would be the perfect time to do so now... Willing to contribute on this. :smile: 

----
This is related to a question I made on IRC ([see logs](http://irclog.iq.harvard.edu/dataverse/2018-11-05#i_77735)).

EDIT:
@pdurbin suggested some clarification on IRC.

The `pom.xml` plugin section I would like to see removed:

```diff
-            <!-- v4.8: The truezip-maven-plugin below deletes two copies of a file that the AWS SDK bundle
-            includes called javamail.providers which breaks system emails. TODO: make our own AWS SDK without Amazon SES.
-            - bsilverstein 8/8/2017 -->
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>truezip-maven-plugin</artifactId>
-                <version>1.1</version>
-                <executions>
-                    <execution>
-                        <id>remove-javamail-providers-from-exploded</id>
-                        <goals>
-                            <goal>remove</goal>
-                        </goals>
-                        <phase>package</phase>
-                        <configuration>
-                            <fileset>
[...]

PR #5289, opened 2018-11-07
5288 Introducing proper deps management and docs about it.

poikilotherm:
~~As suggested by @pdurbin on IRC ([logs](http://irclog.iq.harvard.edu/dataverse/2018-11-14#i_78785)), this PR is primarily about the issues found in #5274.~~ This also paves some of the way to #5288, but lets work in smaller chunks.

Please see my comment below. Splitted this up so it does not break AWS on current Glassfish 4.1 environments.

----

## Related Issues

- connects to #5288 

## Pull Request Checklist

I could use a hand with testing the S3 stuff in some EC2 or so. Don't have this at hand and a deployment with Minio is currently quite an amount of work.


dlmurphy:
I'm finished making commits, all looks good to me!

doc/sphinx-guides/source/developers/documentation.rst
```diff
@@ -78,11 +78,10 @@ GraphViz based images
 
 In some parts of the documentation, graphs are rendered as images via Sphinx GraphViz extension.
 
-This needs `GraphViz <http://graphviz.org/>`_ installed and either ``dot`` on the path or
+This requires `GraphViz <http://graphviz.org/>`_ installed and either ``dot`` on the path or
 `adding options to the make call <https://groups.google.com/forum/#!topic/sphinx-users/yXgNey_0M3I>`_.
 
-On Mac and Linux this should "just work", on Windows this is currently untested. The worst thing that might happen
-is a warning and missing images in your local documentation build.
+This has been tested and works on Mac, Linux, and Windows. If you have not properly configured GraphViz, then the worst thing that might happen is a warning and missing images in your local documentation build.
```
poikilotherm: Great! Good to know this works on a Windows box :smile:

doc/sphinx-guides/source/developers/dependencies.rst
```diff
@@ -259,14 +258,16 @@ Repositories are defined within the Dataverse POM like this:
...
+Typically you will skip the addition of the central repository, but adding it to the POM has the benefit that
+dependencies are first looked up there (which in theory can speed up downloads). You should keep in mind that repositories
 are used in the order they appear.
 
 ----
-.. [#ide] Modern IDEs import your Maven POM and offer import autocompletion for classes based on direct dependencies
-          in the model. You might end up using legacy or repackaged classes because of a wrong scope.
-.. [#ide2] This is going to bite back in modern IDEs when importing classes from transitive dependencies by "autocompletion accident".
+
+.. rubric:: Footnotes
```
poikilotherm: Thanks for adding this - I tried to find other examples, but didn't find one.

poikilotherm:
@matthew-a-dunlap 

> Maybe we should put a note in the top of our pom telling folks to read this doc section before doing edits.

Sure, good idea! Maybe that's even better than my explanation comments. Do feel like adding a commit for this?

> I think the "direct dependency" rules section should be its own section. The background info is pretty vital to understanding this part of maven (and I've definitely learned things reading it) but if folks need to refresh themselves its better if its easier to get to.

I am not opinionated on this. @dlmurphy what do you think?

> In the rules discussing implicit usage, etc, its probably worth pointing to the helpful tools so folks know what they can do.

Sure, good idea. Feel free to add a commit :+1: 

> I feel the lists in Transitive dependencies & managing transitive dependencies overlap somewhat. I'm not sure we need to list all the ways maven can handle them in a list, maybe we can instead add the ways we want to our rules and then discuss the bad ways in a section of text?

The exclude option is currently in use within the real POM, but I wasn't in the mood to add more docs about that, as IMHO this is last resort stuff. IMHO a developer might enjoy a short overview, which is not easy to get from the detailed blogs and docs out there (took me a while to sort this out... :wink: )

> I'm curious what you see as the remaining tech debt in our dependency management.

Let's discuss that in #5288, alright? Maybe the things I collected over there are an inspiration already :wink:

pdurbin:
> Do feel like adding a commit for this?

Done in 5b49f50 (in pom.xml direct people to new dependencies page)

poikilotherm:
As @kcondon said that this would have broken current setups based on Glassfish 4.1, I once more changed this PR to only address the dep management stuff. Thus reassigning this to #5288 again and hope this can be merged then.

Issue #5202, opened 2018-10-16
Internationalization - alert messages is not toggling with language selection


JayanthyChengan:
The alert info/warning/success messages remains English , and doesn't toggle with Language selection .

![image](https://user-images.githubusercontent.com/12862735/47039452-fa7f4100-d151-11e8-95d4-ef0f9b434bfd.png)

pdurbin:
I just saw "please review the code" at https://github.com/IQSS/dataverse/pull/5299#issuecomment-440353165

mheppler:
There is a lot of discussion about the specifics of this issue in the pull request #5299.

I have just created a new branch `5202-edit-dataverse-info-msg` off of the original branch @JayanthyChengan created, in order to fix the display of the info msg on the Edit Dataverse pg.

<img width="1156" alt="screen shot 2018-11-28 at 10 53 30 am" src="https://user-images.githubusercontent.com/687227/49164325-e5730180-f2fc-11e8-8b1b-da4e0fc11d01.png">

Will put more detail and feedback into the pull request, to keep the conversation going there.

kcondon:
@mheppler @JayanthyChengan Why when I look at this PR is there a *ton* of changes not specifically related to the title of this ticket? It makes it less clear what the scope is. The PR title is more accurate, more broad. Just trying to understand what these changes mean.

kcondon:
OK, I'll take a look. Might have misunderstood the essence of the ticket and the conversation moved to PR.

JayanthyChengan:
@kcondon please let me know, if I can answer any of your questions.  Thanks

scolapasta:
Unfortunately, the solution of using f:view on each page had other conflicting issues (see #5400). So we had to revert that back.

The plan is to go back to solving this a different way. It seems to be the issue is that the DataverseLocaleBean, which is a container managed bean is being used in a non managed way by the BundleUtil. So we'd like to try and see if we can make sure that BundleUtil does that, likely by also making it managed (possibly renaming it with the convention we've used for other such classes - BundleWrapper). The thought is by doing this, it wouldn't be creating a new LocaleBean and would use the one from the session which definitely knows the correct Locale.

mheppler:
[...]

Issue #4679, opened 2018-05-16
Internationalization : remove English static strings from javascript files 


JayanthyChengan:
**Example 1 :** 
/src/main/webapp/file-versions.xhtml [Line 27 ] 
/src/main/webapp/dataset-versions.xhtml [Line 27 ] 

preload_message = "(Loading versions...)"

**Example 2 :** 
/src/main/webapp/loginpage.xhtml [Line 162 ]
"Your Browser does not support javascript. Please use"

**Example 3:**
src/main/webapp/resources/js/dv_rebind_bootstrap_ui.js
 show_info_msg('Edit Dataverse', 'Edit your dataverse and click Save Changes. Asterisks indicate required fields.');
   
**Example 4:**
/src/main/webapp/resources/js/mydata.js [Line 318-322] -- View Next... Results...

mheppler:
While looking at the pull request for #5299, I did some research on how to solve this javascript issue, and found a [Stack Overflow answer](https://stackoverflow.com/a/14575170) which should solve Example 1 and Example 2.

```
<script>
    var message = '#{bundle[some.key]}';
</script>
```

For Example 3, I believe I have a workaround for `dv_rebind_bootstrap_ui.js` which removes that message text, and properly delivers the text via the dataverse pg manage bean -- to be delivered as part of #5299.

That would leave the `mydata.js` text as the last remaining unsolved example in this issue.

JayanthyChengan:
@mheppler  Implemented the above mentioned solution for Example 1 and 2.  Thanks.

And handled mydata.js(Example 4) by adding hidden field in dataverseuser.xhtml.  Could you please review the code?  Thanks

PR #5299, opened 2018-11-09
5202 - Internationalization for breadcrumbs , alert messages

JayanthyChengan:
## Related Issues

- connects to #5202 

## Pull Request Checklist
 - Merged latest from "develop"  branch


src/main/webapp/ThemeAndWidgets.xhtml
```diff
@@ -6,7 +6,7 @@
       xmlns:p="http://primefaces.org/ui"
       xmlns:c="http://xmlns.jcp.org/jsp/jstl/core"
       xmlns:jsf="http://xmlns.jcp.org/jsf">
-
+<f:view locale="#{dataverseLocaleBean.localeCode}">
```
mheppler: When researching this approach (http://jdevelopment.nl/internationalization-jsf-utf8-encoded-properties-files/) it appears that the `html` tag has the `lang` attribute using the same dynamic value. Our xhtml pgs do not have this attribute, but instead it is found on the `dataverse_template.xhtml` file. This PR should also review the use of this attribute and what is best moving forward. Will review with @scolapasta for a more detailed recommendation.

mheppler: While reviewing the use of `lang` in the `dataverse_template.xhtml`, we should also look at the meta element `http-equiv="Content-Language" content="en"`. According to W3 this should no longer be used in HTML5 (https://www.w3.org/International/questions/qa-html-language-declarations).

mheppler:
@JayanthyChengan, Thank you for your efforts in delivering this functionality. Sorry for the delay in getting you more concrete feedback, but I have been reviewing your PR changes with @scolapasta and learning more about the `f:view` tag as it is applied here.

Gustavo and I agree with your approach of applying the `f:view` tag to each page, as it appears to be the only method we could find to apply the language change to the alert messages delivered by the `message.xhtml` custom component pg. We had hoped that applying it to the `dataverse-template.xhtml` would be a more efficient way to solve this functionality being applied to every page in the app, but it didn't appear to work. The `f:view` that is on that page currently isn't sufficient, and we tried moving it around, to no avail. So we can move forward with this approach on every page as a workaround until we find a cleaner, more scalable solution.

Here is our list of to-do's based on our code review.

- [ ] Remove `<f:view locale="#{dataverseLocaleBean.localeCode}">...</f:view>` from the `dataverse-template.xhtml` as it won't be needed with it being applied to every pg
- [ ] Change `lang` attribute in the `html` tag of the `dataverse-template.xhtml` to `lang="#{dataverseLocaleBean.localeCode}"`
- [ ] Add `xml:lang="#{dataverseLocaleBean.localeCode}"` attribute to `html` tag of the `dataverse-template.xhtml`
- [ ] Remove deprecated `<meta http-equiv="Content-Language" content="en"/>` meta element from the `h:head` of the `dataverse-template.xhtml`

Also, with the `f:view` tag being removed from the `dataverse-template.xhtml`, it will be needed on all pages. Here is an inventory of all the pages, some of which you have already taken care of in this pull request.

- [ ] 403.xhtml
- [ ] 404.xhtml
- [ ] 500.xhtml
- [X] ThemeAndWidgets.xhtml
- [ ] confirmemail.xhtml
- [X] dashboard-users.xhtml
- [X] dashboard.xhtml
- [ ] dataset-widgets.xhtml
- [X] dataset.xhtml
- [ ] dataverse.xhtml
- [X] dataverseuser.xhtml
- [X] editdatafiles.xhtml
- [ ] file.xhtml
- [X] guestbook-responses.xhtml
- [X] guestbook.xhtml
- [X] harvestclients.xhtml
- [X] harvestsets.xhtml
- [ ] iframe.xhtml
- [ ] loginpage.xhtml
- [X] manage-groups.xhtml
- [X] manage-guestbooks.xhtml
- [X] manage-templates.xhtml
- [ ] passwordreset.xhtml
- [ ] permissions-manage-files.xhtml
- [X] permissions-manage.xhtml
- [ ] privateurl.xhtml
- [ ] shib.xhtml
- [X] template.xhtml
- [X] /search/advanced.xhtml
- [ ] /oauth2/callback.xhtml
- [ ] /oauth2/convert.xhtml
- [ ] /oauth2/firstLogin.xhtml

JayanthyChengan:
Thanks @mheppler and @scolapasta .  I will do the changes and submit it.

JayanthyChengan:
@mheppler  I have done the changes as mentioned above.  Please review the code.  

Also note that I haven't added <f:view in dataverse.xhtml, since it results in below error:

image

mheppler:
@JayanthyChengan thank you for the updates. I've got two items for you.

First, we need both the `lang` and the `xml:lang` attributes in the `html` on the `dataverse_template.xhtml` page.

- [ ] Add `lang="#{dataverseLocaleBean.localeCode}"` attribute in the `html` tag of the `dataverse-template.xhtml` page

Second, hopefully we can find a solution to add the `f:view` to the `dataverse.xhtml` page. At first I did not see the `javax.servlet.ServletException: /dataverse.xhtml @10,52 locale=#{dataverseLocaleBean.localeCode}" null` error you reported when I was building the branch locally. But after a deleteing and adding my `:Languages` config setting back to the db, I finally got the error to show. But it went away after a clean and build and a page refresh or two.

Reviewing this issue with @scolapasta, we didn't hit on any solid feedback to give you. Gustavo suspected it to could with the location of the `f:view` and how/when `dataverseLocaleBean` is being initialized.

I did some digging on StackOverflow, and [this suggestion by BalusC](https://stackoverflow.com/a/30655151) might shed some light on our problem.

Good luck.

JayanthyChengan:
@mheppler Thanks for the review and stackOverflow link.  I have done the changes, please review the code.

mheppler:
@JayanthyChengan thank you for your latest revisions. I see that the `f:view` was added to the dataverse pg, but it does not appear to be applied to the info msg on the edit view of the page.

screen shot 2018-11-26 at 2 43 20 pm

Also, it appears that an unexpected result of the changes to the `DataverseLocaleBean.java` bean was the language toggle reloading the page entirely. So that if you were on the edit dataverse or edit dataset pg and use the dropdown toggle, the page is reloaded and the user is no longer viewing the edit form.

If you have any ideas or questions, hopefully @scolapasta can better guide from this point on. We've long since past the limits of my JSF know-how.

scolapasta:
OK, so I did some more digging and have some info that could help:

I tried first to see why the info message on the dataverse page does not translate. I think I know why. If you look at the info message on the Dataset page, it calls:

JH.addMessage(FacesMessage.SEVERITY_INFO, JH.localize("dataset.message.editMetadata"));

while the DataversePage calls:

FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO, BundleUtil.getStringFromBundle("dataverse.edit.msg") , BundleUtil.getStringFromBundle("dataverse.edit.detailmsg")));

Notice how Dataset is using JH.localize. That said, if you look at the JSFHelper class, this localize message is deprecated:

    /**
     * @deprecated Localization applies not only to the front end (JSF) but also
     * the API so consider using the newer, more flexible BundleUtil methods
     * instead.
     */
    @Deprecated
    public String localize( String messageKey ) {

So, we can solve the Dataverse page issue using this method (I think), BUT we really should be finding a different solution verall, using the more flexible BundleUtil methods. I also wonder if we changed the code to use these BundleUtil messages (which is what the none messages code already does), if we could then go back to just having the jsf:view on the template, and not every single page.

@JayanthyChengan would you be able to experiment with this? Please let me know if you need more guidance on what to try and I can help figure it out.

mheppler:
@JayanthyChengan thank you for catching that msg was originating from the `dv_rebind_bootstrap_ui.js` and sorry I did not pick that up. Also my apologies for my confusion related to the language toggle page refresh. I was thinking of the behavior on the Manage Harvesting Clients pg, which continues to display the info msg after you toggle because it is it's own independent pg, as opposed to an "edit mode" view that the dataset and dataverse pgs use.

With that said, we can scratch item 3 from your list above, as my mistake. @scolapasta is going to provide some feedback for item 2. That leaves item 1, which I have fixed for you! You should see that I have created a new branch with the changes to the dataverse pg, and sent a [pull request](https://github.com/scholarsportal/dataverse/pull/26) to your branch.

I was able to remove the text from the javascript, and changed the dataverse pg to instead receive the msg from the bundle via the `DataversePage.java` manage bean like the rest of the pages do.

screen shot 2018-11-28 at 10 53 30 am

One suggestion that I will make about the French bundle is that now there are two hyphens and no trailing period on the last sentence. I didn't want to edit the bundle myself, but instead thought a note here would be more appropriate.

Let me know if you have any questions. I have also made a comment regarding the changes in the issue for javascript Internationalization #4679. (Also added a reference to a fix for a few of the other text+javascript examples in your comment!)

scolapasta:
@JayanthyChengan I would make the changes to not use localize in this PR. Since that helper is used only for messages (I'm pretty sure this is true) and that what's this issue is for, it makes sense to me to do this fully the first time. I also think that this may allow for the jsf:view to only be on the template* and would appreciate if you could experiment some with that to get it to work. 

* the reason I suspect this is that this only uses:
Locale locale = facesContext.getViewRoot().getLocale();
to get the locale whereas I think the session scoped bean will have it captured.

JayanthyChengan:
@scolapasta 
In this PR https://github.com/IQSS/dataverse/pull/5358, 
- replaced JH.localize in all java programs with BundleUtil.getStringFromBundle()
- added <f:view only for template ( tried with adding xml:lang , @lang to html tag)

 src/main/webapp/dataverse_template.xhtml
 src/main/webapp/manage-templates.xhtml
 src/main/webapp/template.xhtml

But I don't see the language toggle in the messages and breadcrumbs , for example in dataverseuser.xhtml 
(_Why have a Dataverse account? To create your own dataverse and customize it, add datasets, or request access to restricted files_)
image



After adding <f:view to dataverseuser.xhtml page, the messages toggle with language selection

image

scolapasta:
@JayanthyChengan so unfortunately, not using localize did not solve the issue with having to use f:view on every page. Oh well, at this point, we can leave that for some other day.

Could you make one of these PRs (or even a new one) have all the changes, and then close the other one(s)? That will make it straightforward to do what should be one final review and then QA.

The final PR should have `f:view` on all the pages you can navigate to.  Also, @mheppler noticed that you added the `lang` and `xml:lang` attributes to `template.xhtml` and `manage-template.xhtml` and they are not needed there (only needed  on `dataverse_template.xhtml`).

JayanthyChengan:
@scolapasta @mheppler 
closed PR https://github.com/IQSS/dataverse/pull/5358 and merged the code to here.
Please review it,  let me know if there is any corrections.  Thanks

kcondon:
@JayanthyChengan One down, one to go

Issue #5300, opened 2018-11-09
Internationalization - date format


JayanthyChengan:
to customize date display with language selection

mheppler:
Implemented the `LONG` date and time format as outline in the [`DateFormat` class documentation](https://docs.oracle.com/javase/tutorial/i18n/format/dateFormat.html).

<img width="543" alt="screen shot 2018-12-05 at 12 40 37 pm" src="https://user-images.githubusercontent.com/687227/49532641-00102200-f88b-11e8-900f-833c04da2e53.png">

In addition to allowing the time stamps in the Notification table on the Account pg, and the Version Details popup on the Dataset pg to be translated using the language, it will improve format consistency, as requested in _Date/Time Stamp Format - Consistency Across App #1391_.

kcondon:
@JayanthyChengan Loading the homepage or clicking on an existing dataverse results in a server log error about an unparseable date, though the page does load:

[2018-12-05T17:30:34.876-0500] [glassfish 4.1] [SEVERE] [] [] [tid: _ThreadID=51 _ThreadName=Thread-9] [timeMillis: 1544049034876] [levelValue: 1000] [[
  java.text.ParseException: Unparseable date: "Dec 5, 2018"
	at java.text.DateFormat.parse(DateFormat.java:366)
	at edu.harvard.iq.dataverse.util.DateUtil.formatDate(DateUtil.java:38)
	at edu.harvard.iq.dataverse.search.SolrSearchResult.getDateToDisplayOnCard(SolrSearchResult.java:889)
	at sun.reflect.GeneratedMethodAccessor2765.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at javax.el.BeanELResolver.getValue(BeanELResolver.java:363)
	at com.sun.faces.el.DemuxCompositeELResolver._getValue(DemuxCompositeELResolver.java:176)
	at com.sun.faces.el.DemuxCompositeELResolver.getValue(DemuxCompositeELResolver.java:203)
	at com.sun.el.parser.AstValue.getValue(AstValue.java:140)
	at com.sun.el.parser.AstValue.getValue(AstValue.java:204)
	at com.sun.el.ValueExpressionImpl.getValue(ValueExpressionImpl.java:226)
[...]

PR #5302, opened 2018-11-09
5300 dateformat

JayanthyChengan:
## Related Issues

- connects to #5300 

## Pull Request Checklist
 - Merged latest from "develop"  branch


src/main/java/edu/harvard/iq/dataverse/util/DateUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public static String formatDate(Date dateToformat) {
+        try
+        {
+            DataverseLocaleBean d = new DataverseLocaleBean();
+            Locale bundle_locale = new Locale(d.getLocaleCode());
+
+            String formattedDate = "";
+
+            if(bundle_locale.getLanguage().contains("fr"))
```
mheppler: With over half the world's population (*Wikipedia: https://en.wikipedia.org/wiki/Date_format_by_country) using the "dd MMM yyyy" format, there must be a more scalable solution than hardcoding "fr" here. @scolapasta might have some more detailed insight to share.

scolapasta: So 1) we should use DateFormatter rather than SimpleDateFormat. If we use it like below, then we don't need to have the conditional:
dateFormatter = DateFormat.getDateInstance(DateFormat.DEFAULT, currentLocale);
today = new Date();
dateOut = dateFormatter.format(today);

We could use either DEFAULT or MEDIUM. (for example DEFAULT for German would produce: 30.06.2009). If we want to have the text, I think we need to use MEDIUM in that case.

(also, the second method should call the first; i.e. convert the String to date, then pass that to the first method)

JayanthyChengan:
@scolapasta @mheppler - I have implemented the changes.  Please review the code.  Thanks

src/main/java/edu/harvard/iq/dataverse/util/DateUtil.java
```diff
@@ -0,0 +1,67 @@
...
+            return formattedDate;
+        }
+        catch(Exception e) {
+            return null;
+        }
+    }
+
+    public static String formatDate(String dateToformat){
+        String formattedDate = "";
+        DateFormat inputFormat = DateFormat.getDateInstance(DateFormat.DEFAULT);
```
scolapasta: @JayanthyChengan You changed the input format from hardcoded to the date instance dafault? Any particular reason? since this is how we store the date, it's fine to use the hardcoded way which matches how we store, I think. (it feels more readable to me). I can be convinced, though.

kcondon:
@JayanthyChengan  Thanks!

Issue #5165, opened 2018-10-11
resource leaks in S3AccessIO calls


qqmyers:
If I'm following things correctly, a call to something like isAuxObjectCached() (e.g. from ImageThumbConverter.isThumbnailCached()) is calling open() (no parameters, defaults to read mode), which for the S3AccessIO class goes ahead and opens an InputStream for reading. Since this call and others like it never access the InputStream, it looks to me like that stream is left open. A quick scan shows several methods that start with open() in the S3AccessIO class. I didn't see a similar problem in the File class and haven't checked Swift. 
A fix may be as simple as not opening the stream until the getInputStream() call is made...

qqmyers:
@kcondon - I didn't look for the getReadChannel() method after overriding the getChannel() method. That one is also now overridden to make sure the channel is created.

kcondon:
@qqmyers Yep, that did it, thanks!

PR #5305, opened 2018-11-12
5165-resource leaks in S3AccessIO calls

qqmyers:
## New Contributors

## Related Issues

- connects to #5165 resource leaks in S3AccessIO calls

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -651,14 +651,13 @@ public class Access extends AbstractApiBean {
                         || "application/zipped-shapefile".equalsIgnoreCase(df.getContentType())) {
 
                     thumbnailDataAccess = ImageThumbConverter.getImageThumbnailAsInputStream(dataAccess, 48);
+                    if (thumbnailDataAccess != null && thumbnailDataAccess.getInputStream() != null) {
+                        return thumbnailDataAccess.getInputStream();
+                    }
                 }
             }
         } catch (IOException ioEx) {
             return null;
-        }
-        
-        if (thumbnailDataAccess != null && thumbnailDataAccess.getInputStream() != null) {
-            return thumbnailDataAccess.getInputStream();
         }
 
         return null; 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -651,15 +651,14 @@ public InputStream fileCardImage(@PathParam("fileId") Long fileId, @Context UriI
                         || "application/zipped-shapefile".equalsIgnoreCase(df.getContentType())) {
 
                     thumbnailDataAccess = ImageThumbConverter.getImageThumbnailAsInputStream(dataAccess, 48);
+                    if (thumbnailDataAccess != null && thumbnailDataAccess.getInputStream() != null) {
```
pameyer: Not sure if it applies here or not, but in past leak hunting removed several `if (foo.getInputStream() != null)` type statements, where the anonymous open statements appeared to be correlated with accumulating open file descriptors.

qqmyers: @pameyer - I think it should be OK in this case since the StorageIO class caches the InputStream - repeated calls to it don't create more streams as it would if you were directly opening a file stream, etc.
FWIW - In checking on this, I did find that that the code retrieved the channel, created from a stream, in order to create a new stream, rather than just retrieving the stream directly. Added a fix as a new commit.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -215,6 +207,33 @@ public void open(DataAccessOption... options) throws IOException {
...
+        if (super.getInputStream() == null) {
+            throw new IOException("Cannot get InputStream for S3 Object" + key);
+        }
+
+        setChannel(Channels.newChannel(super.getInputStream()));
+
+        return super.getInputStream();
+    }
+    
+    public Channel getChannel() throws IOException {
```
matthew-a-dunlap: Should this `@Override` as well?

qqmyers: Yep...

Issue #5344, opened 2018-11-27
Flyway: Add the database migrator in order to make database migration happen automatically.


MrK191:
Hi, I've managed to make flyway work, now there will be no need to add upgrade scripts since the migration will happen, after user will boot the war file. 
Just tell me if you want this feature then I will make PR.

poikilotherm:
As this involves new deps in `pom.xml` it might be a good idea to rely on the merge of #5289, too.

MrK191:
> As this involves new deps in `pom.xml` it might be a good idea to rely on the merge of #5289, too.

Yeah I will wait for the merge and then apply all those rules to this PR.

pameyer:
If I'm remembering correctly, when flyway came up in previous discussions there was some uncertainty about the potential for certain postgresql versions being restricted to flyway enterprise.  Is this (still?) a concern?

pdurbin:
@MrK191 heads up that pull request #5317 was merged. I don't think you should be blocked on waiting for pull request #5289 to be merged if you are only adding one dependency to pom.xml. There are 9 pull requests in QA right now and I don't know what order they will be worked on.

dheles:
I'm in the process of an ansibilized upgrade to which I'm too committed to change horses now, but I will share that once DSpace incorporated Flyway, it made upgrading *much* easier, so i'd be very much in support of this approach moving forward.

pdurbin:
@dheles thanks for your input. You're making me think that I should take a look at DSpace's dev guide see how they explain to developers how to use Flyway. I read what @MrK191 wrote in his pull request but I'm still a little confused about what I'm suppose to do as a developer. It all feels a bit magic at the moment.

@MrK191 heads up that after pull request #5289 by @poikilotherm was merged there were merge conflicts in your branch so in 91f6574 I took the liberty of doing two things:

- resolving the merge conflicts
- adjusting your changes to pom.xml to be as small as possible to make code review easier. I overwrote pom.xml with the latest from "develop" and then put in the small changes I think you intended. Previously the "diff" was bigger because you included a number of formatting changes (long lines, whitespace, etc.) in your pull request when it first reached us.

landreev:
@MrK191 et al, 
So to recap, this looks promising and we are interested in incorporating FlyWay going forward. 
[...]

PR #5349, opened 2018-11-28
Flyway integration issue - #5344

MrK191:
## New Contributors

## Related Issues

- connects to #5344

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/flyway/StartupFlywayMigrator.java
```diff
@@ -0,0 +1,28 @@
...
+@Startup
+@Singleton
+class StartupFlywayMigrator {
+
+    @Resource(lookup = "jdbc/VDCNetDS")
+    private DataSource dataSource;
+
+    @PostConstruct
+    void migrateDatabase() {
+
```
poikilotherm: IMHO a check on `dataSource == null` should be done here or a `try/catch` for NPEs, so the user gets a reasonable error message. That should finally result in an `EJBException` getting thrown, so deployment fails.

src/main/java/edu/harvard/iq/dataverse/flyway/StartupFlywayMigrator.java
```diff
@@ -0,0 +1,28 @@
...
+import org.flywaydb.core.Flyway;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.Resource;
+import javax.ejb.Singleton;
+import javax.ejb.Startup;
+import javax.sql.DataSource;
+
+@Startup
+@Singleton
```
poikilotherm: While reading into Flyway in a Java EE context, I [stumbled over the annotation](https://web.archive.org/web/20180825121637/http://are-you-ready.de/blog/2017/06/08/integrating-flyway-with-java-ee-and-using-its-datasource)
`@TransactionManagement(value = TransactionManagementType.BEAN)`.

The explanation sounds reasonable - should this be added here?

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
```diff
@@ -2,40 +2,38 @@
...
+How to Create SQL
+-----------------
 
-If no SQL upgrade script exists, look at https://github.com/IQSS/dataverse/milestones to figure out the name of the next milestone and create a script using the naming convention above.
-
-As with any task related to Dataverse development, if you need any help writing SQL upgrade scripts, please reach out using any of the channels mentioned under "Getting Help" in the :doc:`intro` section.
+We assume you have already read the :doc:`version-control` section and have been keeping your feature branch up to date with the "develop" branch.
 
-Please note that we are aware of the problem of merge conflicts in the SQL update script as well as how the next version number can change at any time. Please see the :doc:`making-releases` section for how we are running an experiment having to do with release notes that might help inform an improvement of our process for developing SQL upgrade scripts.
+First, check https://github.com/IQSS/dataverse/tree/develop/src/main/resources/db/migration to see if a SQL numbers are not colliding with your's.
```
poikilotherm: IMHO we could opt for versions based on `VYYYYMMDDIDX` instead to avoid this for the majority of cases?.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
```diff
@@ -2,40 +2,38 @@
...
 .. contents:: |toctitle|
 	:local:
 
-Location of SQL Upgrade Scripts
--------------------------------
+Location of SQL Upgrade Scripts and Flyway SQLs
+-----------------------------------------------
 
-``scripts/database/upgrades`` is the directory where we keep or SQL upgrade scripts.
+``scripts/database/upgrades`` is the directory where we keep SQL upgrade scripts - legacy way of upgrading schema.
```
poikilotherm: Couldn't those be migrated into Flyway someway, too? It would be really cool to get people currently on older releases onto the latest release as easy as possible.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
```diff
@@ -2,40 +2,38 @@
...
-If you are creating a new database table (which maps to an ``@Entity`` in JPA), you do not need to create or update a SQL upgrade script. The reason for this is that we use ``create-tables`` in ``src/main/resources/META-INF/persistence.xml`` so that new tables are automatically created by Glassfish when you deploy your war file.
+How to Determine if You Need to Create SQL
+------------------------------------------
 
-If you are doing anything other than creating a new database table such as adding a column to an existing table, you must create or update a SQL upgrade script.
+If you are creating a new database table (which maps to an ``@Entity`` in JPA), you do not need to create. The reason for this is that we use ``create-tables`` in ``src/main/resources/META-INF/persistence.xml`` so that new tables are automatically created by Glassfish when you deploy your war file.
 
-How to Create or Update a SQL Upgrade Script
---------------------------------------------
+If you are doing anything other than creating a new database table such as adding a column to an existing table, you must create SQL.
```
poikilotherm: Maybe we should place a hint here that adding a value to an `@Entity` is just the same?
Maybe a practical example is usefull here, too.

pom.xml
```diff
@@ -245,6 +245,11 @@
             <artifactId>jbcrypt</artifactId>
             <version>0.3m</version>
         </dependency>
+        <dependency>
+            <groupId>org.flywaydb</groupId>
+            <artifactId>flyway-core</artifactId>
+            <version>5.2.1</version>
```
poikilotherm: Could you please opt for adding a version property here? Also, v5.2.3 has been released in the meantime.

pom.xml
```diff
@@ -718,7 +724,7 @@
             <plugin>
                 <!-- https://stackoverflow.com/questions/46177921/how-to-run-unit-tests-in-excludedgroups-in-maven -->
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.22.0</version>
+                <version>3.0.0-M1</version>
```
poikilotherm: I can't find good resources about what might be affected by this version change, as this is really new. The official release notes just show a few improvements/fixes, so everything should be just fine.

Dunno what IQSS policy is on this. You could upgrade to 2.22.1, too.

pom.xml
```diff
@@ -609,7 +615,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-war-plugin</artifactId>
-                <version>2.3</version>
+                <version>3.2.2</version>
```
poikilotherm: Wow, that's a big step forward. Any side effects to be expected?

pom.xml
MrK191: The reason for the upgrade is that the maven build was not passing on my pc, after reaserch I found out that it was realted to OpenJDK 8 bug for Debian and at the same time some bug in the surefire plugin which was exactly fixed in this version.

pom.xml
MrK191: I would have to do more search if anyone had problem with new version, I didn't encountered any during war building.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
```diff
@@ -2,40 +2,38 @@
 SQL Upgrade Scripts
 ===================
 
-The database schema for Dataverse is constantly evolving. As other developers make changes to the database schema you will need to keep up with these changes to have your development environment in working order. Additionally, as you make changes to the database schema, you must write SQL upgrade scripts when needed and communicate with your fellow developers about applying those scripts.
+The database schema for Dataverse is constantly evolving. As other developers make changes to the database schema you will need to keep up with these changes to have your development environment in working order. Additionally, as you make changes to the database schema, you must write SQL when needed and communicate with your fellow developers about making those SQLs.
```
landreev: This may be nitpicking, but I don't fully get why it is necessary to change "upgrade scripts" and "scripts" to "SQLs".  I don't see how it makes the text any better; and it's not even a proper use of the word. 
?

MrK191:
@landreev From my perspective the user first has to upgrade to the version prior to flyway with sql scripts. With new version flyway will do "baseline" on existing db schema and after that you can just add sql scripts.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
MrK191: I think it should be responsibility of the reviewer, there could be even more mess with this system, E.g. user makes PR in 2019.02.15 with same sql name (with id 1) it was merged in 2019.02.20, now user makes pr in 2019.02.16 (with id 1) it was merged in 2019.02.19. Now flyway will crash because it was out of order and not only you need to track id you also need to track dates now.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
MrK191: I think we should stick with the old way till the user is in the newest version, trying to be back-compatible IMO introduces too much mess, you would have to came up with some way to pass the previous app version and then make migration according to that. And the script which we have already does that.

doc/sphinx-guides/source/developers/sql-upgrade-scripts.rst
MrK191: I'm not exactly sure what do you mean by the same, adding a field in `@Entity` annotated class without any SQL script wouldn't work since our DDL Generetion is set to `create-tables` only.

Issue #5145, opened 2018-10-05
expand logging for draft edits


pameyer:
Short version - change log for draft datasets could use some additional information.

Longer version - `UpdateDatasetVersionCommand` (and possibly `CreateDatasetVersionCommand`) provide information about which dataset version was changed, and who changed it - but there doesn't appear to be accessible information about what was changed.  Not sure if it makes more sense to add this to the action log record `info` field, or add a new logger / log file.

pdurbin:
Related: #1002

pameyer:
@pdurbin Thanks - that is closely related (aka #1002 would eliminate the need for this).

pameyer:
It looks like special characters in metadata fields get mangled in the logs (eg - `Galápagos` is logged as `Gal?pagos`).  

Any ideas if this is a config issue or a code issue?

pameyer:
Thanks to @sekmiller for pointing out that special characters didn't get mangled on his system, so that is a config issue.

pameyer:
In my hands, 2e1d6073f1003a55689d6d9436d50b9568368622 is still losing newline characters in the edit log.

djbrooke:
Moving back to team dev based on @pameyer's most recent comment.

sekmiller:
@pameyer, can you attach a copy of the log file that your machine is producing?
 thanks!

sekmiller:
I was able to get the log to use the raw version of the metadata.

pameyer:
@sekmiller Thanks!

pameyer:
For future reference (aka - it didn't make it into the issue from when we initially talked it over): this code enables logging, and there isn't currently a setting to disable the creation of edit logs.  This is something that might get revisited in the future however.

PR #5388, opened 2018-12-10
5145 expand logging of dataset updates

sekmiller:
## New Contributors

## Related Issues

- connects to #5145 Expand logging of dataset updates

## Pull Request Checklist

- [x] Unit [tests][x] none
- [x] Integration [tests][x]: None
- [x] Deployment requirements, [x]: None
- [x] [Documentation][docs] None
- [x] Merged latest from "develop" [branch][x] and resolved conflicts


src/main/java/edu/harvard/iq/dataverse/batch/util/LoggingUtil.java
```diff
@@ -49,6 +52,21 @@ public static void saveJsonLog(String jobJson, String logDir, String jobId) {
 		    logger.log(Level.SEVERE, "Error saving json report: " + e.getMessage());
 	    }
     }
+    
+    public static void saveDraftEditLog(String changes, String logDir, String identifier, String datasetId) {
```
matthew-a-dunlap: I think as we add more logs, we should centralize this logic to be usable for all future logging. Having a "saveLog" that takes a relative directory name and a string to log would be more reusable. The code for creating the log name can be held in the bean that needs this.

This is somewhat self-serving as I'm going to write my own log that does almost all the same things.

Issue #5327, opened 2018-11-19
Do not require setup of PID credentials for development environments 


djbrooke:
Developers wanting to hack on Dataverse or evaluate Dataverse currently need to set up a test account for PID minting. This makes it harder to hack on Dataverse, so consider removing this dependency from the process.

Production installations and those wanting to test the full publish flow would still need some way to mint PIDs. 

Other ideas welcome, but the goal is reducing what's needed to work on and/or run the code in a development/test environment.

matthew-a-dunlap:
I just spun up a connected issue on not requiring an email server for development #5328

poikilotherm:
This *might* be related to #4106, #5082 and others. @4tikhonov and I discussed plans about implementing an external service for PID minting that is used via some REST API or similar. That would offload maintenance of PID providers from the main project and it might offer a solution to have a dummy/local only PID minter for demo/dev/... purposes available.

Maybe we can as @4tikhonov during the next Community Call how far they got at DataverseEU with this? (I had no time to look into this yet, because of priority goal #5292 etc.)

djbrooke:
We'll use internal mock implementation of a PID provider.

matthew-a-dunlap:
for what its worth, this is the current error on my dev machine:

![screen shot 2018-12-18 at 11 55 47 am](https://user-images.githubusercontent.com/8901015/50169671-0dca9c00-02bc-11e9-962e-2e18bbdce8e0.png)

PR #5399, opened 2018-12-13
add fake PID provider for devs #5327

pdurbin:
connects to #5327

connects to #5409

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
@@ -1,0 +1,66 @@
+package edu.harvard.iq.dataverse.pidproviders;
+
+import edu.harvard.iq.dataverse.AbstractGlobalIdServiceBean;
+import edu.harvard.iq.dataverse.DvObject;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.ejb.Stateless;
+
+@Stateless
+public class FakePidProviderServiceBean extends AbstractGlobalIdServiceBean {
+
+    @Override
+    public boolean alreadyExists(DvObject dvo) throws Exception {
+        return true;
+    }
+
+    @Override
+    public boolean registerWhenPublished() {
+        return false;
+    }
+
+    @Override
+    public List<String> getProviderInformation() {
+        ArrayList<String> providerInfo = new ArrayList<>();
+        String providerName = "FAKE";
+        String providerLink = "http://dataverse.org";
+        providerInfo.add(providerName);
+        providerInfo.add(providerLink);
+        return providerInfo;
+    }
+
+    @Override
+    public String createIdentifier(DvObject dvo) throws Throwable {
+        return "fakeIdentifier";
+    }
+
+    @Override
+    public Map<String, String> getIdentifierMetadata(DvObject dvo) {
+        Map<String, String> map = new HashMap<>();
+        return map;
+    }
+
+    @Override
+    public String modifyIdentifierTargetURL(DvObject dvo) throws Exception {
+        return "fakeModifyIdentifierTargetURL";
+    }
+
+    @Override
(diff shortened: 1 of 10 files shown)


src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.ejb.Stateless;
+
+@Stateless
+public class FakePidProviderServiceBean extends AbstractGlobalIdServiceBean {
+
+    @Override
+    public boolean alreadyExists(DvObject dvo) throws Exception {
+        return true;
```
poikilotherm: See my comment above.

This could be extended, too: when the fake provider would add a number to a simple database field, that could be used for constantly growing index number usable as a "real" faked PID. Otherwise things like UUIDs or similar could be used, which are garantied to be doublication free. It could also be an option to lookup the highest number in the database, increment and use. That is slow, but for demo / dev purposes this could be alright.

src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public boolean registerWhenPublished() {
+        return false;
```
poikilotherm: Maybe this should be a setting instead? It might be usefull during demos of the code.

src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public List<String> getProviderInformation() {
+        ArrayList<String> providerInfo = new ArrayList<>();
+        String providerName = "FAKE";
+        String providerLink = "http://dataverse.org";
```
poikilotherm: Could this be replaced with the dataverse FQDN/URL? That way we could generate fake, but working PID URLs pointing to the instance, couldn't we?

src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
```diff
@@ -0,0 +1,66 @@
...
+        providerInfo.add(providerName);
+        providerInfo.add(providerLink);
+        return providerInfo;
+    }
+
+    @Override
+    public String createIdentifier(DvObject dvo) throws Throwable {
+        return "fakeIdentifier";
```
poikilotherm: See above about providing fake but usable PIDs.

src/main/java/edu/harvard/iq/dataverse/pidproviders/FakePidProviderServiceBean.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void deleteIdentifier(DvObject dvo) throws Exception {
+        // no-op
+    }
+
+    @Override
+    public Map<String, String> lookupMetadataFromIdentifier(String protocol, String authority, String identifier) {
+        Map<String, String> map = new HashMap<>();
```
poikilotherm: Couldn't this at least return some stuff from the `DvObject` if the identifier is fake but usable?

pdurbin:
At standup just now I told @landreev and others that I'd leave a comment here to note that @poikilotherm and I have talked about his comments in IRC and he and I are on the same page with this pull request being an incremental improvement. In the future, if this fake PID provider proves useful to developers, we can hack on the code some more and make it better.

@sekmiller pointed out that the fake PID providers doesn't support file DOIs which is a good catch but I don't believe it needs to. The REST Assured test suite passes without it. If we ever need to add support for file DOIs, Steve said we need to write some code in the `publicizeIdentifier` method.

@matthew-a-dunlap said he'd do some code review (thanks!) and I'd like to point out that I added a "connect to" the new issue I just created about phoenix tests failing: #5409.

Issue #5400, opened 2018-12-13
Internationalization - Duplicate "html", "head" tags, broken 403, 404, 500 pgs after #5299 was merged


pdurbin:
After pull request #5299 was merged DatasetsIT.testPrivateUrl started failing.

Here's a screenshot:

<img width="1421" alt="screen shot 2018-12-13 at 6 34 38 pm" src="https://user-images.githubusercontent.com/21006/49974107-cfbd2900-ff05-11e8-84be-d0cdd1b99518.png">

https://build.hmdc.harvard.edu:8443/job/phoenix.dataverse.org-apitest-develop/306/edu.harvard.iq$dataverse/testReport/edu.harvard.iq.dataverse.api/DatasetsIT/testPrivateUrl/

pdurbin:
It's failing at https://github.com/IQSS/dataverse/blob/ce30a4f523ae82bf4f5e70e1f2a6c8464684dd62/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java#L772

with

`Caused by: java.lang.IllegalArgumentException: Unsupported assignee identifier 'doesNotExist'
	at edu.harvard.iq.dataverse.RoleAssigneeServiceBean.getRoleAssignee(RoleAssigneeServiceBean.java:91)`

at https://github.com/IQSS/dataverse/blob/ce30a4f523ae82bf4f5e70e1f2a6c8464684dd62/src/main/java/edu/harvard/iq/dataverse/RoleAssigneeServiceBean.java#L91

pdurbin:
These lines are failing...

        String title = getDatasetAsUserWhoClicksPrivateUrl.getBody().htmlPath().getString("html.head.title");
        assertEquals("Darwin's Finches - " + dataverseAlias, title);

... because we have two "html" tags on our dataset page:

![screen shot 2018-12-14 at 1 29 34 pm](https://user-images.githubusercontent.com/21006/50020583-650efa80-ffa4-11e8-9b9b-c557ba8da9cb.png)

This is obviously not great. @djbrooke and @kcondon I'd like to flag this as a show stopper for releasing 4.10.

pameyer:
@pdurbin I don't think this is due to having two html tags - the first one you've flagged is the doctype.

I think the problem is that the opening `<html>` tag is before the `<head>` tag; so the DOM doesn't have a valid `<title>`.

For what it's worth, I'd also vote to flag this as a show-stopper.

[...]

PR #5405, opened 2018-12-14
5400 duplicate html tags

pdurbin:
connects to #5400


pdurbin:
@pameyer thanks but yes this shouldn't go to QA. I spoke with @scolapasta about it. It's a confusing situation but I'm sure it'll get sorted out eventually.

mheppler:
This PR appears to be a little premature. Removing the `f:view` tag required for internationalization doesn't seem like a proper solution. It also only removes it from three pages, yet it still remains on 30+ other pages. Those other pages also have the duplicate `html` and `head` tags.

Vote to close.

pdurbin:
Closing in favor of pull request #5407.

Issue #5457, opened 2019-01-11
External tools can't see the dataset context


qqmyers:
Currently, the external tools interface only sends a reference for the relevant file and a tool can't easily discover which dataset and datasetversion the file is included in, and therefore, if I understand correctly, cannot retrieve the file metadata (name, tags, description) since these are associated with a give filemetadata entry that is specific the dataset version.

For QDR, we wanted to display the dataset name (with a link), and file creation date in the annotation viewing tool, so I extended the external tool interface to allow tools to request the dataset and datasetversion ids from Dataverse for inclusion in the called URL, mirroring the way the fileid, apikey, and siteurl can be requested.

I'll submit a PR so this can be merged in if there's broader interest.

pdurbin:
@qqmyers added documentation in 60f9b10. Thanks!

PR #5465, opened 2019-01-14
5457-provide dataset/version/filemetadata to External tools

qqmyers:
## New Contributors

## Related Issues

- connects to #5457 External tools can't see the dataset context

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
@@ -235,6 +235,10 @@ public class FileDownloadServiceBean implements java.io.Serializable {
                 dataFile = guestbookResponse.getDataFile();
             }
         }
+        //For tools to get the dataset and datasetversion ids, we need a full DataFile object (not a findCheapAndEasy() copy)
+        if(dataFile.getFileMetadata()==null) {
+            dataFile=datafileService.find(dataFile.getId());
+        }
         ExternalToolHandler externalToolHandler = new ExternalToolHandler(externalTool, dataFile, apiToken);
         // Back when we only had TwoRavens, the downloadType was always "Explore". Now we persist the name of the tool (i.e. "TwoRavens", "Data Explorer", etc.)
         guestbookResponse.setDownloadtype(externalTool.getDisplayName());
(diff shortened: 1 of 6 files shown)


src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
```diff
@@ -235,6 +235,10 @@ public void explore(GuestbookResponse guestbookResponse, FileMetadata fmd, Exter
                 dataFile = guestbookResponse.getDataFile();
             }
         }
+        //For tools to get the dataset and datasetversion ids, we need a full DataFile object (not a findCheapAndEasy() copy)
```
scolapasta: @qqmyers this looks fine but have you seen any performance issue here? If so, we could right some helper method like doesExternalToolNeedDataset(externalTool). But if there's trivial performance impact, not worth it.

qqmyers: @scolapasta - I think it's trivial - this is only called during tool launch and it is then a lookup of one datafile object. I didn't see any obvious delay launching a tool and I think the main concerns previously have been for cases where many datafiles were being looked up, e.g. when scanning through the list in a dataset...

scolapasta: OK< great. That is what I assumed as well, and why I had already moved to QA. :)

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
```diff
@@ -174,7 +174,9 @@ public JsonObjectBuilder toJson() {
         // from https://swagger.io/specification/#fixed-fields-29 but that's for URLs.
         FILE_ID("fileId"),
         SITE_URL("siteUrl"),
-        API_TOKEN("apiToken");
+        API_TOKEN("apiToken"),
+        DATASET_ID("datasetId"),
+        DATASET_VERSION("datasetVersion");
```
pdurbin: @qqmyers can you please add these two new reserved words (datasetId and datasetVersion) to the list of possibilities? The current list can be found at http://guides.dataverse.org/en/4.10.1/installation/external-tools.html

Issue #5434, opened 2019-01-04
File upload: large file uploads should show up in the file table after each ingest, before save


sbarbosadataverse:
See attached screen shot.
When uploading large files, after the blue bar is complete, files are not listed below where is says "file you upload will appear here"

<img width="1959" alt="screen shot 2019-01-04 at 1 01 26 pm" src="https://user-images.githubusercontent.com/8322346/50703071-e9dbba00-1020-11e9-87ea-d8e43e9582e1.png">

pdurbin:
I just wanted to note that Harvard Dataverse is running 4.9.4 at the moment.

pdurbin:
@sbarbosadataverse is this any better now that Harvard Dataverse is running 4.10.1?

The reason I thought to circle back to this issue is that @Mahsanas just reported a very similar issue with 4.9.4 in IRC:

"Hi, We seem to have a problem when uploading files larger than 100 MB. The upload progress bar completes, but the files does not move from the top widnow to the bottom winodw (where the uploaded files usually appear)."

Details at http://irclog.iq.harvard.edu/dataverse/2019-01-09#i_84292 and kudos to @pameyer for helping to troubleshoot.

qqmyers:
Just saw the chat log - PR #4522 started catching errors such as proxy timeouts that caused this, but I recently discovered that the changes to add provenance upload broke that error checking code. I'm working on a PR to fix it again. In the meantime, if there's a silent error, check for things like a 502 Proxy timeout error or other network error (visible in the browser log) - these cause a problem that the Dataverse server is not aware of and hence can't report.

qqmyers:
Just created PR #5466 - I think it is complete, but I fixed this as part of another branch, so it's possible I've missed some code. If anyone sees issues, let me know.

Mahsanas:
This was very helpful. I was able to look at the logs on my browser and see
that it is a 400 error (Bad Request) when uploading large files.
It might be coming from my Apache settings. If I remember correctly we did
not have any issue uploading large files before fronting glassfish with
Apache.

mheppler:
Suggested to @qqmyers that his PR gets connect to a new issue, specific to the error logging fix, so that this issue reporting large file upload problems can remain open, and continue to be diagnosed.

Confirmed with @sbarbosadataverse that she was still seeing these issues on Wed Jan 9, the day after 4.10.1 was deployed to production.

With @landreev returning to the office tomorrow, I am going to check in with him to see if he has any ideas.

[...]

Issue #5459, opened 2019-01-11
Support for retrieving and displaying Hypothesis annotations


qqmyers:
This is a placeholder issue for discussion: QDR has been using http://hypthes.is to allow researchers to annotate specific parts of published research papers. Those annotations are nominally part of the research record along with the data. With the data stored in Dataverse, QDR wanted a way to retrieve, store, and display these annotations as part of the dataset. @adam3smith can provide more detail on Hypothesis and what QDR has done with it. To support the integration with Dataverse, I've developed code mirroring the dropbox integration to allow a file containing annotations (in json), about a specific source document, and created within a specific group, to be uploaded into Dataverse, and to then use the external tools mechanism, extended as in #5456 and #5457, to create a previewer consisting of a static html page with associated javascript (which could be hosted externally, but which we just add as a resource in the Dataverse webapp).

These capabilities may not be of broad enough interest to include in Dataverse, but if there's interest, I can submit the code and develop documentation. (As with dropbox, the annotation retrieval code only shows if you configure with a hypothesis key, and the tool only appears if you register it as with other tools.). I'm hoping that what QDR has now will be out on the production server in the next few days, which may make it easier to understand what it all looks like...

pdurbin:
@qqmyers I just spoke with @kcondon about #5457 (merged) and #5456 (in QA) and I do think it would be nice to have Hypothesis described as an external tool on a future version of the list at http://guides.dataverse.org/en/4.10.1/installation/external-tools.html#inventory-of-external-tools

The tricky part is that we want the external tool authors to maintain the documentation and I'm not sure if Hypothesis would be open to this or not. Here's an example of how @whole-tale recently updated the list to include their external tool: pull request #5417 . If you know anyone at Hypothesis I'd be happy to chat with them at http://chat.dataverse.org or here or elsewhere about this.

qqmyers:
@pdurbin - This is definitely a tricky one. As of now, the ability to retrieve Hypothesis files is more tightly coupled - like the dropbox integration, while the display 'external tool' is so simple it's just an html template with a javascript. 

The former could potentially be pulled out some if the dropbox-style input became extensible (did the github integration do anything there?). That might at least allow the source to be managed separately (perhaps by Hypothesis) with users just having to drop a jar file in to use it with Dataverse. 

[...]

Issue #4410, opened 2018-01-11
Dataset Linking: As a dataverse owner, I want to link a related dataset to my dataverse to help relevant researchers find it.


dlmurphy:
This issue is narrowed down from the earlier, broader issue #2053.

Currently, only Super Users have the ability to link other users' datasets to their dataverses. This functionality used to be available to all users before Dataverse 4.0. Let's restore this functionality so all users are able to link others' datasets to their dataverses.

Issue #1840 also seems relevant.

pdurbin:
Huh, I didn't realize this was a feature of DVN 3.x but sure enough "Link Datasets/Studies" appears on the "3.6 & 4.0 Feature Comparison" spreadsheet at https://docs.google.com/spreadsheets/d/1ftHM6E4b9Dft_AvA6KTzeklJyLguItz2_0hNAFL8-VM/edit?usp=sharing

Please note that even superusers cannot delete links. That's what #1364 is about.

mheppler:
Today in Slack @thegaryking brought up this feature, requesting:
>trying to add an existing data set to my dataverse... the options here are new dataverse and new data set. can we add ‘existing dataverse’ and ‘existing dataset’?

Also, related to the feature's name, "link" he said:
>‘add’ sounds more coherent to the user i think. i’m just adding to my dataverse

All good feedback to consider when adding this feature for all dataverse admin's. Bonus feature request, also from Gary, to have the system suggest datasets to dataverse admin's:
>“we found this data set that’s like the ones in your dv, would you like to add it?“

Some related issues to also consider:
- Linked/Saved Search: Review Functionality + Usability Test #1840 
- Saved Search - Improve Performance & User Experience #1939 
- Linked Search: Creating linked searches fails in production with blank popup due to timeout. Works on smaller db's like demo. #2167

dlmurphy:
After much deliberation, some updates to this issue:

- This issue will cover dataset linking. There will be a separate issue to cover dataverse linking soon.

- The main impact of this issue is to open up dataset linking to all users, not just superusers.

- A user creating a new link will need their account to have a role that includes the "Add Dataset" permission on the dataverse that is doing the linking.

- A user creating a new link will not need any special permissions on the dataset BEING linked.

- Any user can link to any published dataset.

- For now, there will not be a way to remove links in the UI, but there is a currently existing way to do it via API.

[...]

PR #5466, opened 2019-01-14
5434 - File upload error reporting fix for large files

qqmyers:
which broke error reporting. This update should now add error reporting
to provenance uploads as well.

## New Contributors

## Related Issues

- connects to #5471 Broken error checking 

Relates to #5434 File upload: large file uploads should show up on file page after each ingest, before save

## Pull Request Checklist

### DIFF
--- a/src/main/webapp/editFilesFragment.xhtml
+++ b/src/main/webapp/editFilesFragment.xhtml
@@ -11,6 +11,7 @@
     xmlns:o="http://omnifaces.org/ui"
     xmlns:iqbs="http://xmlns.jcp.org/jsf/composite/iqbs">
 
+    <script type="text/javascript" language="javascript" src="#{resource['js/fileupload.js']}"></script>
     <!-- Static Tab Layout -->
     <div data-widget="content" class="ui-tabs ui-widget ui-widget-content ui-corner-all ui-hidden-container ui-tabs-top" id="datasetForm:tabView">
         <ul role="tablist" class="ui-tabs-nav ui-helper-reset ui-helper-clearfix ui-widget-header ui-corner-all">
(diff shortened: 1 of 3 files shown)


src/main/webapp/provenance-popups-fragment.xhtml
```diff
@@ -48,19 +49,11 @@
                                           fileUploadListener="#{provPopupFragmentBean.handleFileUpload}" 
                                           update="provPopupOutput,assignMessages" 
                                           label="#{bundle['file.editProvenanceDialog.selectToAddBtn']}"
+                                          onstart="javascript:uploadWidgetDropRemoveMsg();uploadStarted();"
+                                          onerror="javascript:uploadFailure();"
                                           oncomplete="bind_bsui_components();PF('provUpload').hide();"
                                           fileLimit="1" 
                                           invalidSizeMessage="#{bundle['file.edit.error.file_exceeds_limit']}"/>
-                            <script>
-                                function uploadStarted() {
```
qqmyers: FWIW - this uploadStarted() method was overwriting the one in editFilesFragment.xhtml...

src/main/webapp/editFilesFragment.xhtml
```diff
@@ -77,113 +78,12 @@
                                 </p>
                     
                     <script>
-//<![CDATA[
```
mheppler: Jim, any particular reason to remove this `CDATA` section from the script?

qqmyers: I think it was added when there were parse errors due to the content of the javascript. With most of that now in a separate file, I didn't think it was needed.

mheppler:
@qqmyers Could you please create a new issue, specific to the error logging vs provenance `uploadFinished()` fix, to connect this PR to, rather than #5434 which may be a larger issue.

Also, I have asked @matthew-a-dunlap to take a look at the provenance code to make sure he's OK with the changes.

qqmyers:
@mheppler - new issue and edited the PR to point at it - let me know if that's not what you wanted.

matthew-a-dunlap:
I have this branch running to test, specifically related to the provenance changes. TBH I'm not 100% sure of the need for that provenance js code in the first place.

That being said, I'm running into the attached error trying to do normal file uploads via http. Develop branch is working for me fine.

[uploaderror.log](https://github.com/IQSS/dataverse/files/2766200/uploaderror.log)

qqmyers:
@matthew-a-dunlap - as far as I can tell, the only thing this PR is doing is showing an existing (perhaps relatively new) issue: The error code is is getting triggered by PrimeFace uploading when it calls the onerror method of the p:fileUpload element. The HTTP status I see is 200, meaning the call went through fine (versus 502 errors for example). Without the javascript error code, this would be a silent issue on the browser side (so it could be occurring on develop and only showing in the log?)

The duplicate element appears to be related to dataset linking. I haven't followed what changes have been in that area lately, but I wonder if some change there is causing an issue when the page is being refreshed during upload. In testing last night, I only saw the error code show a 200 error, with similar log entries as you show, when I uploaded files in multiple actions (e.g. multiple hits on 'the select files to upload' button). I think I saw silent failures like this when I was working on issue #5459, before I repaired the error code, which would be consistent with it being an existing issue (again perhaps a recent one).

It's possible that there's some interaction with the javascript involved but I'm not sure how. The only code called before the error is returned is the uploadStarted(), which adds a 'upid' attribute to some nodes and starts mutators (event watchers) but doesn't otherwise change/add nodes (and it definitely doesn't touch the data linking code which is in the underlying dataset.xhmtl file rather than the editFilesFragment.xhtml. I don't think any code in this PR changes what parts of the page/form are refreshed, etc.

If the log entries are showing up in the develop branch, we should create a separate issue. If not, I can try removing parts of the uploadStarted() method to see what is triggering the issue. 

For the longer term, I could also change the error handler to not report status 200 errors, or to highlight them - thinking beyond the current issue, I think they will always mean a bug/something dataverse could/should handle, versus problems caused by proxies, etc. that are config/network issues, so they could be hidden (let the log report the problem) or flagged differently to ask users to report them to dataverse, etc.)

matthew-a-dunlap:
@qqmyers - Ahh I missed that it was a 200. My experience with this branch is that on the create datatset page the file uploads are now broken. I've had it succeed rarely, most of the time it throws the red error message with a 200 status code. Saving after that the file is not part of the dataset. Uploading after dataset creation seems to work every time. 

With the log messages, the duplicate component ID message may be a red herring. I think I'm mostly worried about the large javax.enterprise message hitting our logs at severe level. If that 750kbs message is intended we should probably set it to a lower level and only turn it on when we need to debug. Maybe this was in here before the code move-around but we don't want to fill up our logs too much.

qqmyers:
@matthew-a-dunlap - I think I found the issue: my change restored functionality that was evidently obsolete. The original uploadStarted method in editFilesFragment clicks the uploadStarted button, but the method in the prov popup that replaced it did not. Removing that (latest commit) appears to fix the error in create mode. I'll leave the error code where it will show a 200 error if one happens (this has no effect on what's logged, so it won't affect log size if some other 200 error shows up.)

matthew-a-dunlap:
@qqmyers Interesting! The prov popup code was not intended to override this functionality. Any idea why we didn't run into upload issues before adding the popup?

qqmyers:
@matthew-a-dunlap - not really, other than the possible hint that the duplicate item is in data linking and I thought (?) there were updates to that code after provenance was merged. If I"m following the code correctly that button only set uploadInProgress=true which is set true by the upload handlers anyway, so it's not clear to me that the call was needed. The button has update="", so I also don't really understand why making the call causes duplicate elements in the first place unless "" doesn't mean don't update anything.

mheppler:
@qqmyers @matthew-a-dunlap Since the linking popup on the dataset was resulting in the duplicate ID errors, I looked in there for what changed recently (see #4410).

When I started searching for causes of `Duplicate component ID` errors, I found a reference in this [PrimeFaces GitHub issue](https://github.com/primefaces/primefaces/issues/1691) which links to this [Stack Overflow question](https://stackoverflow.com/questions/2101755/binding-attribute-causes-duplicate-component-id-found-in-the-view) about "binding attribute causes duplicate component ID".

It looks like the new dataset linking popup code includes a `p:autoComplete` component with a `binding="#{DatasetPage.selectedDataverseMenu}"` attribute. I don't know enough about what it is used for, and if this is the cause of the excessive error logging, but it might be worth looking into.

To be fair, it looks like we use the `binding` attribute all over the app. This is the only instance of it on the `dataset.xhtml` pg though.

src/main/webapp/editFilesFragment.xhtml
```diff
@@ -377,7 +276,6 @@
                     
                     <p:message for="fileUpload" id="uploadMessage" display="text" />
                     
-                    <p:commandButton id="uploadStarted" action="#{EditDatafilesPage.uploadStarted()}" update="" style="display:none"/>
```
landreev: This was part of a legacy fix/hack... But it looks like we are not using it anymore for anything useful... So should be ok to remove as well.

src/main/webapp/provenance-popups-fragment.xhtml
landreev: @matthew-a-dunlap Could you confirm that this method was not needed, for the whole provenance upload thing to work? (looks like not - I'm not even seeing a button with the id "provUploadStarted"... but wanted to double-check)

matthew-a-dunlap: @landreev I am pretty sure it doesn't, I'm hoping to check in on it while waiting for my flight tomorrow.

matthew-a-dunlap:
I'm away on vacation / a conference this coming week. I may have time to dig into this more but it is unsure. If someone else has a chance to dig in, provenance can be enabled via command line http://guides.dataverse.org/en/latest/installation/config.html#provcollectionenabled

qqmyers:
@matthew-a-dunlap  - I've fixed the first error - the prov on start call shouldn't call uploadWidgetDropRemoveMsg (which deletes the 'drag and drop files here message on the upload files fragment). The second error -also fixed, is not specific to this branch. The prov oncomplete method is trying to hide the provUpload div (which has the select files button on the prov fragment) but it is deleted/replaced in the DOM by the 'upload complete'/preview/remove elements. So- hiding is not needed.

With these commits, I think everything found so far has been addressed.

matthew-a-dunlap:
@qqmyers This looks great! I tested many of the permutations of the prov functionality across dataset-create / file / edit-metadata and the all look good. Thanks for cleaning up those FileUpload components, this was my first major piece of jsf work and pieces got left behind as I was reusing code.

Issue #4970, opened 2018-08-16
Count of all file downloads from metrics API doesn't always match what's in UI (on old homepage metrics bar)


jggautier:
For some, but not all, Dataverse repositories running Dataverse 4.9.1-4.9.2, the count of all file downloads we get from [the new metrics API](http://guides.dataverse.org/en/latest/api/metrics.html#downloads-tomonth) doesn't match the download count of all files in an installation that's displayed in the metrics bar on the homepage:

![sefsef](https://user-images.githubusercontent.com/18374574/44231621-bb8e3880-a16c-11e8-9f07-38ef0f13a562.png) vs. `{"status":"OK","data":{"count":11120}}`

**Dataverses running 4.9.1-4.9.2 with different counts**
- Harvard Dataverse
  - Homepage count: https://dataverse.harvard.edu
  - API count: https://dataverse.harvard.edu/api/info/metrics/downloads/toMonth

- UAL Dataverse
  - Homepage count: https://dataverse.library.ualberta.ca
  - API count: https://dataverse.library.ualberta.ca/api/info/metrics/downloads/toMonth

- Qualitative Data Repository
  - Homepage count: https://data.qdr.syr.edu
  - API count: https://data.qdr.syr.edu/api/info/metrics/downloads/toMonth

**Dataverses running 4.9.1-4.9.2 with matching acounts**
- UNB Libraries Dataverse
  - Homepage count: https://dataverse.lib.unb.ca
  - API count: https://dataverse.lib.unb.ca/api/info/metrics/downloads/toMonth

jggautier:
I thought the difference might be caused by the query the API uses (or some version of it) and the number of entries in the database's guestbookresponse table that don't have timestamps:

_select to_char(date_trunc('month', guestbookresponse.responsetime), 'Mon YYYY') as months, count(guestbookresponse.id) AS new_datasets, 
sum(count(guestbookresponse.id)) over (order by date_trunc('month', guestbookresponse.responsetime)) as cumulative
from guestbookresponse
join dvobject on dvobject.id = guestbookresponse.datafile_id
where dvobject.publicationdate is not null
and guestbookresponse.responsetime is not null
group by date_trunc('month', responsetime)
order by date_trunc('month', responsetime) desc
limit 12;_

qqmyers:
@jggautier For QDR, it looks like the GUI counts entries where dvobject.publicationdate is null. Removing that clause from the query above makes the counts match.

jggautier:
[...]

Issue #5447, opened 2019-01-09
Homepage Count Updates


djbrooke:
- [x] Search input watermark dataset count (27400) is off by 53,827 compared to "Datasets (81,227)" facet (Expect to be a count of all dataverses and datasets that will be searched – including Harvested)
- [x] Browse by subject # expected to be dataverses+datasets including harvested
- [x] Allow queries for all datasets (harvested + native), as well as datasets harvested and files harvested. This really means implementing an ability to specify harvested or both for all (file/dataset) metrics.

matthew-a-dunlap:
As per discussion with @TaniaSchlatter , our metrics need to provide values for harvested / non-harvested . 
<img width="920" alt="50990068-168b4800-14df-11e9-9bd6-d5c68be91f88" src="https://user-images.githubusercontent.com/8901015/50990895-518e7b00-14e1-11e9-8b0e-a5d4d33c00ef.png">

matthew-a-dunlap:
Code Todo:
[ ] Add admin command to clear caches (needed with metric renaming happening)
[ ] Add query param to all metrics to specify local/remote/all (default is local for backwards compatibility)

djbrooke:
Closing in favor of #5445. Will do some group design QA and stakeholder verification in that issue.

Issue #5445, opened 2019-01-09
Dynamic Custom Homepage - ROUND TWO


mheppler:
_Misc HTML + CSS + layout improvements_
- [x] Make symbols in the big red buttons gray. See mockup.
- [x] Make the rules (lines) that are above "Find data across...", above “Activity” and above "Looking for other online..." bolder – if "hairline" then 1px, if 1px, then 1.5 or two ([see mockup](https://docs.google.com/document/d/1AtvODFY8aNQSE9BleeQEf5GkFsDWrziUB-IX-1xA0Y8/edit))
...
- [x] “Browse by subject” font to be the same as “Harvard Dataverse is a digital data repository…” 
...
- [x] Recent datasets: change titles to “Datasets from Journal dataverses” and "Datasets from other Dataverses”
- [x] Three columns of datasets  – one on left is journals, and two on right are latest not journals. Header runs across both “other” columns (see mockup, [previous version](https://github.com/IQSS/dataverse/pull/5249/commits/5256dc7cf46d97bfd370ec5811d2ade5e807a2ff))
- [x] Name of dataverse should be bold
- [x] Remove dash before date in display of datasets
- [x] "All recent journal activity" needs to go to search results showing journal dataverses and datasets, both facet boxes checked
- [x] "All recent researchers activity" link – Change text and behavior. Text to say: "All recent activity". Behavior to load the regular search page (same as “all subjects”)
...
- [x] Activity section: add more counts (UI), and have them be correct, responsive layout (see new mockup below, #5447)
...
- [x] "Looking for other repositories...?" section needs line break after DASH for responsive layouts

_Javascript fixes_
- [x] Handle cases with no JS through no script tag
- [x] Remove "Other" in JS
- [x] “Search over ###### datasets...” placeholder count needs a comma
- [x] Subject counts (1146) need commas
- [x] Activity counts (4936934) need commas
- [x] Search input watermark dataset count quickly blinks "75,000" before changing to comma-less count
- [x] Recent dataset dates (Wed Jan 09 2019) need comma (e.g. Wed Jan 09, 2019), remove day of week, remove "0" if the day is a single digit

_Other customization fixes_
- [ ] Header: add 2px above and below Harvard logo (or make the logo slightly smaller)
- [ ] Header & footer: change both to solid background #ececec

_Homepage template fixes_
- [x] Change page title from "Harvard Dataverse Dataverse" to "Harvard Dataverse"

_Additional curation efforts_
[...]

Issue #5398, opened 2018-12-13
Add Metrics API to get datasets by subject AND to month


scolapasta:
Currently, we can get datasets up to a certain month (and determine how many new by subtracting the previous month) or by subject, but one of the desired visualizations is to get by subject for a specific month, i.e. am metric to get to a certain month AND by subject.

pdurbin:
After standup today, @landreev and I discussed that while 0.1 of the metrics aggregator can handle the 6 existing metrics endpoints, it should be updated to handle the new 7th endpoint that's being created.

Here's the current config file for the metrics aggregator with the 6 endpoints: https://github.com/IQSS/metrics.dataverse.org/blob/v0.1.0/config.json.sample

landreev:
Made a pull request (#5451); this should *not* go into CR just yet - as I'm adding this new metrics to the aggregator in the other code repository, as described above. (thanks @pdurbin)

matthew-a-dunlap:
This work is going to be incorporated into the new homepage/metrics changes on #5398, as this code will be changed again in that PR they will be reviewed together along with #5447.

landreev:
@scolapasta 
@pdurbin 
@matthew-a-dunlap 
I modified the python scripts of the metrics aggregator (https://github.com/IQSS/metrics.dataverse.org) to support this new type of metrics. 
I had to come up with a new format for the aggregated .tsv file. 
We had the following 2 types supported: 
The monthly counts, that were aggregated like this (for example):
```
datasets-toMonth.tsv
month	count
...
2018-11	 1
2018-12	 4
2019-01	 6
```
and the totals by subject/category, like this (for example):
```
datasets-bySubject.tsv:
name	count
Agricultural Sciences	4
Social Sciences	2
```
So the by-the subject-and-by-month will now be aggregated like this:
```
datasets-bySubject-toMonth.tsv:
month	  name	count
2018-11	  Social Sciences	1
2018-12	  Agricultural Sciences	2
2018-12	  Social Sciences	2
2019-01	  Agricultural Sciences	4
2019-01	  Social Sciences	2
```

The PR for the aggregator code: https://github.com/IQSS/metrics.dataverse.org/pull/2. I requested review from Phil, since he wrote the original package.

kcondon:
The numbers from this new endpoint are internally consistent with bySubject when chosen to the current date. There is some discrepancy between the endpoint and facets but that will be handled within the homepage ticket. Closing this but will be merged when pr for homepage is merged.

[...]

PR #5475, opened 2019-01-17
5445 dynamic custom hmpg redux

mheppler:
## Related Issues

- connects to #5445 Dynamic Custom Homepage - ROUND TWO
- connects to #5447 Homepage Count Updates
- connects to #4970
- connects to #5398

## Pull Request Checklist


matthew-a-dunlap:
This PR includes changes for 4 different issues. As they were all metrics related, they were combined. These are the code changes:
- Metrics: All dataset and file endpoints support the dataLocation query parameter. Added to support new homepage needs in #5447
- Metrics: Metrics API to get datasets by subject AND to month, to support future visualization in #5398.
- Metrics: Querying download totals by month now returns the historic download counts that do not have a timestamp. Querying a date before the earliest timestamped record will return 0. Fixes #4970.
- Metrics: Metric table redone to support more metrics variations. Backend code was also rewritten a fair bit to match. Update script created to clear all caches and change the table.
- Metrics: Previously unreported bug solved where if the same metric is called twice when it needed to be cached, it would create two entries in the db. Subsequent calls would then throw an error about multiple caches. Multiple copies of the metric are still cached, but now when the system uses the cache it'll clear out duplicates and not error.
- Javascript: to get totals of datasets + dataverses by subject, the two separate metrics endpoints are called and summed on the user side. Running speed tests on this it seemed to take about 10ms. Developed for #5445

Its also worth noting that for the homepage to run optimally, the metrics cache timeout option should be set to update daily. The default is every 7 days.

PR #5487, opened 2019-01-24
Alias passwords #5412

### LINKED ISSUE #5412: Glassfish logs JVM options (including passwords) on startup
As appears to be its default behavior, Glassfish logs the JVM options to server.log on startup. Unfortunately, this includes doi and rserve passwords. I have been unable to find a way to avoid this.

Example log entry:
`[2018-12-19T18:40:00.967+0000] [] [INFO] [NCLS-GFLAUNCHER-00005] [javax.enterprise.launcher] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1545244800967] [levelValue: 800] [[
  JVM invocation command line:
...
-Ddataverse.rserve.password=[PASSWORD]
-Ddoi.password=[PASSWORD]
`
glassfish code responsible:
https://github.com/eclipse-ee4j/glassfish/blob/glassfish-main-aggregator-5.1.0-RC1/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherLogger.java#L159

The log level in question is only INFO, so I'd be happy to simply raise it to WARNING, but neither logging.properties nor asadmin list-log-levels seen to contain the relevant logger.

Alternately, moving the passwords to the database* would help.
* preferably encrypted

dheles:
Fixes #5412

### DIFF
--- a/scripts/installer/glassfish-setup.sh
+++ b/scripts/installer/glassfish-setup.sh
@@ -59,9 +59,9 @@ function preliminary_setup()
   for alias in "rserve_password_alias ${RSERVE_PASS}" "doi_password_alias ${DOI_PASSWORD}" "db_password_alias ${DB_PASS}"
   do
       set -- $alias
-      echo "AS_ADMIN_ALIASPASSWORD=$2" > ${FILES_DIR}/$1.txt
-      ./asadmin $ASADMIN_OPTS create-password-alias --passwordfile ${FILES_DIR}/$1.txt $1
-      rm ${FILES_DIR}/$1.txt
+      echo "AS_ADMIN_ALIASPASSWORD=$2" > /tmp/$1.txt
+      ./asadmin $ASADMIN_OPTS create-password-alias --passwordfile /tmp/$1.txt $1
+      rm /tmp/$1.txt
   done
 
     ###
--- a/scripts/installer/glassfish-setup.sh
+++ b/scripts/installer/glassfish-setup.sh
@@ -55,6 +55,15 @@ function preliminary_setup()
   ./asadmin $ASADMIN_OPTS create-jvm-options "-XX\:PermSize=256m"
   ./asadmin $ASADMIN_OPTS delete-jvm-options -client
 
+  # alias passwords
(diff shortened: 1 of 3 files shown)


pameyer:
On f88da2b07751045687fd7dc37127282062bbdcc1:

Non-interactive installer has several new warnings (looks like it tries to create temporary password files in the glassfish files directory, fails, and then fails to create the alias because the temporary file doesn't exist - rserve_password, doi_password, db_password).  Ends up with an installed application despite the warnings (but this may be due to lax postgres security in a disposable dev environment).

dheles:
Thanks for the catch, @pameyer . Adjusted to use /tmp for the password files. Tested in Ansibilized and straight Vagrant environments. AFAICT, everything checks out now.

scripts/installer/glassfish-setup.sh
```diff
@@ -65,7 +74,7 @@ function preliminary_setup()
   ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.rserve.host=${RSERVE_HOST}"
   ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.rserve.port=${RSERVE_PORT}"
   ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.rserve.user=${RSERVE_USER}"
-  ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.rserve.password=${RSERVE_PASS}"
+  ./asadmin $ASADMIN_OPTS create-jvm-options '\-Ddataverse.rserve.password=${ALIAS=reserve_password_alias}'
```
poikilotherm: Beware: there is a typo in `reserve_password_alias` - this should be `rserve_password_alias`.

pdurbin: @poikilotherm good catch! I fixed the typo in 5de567d. I have not tried this code myself.

landreev: I have tried it, it works.

pdurbin:
Just testing if I can see this comment in Slack.

Issue #5534, opened 2019-02-13
Logging: Verbose debug logging in server log happening on some machines post 4.10.1 release


kcondon:
Several developer machines and at least one test box are showing verbose debug logging in the server log, for example when loading the homepage or where dataset cards are rendered. It appears in other places too but we have not enumerated them. They seem to be lower-level HTTP messages and message within message.

Testing indicates they may have started appearing with this merge:
https://github.com/IQSS/dataverse/pull/5484 5483-netbeans-code-coverage

Here are some examples:

[2019-02-13T12:14:58.355-0500] [glassfish 4.1] [INFO] [] [] [tid: _ThreadID=83 _ThreadName=Thread-8] [timeMillis: 1550078098355] [levelValue: 800] [[
  12:14:58.355 [http-listener-1(51)] DEBUG org.apache.http.wire - http-outgoing-0 << "Fralin_Greek_Vases_20151120-kantharos.stl[0xff]"?[\n]"]]

[2019-02-13T12:14:58.354-0500] [glassfish 4.1] [INFO] [] [] [tid: _ThreadID=83 _ThreadName=Thread-8] [timeMillis: 1550078098354] [levelValue: 800] [[
  12:14:58.354 [http-listener-1(51)] DEBUG o.a.h.impl.execchain.MainClientExec - Connection can be kept alive indefinitely]]

[2019-02-13T12:14:58.222-0500] [glassfish 4.1] [INFO] [] [] [tid: _ThreadID=83 _ThreadName=Thread-8] [timeMillis: 1550078098222] [levelValue: 800] [[
  12:14:58.222 [http-listener-1(51)] DEBUG o.a.h.i.c.DefaultManagedHttpClientConnection - http-outgoing-0: set socket timeout to 600000]]

@qqmyers Any thoughts on this?

kcondon:
It looks like Apache httpClient logging is turned on in some way?
https://hc.apache.org/
https://hc.apache.org/httpcomponents-client-ga/logging.html

matthew-a-dunlap:
I'm snooping through our closed PRs now to see what may be indicative of this issue.

qqmyers:
@kcondon - I see this on one of our 3 machines at QDR and not the other two, even running the same code. I've looked through our config and haven't been able to find out why its occurring.  I recall trying to look at wire logging during development but didn't get it to work - so it's possible I added and forgot to remove some code or a jar via maven that didn't work for me but is popping up on some machines(some load order aspect?)

[...]

PR #5556, opened 2019-02-15
remove DuraCloud/Chronopolis feature (floods logs) #5534

pdurbin:
connects to #5534


pom.xml
```diff
@@ -543,44 +543,6 @@
             <artifactId>unirest-java</artifactId>
             <version>1.4.9</version>
         </dependency>
-        <!-- BagIt export -->
-        <dependency>
-			<groupId>org.apache.commons</groupId>
-			<artifactId>commons-compress</artifactId>
-			<version>1.18</version>
-		</dependency>
```
qqmyers: Unless this is the cause (probably isn't), I'd leave it (the commons-compress) in. It's used in the /BagIt export classes. They aren't called outside the DuraCloud archiver right now, but not having this dependency will break those classes.

qqmyers:
I think this should work. I suspect that the duracloud includes are turning this on somehow and that either excluding more classes, or adding a logback.xml file to turn off the extra logging can fix the problem, but just dropping it for now and waiting until I can discover the root issue should work. Note the one comment that there's a commons-compress include that is used by the bagit code that should probably stay.

djbrooke:
Thanks @pdurbin for the PR and @qqmyers for the discussion. This is good to have available and we'll reopen, QA, and merge this if we're pressed to cut a release, but I'm OK with closing it for now while @qqmyers (and @matthew-a-dunlap and anyone else) digs into the root cause of the increased logging.

Issue #5515, opened 2019-02-05
Add Admin API to change a username


scolapasta:
We want to add an admin API that can be used to change an account's username. This would require dealing with the role assignment table and the actionlogrecord, but other connections (like creator id, etc) should be accounted for by foreign keys.

pdurbin:
The specific SQL update statements can be found at https://github.com/IQSS/dataverse/issues/1445#issuecomment-251712762

We will, of course, want all SQL updates to happen in a single transaction. I'm actually not sure how to accomplish this. I'm interesting in learning because I feel like there are occasions when only some SQL statements are run, leaving "bad data" in the database from partially completed actions. I assume it's some EJB magic?

matthew-a-dunlap:
The branch for this issue was combined with #5514 as they both needed the same changes to actionLogRecord. Please refer to the new PR for any future changes.

kcondon:
see #5514 for a list of issues for both merge and rename.

Issue #5514, opened 2019-02-05
Add an admin API to merge accounts


scolapasta:
We want to add an admin API that can be used to merge two acounts (i.e. move one account into another). This would require changing foreign keys as well as dealing with the role assignment table.

matthew-a-dunlap:
I was hoping to get this done before my 2 weeks away from Dataverse, but it turned out to be a bit more complicated than I thought.

The code still needs tests and all the objects connected to the objects deleted in the command need to be altered. I'm not sure which need their foreign keys changed and which just need to be deleted. On friday Gustavo pointed me to https://github.com/IQSS/dataverse/blob/develop/scripts/migration/scrub_duplicate_emails.sql as reference for how we did this in the past but I didn't get to dig in. Specifically this part:

```
FROM information_schema.table_constraints tc
JOIN information_schema.key_column_usage kcu ON tc.constraint_name = kcu.constraint_name
JOIN information_schema.constraint_column_usage ccu ON ccu.constraint_name = tc.constraint_name
WHERE constraint_type = 'FOREIGN KEY' AND ccu.table_name='vdcuser';
```

It may be possible to manage this instead by using cascade inside java, investigating that was going to be my next step.

scolapasta:
One last thing, should we be changing the actionlogrecords associated with the merged username? (I'm not sure). At the very least we should add a describe to this command with both usernames since there is not dvobject associated.

matthew-a-dunlap:
The branch for this issue was combined with #5515 as they both needed the same changes to actionLogRecord. Please refer to the new PR for any future changes.

kcondon:
I'll use this ticket to record both rename and merge issues.
1. Renamed/merged user remains logged in if happened while logged in. Generally harmless since cannot save a new dataset but can download public files without incrementing count or recording download in guestbookrepsonse.
2. Cannot log in to merged account. No error in logs but ui does not allow log in as if incorrect password.

PR #5569, opened 2019-02-21
5515 change username api

matthew-a-dunlap:
## Related Issues

- connects to #5515

## Pull Request Checklist

- [x] Integration [tests][]


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1207,6 +1207,13 @@ POSTed JSON example::
       "email": "pete@mailinator.com"
     }
 
+Change User Identifier
+~~~~~~~~~~~~~~~~~~~~~~
+
+The body of the PUT request is the new identifier for the user. Changes identifier for user in ``AuthenticatedUser``, ``BuiltinUser``, ``AuthenticatedUserLookup`` & ``RoleAssignment``.
+
+    PUT http://$SERVER/api/admin/authenticatedUsers/changeIdentifier/$identifier
```
pdurbin: I would much prefer a full curl example showing what the body of the PUT request is supposed to look like.

matthew-a-dunlap: Done!

scolapasta: Is this the usual convention for these kinds of apis? Just seems easier to me if it was:
 PUT http://$SERVER/api/admin/authenticatedUsers/changeIdentifier/$oldidentifier/$newidentifier

but I'm not sure if that's atypical.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -310,6 +312,37 @@ public Response deleteAuthenticatedUser(@PathParam("identifier") String identifi
...
+            } else if(null == newIdentifier || newIdentifier.isEmpty()) {
+                return error(Response.Status.BAD_REQUEST, "New identifier provided to change is empty.");
+            }
+            
+            AuthenticatedUser authenticatedUser = authSvc.getAuthenticatedUser(oldIdentifier);
+            if (authenticatedUser == null) {
+                return error(Response.Status.BAD_REQUEST, "User " + oldIdentifier + " not found in AuthenticatedUser");
+            }
+            
+            BuiltinUser builtinUser = builtinUserService.findByUserName(authenticatedUser.getUserIdentifier());
```
scolapasta: do we need to check builtinuser? i.e. I could imagine a scenario where we would want to change the username for a Shib or Github user.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -310,6 +312,37 @@ public Response deleteAuthenticatedUser(@PathParam("identifier") String identifi
...
+            if (builtinUser == null) {
+                return error(Response.Status.BAD_REQUEST, "User " + oldIdentifier + " not found in BuiltinUser");
+            }
+            
+            List<RoleAssignment> raList = roleAssigneeSvc.getAssignmentsFor("@" + oldIdentifier); //only AuthenticatedUser supported
```
scolapasta: would it make sense to put this directly in the command? (rather than call before and send as a param), same with builtinuser if we decide we don't need to validate it exists...

src/test/java/edu/harvard/iq/dataverse/api/AdminIT.java
```diff
@@ -463,6 +463,46 @@ public void testConvertOAuthUserToBuiltin() throws Exception {
...
+        changeAuthIdResponse.then().assertThat()
+                .statusCode(OK.getStatusCode());
+        
+        String newUsernameBad = ""; //one character, should fail before bean validation even
+        Response changeAuthIdResponseBad = UtilIT.changeAuthenticatedUserIdentifier(usernameOfUser, newUsernameBad, superuserApiToken);
+        changeAuthIdResponseBad.prettyPrint();
+        changeAuthIdResponseBad.then().assertThat()
+                .statusCode(BAD_REQUEST.getStatusCode());
+        
+        String newUsernameBad2 = "z"; //one character, should fail bean validation
```
scolapasta: should we test trying to change it into a user that already exists?

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
matthew-a-dunlap: I went with this route because the commands don't have access to the service EJB's (tho maybe I misunderstood)

src/test/java/edu/harvard/iq/dataverse/api/AdminIT.java
matthew-a-dunlap: Yeah good call

scolapasta: Actually, beyond just testing, maybe make sure we catch it as an user friendly error.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
scolapasta: commands have access to ejbs via command context

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
scolapasta: so after discussion, let's move this get into the command (with the role assignment gets)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ChangeUserIdentifierCommand.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        
+        AuthenticatedUserLookup aul = au.getAuthenticatedUserLookup();
+        aul.setPersistentUserId(newIdentifier);
+        
+        for(RoleAssignment ra : raList) {
+            if(ra.getAssigneeIdentifier().charAt(0) == '@') {
+                ra.setAssigneeIdentifier("@" + newIdentifier);
+            } else {
+                throw new IllegalCommandException("Original userIdentifier provided does not seem to be an AuthenticatedUser", this);
```
scolapasta: Do we need this exception check? (raList should all start with "@", no?)

scolapasta:
One last thing, should we be changing the actionlogrecords associated with the old username? (I'm not sure). At the very least we should add a describe to this command with both usernames since there is not dvobject associated.

matthew-a-dunlap:
closing in favor of a branch combining #5514 and #5515 https://github.com/IQSS/dataverse/pull/5626

Issue #5207, opened 2018-10-17
Internationalization - Facet Category/Facet Label


JayanthyChengan:
Noticed the FacetCategory and FacetLabel remains in English and not toggling with language.
 
So, thought of adding those terms as key=value pairs in the bundle property files and implement as below in search-include-fragment.xhtml
 
 
https://github.com/IQSS/dataverse/blob/develop/src/main/webapp/search-include-fragment.xhtml#L177
 
<h:outputText value="#{facetCategory.friendlyName}" styleClass="facetCategoryName"/>
 
modify to
 
<h:outputText value="#{bundle[facetCategory.friendlyName]}" styleClass="facetCategoryName"/>
 

**Example:**

**FacetCategory:**
 
_#search-include-fragment.xhtml bundle[facetCategory.friendlyName]_
Dataverse\u0020Category=Catégorie Dataverse
Publication\u0020Date=Date de publication
Author-Name=Nom \u2014 Auteur
Subject=Sujet
Deposit\u0020Date=Date de dépôt
File\u0020Type=Type de fichier
File\u0020Tag=Libellé de fichier
Access=Accès
Keyword-Term=Mot-clé \u2014 Terme
Author\u0020Affiliation=Affiliation de l'auteur
Language=Langue
Kind\u0020of\u0020Data=Type de données
Publication\u0020Status=Statut de publication
 
FacetLabel:
 
Researcher=Chercheur
Research\u0020Project=Projet de recherche
Journal=Revue
Organizations\u0020or\u0020Institutions=Organisation ou établissement
Teaching\u0020Course=Cours
Uncategorized=Sans catégorie
Research\u0020Group=Groupe de recherche
Laboratory=Laboratoire
Agricultural\u0020Sciences=Sciences de l'agriculture
Arts\u0020and\u0020Humanities=Arts et sciences humaines
Astronomy\u0020and\u0020Astrophysics=Astronomie et astrophysique
Business\u0020and\u0020Management=Affaires et gestion
Chemistry=Chimie
Earth\u0020and\u0020Environmental\u0020Sciences=Sciences de la terre et de l'environnement
Engineering=Génie
Medicine,\u0020Health\u0020and\u0020Life\u0020Sciences=Médecine, santé et sciences de la vie
Computer\u0020and\u0020Information\u0020Science=Informatique et science de l'information
Law=Droit
Mathematical\u0020Sciences=Sciences mathématiques
Physics=Physique
Social\u0020Sciences=Sciences sociales
[...]

PR #5618, opened 2019-03-08
Internationalization - Facet Category/Facet Label

JayanthyChengan:
For both facet categories and Facet labels, key/value pairs are added to bundle files based on what solr returns and fetched into UI ( search-include-fragment.xhtml )

## Related Issues

- connects to #5207


src/main/java/Bundle_fr.properties
```diff
@@ -8,8 +8,8 @@ datasets=Ensembles de donn
 newDataset=Nouvel ensemble de donn�es
```
pdurbin: When will this entire Bundle_fr.properties file be removed? It's the plan for installations who want French support to get the Bundle_fr.properties file from https://github.com/GlobalDataverseCommunityConsortium/dataverse-language-packs ?

src/main/java/Bundle.properties
```diff
@@ -2289,3 +2289,932 @@ rtabfileparser.ioexception.boolean=Unexpected value for the Boolean variable ({0
...
+Organization\u0020or\u0020Institution=Organization or Institution
+Teaching\u0020Course=Teaching Course
+Research\u0020Group=Research Group
+Laboratory=Laboratory
+Department=Department
+Uncategorized=Uncategorized
+
+#astrophysics
+Image=Image
+Mosaic=Mosaic
```
pdurbin: Do we really want strings that were formerly in astrophysics.tsv (for example) in the main Bundle.properties file? Should they go in src/main/java/astrophysics.properties instead? Also, does metadatacustomization.rst need to be updated to mention facets and where to put the strings when creating a custom metadata block?

src/main/java/Bundle_fr.properties
JayanthyChengan: @pdurbin - Yes, I would like to create a separate issue to remove all french language files and Bundle_zh_CN.properties from codebase.  Also to move all the English property files into a specific folder.  Eg:  under src/main/java/english/*.properties , so this folder can be synchronized automatically (through Jenkins setup) to the English folder in dataverse-language-packs repository.  Later, translators can work for other languages.  Please add your comments.  Thanks

src/main/java/Bundle.properties
JayanthyChengan: @pdurbin - The key added to the bundle.properties are the values returned from solr for facets.  Is there any way to identify which metadata property file it corresponds to?

scolapasta:
closing in favor of #5697

PR #5621, opened 2019-03-11
use dataset thumbnail if available

qqmyers:
## New Contributors

## Related Issues

- connects to #5641 (edited by @djbrooke) 

## Pull Request Checklist


src/main/webapp/dataset.xhtml
```diff
@@ -43,7 +43,7 @@
                 <meta property="og:title" content="#{DatasetPage.title}" />
                 <meta property="og:type" content="article" />
                 <meta property="og:url" content="#{DatasetPage.dataverseSiteUrl}/dataset.xhtml?persistentId=#{dataset.globalId}" />
-                <meta property="og:image" content="#{DatasetPage.dataverseSiteUrl.concat(resource['images/favicondataverse.png'])}" />
+                <meta property="og:image" content="#{DatasetPage.dataset.getDatasetThumbnail() == null ? DatasetPage.dataverseSiteUrl.concat(resource['images/favicondataverse.png']): DatasetPage.dataverseSiteUrl.concat('/api/datasets/:persistentId/thumbnail?persistentId=').concat(DatasetPage.dataset.getGlobalId().asString())}" />
```
mheppler: @qqmyers -- one question I have about this `og:image` file path is if it requires the `dataverseSiteUrl` in your installation? In my initial local and AWS tests, it seems that a relative path with just `content="#{resource['images/...']}"` to the default image file works with no problem.

qqmyers: @mheppler - I'm not sure how you're testing, but it may be that relative URLs are OK for some uses. I think we went with a full URL based on https://stackoverflow.com/a/9858694 and looking at https://developers.facebook.com/tools/debug/og/object/ which gives an error such as 
image
with a relative URL. 

That said, Facebook also complains about the article:author field not being a profile (minimally a URL for the person), and doesn't like images less than 200x200... other things that could/should be fixed but that don't appear to cause trouble in slack for example...

mheppler: Ah, that might explain the lack of complaints on my end. I haven't checked my local development on the Facebook yet. (That's on today's agenda.) The dataset thumbnail image size was also something I was going to experiment today, so I am glad to hear you've already gone down that route.

qqmyers:
FWIW: I just resolve the conflict, but, with the changes to make the supplied images higher resolution, this could/should wait until one can specify a higher resolution dataset thumbnail as in #5679

pdurbin:
Over at https://github.com/IQSS/dataverse/issues/10220#issuecomment-1897046254 mentioned that this PR increases the dataset logo size (that's my understanding anyway), so it may be a solution for this issue:

- #9418

jp-tosca:
Hi @qqmyers 👋🏼 I was thinking that this being a solution on the JSP it would be better to implement this logic on `/api/datasets/:persistentId/thumbnail` so this solution could also be available to the SPA?

jp-tosca:
I opened https://github.com/IQSS/dataverse/issues/10586 to keep track of this for the SPA

jp-tosca (approved):
Looks good 👍🏼. Adding https://github.com/IQSS/dataverse/issues/10586 to keep track of this for the SPA. probably we need to remove this change from the JSF when the endpoint is updated.

PR #5626, opened 2019-03-12
5514 5515 merge and rename

matthew-a-dunlap:
## New Contributors

## Related Issues

- connects to #5514 
- connects to #5515

## Pull Request Checklist


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1207,6 +1207,27 @@ POSTed JSON example::
...
 
+
+Combining Accounts
+~~~~~~~~~~~~~~~~~~
+
+If a user has created multiple accounts and has been performed actions under both accounts that need to be preserved, these accounts can be combined.  One account can be merged into another account and all data associated with both accounts will be combined in the surviving account::
+
+    POST https://$SERVER/api/admin/authenticatedUsers/$continuingIdentifier/mergeIntoUser/$toMergeIdentifier
+
+Example: ``curl -X POST http://$server/api/admin/authenticatedUsers/toKeepIdentifier/mergeIntoUser/toMergeIdentifier``
```
scolapasta: Since the definition line uses terms like continuingIdentifier and toMergeIdentifier, I think the example would be more readable if it just used something more real world, e.g. 
//$server/api/admin/authenticatedUsers/jsmith/mergeIntoUser/jsmith1

Additionally (and my example I think helps illustrate this), it's a little strange to call it mergeinto and have the order be this way, since it reads the other way.

Thoughts? @sekmiller @matthew-a-dunlap @dlmurphy

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MergeInAccountCommand.java
```diff
@@ -0,0 +1,170 @@
...
+        if (consumedBuiltinUser != null){
+            ctxt.em().remove(consumedBuiltinUser); 
+        }
+        
+        
+    }
+    
+    @Override
+    public String describe() {
+        return "User " + consumedAU.getUserIdentifier() + " ( type: " + consumedAU.getAuthenticatedUserLookup().getAuthenticationProviderId() + " ) merged into " +ongoingAU.getUserIdentifier();
```
scolapasta: getAuthenticationProviderId is good but only provides "shib" and not more specific. We can either add in logic to parse the persistentuserid to get "Harvard IDP" OR maybe we just put in the whole persistentuserid (is that useful?) since that is onfo that would otherwise be lost?

@matthew-a-dunlap thoughts?

doc/sphinx-guides/source/api/native-api.rst
sekmiller: I don't have any strong feelings here. I think we tend to use "real world" examples elsewhere in the guide, as long as the descriptive prose along with the example clearly states which account is going away and which one is continuing with all of the data from both.

dlmurphy: I agree, in general an example should be as close to a real-life scenario as possible. So for this example, I'd use those jsmith and jsmith1 usernames; I think that's more instructive. Worth noting that the example for ChangeIdentifier has the same problem. Let's use jsmith and jsmith1 for that too.

Re: the other concern, I discussed with Gustavo and we're thinking the mergeIntoUser command should read left-to-right in the same way that changeIdentifier does. So I'd reverse the order of arguments in the mergeIntoUser command so the final retained username is used in the second argument.

scolapasta: Note that the "read left to right" change is a not just a doc change, but also a change to the API call itself, which currently works in the above right to left fashion.

matthew-a-dunlap: I definitely agree the text `mergeIntoUser` reads as if the command goes the other way. Order wise it feels weird no matter how its set up so I'm fine with that solution.

I can change the example as well.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MergeInAccountCommand.java
matthew-a-dunlap: I'm going to just add the whole persistentUserId, seems the easiest and most useful.

PR #5656, opened 2019-03-16
Fixed in password resetting email link.

adoal:
Use `siteUrl` to generate correct password resetting link instead of hard-coded 'https://...'
So that password resetting works if the site runs without https.


djbrooke:
Hey @adoal, thanks for the PR. We'll take a look early next week. I'll let more technical members of the team weigh in on the PR itself, but I know that we do highly recommend HTTPS for Dataverse. That being said, this may be a good change regardless.

Issue #5724, opened 2019-04-08
Breaking API changes in 4.12: RSpace (and other?) integrations broken


richarda23:
Hi
 Since 4th April our acceptance tests for the Java Dataverse SDK are failing running against demo dataverse, I expect this is due to updating to 4.12 . I'm wondering if there are any known changes to the returned data structure from a call to `https://demo.dataverse.org/api/v1/dataverses/<dataverseAlias>`? From the error messages we are getting it may be something to do with cardinality of `dataverseContacts` field.

Just more broadly, is there a published changelog of API changes when a new version of Dataverse is released? Is it possible to access a demo version of the previous version (4.11 in this case) for a short time after a new release for comparison? 

Thanks Richard

pdurbin:
@richarda23 hi! I'm sorry to hear we may have broken something. I guess is that pull request #5655 could have changed something you're relying on. I haven't looked deeply at it. If you spot something, please let us know.

When you say Java SDK I assume you mean https://github.com/IQSS/dataverse-client-java . It would be great to have some public continuous integration set up for it. It looks like we talked about this a bit at https://github.com/IQSS/dataverse-client-java/issues/5 . (I have some news I'll share on that issue.)

We don't publish any changelogs (we probably should, especially if you ask the https://keepachangelog.com people :smile: ). We do tag all issues and pull requests with GitHub milestones, if that helps. It's sometimes a lot to wade through.

IQSS generally only has the latest release in production and on our demo site. For public APIs like this I will admit that sometimes I poke around installations that are out there in the community. You can get a list from the bottom of https://dataverse.org/metrics

I hope this helps. Thanks for the bug report!

poikilotherm:
It should be fairly easy to get a v4.11 instance running with [IQSS/dataverse-kubernetes](https://github.com/IQSS/dataverse-kubernetes).

richarda23:
Hi Phil, thanks for quick reply

We have a lot of integrations with other software now and aren't really able to follow every single issue for all projects on the offchance that a change might affect the API.
An API changelog can be really simple/low effort, we just use a public GoogleDoc for ours to enumerate changes, e.g. https://docs.google.com/document/d/1HjZBnhPF3SQr3_Mt9wCMl625fH5hydifLp_Quo-3bTo/edit <https://docs.google.com/document/d/1HjZBnhPF3SQr3_Mt9wCMl625fH5hydifLp_Quo-3bTo/edit>. 

[...]

PR #5737, opened 2019-04-09
consistent format for dataverseContacts JSON #5724

pdurbin:
Connects to #5724.

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -39,12 +39,14 @@ Download the :download:`JSON example <../_static/api/dataverse-complete.json>` f
 
 .. literalinclude:: ../_static/api/dataverse-complete.json
 
+.. _view-dataverse:
+
 View a Dataverse
 ~~~~~~~~~~~~~~~~
 
-|CORS| View data about the dataverse identified by ``$id``. ``$id`` can be the id number of the dataverse, its alias, or the special value ``:root``. ::
+|CORS| View data about the dataverse identified by ``$id``. ``$id`` can be the id number of the dataverse, its identifier (a.k.a. alias), or the special value ``:root`` for the root dataverse.
 
-    GET http://$SERVER/api/dataverses/$id
+``curl $SERVER_URL/api/dataverses/$id``
 
 Delete a Dataverse
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


djbrooke:
Closing this test-only PR as the tests are included in #5739.

pdurbin:
Reopening as this is the "as usual" one we want (not a weird hotfix to master etc.)

pdurbin:
Maybe @michbarsinai knows how to make it work but it was way easier to just take it out.

michbarsinai:
@matthew-a-dunlap what lambda bug?

pdurbin:
@michbarsinai please see https://github.com/IQSS/dataverse/issues/5724#issuecomment-481729978 for my take. In 50c5c2d I modified the `public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts)` method, taking out

        return dataverseContacts.stream()
                .map(dc -> jsonObjectBuilder()
                .add("displayOrder", dc.getDisplayOrder())
                .add("contactEmail", dc.getContactEmail())
                ).collect(toJsonArray());

... and replacing it with this non-lamda function:

        JsonArrayBuilder jsonArrayOfContacts = Json.createArrayBuilder();
        for (DataverseContact dataverseContact : dataverseContacts) {
            NullSafeJsonBuilder contactJsonObject = NullSafeJsonBuilder.jsonObjectBuilder();
            contactJsonObject.add("displayOrder", dataverseContact.getDisplayOrder());
            contactJsonObject.add("contactEmail", dataverseContact.getContactEmail());
            jsonArrayOfContacts.add(contactJsonObject);
        }
        return jsonArrayOfContacts;

It's the weirdest thing. If you look at 50c0de8 (and earlier commit) you can see that the lambda version is emitting JSON for dataverse contacts from a unit test. But the later commit to remove the lambda was the quickest way to get the deployed app to emit JSON for dataverse contacts. I hope this makes sense. Again, in production the lambda would only ever emit an empty JSON array for dataset contacts. Weird.

michbarsinai:
Weird indeed. We had issues with lambdas on classes weaved by our JPA implementation. These were fixed in later versions of EclipseLink, but I'm not sure whether we upgraded or not. This is probably not directly related, since JsonPrinter is  not an entity class. But maybe some other byte code level mechanism is involved (e.g. dependency injection)?

pdurbin:
I have no idea but if we figure it out let's put it in a JavaOne^h^h^h CodeOne talk next year. 😄

Issue #5671, opened 2019-03-20
API Token and Data Curation Tool


lubitchv:
Data Curation Tool is an external Tool for editing variable level metadata. It needs to get an API Token to update data. But when users create account in Dataverse or an account is created for them API Token is not generated automagically . So then API Token is empty users cannot edit metadata and have to be reminded to create API Token.

But there is a possibility to generate such token automatically in  ConfigureFragmentBean in getConfigurePopupToolHandler of "Configure" button such that when the token is empty and user is authenticated the token can be generated automatically. In this case users would not lose the data that they started to update in DCT only to learn that they forgot to generate the token.

I would like to know what dataverse community thinks about it. Is it ok to generate token in Configure section? 
This issue is related to #4174 and #4448

scolapasta:
Hi @lubitchv we had a previous issue for this, #4310 where at the time we were leaning towards having the user "opt in" to the token being created. This would of course require a UI around this opt in capability. I think having it be automatically generated would be fine too, but there should probably be messaging so the user is aware of it happening. (since an api token is effectively another password and the user needs to understand it's security needs).

I'd have to refresh my knowledge on the code, but I think getConfigurePopupToolHandler, just by displaying the button? if so, I'm not sure if that's the best place for it, since it would then be generated for *every* user who visits a dataset. Unless we're ok with that, but if so, we should probably just generate at user creation time.

(I could be mis recalling when that code is called)

pdurbin:
My first thought is this:

- Users who are created via the GUI do not automatically get an API token.
- Users who are created via API do automatically get an API token.

This is inconsistent and was observed recently in 5f383da ("not all users have apiTokens").

Should we introduce a system-wide boolean setting called something like `:GenerateApiTokenOnCreate`? Installations of Dataverse that have security concerns with API tokens being created could set this to "false". Developers would set this to "true" because API tokens are used in integration tests. Scholars Portal and others could set it to "true" to make sure their users all have API tokens. (Perhaps an API endpoint could also be added to generates API tokens for existing users.)

These are my first thoughts but I'd love to hear some discussion from the community about this! 😄

lubitchv:
@scolapasta  not exactly. getConfigurePopupToolHandler has a check for fileId is null or toolHandler is not null.In that case it does not go to retrieve API Token. This is also the case when user just visit dataset. I suggest to add the code in case retrieval of Api Token brings null. It should only work if user actually press on External Tool in drop down.

[...]

PR #5755, opened 2019-04-11
5671 api token

lubitchv:
When a user use External tool in Configure button, the new API Token is generated. The message popups to confirm or cance. When token is generated an e-mail is sent to the user with a warning that it should be treated like a password.

The email, subject and the message  are added in Bundle.properties and will need to be translated to other languages. 

## Related Issues #5757 

- connects to #5671

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/ConfigureFragmentBean.java
```diff
@@ -35,15 +41,21 @@
...
     @EJB
     AuthenticationServiceBean authService;
+    @EJB
+    UserNotificationServiceBean userNotificationService;
     
-    public String configureExternalAlert() { 
+    public String configureExternalAlert() {
+        generateApiToken();
+        String httpString = "window.open('" + toolHandler.getToolUrlWithQueryParams()+  "','_blank'" +")";
+        RequestContext.getCurrentInstance().execute(httpString);
```
qqmyers: I think this is deprecated. 
PrimeFaces.current().executeScript(httpString);
should do the same thing...

lubitchv: You are right. I replaced RequestContext with PrimeFaces as you recomended.

src/main/java/edu/harvard/iq/dataverse/ConfigureFragmentBean.java
```diff
@@ -55,9 +51,9 @@
     
     public String configureExternalAlert() {
         generateApiToken();
+        PrimeFaces.current().executeScript("location.reload(true)");
```
scolapasta: @lubitchv could you briefly explain what this change does?

lubitchv: It reloads the dataverse page. Without reloading the page one would not see the notification. There is also another reason for that. If user has several files in DataSet and he uses external tool for the first file and later returns for the page to use external tool for the second file, the fileId will remain of the first file end external tool will show the user content of the first file and not of a second. It confuses users tremendously. We had an experience that many of users do not pay attention to the message that asks them to reload the browser. So I thought it is better to do for them automagically.

scolapasta: @lubitchv OK, we'll have @kcondon test that and see how it works. Can you also then remove the message from the bundle (I'm assuming, of course, it was only used in this one place)?

Issue #3575, opened 2017-01-17
Usernames are case sensitive


pdurbin:
@dlmurphy asked the following at https://github.com/IQSS/dataverse/pull/3539#issuecomment-270491499

> Under Create Account it mentions that the Username field supports characters "a-Z". This implies that uppercase letters are allowed and that Dataverse distinguishes between lowercase and uppercase letters in usernames. Are both of these implications accurate? Would bob11 and BOB11 be separate usernames?

I posted my answer at https://github.com/IQSS/dataverse/pull/3539#issuecomment-270658359

> I tested bob11 vs. BOB11 and it's apparently allowed. There's always a "dataverseAdmin" account and I was able to create a second account as "DATAVERSEADMIN". I'll let you think about if this is a bug of a feature but I will say that we disallow this for the "alias" of a small d dataverse.

We discussed this a bit more today an I'm fairly sure that case-insensitive usernames is a usability issue. I can imagine a poor user saying "When you assign me permission to your dataset please be sure to choose `@jharvard` and not `@JHARVARD` because the upper-case version isn't me!"

#1445 is highly related.

dlmurphy:
Thanks for making this issue, Phil.

Another concern is the potential for abuse. Example: someone with the username @MichelleObama (with two lowercase L's) could be spoofed by someone with the username @MicheIIeObama (with two uppercase i's). It's also possible that this sort of confusion could arise by accident as well.

pdurbin:
Yeah, we should probably fix this at some point but it hasn't been a priority. Let's make a new issue when we decide to pick this up.

dliburd:
This issue still exists in Dataverse. Also Julian and I noticed that the search field on the Dataverse permissions page is case sensitive. Eg: searching for "bob" will not find the "Bob11" account.

Just looking at usernames that begin with the letter 'a', I came across 16 instances of usernames that are identical except for capitalization differences.

djbrooke:
We should display our regular message about the username already being taken when someone tries to create a username with different capitalization. 

<img width="923" alt="screen shot 2019-02-13 at 10 51 04 pm" src="https://user-images.githubusercontent.com/7773392/52761597-fa009500-2fe1-11e9-8fee-a94ef6631a21.png">

pdurbin:
I mentioned in standup that it would be great if JPA supported a case insensitive uniqueness constraint so I just opened https://github.com/eclipse-ee4j/jpa-api/issues/209

[...]

PR #5783, opened 2019-04-23
make usernames case insensitive #3575

pdurbin:
connects to #3575


pdurbin:
> Maybe update the named query countOfIdentifier on AuthenticatedUser to compare on lowercase.

Ah, like you did in 95a8086. Thanks! I'll take a look.

pdurbin:
> fails with no feedback

Here's the stack trace as of 0a0dc25: [stacktrace-0a0dc25.txt](https://github.com/IQSS/dataverse/files/3116994/stacktrace-0a0dc25.txt)

sekmiller:
When I said no feedback I meant that if you try to create an account that has an identifier that is already taken, but with a different case from the existing account, the save fails at the database level and the user doesn't get an error message. It just fails. and they can hit Save again and wonder why nothing is happening. If you add "LOWER(" to the named query I mentioned above then the user sees an error message of "This username is already taken".

pdurbin:
@sekmiller yes. I get you. Thanks! I just pushed 392d26e. Can you please take a look and move it QA if you feel like we're ready? Thanks!

sekmiller:
The test for existing usernames on account create looks good.

Also, if we're saying that username is case insensitive we should probably allow you to log in whatever case you like, so if my username is sekmiller I should also be able to log in as SekMiller or SEKMILLER, etc.

pdurbin:
@sekmiller you're saying I should be able to log in with DATAVERSEADMIN. Obviously, this only applies to built-in accounts, not Shib or OAuth.

@djbrooke thoughts?

scolapasta:
Not Danny, but yes, we should allow this. If we're saying usernames are case insensitive then we should treat them that way functionally. (it's ok to preserve a user's chosen format for presentation). We do the same for dataverse alias, e.g. both dataverse.harvard.edu/dataverse/PSI and dataverse.harvard.edu/dataverse/psi will get you to PIS's dataverse.

djbrooke:
@pdurbin thanks for tagging me. I think it's fine if people log in with DaTaVeRsEADMIN vs dataverseadmin vs. whatever other capitalization. 

Regarding the title of this PR - is there any internationalization/localization concern forcing lower case?

scolapasta:
@pdurbin @sekmiller one more note - we should basically find all places in the code that look for username and make sure they're case insensitive - create user and user login are the two obvious ones, but another one that comes to mind is when you are assigning roles and looking for the user. (a good approach here would be to look for all the queries that involve username)

pdurbin:
I suppose a corollary of all this is that I should be able to use the changeIdentifier API to rename dataverseAdmin to DATAVERSEADMIN. Same case but stored differently in the database.

I'm beginning to feel like this issue needs more definition of done.

djbrooke:
@pdurbin sounds good! Happy to discuss tomorrow.

scolapasta:
Sure. I think this is lower priority, but at some point someone may want to change it and it should be an easy add. (ChangeUserIdentifierCommand first checks if the newusername already exists, so we could add a 2nd check there to see that if it does, it still allows the change if it's just a case variant of the old username). Should be a one line change.

scolapasta:
I went ahead and made the change, though used slightly different logic:
```if (authenticatedUserTestNewIdentifier != null && !authenticatedUserTestNewIdentifier.equals(au)) {```

kcondon:
@pdurbin Please update from /develop -version was changed. Thanks!

pdurbin:
@kcondon sure, done in cb95216a3

Issue #5796, opened 2019-04-25
UZH: Unit Testing (Mocking)


rschlaefli:
This issue is about improving tests that could benefit from mocking (see #5732). See #5619 for additional details.

poikilotherm:
I just added some notes to PR #5797 Feel free to get inspired or ignore.

PR #5797, opened 2019-04-25
5796 mocking

rschlaefli:
This PR adds initial mocking with Mockito in several places that I found it could be useful.

**Important**: The larger test suite depends on a change to **production** code in that the entity manager in `passwordreset/PasswordResetServiceBean.java` needs to be *protected* instead of *private*. I found that throughout the codebase, it is already often the case that the EM is only *protected*, so found it reasonable to change this for improved testability. If you prefer or use any other kind of dependency injection in such cases, I can still amend this. Without this production change, we could not unit test or *cleanly* mock anything related to database interactions.

As for the specific changes:
- Add a test suite for `passwordreset/PasswordResetServiceBeanTest.java`
  - Executes/covers 86% of instructions compared to 0% before the change
  - However, this is only a simple metric and the assertions and evaluations are on a relatively rudimentary level (as far as I got with limited understanding of the entire codebase)
    - For example, the test ensures that an email would have been sent or not sent in the respective cases, or that the returned object correctly includes `isChanged` or `isEmailFound`
  - All externalities that are used within the class are mocked
- Add initial test suite for `makedatacount/DatasetMetricsServiceBeanTest.java`
  - Test different behaviors of `testGetDatasetMetricsByDatasetMonthCountry`
  - The entity manager is already accessible here, so no need for a change

## Related Issues

- connects to #5796

## Pull Request Checklist


poikilotherm:
Just my 2 cents about dependency injection during mocking:

There seem to be two "schools" how to do mocking: implicit and explicit declaration.

* Implicit: make class properties accessible to test code to inject mocks.
* Explicit: create a constructor that receives mock objects and injects them during object creation time. (see an example [test here](https://github.com/peerpub/peerpub/blob/1b6ad4b30496fda364d9a34ebbc58c9e891b5bb5/src/test/java/de/fzj/peerpub/doc/doctype/DocTypeAdminCtrlTest.java#L59) and [class here](https://github.com/peerpub/peerpub/blob/1b6ad4b30496fda364d9a34ebbc58c9e891b5bb5/src/main/java/de/fzj/peerpub/doc/doctype/DocTypeAdminCtrl.java#L59))

Both have their supporters. I am counting me to the "explicit faction", as I like it clean and explicit. Creating the test object should be shorter to write, too. I also like to use things like [Lombok](https://projectlombok.org), which can be a real timesaver for these cases with many properties. It also will reliably break your build when you forget to refactor the test when refactoring the class.

There are a lot of articles and posts on this out there... Again, just my 2 cents from a community member.

My 2 cents on how you create the mocks:

Please cosnider taking a look at Jupiter Mockito Extension. It really saves you from creating a lot of the boilerplate in setup...  Good tutorial [here](https://www.baeldung.com/mockito-junit-5-extension).

rschlaefli:
Hi @poikilotherm,

Thanks for the inspiration! I did not know about the Jupiter Mockito extension, that would really save a lot of code. I will add it to this PR since I already use JUnit 5 anyways :)

About the implicit vs. explicit declaration: I took the path that I thought was more coherent regarding other production classes (many already have protected properties that are accessible like this, so I made a private one protected without any other production source changes). I thought that the maintainers would be more critical regarding the production change if I were to create an entirely new constructor or add other explicit dependency injection primitives.

I agree that the approach here ("implicit") does not feel very "clean". I think how to do it should be a general decision to be made by the maintainers, as mocking does not yet seem to be used very often throughout the codebase, and a general approach should be defined (@pdurbin)? Also, something like Lombok is not currently used, as far as I can see (but might be considered?).

pdurbin:
@rschlaefli I just looked at your pull request as of b45d4d5 and it looks fine. I'm wondering why you had to change "private" to "protected" in the code itself but I'm not especially bothered by that. I don't have a strong opinion about implicit vs explicit. I'm happy to see that code coverage is up. Are you done making commits? If so, I'll drag this over to QA. If not, I'll leave in in code review until you're done. Thanks!

@poikilotherm thanks for weighing in!

rschlaefli:
Hi @pdurbin,

Thanks for the feedback.

> I'm wondering why you had to change "private" to "protected" in the code itself but I'm not especially bothered by that. I don't have a strong opinion about implicit vs explicit. I'm happy to see that code coverage is up.

Did you look at the latest commits or b45d4d5 specifically? The later commits made it necessary to make the production change in `passwordreset/PasswordResetServiceBeanTest.java` to be able to "implicitly" inject the dependency I wanted to mock (i.e., the `EntityManager`). Otherwise, I would not have been able to test the component without it trying to make real database requests.

If you do not want me to change anything regarding explicit mocking or code refactoring, you can move this to QA. Maybe http://guides.dataverse.org/en/latest/developers/testing.html could be extended if you develop any kind of preference in the future such that there is no uncertainty on how mocking should be approached?

pdurbin:
@rschlaefli I looked at "files changed" (the diff) and the amount of code coverage increase. I agree with you that if we develop a preference some day we should add it to that testing page. Thanks for the pull request!

Issue #2073, opened 2015-04-23
Moving Datasets: Provide way in the UI for a user to do this


eaquigley:
Have discussed this in the past but did not do if for 4.0. If a user uploads their dataset into the wrong dataverse, could we have a way in the UI for them to change this. 

Additionally, I'd like to track how many RT tickets we have about this now that 4.0 is out and see if it is happening as frequently as it was in 3.6 to determine how necessary this is now.

eaquigley:
This needs to be designed during 4.0.2 to be developed during 4.0.3.

eaquigley:
Things to think about:
- When a user moves a dataset into a dataverse, if that dataverse doesn't have the metadata domains the dataset uses, the user needs to be told to go to the dataverse and enable the metadata domains needed for the dataset.

pdurbin:
We used to have a "move dataverse" command but it was problematic and #283 is still open. The challenges listed in the Google doc mentioned in that issue still apply: [Implications of moving dataverses and datasets](https://docs.google.com/document/d/1W8AXMtFJ32XCILCe1PX2BTIER6hxGiOLZMLhbFPQe7A/edit?usp=sharing).

It sounds like in production we've moved a few things manually. I'd be interested in knowing what the steps have been.

pdurbin:
See also the "How to move dv's, ds" doc at https://docs.google.com/document/d/1A2v_ylLIRLIV6UpkDAJNNiFQt7q3SrAtAvQmg4nHBoY/edit?usp=sharing mentioned at https://groups.google.com/d/msg/dataverse-community/polE7z9_-NE/3xwAxgIHGgAJ

Here are the related issued:

- Moving Dataverses + Datasets: Provide way in the UI for a user to do this - https://github.com/IQSS/dataverse/issues/2073
- Move Dataset: Allow super user to move a dataset to another dataverse, one at a time. - https://github.com/IQSS/dataverse/issues/2226
- Move Dataverse: Allow super user to move a dataverse and reindex it and all child objects. - https://github.com/IQSS/dataverse/issues/2278

PaulBoon:
We at DANS felt the urge to have a GUI to do the moving of datasets and therefore we have developed a separate web application that must be deployed on the same Glassfish as Dataverse. 
Our Github repo: https://github.com/DANS-KNAW/dataverse-datasetmover
It uses the 'move' API introduced in version 4.8.6. 

It would however be much better if this functionality was available in the Dataverse GUI. 
[...]

PR #5798, opened 2019-04-25
2073 dataset move ui

pdurbin:
connects to #2073


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1020,7 +1021,11 @@ public Response moveDataset(@PathParam("id") String id, @PathParam("targetDatave
                     ));
             return ok("Dataset moved successfully");
         } catch (WrappedResponse ex) {
-            return ex.getResponse();
+            if (ex.getCause() instanceof UnforcedCommandException) {
+                return ex.refineResponse("Use the query parameter forceMove=true to complete the move.");
```
sekmiller: Can we add this response text to the Bundle>

pdurbin: @sekmiller as I mentioned in person, I had this API only message in the bundle originally but decided to take it out because the rest of the API only messages are hard coded as English in the Java. Like we said, let's get a ruling at standup. 😄

djbrooke: We decided that we want these API only messages in the bundle. We'll need a task in the future to make this change throughout the API, and in the meantime we'll pick these off for new APIs and for APIs that we're working on.

pdurbin: In 926e955 I moved the English in the API I touched to the bundle. In 4c76181 I encouraged other developers to do the same with a note in the dev guide.

Issue #2202, opened 2015-05-26
Dataset - Too Many "Unknown" Files, Friendly File MIME Type Display Names


mheppler:
As referenced in #2192, there are files in production that need friendly MIME Type labels.

From @landreev 

> The file in question is ./src/main/java/MimeTypeDisplay.properties

We should identify as many of these as possible, and give them friendlier display names that the one that @pdurbin  found.

![documentation_and_metadata_-_training_materials_dataverse_-_2015-05-24_10 00 24](https://cloud.githubusercontent.com/assets/21006/7788123/327273d0-01fc-11e5-8b89-608b7b05a7d8.png)

landreev:
Also, I believe we should extend this "friendly name" functionality, to support wild cards. 
As in: 

image/jpeg=JPEG Image
image/gif=GIF Image
image/bmp=Windows Bitmap Image
image/*=Graphic Image 

i.e., we provide friendly names for the types we know about; and a generic name for an image of type image/blah-blah that's not specifically listed. 
We can do the same with MS documents and other types of files. Because we'll always be encountering file types we don't know about.

mheppler:
Currently, the File Type values that are delivered from MimeTypeFacets.properties are lower case (see attached). I suggest that we capitalize them.
![screen shot 2015-08-18 at 12 01 31 pm](https://cloud.githubusercontent.com/assets/687227/9335262/05166d2c-45a1-11e5-97cb-21d8733c6c21.png)

pdurbin:
@scolapasta I'm passing this to you for a decision of what to do for 4.2.

mheppler:
Related to #3333

landreev:
This was briefly discussed at the tech hours. 
What needs to be done can be summarized as 3 sub-tasks:

1) Curation work to better handle the mime types already in the database. (provide better user-friendly type/facet names; see the example below). 
2) Investigate/find a file type identification solution better than the currently used Jhove. It was not actively maintained for a long time; but it appears that it came back to live, possibly under a different team, with new versions being released. So one lead is to just try the latest version. 
3) Provide an API for re-identifying the files already in a Dataverse. That would allow to try an assign better mime types to the files currently in production typed as "application/octet-stream" (which is a fancy way to say "type unknown"). 

An example of "curation" work, in 1) above:
[...]

PR #5853, opened 2019-05-16
 Dataset - Too Many "Unknown" Files, Friendly File MIME Type Display Names #2202 

pdurbin:
#2202


src/main/java/propertyFiles/MimeTypeFacets.properties
```diff
@@ -26,35 +60,118 @@ application/x-stata-6=Data
...
+# FITS
+image/fits=FITS
 application/fits=FITS
+# Shape
+application/dbf=Shape
+application/dbase=Shape
+application/prj=Shape
+application/sbx=Shape
+application/shp=Shape
+application/shp=Shape
```
scolapasta: This seems like a duplicate line.

mheppler: Corrected. Good catch.

PR #5858, opened 2019-05-17
no need for SQL create scripts with flyway #5854

### LINKED ISSUE #5854: Migration Script: add v4.14 create script and version to versions file to keep migration script current.
I'm not sure whether this is still needed but I'd neglected to add the v4.14 create script to the v4.14 release so that the db migration script doesn't know about v4.14. This issue is to confirm we still need this and to add it.

pdurbin:
fixes #5854


kcondon:
@pdurbin Thanks!

PR #5908, opened 2019-06-03
set INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL to false #5907

### LINKED ISSUE #5907: validation error on root dataverse, Internal Server Error after logging in when deploying to Payara 5
We're getting a strange error when browsing to the root dataverse when Dataverse is deployed to Payara 5 rather than Glassfish 4.1:

58833050-3c47f380-861e-11e9-8230-cd1f1c13f3a3

The theory by the non-JSF experts (myself included) over at https://github.com/IQSS/dataverse/issues/4172#issuecomment-498426325 is that it has something to do with this:
         
    javax.faces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL

Should we try changing it to "false" like this?

```
[pdurbin@tabby dataverse]$ git diff
diff --git a/src/main/webapp/WEB-INF/web.xml b/src/main/webapp/WEB-INF/web.xml
index ff617a9fa..fc9c32ae9 100644
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -42,7 +42,7 @@
         <param-name>
             javax.faces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL
         </param-name>
-        <param-value>true</param-value>
+        <param-value>false</param-value>
     </context-param>
     <context-param>
         <param-name>javax.faces.FACELETS_SKIP_COMMENTS</param-name>
[pdurbin@tabby dataverse]$ 
```

pdurbin:
closes #5907

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -42,7 +42,7 @@
         <param-name>
             javax.faces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL
         </param-name>
-        <param-value>true</param-value>
+        <param-value>false</param-value>
     </context-param>
     <context-param>
         <param-name>javax.faces.FACELETS_SKIP_COMMENTS</param-name>


scolapasta:
We definitely  INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL to true purposefully, though I don't recall the exact thing that needed it.(possibly the way we handle "required" validation?)

Regardless a different solution was found, so am closing this PR.

pdurbin:
> Regardless a different solution was found, so am closing this PR.

The different solution was pull request #5933 but it caused new problems (unable to create a dataverse) so we reverted it with pull request #5944.

Should we revisit this fix, perhaps after absorbing this very long blog post? http://balusc.omnifaces.org/2015/10/the-empty-string-madness.html

PR #5913, opened 2019-06-04
use "root" as default alias, handle page=0 #5907

pdurbin:
Related to #5907, an alternative to pull request #5908.

This pull request feels like progress in getting Dataverse to work properly on Payara 5 in the sense that there is no validation error when you visit the homepage and after logging in as dataverseAdmin, one does not see an Internal Server Error. However, the "Create Dataverse" button doesn't work. The page flashes and you land on http://ec2-34-207-192-222.compute-1.amazonaws.com:8080/dataverse.xhtml?ownerId=1 (for example) but you are not redirected to the "create dataverse" page.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
@@ -307,6 +307,10 @@ public class SearchIncludeFragment implements java.io.Serializable {
         String allTypesFilterQuery = SearchFields.TYPE + ":(dataverses OR datasets OR files)";
         filterQueriesFinalAllTypes.add(allTypesFilterQuery);
 
+        if (page == 0) {
+            logger.info("Variable 'page' was zero. Changing it one. What does this have to do with upgrading from Glassfish 4.1 to Payara 5?");
+            page = 1;
+        }
         int paginationStart = (page - 1) * paginationGuiRows;
         /**
          * @todo
--- a/pom.xml
+++ b/pom.xml
@@ -292,7 +292,8 @@
         <dependency>
             <groupId>org.omnifaces</groupId>
             <artifactId>omnifaces</artifactId>
-            <version>1.7</version> <!-- Or 1.8-SNAPSHOT -->
+            <!-- "For users who are still on JSF 2.2, use 2.7.1 instead." http://omnifaces.org -->
+            <version>2.7.1</version>
         </dependency>
         <dependency>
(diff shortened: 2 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
```diff
@@ -307,6 +307,10 @@ public void search(boolean onlyDataRelatedToMe) {
         String allTypesFilterQuery = SearchFields.TYPE + ":(dataverses OR datasets OR files)";
         filterQueriesFinalAllTypes.add(allTypesFilterQuery);
 
+        if (page == 0) {
```
scolapasta: @pdurbin the log message makes me think we don't know why this is here? Can we find out if this is needed and why? I'd rather be sure of why we're adding in logic than just logging a  ¯\_(ツ)_/¯

pdurbin: @scolapasta yes, the message could be better. It is needed (or something like it) to avoid a 500 error like this:

58897303-ffcfd280-86c5-11e9-8e6e-4c5cd174ccb6

poikilotherm: Hey guys, I just created bd25409, which is likely to eliminate the root cause.

@pdurbin I had no chance to test this yet, would you be so kind?

pdurbin: @poikilotherm I just tested your fix in bd25409 but by cherry-picking it into e4946bc but I'm still getting the same 500 error and `java.lang.IllegalArgumentException: paginationStart must be 0 or greater at edu.harvard.iq.dataverse.search.SearchServiceBean.search(SearchServiceBean.java:148)
` in the logs. Here are some screenshots:

Screen Shot 2019-06-11 at 2 13 14 PM

Screen Shot 2019-06-11 at 2 13 10 PM

pdurbin:
@scolapasta 00fb3d0 has two fixes in one commit which was probably misguided of me. I just created pull request #5933 as a smaller chunk.

pdurbin:
> @scolapasta [00fb3d0](https://github.com/IQSS/dataverse/commit/00fb3d02e14e940f43b24dde6ce15f34354b8b95) has two fixes in one commit which was probably misguided of me. I just created pull request #5933 as a smaller chunk.

I'm closing this pull request in favor of pull request #5933.

Issue #5846, opened 2019-05-15
UZH - UI Testing / Additional Test Improvements


rschlaefli:
Hi all!

We have received our final assignment regarding testing improvements in the context of #5619. This will be the "root" issue for our tasks, which are the following:

- Implement UI tests for some parts of the application (e.g., using Cypress or Selenium)
- Improve some of the code we have already committed/merged
- And finally, add some more new testing/coverage

We were wondering if you have any previous experience with UI testing in the dataverse project or if there is any interest in starting to do something like it? We do not have to get anything merged for that task but could obviously create PRs if it were of interest to you. What it would/will probably look like is that the UI interactions are executed against mocked API responses to ensure their correct behavior (not "end-to-end").

In the same context we are also wondering if it is possible to only start the web interface without having to bootstrap the entire stack? Our experience with Java EE etc. is limited to this project.. We managed to run the entire application with the `docker-aio` but if it were possible to run the web interface separately that would ease the process a fair bit.

A final question: do you still want us to create one issue per PR, now that you have migrated to Github projects? It seems that you can add pull requests to the board as well and that issues are no longer grouped in any way. It might now be better to relate all PRs to this issue?

pdurbin:
> We were wondering if you have any previous experience with UI testing in the dataverse project or if there is any interest in starting to do something like it?

Yes, we are extremely interested in automated UI testing. If you look at 0d8150c you can see that we used to have a folder called "sauce_test" because we were playing around with Selenium and free test execution (free for open source) through "Open Sauce": https://saucelabs.com/blog/Announcing-Open-Sauce-free-unlimited-testing-for-Open-Source-projects

That was years ago. The automated UI tests have fallen into disrepair.

More to come but I wanted to show you the old broken tests that are still in the code base with that commit above. Thanks!!

pdurbin:
@rschlaefli I wanted to mention that I left a couple comments in #4202 and #4207 and than on Tuesday's community call, I asked @4tikhonov if now is a good time for him to coordinate with you. Unfortunately it isn't. He's busy working with @poikilotherm on getting the infrastructure in place to host Dataverse on Docker (that's my understanding anyway). It's not very well captured in the notes at https://groups.google.com/d/msg/dataverse-community/hekvbHfD-3w/uaTHr5rEAgAJ but I thanked you and @alexscheitlin and @nikzaugg for all your great work!

[...]

PR #5922, opened 2019-06-06
5846 Initial Framework for UI Testing

rschlaefli:
This pull requests includes the results of my initial investigation into UI testing the dataverse project using Cypress (see https://github.com/IQSS/dataverse/issues/5846 for the overall discussion and https://www.cypress.io/ for a general overview of the tool).

ezgif com-optimize

Some important points to consider regarding the code added in this PR:

- Cypress is based on JavaScript/NodeJS and thus requires an installation of `node`. It runs its tests in a browser-instance based on chromium/chrome/electron, depending on the browsers that are installed. Cypress can be installed by running `npm install` within `tests/`, and run locally using `npm run cypress:dev` or in CI using `npm run cypress:ci`.
  - Cypress is not (yet) usable for cross-browser testing but I think they are working on it.
- Cypress is configured with the `cypress.json` file located in `tests/cypress.json`. This "default" configuration can be overridden with environment variables (e.g., in CI):
  - Define `CYPRESS_baseUrl=https://staging.app.com` to override the default setting. Could be useful to point Cypress to another instance of dataverse when running in CI.
  - See also https://docs.cypress.io/guides/guides/environment-variables.html#Setting
- To setup the integration with the Cypress dashboard (see https://www.cypress.io/dashboard/), update `projectId` in `cypress.json` or set it dynamically using environment variable as described above.
  - The dashboard (free for OSS) displays screen recordings and other meta information that have been created by Cypress when run in CI. The dashboard does not actually execute the tests! 
  - To receive recordings, the CI runner also depends on the `CYPRESS_RECORD_KEY` environment variable, the value of which is shown when creating the Cypress project.
  - For an example of what the dashboard could do, see https://dashboard.cypress.io/#/projects/ufkudc/runs, which displays some of the runs executed when I was developing.
- The dataverse application stack needs to be in pristine state prior to testing, as the UI frameworks expects that the environment is always initialized the same way. Naive example: there always needs to be "one notification to be read", which is then marked as read by the UI tests.
  - When running a Dataverse instance in CI (`docker-aio` or `dataverse-kubernetes`), this will implicitly be the case, as the instance is spun up separately in each job.
  - When executing against a `phoenix` instance, this needs to be ensured manually, or cleanup API calls need to be added to `beforeEach` in the respective UI tests.
  - See https://docs.cypress.io/guides/references/best-practices.html for best practices in general and regarding this point (the Cypress docs are quite good).
- When running UI tests as developed for this PR, one can run into some incompatibilities with the frame busting techniques applied in Dataverse. The following line in `dataverse_template.xhtml` would need to be disabled/removed when testing, or refactored in general: 
  - `if (window !== top) top.location = window.location;`

## Related Issues

- #5846 UZH - UI Testing / Additional Test Improvements

## Pull Request Checklist

### DIFF
--- a/tests/cypress.json
+++ b/tests/cypress.json
@@ -1,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:8084",
+  "projectId": "ufkudc"
+}
--- a/tests/cypress/plugins/index.js
+++ b/tests/cypress/plugins/index.js
@@ -1,0 +1,18 @@
+
+// ***********************************************************
+// This example plugins/index.js can be used to load plugins
+//
+// You can change the location of this file or turn off loading
+// the plugins file with the 'pluginsFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+module.exports = (on, config) => {
+  // `on` is used to hook into various events Cypress emits
+  // `config` is the resolved Cypress config
+}
--- a/tests/cypress/util.js
+++ b/tests/cypress/util.js
@@ -1,0 +1,31 @@
+/**
+ * Function that performs a programmatic login
+ * @param {Cypress} cy
+ */
+export function login(cy, cb) {
+  // need to visit the login page to get the CSRF token
+  cy.visit('/loginpage.xhtml')
+
+  // a session needs to have been setup
+  cy.getCookie('JSESSIONID').should('exist')
+
+  // extract the CSRF token from the view state
+  cy.get('#loginForm input[name="javax.faces.ViewState"]').then(viewState => {
+    cy.request({
+      method: 'POST',
+      url: 'http://localhost:8084/loginpage.xhtml',
+      form: true,
+      body: {
+        'javax.faces.partial.ajax': true,
+        'javax.faces.source': 'loginForm:login',
+        'javax.faces.partial.execute': '@all',
+        'javax.faces.partial.render': '@all',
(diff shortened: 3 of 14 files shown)


tests/cypress/plugins/index.js
```diff
@@ -0,0 +1,18 @@
...
+// the plugins file with the 'pluginsFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+module.exports = (on, config) => {
```
rschlaefli: Just as a note prior to review, these stub files are not doing anything, but are recreated on every run of Cypress if they are not there (as are the folders). I would suggest to merge them as such to prevent this behavior.

tests/cypress/util.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Function that performs a programmatic login
+ * @param {Cypress} cy
+ */
+export function login(cy, cb) {
```
rschlaefli: Perform a login without having to explicitly fill in the login form (as a util in all tests that don't test the login page). It is still necessary to visit the login page to get a session id as well as the CSRF/ViewState (as far as I found out).

tests/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:8084",
+  "projectId": "ufkudc"
```
rschlaefli: Will need to be replaced if dataverse were to use Cypress (and dashboard) at some point

PR #5933, opened 2019-06-11
prevent JSF validation error on Payara 5 #4172

pdurbin:
The issue for supporting Payara 5 at #4172 and the issue for not being locked to Glassfish 4.1 at #2628 are both related but this is a "small chunk" fix that I don't believe causes any problems on Glassfish 4.1. Thank you to @smillidge for the suggestion at https://github.com/IQSS/dataverse/issues/4172#issuecomment-498446352 

Here is the error that I am trying to prevent on Payara 5.192 when visiting the home page of Dataverse:

58837007-51c21b00-8628-11e9-9040-c25cb33dc276

I see #4172 as representing full support for Payara 5 in the sense of an updated Dataverse Installation Guide, etc. so this pull request is not intended as a fix for that issue. Rather, the idea is to do no harm to existing Dataverse installations on Glassfish 4.1 but help move the needle on Payara 5 support.

PR #5940, opened 2019-06-12
avoid "paginationStart must be 0 or greater"

pdurbin:
This has something to do with upgrading from Glassfish 4.1.

This is the error we're trying to avoid:

java.lang.IllegalArgumentException: paginationStart must be 0 or greater at edu.harvard.iq.dataverse.search.SearchServiceBean.search(SearchServiceBean.java:148)


pdurbin:
I made this pull request because pull request #5933 was merged yesterday and this is the next logical step for something that is obviously broken when we try to move to something newer than Glassfish 4.1, such as Payara 5. I mentioned "java.lang.IllegalArgumentException: paginationStart must be 0 or greater at edu.harvard.iq.dataverse.search.SearchServiceBean.search(SearchServiceBean.java:148)" and the user experience is that the following 500 error is what you see on a fresh installation of Dataverse after you have logged in as "dataverseAdmin" and land on the home page:

58897303-ffcfd280-86c5-11e9-8e6e-4c5cd174ccb6

PR #5987, opened 2019-07-02
4425 add on success method to commands

### LINKED ISSUE #4425: Command Engine: Move indexing and other actions into "onSuccess" method
As a developer, when calling a command such as CreateDatasetCommand, I want to be able to do things like index the dataset **after** the dataset as been successfully created and the EJB transaction has completed, so that I will never have a scenario where my dataset was indexed but it does not exist in the database. 

@scolapasta and I started a little bit of work related to this issue but determined it needed a bigger discussion surrounding its implementation. The current work on branch `4425-on-success` is not what we think the correct solution will be, but just initial work.

sekmiller:
## New Contributors

## Related Issues

- closes #4425 move indexing to onSuccess method to be run after command completes

## Pull Request Checklist

- [x] Unit [tests][x] completed
- [x] Integration [tests][x]: completed
- [x] Deployment requirements: [x] None
- [x] Merged latest from "develop" [branch][x] and resolved conflicts


michbarsinai:
All in all looking good, and seems like the code is handling errors that it didn't handle before. Some questions:

* EJBDataverseEngine:255 -  I'm not sure I was able to follow the full logic here... is this always true on the last non-inner command executed in a command sequence?
* EjbDataverseEngineInner - This class needs 1-2 lines of JavaDoc, takes time to understand why it's there.
* Index.java:228-234 - Similar lines appear in many places in the code. Maybe could be consolidated into a single method. OR - we might need an `onFailure` handler as well.
* Command:46 - the onSuccess methods needs some JavaDoc, Command is a class new developer refer to a lot.
* DeleteDataverseLinkingDataverseCommand:56 - new code does not re-index editedDV. Is that intentional?

This PR improves on the current design. I think it needs a bit more documentation since it's the core of the code, and people will be reading it.

src/main/java/edu/harvard/iq/dataverse/EjbDataverseEngine.java
```diff
@@ -250,6 +250,8 @@ private DvObject getRetType(Object r){
                     getContext().beginCommandSequence();
                 }
                 getContext().addCommand(aCommand);
+                //This list of commands is hel by the outermost commands context
```
mheppler: @sekmiller typo ("hel")?

sekmiller: "held". thanks

PR #6022, opened 2019-07-15
remove extra tab character for publicationIDNumber #5766

### LINKED ISSUE #5766: OAI-ORE exports are missing ID Number
The citation.tsv file modified for release 4.10 (and continues through 4.12) has one line with an extra tab. The line for the field "publicationIDNumber" (displayOrder=31) has an extra tab before the "purl" URL, so the termURI field for that line is really empty - the contents is in the next field over.

Not sure what this affects, but noticed the mis-alignment when I compared this citation.tsv file with the one we have for 4.9.4

pdurbin:
Closes #5766

Head up that I did not remove the extra "arxiv" mentioned in #5766. (Update: it was later removed in 4cbfef1.)

Also, I didn't test this at all. I just removed the tab character.

### DIFF
--- a/doc/release-notes/5766-citation-tsv
+++ b/doc/release-notes/5766-citation-tsv
@@ -1,4 +1,4 @@
 Update citation metadata block
 curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "C
 
-Run ReExportall to update the citation metadata block to remove an errant tab: http://guides.dataverse.org/en/4.10/admin/metadataexport.html?highlight=export#batch-exports-through-the-api
+Run ReExportall to update the citation metadata block to remove an errant tab: http://guides.dataverse.org/en/4.16/admin/metadataexport.html?highlight=export#batch-exports-through-the-api
--- a/doc/release-notes/5766-citation-tsv
+++ b/doc/release-notes/5766-citation-tsv
@@ -1,0 +1,4 @@
+Update citation metadata block
+curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "C
+
+Run ReExportall to update the citation metadata block to remove an errant tab: http://guides.dataverse.org/en/4.10/admin/metadataexport.html?highlight=export#batch-exports-through-the-api
--- a/doc/release-notes/5766-citation-tsv
+++ b/doc/release-notes/5766-citation-tsv
@@ -1,4 +1,4 @@
 Update citation metadata block
-curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "C
+curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "Content-type: text/tab-separated-values"
 
 Run ReExportall to update the citation metadata block to remove an errant tab: http://guides.dataverse.org/en/4.16/admin/metadataexport.html?highlight=export#batch-exports-through-the-api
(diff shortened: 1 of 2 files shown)


kcondon:
@pdurbin Does this require existing installations to reimport citation.tsv in production as part of upgrade instructions and also to reexport all to fix ORE exports? If so, would you add this to the release notes file?

pdurbin:
@kcondon reimporting the citation.tsv file won't remove the tab character I removed. I did not investigate writing a SQL script to remove the tab character.

I also did not test the original bug in #5766 which, as you say, has something to do with ORE exports. I was hoping that you or @jggautier could help with this. I'm happy to spin up a server running 4.15.1 (the released version) as well as a server running this branch so that the bug and fix can be confirmed.

Once we've established that the bug can be reproduced and the fix works I'm happy to write up some release notes explaining what to do. I'm not confident enough in my SQL skills to say that I'd be able to quickly write a script to automate this but perhaps another developer on the team could help. I'm trying to keep to the "small" estimate.

kcondon:
@pdurbin I guess what I'm saying as the designer of the fix there is a nominal responsibility to inform downstream folks about the implications of the fix.

pdurbin:
@kcondon that makes sense. I'm at capacity at the moment. Please feel free to move this pull request out of the QA column if you want someone else to confirm the ORE export bug and test the fix before you do.

kcondon:
@pdurbin What I really want is for the developer to understand what the fix does and what other actions need to take place to deploy it and then communicate that.

djbrooke:
Assigning to myself for now, happy to work collaboratively on this to determine the next steps once you're both free. Thanks in advance!

kcondon:
So we have confirmed the changed citation.tsv fixes the mapping for publicationId field. Also, for existing dataverses, release notes should indicate importing the updated citation.tsv and then reexporting to update existing datasets. These two steps are identical to those same steps in v4.10 upgrade notes.

djbrooke:
Thanks, I'll add the release notes file.

djbrooke:
@kcondon - I'm moving this back to QA, but note that the only change that I made since you last saw this was adding release notes. Let me know if I can be of any help.

doc/release-notes/5766-citation-tsv
```diff
@@ -0,0 +1,4 @@
+Update citation metadata block
+curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "C
```
pdurbin: @djbrooke this curl command seems incomplete. It ends with ` -H "C`

pdurbin: Also, does it work? Is removing a tab character one of the supported operations for reloading a metadata block? I thought the main support operation was adding (but not removing) additional controlled vocabulary. @scolapasta and @sekmiller know better than I do what's supported.

djbrooke: Thanks, updated in https://github.com/IQSS/dataverse/pull/6022/commits/253f12d93e90a4c23d7017609e99d204af6296a0

djbrooke: I talked with @kcondon about this a bit yesterday and I know he worked with @scolapasta as noted above, so his knowledge has been involved. Will resolve this convo. If the fix for this small changes does not work as expected, we'll revisit.

pdurbin: @djbrooke oh! I missed the note at https://github.com/IQSS/dataverse/pull/6022#issuecomment-512999475 ! That's fantastic that this is a supported operation!

Issue #5303, opened 2018-11-09
Internationalization - Terms of Use


JayanthyChengan:
Requirement for internaionalization:
terms of use - 

![image](https://user-images.githubusercontent.com/12862735/48286367-6728eb00-e433-11e8-9feb-534000c83e6a.png)

pdurbin:
@JayanthyChengan thanks for pull request #5304 

I took a quick look and the first thing that struck me is that you are changing how "General Terms of Use" is configured in the application. We've always configured it with the ":ApplicationTermsOfUse" database setting like this:

![screen shot 2018-11-13 at 12 04 00 pm](https://user-images.githubusercontent.com/21006/48429975-3c44dc80-e73c-11e8-99bf-0f2f89090dd1.png)

Thanks from http://guides.dataverse.org/en/4.9.4/installation/config.html#applicationtermsofuse and I believe #2669 was recently fixed by pull request #5269, by the way.

Please edit doc/sphinx-guides/source/installation/config.rst in your pull request to remove the old way and add the new way to set General Terms of Use.

Also, while you're on that branch, please take a look at the preferred placement of curly braces at http://guides.dataverse.org/en/4.9.4/developers/coding-style.html#braces-placement . In Java it's valid to leave out the braces in the following code...

        if(appTermsOfUse != null)
            return appTermsOfUse;
        else
            return saneDefaultForAppTermsOfUse;

... but we prefer to keep them in like this:

        if (appTermsOfUse != null) {
            return appTermsOfUse;
        } else {
            return saneDefaultForAppTermsOfUse;
        }

If you reformat the lines you changed as advised at http://guides.dataverse.org/en/4.9.4/developers/coding-style.html#format-code-you-changed-with-netbeans Netbeans will add the curly braces in for you automatically.

JayanthyChengan:
@pdurbin - updated the docs.  I noticed that in this approach, by default, any new dataverse installation will display the "Dataverse General Terms of Use" from termsofuse.properties instead of "There are no Terms of Use for this Dataverse installation".  Is that ok? , if not , how can we design?

pdurbin:
[...]

PR #6042, opened 2019-07-19
Internationalization - Application Terms of Use

JayanthyChengan:
## Related Issues

- https://github.com/IQSS/dataverse/issues/5303


scolapasta:
@JayanthyChengan overall this looks good.

I think we should not have EN for lang for all settings, especially since not all settings are user text, so blank lang, which is default, but then you can add language specific ones if you need. So the logic should be if you don't provide a lang, it returns the default (and most settings would do that); if you do provide a language it looks for that language, but if not available returns the default.

JayanthyChengan:
@scolapasta  - Thanks for your review.  I would like to add some notes of what I tried.
As a first step, I added new column “lang” and when a new row is added(:ApplicationTermsOfUse , "french content" , "fr" ) , it throws the below exception,

INSERT INTO setting (name, content, lang)
VALUES
(:ApplicationTermsOfUse , "french content" , "fr" );
 
**org.postgresql.util.PSQLException: ERROR: **duplicate key value** violates unique constraint "setting_pkey"**

 
So, I added “lang” to the primary key (name,lang), and  updated to  default value “en” as primary key cannot be NULL
 
https://github.com/IQSS/dataverse/blob/b9bed069c01a5587cc1a75c551ff4352dd9fc755/scripts/database/upgrades/upgrade_v4.15.1_to.sql
 
setting_pkey PRIMARY KEY (name,lang)
 
Please add your suggestion.

JayanthyChengan:
@scolapasta - Thanks Gustavo and I implemented the suggested solution using coalesce, it worked.  Committed the code,  please review.  Thanks

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -2383,6 +2383,7 @@ staticSearchFields.dvObjectType=Type
 staticSearchFields.fileTag=File Tag
 staticSearchFields.fileAccess=Access
 staticSearchFields.publicationStatus=Publication Status
+staticSearchFields.subject_ss=Subject
```
scolapasta: @JayanthyChengan is this supposed to be part of this PR?

scripts/database/upgrades/upgrade_v4.15.1_to.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE ONLY setting DROP CONSTRAINT setting_pkey ;
+
+ALTER TABLE setting ADD COLUMN ID SERIAL PRIMARY KEY;
```
scolapasta: does this add values for the id for existing settings? (it must, right?)

Also, can you update the file name to reference 4.16 (the latest release?)

JayanthyChengan: Yes, it adds serial number to all existing rows.  I will rename the file and commit it.

src/main/java/propertyFiles/Bundle.properties
JayanthyChengan: @scolapasta  - Missing term related to facet Internationalization.  hope it's okay to add with this PR.

scripts/database/upgrades/upgrade_v4.15.1_to.sql
JayanthyChengan: committed the changes.

kcondon:
@JayanthyChengan Would you refresh from develop? The app version is still 4.15.1.

Thanks!

JayanthyChengan:
@kcondon - merged the latest develop branch.

kcondon:
@JayanthyChengan thanks. I am seeing a 500 error when I load the app in a browser with a sql error in logs:
Call: SELECT ID, CONTENT, LANG, NAME FROM SETTING WHERE ((NAME = ?) AND (LANG IS NULL))
        bind => [1 parameter bound]
Query: ReadAllQuery(name="Setting.findByName" referenceClass=Setting sql="SELECT ID, CONTENT, LANG, NAME FROM SETTING WHERE ((NAME = ?) AND (LANG IS NULL))")

JayanthyChengan:
Hi @kcondon - there is a setting table update, I am not sure whether you executed the db update script?  Let me know if the error still exists.  Thanks
 https://github.com/IQSS/dataverse/blob/f199fce7b3146877d9576bb474fcd3f3186cdfd7/scripts/database/upgrades/upgrade_v4.16_to_4.16.1.sql

kcondon:
@JayanthyChengan We now use an automatic db update management s/w called Flyway. We no longer use the updates dir method. If you would use the Flyway method, that would be great: http://guides.dataverse.org/en/4.16/developers/sql-upgrade-scripts.html# Thanks!

JayanthyChengan:
@kcondon  - Added the db upgrade file as per Flyway method.  Please verify it now.  Thanks

kcondon:
@JayanthyChengan Almost there! The new script needs to be renamed: you need to use a capital V rather a v at the beginning of the script name, otherwise flyway ignores the script. Annoying, I know.

JayanthyChengan:
@kcondon - got it, and renamed the file .

Issue #4201, opened 2017-10-13
Test Automation: Investigate, specify, prototype a performance test suite that can be run at build time.


kcondon:
Produce a performance test suite that can run at build time.
Initially, it should measure the performance of the most frequently accessed pages: dataset, dataverse.

There are some open questions about what constitutes an effective performance test suite: whether it needs to be deterministic of production performance, requiring the same servers and data or whether it can be used as a relative measure of performance using a range of real and/or synthetic data, such as larger lists of parent/child objects, eg. dataverse with datasets, datasets with files.

pdurbin:
I'd recommend watching the video created https://github.com/BU-NU-CLOUD-SP18/Dataverse-Scaling#our-project-video that was created by BU students Ashwin Pillai, James Michael Clifford, Ryan Morano, and Patrick Dillon in April 2018.

They're using JMeter and here are some screenshots from around 17:38:

![screen shot 2018-04-30 at 1 42 42 pm](https://user-images.githubusercontent.com/21006/39441952-9c7a4a18-4c7d-11e8-806f-7137605366e0.png)
![screen shot 2018-04-30 at 1 44 33 pm](https://user-images.githubusercontent.com/21006/39441953-9c86434a-4c7d-11e8-8df6-171dbbba3956.png)
![screen shot 2018-04-30 at 1 44 48 pm](https://user-images.githubusercontent.com/21006/39441954-9c9225d4-4c7d-11e8-8840-c3ba1c02d8cf.png)
![screen shot 2018-04-30 at 1 45 04 pm](https://user-images.githubusercontent.com/21006/39441955-9c9d616a-4c7d-11e8-9008-a4828433fbdb.png)
![screen shot 2018-04-30 at 1 45 25 pm](https://user-images.githubusercontent.com/21006/39441957-9cb90e9c-4c7d-11e8-8722-bf47278d8ece.png)

rockash:
The above tests have been run on a Minishift deployment of the containerized Dataverse application used for our class group project [Dataverse Scaling](https://github.com/BU-NU-CLOUD-SP18/Dataverse-Scaling.). I have documented steps for creation of my simple Jmeter test to test the Dataverse deployment. I have also added two Jmeter ".jmx" files. 

"**Dataverse Load testing template.jmx**" is a file which can be used with the steps for test creation on Jmeter to make a customized test plan as needed. 

[...]

Issue #6035, opened 2019-07-18
Memory leak when dataverse is stressed with GET requests


PaulBoon:
Our Dataverse was coming to a grinding halt with a CPU load of 400%

The high CPU load seems to be a result of a high memory usage. 
Probably the GC kicks in when the memory is getting full. 
When glassfish or the whole server is restarted memory usage is back to low.

We have looked in the logs and finally discovered that the site got scraped and someone was firing a GET request every second visiting verse pages with type=file and fq0, fq1 and fq2 parameters. 
 
The scenario to produce leakage has been reduced to a simple GET request on a verse (non- homepage also) every second. If this is done for 10 minutes we get a 200M leak. 

WE are running our fork on 4.10.1 on production, but we could reproduce this with the 4.14 war file from IQSS

PaulBoon:
Opened a GitHub issue at IQSS: https://github.com/IQSS/dataverse/issues/6035

djbrooke:
Hey @PaulBoon! I've queued this for the IQSS team to work on in our next sprint, beginning on Wednesday. Two quick questions:

- Do you plan to do any work in this area and produce a pull request? If so, we'll either wait or coordinate with you. Let me know.
- Is there any additional information that you can provide here that would be helpful? 

Thank you again for reporting and troubleshooting this with @kcondon.

PaulBoon:
Hi @djbrooke, Great to hear that you have planned it in your next sprint. 

I am still working on my own sprint, with this memory leakage as an incident. My plans are to keep a look at this issue, while working on other stuff before going on holidays. So I don't think I can be of help much right now.

pdurbin:
@PaulBoon thanks for opening this issue. Do you have a JMeter script I can add to pull request #6043 that reproduces the memory leak?

PaulBoon:
@pdurbin The 'script' that I used is very simple, but I can put it here so you can have a look. I used apache-jmeter-5.1.1 to produce it. 
[Dataverse Memory Leak.jmx.zip](https://github.com/IQSS/dataverse/files/3425781/Dataverse.Memory.Leak.jmx.zip)

PaulBoon:
Maybe it would be interesting to see if the known Solr leak is involved, by applying the fix and test. But I don't see how it could explain the leaking I saw if I just visited the loginpage every second, or other pages. 

To me this suggests that it is something very basic, like primefaces related. 
[...]

PR #6043, opened 2019-07-22
stub out jmeter #4201

pdurbin:
Related to #4201

Running `mvn verify` will create a CSV like this:

Screen Shot 2019-07-22 at 11 14 32 AM

Update: I added a few docs:

Screen Shot 2019-07-22 at 3 01 13 PM

I've been looking at these two pages:

- https://github.com/jmeter-maven-plugin/jmeter-maven-plugin
- https://www.blazemeter.com/blog/how-use-jmeter-maven-plugin/

### DIFF
--- a/doc/sphinx-guides/source/developers/testing.rst
+++ b/doc/sphinx-guides/source/developers/testing.rst
@@ -215,6 +215,19 @@ Locust
 
 Load and performance testing is conducted on an as-needed basis but we're open to automating it. As of this writing Locust ( https://locust.io ) scripts at https://github.com/IQSS/dataverse-helper-scripts/tree/master/src/stress_tests have been used.
 
+JMeter
+~~~~~~
+
+We are starting to experiment with JMeter: https://jmeter.apache.org
+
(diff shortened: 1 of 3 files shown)


pdurbin:
@landreev I'm glad you like JMeter too but please not that I did change the pom.xml so I wouldn't say it's documentation only. I don't believe we are using `mvn verify` currently so I don't think any behavior will change for us. It's basically a no-op given our current use of Maven and I think it's safe to merge so we can continue to play with JMeter.

djbrooke:
@scolapasta can you work with @pdurbin and @landreev to discuss implementation of jmeter or something else that allows better detection of leaks?

scolapasta:
Per discussion yesterday with @landreev and @pdurbin we're going to close this PR. 

The work in the branch is a good start for exploring JMeter and automated stress testing, bur we'll want a more robust test written for the initial PR than the one provided here. It could possibly be the one provided by @PaulBoon in #6035 or a variation of that.

Issue #5473, opened 2019-01-17
Dataverse Facets - Duplicate French Values Listed (Category, File Type)


sbarbosadataverse:
In my demo yesterday I noticed one category facet in "french" for one dataverse. When I navigate to that dataverse, that category is not an option in French, as expected and not included anywhere else in the metadata for that dataverse and its dataset. See screen shots. 

<img width="533" alt="screen shot 2019-01-17 at 1 44 59 pm" src="https://user-images.githubusercontent.com/8322346/51341793-287d6580-1a60-11e9-81be-320e72aebffb.png">
<img width="1334" alt="screen shot 2019-01-17 at 1 45 09 pm" src="https://user-images.githubusercontent.com/8322346/51341794-287d6580-1a60-11e9-8aec-ff063d333fc4.png">

mheppler:
There are now three facets in production in French.

```
Chercheur (1)
Groupe de recherche (1)
Projet de recherche (1)
```

With three different dataverses.

https://dataverse.harvard.edu/dataverse/amsre_footprint

scolapasta:
I cleaned up the ones in production by reindexing via admin API. They seemed to have been created in later January / early February. Is it possible we had multiple languages enabled briefly? 

In other words, I didn't do anything about what is likely the underlying cause. but fixing individual one is easy.

mheppler:
Was able to replicate the duplicate translated category values in my localhost.

![Screen Shot 2019-04-02 at 1 18 19 PM](https://user-images.githubusercontent.com/687227/55422632-10d75880-554a-11e9-9aa0-959c99051e17.png)

1. Toggle language to French
2. Create dataverse
3. Publish dataverse
4. Toggle language to English, find French category value listed as duplicate facet

mheppler:
Copied comment from related/duplicate issue that I open and closed.

_Search - French Controlled Vocabulary Facets #5752_

> Found more French facets in [production](https://dataverse.harvard.edu/dataverse/harvard?q=&fq0=fileTypeGroupFacet%3A%22Texte%22&types=files&sort=dateSort&order=desc).
> 
> <img alt="Screen Shot 2019-04-10 at 3 33 42 PM" width="1184" src="https://user-images.githubusercontent.com/687227/55907948-267dfb00-5ba6-11e9-9dcc-3488e5e563f2.png">

JayanthyChengan:
@scolapasta - please review PR https://github.com/IQSS/dataverse/pull/6051

PR #6051, opened 2019-07-24
5473 duplicate french values

JayanthyChengan:
## Related Issues

- https://github.com/IQSS/dataverse/issues/5473

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Dataverse.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataverse.java
@@ -111,23 +111,23 @@ public class Dataverse extends DvObjectContainer {
 
         public String getFriendlyCategoryName(){
        switch (this.dataverseType) {
-            case RESEARCHERS:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchers");
-            case RESEARCH_PROJECTS:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchProjects");
-            case JOURNALS:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.journals");           
-            case ORGANIZATIONS_INSTITUTIONS:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.organizationsAndInsitutions");           
-            case TEACHING_COURSES:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.teachingCourses");
-            case LABORATORY:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.laboratory");
-            case RESEARCH_GROUP:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchGroup");
-            case DEPARTMENT:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.department");
-            case UNCATEGORIZED:
+           case RESEARCHERS:
+               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchers", null , "en");
+           case RESEARCH_PROJECTS:
+               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchProjects",null, "en");
+           case JOURNALS:
+               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.journals", null, "en");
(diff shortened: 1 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/Dataverse.java
```diff
@@ -111,23 +111,23 @@ public void setDataverseType(DataverseType dataverseType) {
 
         public String getFriendlyCategoryName(){
        switch (this.dataverseType) {
-            case RESEARCHERS:
```
scolapasta: We need to be careful that this (and other changes in this PR) are used by places in the app that do want the localized version. Maybe we should change the getIndexableCategoryName instead? (ideally, both it and this would call one method, and pass parameters to either get localized or default)

src/main/java/edu/harvard/iq/dataverse/Dataverse.java
```diff
@@ -111,23 +111,23 @@ public void setDataverseType(DataverseType dataverseType) {
...
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.teachingCourses");
-            case LABORATORY:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.laboratory");
-            case RESEARCH_GROUP:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchGroup");
-            case DEPARTMENT:
-                return BundleUtil.getStringFromBundle("dataverse.type.selectTab.department");
-            case UNCATEGORIZED:
+           case RESEARCHERS:
+               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchers", null , "en");
```
scolapasta: can we have this get the value from the default bundle and not hard coded "en"?

JayanthyChengan: @scolapasta  - committed new code. please verify

src/main/java/edu/harvard/iq/dataverse/Dataverse.java
JayanthyChengan: @scolapasta  - refactored mehtods in Dataverse.java and FileUtil.java.  Please add your comments

src/main/java/edu/harvard/iq/dataverse/Dataverse.java
```diff
@@ -109,37 +109,48 @@ public void setDataverseType(DataverseType dataverseType) {
...
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.laboratory",null, "en");
-           case RESEARCH_GROUP:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.researchGroup",null, "en");
-           case DEPARTMENT:
-               return BundleUtil.getStringFromBundle("dataverse.type.selectTab.department",null, "en");
-           case UNCATEGORIZED:
+    public String getFriendlyCategoryName(){
+        String key = getFriendlyCategoryKey();
+
+        if (key.equals(uncategorizedString)) {
```
scolapasta: @JayanthyChengan why would this not use a value from the bundle? I *think* in the old code, it may have been because this was used by the getIndexable... function and that wanted to make sure it could have a set string and know to ignore it? (this is a guess). But now the getIndexable method uses the key method, the friendly method could treat it the same (and replace with localized text), while the getindexable can do what it needs to and ignore it (based on ley value, instead if text value. Does that make sense?

(I've asked the team if they know of anywhere else where we would hide "Uncategorized" besides facets)

JayanthyChengan: @scolapasta  - replaced "uncategorizedString" string with bundle key "dataverse.type.selectTab.uncategorized"

kcondon:
@JayanthyChengan @scolapasta 
OK, these changes now allows for the facet to be translated, regardless of which language was used to create it. However, it still writes the creating language to the index so that there are still two distinctly separate facets for department in french and english, even if now they are translated to be in the same language. Also, those server log errors have returned about not finding keys in the bundle, even though I have the updated bundles in place.

kcondon:
@JayanthyChengan Worked with @scolapasta this am to determine this branch was doing what was expected, so merging. Thanks!

Issue #5028, opened 2018-09-06
Dataset Explore


djbrooke:
Dataverse has the infrastructure to support community built, file-level external tools (TwoRavens, Data Explorer). Similar infrastructure should be added at the dataset level to support the Code Ocean integration funded by the recent Sloan grant. Additionally, supporting Binderverse (#4714), SBGrid's reprocessing tool (ping @pameyer), and other future tools should be considered.

pdurbin:
This issue represents the initial step toward integration of Dataverse with Code Ocean (Sloan grant) and hopefully related tools such as Binder and Whole Tale (community efforts) so I'm going to leave a bit of a brain dump of recent happenings.

On Tuesday during our regular community call, the Code Ocean team shared their screen and we talked through the future integration at a pretty high level. Notes can be found at https://groups.google.com/d/msg/dataverse-community/HPLziKZbOAc/q_XEqyKEBwAJ

Yesterday @djbrooke and I called in to the first meeting of the Open Science Infrastructure Working Group organized by @craig-willis from @whole-tale to discuss a variety of computation and reproducibility topics with @aprilcs (Code Ocean) @donsizemore and @tlchristian (Odum) @choldgraf and @aculich (Binder #4714) @craig-willis , @Xarthisius and @amoeba (Whole Tale #5097) and others. Notes at https://docs.google.com/document/d/1bOVWBfhOiKGU2dYoHN_Pkpv5zPI6y819UISfkTqYpoQ/edit?usp=sharing

This morning I attended a fantastic Code Ocean workshop given by @aprilcs and I'd be happy to walk anyone through her tutorial which is captured very nicely in her slides ( http://bit.ly/harvard-oa-week ), example capsule ( https://bit.ly/2NanJLc ), and git repo ( https://github.com/aprilcs/candy_trade ). I'd also like to note that she features https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/EZSJ1S as an example of a dataset that has an excellent README describing how to reproduce results using code and data in the dataset.

@mheppler and discussed this issue this morning and to me the next logical step is to make a decision on where on the dataset page the button should go that brings the user from the dataset to Code Ocean, Whole Tale, Binder, etc., sort of like how the "Explore" button can bring you to multiple [external tools](http://guides.dataverse.org/en/4.9.4/installation/external-tools.html) at the file level such as Data Explorer and Two Ravens:

[...]

Issue #4429, opened 2018-01-22
Revise Explore Tools documentation to reflect modularity


dlmurphy:
With #3657 we've moved toward treating external explore tools as modular. This new way of looking at external tools in Dataverse requires we rethink the way we talk about them in our documentation. Let's review all mentions of TwoRavens, GeoConnect/Worldmap, and other explore tools throughout all guides to make sure that it's clear that these tools are considered modular, and that any number of other modular explore tools may be present in any given Dataverse installation.

dlmurphy:
Note that Geoconnect is not considered modular in the same way as TwoRavens. At this point, Geoconnect is still a core feature of Dataverse that we maintain. Docs should reflect this.

pdurbin:
I just reviewed the entire API Guide as part of pull request #6107 and there's nothing to fix there for this issue so I removed the GitHub Issue label for that guide.

mheppler:
Going to close this issue, as the review of the guides suggested will be completed as part of issue #6919 for preview tools, and anything not resolved there, can be added to issue #5597 for Geoconnect/WorldMap.

PR #6059, opened 2019-07-26
5028 Add dataset level external tools

pdurbin:
For #5028


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2016,6 +2021,7 @@ private String init(boolean initFull) {
         
         configureTools = externalToolService.findByType(ExternalTool.Type.CONFIGURE);
         exploreTools = externalToolService.findByType(ExternalTool.Type.EXPLORE);
```
qqmyers: findByType isn't filtering by scope yet - should these use findByTypeAndScope with file scope?

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
```diff
@@ -131,6 +147,7 @@ public static ExternalTool parseAddExternalToolManifest(String manifest) {
         String displayName = getRequiredTopLevelField(jsonObject, DISPLAY_NAME);
         String description = getRequiredTopLevelField(jsonObject, DESCRIPTION);
         String typeUserInput = getRequiredTopLevelField(jsonObject, TYPE);
+        String scopeUserInput = getRequiredTopLevelField(jsonObject, SCOPE);
         String contentType = getOptionalTopLevelField(jsonObject, CONTENT_TYPE);
         //Legacy support - assume tool manifests without any mimetype are for tabular data
         if(contentType==null) {
```
qqmyers: This is going to set a content type on dataset level tools

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
```diff
@@ -139,26 +156,29 @@ public static ExternalTool parseAddExternalToolManifest(String manifest) {
...
         JsonArray queryParams = toolParametersObj.getJsonArray("queryParameters");
         boolean allRequiredReservedWordsFound = false;
-        for (JsonObject queryParam : queryParams.getValuesAs(JsonObject.class)) {
-            Set<String> keyValuePair = queryParam.keySet();
-            for (String key : keyValuePair) {
-                String value = queryParam.getString(key);
-                ReservedWord reservedWord = ReservedWord.fromString(value);
-                if (reservedWord.equals(ReservedWord.FILE_ID)) {
-                    allRequiredReservedWordsFound = true;
+        if (scope.equals(Scope.FILE)) {
```
qqmyers: else Scope.DATASET - just not written yet?

doc/sphinx-guides/source/installation/external-tools.rst
```diff
@@ -35,14 +35,17 @@ External tools must be expressed in an external tool manifest file, a specific J
...
 In the example above, a mix of required and optional reserved words appear that can be used to insert dynamic values into tools. The supported values are:
 
-- ``{fileId}`` (required) - The Dataverse database ID of a file the external tool has been launched on.
-- ``{siteUrl}`` (optional) - The URL of the Dataverse installation that hosts the file with the fileId above.
+- ``{fileId}`` (required for file tools) - The Dataverse database ID of a file from which the external tool has been launched.
+- ``{siteUrl}`` (optional) - The URL of the Dataverse installation from which the tool was launched.
 - ``{apiToken}`` (optional) - The Dataverse API token of the user launching the external tool, if available.
-- ``{datasetId}`` (optional) - The ID of the dataset containing the file.
+- ``{datasetId}`` (optional) - The ID of the dataset.
+- ``{datasetPid}`` (optional) - The Persistent ID (DOI or Handle) of the dataset.
```
qqmyers: Why this and not file PID too? Since Dataverse is always generating this, I'm not sure it helps anyone to have both id and pid (a tool can get the pid), but it seems like file and dataset should be the same.
Also - right now we check for fileId existing (for file scope) as a required condition. For dataset, and file if it is made parallel, the condition will have to change to be that id or pid is required.

src/main/resources/db/migration/V4.15.1.1__5028-dataset-explore.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE externaltool ADD COLUMN IF NOT EXISTS scope VARCHAR(255);
+UPDATE externaltool SET scope = 'FILE';
+ALTER TABLE externaltool ALTER COLUMN scope SET NOT NULL;
```
qqmyers: should content type now be nullable (in the db and class definition) since it isn't used for scope dataset?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
pdurbin: Yes, they probably should.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
pdurbin: Yes, this is a bug. We don't want dataset tools to have a content type. Good catch. Thanks.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
pdurbin: Well, I'm not sure. Should dataset tools have any reserved words that are required?

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
scolapasta: @qqmyers eventually we may use this - for example, show the dataset level explore only if the dataset has at least one file of content type X - but for now we're going with simple.

doc/sphinx-guides/source/installation/external-tools.rst
pdurbin: Good point. The focus right now is on dataset level tools but sure, if a file PID exists it would hurt to expose it. Tool makers should just understand that they might not always get a file PID.

I'm not sure if I grok your other comment but thanks for making it. 😄 Someone else will be picking this up while I'm on vacation, I suspect.

src/main/resources/db/migration/V4.15.1.1__5028-dataset-explore.sql
pdurbin: As mentioned above, due to a bug in the pull request, dataset tools get the default content type. Yes, I think this field should be made nullable (and the bug fixed).

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
qqmyers: without an else statement, allRequiredReservedWordsFound is false and an exception happens so this method doesn't work for scope dataset as is. 
w.r.t. reserved words - it would be odd/bad to configure a dataset tool to not get the dataset id (or pid) in the same way that a file tool needs a file id (or pid if that's implemented).

qqmyers: ah sorry - I see the exception is in the if statement - probably should not define the allRequiredWordsFound outside the if clause then (to avoid confusing people like me :-) )

djbrooke:
Thanks @sekmiller for picking this up. Question for you, @scolapasta and @pdurbin:

Was there any discussion about how/if Dataset explorations should be recorded in the guestbook table? At the file level, I believe the explore tool name is added. Happy to discuss and sorry if I missed some communication or the spot in the code where this was reflected.

scolapasta:
It's a good question - it has not been discussed. It's a challenge because guestbook responses are at the file level. 

The other thing that we need to think about are restrictions? if one or more files are restricted, do we show this button?

djbrooke:
Thanks @scolapasta. When the Dataset explore is initiated, does the tool download all the files or does it download them as needed (or will this differ from tool to tool)? If it's the former, I'd be OK with guestbook records being created for all files in the dataset when a dataset explore is initiated and reflecting the dataset explore tool's name in the table.  If the latter, we can create the records as the tool pulls them. Better ideas or corrections welcome.

I'm not super worried about the some-restricted-files use case right now as the tools being focused on at this time are geared towards replication of datasets from journals which are mostly (all?) public. We will definitely need to worry about it in the future for these replication tools, both for restricted files and for files in TRSAs, and also as more tools are added with additional use cases that are not replication focused. 

Note that we do need to provide the explore option pre-publish, so we'll already be dealing with files that can only be accessed by certain users, not sure if there's something that can be leveraged there.

qqmyers:
If the dataset level tools work as the file level ones, they'll retrieve files as needed and either be restricted to public files, or, if configured to be sent an api key, the ones the user launching them can see. So I think security is always maintained, but the question of whether the button should be shown is separate and depends somewhat on whether/how well Dataverse can determine if the tool can retrieve all files it will need and therefore be able to run. A binary yes if the tool can get all files /no if it can't wouldn't be hard to calculate, but if tools only depend on some files, one would need a way to decide if it can get the 'right' files...

scolapasta:
I think it will depend on the tool, but we should err on the side of a tool pulling what is needed. But as currently designed it does present a challenge.

The way file explore currently works (correct me if I'm wrong @sekmiller @qqmyers) is that wen you click the explore button you get the popup, you fill it out and a guestbookresponse / download is created. When the explore tool downloads the file, it sends an extra parameter to not write another guestbook entry (in order to not double count). 

I think once we have the concept of guestbook entries existing independent of downloads, it should be easier, because then download record wouldn't be created until the tool actually fetches a file (and attaches to the existing download). But that doesn't seem in scope for this issue.

Re: restricted, my question is really about when to show the button. Draft shouldn't be an issue, because if you can see the draft you can download all the files. We could go with something simple  like if all files are available to you, show the button, or if at least  one file is available, show the button. I lean towards the former for now, as that would still work for all cases were the files are all public and seems "safer". (not from a security perspective, but from a "will the user be confused" perspective).

scolapasta:
@qqmyers yes, I think we're saying the same thing (our comments crossed paths).

djbrooke:
Thanks @qqmyers and @scolapasta. If I'm understanding correctly, the only truly scalable solution here seems to be to always show the button and rely on the external tool to provide the messaging about why it can't work with the files in the dataset, instead of trying to add logic to the button display.

src/main/resources/db/migration/V4.15.1.1__5028-dataset-explore.sql
landreev: So what is the status of this - is this being addressed? I.e., is the bug in ExternalToolServiceBean getting fixed in this PR, and is ContentType being made nullable?

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
landreev: If I understand @scolapasta's suggestion - we may have a use for the content type, with *some* dataset-level external tools; but it sounds like it still needs to be nullable in the table; and it still should be null by default. 
In other words, if we adopt this scheme, then we definitely DON'T want to set the type to tab-delimited by default.

doc/release-notes/5028-dataset-explore
```diff
@@ -0,0 +1,5 @@
+There is a DB migration at: 
+
+https://github.com/IQSS/dataverse/blob/5028-dataset-explore-btn/src/main/resources/db/migration/V4.15.1.1__5028-dataset-explore.sql
+
+We will need to reflect this in the release notes. Verify at release time.
```
landreev: Since this is a flyway script, and no interactive steps are needed for this migration - do we even need to mention this script in the release notes?

djbrooke: I added this because I thought it was needed. I'll remove it.

djbrooke: Removed in https://github.com/IQSS/dataverse/pull/6059/commits/c0df37dd564214fd53f3e05b722b97d3a453b0b6

qqmyers:
I'd add checking that the file-level tools still work after a dataset-level tool is added to the database, i.e. to make sure the file-level code to list tools properly ignores the dataset-level one(s).

scolapasta:
@djbrooke asked me to try and summarize what is remaining on this issue. I reviewed all the comments (and did a quick review myself, and this is what I could come up with (but feel free to suggest other things if I missed something):
- [ ] allow content type to be null (since Dataset level tools won't require it (and in theory, there may be a space for a file level tool that could work any any tool?
- [ ] add File PID as parameter
- [ ] require proper parameter based on scope (i.e. currently requires fileID for files; should require either dataset id or dataset pid for dataset scope)
- [ ] on DatasetPage change the finds for file tools to use findByScopeAndType

doc/sphinx-guides/source/installation/external-tools.rst
```diff
@@ -59,6 +62,15 @@ To list all the external tools that are available in Dataverse:
 
 ``curl http://localhost:8080/api/admin/externalTools``
 
+Showing an External Tool in Dataverse
```
scolapasta: What's the purpose / need for this API? (besides having it in a test?)

pdurbin: It's for sysadmins to check which external tools have been installed.

scolapasta: Sure, for curl http://localhost:8080/api/admin/externalTools, which lists the tools, I'm asking specifically about this new one that takes an id. i.e if you have the id already then you already know about the tool. Doesn't seem to add any value, but maybe I'm missing something.

pdurbin: I can imagine a somewhat paranoid sysadmin or tool maker adding a tool and then wanting to double check that correct values were saved in the database. For example, here I'm asserting that the tool that was just added is a dataset level tool but you could make assertions on other things:

        Response getTool = UtilIT.getExternalTool(id);
        getTool.prettyPrint();
        getTool.then().assertThat()
                .body("data.scope", CoreMatchers.equalTo("dataset"))

To me, this is all basic CRUD stuff. We're doing a read of a single item. I meant to add this back when I first implemented external tools but forgot about it. I'm trying to correct an oversight, something I forgot to add.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
```diff
@@ -150,10 +150,10 @@ public static ExternalTool parseAddExternalToolManifest(String manifest) {
         String scopeUserInput = getRequiredTopLevelField(jsonObject, SCOPE);
         String contentType = getOptionalTopLevelField(jsonObject, CONTENT_TYPE);
         //Legacy support - assume tool manifests without any mimetype are for tabular data
-        if(contentType==null) {
-            contentType=DataFileServiceBean.MIME_TYPE_TSV_ALT;
+        if (contentType == null) {
```
scolapasta: Should this only be true for file scope tools?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1893,5 +1896,35 @@ public Response getMakeDataCountMetric(@PathParam("id") String idSupplied, @Path
         }
     }
 
+    @GET
+    @Path("{id}/externalTools")
+    public Response getExternalTools(@PathParam("id") String idSupplied, @QueryParam("type") String type) {
```
scolapasta: I get the idea of moving these to each of the Dataset and File APIs, but there is a lot of redundancy - it could be a nice clean up to have the core of the code centralized.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
```diff
@@ -45,45 +45,50 @@
...
-        if (toolsFromQuery != null) {
-            externalTools = toolsFromQuery;
-        }
-        return externalTools;
+    /**
+     * @param type explore or configure
+     * @param contentType file content type (MIME type)
+     * @return A list of tools or an empty list.
+     */
+    public List<ExternalTool> findFileToolsByScopeAndContentType(Type type, String contentType) {
```
scolapasta: this method should by not be "by scope" since scope is file or type (personally I think it's overkill to have separate methods for file and dataset, but at least they wrappers around a single method that takes in scope as a parameter)

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
```diff
@@ -45,45 +45,50 @@
...
     /**
-     * @param type
+     * @param type explore or configure
      * @return A list of tools or an empty list.
      */
-    public List<ExternalTool> findByType(Type type) {
-        return findByType(type, null);
+    public List<ExternalTool> findDatasetToolsByType(Type type) {
+        String nullContentType = null;
+        return findByScopeTypeAndContentType(ExternalTool.Scope.DATASET, type, nullContentType);
```
scolapasta: you could just pass null in the last argument; it's fairly self explanatory

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
pdurbin: Whoops! Fixed in 94f812724 by removing "scope". Good catch!

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
pdurbin: Calling the variable "nullContentType" was a deliberate choice to improve readability of the code. We can agree to disagree. 😄

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolServiceBean.java
pdurbin: @scolapasta I don't have a strong opinion about this. I'll leave it alone unless you want me to change it.

Right now the docs reflect current state of the code (the "if" removed) in that I removed the statement in parenthesis highlighted in blue below.

Screen Shot 2019-08-29 at 4 15 43 PM

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: Good idea. In e92c8f12d I reduced the duplicate code, consolidating it.

pdurbin:
@scolapasta I just pushed some improvements based on your feedback. Please take another look.

kcondon:
OK, so since I was a bit confused when I tested this, I'm noting the scope with help from Phil:
1. add dataset level tool
2. uses either dataset id or dataset pid
3. end user interrogate tools endpoints for both file and datasets
4. add filepid as an option
5. list specific external tool by id
6. regression test existing file tool endpoints
7. ExternalToolsIT (API tests) now being added to the API test suite.
8. contentType column changed to nullable=true (because it doesn't make sense to have contentType for dataset level tools)
9. In previous releases, not specifying a contentType for a file level tool would make the tool operate on tabular files. It would be given the contentType we use for tabular files. This no longer happens.
10. Existing tools in the database table are converted (by flyway) to be scope=file. (scope can't be null).

Suggested changes:
1. Update dataset tool doc to indicate config tool type is not yet supported by UI but can be configured using API, just won't be callable yet.
2. Add sample json file for configuring dataset level tool
3. Explain the meaning and variations of the key/value pairs in the json file. Specifically, that the json defines the tool interface, contains static key/value content as well as dynamic key/value content. The dynamic content has user definable key names and reserved keyword placeholder values as for dynamic content that is filled in by dataverse and passed to the external tool when the button is clicked. Some keys are required, some are optional.

pdurbin:
Decisions from standup today:

- Change doc for adding tools to say that there is no dataset-level button for "configure" tool.

To further discuss:

- For new end user "what tools can operate on this file or dataset?" endpoint, respect permissions.
- Private URL and API tokens.

pdurbin:
@kcondon @TaniaSchlatter @scolapasta @mheppler in e2e34d3 I just pushed a fairly significant reorganization of our documentation for external tools with an eye toward these goals or thoughts that I expressed during design stand up this morning:

- As we did with the API Guide, let's acknowledge that there are multiple audiences for external tools. In the commit above I differentiate between people like @donsizemore @shlake etc. who will be installing external tools (Admin Guide) and people like @xarthisius who may have never installed Dataverse but who develops external tools for Dataverse (Developer Guide). I would further like to consider end users as a audience for external tools but I stopped short because this seems a bit controversial and because I ran out of time tonight and I need to get home. I will note two things. 1. http://guides.dataverse.org/en/4.16/user/find-use-data.html#explore-data is the current entry point for end users from the User Guide into the world of external tools and it could really use a rewrite. 2. There's already an issue at #4429 (Revise Explore Tools documentation to reflect modularity) to think about doing this rewrite, especially with regards to how to document tools like Geoconnect that are not external tools. Ok, a third thing... I don't think we document configure tools at all in the User Guide. I know they're new and there's only one (awesome!) one. But we probably should.
- The better we document the process for creating external tools, the more likely toolmakers will have an easier time creating them. Better docs, more tools. We hope.

I still need to do some cleanup of the docs but here's a preview of one of the tables I added:

Screen Shot 2019-09-09 at 6 12 07 PM

kcondon:
@pdurbin Looks good but at some point this should be a separate doc ticket! l)

pdurbin:
@kcondon heh.

At standup I noticed that one of the tables I added wasn't rendering. @TaniaSchlatter and @scolapasta this what I was trying to show you:

Screen Shot 2019-09-10 at 5 27 23 PM

As we discussed, my next steps are:

- Move the new "Building External Tools" to the API Guide to allow the Developer Guide to continue to have a main focus on the core (Java) application.
- Continue to improve the "Creating an External Tool Manifest" section, especially the dense and complicated text in the screenshot below:

Screen Shot 2019-09-10 at 5 31 10 PM

landreev:
The name of the flyway script - "V4.16.1__..." may be breaking our convention - it should have been "V4.16.0.1__...", correct? 
Also, "V4.16.0.1__..." is already being used in #6042, so this one should probably be "0.2".

pdurbin:
@landreev woof, thanks for the heads up. I guess I'll keep and eye on that other pull request. I assume it'll get merged first.

scolapasta:
For the end user "what tools can operate on this file or dataset?" endpoint, I still don't see the utility of it from the API. What is returned is NOT which tools can operate, but which tool will show up in the UI. You can always use the tools on the file (or dataset) if you are able to construct the URL, this is just constructing the URL so it can display the link on the UI.

In other words this is an API for a UI only (by design) feature.

The only argument I see for keeping this API is for testing (code coverage) which is valid. In that case it should moved, ideally, to a new API prefix ("test")  We will likely be getting more of these test only APIs as we focus on Integration tests. I also don't think these therefore need to be documented in a user or admin facing guide as they would not be called by anyone but the integrations tests and could be changed / removed at any moment (if the underlying code changes).

pdurbin:
@scolapasta instead of a brand new API namespace called "test", can we use "admin/test"? That way, we don't have to rewrite the "Securing Your Installation" part of the Installation Guide which only mentions blocking "admin" and "builtin-users":

Screen Shot 2019-09-11 at 5 06 53 PM

Also, I can imagine UIs other than JSF. Maybe it's React or Vue or Angular and calling into Dataverse APIs to construct the GUI, including dropdowns of "Explore" buttons. But this is a conversation for another day, maybe a future tech hours! 😄

pdurbin:
@kcondon unfortunately PDF versions of the guide are failing in this branch as well as "develop" so I just opened #6168.

Issue #6092, opened 2019-08-14
File metadata updates not reindexed


qqmyers:
On v4.15.1 (demo.dataverse.org today and locally), it looks to me like edits of file metadata (I checked name and category tags specifically) are not being reindexed, which manifests itself as searches failing to find files by their updated name and the Filter by Tag facet either not appearing or not adding a new tag. It does look like file metadata is being indexed during upload (so I can make the Filter by tag facet appear by adding tags during upload, but subsequent tag edits don't change anything).

Any guidance /help appreciated. Otherwise, I'll start looking into the code.

qqmyers:
I think this is limited to edits made from the file page. Those appear to go through EditDatafilepage's save() method without invoking a command or reindexing directly. 

Also - I see evidence that the reidex command being commented out for #4425 (https://github.com/IQSS/dataverse/blame/develop/src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java#L1437) is actually used when you select add metadata for files selected on the dataset page and then use the edit/tag dialog after that. I.e. removing that line should be causing a similar bug in dev and will affect post 4.15.1 releases if not reverted.

I can put in a fix for both issues if that's useful.

pdurbin:
@qqmyers yes, please fix all the bugs. If you have any question, please don't hesitate to ask!

It sounds like we need more API tests. There are some search/index related tests in "SearchIT": https://github.com/IQSS/dataverse/blob/v4.15.1/src/test/java/edu/harvard/iq/dataverse/api/SearchIT.java

PR #6094, opened 2019-08-14
reindex when datasetUpdateRequired==false

qqmyers:
and no Commands (which would trigger reindexing) are called.

## New Contributors

## Related Issues

- ## #6092 File metadata updates not reindexed

## Pull Request Checklist

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
@@ -5,6 +5,7 @@ import edu.harvard.iq.dataverse.api.AbstractApiBean;
 import edu.harvard.iq.dataverse.authorization.AuthenticationServiceBean;
 import edu.harvard.iq.dataverse.authorization.Permission;
 import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
+import edu.harvard.iq.dataverse.batch.util.LoggingUtil;
 import edu.harvard.iq.dataverse.branding.BrandingUtil;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
```diff
@@ -1377,6 +1379,12 @@ public String save() {
                         dataset = datasetService.find(dataset.getId());
                     }
                 }
+                //We also need to reindex since no commands that include reindexing have been called.
+                indexService.indexDataset(dataset, true);
```
qqmyers: To fix the bug in 4.15.1, I could have added this at line ~1430, but reindexing is also needed at ~1438 for another case (see issue). I think adding it here catches both cases and avoids reindexing when it isn't needed (as would be the case if 1438 were uncommented).

I also tried to mimic the new exception handling/logging from #4425.

scolapasta: So, really this logic *should* go through a Command. Not sure how it got in without that. We'll need to decide if we want his simple fix or require the refactoring to a command.

(my guess is we'll seetle on accepting this but open up a new issue for the refactoring - being in a command is important for permissions, automatic logging, and allowing same code for UI and API.

qqmyers: +1 - it might be as easy as always forcing datasetUpdateRequired == true  as was done with prov around line 1178 (or just removing that flag completely).

qqmyers:
It's been too long, but I think with #6248, this became obsolete, so I'm closing it.

PR #6101, opened 2019-08-19
6007 CORS

### LINKED ISSUE #6007: CORS problem for "PUT" API
At present Dataverse does not allow to use "PUT" APIs from external websites. But external tools sometimes will need such kind of ability. For example Data Curation Tool will need to use EditDDI but prevented by CORS. 
The easy solution would be to add the following lines into ApiBlockingFilter.java
```
((HttpServletResponse) sr1).addHeader("Access-Control-Allow-Origin", "*");
((HttpServletResponse) sr1).addHeader("Access-Control-Allow-Methods","PUT");
((HttpServletResponse) sr1).addHeader("Access-Control-Allow-Headers", "Content-Type, X-Dataverse-Key");
```
Regarding security, in order to use "PUT" API one usually will need an API token, website that does not have a token will not be able to use the API and benefit from ("Access-Control-Allow-Origin", "*").
I would like to know what Dataverse Community thinks about it

lubitchv:
External tools such as DataCurationTool needs to use "PUT" Api such as EditDDI from its application, but the application may be located in different domain from dataverse (at github for example). At the moment it is prevented by CORS. ApiBlockingFilter.java was changed to allow to call "PUT" api.

## Related Issues

- closes #6007 

## Pull Request Checklist


djbrooke:
@lubitchv - my apologies for not mentioning the existing solution on your original issue #6007 before you started the work here. Thankfully this diff looks small. :) Thanks @pdurbin for the review. 

@scolapasta - I would be interested in your guidance here about which solution is better for both current needs and for anticipated future integrations.

michbarsinai:
Seems that this change will allow CORS on all PUT endpoints. I get the point in the GitHub issue about requiring API token, but I think this might be too broad, and we can use the existing allowCors() on a per-endpoint basis.
As far as I understand, CORS is required only when we want JS code running in browsers to make API calls. Other systems can make these calls as they like (as long as they have API Token). I'm probably missing some context here, but I'm not sure why this is necessary.

scolapasta:
I discussed with more extensively with @michbarsinai. The reasons we want to be careful with CORS are for security (which is handled by API tokens) and server load. For some smaller installations, it may not be great to allow CORS from any site. So let's go with this solution:

• add setting to disable allowing of CORS (i.e. without the setting the default is allow CORS)
• user this idea of adding to the ApiBlockingFilter, but a) make sure to check the setting, b) allow for all CORS (not just PUT)
• remove the current allowCors methods, as they would now be redundant

@lubitchv is that something you can work on?

lubitchv:
Hi @scolapasta, yes I can work on it. Although it is not clear to me what do you mean by settings disabling or allowing CORS, should it be in jvm-options (not allow CORS) or in some other place?

scolapasta:
@lubitchv It can be a setting in the settings table.

lubitchv:
Ok @scolapasta, I understand, will do it that way.

lubitchv:
I added AllowCors setting and also a check in ApiBlockingFilter. I also removed allowCors function.

src/main/java/edu/harvard/iq/dataverse/api/Meta.java
```diff
@@ -178,8 +174,6 @@ public String dataset(@PathParam("datasetId") Long datasetId, @QueryParam("exclu
             throw new ServiceUnavailableException();
         }
 
-        response.setHeader("Access-Control-Allow-Origin", "*");
-
```
pdurbin: In 8c7fb34 we flagged parts (or all?) of this API as deprecated and there's a test that calls `getMetaDatafileDeprecated` which seems to exercise the `@Path("datafile/{fileId}")` method. I don't know if `/api/meta` is still used by anything. I don't believe it's even documented. And Meta.java does not extend AbstractApiBean so I don't think it has access to allowCors or any other methods in there. Since CORS is now (in this pull request) enabled in ApiBlockingFilter, this "meta" API will still be subject to the new ":AllowCors" setting, right?

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -514,8 +506,6 @@ public DownloadInstance tabularDatafileMetadataPreprocessed(@PathParam("fileId")
             logger.fine("Preprocessed data for tabular file "+fileId);
         }
         
-        response.setHeader("Access-Control-Allow-Origin", "*");
-
```
pdurbin: This "dataaccess" API is subject to ":AllowCors", right? I assume so since it looks like the setting affects a low level, the ApiBlockingFilter.

lubitchv: I think so. I did not test that specific API but  I tested tabularDatafileMetadata API from dataaccess and it works with the new setting.

pdurbin: Ok, it should be easy enough to test, right? With `curl -i` you can see the headers.

lubitchv: Ok, thanks, I just run it and it has the required headers.

src/main/java/edu/harvard/iq/dataverse/api/Meta.java
lubitchv: Yes, it will.

Issue #6111, opened 2019-08-22
Update Codebook Docs


djbrooke:
We should expand on the HTML Codebook docs added in #6081. 

PR from @jggautier incoming!

PR #6115, opened 2019-08-23
Edited Supported Metadata Export Formats section [ref: #6111]

jggautier:
Reworded Supported Metadata Export Formats section, added info about DDI HTML Codebook   #6111

- connects to #6111 Update Codebook Docs

### DIFF
--- a/doc/sphinx-guides/source/user/dataset-management.rst
+++ b/doc/sphinx-guides/source/user/dataset-management.rst
@@ -25,16 +25,18 @@ For more details about what Citation and Domain Specific Metadata is supported p
 Supported Metadata Export Formats
 ---------------------------------
 
-Once a dataset has been published its metadata is exported in a variety of formats. A button on the dataset page's metadata tab will allow a user to export the metadata of the most recently published version of the dataset. Currently supported export formats are:
+Once a dataset has been published, its metadata can be exported in a variety of other metadata standards and formats, which help make datasets more discoverable and usable in other systems, such as other data repositories. On each dataset page's metadata tab, users can download metadata in the these metadata exports:
 
 - Dublin Core
-- DDI (Data Documentation Initiative)
-- DDI HTML Codebook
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -25,17 +25,19 @@ For more details about what Citation and Domain Specific Metadata is supported p
 Supported Metadata Export Formats
 ---------------------------------
 
-Once a dataset has been published its metadata is exported in a variety of formats. A button on the dataset page's metadata tab will allow a user to export the metadata of the most recently published version of the dataset. Currently supported export formats are:
+Once a dataset has been published, its metadata can be exported in a variety of other metadata standards and formats, which help make datasets more discoverable and usable in other systems, such as other data repositories. On each dataset page's metadata tab, users can download metadata in these metadata exports:
```
pdurbin: Instead of...

"users can download metadata in these metadata exports:"

... I would suggest...

"users can download metadata in the following formats."

jggautier: I didn't want to say "formats" because formats to me are XML, HTML, JSON, etc. What's being listed are the metadata standards: Dublin Core, DDI, etc. But, now we're listing an export that uses the same standard, DDI, but different format, HTML. I was hoping that writing that "users can download metadata in these metadata exports" would be quicker than writing that "users can download metadata in these metadata standards and formats".

But I agree the current phrasing is awkward. And the guides never directly make the distinction between "formats" and "standards" as I understand them (or at all). Here are a few other options I think can of:

- Write "users can download metadata in the following standards and formats", and in the list include the formats for each export
- Write that "users can download metadata in these metadata standards and formats"
- Write "users can download metadata in the following standards " and remove mentions of format in the paragraph above the list. ("Once a dataset has been published, its metadata can be exported in several metadata standards, which help..."

Issue #6140, opened 2019-09-03
Housekeeping: remove orphaned Java Endorsed API JAR moving from Maven POM


poikilotherm:
Related to #5360. Needed by #6132. 

----

In the Maven POM, there is a implicit (not transitive nor direct) dependency:

https://github.com/IQSS/dataverse/blob/09fe94bdc6f4f5e79c61a203b7df8736692657ea/pom.xml#L705-L708

This is present since the initial commit. I removed it during #6133, but @pdurbin asked me to separate it. I am more than happy to create a PR with 7bc6c6f in it, so we can get #6134 working in my K8s development stuff.

I would appreciate this to be done quickly. This is a blocker.

poikilotherm:
This is solved by merging 6141. Closing.

PR #6141, opened 2019-09-03
6140 - remove orphaned Endorsed Java API specials

poikilotherm:
Since 2013, the initial commit d6868ad, added the endorsed Java API.
These days, there is no reference to be found installing the resulting files.
The API was not included in the WAR file.

## Related Issues

- #6140

## Pull Request Checklist

- [X] Merged latest from "develop" [branch][] and resolved conflicts

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,6 @@
     <packaging>war</packaging>
     <name>dataverse</name>
     <properties>
-        <endorsed.dir>${project.build.directory}/endorsed</endorsed.dir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <additionalparam>-Xdoclint:none</additionalparam>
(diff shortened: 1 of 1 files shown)


pdurbin:
@poikilotherm I think I'm ready to move this over but how should QA test it? Can you please write a little bit about this here? Thanks!

poikilotherm:
Actually there is not a single code reference. No docs reference. Just nothing. So I think it should be fine to just do the normal tests QA does for every PR/merge.

pdurbin:
@poikilotherm ok. Obviously, testing is different for pull requests for docs (Sphinx) than anything touching the application. Since this pull request touches the application, I agree that we're looking to make sure that nothing breaks. Like you said, there should be no business logic changes so a smoke test is probably enough. I'll move this over. Thanks.

poikilotherm:
Thx for merging! :smile:

PR #6143, opened 2019-09-05
6035 memory leaks

### LINKED ISSUE #6035: Memory leak when dataverse is stressed with GET requests

landreev:
## New Contributors

## Related Issues

closes #6035

## Pull Request Checklist


landreev:
One specific area I want somebody who understands how different types of authentication work to review:
Our strategy is to keep the anonymous sessions short - by keeping the default session timeout (the value configurable in web.xml) short; and then increase the timeout for logged-in users, to something configurable as a setting. 
It looks like this needs to happen in several places, not just the login page. I'm doing it in the following places: 
```
src/main/java/edu/harvard/iq/dataverse/LoginPage.java
src/main/java/edu/harvard/iq/dataverse/authorization/providers/builtin/DataverseUserPage.java
src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/OAuth2FirstLoginPage.java
src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/OAuth2LoginBackingBean.java
src/main/java/edu/harvard/iq/dataverse/confirmemail/ConfirmEmailPage.java
src/main/java/edu/harvard/iq/dataverse/passwordreset/PasswordResetPage.java
src/main/java/edu/harvard/iq/dataverse/privateurl/PrivateUrlPage.java
src/main/java/edu/harvard/iq/dataverse/Shib.java
```
It's possible that it's not strictly necessary in some of the places above (in some cases the user may be redirected back to the login page, for example, and the session timeout will be increased there - ?) - but it shouldn't hurt either. I'm more concerned about missing something.

doc/sphinx-guides/source/developers/tools.rst
```diff
@@ -145,6 +145,139 @@ For example...
...
+will output a list of all classes, sorted by the number of instances of each individual class, with the size in bytes. 
+This can be very useful when looking for memory leaks in the application. Another useful tool is ``jstat``, that can be used in combination with ``jmap`` to monitor the effectiveness of garbage collection in reclaiming allocated memory. 
+
+In the example script below we stress running Dataverse applicatione with GET requests to a specific dataverse page, use ``jmap`` to see how many Dataverse, Dataset and DataFile class object get allocated, then run ``jstat`` to see how the numbers are affected by both "Young Generation" and "Full" garbage collection runs (``YGC`` and ``FGC`` respectively):
+
+(This is script is provided **as an example only**! You will have to experiment and expand it to suit any specific needs and any specific problem you may be trying to diagnose, and this is just to give an idea of how to go about it)
+
+.. code-block:: bash
+
+   #!/bin/sh
```
pdurbin: We could refactor this into a downloadable script and also continue to inline it. Here's an example of where we do that (the "clear timer" script): http://guides.dataverse.org/en/4.16/admin/troubleshooting.html#deployment-fails-ejb-timer-service-not-available

kcondon:
Tested, this is ready to go, pending a release notes update.

PR #6177, opened 2019-09-18
6175 an exception is logged as fine entry during email verification

### LINKED ISSUE #6175: An exception is logged as FINE entry during email verification
in the `findSingleConfirmEmailDataByUser` method (edu.harvard.iq.dataverse.confirmemail.ConfirmEmailServiceBean#findSingleConfirmEmailDataByUser in src/main/java/edu/harvard/iq/dataverse/confirmemail/ConfirmEmailServiceBean.java) if the user is not found exceptions are catched and logged as FINE log entry:

```
try {
  confirmEmailData = typedQuery.getSingleResult();
} catch (NoResultException | NonUniqueResultException ex) {
  logger.fine("When looking up user " + user + " caught " + ex);
}
```

For me it is strange. I think FINE entries are for those information which happen on the main, expected workflow. If we run into an exception, it should not be the main way, and we should make a distinction putting this into at least a warning or a higher level of log entry.

In this particular case it would be useful to log the SQL query if it is possible.

The current resulted log entry is something like this:

```
When looking up user [AuthenticatedUser identifier:@<useridentifier>] 
  caught javax.persistence.NoResultException: getSingleResult() did not retrieve any entities.]]
```
(It is a single line, I just break it because of the display of the ticket.)

The message is misleading, because this query runs on the ConfirmEmailData table, and it means that the user ID is not found in the table used for storing user data temporary in the email confirmation process, and not as the log entry might suggest the user is not existing in the system.

pkiraly:
## Related Issues

- Fixes #6175: An exception is logged as FINE entry

## Pull Request Checklist

- [ ] Unit [tests][] None
- [ ] [Documentation][docs] None


pdurbin:
@pkiraly I added "fixes" above so that the issue will be closed automatically if/when this pull request is merged. Thanks!

pdurbin:
@pkiraly can you please take a look at the feedback from @kcondon at https://github.com/IQSS/dataverse/issues/6175#issuecomment-532691728

It sounds like normal operations on the user account page are causing a lot of noise in server.log now. Are you seeing the same thing? Thanks.

pdurbin:
@pkiraly I'm looking at some of your old PRs. Sorry it has taken so long. This one doesn't have a branch anymore:

```
error: pathspec '6175-an-exception-is-logged-as-FINE-entry' did not match any file(s) known to git
```

I'm going to go ahead and close it. If you'd like to make a fresh PR, please go ahead.

Issue #6124, opened 2019-08-28
Add code coverage for integration tests


scolapasta:
This issue is related to #4960. 

That issue was closed once documentation was added, but let's go further and actually add this to our process. In particular, with our new focus on integration tests over unit tests, this should help us fine which areas of the application are not being exercised by the current test so we can add new tests.

- Percentage of coverage
- Areas exercised and those not exercised

It would be great to include some coverage metric in the readme like we do for coveralls.

pdurbin:
In pull request #6186 I improved the docs on how to get a code coverage report for API tests out of Dataverse. Sadly, this has not yet been automated. Just today @poikilotherm asked me if we plan to use Maven for this and we probably should but right now we're using a script and some documented commands. This issue isn't a sprint anymore though so I guess it'll have to wait unless someone from the community picks it up. Here are the new and improved docs from the pull request I mentioned: http://guides.dataverse.org/en/4.17/developers/testing.html#measuring-coverage-of-integration-tests

donsizemore:
Currently, Jenkins/Ansible are spitting out this:

![Screen Shot 2020-02-25 at 07 56 26](https://user-images.githubusercontent.com/12831666/75249468-62ad5080-57a4-11ea-9c21-754c41ac62f2.png)

and this: https://jenkins.dataverse.org/job/IQSS-dataverse-develop/364/jacoco/

next to review with @pdurbin and @scolapasta and decide which statia to publish.

pdurbin:
> It would be great to include some coverage metric in the readme like we do for coveralls.

I just split this off into its own small chunk:

-  As a developer, I want a button in the README saying the percentage code coverage for the API test suite (integration tests) #6682

djbrooke:
Much of this is now available at the bottom of the readme! Thanks @donsizemore!

Issue #5483, opened 2019-01-23
As a developer, I want to see code coverage from Netbeans


pdurbin:
Over at http://guides.dataverse.org/en/4.10.1/developers/testing.html#observing-changes-to-code-coverage we talk about how developers can check code coverage right from Netbeans. Here's what's written there:

![screen shot 2019-01-23 at 5 13 10 pm](https://user-images.githubusercontent.com/21006/51640690-32a4d580-1f32-11e9-9136-e49f783f6772.png)

Here's how the red/green is supposed to look:

![screen shot 2019-01-23 at 5 11 52 pm](https://user-images.githubusercontent.com/21006/51640714-405a5b00-1f32-11e9-90d7-969bc8a0b4d3.png)

I think the upgrade to JaCoCo 0.8.2 in pull request #5289 broke JaCoCo/Netbeans integration. the  caused it. There's a comment at https://github.com/jacoco/jacoco/issues/772 that says this:

> AFAIK NetBeans integration reads XML Report, which was intentionally changed in version 0.8.2 in a perfectly valid way by addition of new attribute - see https://github.com/jacoco/jacoco/pull/702. NetBeans integration must be fixed accordingly and ideally be less strict to cope with such updates in future. This problem should reported to developers of NetBeans integration, since it is not developed by us - as you can see on page https://www.jacoco.org/jacoco/trunk/doc/integrations.html

I just tried JaCoCo 0.8.1 and the integration works again. I'll create a pull request.

pdurbin:
I reported the problem on the Netbeans mailing list and based on replies it sounds like I'm on the right track: Does JaCoCo 0.8.2 break NetBeans code coverage? https://lists.apache.org/thread.html/509bfddb9273726a6e8dff1ca419790b40858d6a648b4413142e2abe@%3Cusers.netbeans.apache.org%3E

pdurbin:
In the Netbeans mailing list thread above @geertjanw asked me to open an issue so I did: https://issues.apache.org/jira/browse/NETBEANS-1997

PR #6186, opened 2019-09-18
more docs on code coverage for integration tests #6124

pdurbin:
Relates to #6124

As I was saying at standup today, the first step is to improve the documentation for how to measure code coverage for integration tests. That's what this pull request is about.

I have lots of hopes and dreams for where to go from here. I'm still using the very helpful script @pameyer provided. We should probably put the logic in Maven as well (but it doesn't hurt to keep the script and docs because users might want to instrument an existing war file).

I'm hoping these docs will be useful to @donsizemore and he and I work on automating some of this using Ansible in https://github.com/IQSS/dataverse-ansible/issues/67

But the first step, again, is better docs.

### DIFF
--- a/doc/sphinx-guides/source/_static/util/instrument_war_jacoco.bash
+++ b/doc/sphinx-guides/source/_static/util/instrument_war_jacoco.bash
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-JACOCO_HOME=${HOME}/local/jacoco-0.8.2/lib/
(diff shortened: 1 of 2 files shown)


donsizemore:
@pdurbin looks good to me, but why are we using Jacoco-0.8.1 instead of 0.8.4? An idle question at best...

pdurbin:
> why are we using Jacoco-0.8.1

Because of #5483 and https://issues.apache.org/jira/browse/NETBEANS-1997

PR #6206, opened 2019-09-23
Update Whole Tale ext tool entry

Xarthisius:
As per https://github.com/whole-tale/wt-design-docs/issues/95#issuecomment-534252381 I'm updating the scope of Whole Tale external tool.


pdurbin:
The dataverse-external-tools.tsv file @Xarthisius modified in this pull request is the "source" two tables:

- Inventory of External Tools in the Admin Guide
- Examples of External Tools in the API Guide.

It changes "Scope" from "file" to "dataset" for Whole Tale. It'll be nice to ship the new dataset explore feature with an example of a working dataset level tool! Thanks, @Xarthisius !

PR #5863, opened 2019-05-21
2043 split gbr table

mdmADA:
## Related Issues

- #2043 - Guestbook - Enabled for Request File Access

Submitting code to split the guestbookresponse table into 2 tables: guestbookresponse and filedownload

This is the first step in moving towards enabling the Guestbook at Request Access so that dataset owners can ask questions of users when they request access and use their answers as a basis to grant or reject access.

I created the FileDownload.java class and modified the GuestbookResponse.java class and created a one-to-one relationship between the two with the GuestbookResponse as the 'parent' and FileDownload as 'child'. guestbookresponse_id in filedownload is a foreign key to guestbookresponse.id and is also the primary key of filedownload.

**Questions I have:**

**1.** Not sure if, when a file is not restricted and has no guestbook, **should 'Guest' be written** to the guestbookresponse.name field? It was (verified) in previous versions (ex. 4.6.1) but isn't now.

I know 'GUEST' has been removed from prefilling the Guestbook pop-up field but the scenario I am asking about is when a datafile is open and the dataset has no guestbook so there is no pre-filled 'Guest' entry at all.

I think writing 'Guest' (or language specific equivalent) into the name when there is no guestbook is a good idea otherwise, when I look at the table and there is nothing written to any of the user-specific-fields, I wonder if that is because there was an issue/bug that prevented the details from being written. Having 'Guest' indicates to me as an admin user that the download process worked properly.

**2.** Not sure if the **timestamp fields are required in BOTH** guestbookresponse AND in filedownload or just one of them.

**3.** Not sure if the DATAFILE_ID, DATASET_ID, DATASETVERSION_ID etc. **should be moved** to filedownload since these fields ARE related to the filedownload itself.

**4.** In my meeting with @scolapasta so long ago now, we said the **assumption** is that if there is a download there will be a guestbookresponse
but Q: Is there a case where there will be a download without a guestbook response?

**5.** I am not sure **how to include the database scripts** related to the filedownload creation and guestbookresponse alteration (and inserting any guestbookresponse fields into the new filedownload fields) in the pull request:

#upgrading to version xxx
CREATE TABLE FILEDOWNLOAD(
DOWNLOADTYPE VARCHAR(255),
DOWNLOADTIMESTAMP TIMESTAMP,
SESSIONID VARCHAR(255),
GUESTBOOKRESPONSE_ID BIGINT NOT NULL,
PRIMARY KEY (GUESTBOOKRESPONSE_ID)
);

[...]

Issue #2043, opened 2015-04-17
Guestbook - Enabled for Request File Access


posixeleni:
Users with restricted files would like to enable guestbook like functionality so that they can find out who and why someone would like to access their data. This will help inform the user on whether or not they should grant or reject access to a file.

This suggestion was requested by a user in our Google Community Forum: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/dataverse-community/Xn-SsIEaRf8/LwtvL8exXEAJ

sbarbosadataverse:
@eaquigley 
@pdurbin 

Contact from user today regarding this function. In 3.6, clicking on username allowed admins to see: names, affiliations, and email. Also mentioned they can't grant access to their users until this is resolved which bumps it up to high/critical request
We need a function that resembles this.

mcrosas:
@eaquigley can you this to the In Design milestone, so we can push it to an upcoming release as soon as possible? This is urgent. From users the inability to verify users information is equivalent to the system being down because they cannot process requests.
@scolapasta this is high priority for an upcoming release.

eaquigley:
FRD and mock up needs to be done

eaquigley:
FRD: https://docs.google.com/document/d/1qoDTXXCIBtMgLbh_xSISNkfh5dvh_afB3kDNcEUIOEo/edit?usp=sharing

eaquigley:
FRD has been completed: https://docs.google.com/document/d/1qoDTXXCIBtMgLbh_xSISNkfh5dvh_afB3kDNcEUIOEo/edit?usp=sharing 

Now just needs to be scheduled for development. No need for mockups as it is adding an extra step to an existing workflow.

stevenmce:
Per the community call on 28 Feb 2017, we would like to highlight this issue for possible inclusion in an upcoming release - particularly the requirement above of "having the guestbook recording entries that a dataset owner can review when a restricted file is requested for access." ADA can assist in describing the required workflow for this request process if that would be of assistance - we have an immediate need for this functionality in order to support of a number of data owners who use our service.

Steve McEachern, Australian Data Archive

pdurbin:
[...]

PR #6207, opened 2019-09-24
2043 alter fileaccessrequests tbl

mdmADA:
Submitting flyway script to add guestbookresponse_id column to fileaccessrequests table to allow File Access Request with Guestbook.

When a File Access Request with Guestbook is created in forked code, the guestbook data is written to the guestbookresponse table with 'Request Access' as the Download type. This worked for that code but the preferred solution, in consulting with @scolapasta is to write the corresponding guestbookresponse_id to the fileaccessrequests table.

Adding this column to fileaccessrequests will affect the @ManyToMany relationship created in DataFile.java. Feedback as to the preferred way to manage this third column in the @ManyToMany fileaccessrequests relationship in DataFile.java is appreciated.

I hope the flyway script is correct.


scolapasta:
@mdmADA Can you explain a little more what you mean about affecting the @manytomany relationship? I just want to verify that this PR can be merged now independent of that (and that figured out later, when this new column is "used"), or if if you think that needs to be resolved first?

mdmADA:
Hi @scolapasta - in my original forked code, to implement the guestbook-at-request functionality, when the guestbook is filled out at request access, it is then written to the guestbookresponse with downloadType='Request Access'.

But now in the main dataverse codebase, since guestbookresponse has been split into guestbookresponse and filedownload, without changing any code right now, the 'request access' will be written to filedownload with downloadType='Request Access'. 

Since the only things being written to filedownload should only be records related to actual file downloads, it was agreed that instead of writing 'Request Access' to the filedownload table, to add the column guestbookresponse_id to the fileaccessrequests table. 

So then on 'Request Access', the guestbookresponse would be written to the guestbookresponse table and the guestbookresponse_id written to fileaccessrequests along with the other column values (datafile_id, authenticateduser_id, guestbookresponse_id). (In the case of no guestbook-at-request-access, the guestbookresponse_id in fileaccessrequests will be null.)

This guestbookresponse_id in fileaccessrequests will also allow dataverse to pull the guestbookresponse associated with a specific fileaccessrequest to display to the user who will be granting/rejecting access.  

The issue I need to figure out is the @ManyToMany  relationship for fileaccessrequests currently specified in DataFile:
    
   *****
   @ManyToMany
    @JoinTable(name = "fileaccessrequests",
    joinColumns = @JoinColumn(name = "datafile_id"),
    inverseJoinColumns = @JoinColumn(name = "authenticated_user_id"))
   
    private List<AuthenticatedUser> fileAccessRequesters;
    public List<AuthenticatedUser> getFileAccessRequesters() {
        return fileAccessRequesters;
    }
    public void setFileAccessRequesters(List<AuthenticatedUser> fileAccessRequesters) {
        this.fileAccessRequesters = fileAccessRequesters;
    }
   *******
  
   Since fileaccessrequests now has the third guestbookresponse_id column, how do I change the above @ManyToMany relationship in DataFile to write the guestbookresponse_id value to that column? I am not sure it's possible which is why I am asking for some advice.


  After a bit of research, to replace this @ManyTonMany code above, I implemented a new class FileAccessRequest.java to replace that @ManyToMany code in DataFile. The FileAccessRequest is mapped to the fileaccessrequests table with (and the getter and setter methods):

    @ManyToOne(cascade = {CascadeType.PERSIST, CascadeType.MERGE})
    @JoinColumn(name = "datafile_id")
    private DataFile dataFile;
    
    @ManyToOne(cascade = {CascadeType.PERSIST, CascadeType.MERGE})
    @JoinColumn(name = "authenticated_user_id")
    private AuthenticatedUser user;

    @ManyToOne(cascade = {CascadeType.PERSIST, CascadeType.MERGE})
    @JoinColumn(name = "guestbookresponse_id")
    private GuestbookResponse guestbookResponse; 

   The @OneToMany set of fileAccessRequests is also added to DataFile, AuthenticatedUser and GuestbookResponse.

   This requires a lot more modification of the code, though, and I can't quite get it to work 
   properly.  It will also require, I believe, a generated id column for the fileaccessrequests table so there would be 4 columns: id, datafile_id, authenticated_user_id, guestbook_response_id.

  
 **So my questions are:**
  1. Is there a way to add the guestbookresponse_id into the code so that the guestbookresponse_id is written to the fileaccessrequests table:

    @ManyToMany
    @JoinTable(name = "fileaccessrequests",
    joinColumns = @JoinColumn(name = "datafile_id"),
    inverseJoinColumns = @JoinColumn(name = "authenticated_user_id"))

 2. If 1. isn't possible, is the FileAccessRequest (with possibly a corresponding  FileAccessRequestServiceBean) the right way to go?

I hope my explanation and questions are understandable (maybe?)...

scolapasta:
Ah, I understand now. And yes, I had forgotten that the fileaccessrequest table was generated directly via the @manytomany, i.e. there was no fileAccessRequest entity.

So, with all that, I think that you are correct, the way to move forward is to make it it's own entity and shift the relationships as you suggested above.

You also said, that you can't quite get it to work properly? How can we help? What specifically is not working there.

Another question: should the guestbookResponse relationship be one to one and not many to one? (i.e. one file can have many requests, as can one user, but a guestbookresponse should only be associated with at most one file request no?)

stevenmce:
HI Gustavo,

Re " should the guestbookResponse relationship be one to one and not many to one?"

FWIW, it seems to me that Dataverse should probably look at moving towards having both dataSET requests, as well as dataFILE requests. (Given the move to dataSET level tools in other areas) So the one-to-one relationship probably isn't something you would want, as it would inhibit this.

(We in fact had a discussion with one of our data owners last week about this - they were interested in dataSET level reporting, and thus trying to figure out how to remove all the repeated dataFILE level guestbookResponses related to a single request for access to the dataSET).

Cheers,
Steve

mdmADA:
Thanks for your response @scolapasta. I am just getting back to this again. But I think you are correct, yes, that one guestbookresponse is associated with only one file request.

I will modify my code accordingly and see how I go.

mdmADA:
Hi @scolapasta - In light of @stevenmce's comments regarding dataset-level requests, do you recommend going forward with the datafile-level request and visit the dataset-level requests in a future iteration? Or should dataset-level requests be considered/designed for in my current code?

Thanks!

scolapasta:
Actually, it's an interesting point. When we had this all in one table, and you downloaded a zip file, we had to have repeated guestbook entries because we wanted to count each file in the zip as one download. But since we've now separated, in fact have one guestbook be associated with several downloads.

@stevenmce's point was about fileaccessrequests which could be similar. if I request several files at the same time they could all be tied to one guestbookresponse. But you would have to change the relationship from guestbookresponse and download, as outlined above, as well, or you woudn't have a guestbookresponse to "attach" to each download as they happened.

The other challenge their is how to handle downloads at separate times. For example, I request files A and B, and fill out guestbookresponse 1. I am granted access, so I go and download File A. That download is associated with guestbookresponse 1. We'd still need to track that guestbookresponse 1 is the guestbook response to associate with a future download of File B.

I'd be perfectly happy with (and prefer) this approach, though of course it's more work and probably still requires some more details in the design.

I'm also going to ping @sekmiller and @landreev as they've had more direct experience working with guestbooks and downloads to make sure I'm not missing anything.

pdurbin:
@mdmADA hi, I just noticed that your `2043-alter-fileaccessrequests-tbl` branch has been deleted. Are you still interested in getting this pull request merged? If so, we'll probably need a new branch.

qqmyers:
I'm not sure this is needed with #5863 merged (could be wrong). FWIW - I'm working on an update of #2043 for QDR.

pdurbin:
Closing in favor of this new PR:

- #9599

PR #6244, opened 2019-10-02
Don't return file when some derived info is requested.

### LINKED ISSUE #6240: Thumbnail api call returns original file when thumbnail not available
Calling /api/access/datafile/\<id\>?imageThumb=400 works for file types that have a thumbnail/where a thumbnail can be generated, but it just sends the original file for filetypes for which thumbnails can't be generated. It seems like a 404 would be a better default (avoids downloading large files accidentally and I think it stops the bad thumbnail call from counting a download). 
For a quick look, it appears that not being able to create a thumbnail causes the execution to hit this in the Access class:
```
            } else {
                // Service unknown/not supported/bad arguments, etc.:
                // TODO: throw new ServiceUnavailableException(); 
            }
```
Unless there are objections, or better ideas, I can look into throwing this exception or otherwise replying with a 404.

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Note that we use the "closes" syntax below to trigger Github's automation to close the corresponding issue once the pull request is merged.

Thanks for your contribution to Dataverse!

## Related Issues

- closes #6240 Thumbnail api call returns original file when thumbnail not available

## Pull Request Checklist


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -312,66 +313,78 @@ public DownloadInstance datafile(@PathParam("fileId") String fileId, @QueryParam
             downloadInstance.setDataverseRequestService(dvRequestService);
             downloadInstance.setCommand(engineSvc);
         }
+        boolean serviceRequested = false;
+        boolean serviceFound = false;
```
qqmyers: The main change is to run through the loop of query params and to set serviceRequested = true if any of the ones required by a service are sent, and to then run the checkIfServiceSupportedAndSetConverter ~as before (now only run on service params) to decide if the specific service needed exists.

After the loop, a check to see if a service was requested but not found, and, if so, to throw a not found exception. Otherwise, we're good to go (as before)

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -312,66 +313,78 @@ public DownloadInstance datafile(@PathParam("fileId") String fileId, @QueryParam
...
-                                    if (variable != null) {
-                                        if (downloadInstance.getExtraArguments() == null) {
-                                            downloadInstance.setExtraArguments(new ArrayList<Object>());
+            logger.fine("is download service supported? key=" + key + ", value=" + value);
+            // The loop goes through all query params (e.g. including key, gbrecs, persistentId, etc. )
+            // So we need to identify when a service is being called and then let checkIfServiceSupportedAndSetConverter see if the required one exists
+            if (key.equals("imageThumb") || key.equals("format") || key.equals("variables")) {
+                serviceRequested = true;
+                //Only need to check if this key is associated with a service
+                if (downloadInstance.checkIfServiceSupportedAndSetConverter(key, value)) {
```
qqmyers: because this used to run for any params like :persistentId, gbrecs, key it wasn't possible to just put an exception in an else clause (as was commented out in the previous code.

Issue #4717, opened 2018-05-25
Code Cleanup - better handling of external links


pameyer:
Splitting out from https://github.com/IQSS/dataverse/issues/3337#issuecomment-391737405 :

Short version - it would probably be a good idea to add `rel="noopener"` to external links, in addition to `target="blank"`.

Another thought that's almost certainly out of scope for this issue, but related to the discussion of consistency with handling external links, would be adding rel="noopener" to external links (recommended by lighthouse for performance / security - https://developers.google.com/web/tools/lighthouse/audits/noopener).

mheppler:
As @pdurbin noted in his PR comment, we have started to address this issue in _adds rel="noopener" attribute to the link formatting in citation.tsv_ #6249. I will be diligent when reviewing PR that include UI changes which will provide an opportunity to add these `rel="noopener"` attribures.

I will start a checklist of the places it has been address here, so that if/when this issue gets picked up, we know where this is already done.

- [x] citation.tsv (#6249)

PR #6249, opened 2019-10-03
adds rel="noopener" attribute to the link formatting in citation.tsv

landreev:
This was discussed at the standup meeting this AM.
This adds one more attribute to the html formatting for external urls that we provide in ```citation.tsv```:

```<a href="#VALUE" target="_blank" rel="noopener">#VALUE</a>```

this is considered a better style/the recommended way, vs. just ```target="_blank"```.

(we've been meaning to add this for a while; and since we are recommending the installations to refresh the citation block in 4.17, could be a good time to do it now)


pdurbin:
Yeah, looks good. This appears to be a partial solution for #4717.

PR #6273, opened 2019-10-11
Update org.apache.commons:commons-compress

djbrooke:
Update of this helpful dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -552,7 +552,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-compress</artifactId>
-            <version>1.18</version>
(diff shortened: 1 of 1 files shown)


kcondon:
@djbrooke @pdurbin Do you know what the apache commons lib is used for? Might be good just to smoke test that area.

djbrooke:
Let's move forward as is right now.

qqmyers:
FWIW - compress is used directly by the BagIt/archiving code, so the dependency is real and independent of the use in POI. 

Also - I've updated to 1.19 locally and can confirm the bag code works with it. (Have been having issues with other old libs interfering with the google cloud api we use to send bags to that cloud, but that's a separate issue...)

PR #6285, opened 2019-10-17
sleep after create and before publish of dv #6270

### LINKED ISSUE #6270: SearchIT.testNestedSubtree failing on jenkins.dataverse.org
From https://jenkins.dataverse.org/job/IQSS-dataverse-develop/214/testReport/

Screen Shot 2019-10-11 at 1 41 38 PM

This failure was reported on 8991919 but @donsizemore and I have been seeing it for several days if not weeks.

pdurbin:
closes #6270


scolapasta:
If after create the index is not being updated correctly (i.e. it doesn't until after publish) isn't that a bug we want to fix? I guess I don't understand from the comment exactly what's happening here (i.e. why would the index for create be queued up after the index from publish (another command) ever?)

I'm hesitant that this change hides a possible real bug just to make the tests pass.

PR #6314, opened 2019-10-24
improve API blocking documentation #6313

### LINKED ISSUE #6313: API: Blocked-key does not seem to work.
According to our guides, one can block api endpoints, then allow remote access by using a key identified by the setting :BlockedApiKey and accessed by passing its value by adding an argument, unblock-key=theKeyYouChose . When troubleshooting a recent user issue, this appeared not to work when using the default localhost/admin/test policy along with the example key setting provided in the guide:

http://guides.dataverse.org/en/4.17/installation/config.html#blockedapikey

pdurbin:
closes #6313

@kcondon for now, as we discussed, all I did was add links between the various information scattered across the config page. I'm happy to add or update content as well. Please let me know.

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -19,14 +19,16 @@ Changing the Superuser Password
 
 The default password for the "dataverseAdmin" superuser account is "admin", as mentioned in the :doc:`installation-main` section, and you should change it, of course.
 
+.. _blocking-api-endpoints:
+
 Blocking API Endpoints
 ++++++++++++++++++++++
 
 The :doc:`/api/native-api` contains a useful but potentially dangerous API endpoint called "admin" that allows you to change system settings, make ordinary users into superusers, and more. The ``builtin-users`` endpoint lets people create a local/builtin user account if they know the ``BuiltinUsers.KEY`` value described below.
 
-By default, all APIs can be operated on remotely and a number of endpoints do not require authentication. https://github.com/IQSS/dataverse/issues/1886 was opened to explore changing these defaults, but until then it is very important to block both the "admin" endpoint (and at least consider blocking ``builtin-users``). For details please see also the section on ``:BlockedApiPolicy`` below.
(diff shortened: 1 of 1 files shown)


kcondon:
@pdurbin Thanks, I thought we'd also talked about adding/modifying a sentence or potentially adding an example to BlockedApiKey in case someone, like me, goes there directly without seeing the chain of info from the intro to policy to key. Like: The blocked api key is used in conjunction with BlockedAPIPolicy and will not be enabled unless the policy is set to unblocked-key.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -19,14 +19,16 @@ Changing the Superuser Password
...
 
-By default, all APIs can be operated on remotely and a number of endpoints do not require authentication. https://github.com/IQSS/dataverse/issues/1886 was opened to explore changing these defaults, but until then it is very important to block both the "admin" endpoint (and at least consider blocking ``builtin-users``). For details please see also the section on ``:BlockedApiPolicy`` below.
+By default, all APIs can be operated on remotely and a number of endpoints do not require authentication. https://github.com/IQSS/dataverse/issues/1886 was opened to explore changing these defaults, but until then it is very important to block both the "admin" endpoint (and at least consider blocking ``builtin-users``). For details please see also the section on :ref:`:BlockedApiPolicy` below.
 
-It's also possible to prevent file uploads via API by adjusting the ``:UploadMethods`` database setting.
+It's also possible to prevent file uploads via API by adjusting the :ref:`:UploadMethods` database setting.
```
poikilotherm: It is very easy to lock yourself out if you don't use the commands in the correct order.
Could you add the full example here, so this is not left as an excersice to the admin?
You can copycat from https://github.com/IQSS/dataverse-kubernetes/blob/6f83afbeeb249542e87b9d04ae504d2ff319abfe/docker/dataverse-k8s/bin/bootstrap-job.sh#L60-L62 :wink:

pdurbin: @poikilotherm good catch. Fixed in bc08e28

pdurbin:
@kcondon as we've been discussing, I added a couple more commits to this pull request. bc08e28 is the latest and here is the HTML version of how it looks:

Screen Shot 2019-10-24 at 2 38 58 PM

...

Screen Shot 2019-10-24 at 2 39 28 PM

Issue #6322, opened 2019-10-29
Fix API Test Suite


djbrooke:
Our handy indicator at the bottom of the readme indicates that we have some test failures. We should fix them before releasing 4.18. More info here:

https://jenkins.dataverse.org/job/IQSS-dataverse-develop/lastCompletedBuild/testReport/

pdurbin:
My pull request is in at #6324 and I'm taking myself off this unless there is more swarming to do.

pdurbin:
Three new-ish API test classes were missing from phoenix so I just added them:

- EditDDIIT
- ExternalToolsIT
- AccessIT

PR #6271, opened 2019-10-11
6214 user notification following ingest

alejandratenorio:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Note that we use the "closes" syntax below to trigger Github's automation to close the corresponding issue once the pull request is merged.

Thanks for your contribution to Dataverse!

## Related Issues

- closes #6214

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #6324, opened 2019-10-29
ingest now completes, fix assertions #6322

pdurbin:
partial solution for #6322

Here's how the output looks now, and I changed the assertions to match:

## assert from line 337 

```
{
    "status": "OK",
    "data": {
        "notifications": [
            {
                "id": 818,
                "type": "SUBMITTEDDS"
            },
            {
                "id": 816,
                "type": "INGESTCOMPLETED"
            },
            {
                "id": 815,
                "type": "SUBMITTEDDS"
            },
            {
                "id": 812,
                "type": "CREATEACC"
            }
        ]
    }
}
```

## assert from line 389

```
{
    "status": "OK",
    "data": {
        "notifications": [
            {
                "id": 827,
                "type": "SUBMITTEDDS"
            },
            {
                "id": 825,
                "type": "SUBMITTEDDS"
            },
            {
                "id": 823,
                "type": "INGESTCOMPLETED"
            },
            {
                "id": 822,
                "type": "SUBMITTEDDS"
            },
            {
                "id": 819,
                "type": "CREATEACC"
            }
        ]
    }

```

I suppose this means that ingest is now succeeding while previously it was not. I'm not sure why. Here's the transition on phoenix where it flipped from passing to failing:

- passing: https://build.hmdc.harvard.edu:8443/job/phoenix.dataverse.org-apitest-develop/437/testReport/
- failing: https://build.hmdc.harvard.edu:8443/job/phoenix.dataverse.org-apitest-develop/438/testReport/

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/InReviewWorkflowIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/InReviewWorkflowIT.java
@@ -338,10 +338,11 @@ public class InReviewWorkflowIT {
                 // TODO: Test this issue from the UI as well: https://github.com/IQSS/dataverse/issues/2526
                 .body("data.notifications[0].type", equalTo("SUBMITTEDDS"))
                 //.body("data.notifications[0].reasonsForReturn[0].message", equalTo("You forgot to upload any files."))
-                .body("data.notifications[1].type", equalTo("SUBMITTEDDS"))
+                .body("data.notifications[1].type", equalTo("INGESTCOMPLETED"))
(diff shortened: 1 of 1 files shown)


landreev:
> I suppose this means that ingest is now succeeding while previously it was not.

My understanding is that the ingest was always succeeding - it's just that we now have a new ingest notification, that we didn't have before. (Was introduced in the PR merged last week - so this explains why the test started failing last week). 

But the fix in this PR - adjust the expected notifications to assert - seems like the right thing to do.

pdurbin:
> it's just that we now have a new ingest notification

Right, right, from pull request #6271. Thanks, @landreev

PR #6333, opened 2019-10-31
Search API no longer limited to published data #6301

### LINKED ISSUE #6301: View unpublished data through search API
**Given** that I am searching for information in Dataverse
**When** I perform a search through the **Search native API**
**Then** I would like to have the ability to see `Published`, `Unpublished` or `both` information

**What is currently happening:**
I am only able to search for published information

**What we would love to have**
A more flexible approach for search that allows us to search for any kind of data, having `only published` by default and then having a potential flag that can be set on the request

Thanks a lot! 🚀

pdurbin:
closes #6301


pdurbin:
@sekmiller good idea. Implemented in eb861c7

pdurbin:
@sekmiller good catch. Fixed in 7ebcc3f

Issue #6274, opened 2019-10-15
As a dataverse administrator, I want to remove my name from contributors list of a dataset version


tcoupin:
All editors of a dataset appears in contributors list of the published dataset version.

On the following snapshot, only the scientist should be displayed (Christian DEPRAETERE) et not the administrators who make only small adjustements.
![image](https://user-images.githubusercontent.com/7457386/66828870-5f7e9a00-ef52-11e9-98d6-ead6ec138e22.png)

tcoupin:
A temporary workaround can be to deletes row from datasetversionuser table. What do tou think about that ?

djbrooke:
Hi @tcoupin, in #5541 there were changes that allowed updates to be made by superadmins without making a version change. This was included with 4.12. I'd suggest using this feature instead of removing records that were created. 

I acknowledge that the use case that you present is slightly different, but I'm concerned about providing extended functionality in this area because each additional option for editing without a record reflected in the versions view takes away from what the versioning system was meant to support in the first place. Suggestions/discussion welcome!

tcoupin:
Hey @djbrooke,
I use this feature but my name is in the contributors list. Since this feature is for superusers only, we can remove names when publishing whitout version increase.

djbrooke:
Hi @tcoupin, I'm not sure I understand. Are you saying that you are able to use the feature to publish without incrementing the version, but your name still appears on the contributor list?

tcoupin:
Yes. I test on a new dataset with 2 accounts, one as simple user and one as superuser.

djbrooke:
Thanks @tcoupin, I just tested this in this way:

- Added https://demo.dataverse.org/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2FSM6V9Z to demo as my Danny Brooke account
- Published
- Logged out
- Logged in with a superuser account
- Made a metadata edit
- Published with the "Change Current Version" option selected

A new version was not created (as expected) and I did see the superuser account added to the contributors list as you mention. 

@qqmyers and @adam3smith, do you have any thoughts on this? I suppose I was expecting for the superuser account to not be added to the contributors list for the current version, but thinking about it more, this is correct and makes sense. Just wondering if it came up as you were developing #4760/#5541.

scolapasta:
I can see both sides of this. A change *was* made to that version, so we should track it as if it had happened when in draft. But I also see the idea of having a superuser make changes for the contributor and not needing to show up as a contributor.

[...]

PR #6340, opened 2019-11-04
Rename 'contributors' by 'edited by' in dataset version table.

tcoupin:
## Related Issues

- #6274 : As a dataverse administrator, I want to remove my name from contributors list of a dataset version


pdurbin:
I'm sure people can imagine what this change looks like but here's a screenshot from http://ec2-18-234-42-196.compute-1.amazonaws.com/dataset.xhtml?persistentId=doi:10.5072/FK2/C3SJVC

Screenshot from 2019-11-04 06-17-14

djbrooke:
Hi @tcoupin, I'm going to close this PR without review. As I mentioned in #6274, we need to better understand this whole workflow before making individual small changes and we don't have the design resources to prioritize this larger effort right now. If you want to provide a link to the one line code change in the issue that's fine, but we won't merge anything right now.

PR #6344, opened 2019-11-05
explain how developers can run the API test suite on EC2

pdurbin:
Screen Shot 2019-11-05 at 2 53 48 PM


djbrooke:
I assigned myself and @TaniaSchlatter because we are both interested in spinning up branches to play with code changes that are working their way through the development process. I thought this was that. This is not that, so I'm removing us from the issue. We'll schedule some time in the future to go through the existing docs.

pdurbin:
@djbrooke for this to be that you would skip this step:

-  Edit ``ec2config.yaml`` to change ``test_suite`` to ``true``.

djbrooke:
@pdurbin this isn't what I'm looking to make easier for me and others. What I'm looking to make easier is everything in the "deployment" section. This PR is fine to move along, but it's not what I'm looking for.

PR #6347, opened 2019-11-06
#6346 disable solr service account after installation

### LINKED ISSUE #6346: docs: disable solr service account shell
in http://guides.dataverse.org/en/latest/installation/prerequisites.html#installing-solr we tell installationators to create a service account for solr. in http://guides.dataverse.org/en/latest/installation/prerequisites.html#securing-solr we tell them to firewall solr off from the Big Scary Internet. we should also tell them to set the service account's shell to /sbin/nologin:

`$ sudo usermod -s /sbin/nologin solr`

Pull request forthcoming.

donsizemore:
Closes #6346


djbrooke:
Thanks @donsizemore for the PR and the discussion about this!

PR #6378, opened 2019-11-16
MakeDataCount only retrieves 25 citations

### LINKED ISSUE #6138: MakeDataCount only retrieves 25 citations
The default response to https://api.datacite.org/events?doi=<Dataset DOI>&source=crossref as called in MakeDataCountApi.updateCitationsForDataset only returns 25 responses because it paginates.

(Try curl https://api.datacite.org/events?doi=10.33564/FK2LJNVAG&source=crossref to see this.)

It looks like you can add &page[size]=1000 (1000 is the max) which may be sufficient for most datasets, but it looks like you have to handle pagination to get the max they'll send (10K entries). Info at https://support.datacite.org/docs/pagination.

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Note that we use the "closes" syntax below to trigger Github's automation to close the corresponding issue once the pull request is merged.

Thanks for your contribution to Dataverse!

## Related Issues

- closes #6138 MakeDataCount only retrieves 25 citations

## Pull Request Checklist


mheppler:
Took a look at reviewing to get this moving to QA, but the changes are much more complex than anticipated, so I am requesting a review from the original developer, @pdurbin.

qqmyers:
@pdurbin - Hmm. At QDR, I tested prior to the other PR on filtering. Since QDR submits is-part-of /has-part relationships, I could see that, for a dataset with 1000 files, this PR was returning 2000 items - demonstrating it got the first page of 1000 (versus 25) and then retrieved the second and third (with a few other events on the third page of results). 
It would be a partial test if there are any datasets with more than 25 citations. One option would be to revert the findDatasetOrDie() call to allow you to request citations for a QDR dataset (test or real) from the test Dataverse. Another might be to add a bunch of relationships manually at Datacite for a given test dataset. For either of those doi:10.33564/FK2LJNVAG might be useful - it should have 2K+ relationships. It's registered with the test server which you can ping with curl 'https://api.test.datacite.org/events?doi=10.33564/FK2LJNVAG&source=                                                                          crossref&page[number]=2&page[size]=1000' (same call that Dataverse is making internally).  curl 'https://api.test.datacite.org/dois/10.33564/FK2LJNVAG' will get you json that I think you could cut/paste from to add is-part-of relationships to some local test dataset in the DataCite Fabrica GUI. (I may end up trying this to add different types of relationships to test 'cites' with test datasets - haven't done it yet to know that it works.)

kcondon:
Need a better way to test, couldn't spoof the local dv to access that test dataset and not sure how to access/use Fabrica. Passing back to Dev for potential follow up.

pdurbin:
@qqmyers if we wanted the following URL to work:

http://localhost:8080/api/datasets/:persistentId/locks?persistentId=doi:10.33564/FK2LJNVAG

I know we'd need to hack on the dvobject table to swap in "10.33564" and "FK2LJNVAG" on one of the datasets but do you recall what other settings we'll need to change? Stuff about "shoulder" or whatever?

qqmyers:
@pdurbin Since you're not trying to mint anything, I think you can just do what you say. The only thing you're trying to trick is the findDatasetOrDie call which will just bounce off the dvobjects table authority and identifier (AFAIK).

kcondon:
@qqmyers @pdurbin So that is what I did, even changed the authority and shoulder to match. Even, tried setting FAKE, creating the target DOI, publish, then switch back to DataCite. Dataverse now thinks it has that dataset but when calling the fetch url, it returns OK but 0 citations.

qqmyers:
@kcondon - sorry - possibly a missed step - have you set doi.mdcbaseurlstring to api.test.datacite.org and restarted glassfish? If not, it will default to using the production datacite site and find nothing. That's generally required for any testing with MDC and Datacite with test DOIs, not specific to this pull, but I don't think it's been mentioned here.

kcondon:
@qqmyers have this: 
```
       <jvm-options>-Ddoi.baseurlstring=https://mds.test.datacite.org</jvm-options>
 ```

qqmyers:
`<jvm-options>-Ddoi.baseurlstring=https://mds.test.datacite.org</jvm-options>` is just for minting/publishing. The new doi.mdcbaseurlstring that has to point to api.* instead of mds is for the MDC uploads/downloads. (we decided on a separate jvm-option rather than changing to one with the common domain to use in the earlier PR).
In any case I think MDC for test needs:
`<jvm-options>-Ddoi.mdcbaseurlstring=https://api.test.datacite.org</jvm-options>`

If it's not working with that, I'll check the code again - my testing is basically with this and the filtering PR together so there could be a typo when I tried to pull them apart for the PRs.

qqmyers:
and the call you're making is
` curl -X POST 'http://localhost:8080/api/admin/makeDataCount/:persistentId/updateCitationsForDataset?persistentId=doi:10.33564/FK2LJNVAG'` ?

Does 
`curl 'https://api.test.datacite.org/events?doi=10.33564/FK2LJNVAG&source=crossref&page[number]=2&page[size]=1000'` show results? (It takes several seconds for me, but I see 1K entries).

kcondon:
@qqmyers Yes, I'm making that call.

The other results in:
curl "https://api.test.datacite.org/events?doi=10.33564/FK2LJNVAG&source=crossref&page[number]=2&page[size]=1000"
curl: (3) [globbing] bad range in column 86

qqmyers:
Hmm - I'm at a loss as to how that should not give any error yet have 0 counts. The parsing code does try to findByGlobalId, but I don't see that that tests the global authority/shoulder settings, so I think it should find the hacked entries in the dvobject table.

There is a little logger.fine() logging you could turn on. Beyond that, I guess I can try to recreate this more exactly - make sure my code here is exactly as in the PR and then pick a test DOI from a different server, etc.

kcondon:
@qqmyers I have one last thing to try, which may have been user error.

qqmyers:
@kcondon - updated. Hope you found the issue. I'm going to head off for a while...

kcondon:
@qqmyers It works! Was user error, got a bit confused between test/prod and doi.baseurlstring and doi.mdcbaseurlstring. Sorry for causing you to go on a goose chase.

PR #6379, opened 2019-11-16
MakeDataCount counts all relationships as citations

### LINKED ISSUE #6139: MakeDataCount counts all relationships as citations
As written, I think the MakeDataCountApi citation counting will also report the "unique-resolutions-machine" and similar relationships between a dataset and the report sent in to register its views and downloads. Those events have the same structure and the code only checks subj-id and obj-id and doesn't filter on "source-id" or "relation-type-id". 

The example URL in the code (curl https://api.datacite.org/events?doi=10.7910/dvn/hqzoob&source=crossref) shows the type of view/download events that I think will get picked up. 

I have not confirmed this specifically, but instead noted that this code does pick up the is-part-of relationships between files and a dataset that we've added for QDR ( #2778 is related) and those are reported as citations. That lead me to inspect the code and, unless I missed something, I think it will count the reported views/downloads as well.

Minimally, I think it should filter those out, but there's a potentially larger question of whether all of the relationships DataCite will report should count as citations. That may be a MakeDataCount question rather than for Dataverse alone. (My guess is that there are few systems that give DOIs to datasets and files as Dataverse can, and few if any of those actually reporting the ispartof/haspart relationship to DataCite as we've started to do in QDR and as is planned in #2778).

qqmyers:
This PR implements a whitelist of relationships to count. Given the large number of relationships from multiple sources (see the links in the issue), with most of those being structural, I thought a whitelist would be the better approach. Currently that includes 4 relationships - cites/references and their inverses.

## Related Issues

- closes #6139 MakeDataCount counts all relationships as citations

## Pull Request Checklist


djbrooke:
@qqmyers - I moved this back to community dev since there was some discussion on the community call about whether or not to make this configurable. Let us know when it's ready for another look. Thank you!

qqmyers:
OK - to get to something that can hopefully go forward now as a good incremental step, I've made the relationship checks directional as discussed above (e.g. counting dataset is cited-by <uri> or <uri> cites dataset but not the opposite directions) and added the supplements/is-supplemented-by pair.

That should cover existing cases and let us get a version out that filters the view/download reports which is a problem for all current MDC-enabled versions.

landreev (approved):
(The PR appears to be straightforward  - I only kept it in CR because I was educating myself on how external citations work, and how our usage metrics are reported back etc. (But also because we were busy finishing - and then fixing - 5.1 anyway).

PR #6399, opened 2019-11-22
do not hardcode https for SWORD #6348

### LINKED ISSUE #6348: Do not hardcode https for SWORD
Hardcoding https makes non-production (test installations, development, etc.) use of Dataverse harder. Spun off from #6296, which has more details.

pdurbin:
closes #6348


pdurbin:
@kcondon how to test:

- Make sure SWORD works fine, in general. It's a good sign that the API test suite is passing according to https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/view/change-requests/job/PR-6399/
- Compare **responses** (XML output) from SWORD curl commands in "develop" vs this pull request. Since 4.0, the URLs in the responses are always `https`. Now they can be configured to be `http` (no s) based on the `dataverse.siteUrl` JVM option.

Issue #6226, opened 2019-09-27
Revise PR Template


djbrooke:
In https://github.com/IQSS/dataverse/pull/6223#discussion_r329180952 @pdurbin mentioned that he just deletes most of the PR template (https://github.com/IQSS/dataverse/blob/develop/PULL_REQUEST_TEMPLATE.md). I do the same. 

Let's think about this, get some opinions (from other people or other projects) and change what's in the template.

poikilotherm:
As discussed with @pdurbin [on IRC today](http://irclog.iq.harvard.edu/dataverse/2019-11-21#i_113036), more information in PRs would be usefull when reaching the QA column. @pdurbin, me and (as far as I understood it) most certainly @djbrooke and @kcondon, too, would like to communicate more about what things changed, what needs to be tested, etc.

It would be very helpfull to include a good and easy pattern (needs to be first timers compatible) on how to most efficiently draw a picture for reviewers, testers etc.

The K8s people have some nice ideas for this, maybe it's an inspiration: 
https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md

(These `/kind` thingies have bots attaching labels to the PR, which is not doable for normal contributors)

poikilotherm:
In addendum: it might be a good idea to talk about where to put updates that happen after creating the PR in code review. 

- Should QA browse through all of this? 
- Should the author provide an enhanced description before it moves from review to QA? In small PRs this is no problem, but what about longer reviews with lots of chatter? 
- When using Waffle, the policy was "discuss in the issue". Nowadays that changed to "use the code review functions of GitHub" (which changed for good), but does it help QA?

djbrooke:
Thanks @poikilotherm for the thoughts and for the links to the k8s example. I like to encourage face to face discussion whenever possible but that doesn't mean we can't to better here. I looked around at a few other projects (Canvas, EdX) and the k8s example was better. Just needs to add some QA-specific info. This is what I'm going to do:

- Create a proposed PR template based on k8s (will link branch here), add some proposed thoughts for QA as well
- Ask @scolapasta and the developers to discuss the proposed template during tech hours and make any changes.

From there, we'll get it out to the larger community for comment.

Issue #5974, opened 2019-06-26
As a Dataverse admin, I would like to hook up my installation to an IDM/IAM


poikilotherm:
After some recent chat on IRC with @pdurbin [here](http://irclog.iq.harvard.edu/dataverse/2019-06-26#i_98796) and [here](http://irclog.iq.harvard.edu/dataverse/2019-06-25#i_98437), I am opening this issue quickly, as @pdurbin pointed out [slide 32](https://mfr.osf.io/render?url=https://osf.io/8yh3e/?action=download%26mode=render#page=32) by @djbrooke regarding future changes on the auth flow.

## Motivation
We have had a discussion at FZJ about what auth methods we will allow and the result was that we will need to provide multiple options, including reusing the already existing accounts from AD. We are part of DFN AAI/eduGain, so there is a SAML IdP present, too.

While working on IQSS/dataverse-kubernetes, I tried to integrate the SAML IdP in IQSS/dataverse-kubernetes#14 using mod_auth_mellon instead of Shibboleth. Shibboleth as an SP IMHO is a nightmare. You can only use it with mod_shib, have to run it on the same host/container as Apache does and for usage with Dataverse, you are tied to using mod_ajp. As Dataverse is very tightly integrated with Shibboleth specials (reading sessions, etc), I had no luck with mod_auth_mellon... :cry: 

## Proposal
All of this is not very container friendly, which is why I would like to propose integrating a "new" option: usage of an [IDM/IAM](https://en.wikipedia.org/wiki/Identity_management).

Most existing solutions are capable of OAuth2 and/or OpenID Connect. At least OAuth2 support is already present in Dataverse.

Steps to be taken:
1. Use [current OAuth2 lib](https://github.com/scribejava/scribejava) to read OpenID Connect data:
    - [ ] Provide integration tests for Github, Google and ORCID to ensure nothing breaks (using stubs) See #5985 
    - [x] Update to latest version (6.6.3) from v3.1.0 (See #5991)
    - [X] Write new AuthenticationProvider for interaction with a custom endpoint (see #6432)
    - [X] Receive and parse user data following Open ID connect standard (see #6432)
    - [X] Provide documenation on usage and technical details (see #6432)
    - Stick with current "flow" based on first login page, etc for this first step.
    - This might be referred to as "OpenID Connect Core".
2. Create automated testing with Unity IDM provided by Helmholz Data Federation
3. Make use of email adress verification status claim available from OIDC
[...]

PR #6433, opened 2019-12-04
6432 - implement a first simple

### LINKED ISSUE #6432: Implement first (simple/PoC) version of a OIDC auth provider
This is part of epic #5974.

----

Since the refactoring of OAuth2 infrastructure in #5991 and #6364 it was made possible to introduce a new authentication mechanism using the [Nimbus OAuth 2.0 SDK with OpenID Connect extensions](https://connect2id.com/products/nimbus-oauth-openid-connect-sdk).
(ScribeJava has a long missing feature request for OIDC support, which has never been attacked).

This first implementation should enable us:
* load a configuration
* retrieve metadata
* login at the provider
* retrieve standardized claims like email, name, ...
* allow multiple providers so it can be reused for different IdPs without touching codebase

poikilotherm:
This pull request adds very basic support for authentication with an OpenID Connect capable identity provider.

We need this is as a basis to solve #5974, which is about using a more sophisticated user authentication system for Dataverse.

- Closes #6432

This has been tested so far with an internal IdP based on Unity IDM.
I think about introducing integration testing this feature by using [Testcontainers](https://www.testcontainers.org), but this seems like beyond scope for this PR and keeping that narrow.

We should test this with Google, maybe with ORCID.

**Does this PR introduce a user-facing change?**:
Kind of. It reuses all the existing OAuth2 UI stuff, but of course enables new buttons for providers. It might become an issue in case someone really wants to add a big bunch of stuff, as the list grows in that case.

It might affects the order of auth providers, too, as the sorting feature for the UI needed to be refactored.

**Is there a release notes update needed for this change?**
Sure. But maybe this should be a separate PR, as there are some more things still to come.

**Additional documentation**:
This PR is a demo for #6226 on @pdurbins [demand on IRC](http://irclog.iq.harvard.edu/dataverse/2019-12-04#i_113824)


src/main/java/edu/harvard/iq/dataverse/LoginPage.java
```diff
@@ -129,16 +125,20 @@ public void init() {
...
-                    infos.add(authenticationProvider.getInfo());
-                }
+        List<AuthenticationProvider> idps = new ArrayList<>(authSvc.getAuthenticationProviders());
+        
+        // sort by order first. in case of same order values, be deterministic in UI and sort by id, too.
+        Collections.sort(idps, Comparator.comparing(AuthenticationProvider::getOrder).thenComparing(AuthenticationProvider::getId));
+        
+        for (AuthenticationProvider idp : idps) {
+            if (idp != null) {
+                infos.add(idp.getInfo());
```
pdurbin: Is the change above necessary for OIDC to work? I don't think so. It does look like a nice cleanup though. I like how it's deterministic. Perhaps this change should be documented.

src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/AbstractOAuth2AuthenticationProvider.java
```diff
@@ -207,6 +214,14 @@ public AuthenticationProviderDisplayInfo getInfo() {
         return new AuthenticationProviderDisplayInfo(getId(), getTitle(), getSubTitle());
     }
     
+    /**
+     * Used in {@link LoginPage#listAuthenticationProviders()} for sorting the providers in the UI
+     * TODO: this might be extended to use a value set by the admin when configuring the provider via JSON.
+     * @return an integer value (sort ascending)
+     */
+    @Override
+    public int getOrder() { return 100; }
```
pdurbin: What if there are more than 100 authentication providers? Some have the same number? Is it ok to have two 42s?

src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/oidc/OIDCAuthProvider.java
```diff
@@ -0,0 +1,263 @@
+package edu.harvard.iq.dataverse.authorization.providers.oauth2.oidc;
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.nimbusds.oauth2.sdk.*;
```
pdurbin: In general, we try not to use star imports. I regret to say that we don't mention this at http://guides.dataverse.org/en/4.18.1/developers/coding-style.html 😄

pdurbin:
> **Suggestions on how to test this**:
> We should test this with Google, maybe with ORCID.

@poikilotherm can you also provide an OIDC environment for testing the actual feature this pull request introduces?

src/main/java/edu/harvard/iq/dataverse/LoginPage.java
poikilotherm: Actually yes it is. From this function, `authSvc.getAuthenticationProviderIdsSorted()` was called, which is deleted below. This was done in 3fd326c23 and here's what I wrote in the commit log:

> Refactor LoginPage.listAuthenticationProviders() to represent actual provider configuration.
>
> Before this refactoring, a static list of providers has been used, no matter if they where
enabled or not. This is not acceptable for the new OIDC provider, as we might have multiple
of 'em and it retrieves metadata on creation.
>
> So the idea was to not create some static list of IDs (which had been used to look up the real
provider from the "registry"), but instead get the registred providers, sort them and
retrieve the display info.
>
> Sorting is done by a numerical value first, then by id attribute (which is set by factories),
so we have a deterministic, unchanging order for the user.
>
> This lead to the creation of a new method AuthenticationProvider.getOrder(), defaulting to 1
and overridden by implementations. This opens up the possibility to change ordering by
configuration (as suggested in an old TODO comment of the refactored code)

src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/AbstractOAuth2AuthenticationProvider.java
poikilotherm: Actually it doesn't matter if some have the same number, as they are sorted by number first, then by name.

This is mostly about sorting Builtin first (order = 1), Shib second (order=20) and OAuth last (all order = 100), as it has been sorted before my refactoring.

Once we have a configurable value for this, you could also realize sth. like provider A first, then all the others.
[Please see also my comment above.](https://github.com/IQSS/dataverse/pull/6433/files/511827429f280ee7ad997fa68fc85694a556640d?file-filters%5B%5D=.java&file-filters%5B%5D=.properties#r354178859)

src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/oidc/OIDCAuthProvider.java
poikilotherm: I'm sorry, IntelliJ IDEA does this by default.
Should I mention checkstyle again to catch this sort of stuff?
We could create a Jenkins job for it...

poikilotherm: Oh and should I change this in a new PR as this one is already merged?

poikilotherm:
I can report great success for using Google OpenID Connect support. Worked flawless, does what it should do. I love standards... :smile: 

Should we talk to @alejandratenorio and @Gerafp for testing it with Microsoft Azure AD?

PR #6422, opened 2019-11-26
Can't preview in page for files in prior public dataset versions

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Note that we use the "closes" syntax below to trigger Github's automation to close the corresponding issue once the pull request is merged.

Thanks for your contribution to Dataverse!

## Related Issues

- closes #6421 Can't preview in page for files in prior public dataset versions

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

Issue #6421, opened 2019-11-26
Can't preview in page for files in prior public dataset versions


qqmyers:
Related to #6419 (and I asked about this there but am creating a separate issue), if there is more than 1 published version of a dataset, files in the earlier versions can't be previewed in page. This means that files that have been deleted can't be previewed at all and current files can only be previewed if they are accessed via the latest published version. I'm not sure if this was intended or is needed (these are still publicly accessible files). 
The related PR resolves this and still keeps all of the other limits to avoid previews of files that have restrictions/conditions/are in draft datasets/are in datasets with access terms...

scolapasta:
@qqmyers I can't remember if there was a specific issue in allowing previewers on older versions or if it was just part of the "previewers on current versions adds value; let's get that in while we consider older versions".

Since this only adds the previewers for public files still, let's test this and merge.

PR #6420, opened 2019-11-26
compare versus correct version

qqmyers:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Note that we use the "closes" syntax below to trigger Github's automation to close the corresponding issue once the pull request is merged.

Thanks for your contribution to Dataverse!

## Related Issues

- closes #6419 File Previews don't show in page if a draft version exists

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

Issue #6419, opened 2019-11-26
File Previews don't show in page if a draft version exists


qqmyers:
When viewing an unrestricted file in a published dataset, if a draft dataset exists, the preview tab does not display. And, if that is fixed, the previewer tries to load the draft dataset version and fails.

Digging in, it looks like there are two places where getLastestFilemetadata is getting called (directly or indirectly) and leading to these failures. In the FilePage, the init() method correctly picks the filemetadata for the requested version, but the preview() method was using file.getFileMetadata() which gets the draft version. (The second issue above).
In file.xhtml, there is a render check requiring the FilePage.filemetadata.datasetversion (e.g. 1.0) to be equal to the latestversion of the dataset (e.g. draft in this case).

I'll submit a PR with updates for both which work for me to make the preview tab show for this case. (Question - was the design to not allow prior versions to be previewed? I.e. is checking that the datasetversion is the latest public version what was intended?).

pdurbin:
@qqmyers sounds find to me but just a heads up that we went back and forth with this notion of having the Preview tab with non-public files or not in pull request #6325 . See c35268b and 29a7edc especially and https://github.com/IQSS/dataverse/pull/6325#issuecomment-548575376

qqmyers:
@pdurbin - this is not about showing non-public files. The presence of a draft version makes the published version un-previewable.

pdurbin:
Oh dear. Yes, a pull request would be much appreciated! 🎉

qqmyers:
@pdurbin - do you know the answer to my question above? I think the version test in file.xhtml (after my fix for the draft issue) will still stop you from previewing a file in v1.0 if a later published version exists (e.g. v2.0).

pdurbin:
@qqmyers sorry, I don't know. And I forget what `getLatestVersionForCopy` in your pull request does. I'm hoping @sekmiller or @scolapasta will take a look. 😄 

At a high level, my understanding is that for the Preview tab we are reusing some logic from when we show "Download URL" like this (no guestbook, no restrictions, etc.):

![Screen Shot 2019-11-26 at 4 04 25 PM](https://user-images.githubusercontent.com/21006/69672626-7d1d4400-1066-11ea-8496-a94282082baf.png)

But I could be off or the thinking could have changed. I'm not sure. So I should probably stop making noise in here. 😄

qqmyers:
getLatestedVersionForCopy just returns the latest version that is released or archived, so it doesn't pick up a draft version.

PR #6438, opened 2019-12-05
Cleaned up preview tab on file pg [ref #6437]

### LINKED ISSUE #6437: File Preview - UI clean up
There was some UI clean up that was missing in #3758. After reviewing with @TaniaSchlatter the to-do list is as follows:

- "Previews" btn text label should be "Preview Tools"
- White btn background for "access" is not appropriate on the Preview btn, it should be default gray
- Dropdown menu is aligned right, when it should be aligned left
- "Open {Preview Tool Name}" btn is small, not styled with Bootstrap, and breaks to a second line
- "Open {Preview Tool Name}" btn text label should be "Explore on {Preview Tool Name}", and needs an explore icon
- Preview tool iframe border needs styling, [responsive Bootstrap classes](https://getbootstrap.com/docs/3.4/components/#responsive-embed)

Screen Shot 2019-11-18 at 11 29 54 AM

mheppler:

The new preview tab on the file pg needed some UI clean up.

Closes #6437 _File Preview - UI clean up_

Thank you for reviewing.

Single tool, multiple tools, with and without guestbook/terms of access

Sure does.

No.

**Additional documentation**:

N/A

### DIFF
--- a/src/main/java/propertyFiles/Bundle.properties
+++ b/src/main/java/propertyFiles/Bundle.properties
@@ -1498,7 +1498,6 @@ file.rsyncUpload.rsyncUploadDisabledDueFileUploadedViaHttpAndPublished=Upload wi
 
 file.metaData.dataFile.dataTab.variables=Variables
 file.metaData.dataFile.dataTab.observations=Observations
-file.metaData.viewOnWorldMap=Explore on WorldMap
 file.addDescription=Add file description...
 file.tags=Tags
 file.editTags=Edit Tags
@@ -1652,7 +1651,7 @@ file.dataFilesTab.versions.viewDetails.btn=View Details
 file.dataFilesTab.versions.widget.viewMoreInfo=To view more information about the versions of this dataset, and to edit it if this is your dataset, please visit the <a href="/dataset.xhtml?persistentId={0}" title="{1}" target="_blank">full version of this dataset</a> at the {2}.
 file.dataFilesTab.versions.preloadmessage=(Loading versions...)
 file.previewTab.externalTools.header=Available Previews
-file.previewTab.button.label=Previews
+file.previewTab.button.label=Preview Tools
 file.previewTab.previews.not.available=Public previews are not available for this file.
 file.deleteDialog.tip=Are you sure you want to delete this dataset and all of its files? You cannot undelete this dataset.
 file.deleteDialog.header=Delete Dataset
@@ -1793,6 +1792,8 @@ file.DatasetVersion=Version
 
 file.previewTab.tool.open=Open
 file.previewTab.header=Preview
+file.previewTab.presentation=File Preview Tool
+file.previewTab.exploreBtn={0} on {1}
 file.metadataTab.fileMetadata.header=File Metadata
 file.metadataTab.fileMetadata.persistentid.label=Data File Persistent ID
 file.metadataTab.fileMetadata.downloadUrl.label=Download URL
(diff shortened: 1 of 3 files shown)


mheppler:
@qqmyers I was hoping you can help clarify a problem with the Preview tab render logic on the file pg in this branch, as well as develop. @kcondon was testing this PR and found that your changes to the render logic were not behaving as expected in my branch.

In your PR #6420 which was merged into develop on 12/2, you changed the render logic of the tab on file.xhtml to:

```
... and FilePage.fileMetadata.datasetVersion == FilePage.fileMetadata.datasetVersion.dataset.getLatestVersionForCopy()
```

But, in another commit on 12/2 37906d3800d6876ebd5fea6911c80d1a87bffac1, directly to the develop branch (I'd scold you but I did the same thing last week), you changed it again to:

```
... and FilePage.fileMetadata.datasetVersion.isPublished()
```

That latest commit seems to have revert the bug fix of your PR. Can you provide some guidance here on what that latest commit does? Was it an unintended commit? Thanks.

qqmyers:
@mheppler - I don't think I have privs to commit to dev - these should have been from 2 PRs. I did try to fix a merge issue and ended up making a null PR, but I then added a corecting commit.

FOr the two PRs: The first one corrected an issue where the previewers would not show if there was a draft version since the current version (i.e. a published v 2.0) would not equal the latest version, but would equal the latestversionforcopy (basically the latest published/public one). 
The second issue/PR was that, even with this fix, one could not preview files in earlier versions (e.g. a file only in the published v1 version). The second commit is to replace the comparison between the current version and the latest/latestforcopy with a simple test to see if the current version is public. Since this PR changes the same line as the other one, it led to a merge conflict (which I then bungled before correcting).

After both are in, I think the FilePage.fileMetadata.datasetVersion.isPublished() should fix both issues.

mheppler:
Ahh, my apologies. I didn't notice there were two PRs (#6420, #6422) to the same line of code, on the same day, from the same developer.

I'll tell you what I am seeing in develop/my branch, and you can tell me if it's behaving as expected.

- preview tab displays on file pg for file in V1 of a dataset with no other versions, or drafts
- preview tab displays for file in V1 of a dataset with no other versions, BUT has a draft (fixed in #6419)
- preview tab does NOT display for file in DRAFT of a dataset
- preview tab does NOT display for file in V1 or V2 of a dataset with no draft (still a bug from #6421?)

kcondon:
@qqmyers Jim, have seen a couple issues that were not addressed by your fixes. Not sure whether you would have included them if known: 1. If file is restricted or request for access is specified. 2. if tou specified, 3. if gb specified. In any of these cases, preview tab will not appear.

qqmyers:
@mheppler  - I'm surprised if you don't get a preview for v1 or v2 when there's no draft, unless it is due to one of the issues @kcondon brings up. For those issues, my understanding is that those limitations were 'by design' at this point - awaiting thinking about how to display the terms/ask for gb info, etc. when the file isn't public. (These should still be viewable through the external preview button which has a way to popup terms, etc. - it's just the in page display). So I took them as out of scope at present and just tried to fix issues where I saw valid public datasets not getting preview displays.

mheppler:
@qqmyers my apoloogies, @kcondon helped me diagnose what I was seeing, which was the expected behavior because those files were in datasets that had terms or guestbooks.

Someday this we will build out _File Page needs a Term Tab to define terms at the file level #4391_ and all this preview tab render logic will be a lot easier to determine per file.

PR #6464, opened 2019-12-16
get tabs working on user page on Payara #6463

### LINKED ISSUE #6463: Glassfish Upgrade: issues with javascript and  "//<![CDATA[" and "//]]>"
Split off from #6230 

First the issues:
• user page - if you don't start on API token tab, clicking on that tab is blank
• user page - if you don't start on My Data tab, clicking on that tab is blank
• dataset page - click on versions tab, nothing happens
• dataset page - selecting files on files table, links broken in "## files are currently selected.   Select all ## files in this dataset.  Clear selection." msg above table
• dataset page - selecting edit dataset does nothing
• edit files page - upload doesn't add file to list; if you add a 2nd file, 1st file gets added (true for n+1 and n)
• create Dataset - if validation errors, block ui does not get removed, errors do not appear
• dataset pg + file pg, clicking tabs, in FF (68.2.0) multiple errors in console log: `XML Parsing Error: mismatched tag. Expected: </update>`

Basically, all these issues are caused by XML Parsing error due to the "//<![CDATA[" and closing "//]]>" on various pages. If we remove them, things work again. For example, remove them from my_data_fragment.xhtml and that page works again. remove them from dataset-versions.xhtml and that fixes not only the versions tab, but also the files table links.

That said, I'm not sure removing them is the exact solution we want. They are there to help XML parsing of javascript on xhtml pages (if they contain certain characters, e.g. "<" or "&"). But once we figure out how we want to solve one of these, we can solve all of them (*).

pdurbin:

On Payara, users can't use the "API Token" tab. It's blank.

Closes #6463

I thought I'd try simply deleting all the extra Javascript in the xhtml page and it seems to have worked. I can get to the tabs now in both Firefox and Chrome. I'm not sure if I deleted too much, though. Maybe we need some that Javascript? If so, I could try a smaller change.

Also I was kind of hoping "edit user - says update successful, but changes not saved to db (somewhat similar to dataverse)" would be magically fixed but it isn't. 😢 

Try to replicate this:  user page - if you don't start on API token tab, clicking on that tab is blank; same behavior with My Data tab #6463 

No

No

**Additional documentation**:

No

### DIFF
--- a/src/main/webapp/dataverseuser.xhtml
+++ b/src/main/webapp/dataverseuser.xhtml
@@ -17,15 +17,6 @@
             <ui:param name="showDataverseHeader" value="false"/>
             <ui:param name="loginRedirectPage" value="dataverseuser.xhtml"/>
             <ui:define name="body">
-                <script>
-                    //<![CDATA[
-                    window.onpageshow = function(event) {
-                        if (event.persisted) {
-                            window.location.reload();
-                        }
-                    };
-                    //]]>
-                </script>
                 <f:loadBundle basename="propertyFiles.Bundle" var="bundle"/>
                 <f:metadata>
                     <f:event type="preRenderView" listener="#{facesContext.externalContext.response.setHeader('Cache-Control', 'no-cache, no-store')}"/>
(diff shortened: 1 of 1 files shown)


scolapasta:
@pdurbin I tried this and it seems to have "sort of" worked for API token tab, and not at all for My Data. Is that what you're seeing?

I say sort of, because one of those scripts "removes "hidden" class from div containing API token to stop javascript browser exploits". So without it, while you see some text on API token, the actual token itself is hidden.

I was actually able to achieve the same result by just removing / moving that script and leaving the rest. But of course, we do need the logic of this script. Note #6470 seems to also be about a conflict with some embedded javascript.

scolapasta:
Closing, as I've learned more about the issue and we'll need a more surgical solution.

PR #6478, opened 2019-12-20
#6426 get actual prod date for export

### LINKED ISSUE #6426: "Production date" hard coded in JSON metadata export
In recent versions of Dataverse, including the latest 4.18.1, the"JSON" metadata export of every dataset includes `productionDate: "Production Date"`.

<kbd>Screen Shot 2019-12-02 at 1 32 54 PM</kbd>

It seems to be hard coded, appears only in the "JSON" export, and doesn't seem to affect how the "Production Date" metadata (that depositors fill out) is recorded (see `typeName: productionDate` at the bottom of the screenshot).

But it could confuse people who are trying to make sense of the metadata in the JSON export and I think should be removed. Then I think each installation will need to re-publish each published dataset's JSON metadata export.

sekmiller:
Fixes a formerly hard-coded item in the json export (Production Date)

Closes #6426 

none
must do a re-export or republish in order to see the effect.

no

suggest a re-export all to see the effect of bug fix

**Additional documentation**:
None

### DIFF
--- a/doc/release-notes/6426-reexport-all
+++ b/doc/release-notes/6426-reexport-all
@@ -1,0 +1,14 @@
+## Notes for Dataverse Installation Administrators
+
+### Run ReExportall
+
+We made changes to the JSON Export in this release (#6246). If you'd like these changes to reflected in your JSON exports, you should run ReExportall as part of the upgrade process. We've included this in the detailed instructions below.
+
+
+
+
+
+
+xxx. (Optional) Run ReExportall to update JSON Exports  
+
+   <http://guides.dataverse.org/en/4.19/admin/metadataexport.html?highlight=export#batch-exports-through-the-api>
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
@@ -675,7 +675,7 @@ public class DatasetVersion implements Serializable {
     }
 
     public String getProductionDate() {
-        String retVal = "";
+        String retVal = null;
         for (DatasetField dsfv : this.getDatasetFields()) {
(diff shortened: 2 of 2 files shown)


djbrooke:
Release note about exportall added in https://github.com/IQSS/dataverse/commit/80c429bbc3bb3cbf0ac49d49758b8d5d14eb8331

doc/release-notes/6426-reexport-all
```diff
@@ -0,0 +1,14 @@
+## Notes for Dataverse Installation Administrators
+
+### Run ReExportall
+
+We made changes to the JSON Export in this release (#6246). If you'd like these changes to reflected in your JSON exports, you should run ReExportall as part of the upgrade process. We've included this in the detailed instructions below.
```
pdurbin: "We made changes to the JSON Export" sounds a little scary to me but we can reword this when we write the actual release note. What we're really doing is populating a data field with a date instead of the word "Production Date".

doc/release-notes/6426-reexport-all
```diff
@@ -0,0 +1,14 @@
...
+
+
+
+
+
+
+xxx. (Optional) Run ReExportall to update JSON Exports  
+
+   <http://guides.dataverse.org/en/4.19/admin/metadataexport.html?highlight=export#batch-exports-through-the-api>
```
pdurbin: In the actual release note, I'd suggest removing `?highlight=export` from this URL.

kcondon:
@sekmiller This did not work for me. Production Date remains in json afte reexortall

sekmiller:
After re-export all I wasn't seeing "productionDate":"Production Date", I was seeing "productionDate":"",
My current fix removes production date from the export entirely if it is missing from the dataset version.
Again a re-export all would be required to see the result.

kcondon:
empty prod date not included, prod date was working, testing error.

PR #6488, opened 2020-01-07
Iqss/6485 - Multiple File Stores

qqmyers:
**What this PR does / why we need it**: Implements the ability to configure multiple file stores.

**Which issue(s) this PR closes**:

Closes #6485 

**Special notes for your reviewer**: The install script should set up a file store, backward compatible with an existing installation. The docs also describe how to do this.

**Suggestions on how to test this**: Multiple things that could be tested:
Basic install comes up with a single file store working as before (can upload/download files)
Ability to configure a second file store (e.g. with a different file path) 
Ability to configure an additional s3 store 
Ability to configure another s3 store using a separate aws profile
Ability to change the default store for new files (new store allows upload/download, files in other stores are still downloadable)
Ability to list stores available and use the admin API and/or superuser UI (both described in docs) to set/reset to default which store is used for a given dataverse.
Verify that a store set for a given dataverse works for datasets created/edited within that dataverse.
With a s3 store configured with the id 's3', verify that rsync still works
Without an s3 store configured with the id 's3', verify that rsync upload does not appear as an option.

**Does this PR introduce a user interface change?**: For superusers, adds an entry in the 'Edit'/'General Information' form for a dataverse to change the store for a given dataverse.

**Is there a release notes update needed for this change?**: Yes - existing installations will have to update some jvm options to move to this version. The config docs describe the changes needed for a file store or s3 store.

**Additional documentation**:
DCM/rsync has hardcoded using a vanilla s3 store. Since I don't have a good way to test any rsync changes, I didn't try to update it to support multiple stores, but did try to make changes to enable an existing instance to keep using DCM/rsync. Adding the jvm options to define an s3 store with id 's3', as described in the docs, should be enough to make DCM work with this update. It should be possible to make the rsync capability available per store, and support stores with specified endpoints, etc. in the future.

PR #6025, opened 2019-07-16
5586 duplicate component names

sekmiller:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Thanks!

## Related Issues

- #5586 JSF Errors duplicate component ids

## Pull Request Checklist

- [x] Unit [tests][x]none UI only
- [x] Integration [tests][x]: None
- [x] Deployment requirements: None
- [x] [Documentation][docs] None
- [x] Merged latest from "develop" [branch][x] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

Issue #5919, opened 2019-06-05
FITS File Upload: no longer extracting metadata


jggautier:
Harvard Dataverse and Demo Dataverse, both on v4.14, aren't extracting metadata from FITS files anymore.

On Harvard Dataverse, it looks like the last published file where metadata was extracted was on June 19, 2018. Since then, no uploaded file whose file type is FITS has had its metadata extracted.

https://github.com/IQSS/dataverse/issues/2904 is related

jggautier:
In [this dataset on Demo Dataverse](https://demo.dataverse.org/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2FKG5TTZ) I tested publishing files whose metadata Harvard Dataverse has and has not extracted.

meghangoodchild:
In `[this dataset on Demo Dataverse](https://demo.dataverse.org/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2F8HWLO3)`, I tried uploading a FITS file and the metadata was not extracted. It is also not working on our instance at Scholars Portal Dataverse (currently on 4.17). Would anyone be able to look into this? Thanks!

djbrooke:
@meghangoodchild - we'll bring this into a sprint and fix this in a release soon.

qqmyers:
FYI : https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java#L169 looks suspicious.  The next few lines take the storage identifier and strip any tmp:// prefix beofre turning it into a file path. Line 169 doesn't strip the prefix before prepending the tempfile dir and then sends that to the metadata extractor. 

I just saw this while working on other things and haven't tested to see if this is really the issue.

qqmyers:
Probably not the real issue - just saw that at least single file uploads don't have tmp:// at the start of their storageidentifiers in this method, so this may be a red herring or only a secondary problem (presumably some files have tmp:// in their storageidentifiers when they go through this method?).

djbrooke:
- In the short term we'd like to restore the functionality to the way it was before
- Longer term, once we have better support for file level metadata, we can remove the hacky extraction in place right now (stuffing it into the description, storing it in dedicated database tables and rolling it up to the dataset metadata) and investigate moving that to the representations that we want

mheppler:
Localhost server log errors from uploading two FITS files. Checked the temp directory path locally,  files are not there.

```
[...]

Issue #5586, opened 2019-02-28
JSF errors when adding a dataset: duplicated component IDs


poikilotherm:
Steps to reproduce:
1. Use Firefox 60+
2. Start a fresh installation, preferrably using Minikube and IQSS/dataverse-kubernetes
3. Create a dataset and watch the errors in the logs ([exception-log.txt](https://github.com/IQSS/dataverse/files/2914803/exception-log.txt)).

I am not sure if this is something related to my setup and could not yet investigate further. Maybe others have seen those, too?

mheppler:
@poikilotherm Thank you for catching this, reporting it and including your log file. Combing through that, I found the offending code snippet.

```
      +id: linkDatasetForm
       type: org.primefaces.component.dialog.Dialog@51e315a1
        +id: j_idt1870
         type: 
                        <div class="form-horizontal">
                            <p class="help-block">
                                
        +id: j_idt1871
         type: javax.faces.component.html.HtmlOutputFormat@6df3054
          +id: j_idt1872
           type: org.omnifaces.component.output.Param@5a8a3901
            +id: j_idt1873
             type: org.primefaces.component.commandlink.CommandLink@36752614
        +id: j_idt1874
         type: 
                            </p>
                            <div class="form-group">
                                <label class="col-xs-3 control-label">
                                    
        +id: j_idt1875
         type: javax.faces.component.html.HtmlOutputText@7ff27366
        +id: j_idt1876
         type: 
                                </label>
                                <div class="col-xs-8">
                                    
        +id: linkNameContent
         type: org.primefaces.component.fragment.Fragment@d0a15c
          +id: dataverseLinkName
           type: org.primefaces.component.autocomplete.AutoComplete@b19c276
            +id: j_idt1877  <===============
             type: org.primefaces.component.column.Column@14c30087
              +id: j_idt1878
               type: javax.faces.component.html.HtmlOutputText@78c6383a
            +id: j_idt1879
             type: org.primefaces.component.column.Column@7d19bc3f
              +id: j_idt1880
               type: javax.faces.component.html.HtmlOutputText@38d6db98
            +id: j_idt1877  <===============
             type: org.primefaces.component.column.Column@349d233d
```

The duplicate id of `j_idt1877` is being generated by the `p:autoComplete` component in the linkDatasetForm popup on the dataset pg. We do not have control of that id and the elements it is applied to, because it all generated by PrimeFaces.

[...]

PR #6490, opened 2020-01-07
Iqss/6489 - Direct uploads to S3 using presigned URLs

### LINKED ISSUE #6489: Support larger file uploads to S3 stores via direct upload using presigned URLs
One of the bottlenecks to handling large files (e.g. 1-100GB+) in Dataverse is that the existing implementation streams them through the Glassfish server and creates a local temporary file. 

Working for TDL, I've developed a mechanism to directly upload files to S3 stores using presigned URLs. The overall concept is analogous to how the direct s3 download works, in which Dataverse creates a presigned URL allowing download of a file that it then sends to the client's browser, which redirects to that URL and directly retrieves the file content from the s3 store. To support upload, I added an API call to request a presigned direct upload URL and have then adapted the DVUploader to use it. I've also used Javascript to catch upload requests in the Dataverse web interface and, for a store that supports direct upload, use Ajax calls to transfer the file directly. I then extended the existing upload methods to be able to accept the storageId given to that file instead of the file content stream itself. 

In the initial implementation for TDL, I disabled any ingest processing that requires access to the file contents (i.e. unzip, format conversion, metadata extraction, mime-type determination). Per discussion on the DV Community call, my intent is to restore all of that capability, perhaps up to a file size limit and perhaps excluding the unzip step. (The primary concern is that the benefit of being able to stream larger files direct to S3 will be negated if Dataverse then retrieves the file, unzips its content, etc. A size limit would be a simple way to handle this concern to start.)

TDL is currently testing this code and we've uploaded a 39GB file this way already. We have not yet tested performance relatively to the existing uploads (for smaller files where both should work) but hope to do so. The current implementation for the web interface streams the file to S3 and then calculates an MD5 hash locally on the user's machine before triggering Dataverse to update the dataset so nominally the overall performance should be similar to the raw performance of the S3 upload plus the local machine's MD5 hash performance. (As with the upload to Glassfish, multiple files upload in parallel in the web interface, helping to maximize use of available bandwidth.)

One additional benefit of this approach is that it is ~transparent to the user. There is no visible change to the file upload process through the Dataverse web interface.

I'll be creating a PR, possibly building off of #6488 that supports multiple stores. Whether this is something that makes sense to merge in in its current form is TBD, but the approach looks promising enough that I wanted to get it on the radar.

qqmyers:
**What this PR does / why we need it**: This PR supports direct uploads to S3 using presigned URLs

Closes #6489 

**Special notes for your reviewer**: This PR builds on #6488. It could be modified to work independent of multiple stores, but that's not how it was developed and I suspect multiple stores will be merged before this. As of now, this branch is up to date with dev plus #6488.

 The initial state of the PR bypasses all ingest processing for direct uploads. I'll be adding that back in. Up to you guys if you want to test prior to that. Also owe some documentation.

Once a store is configured to do direct uploads via a jvm option, the store should work as normal w.r.t. user interface uploads. Testing should probably focus on making sure that's true. Once the PR is updated to support ingest processes, testing could check that the processing is the same between a normal store and a direct upload one.
There will also be a version of the DVUploader that can use the direct upload capability when a store supports it (next PR). That could be used to test the API calls directly.
Performance would be nice to test as well - overall speed and max file size.

None beyond that in #6488. That said, some elements, like the upload progress bar are being controlled via Javascript rather than by PrimeFaces when direct upload is used, so there could be some unintentional/minor style differences (something else to test...).

Could be worth mentioning, but everything beyond #6488 is optional, so there are no required upgrade steps.

**Additional documentation**:


djbrooke:
Thanks @qqmyers, very exciting to see this. I'll add this to the code review column so that someone more technical than me can make the call about whether to provide feedback now or wait until #6488 is merged.

qqmyers:
I've modified the code to do tabular and FITS processing (still to be tested), up to a configurable size limit. 
Looking at other ingest functionality, e.g.the unzip and MIMEtype determination code, - it looks like these could be done much more efficiently than a naive implementation that pulls the whole file from S3 to create a temp file that then gets used as in the normal upload (creating temporary local files for unzipped items, etc.). For example, most of the MIME tests just look at the first few (from 5 to <1K) bytes and S3 has a call to just retrieve part of the object that could be used. Similarly, I think that unzipping could be done with more streaming to limit local disk use. (It's also possible to read a zip index and pull a single file out by grabbing the right bytes - so it might be possible to just store the zip itself.)

Given that, I guess I'd suggest that this PR get looked at ~as is and, assuming that unzipping and mimetype processing makes sense, to create a new issue for those to be added at some point as a separate PR.

pdurbin:
> This PR builds on #6488

Ok, I think this means that this link will allow a code review to reason about this pull request in isolation, as a smaller pull request once the other one is merged first: https://github.com/TexasDigitalLibrary/dataverse/compare/IQSS/6485...TexasDigitalLibrary:IQSS/6489

> some elements, like the upload progress bar are being controlled via Javascript rather than by PrimeFaces when direct upload is used, so there could be some unintentional/minor style differences

I thought I'd just flag this early for @TaniaSchlatter and @mheppler to look at. UI impact.

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -6,7 +6,29 @@ Big data support is highly experimental. Eventually this content will move to th
...
+
+To configure these options, an administrator must set two JVM options for the Dataverse server using the same process as for other configuration options:
+
+``./asadmin create-jvm-options "-Ddataverse.files.<id>.download-redirect=true"``
+``./asadmin create-jvm-options "-Ddataverse.files.<id>.upload-redirect=true"``
+
+
+With multiple stores configured, it is possible to configure one S3 store with direct upload and/or download to support large files (in general or for specific dataverses) while configuring only direct download, or no direct access for another store.  
+
+At present, one potential drawback for direct-upload is that files are only partially 'ingested', tabular and FITS files are processed, but zip files are not unzipped, and the file contents are not inspected to evaluate their mimetype. This could be appropriate for large files, or it may be useful to completely turn off ingest processing for performance reasons (ingest processing requires a copy of the file to be retrieved by Dataverse from the S3 store). A store using direct uplod can be configured to disable all ingest processing for files above a given size limit:
```
pdurbin: Typo. "uplod". Also, do we consider unzipping zip files to be part of ingest? Is processing of FITS files part of ingest? (I thought special handling of FITS files is broken right now: #5919.) My sense is that when we say "ingest" we are usually only talking about tabular files, not zip files or FITS files. Please see http://guides.dataverse.org/en/4.18.1/user/tabulardataingest/ingestprocess.html

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -18,7 +40,7 @@ Install a DCM
 
 Installation instructions can be found at https://github.com/sbgrid/data-capture-module/blob/master/doc/installation.md. Note that shared storage (posix or AWS S3) between Dataverse and your DCM is required. You cannot use a DCM with Swift at this point in time.
 
-.. FIXME: Explain what ``dataverse.files.dcm-s3-bucket-name`` is for and what it has to do with ``dataverse.files.s3-bucket-name``.
+.. FIXME: Explain what ``dataverse.files.dcm-s3-bucket-name`` is for and what it has to do with ``dataverse.files.s3.bucket-name``.
```
pdurbin: I'm not sure what's going on here. Maybe the JVM option was mis-documented? Also, I think `..` means it's commented out but I'm not sure.

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -100,6 +122,7 @@ Optional steps for setting up the S3 Docker DCM Variant
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Before: the default bucket for DCM to hold files in S3 is named test-dcm. It is coded into `post_upload_s3.bash` (line 30). Change to a different bucket if needed.
+- Also Note: With the new support for multiple file store in Dataverse, DCM requires a store with id="s3" and DCM will only work with this store.
```
pdurbin: Heads up to @pameyer about this if we merge this. Affects DCM.

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -132,7 +158,7 @@ Optional steps for setting up the S3 Docker DCM Variant
 
     - S3 bucket for Dataverse
 
-      - ``/usr/local/glassfish4/glassfish/bin/asadmin create-jvm-options "-Ddataverse.files.s3-bucket-name=iqsstestdcmbucket"``
+      - ``/usr/local/glassfish4/glassfish/bin/asadmin create-jvm-options "-Ddataverse.files.s3.bucket-name=iqsstestdcmbucket"``
```
pdurbin: Why was this changed from a hyphen to a period?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -208,10 +208,47 @@ As for the "Remote only" authentication mode, it means that:
...
+By default, a Dataverse installation stores all data files (files uploaded by end users) on the filesystem at ``/usr/local/glassfish4/glassfish/domains/domain1/files``. This path can vary based on answers you gave to the installer (see the :ref:`dataverse-installer` section of the Installation Guide) or afterward by reconfiguring the ``dataverse.files.directory`` JVM option described below.
+
+Dataverse can alternately store files in a Swift or S3-compatible object store, and can now be configured to support multiple stores at once. With a multi-store configuration, the location for new files can be controlled on a per-dataverse basis.
+
+The following sections describe how to set up various types of stores and how to configure for multiple stores.
+
+Multi-store Basics
++++++++++++++++++
+
+To support multiple stores, Dataverse now requires an id, type, and label for each store (even for a single store configuration). These are configured by defining two required jvm options:
```
pdurbin: If single store configurations need to do something when they upgrade, I hope there's a release note about this.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -208,10 +208,47 @@ As for the "Remote only" authentication mode, it means that:
...
+    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.<id>.label=<label>"
+
+For backward compatibility, the id and type should be the same, and the label can be set to the same value as well. For example, the following would define a backward compatible file store:
+
+.. code-block:: none
+
+    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.file.type=file"
+    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.file.label=file"
+  
+or a file or swift store, this is all that is needed for backward compatibility. For an s3 store, any additional options must be changed to conform to the new, more consistent naming convention using a . after the store's id ('s3' for backward compatibility) For example:
```
pdurbin: We typically put information like this (what to change after you deploy a new version) in release notes. I guess it's ok to have it here too but I prefer to think of the audience as being someone who is installing Dataverse for the first time. This could be their first day with Dataverse. They probably don't need information about how things used to be (Don't Make Me Think).

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -224,12 +261,13 @@ First, run all the following create commands with your Swift endpoint informatio
 
 .. code-block:: none
 
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.defaultEndpoint=endpoint1"
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.authType.endpoint1=your-auth-type"
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.authUrl.endpoint1=your-auth-url"
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.tenant.endpoint1=your-tenant-name"
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.username.endpoint1=your-username"
-    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.swift.endpoint.endpoint1=your-swift-endpoint"
+    ./asadmin $ASADMIN_OPTS create-jvm-options "\-Ddataverse.files.<id>.type=swift"
```
pdurbin: I'm confused. Don't we need to explain in this section (Swift Storage) what to put for `<id>`?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -267,8 +305,8 @@ You also have the option to set a custom expiration length, in seconds, for a ge
 In this example, you would be setting the expiration length for one hour.
 
 
-Setting up Compute
-+++++++++++++++++++
+Setting up Compute with Swift
```
pdurbin: I appreciate this clarification that the Compute button is currently tied to Swift. 😄

qqmyers:
the DVUploaderv1.0.7.jar - just released and available from the dataverse-uploader repository can now do the direct upload when using the new -directupload flag. For example - the following should work with appropriate values inserted. (-recurse goes through subdirs, -trustall allows you to have a self-signed cert for testing).

java -jar DVUploader-v1.0.7.jar -server=https://dataverse-dev.tdl.org -key=<key> -did=doi:10.33536/FK2/BECWEL -directupload -recurse -trustall <local_dir>

doc/sphinx-guides/source/developers/big-data-support.rst
qqmyers: My change is just to update the option name to what it will be. Whether the fixme is still considered relevant, I don't know.

doc/sphinx-guides/source/developers/big-data-support.rst
qqmyers: When there was only one possible s3-bucket-name, a hyphen made sense. Now that there can be multiple S3 stores, we need options for each store's bucket name. I chose to standardize on
-Ddataverse.files.<id>.bucket-name 
and to have all of the options for all three types of stores be of the form
-Ddataverse.files.<id>.<option name>
rather than have a mix of . and hyphens depending on type. 
If there's an argument for a better convention, I'm open.

doc/sphinx-guides/source/installation/config.rst
qqmyers: I'm fine with it being in release notes instead. I figured I'd put it here since I don't know what version this might become part of.

doc/sphinx-guides/source/installation/config.rst
qqmyers: Whatever makes sense. The general concept is that all options are now of the form
-Ddataverse.files.\<id\>.\<option name\>
so there's nothing swift specific about the naming per se. I could say more in the multistore basics section?

doc/sphinx-guides/source/developers/big-data-support.rst
qqmyers: What do you want to call the rest of it? Processing on upload?
For FITS, this branch should now be just-as-broken as the other. I just assured that the same processing is triggered when there is no temp file - handing the code a stream to the final file instead.

doc/sphinx-guides/source/developers/big-data-support.rst
qqmyers: FWIW: As long as there's a store with the id 's3', as recommended for backward compatibility, DCM should be unaffected/ will continue to work on an instance that upgrades from a single s3 store today.

In looking through the DCM code, I did see that it creates an AWS client without using any of the s3 store options that have (since?) been created to support minio, etc. Given that, I think DCM can only currently work with a vanilla AWS S3 server and this branch won't affect that.

djbrooke:
Hey @qqmyers, we discussed this briefly after standup this morning and we'll plan to merge this after #6488. Let me know if anything changes in terms of priority from your perspective and whether or not we should revisit this decision.

pdurbin:
@qqmyers hi! I hope you're enjoying Tromso! I loved @poikilotherm 's tweet at https://twitter.com/poi_ki_lo_therm/status/1220345822073761798

@hackdna and I were just talking at #6535 at http://irclog.iq.harvard.edu/dataverse/2020-01-23#i_116809 and I was wondering... could pre-signed URLs help with uploads from Dropbox?

qqmyers:
I don't think it helps as is. The direct upload allows the local browser to directly PUT the file to S3. For DropBox, it's being a passive endpoint that the Dataverse server uses - Dataverse does a GET and then streams the data to a local temp file/ there's no party like the browser that could use a presigned URL. Things could potentially be written in the server to see if the data is going to an S3 store with direct upload capabilities and, if so, to at least send the stream directly to S3 from the server (cutting out the temp file). If the dropbox issue is something that is causing the GET to Dropbox to fail (a timeout somewhere, etc.), then recoding to use direct upload might help, e.g. if  the network to S3 is faster and streaming from DropBox could finish sooner. If the error is later in the processing (which may not be too likely, the Dropbox uploads reuse the same code as normal uploads through the UI w.r.t. unzip, ingest, etc.) then changing the initial upload to use direct presigned S3 wouldn't fix it. (I can't see the initial issue for the bug - has the problem been identified?)

qqmyers:
@djbrooke - I think this is all up-to-date. I've compiled the latest merged version but haven't directly tested it. I have tested TDL's v4.18.1 dev instance with the latest changes from the multi-store and direct upload branches combined though, so I think this is ready to go through Q/A, unless you know of some particular issue.

djbrooke:
Thanks @qqmyers! I'm going to move this to code review. Three quick questions/thoughts about the implementation:

- Is there a way to disable this and enable this for certain users/groups? This may not be a TDL use case but it's something that we need to consider in Harvard Dataverse since we're open to the world and may not want to support large files because of transfer/storage costs. 
- Are there any new JVM options that need to be documented in a release note, or is there any other content that should be added in the release notes? If so, can that be added using the 6489-release-note.md format in the release note directory?
- Thinking towards implementation, if we add this and don't have ingest for now, we'll want to reingest in the future when ingest is optimized/modularized. We have a reingest API endpoint and all! This would update/add the UNF for a citation so it would require republishing. I'm not sure of the preferred way to identify files that were brought in this way, and we'd have to think about how to communicate that authors can reingest, republish, and get the benefits of explore/UNFs/etc.

qqmyers:
@djbrooke 

1) The potential for direct upload is configured by jvm-option per store, and which Dataverses  use a given store is configurable by superadmin, so, in general, there are ways to control who can use it. Direct uploads are also controlled by the max size limit, which is now configurable per store. 
2) Turning this on requires setting a direct upload jvm option per S3 store, but directupload is optional. I can add a release note to mention that. This does assume multi-store configuration changes, which involve several jvm-option changes, but those should be described in that PR's notes already.
3)Ingest and FITS metadata extraction are enabled in this version, up to a max file size limit (currently defined by an optional property - could also be added to a release note, or made per store as a jvm-option if a global limit isn't acceptable). That said, the question of what one does if the limit changes, or how to reprocess for other reasons (e.g. inspecting file content to determine a mimetype is not enabled at all right now), is an open question. I did do work for QDR to make ingest something that can be redone in the UI - one can uningest to remove errors and ingest/reingest, with the changes triggering creation of a draft as usual if needed. Combined with the ability to republish without version change (or with one), that work might provide a partial mechanism for this (I can make a PR if desired.).

qqmyers:
@djbrooke - I misspoke in the (now edited) comment above: maximum file size is still controlled by the single :MaxUploadFileSizeInBytes property, and the max size to run ingest on is controlled by a per store jvm option. Since the store is known when the upload UI is rendered, the max size allowed could be made store-specific, e.g. so that a normal S3 store is limited to <2.5GB and a direct upload store has a larger limit (but is only available to users in certain dataverses). If that's needed, let me know and I can switch the max upload size logic to be per store (as a jvm option unless you want store specific properties).

landreev:
@qqmyers Could you please sync the branch with develop.
Also, a couple of tests in DataAccessTest.java are failing:
Screen Shot 2020-02-25 at 2 32 34 PM
(still failing after syncing with develop locally)
But even if I disable these tests, I'm having trouble getting direct downloads to work. 
I have s3 configured and set as the default driver (uploads and downloads are working). I have `dataverse.files.s3.upload-redirect` set to true...
But `EditDatafilesPage.directUploadEnabled()` is returning false; and inside the method, `this.dataset.getDataverseContext().getStorageDriverId()` is returning null. (?)

I was wondering if I was doing it in the wrong order - i.e., if we needed to merge #6670 first, in order to get this PR to work... But I'm not seeing anything relevant there either.

landreev:
@qqmyers  the thing with `getDataverseContext().getStorageDriverId()` returning null for the root dataverse - I can work around that by explicitly setting the storage driver to s3 for a dataverse. Still can't get the actual direct upload to work though.

qqmyers:
@landreev - I was out yesterday afternoon but am looking into these issues now. I think merging #6670 first would help with the tests, but probably not the main issue. A lot of what was done in fixing #6488 affected how defaults/inheriting stores were handled so there's probably something still making old assumptions in the direct upload code.

qqmyers:
Found two places where getStorageDriverId() had to be updated to getEffectiveStorageDriverId() which handles inherited storage.  That should address the null pointers when the driver isn't set locally. Not sure that these would affect the case where it was set locally though, so I'll keep looking/ see if I can upload directly.

landreev:
> Found two places where getStorageDriverId() had to be updated to getEffectiveStorageDriverId() which handles inherited storage. That should address the null pointers when the driver isn't set locally. Not sure that these would affect the case where it was set locally though, so I'll keep looking/ see if I can upload directly.

OK, that makes sense. I saw the getEffectiveStorageDriverId() method in DataAccess last night, and it looked like it was doing the right thing, but figured maybe it wasn't always called. 
When I have the correct s3 driver selected for a dataverse, and have the upload-redirect set - what am I supposed to be seeing? You mention that it may be cosmetically different, because your javascript goes around primefaces. What about the progress bar - is it supposed to look completely different from what primefaces shows us, or is it the same component underneath? This is what I get when trying to upload a file: 
Screen Shot 2020-02-25 at 5 25 32 PM
i.e., a progress bar that looks suspiciously identical to the primefaces thing; that completes, but then the whole thing freezes and nothing else happens. 
Without the upload-direct option, upload works as expected and files get saved in the s3 bucket.  

Just sending this along in case it helps understand what's going on on my system. May still be something I didn't set up right. (I'm working on other things now; but will move #6670 along asap)

qqmyers:
@landreev - it should look ~the same. The javascript retrieves the directupload URL, runs that showing progress from 0-50% and it may pause there if is a delay in getting the final http response (i.e. some testing at TDL is with a proxy that has to finish resending the data before it responds to the browser, so there's a delay between the javascript seeing all bytes submitted and the next step starting). Then, the code does an md5 hash with progress going from 51-100% at which point is sends a final message to Dataverse to do the equivalent of primefaces finishing the file upload at which point ingest/other processing happens and the file should show up in the normal lower panel, awaiting your save of all files. If there's no proxy, there may be no visible pause at 50% but you may notice a difference between the rate for the first half (depends on network speed) and second half (should just be local disk speed).

For debugging, looking in the browser's javascript console to see what errors there may be and/or what HTTP calls succeeded/failed could help. If it's an issue with Dataverse (which it could be at 100%) then the Dataverse log could help too.

One other note: Just like multistore, there are enough differences between uploading via API, uploading during dataset creation, and later uploads that all three should be tested (and noting which way you tried things could help in debugging).

qqmyers:
At TDL, uploading to an existing dataset is working - with all of the pre #6670 changes for multi-store and directupload applied to v4.18.1. 

Given that you've got normal upload working, I can't think of any DV config issues that could be involved. I guess there could be network issues to S3 (do direct downloads work?), e.g. a firewall allowing the DV server but not browsers on other machines to get to the S3 address. Other possibilities could be browser differences - which should throw errors in the javascript if they exist - or some interaction with other changes since 4.18.1/mistakes in my cherry-picking across branches.

landreev:
Just to update on the status: direct upload is now working on our end too. It looks like the reason it wasn't working early on was because it didn't like the CORS rules on our test bucket. 
Other, less serious issues like the failing tests appear to have been fixed by the recent merges. 
Kevin is out today and Mon., but this will go through QA when he's back next week.

qqmyers:
@landreev - awesome! I hadn't gotten back to testing so I'm glad CORS was the (current) issue.

djbrooke:
I'm removing myself from this issue for now, as my main concern was about how we'd implement this in Harvard Dataverse. My understanding, per @landreev's demo, is that we'll be able to set up two stores pointing to the same storage, one with "S3, Direct Upload Enabled" and one with "S3, No Direct Upload." We can assign the S3 direct upload store to those dataverses where we want to allow larger uploads. 

@qqmyers, @landreev - apologies if I missed this in the previous PR, but since we're now allowing :MaxFileUploadSizeInBytes for each store, will the depositor see the appropriate, dataverse/store level file size limit on the upload screen?

qqmyers:
@djbrooke - hidden in one of the comments above was my backtracking - there is still only one global maxfileuploadsizeinbytes. If that's something we should add now, I need to know whether we want store specific properties or to move this to being a storespecific jvm option. (I think it makes sense for this to be store specific (maybe with a global default?))

djbrooke:
Thanks @qqmyers. I think we'd want this to settable on a store-by-store basis, however that may be implemented. Thanks for being willing to add it. 

@landreev, @scolapasta I'll defer to you regarding the technical implementation of this, whether as a store specific JVM option or something else. Tagging @kcondon as well since he has experience with configuration management and may have thoughts.

landreev:
Oh wait, this hasn't been done yet? - I thought it was ready to go, was wondering why it hasn't been moved into QA. 
I vote for implementing it as a setting - another JVM option would be consistent with the other settings; but the current installation-wide limit is a setting, so let's go with that? 
But would be ok with a per-store JVM option too.
 
> @landreev, @scolapasta I'll defer to you regarding the technical implementation of this, whether as a store specific JVM option or something else. Tagging @kcondon as well since he has experience with configuration management and may have thoughts.

qqmyers:
Just testing now. FWIW - I thought a 'compound' json property value would be better than a setting per store, so right now (not yet in the PR) you can set :MaxFileUploadSizeInBytes to '4000000000' (4GB) as a universal limit (as before) or to '{"default":"4000000000","s3tacc":"44000000000"}' to set a default and a store(s) specific limits. I'll commit that as soon as I make sure it works but if that's not OK, let me know.

landreev:
Makes sense.

> Just testing now. FWIW - I thought a 'compound' json property value would be better than a setting per store, so right now (not yet in the PR) you can set :MaxFileUploadSizeInBytes to '4000000000' (4GB) as a universal limit (as before) or to '{"default":"4000000000","s3tacc":"44000000000"}' to set a default and a store(s) specific limits. I'll commit that as soon as I make sure it works but if that's not OK, let me know.

qqmyers:
OK - this should be ready again: an S3 Tag is used to mark temporary files on stores that support it (AWS, but not minio), progress bar should be green instead of blue, and file size limits can be defined per store. The tags and file size limits are described in the docs. I think that's everything - if not, let me know.

landreev:
Everybody ok with this going into QA? (yay!)
@scolapasta ?

kcondon:
@qqmyers This build does not appear to recognize dta (stata) file type for normal uploads to local (file) or s3 storage. The result is no ingest is performed. Direct upload has not yet been configured. This works fine with develop branch.

qqmyers:
Odd - no log errors? Are the files assigned the same mimetype in this branch as in develop?

kcondon:
@qqmyers No log errors and it works on develop, shows tabular type there, post ingest.

qqmyers:
Can you run with fine logging configured for util/FileUtil.java? For normal uploads, there shouldn't be a difference and most of the code to recognize the mimetype /decide if a file is ingestable is in that class - it would be helpful to know at what point the processing is going wrong.

kcondon:
@qqmyers No log entries with fine enabled, assuming I've done it correctly:

```
/usr/local/glassfish4/bin/asadmin list-log-levels
ShoalLogger	<CONFIG>

edu.harvard.iq.dataverse.util.FileUtil.java	<FINE>
```

qqmyers:
Hmm - One early difference I see is that when I upload a .dta file, when the bottom panel opens on the upload form (where you can edit the name, add a description) the filetype is Unknown versus something like Stata14 binary (what I see in other branches). So something has changed in being able to pick up the file type...

kcondon:
@qqmyers Doesn't it attempt to read the first bytes of the file to determine type? Is it still doing that?

qqmyers:
It should skip that for directupload but not otherwise - that's why I was hoping to see some fine logging comments.

kcondon:
@qqmyers There is some logging after all, it was lost in all the popup noise:
3 files, one .dta, one .jpg, one .txt

```
[2020-03-05T15:28:51.949-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131949] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  Will attempt to save the file as: /usr/local/glassfish4/glassfish/domains/domain1/files/temp/tmp505617200448658709upload]]

[2020-03-05T15:28:51.952-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131952] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  mime type supplied: application/octet-stream]]

[2020-03-05T15:28:51.952-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131952] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  Attempting to identify potential tabular data files;]]

[2020-03-05T15:28:51.980-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131980] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  determineFileType: tabular data checker found application/x-stata]]

[2020-03-05T15:28:51.980-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131980] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  fileExtension=dta]]

[2020-03-05T15:28:51.981-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131981] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  returning fileType application/x-stata]]

[2020-03-05T15:28:51.981-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131981] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  File utility recognized the file as application/x-stata]]

[2020-03-05T15:28:51.983-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131983] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID value: eeca210b-87b7-4393-85ab-734028916e30]]

[2020-03-05T15:28:51.984-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131984] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID (last 6 bytes, 12 hex digits): 734028916e30]]

[2020-03-05T15:28:51.984-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131984] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  (not UUID) timestamp in hex: 170ac62f790]]

[2020-03-05T15:28:51.985-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440131985] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  timestamp/UUID hybrid: 170ac62f790-734028916e30]]

```

```
[2020-03-05T15:28:52.706-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132706] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  Will attempt to save the file as: /usr/local/glassfish4/glassfish/domains/domain1/files/temp/tmp2407269516140127995upload]]

[2020-03-05T15:28:52.706-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132706] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  mime type supplied: image/jpeg]]

[2020-03-05T15:28:52.707-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132707] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  Attempting to identify potential tabular data files;]]

[2020-03-05T15:28:52.708-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132708] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  determineFileType: tabular data checker found null]]

[2020-03-05T15:28:52.708-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132708] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  begin isGraphMLFile()]]

[2020-03-05T15:28:52.712-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132712] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  XML error - this is not a valid graphML file.]]

[2020-03-05T15:28:52.712-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132712] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  end isGraphML()]]

[2020-03-05T15:28:52.712-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440132712] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isFITSFile] [[
  attempting to read 6 bytes from the FITS format candidate stream.]]

[2020-03-05T15:28:53.113-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133113] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  fileExtension=jpg]]

[2020-03-05T15:28:53.113-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133113] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  returning fileType image/jpeg]]

[2020-03-05T15:28:53.114-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133114] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  File utility recognized the file as image/jpeg]]

[2020-03-05T15:28:53.116-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133116] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID value: ab15f604-85c7-4cb5-a49f-a410bcf7b51e]]

[2020-03-05T15:28:53.116-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133116] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID (last 6 bytes, 12 hex digits): a410bcf7b51e]]

[2020-03-05T15:28:53.116-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133116] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  (not UUID) timestamp in hex: 170ac62fbfc]]

[2020-03-05T15:28:53.116-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(1)] [timeMillis: 1583440133116] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  timestamp/UUID hybrid: 170ac62fbfc-a410bcf7b51e]]
```

```
[2020-03-05T15:28:53.819-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133819] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  Will attempt to save the file as: /usr/local/glassfish4/glassfish/domains/domain1/files/temp/tmp2556543048768298943upload]]

[2020-03-05T15:28:53.820-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133820] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  mime type supplied: text/plain]]

[2020-03-05T15:28:53.820-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133820] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  Attempting to identify potential tabular data files;]]

[2020-03-05T15:28:53.820-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133820] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  determineFileType: tabular data checker found null]]

[2020-03-05T15:28:53.821-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133821] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  begin isGraphMLFile()]]

[2020-03-05T15:28:53.823-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133823] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  XML error - this is not a valid graphML file.]]

[2020-03-05T15:28:53.823-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133823] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isGraphMLFile] [[
  end isGraphML()]]

[2020-03-05T15:28:53.823-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133823] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: isFITSFile] [[
  attempting to read 6 bytes from the FITS format candidate stream.]]

[2020-03-05T15:28:53.835-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133835] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  fileExtension=txt]]

[2020-03-05T15:28:53.835-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133835] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileTypeByExtension] [[
  MimetypesFileTypeMap type by extension, for ShortFile.txt: text/plain]]

[2020-03-05T15:28:53.836-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133836] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  mime type recognized by extension: text/plain]]

[2020-03-05T15:28:53.836-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133836] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: determineFileType] [[
  returning fileType text/plain]]

[2020-03-05T15:28:53.836-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133836] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: createDataFiles] [[
  File utility recognized the file as text/plain]]

[2020-03-05T15:28:53.848-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133848] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID value: d8ea48db-8614-43c2-abb4-bb355d724770]]

[2020-03-05T15:28:53.848-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133848] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  UUID (last 6 bytes, 12 hex digits): bb355d724770]]

[2020-03-05T15:28:53.848-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133848] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  (not UUID) timestamp in hex: 170ac62fed8]]

[2020-03-05T15:28:53.849-0500] [glassfish 4.1] [FINE] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=53 _ThreadName=jk-connector(2)] [timeMillis: 1583440133849] [levelValue: 500] [CLASSNAME: edu.harvard.iq.dataverse.util.FileUtil] [METHODNAME: generateStorageIdentifier] [[
  timestamp/UUID hybrid: 170ac62fed8-bb355d724770]]

```

qqmyers:
That helped - try the latest commit - looks like I had a misplaced copy of an else statement from some merge.

kcondon:
@qqmyers That did it, thanks. On to the rest!

kcondon:
@qqmyers Not able to upload files with direct s3 upload enabled. UI displays empty list, log file dumps large amount of html/script error. I believe @mheppler helped with such an issue a while back:

[2020-03-06T10:44:33.778-0500] [glassfish 4.1] [SEVERE] [jsf.duplicate_component_i
d_error] [javax.enterprise.resource.webcontainer.jsf.application] [tid: _ThreadID=
52 _ThreadName=jk-connector(2)] [timeMillis: 1583509473778] [levelValue: 1000] [[
  JSF1007: Duplicate component ID datasetForm:j_idt2038 found in view.]]

qqmyers:
@kcondon - I just tried doing this on an ansible build of the branch as AWS and didn't have this problem. I found two other minor issues, one testing related:

- The FakePidProvider always returned true for alreadyExists() which caused direct upload to fail in create mode only (direct upload assigns an identifier to the dataset which triggers this to be called whereas without directupload the fact that the dataset has no identifier gets an automatic false which avoids the failure otherwise). I changed this to false.
- The default storagedriver created for the root dataverse was "" rather than null, which broke the logic for deciding if directupload was enabled for that case and made the UI use the normal upload.  This is now fixed. It would not have affected other dataverses or the root after it's storagedriver had been changed once.

The endpoint and proxy being null are OK (when they're not set) - just debugging log messages which I've now deleted.

I'm not sure how you got the duplicate component issue - maybe there's some sequence of actions that causes page reloads to cause this, but I'm not sure if it is direct upload related or not. I guess it's possible that some of the callbacks I added cause trouble, but the mirror what others do (see editFilesFragment.xhtml L524-526 - they're not trying to update parts of the page. @mheppler may be able to spot something, but, if not, if there's a reproducible procedure, I can look into as well.

qqmyers:
FWIW: I have seen this in the browser console when viewing the create dataset page, independent of whether directupload is on or not - they appear to be non-fatal, but perhaps are somehow related to what you've seen? 
image

kcondon:
@qqmyers Thanks for fixing those. As for a reproducible procedure, it happens every time for me:
1. create a dataverse whose storage is s3 direct
2. create a dataset in that dataverse, upload 3 files, 1 stata (50by1000), 1 jpg, 1 txt, see issue

I do have 2 s3 storageand 2 file storage defined with 1 s3 defaulting and second the direct upload, all confirmed to work before configuring direct upload. I only have specified direct upload true, nothing in addition. Those don't seem likely to affect ui except maybe my not setting an upload limit? I think that may post text to ui.

qqmyers:
Hmm - when are you seeing a blank list? After choosing the files (dialog, drag/drop?)? After the files upload / progress bar happens? Or?

I'll keep testing and try to add other stores, match you set of files, etc.

qqmyers:
@kcondon - I can now occasionally see a race condition on the dataset create form that gives the duplicate component error. I'll dig into that.

kcondon:
@qqmyers Progress!? Sorry, did not see your earlier question: I am seeing various things but mostly consistent: I'm seeing blank progress bars on upload, blank list on save.

qqmyers:
Looks like the problem is related to the <p:fragment id="linkNameContent">. I see the following the matches the idt_2038 entry you see in the logs:
       +id: linkNameContent
         type: org.primefaces.component.fragment.Fragment@1c9742b0
          +id: dataverseLinkName
           type: org.primefaces.component.autocomplete.AutoComplete@72106de7
            +id: j_idt2038  <===============
             type: org.primefaces.component.column.Column@77d60f94
              +id: j_idt2039
               type: javax.faces.component.html.HtmlOutputText@15c3e76f
            +id: j_idt2040
             type: org.primefaces.component.column.Column@33e7b5a5
              +id: j_idt2041
               type: javax.faces.component.html.HtmlOutputText@42e9996b
            +id: j_idt2038  <===============
             type: org.primefaces.component.column.Column@584940d
              +id: j_idt2039
               type: javax.faces.component.html.HtmlOutputText@ff6e34d
            +id: j_idt2040
             type: org.primefaces.component.column.Column@1b2358b
              +id: j_idt2041
               type: javax.faces.component.html.HtmlOutputText@f9ed617
            +id: j_idt2038  <===============
             type: org.primefaces.component.column.Column@2e1b0f15
              +id: j_idt2039
               type: javax.faces.component.html.HtmlOutputText@164f7d10
            +id: j_idt2040
             type: org.primefaces.component.column.Column@6c10c3f
              +id: j_idt2041
               type: javax.faces.component.html.HtmlOutputText@7f165c9e
            +id: j_idt2038  <===============
             type: org.primefaces.component.column.Column@90397ce
              +id: j_idt2039
               type: javax.faces.component.html.HtmlOutputText@64c9cc03
            +id: j_idt2040
             type: org.primefaces.component.column.Column@6ae79451
              +id: j_idt2041
               type: javax.faces.component.html.HtmlOutputText@27d8d73f
          +id: j_idt2042
           type: org.primefaces.component.message.Message@327089ed
        +id: j_idt2043

and commenting out that fragment removes the problem. 
Not sure why this occurs yet and whether it is something wrong there or with what I'm updating with the commands.

Another thing that I think is going on with small files is that it takes some time for S3 to report that the file exists, so if it doesn't take a little time to calculate the md5 and for dataverse to check on it, there can be an error that causes the file not to show in the lower panel. I can look into adding a delay/retry on that.

For the main issue - if anyone has ideas...

landreev:
Tested this today - it's still working for me just fine. Will try it on the same box that @kcondon was using for testing. 
I'm more confused now than I was when I started testing though. @qqmyers - that issue with PrimeFaces complaining about the duplicate ids related to the autoComplete component - you mentioned you started seeing it intermittently... was there anything special at all you had to do to trigger the condition? I haven't been able to make it happen even once.

landreev:
OK, just realized, finally, that Kevin was using a different test bucket, from the one I was using. So that may explain why it wasn't working for him at all. 
This still doesn't address or explain the intermittent (?) problem with the autoComplete component. 
We've seen that problem with autoComplete component before - it appears to be kinda buggy. I'll need to check but we may have "fixed"/worked around it by randomly changing things on the page until it went away (sigh).

qqmyers:
@landreev - one thing I'm not sure I mentioned - it looks like this only occurs in create mode for me. Not sure if that's because the linking stuff doesn't appear otherwise or what. I'm not sure if its 100% of the time in create mode but it may be. 
I'm not sure what the fixes were earlier but I seem to recall parts of the page getting wrapped as ui:fragment as well as changes to what gets updated in the commands.
Also - I should have a fix for files not showing up - basically doing a retry to wait for S3 to respond correctly that the file exists. I'll commit that but it shouldn't affect the other issue.

landreev:
OK, we'll retest it tomorrow on dvn-build. I have a good guess as to what it is.
Sorry guys; I was not aware until today that it was not working for Kevin at all; I thought the branch was in a very good shape last week - it was working for me and I was able to demo it to other people here. So I didn't think of checking on it.
We'll figure it out tomorrow.

landreev:
> @landreev - one thing I'm not sure I mentioned - it looks like this only occurs in create mode for me. ...

Thanks, that probably explains it. There's a chance we are only using that autocomplete in create mode - to pick the parent dataverse for a new dataset (?)

qqmyers:
Another clue - I think it only happens when you upload multiple files. They are all done in parallel, so perhaps their partial updates are interfering in a way that doesn't happen if each is uploaded sequentially?

kcondon:
@qqmyers Updated list of known issues. I will update this list with anything new as it comes in:
[ ] Failure to upload multiple files on dataset create using direct S3, duplicate component ID
[x] Failure to upload files at all to direct S3. Resolved, needed CORS on S3 configured.
[ ] Stata tabular file type not detected, type unknown, therefore not ingested using direct S3
[ ] 1000 file upload to direct S3 fails, UI freezes, works on normal S3.

landreev:
To reiterate what I said during the standup meeting yesterday: if resolving that conflict with the p:autoComplete component in Create mode is at all problematic, I'm not sure we should be spending too much effort on it. (if nothing else, there's a chance this is something that will be fixed by the upcoming PrimeFaces 8 upgrade). 
So I'm proposing to just disable file upload on create, if direct upload is configured. (upload button replaced with a message "This dataverse is configured for direct upload; create the dataset without files, then upload away" ... or something like that)
This functionality is primarily for those who need to upload very large files, and presumably are willing to wait. The upload on create gives somebody who wants to create a compact dataset with few small files an option of doing it in one quick step. These don't appear to be overlapping demographics.

landreev:
I have added a paragraph on CORS to the configuration guide. Please edit for style/grammar if needed.

mheppler:
@landreev @qqmyers sorry if I'm slow to dig into the duplicate ID errors. It looks like these are the same errors reported in **JSF errors when adding a dataset: duplicated component IDs #5586**.

In that issue, I linked to this [Stack Overflow answer](https://stackoverflow.com/a/2101768) which includes an outline of potential root causes of these errors.

> Duplicate component ID errors may occur when:
> 
>   -  Same ID is used on different components inside the same `NamingContainer.`
>   -  Physically different components are bound to the same property of the same bean.
>   -  The `<f:subview>` is been declared in the include page instead of the parent page.
>   -  The same include page is included multiple times inside the same `NamingContainer`.
>   -  A component is been dynamically created without having an explicit ID assigned.

The PR #6025 from @sekmiller which closed that issue used render logic to hide the offending link dataset popup. Not sure if that fix has been undone in the branch or what.

Also, I have seen no evidence as of yet to suggest PrimeFaces 8 contains any magic cure all.

qqmyers:
@landreev @kcondon @mheppler - I've been trying to dig into this as well and have discovered that the key issue is having a binding= line in the autoComplete. Despite the fact that the autoComplete for dataset linking (which is the one causing problems - not the one to select a host dataverse at the top of the form) is not rendered at all in create mode, it still seems to be getting processed. Removing the binding= line allows the directupload to work. As far as I can tell, it breaks linking instead (on published datasets where that shows), but, since the other autoComplete doesn't use binding, I'm assuming there's a reasonable way to work around using it. Perhaps not, but I was going to look at that next (suggestions welcome). If that can't be done easily, I guess we can switch off direct upload in create (might just be able to allow normal upload during create and add a message as suggested, versus completely turning upload off).

qqmyers:
Actually, it doesn't look like removing binding causes any problems, so I've just removed it and related things. Let me know if this doesn't resolve the problem. I'll move on to looking at the many files/large files issues.

sekmiller:
Based on my check in notes the binding was put in to facilitate the display of success/failure  messages. My direct recollection of almost two years ago is a bit hazy.

sekmiller:
I agree with @qqmyers. It looks like removing the binding seems to have no ill effects. Success and error messages are displaying as expected

qqmyers:
Argh - regarding file size - it looks like Amazon has a 5GB limit on the size of a single upload. Not sure if it is configurable. Minio S3 doesn't and I continue to be able to upload > 5 GB files with Minio. I thought I tried against AWS at some point but evidently not with a large file, presigned, and not using the aws cli). 

The right way to handle larger uploads to AWS is to break the file into parts that are sent in separate http requests. That looks to be possible (using presignedURLs) but not well documented/supported at the moment - probably not something to hold up 4.20 for. 
I'm still looking into why large numbers of files fails. If there's an obvious fix for that I can add that, and I may be able to add some error checking to catch the errors from CORS/AWS size limit.
With those fixes, which I hope can be quick, maybe this can go as is as an 'experimental' or 'limited' feature for now and larger files at AWS can be a separate PR?

kcondon:
@qqmyers Thanks for the info and continuing to look into it. I can test using some lesser number of files such as 50 to see whether it behaves reasonably well. That might be enough for a first effort but would need to check in with @djbrooke

qqmyers:
I think the change I committed helps with larger numbers of files (I've tested ~100 so far). 

I also have a handle on the progress bar flashing - the progress bar code still assumes one file/changes all the bars at once.

I also noticed in testing that drag-n-drop doesn't work - must be a separate event to catch in the js to trigger processing.

So far I haven't been able to reproduce #6 above.

qqmyers:
OK - may be worth testing again. I'm not yet catching any failed http calls but assuming the calls succeed, it should now handle more files, with progress showing for ~4 at a  time, no flashing, drag and drop should work, and the long pause with the twirling icon at the end should be gone. Based on debugging with @kcondon , its possible that the root cause for #6 could also be fixed by the code I added a couple days ago to do a retry if Dataverse's first call to get the new file's metadata fails (e.g. some AWS server responds with a  404 before it learns about the new object).

If anything fails without an http error code showing for some call in the browser dev console, let me know (including #6)

kcondon:
@qqmyers Had mixed results when adding files to dataset on create with direct to s3:
first time failed to add files, ui froze at progress bar, no progress, error in logs
second time failed to add 3 of 4 files, exception in logs, multiple retries and failure to add file msgs
third time succeeded as expected.

qqmyers:
I just discovered a cut/paste error in the upload js this morning. That update is definitely needed.

PR #6498, opened 2020-01-09
log expired tokens

### LINKED ISSUE #6495: As a dataverse operator, I'd like API access from expired / invalid tokens to be logged so that it can be monitored
As more external tools and pipelines interoperate with Dataverse, some of these tools may require use of API tokens.  However, long-running integrations have the potential for the token used to expire; breaking the integration.  Logging use of expired API tokens would allow Dataverse operators to monitor these cases, to allow monitoring (and potentially fixing them) without requiring user error reports (or ideally without the users noticing).

pameyer:
Allow sysadmins to log access from expired API tokens.

Closes #6495 

`curl -i -H "X-Dataverse-key: $valid_token" -X GET "http://localhost:8084/api/dataverses/:root"` ; edit database to set expiration date to past, repeat curl call ; check system.log

no

no

**Additional documentation**:


kcondon:
@pameyer This works for an expired token but not an invalid one, which is the other case mentioned in your ticket.

pameyer:
@kcondon Good catch - I should've done a better job with the title.

PR #6513, opened 2020-01-13
only show email portion system email via Harvesting #3619

### LINKED ISSUE #3619: adminEmail field invalid in OAI-PMH Identify response
Unless the _:SystemEmail_ setting contains only a mailbox addr-spec (just the address, no spaces or angle brackets), the Dataverse 4 OAI-PMH provider will return an invalid value for the _adminEmail_ field, since that field should not contain a mailbox display name.

For example, the sample command in the [:SystemEmail section](http://guides.dataverse.org/en/4.6/installation/config.html#systememail) of the Configuration section of the Installation Guide:
> curl -X PUT -d "Support \<support@example.edu>" http://localhost:8080/api/admin/settings/:SystemEmail

would result in this error.

Ideally, the OAI-PMH provider code would check for a value ending in "\<.\*>" and, if present, emit only the string between the angle brackets (i.e., like to sed -e 's/^[^<]*<\\([^>]\*\\)>$/\1/ ').

pdurbin:

Avoids OAI-PMH validation errors like this:

65978606-fff89880-e441-11e9-9f58-5f19382163da

In my local testing, before making this pull request, I was seeing output like this...

`<adminEmail>Support2 &lt;support2@example.edu&gt;</adminEmail>`

... from http://localhost:8080/oai?verb=Identify

Closes #3619

If you do something like this...

curl -X PUT -d "Support2 <support2@example.edu>" http://localhost:8080/api/admin/settings/:SystemEmail

... and then stop and start Glassfish (required to see the change, apparently) and then look for "adminEmail in the output of http://localhost:8080/oai?verb=Identify (or equivalent for your server) you should only see the email portion (support2@example.edu) of :SystemEmail, without the "personal name" (Support2).

To test if Dataverse validates in terms of OAI-PMH, I'd suggest using http://oval.base-search.net where the screenshot above comes from.

No.

Probably not worth it.

**Additional documentation**:

I put a line in the guides, the section on setting up a harvesting server.

### DIFF
--- a/doc/sphinx-guides/source/admin/harvestserver.rst
+++ b/doc/sphinx-guides/source/admin/harvestserver.rst
@@ -22,6 +22,8 @@ You might consider adding your OAI-enabled production instance of Dataverse to
 `this shared list <https://docs.google.com/spreadsheets/d/12cxymvXCqP_kCsLKXQD32go79HBWZ1vU_tdG4kvP5S8/>`_
 of such instances.
 
+The email portion of :ref:`systemEmail` will be visible via OAI-PMH (from the "Identify" verb).
+
 How does it work? 
 -----------------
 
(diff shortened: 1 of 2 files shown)


kcondon:
This fixes the angle bracket in systememail issue. Test data should be: curl -X PUT -d "Support2 <support2@example.edu>" http://localhost:8080/api/admin/settings/:SystemEmail 

My test env does not validate using the specified service but that is outside the scope of this ticket:


Screen Shot 2020-01-15 at 1 45 18 PM

Issue #6230, opened 2019-09-30
Upgrade from Glassfish 4.1


scolapasta:
Dataverse is currently running on a (modified) Glassfish 4.1. This has served us well, but is old and no longer actively supported. Especially in light of the upcoming DataTags integration, we want to upgrade to a new app server for security uogrades, as well as being able to leverage newer features (e.g. Config API) and enable upgrades of other core technology (e.g. Java EE 7 -> Jakarta EE 8).

There are a few options to explore. Some work has already been done for Payara 5, and this tickets will serve for both  for the evaluation of that and some other candidates (TomEE, Glassfish 5, Wildfly), and the actual upgrade.

Note that I'm closing this old issue #2628 with a very similar title, as that is several years old now and was originally about upgrading to 4.1.1. 

Let's use this issue going forward for comments related to upgrading from glassfish 4.1.

(per request, also adding a link to the GH project that references subtasks related to this upgrade: https://github.com/orgs/IQSS/projects/16)

poikilotherm:
@scolapasta may I ask you to link to the GH project? We can rename it to be more technology neutral if it is too Payaraish (too much badass 🐟).

Looking forward to the next community call on this. :smile:

djbrooke:
Thanks for creating, @scolapasta! I'll put a "Large" on this and move it to the team dev column so that we can regularly discuss progress in standup.

pdurbin:
I'm excited to see the following candidates on the list above:

- Payara
- Glassfish 5
- TomEE
- Wildfly

I think we should also consider one more:

- Open Liberty

My reasoning for including Open Liberty is twofold.

First, if you go to https://jakarta.ee and follow the links from the announcement of Jakarta EE 8 to the list of compatible products, Open Liberty is a compatible implementation (Payara is not yet listed but I'm sure will be).

![Screen Shot 2019-09-30 at 4 35 25 PM](https://user-images.githubusercontent.com/21006/65915185-ef441600-e3a0-11e9-8377-9e38490146e6.png)
![Screen Shot 2019-09-30 at 4 35 31 PM](https://user-images.githubusercontent.com/21006/65915186-ef441600-e3a0-11e9-819a-9cca366c9905.png)
![Screen Shot 2019-09-30 at 4 35 49 PM](https://user-images.githubusercontent.com/21006/65915187-ef441600-e3a0-11e9-9d6e-7d9677b4e87e.png)

[...]

PR #6523, opened 2020-01-15
6230 changes for upcoming glassfish upgrade

### LINKED ISSUE #6472: Issues with validation on Dataverse Page
While working on the glassfish upgrade (specifically issue #6447), I noticed some issues with validation. However trying on our demo and production sites, they are not specific to the upgrade.

• Validation messages don't appear - this one is what #6447 is about. It's strange because I see the same behavior in production, but not in demo (i.e. demo works properly). This should be fixed by changing the update attribute from "@all" to "@form :messagePanel". I've merged this into 6230 branch, but we should consider making this change for develop so we can fix our production.
• (I would be curious to understand why it does work correctly in demo and have some others try it to see if demo is the only outlier)
• Dataverse name does not have a <p:message> tag so if you leave it blank you don't get a validation message.
• validation messages know show a lot of extra text (jsf component ids) at the beginning - not sure which version this got in, but checking Odum production, it was not an issue in 4.11. My guess is that it happened during the recent changes related to accessibility.

scolapasta:

These PR makes changes that allow the app to run with JSF 2.3 (specifically 2.3.14) and will allow us to upgrade our app server.

It is in progress for #6230, but we are also still looking for other issues.
It does close #6472 as I fixed it here before realizing it happens in current prod (and opened up this issue)

there are 3 fundamental changes:
• don't store ids (and alias for dataverse) from the viewParam in as the id (or alias) of an object, but have dedicated fields on the backing bean
• for viewparams that refer to fields that can change and used after partial submits, switch to omnifaces view params
• for converters don't use `@inject` or `@ejb` annotation

An overall smoke test could be good, but some specific things based on the changes:
• Initial load of dataverse and dataset page
• edit dataset button works
• converters, so basically things like role permissions autocomplete, facet selection, metadata selection

Note: the purpose of this PR is not to fix anything specifically, but to not introduce new bugs when deploying to a later app server; so, while if you find a bug, please verify if that bug already exists in the current code base.

No.

No. Though eventually there will be, when we upgrade app server, the purpose of this PR is to proactively get these changes in, but still run on glassfish 4.1.

**Additional documentation**:


kcondon:
This looks ok to merge, all functionality appears working.

PR #6564, opened 2020-01-23
6262 show proper number of ds inlinking dv

### LINKED ISSUE #6262: Linking Dataverses does not show proper number of Datasets in facet results
When you link an entire dataverse to another dataverse, it does not display the number of datasets in the linked dataverse, on the target dataverse page:
see https://dataverse.harvard.edu/dataverse/HarvardUniversity
Where I have linked several dataverses with multiple collections and the only dataset facet that shows (1) is for the single dataset I linked.

Merce wants it to display #datasets contained within the linked dataverses.

sekmiller:
This PR will display the datasets owned by a linked dataverse in the linking dataverse

Closes #6262 

None

Link a DV containing datasets and see that the datasets are displayed in the linking DV. 

no

added as 6262-release-notes.md

**Additional documentation**:
NA


scolapasta:
@sekmiller this doesn't handle when there are dataverse in a linked dataverses (nor datasets in those dataverses, etc). We would need some recursion for this, but is likely what is needed to truly resolve this issue.

djbrooke:
Moving back to team dev based on comment from @scolapasta above

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1280,8 +1320,53 @@ private String getTopicClassificationTermOrTermAndVocabulary(DatasetField topicC
...
+                List<Dataverse> linkingDVs = dvLinkingService.findLinkingDataverses(prior.getId());
+                if (!linkingDVs.isEmpty()){
+                    return true;
+                }
+            }
+        }       
+        return false;
+    }
+    
+    private List<String> findLinkingDataversePaths(Dataverse dataverse, Boolean self) {
```
scolapasta: why is this an object and not a primitive boolean?

PR #6572, opened 2020-01-24
5952 - tell admins to reexport following email privacy change

### LINKED ISSUE #5952: RE-Index All when setting/unsetting :ExcludeEmailFromExport
When setting ExcludeEmailFromExport - all cached "export" files should be re-indexed. The setting effects only the metadata for datasets that are created AFTER the setting change. 

Two choices here:

1. update the docs with a note that says after the command is run (and the setting changed), run the reIndexAll command: curl http://localhost:8080/api/admin/metadata/reExportAll   OR

1.  Edit the code to make the re-index happen when the setting is updated (on a change of false-to-true OR true-to-false)

djbrooke:

Changing the setting to expose/not expose emails addresses in exports doesn't make the change, a reexport is required. There was a suggestion to call this out in the docs and it seemed like a helpful addition.

Closes #5952 

I looked and I didn't see other DB settings or JVM options where we'd want admins to run a reexport, but if there are others, let me know and I'll add this there as well. 

Just make sure I didn't make any formatting errors!

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -1351,6 +1351,12 @@ Set ``:ExcludeEmailFromExport`` to prevent email addresses for contacts from bei
 
 ``curl -X PUT -d true http://localhost:8080/api/admin/settings/:ExcludeEmailFromExport``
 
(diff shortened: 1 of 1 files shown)


djbrooke:
@shlake you were the original requester here, so feel free to give this a code (doc) review if you'd like. Thanks for the suggestion!

PR #6577, opened 2020-01-24
6545 solr var meta

### LINKED ISSUE #6545: Solr search and variable level metadata
With Data Curation Tool it is possible to update variable level metadata, but presently, updates in variable label are not reflected in solr search. 
There are also new variable level fields such as literal question, interview instructions, post question, universe and notes that would be needed to add to solr schema and that solr search would be able to reflect updates in these fields.

lubitchv:
Variable level metadata that is updated using DCT  (variable label, literal question, interview instructions, post question, universe) will be searchable with solr. Also small bugs connected to datasetVerionId and DRAFT version with solr and FilePage were fixed. 

Closes #6545


djbrooke:
Hey @lubitchv - we'll review this soon. I'm planning to add some more detailed tooltips. Can you give me write access to your branch?

djbrooke:
@lubitchv nevermind, I already have access. Sorry about the false alarm! Commit incoming...

djbrooke:
@lubitchv I added tooltips based on DDI 2.5 here: 

https://ddialliance.org/explore-documentation

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -818,11 +818,11 @@ advanced.search.datasets.persistentId=Dataset Persistent ID
...
-search.datasets.literalquestion=Literal Question
-search.datasets.interviewinstructions=Interview Instructions
-search.datasets.postquestion=Post Question
-search.datasets.variableuniverse=Variable Universe
-search.datasets.variableNotes=Variable Notes
+search.datasets.literalquestion=Text of the actual, literal question asked.
+search.datasets.interviewinstructions=Specific instructions to the individual conducting an interview.
+search.datasets.postquestion=Text describing what occurs after the literal question has been asked.
+search.datasets.variableuniverse=The group of persons or other elements that are the object of research and to which any analytic results refer.
+search.datasets.variableNotes=For clarifying information/annotation regarding the parent element.
```
jggautier: The notes element can be used as a sub-element within many other elements. I think that's why [the description in the DDI documentation is so vague](https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/field_level_documentation.html). The parent element here would be the variable, right? If so, could we edit the tooltip to read something like this?:

> For clarifying information/annotation regarding the variable.

lubitchv:
@sekmiller But guest user would not get a draft file listed in the search anyway, so he will not be able to get this link anyway. But maybe you have a specific scenario in mind that I am missing.

Although for non guest user who will get listing of a draft file, after pressing the dataset link provided, he will end up with published version and not draft and this is confusing.

src/main/java/propertyFiles/Bundle.properties
djbrooke: Thanks, updated in https://github.com/IQSS/dataverse/pull/6577/commits/f00348b162840ccae8071bb7ecba2c99dc305a32

kcondon:
@lubitchv Hi, so I plan to test this by doing the following. Please let me know if I am missing anything. Also, please add a release note indicating a need to change shema.xml and whether or not a full reindex is needed.
To test:
1. Using DCT tool, modify above mentioned fields and verify they are searchable.
2. Create multiple dataset versions and modify file history too, confirm changes are visible in dataset and file version history.
3. Confirm a draft file url is not accessible by guest and is accessible if you have permission.
4. Determine steps to enable this feature: ie. new schema.xml, restart solr, reindex?

kcondon:
@lubitchv Thanks for the release note!

lubitchv:
@kcondon The plan looks good!

djbrooke:
@lubitchv, we unfortunately had to revert the changes related to indexing as it slowed indexing down to the point where it would no longer complete. This reversion was done in https://github.com/IQSS/dataverse/pull/6704 and @sekmiller should be able to provide more information if needed. We'd still like to get this to Dataverse, but it was a blocker for the 4.20 release, so we pulled it out for now. 

I'll plan to reopen #6545 and if there's anything you need from us please let us know.

Issue #6083, opened 2019-08-09
documentation: update Native API examples


donsizemore:
Related to #2174 and #930 and #1381 but focusing on a smaller chunk: the Native API examples in http://guides.dataverse.org/en/latest/api/native-api.html are sometimes complete, sometimes make assumptions of the user, and in a few cases contradict one another.

A rising graduate student here would like to learn more about Dataverse, Git, and contributing code to open source, so she's agreed to graciously test and update curl examples. Pull request forthcoming.

pdurbin:
@donsizemore fantastic! Please be aware that if she starts immediately, she's very likely to encourage merge conflicts with the improvements I'm hoping to make to the API Guide before my talk on Tuesday: https://groups.google.com/d/msg/dataverse-community/V5WkMGDS4VI/maxXTdmzDwAJ

It could be a good opportunity to learn about how to resolve merge conflicts, I suppose. 😄 

Also, once we know her GitHub username, we can add her to https://github.com/orgs/IQSS/teams/dataverse-readonly/members , assign this issue to her, and move it to the "Community Dev" column on our board at https://github.com/orgs/IQSS/projects/2 . This helps us remember what people are working on. 😄

pdurbin:
@donsizemore hi! Would you and/or the rising graduate student like to review pull request #6107 ? I rewrote a few of the curl commands in the Native API section and I hope you like the direction I'm taking. 😄 Feedback welcome!

donsizemore:
In the spirit of "small chunk" updates, I'm closing this in favor of tickets to accompany smaller batches of documentation changes, say updating the example syntax in the "Dataverses" section, then Datasets, and so on.

PR #6579, opened 2020-01-24
6083 documentation update native api examples

pkiraly:
**What this PR does / why we need it**: Improve the curl examples of the Dataverse related sections of the Native API documentation

**Which issue(s) this PR closes**: It does not close any partucular issue. There is a large issue which talks about the full page of the Native API: #6083 

**Special notes for your reviewer**: This pull request is quite format related, it does not revise fully the content, so later revisions of the sections from content perspective is needed. E.g. the dataverse ID is sometimes referred as $id and other times as $alias.

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -138,9 +138,9 @@ Show Contents of a Dataverse
 
   export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
   export SERVER_URL=https://demo.dataverse.org
-  export ALIAS=root
+  export ID=root
 
-  curl -H X-Dataverse-key:$API_TOKEN $SERVER_URL/api/dataverses/$ALIAS/contents
+  curl -H X-Dataverse-key:$API_TOKEN $SERVER_URL/api/dataverses/$ID/contents
 
 The fully expanded example above (without environment variables) looks like this:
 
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -108,10 +137,10 @@ Show Contents of a Dataverse
 .. code-block:: bash
 
   export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
-  export ALIAS=root
   export SERVER_URL=https://demo.dataverse.org
-  
-  curl -H X-Dataverse-key:$API_TOKEN $SERVER_URL/api/dataverses/$ALIAS/contents
+  export alias=root
```
mheppler: @pkiraly was there a reason to switch the variable formatting from uppercase to lowercase? I believe it would be more consistent with the other variables like $API_TOKEN, and would stand out more in the curl commands, if all lowercase variables were uppercase.

pkiraly: There was inconsistency in the variable case, e.g.:
```
curl -H X-Dataverse-key:$API_TOKEN $SERVER_URL/api/dataverses/$id
```
I left SERVER_URL and API_TOKEN uppercate to suggest they are constant within a script or API session, and made other lowercase if they were not already in lowercase.

Do you think all variables should be uppercase? It might affect other API pages as well.

scolapasta: Hi @pkiraly it is really just a matter of style and while there is some utility in differentiating between configuration variable vs. data variables, I think it'd be better for them all to standout - it makes it significantly more obvious that those are placeholders that need to be changed.

(I'm willing to reconsider if there is a known convention for having these be different, but I couldn't find any in my quick attempt to review online)

As you said, it will likely affect other API pages, and if you have time to make updates, that would be appreciated. If not, then we can just tackle incrementally as we make edits to those pages.

pkiraly: @mheppler, @scolapasta I changed them uppercase. I found one thing which also needs some conversation: the identifier of a dataverse should be ALIAS or ID? Mostly it is ID, but there are some ALIAS-es here and there.

scolapasta: We used to call it alias (and still do in db and code) but it later switched to thinking of it as identifier (as that was more clear to outsiders). So probably best to go with ID.

pkiraly: @scolapasta I replaced ALIAS with ID in the Dataverse related native API calls.

pkiraly:

PR #6580, opened 2020-01-27
Update native-api.rst

### LINKED ISSUE #6563: Guides - Missing quotation marks in curl command for creating dataset via API
Version: 4.18.1

Hello everybody,
After @youssefOuahalou brought up an error, @pdurbin found out that quotation marks are missing in the curl commands in [this section of the API Guide](http://guides.dataverse.org/en/4.18/api/native-api.html?highlight=already%20doi#create-dataset-command). Otherwise, bash cannot parse the URLs properly (and the ampersand (&) in the first of the two paths confuses it).

It should be:

`curl -H "X-Dataverse-key: [admin token]" -X POST "http://[dataverse URL]/datasets/:import?pid=[digital object identifier]&release=[yes or no]" --upload-file "[folder path]"`

BPeuch:
Quotation marks are necessary in this curl command, otherwise bash cannot parse the URLs properly (and the ampersand (&) in the first of the two paths especially confuses it).

Closes #6563

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -330,7 +330,7 @@ Next you need to figure out the alias or database id of the "parent" dataverse i
   export PARENT=root
   export SERVER_URL=https://demo.dataverse.org
 
-  curl -H X-Dataverse-key:$API_TOKEN -X POST $SERVER_URL/api/dataverses/$PARENT/datasets --upload-file dataset-finch1.json
+  curl -H X-Dataverse-key:$API_TOKEN -X POST "$SERVER_URL/api/dataverses/$PARENT/datasets" --upload-file dataset-finch1.json
 
 The fully expanded example above (without the environment variables) looks like this:
 
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -336,7 +336,7 @@ The fully expanded example above (without the environment variables) looks like 
 
 .. code-block:: bash
 
-  curl -H X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx -X POST https://demo.dataverse.org/api/dataverses/root/datasets --upload-file dataset-finch1.json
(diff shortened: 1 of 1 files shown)


pdurbin:
Whoops! I meant to say that I added this commit: 3835baa

Time for coffee. ☕️

BPeuch:
Thanks, @pdurbin!

landreev:
> **What this PR does / why we need it**:
> Quotation marks are necessary in this curl command, otherwise bash cannot parse the URLs properly (and the ampersand (&) in the first of the two paths especially confuses it).
> 
> **Which issue(s) this PR closes**:
> Closes #6563

@BPeuch 
@pdurbin 
I may be missing something, but this PR still doesn't make much sense to me. You are correct, yes, in that you can't have unquoted ampersands in URLs in shell commands. But in the actual changes you guys committed in this PR, there are NO ampersands in the URLs that you're putting the quotes around. 
So no, you DON'T need the double quotes in this example (from the diff):
```
curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X POST "https://demo.dataverse.org/api/dataverses/root/datasets" --upload-file "dataset-finch1.json"
```

On the other hand, I'm still seeing some curl examples in that guide with un-quoted URLs that DO have ampersands in them... -? 
Like this line, for example - which appears to be the line for which #6563 was originally opened - it looks like it hasn't been fixed: 

```
curl -H X-Dataverse-key:$API_TOKEN -X POST $SERVER_URL/api/dataverses/$DV_ALIAS/datasets/:import?pid=$PERSISTENT_IDENTIFIER&release=yes --upload-file dataset.json
```

BPeuch:
Hello, @landreev.

I am responsible for the confusion: I forgot to remove certain local variables from the example before posting it, and there happened to be an ampersand in one of them. I later erased the contents of the  variables (it was especially problematic for the admin token) but forgot to update my message elsewhere. Sorry about that.

The general idea still is that certain characters might need escaping and the guidelines should specify that in case they pop up in a query.

donsizemore:
@landreev @BPeuch the "Datasets" section was next on my list of examples through which to stroll, but there are pending pull requests on the page.

pdurbin:
Yeah, I also noticed that there are no `&` but I figured sending the message that it's often good to put URLs in quotes is probably a good thing. I think in the example @BPeuch or his colleague sent me, there were some extra query parameters and hence a `&` or two.

PR #6597, opened 2020-01-31
docs: add Renku to integrations list

### LINKED ISSUE #6592: Add Renku to list of integrations
At https://twitter.com/pmarrai/status/1222813558758572032 you can see that I just mentioned Renku in a talk...

Screen Shot 2020-01-30 at 10 17 25 AM

... but Renku isn't on the list at http://guides.dataverse.org/en/4.19/admin/integrations.html

We should update https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/admin/integrations.rst to add Renku. Pull requests welcome!

rokroskar:

Adds Renku to the list of integrations in the documentation. 

closes #6592

### DIFF
--- a/doc/sphinx-guides/source/admin/integrations.rst
+++ b/doc/sphinx-guides/source/admin/integrations.rst
@@ -19,9 +19,9 @@ If your researchers have data on Dropbox, you can make it easier for them to get
 Open Science Framework (OSF)
 ++++++++++++++++++++++++++++
 
-The Center for Open Science's Open Science Framework (OSF) is an open source software project that facilitates open collaboration in science research across the lifespan of a scientific project. 
+The Center for Open Science's Open Science Framework (OSF) is an open source software project that facilitates open collaboration in science research across the lifespan of a scientific project.
 
-For instructions on depositing data from OSF to your installation of Dataverse, your researchers can visit http://help.osf.io/m/addons/l/863978-connect-dataverse-to-a-project 
+For instructions on depositing data from OSF to your installation of Dataverse, your researchers can visit http://help.osf.io/m/addons/l/863978-connect-dataverse-to-a-project
 
 RSpace
 ++++++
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/admin/integrations.rst
```diff
@@ -95,6 +111,17 @@ Researchers can launch Jupyter Notebooks, RStudio, and other computational envir
 
 Institutions can self host BinderHub. Dataverse is one of the supported `repository providers <https://binderhub.readthedocs.io/en/latest/developer/repoproviders.html#supported-repoproviders>`_.
 
+Renku
++++++
+
+Researchers can import Dataverse datasets into their Renku projects via the
+command-line interface (CLI) by using the Dataverse DOI. See the `renku Dataset
```
djbrooke: A Dataverse installation can run handles as well, but I'm OK with this as is. We can revisit as we'll be working in this area quite a bit over the coming months.

rokroskar: actually I think what we do is either process the DOI or extract it from a dataverse URL. Are either of those possibly incompatible with how some installations might format URLs? Maybe @Panaetius can chime in about our implementation.

djbrooke: An example of an installation running handles is:

http://hdl.handle.net/11529/10016
https://data.cimmyt.org/dataset.xhtml?persistentId=hdl:11529/10016

rokroskar: right, I remember some discussion about that - I think at the moment we require a DOI for import. I also don't see an option to export JSON-LD on that installation?

rokroskar: so should I qualify the statement in the docs to be clearer about the current requirements?

djbrooke: Yeah, they are on 4.7.1 and it's been added since then (4.8.x)

djbrooke: Yes, it would be good to add some more details. Thanks!

Panaetius: There's 3 way that work, either just a plain DOI, or a DOI that's prefixed with `doi:` or an URL that has the following conditions:
- it contains a `persistentId` entry
- it has the dataverse version page at the `api/v1/info/version` path
 in which case we access `<domain>/api/v1/datasets/export/?persistentId=<persistentId>&exporter=schema.org` to get the data.

So short answer, handles also work. Long answer, I remember some dataverse instances that use handles being weird as they don't have the API enabled or don't use `persistentId`. But it's been a while since I did this.

In your examples, the first URL wouldn't work, the second should work (though the first just redirects to the second).

rokroskar: I tried both, neither seemed to work.

PR #6609, opened 2020-02-04
6524 dataset size api

### LINKED ISSUE #6524: Dataset Size API
We have an endpoint for getting Dataverse size, added in #5848/#5941: 

http://guides.dataverse.org/en/latest/api/native-api.html#report-the-data-file-size-of-a-dataverse

We should add an additional endpoint for getting dataset size that should (generally) follow the same logic as the dataverse API. The display of this on the dataset page (#6400) is still being designed and tested, but we should add the API now to ease UI development in the future and because this may be useful for systems integrating with Dataverse.

sekmiller:
This PR add API endpoints for getting the storage size of a dataset and getting the sum of size of the files available for download in a particular version of a dataset.

Closes #6524  

None

Note that the storage size api requires that the token passed has view unpublished. Likewise for the download size, if the version is :draft there must be a token passed with view unpublished permission. No token is required for a published version.

No

No. (could be included in "additional features/fixes."

**Additional documentation**:
added doc to native-apt.rst


src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -866,45 +866,67 @@ public void obtainPersistentIdentifiersForDatafiles(Dataset dataset) {
...
+    public long findStorageSize(Dataset dataset, boolean countCachedExtras) throws IOException {
+        return findStorageSize(dataset, countCachedExtras, "storage", null);
+    }
+  
     /**
      * Returns the total byte size of the files in this dataset 
      * 
      * @param dataset
      * @param countCachedExtras boolean indicating if the cached disposable extras should also be counted
+     * @param mode String indicating whether we are getting the result for storage (entire dataset) or download version based
```
pdurbin: It would be nice to document the accepted values for "mode".

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetStorageSizeCommand.java
```diff
@@ -0,0 +1,79 @@
...
+ * @author skraffmi
+ */
+@RequiredPermissions({})
+public class GetDatasetStorageSizeCommand extends AbstractCommand<Long> {
+
+    private static final Logger logger = Logger.getLogger(GetDataverseStorageSizeCommand.class.getCanonicalName());
+
+    private final Dataset dataset;
+    private final Boolean countCachedFiles;
+    private final String mode;
```
pdurbin: Instead of a String for "mode" and enum would be nice. More type safe that way. And auto documenting.

djbrooke:
@pdurbin if there are suggested changes let's not start QA yet. @sekmiller feel free to respond to the feedback above and let me know if you want me to edit any docs.

pdurbin:
@djbrooke that's fine but I want to re-emphasize that I don't see any changes as required. Again, they're just suggestions. 😄

djbrooke:
@pdurbin thanks, yeah, I 100% agree and understand, but I don't want @kcondon to start QA and have things shift under him if @sekmiller plans to make the changes.

sekmiller:
I addressed the coding suggestions that @pdurbin made, but I'm going to leave the doc as is for now since it is in line with what is already there for other api endpoints.

PR #6610, opened 2020-02-04
get FITS metadata extract working again #5919

### LINKED ISSUE #5919: FITS File Upload: no longer extracting metadata

pdurbin:

FITS file metadata extraction is an advertised feature of Dataverse...

- https://dataverse.org/software-features
- http://guides.dataverse.org/en/4.19/user/dataset-management.html#astronomy-fits

... but my guess it has been broken since 4.9 and dc6ff86

Closes #5919

This is a draft pull request (Update: now it's real) because while removing the call to "delete" on the temp file fixes the problem allowing FITS metadata to be extracted...

Screen Shot 2020-02-04 at 2 52 14 PM

... I'm sure we need to keep the "delete" in there. 2bc4690 is just the first commit to make it easy for me to have a conversation with @landreev to see what he recommends.

Update: I moved the delete method down. @landreev I had to keep it in the for loop to keep the variables in scope. I couldn't move it all the way down to the return like we talked about.

The screenshot above is from testing with perseus60.fits from https://dataverse.harvard.edu/file.xhtml?persistentId=doi:10.7910/DVN/NTT77F/LASN2S

No.

It might be nice to explain this feature has been restored.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
@@ -277,20 +277,6 @@ public class IngestServiceBean {
                     }
                 }
 
-                // ... and let's delete the main temp file:
-                // Commenting out this "delete" for code review because otherwise "extractMetadata"
-                // below (for FITS files) fails with "Could not open temp file".
-                // See https://github.com/IQSS/dataverse/issues/5919
-/*
-                try {
-                    logger.fine("Will attempt to delete the temp file " + tempLocationPath.toString());
-                    Files.delete(tempLocationPath);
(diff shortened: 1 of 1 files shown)


pdurbin:
@landreev I pushed the commit we talked about and updated the description above. Please note that to keep the variable in scope, I couldn't move the delete all the way to the bottom. It's still in a for loop.

PR #6612, opened 2020-02-05
6522 duplicate dvobjects

### LINKED ISSUE #6522: DB Constraint to prevent duplicate DvObjects for the same physical file
Another issue with files discovered in prod. in recent days: 
There appears to be a number of instances in the db, where TWO dvobjects/datafiles are associated with the same storage identifier (i.e., same physical file).
I have not yet identified the exact scenario that makes it possible. But it must be something going wrong during upload and save. 
Opening the issue to diagnose and fix the problem. 
As always with things like this, this will involve both code fixes and cleanup.

landreev:
This addresses the issue of duplicate datafiles/dvobjects created on upload under some (unknown) conditions, referencing the same physical file. The code change adds the unique constraint on the storageidentifier within dataset that should prevent this from happening in the future. 

Before it can be deployed however, the database must be checked for any existing duplicates. We've already done it in our prod., but we'll have to notify other installations and guide them through the cleanup process.

In addition to the normal release note, the PR contains a "pre-release-note", plus the diagnostic script, for sending out to the partner installation. This should be done *ahead* of the next release, as soon as the branch is merged into develop. 

Closes #6522

Up for review is the approach I chose for dealing with this on remote installations. Specifically, I concluded it was not practical to try and produce a script that would fix any potentially affected installation. There are too many variables involved - whether an affected file is published vs. draft, ingested as tabular, physical file intact, etc. etc. So my approach is to send out the diagnostics script (that has been tested on one remote installation already) that will detect if this condition is present. If any duplicates are found, I'm asking them to send us the output and then we'll assist them with cleanup. 

This involves committing to more support work down the road, but I feel like this is the best and safest way to address this. 

 

We do not know of a way to reproduce the condition that leads to the creation of duplicates on upload. So there is no clear way to test it. 

Mostly the QA should be about reviewing the diagnostic script and information note that we'll be sending out to the remote installations. 

No
Yes
**Additional documentation**:


kcondon:
@landreev The script url in the release notes results in 404: https://github.com/IQSS/dataverse/raw/develop/scripts/issues/6522/find_duplicates.sh

but this one in another file works:

landreev:
@kcondon 
The flyway issue is a conflict with the migration script in the multistore branch, that's using the same number  and has been deployed on build - 4.19.0.1. I didn't realize that branch had a flyway script. 
I'll rename the script in my branch quickly (will be 4.19.0.2). 

The link is just a matter of changing one of them to "develop" right before we merge, as discussed. 


> The script url in the release notes results in 404: https://github.com/IQSS/dataverse/raw/develop/scripts/issues/6522/find_duplicates.sh
> 
> but this one in prerelease notes works: https://github.com/IQSS/dataverse/raw/6522-duplicate-dvobjects/scripts/issues/6522/find_duplicates.sh
> 
> Also getting a flyway failure due to checksum. That usually happens when there is already a script applied? Does this script need to be renamed for 4.20?

landreev:
@kcondon checked in

landreev:
Doh - we may need to revert this PR, on account of harvested files. Those often have remote urls in the storageidentifier field, that are totally not guaranteed to be unique... Our own prod. database has a ton of these, so the constraint doesn't work. 

We don't really use these urls for any practical purposes; so we can probably stop shoving them into the field... But we can't proceed with adding the constraint until we figure it out.

PR #6619, opened 2020-02-06
remove parentheses from dataset contact affiliation #6570

### LINKED ISSUE #6570: New contact affiliation metadata in Search API includes parentheses
In Dataverse 4.19, when Search API results include datasets with Contact Affiliation metadata, the affiliation value is wrapped in parenthesis:

Screen Shot 2020-01-23 at 8 23 58 PM

Was this by design? It reminds me of https://github.com/IQSS/dataverse/issues/5521

pdurbin:

Closes #6570

I copied and pasted a similar method above. Here's where the similar method is now used (one fewer place in this pull request since the Search API uses the new method):

Screen Shot 2020-02-06 at 10 54 23 AM

Add an affiliation to the contact of a dataset and inspect the JSON output from the Search API when finding that dataset. The value should not be wrapped in parentheses, like it is in 4.19.

No.

Yes, added. Technically this is a backward incompatible change but it's a small one and I don't think it's worth adding a configuration option for it. Sorry, API users. You no longer need to strip parentheses from affiliation if you were before. 

**Additional documentation**:

None.

### DIFF
--- a/doc/release-notes/6570-search-api-contact-affiliation.md
+++ b/doc/release-notes/6570-search-api-contact-affiliation.md
@@ -1,0 +1,1 @@
+As reported in https://github.com/IQSS/dataverse/issues/6570 the affiliation for dataset contacts has been wrapped in parentheses in the JSON output from the Search API. These parentheses have now been removed. This is a backward incompatible change but it is hoped that it won't cause any problems for integrations and API users.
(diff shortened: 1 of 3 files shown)


pdurbin:
@sekmiller I just reduced code duplication in e3a25fa

PR #6621, opened 2020-02-06
change DDI schemaLocation to https #6553

### LINKED ISSUE #6553: In DDI export, use https for the DDI schemaLocation URL
While checking for validation issues with the DDI XML that Dataverse exports, the tool I use (https://www.freeformatter.com/xml-validator-xsd.html) reported that the XML failed validation because "White spaces are required between publicId and systemId.: Line = 1, Column = 50". Some Googling led me to [this GitHub issue comment](https://github.com/openpreserve/jhove/issues/227#issuecomment-303433586) about the error message where the solution was to change the http in the URL in the schemaLocation attribute (within the XML's `<codebook>` element)...

```
<codeBook
	xmlns="ddi:codebook:2_5"
	xsi:schemaLocation="ddi:codebook:2_5 http://www.ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd" version="2.5">
```

... to https.

In any dataset's DDI XML export, when I changed the http in http://www.ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd to https, the validation tool doesn't report that "white spaces" error (and is able to continue reporting other errors). Maybe the DDI Alliance changed the URL to use https recently? So the DDI Codebook metadata that Dataverse exports should, too, so it's pointing to the right schema URL.

pdurbin:

We want validation tools like https://www.freeformatter.com/xml-validator-xsd.html to demonstrate to us that DDI exports are valid.

Closes #6553

I did the bare minimum of adding an "s" to "http" but I noticed (with `curl -i`) that redirects are in place that land on this URL with no "www": https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd . So maybe we should use that instead?

I also noticed a couple other places in the code (DDIExporter.java and OAI_DDIExporter.java) where the vanilla "http" URL exists. I left a comment on these so they can be easily seen in this pull request.

I was glad to see that a test broke when I changed the URL. This tells us that the code is being exercised. I updated the file used in the test code to match the new URL.

Create a new dataset, publish it, export as DDI and look at the the `<codeBook xsi:schemaLocation=` line.

I did not try using https://www.freeformatter.com/xml-validator-xsd.html but that would be good too.

No.

No. I doubt this will break any integration or API user's code.

**Additional documentation**:

None.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/export/DDIExporter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/DDIExporter.java
@@ -13,6 +13,8 @@ import javax.xml.stream.XMLStreamWriter;
 import javax.xml.stream.XMLOutputFactory;
 
 /**
+ * This exporter is for the "full" DDI, that includes the file-level,
+ * <data> and <var> metadata.
  *
  * @author Leonid Andreev
  * (based on the original DDIExporter by
@@ -21,15 +23,10 @@ import javax.xml.stream.XMLOutputFactory;
  */
 @AutoService(Exporter.class)
 public class DDIExporter implements Exporter {
-    // TODO: 
-    // move these into the ddi export utility
-    private static String DEFAULT_XML_NAMESPACE = "ddi:codebook:2_5"; 
-    // This redirects to https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd
-    private static String DEFAULT_XML_SCHEMALOCATION = "http://www.ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd";
-    private static String DEFAULT_XML_VERSION = "2.5";
+    public static String DEFAULT_XML_NAMESPACE = "ddi:codebook:2_5";
+    public static String DEFAULT_XML_SCHEMALOCATION = "https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd";
+    public static String DEFAULT_XML_VERSION = "2.5";
     
-    // This exporter is for the "full" DDI, that includes the file-level, 
-    // <data> and <var> metadata.
     @Override
     public String getProviderName() {
         return "ddi";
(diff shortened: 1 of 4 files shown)


jggautier:
Looks good to me!

I deployed branch 6553-https to AWS, http://ec2-54-242-96-185.compute-1.amazonaws.com:8080, created a dataset, clicked on the Metadata tab and the "Export Metadata" dropdown to check the DDI export, saw in the XML that the DDI schemaLocation uses https, copied the XML into that online validator and saw that it doesn't complain about how "White Spaces Are Required Between PublicId And SystemId., Line '1', Column '50'.". (As expected, it complains about known validation errors.)

scolapasta:
@pdurbin the bare minimum does fix the validation, but I think a little more can be done here (still a small) to a) change the other two to https or b) better yet have all three locations use one constant overall. (it feels like a bigger change would be to have OAI_ddi_exporter extend ddi_exporter, for example).

Can you take a look?

pdurbin:
@scolapasta ok. And what about the "www"? Should I remove it?

scolapasta:
Sure, I'd say while you're there, makes sense to.

pdurbin:
@scolapasta @jggautier ok, in 86c385d I removed the "www" and did some code consolidation.

scolapasta:
Looks good! I still think there's room for cleanup (removing the util and have one exporter extend the other), but not in scope for this "small".

kcondon:
OK confirmed xsd link now uses https. Validator runs with errors but they are known and addressed in other tickets per Julian:

Cvc-enumeration-valid: Value 'DVN' Is Not Facet-valid With Respect To Enumeration '[archive, Producer]'. It Must Be A Value From The Enumeration., Line '12', Column '23'.
Cvc-attribute.3: The Value 'DVN' Of Attribute 'source' On Element 'verStmt' Is Not Valid With Respect To Its Type, '#AnonType_sourceGLOBALS'., Line '12', Column '23'.
Cvc-complex-type.2.4.a: Invalid Content Was Found Starting With Element 'contact'. One Of '{"ddi:codebook:2_5":sumDscr, "ddi:codebook:2_5":qualityStatement, "ddi:codebook:2_5":notes, "ddi:codebook:2_5":exPostEvaluation}' Is Expected., Line '41', Column '71'.
Cvc-complex-type.2.4.a: Invalid Content Was Found Starting With Element 'useStmt'. One Of '{"ddi:codebook:2_5":method, "ddi:codebook:2_5":dataAccs, "ddi:codebook:2_5":othrStdyMat, "ddi:codebook:2_5":notes}' Is Expected., Line '49', Column '10'.
Cvc-complex-type.2.3: Element 'useStmt' Cannot Have Character [children], Because The Type's Content Type Is Element-only., Line '49', Column '30'.

PR #6622, opened 2020-02-06
6290 dataset role mgmt api

### LINKED ISSUE #6290: Dataset Role Management API
As Dataverse user, I would like to assign a specific role/user to a specific dataset using the Native Api. As we can do with Dataverse. I didn't found it in the documentation.

Thanks in advance;
Flavio.

sekmiller:
Allows admins to assign roles to users at the dataset level

Closes #6290 Dataset Role Management API

None

None
No

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1084,24 +1086,38 @@ public Response getLinks(@PathParam("id") String idSupplied ) {
...
+                    if (aRole.getAlias().equals(ra.getRole())) {
+                        theRole = aRole;
+                        break;
+                    }
+                }
+                dv = dv.getOwner();
             }
-            DataverseRole theRole = rolesSvc.findBuiltinRoleByAlias("admin");
+            if (theRole == null) {
+                return error(Status.BAD_REQUEST, "Can't find role named '" + ra.getRole() + "' in dataverse " + dataset.getOwner());
```
pdurbin: I think the new-ish rule is that when we touch APIs we should put the English in a bundle.

sekmiller: Added revoke and api messages to the Bundle.

kcondon:
@sekmiller Need GET after -X in list assignments example.

kcondon:
@sekmiller Discussed with @scolapasta and @djbrooke , please add inherited as well as directly assigned roles and indicate which is which.

djbrooke:
Thanks @kcondon for the discussion and @sekmiller for taking another pass at this!

PR #6441, opened 2019-12-06
6300/6396 additional dataset metadata via search api

sekmiller:
**What this PR does / why we need it**: Adds requested dataset information on requests via the Search API. Does not effect searches used by Search Include Fragment (loading the home page, etc.)

**Which issue(s) this PR closes**: 

Closes #6300 and closes #6396 

**Special notes for your reviewer**:  I ended up not adding indexing for this because the fields requested (create date, update date, data source, related publications, etc. aren't really useful as indexes. I made sure that the DB hits for collection the additional info only occur for retrieval of datasets via the SEARCH API.

**Suggestions on how to test this**: Add additional data to a dataset and see that it is returned via the SEARCH API. Run a search with that will return a draft dataset (while including a key that will give you access to that draft) and see that the version state is DRAFT and the published version is also returned if applicable.

**Does this PR introduce a user interface change?**: No.

**Is there a release notes update needed for this change?**: Not really. Could be mentioned as an enhancement

**Additional documentation**:

PR #6623, opened 2020-02-06
in Search API show fileCount for datasets #6601

### LINKED ISSUE #6601: Dataset file count in search results from API
Hi!

We have been recently working with search results and also retrieving dataset information. 

We are wondering whether it would be possible to get the file count for a dataset as part of the information being returned in each dataset information search result item.

We would find it very handy for our application

**Why do we need it?**
When showing dataset information in our application, we want to display the number of files that a dataset has as part of the search result. Currently, the only way to do this operation is by retrieving dataset information and calculating the files' array length.

**Why would this be useful for other people?**
Taking into consideration that people might be performing searches looking for dataset files, it would be handy to know whether a dataset contains files.

Thanks! 🚀

pdurbin:

Search API users have requested counts of files in datasets.

Closes #6601

Despite the noise I made in the issue about reindexing, I went for a database hit, nestled in several lines of other database hits. We already have the datasetVersion in our hands. It seemed pretty quick with my local testing, admittedly only a few files.

Look for "fileCount" in the response for datasets. It might be nice to test a dataset with many files to see if it's slow. A zero is returned, as expected, with no files.

No.

I didn't bother since it seems like a small feature. I'm reminded that some day we might want to keep an API change log. See also http://guides.dataverse.org/en/4.19/api/faq.html#is-there-a-changelog-of-api-functionality-that-has-been-added-over-time

**Additional documentation**:

None. I did update the JSON examples in the Search API docs to show "fileCount".

### DIFF
--- a/doc/sphinx-guides/source/api/search.rst
+++ b/doc/sphinx-guides/source/api/search.rst
@@ -116,6 +116,7 @@ https://demo.dataverse.org/api/search?q=trees
                        "Astronomy and Astrophysics",
                        "Other"
                     ],
+                    "fileCount":3,
                     "versionId":1260,
                     "versionState":"RELEASED",
                     "majorVersion":3,
@@ -291,6 +292,7 @@ The above example ``fq=publicationStatus:Published`` retrieves only "RELEASED" v
                     "subjects": [
                         "Medicine, Health and Life Sciences"
                     ],
+                    "fileCount":6,
                     "versionId": 53001,
                     "versionState": "RELEASED",
                     "majorVersion": 1,
@@ -323,6 +325,7 @@ The above example ``fq=publicationStatus:Published`` retrieves only "RELEASED" v
                     "subjects": [
                         "Medicine, Health and Life Sciences"
                     ],
+                    "fileCount":9,
                     "versionId": 53444,
                     "versionState": "RELEASED",
                     "majorVersion": 1,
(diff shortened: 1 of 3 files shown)


pdurbin:
> Are we always guaranteed to have the dataset/datasetversion db objects instantiated?

Yes, we are. Here's how we talk about this in https://github.com/IQSS/dataverse/releases/tag/v4.19

'The boolean parameter query_entities has been removed from the Search API. The former "true" behavior of "whether entities are queried via direct database calls (for developer use)" is now always true.'

This change was introduced in pull request #6441

landreev:
OK, I remember it now. 
We do have some hacky/optimized methods for instantiating entities with custom queries, instead of relying on the expensive .find() methods (we use them to populate the information in the search cards on the dataverse page, for example). We may consider using them for the Search API as well... but it may require some research, or at least testing. 
Definitely outside the scope for this.
Thanks. 

> Yes, we are. Here's how we talk about this in https://github.com/IQSS/dataverse/releases/tag/v4.19
> 
> 'The boolean parameter query_entities has been removed from the Search API. The former "true" behavior of "whether entities are queried via direct database calls (for developer use)" is now always true.'
> 
> This change was introduced in pull request #6441

pdurbin:
> I meant to ask about the decision not to index the number.

Probably we should have indexed the number (of files, `fileCount`). We now have an open issue about this:

- #8941

PR #6625, opened 2020-02-07
allow file preview and explore tools for Private URL users #6537

### LINKED ISSUE #6537: Support File Previews/Explore Tools for Private URL Users
People viewing a draft dataset using the Private URL feature should be able to use the same explore tools and see the same previews as regular users. As peer reviewers take advantage of this feature, it will be helpful to use all of the tooling. 

To reproduce the issue:

- Create a Private URL for your dataset and copy it to your clipboard
- Add a .pdf file (or other previewable file)
- Sign out of Dataverse and click on the private URL
- Go to the file page and see that the preview is not available
- Click the "Explore" button on the file page and be taken to a window that shows an error message in the previewer

If we look at the URL in the window from launching "Explore" we can see that no token is passed. We get something like this:

- https://qualitativedatarepository.github.io/dataverse-previewers/previewers/PDFPreview.html?fileid=xxxx&siteUrl=https://dataverse.harvard.edu&datasetid=xxx&datasetversion=:draft

Instead of something like this which has the private URL user's key:

https://qualitativedatarepository.github.io/dataverse-previewers/previewers/PDFPreview.html?fileid=xxx&siteUrl=https://dataverse.harvard.edu&key=xxx&datasetid=xxx&datasetversion=:draft

pdurbin:

It's valuable for reviewers who are accessing datasets via a Private URL to be able to preview files.

Closes #6537

I removed a check for `FilePage.fileMetadata.datasetVersion.isPublished()`

Regression test permissions for preview when for non-Private URL users.

No. The Preview tab is unchanged. It's just now always available for Private URL users.

I didn't add one but I'm happy to.

**Additional documentation**:

I didn't write any docs. I can.

### DIFF
--- a/doc/sphinx-guides/source/user/dataset-management.rst
+++ b/doc/sphinx-guides/source/user/dataset-management.rst
@@ -150,7 +150,7 @@ Certain file types in Dataverse are supported by additional functionality, which
 File Previews
 -------------
 
-Installations of Dataverse can install previewers for common file types uploaded by their research communities. The previews appear on the file page. If a preview tool for a specific file type is available, the preview will be created and will display automatically. File previews are not available for restricted files.
+Installations of Dataverse can install previewers for common file types uploaded by their research communities. The previews appear on the file page. If a preview tool for a specific file type is available, the preview will be created and will display automatically. File previews are not available for restricted files unless they are being accessed using a Private URL. See also :ref:`privateurl`.
 
 Tabular Data Files
 ------------------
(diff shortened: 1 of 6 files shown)


djbrooke:
@pdurbin does the PR cover explore as well, or just preview?

pdurbin:
@djbrooke I was pretty focused on just preview. I was looking at the "To reproduce the issue" steps you put in the issue.

djbrooke:
@pdurbin OK, preview tools are important since they included on page load, but we should handle this for explore tools as well. Noted about the steps to reproduce. I'll move this back to team dev for now. Thanks for taking another look when you have a capacity

pdurbin:
@djbrooke ok as of f28dbe1 the Private URL token is passed to external tools at the dataset level.

Also, in 132f2d8 I explained that restricted data can be previewed with a Private URL.

As I mentioned, I did previously test with the file level PNG explore tool. I'll update the title of this pull request.

djbrooke:
@pdurbin perfect, thanks!

PR #6628, opened 2020-02-07
added PIDs for datasets and files to guestbook download

### LINKED ISSUE #6534: Add dataset and file PIDs to guestbook csv download
In all Dataverse 4.x versions, the guestbook csv file that dataverse owners can download identifies the datasets by dataset title:

Screen Shot 2020-01-16 at 3 21 17 PM

I recently helped someone who needed to analyze the data in a dataverse guestbook, and the instability of the dataset titles, the only way that the dataset is identified, became a problem.
- In the guestbook csv, commas are removed from dataset titles (https://github.com/IQSS/dataverse/issues/3449)
- The title shown is always from the first published version of the dataset, so for example, if the dataset title "Japan 1954" in the screenshot above was to change in a later published dataset version to "Replication data for: Japan 1954", the guestbook will still identify the dataset with "Japan 1954".

This instability makes it difficult to work with this data (like combining it with data from other sources).

Since the dataset DOI should be more stable, could that be added to the guestbook?

landreev:
Self-explanatory!

Closes #6534

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
@@ -9,6 +9,7 @@ import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
 import edu.harvard.iq.dataverse.authorization.users.User;
 import edu.harvard.iq.dataverse.externaltools.ExternalTool;
 import edu.harvard.iq.dataverse.util.BundleUtil;
+import edu.harvard.iq.dataverse.util.StringUtil;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
```diff
@@ -226,6 +238,20 @@ public void streamResponsesByDataverseIdAndGuestbookId(OutputStream out, Long da
...
+        if (StringUtil.nonEmpty(protocol) && StringUtil.nonEmpty(authority) && StringUtil.nonEmpty(identifier)) {
+            StringBuilder sb = new StringBuilder();
+            sb.append(protocol);
+            sb.append(":");
+            sb.append(authority);
+            sb.append("/");
+            sb.append(identifier);
+            return sb.toString();
+        } 
+        return "N/A";
```
pdurbin: I'm fine with "N/A" here if there is no PID. Heads up to @jggautier

jggautier: Thanks. I have no preference. Another conventional option is to leave it blank, right?

In the case where some files in a Dataverse repository have file PIDs and some files don't (which I don't think is the case for any Dataverse repository yet), would using "N/A" be better than leaving it blank? I'm trying to remember the details of another guestbook-related conversation we had with a repository admin who said she preferred using a value instead of leaving it blank because it was a clearer to her that a record was written and, i.e. that the code worked.

pdurbin: @jggautier if you're fine with it, so am I. Thanks.

pdurbin:
@landreev heads up that as I moved this to QA I added a released note about the new columns: d2fcfb9d5

Issue #4938, opened 2018-08-08
Solr search result ordering broken


matthew-a-dunlap:
After upgrading solr, the configuration we had to boost search matches on certain Dataverse fields is broken. 

Based upon our configuration, the boosting should be in this order:

```
dvName^170
dvSubject^160
dvDescription^150
dvAffiliation^140
title^130
subject^120
keyword^110
topicClassValue^100
dsDescriptionValue^90
authorName^80
authorAffiliation^70
publicationCitation^60
producerName^50
fileName^40
fileDescription^30
variableLabel^20
variableName^10
text^1.0
```

Yet the result is off (note that this screenshot does not contain all the fields boosted)
![screen shot 2018-08-08 at 4 55 33 pm](https://user-images.githubusercontent.com/8901015/43865418-8c645176-9b30-11e8-8731-ed73a18ec226.png)

There is more information (production examples, etc) captured in https://github.com/IQSS/dataverse/issues/4836 , along with information on solr highlighting which has been fixed.

matthew-a-dunlap:
For whoever takes on this story: I have some best practice solr fixes that were not committed as part of the highlighting fix and should be included with this story. They are captured in https://github.com/IQSS/dataverse/tree/4836-solr-highlight-boost

matthew-a-dunlap:
After spending a while poking around in this I get the feeling that our results are acting more or less the same as before. I found http://beta.dataverse.org/ which is still running 4.8.2 (pre solr upgrade) and have been trying some queries on it. For example, when searching `test`, the datasets with `test` in the title are not inherently higher than those with `test` in other fields.

![screen shot 2018-09-19 at 6 59 20 pm](https://user-images.githubusercontent.com/8901015/45786445-76af6300-bc3e-11e8-80cc-45f031b5d863.png)

I've started reading up on this more and I think we have just misunderstood how boosting works. [This article](https://opensourceconnections.com/blog/2013/07/02/getting-dissed-by-dismax-why-your-incorrect-assumptions-about-dismax-are-hurting-search-relevancy/) has been helpful (though I haven't read it thru completely yet). It is also possible that the boosting was extra broken before the highlighting fix and I didn't notice that change.

I plan to read up on this more tomorrow and hope to have a better understanding of what we can actually do with boosting make results fit our needs.

matthew-a-dunlap:
[...]

PR #6631, opened 2020-02-10
6633 update solr 772

### LINKED ISSUE #6599: Clarification regarding supported Solr version(s)
Although solr 7.3.1  is used in the install example (http://guides.dataverse.org/en/4.19/installation/prerequisites.html#solr), there is no clarity as to exactly what versions are supported/tested.

Can i use the latest 7.x release, since 7.3.1 is now quite old?

What about the latest stable (8.x)?

Thanks in advance!

### LINKED ISSUE #6633: Upgrade from Solr 7.3.1 to 7.7.2
We're using Solr 7.3.1 but Solr 7.7.2 is out. We should upgrade.

Here's the changelog: https://lucene.apache.org/solr/7_7_2/changes/Changes.html

poikilotherm:
We should update to the latest supported version of Solr in the 7.x release train to be on an upstream supported release.

According to [this list](https://www.cvedetails.com/vulnerability-list/vendor_id-45/product_id-18263/Apache-Solr.html), there have been no minor or major security problems.

Closes #6633 

**2020-02-10: this is a WIP**
Basically, I just did a grep and sed to replace any "7.3.1" with "7.7.2":
```
rg "7\.3\.1" --files-with-matches | xargs sed -i '' -e 's/7\.3\.1/7.7.2/g'
mv conf/solr/7.3.1 conf/solr/7.7.2
```
Do we need to add more notes, add docs, ...?

We should discuss about the deprecation warnings (see #6599) and maybe get rid of 'em to prepare for the future. If we change that, we will need to check about reindexing advice for RLN.

We still need to figure out how to test all of this.

Components affected:
- SolrJ updated, so new WAR files needed with next release (4.20)
- Solr updated :laughing: (see also my note about deprecations above)
- Docker AIO -> @pameyer 
- Jenkins Integration Tests -> maybe @donsizemore 
- Vagrant -> @pdurbin 
- Installer and release Makefile -> @scolapasta and @landreev 
- Boosting rules (remember #4938 from last time)

Nope. Nada.

Definitely. Still needs to be take care of. A first draft is included in the commits.

**Additional documentation**:
None so far.


pdurbin:
> We should discuss about the deprecation warnings (see #6599)

How frequent are these warnings? Only on startup? Or are they emitted constantly and there's a risk of filling up a disk?

poikilotherm:
I only see them at startup, when config is parsed.

djbrooke:
@poikilotherm release note looks good, thanks for adding! I think the last time we updated Solr we had some issues with our boosting rules no longer working as expected (#4938), so I added a note in the testing area above. 

@scolapasta @pdurbin I like the upgrade here but I have some questions about what this means for the supported versions question in the original issue, I can ask post standup.

pdurbin:
At standup this morning we decided to create a new issue dedicated to the the upgrade itself and change the "closes" to that new issue. Here it is: #6633 

Previously, pull request was closing #6599 (Clarification regarding supported Solr version(s)) but we don't have a solution for this yet and want to treat that as a separate small chunk.

kcondon:
Quick q: has anyone in Dev reviewed the Solr release notes to see whether there are any obvious points of concern prior to testing?

pdurbin:
@kcondon I haven't. @poikilotherm , have you ?

kcondon:
@pdurbin I guess it was really a request if it hasn't been done. Thanks!

poikilotherm:
I have. I couldn't find anything relevant for us, but this should be verified in-depth by someone else. 4 eyes see more than 2.

pdurbin:
I just looked at the "upgrade notes" at https://lucene.apache.org/solr/7_7_2/changes/Changes.html for all the versions (screenshot below) and I don't see anything I'm concerned about. We don't use fancy features in Solr. I don't think any of these upgrade notes pertain to our use of Solr.

lucene apache org_solr_7_7_2_changes_Changes html

kcondon:
@poikilotherm @pdurbin Thanks for checking!

donsizemore:
I recompiled v4.19 with @poikilotherm's change to pom.xml, deployed it to https://payara5.odum.unc.edu and upgraded Solr to 7.7.2 on that host, FWIW.

Only entry in server.log so far:

`The web application [unknown] registered the JDBC driver [org.apache.solr.client.solrj.io.sql.DriverImpl] but failed to unregister it when the web application was stopped. To prevent a memory leak, the JDBC Driver has been forcibly unregistered.`

poikilotherm:
I see the same error message with current SolrJ 7.3.1, so no blocker :phew:

kcondon:
@poikilotherm Quick q: would it not be simpler to just install the new version rather than upgrade?

poikilotherm:
That might depend on how much time it takes to reindex and how much downtime is ok for your installation. I have no experience with big installations with 100.000+ datasets. Maybe @4thikonov can share experience about his 50.000 dataset import to give us an idea?

poikilotherm:
Ad 1) I have no idea if SolrJ 7.7 is compatible with Solr 7.3. I even have no idea where to look for such a statement. IMHO it might be better to upgrade, as our old SolrJ version has a CVE. We could also test this, but upgrading is always better IMHO...

Ad 2) We are only doing minor upgrades. Index is compatible from what I see in the release docs and upgrade notes. 

So no need to reindex as long as we don't change the schema config or a reindex becomes necessary due to changes in a metadata block like citation.tsv.

If you're asking the paranoid me, I'd do an update test with sample data...

Ad 3) I'm pretty sure that neither Dataverse nor Solr supports partial reindexes. An inplace reindex should be fine, no need to drop and rebuild. That should be much faster too, shouldn't it?

pdurbin:
@poikilotherm heads up that I tweaked your release note at b1fb21d. @kcondon and I worked on it together.

pdurbin:
> I'm pretty sure that neither Dataverse nor Solr supports partial reindexes.

Well, you can reindex individual datasets, for example: http://guides.dataverse.org/en/4.19/admin/solr-search-index.html#reindexing-datasets

poikilotherm:
Thank you everyone involved for a quick solution and all your efforts you spent so eager and willingly! Very much appreciated!

kcondon:
@poikilotherm @pdurbin After some additional testing, post merge, I'm finding some strange behavior:
performance is really poor on our production db copy and the pattern of indexing and logging has changed. After two hours it had indexed 5 dataverses, 27 datasets, and 399 files. With the current system it indexes all the dataverses first, then the datasets and indexes all 6000+ dataverses in 15 minutes. After switching back to prod release and solr the same db, same gfish, solr 7.3.0 on same box as 7.7.2 returned to good performance. I'm not seeing anything in server logs to indicate the issue and gfish/java is showing 100% cpu when running with new solr/war. I'm wondering if I have something misconfigured? I will try the new war with the current solr config to eliminate that from the equation. Any help or suggestions are appreciated.

poikilotherm:
You are testing with full text indexing turned on, right? If so, could you try again with fulltext indexing turned off and see if performance is back again?

Any chance we can create a representative test data set so this could be part of a load test in CI? If you have some time in spare: any chance to test this with current dataverse-sample-data to check if this happens with it, too? Reproducible problems are so much easier to debug... :wink:

kcondon:
@poikilotherm I am not testing with full text indexing and I am using the same db for both current and new solr test scenarios.

kcondon:
@poikilotherm @pdurbin So it looks like it is not the solrj client version, nor solr server but something in the develop branch related to indexing a particular dataverse we have in production, Murray Research Archive. I'm not sure yet what might be the problem but 1. the dv fails to index with no error, 2. batch index is impacted in a weird way, still partly functions, 3. logging is impacted and does not help with understanding the problem. I think 2 and 3 are preexisting and 1 might be related to dv metadata changes in this branch?

pdurbin:
> I do not think it has to do with this pr at this point until we learn more.

I agree. Judging from investigation by @sekmiller the problems mentioned above seem to have been introduced in pull request #6564. #6665 is the issue to watch for updates.

PR #6643, opened 2020-02-12
added API example for updatecurrent

### LINKED ISSUE #6642: Add Documentation for superuser publishing without creating a version using the API
In http://guides.dataverse.org/en/latest/admin/dataverses-datasets.html#make-metadata-updates-without-changing-dataset-version we talk about how superusers can publish without a version update using the UI but we should include an API example as well. PR in a second.

djbrooke:

Adding an API example for a superuser publishing a dataset without a version change

Closes #6642 

Please make sure I got the syntax correct. :)

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/admin/dataverses-datasets.rst
+++ b/doc/sphinx-guides/source/admin/dataverses-datasets.rst
@@ -98,7 +98,9 @@ Forces update to metadata provided to the PID provider of a published dataset. O
 Make Metadata Updates Without Changing Dataset Version
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


qqmyers:
@djbrooke - syntax looks good but I think this functionality is mentioned in two places. You've added the API call in the place where the interface button is mentioned. Your example might go better at http://guides.dataverse.org/en/latest/api/native-api.html#id50 - the Publish a Dataset entry in the Native API guide.

djbrooke:
@qqmyers Thanks! It's already there in the API guide. I went to the Admin guide first and assumed there was no API endpoint until I went to a separate guide and saw it as a smaller note, which I didn't feel was a great experience.

I hear you and agree with you about not wanting to have this reflected in two spots and I'm unwilling to accept that maintenance burden. Our organization of the guides can be better but it's a larger effort. I'll close this out for now and we can revisit if we're able to prioritize the reorganization at a later state, but right now we don't have any capacity for this effort.

PR #6659, opened 2020-02-18
6587 deleted files shown on grant permission

### LINKED ISSUE #6587: Deleted files showing up as duplicates when granting permissions, Add path if it exists
There seems to be this continuous issue when files are deleted from a dataset and what shows up under permissions. Some files from this dataset shows up twice under the "permissions-file" dialogue box so I am never sure if I am giving permissions to a an old deleted file or the file in the most recent published version. Very confusing!

Screen Shot 2020-01-28 at 5 51 10 PM

sekmiller:
This PR allows the admin to deal with file permissions without having to wade through files that have been deleted from previous dataset versions. through the use of a check box she may include deleted files in case there are residual permissions on the deleted files.

Closes #6587 Deleted files showing up as duplicates in Grant File Permissions

Added "Transient" fields to DataFile and Role Assignment Row for ease of use

Add permissions or request access on files, then delete them.

Minor UI change, reviewed with Sonia.

No

**Additional documentation**:
None


src/main/webapp/permissions-manage-files.xhtml
```diff
@@ -25,6 +25,11 @@
             <ui:param name="showMessagePanel" value="false"/>
             <ui:define name="body">
                 <h:form id="rolesPermissionsForm" styleClass="margin-top">
+                    <div class="panel-group">
+                    <p:selectBooleanCheckbox value="#{manageFilePermissionsPage.showDeleted}" itemLabel="#{bundle['dataverse.permissionsFiles.files.includeDeleted']}" onchange="#{manageFilePermissionsPage.init()}" >
```
mheppler: This checkbox for "Include Deleted Files" displays at the top of page, rather than at the top of the table it relates to. I suggest moving it down, next to the file count, above the file table in the Restricted Files panel.

src/main/webapp/permissions-manage-files.xhtml
```diff
@@ -314,7 +321,9 @@
                                     <p:ajax event="toggleSelect"  update=":rolesPermissionsForm:selectionCounter" />
 
                                     <p:column selectionMode="multiple" style="width:2%;text-align:center"/>
-                                    <p:column headerText="#{bundle['dataverse.permissionsFiles.assignDialog.fileName']}">
+                                    <p:column width="98%"  headerText="#{bundle['dataverse.permissionsFiles.assignDialog.fileName']}">
```
mheppler: Not sure about the UX of the second column here for "Draft", while it may help when file has been deleted, it appears to be visual noise for all other use cases, by just have "Restricted" in two columns.

src/main/webapp/permissions-manage-files.xhtml
```diff
@@ -227,7 +233,8 @@
                                     <h:outputText value="#{roleAssignment.assigneeDisplayInfo.title}"/> <h:outputText value=" (#{roleAssignment.assigneeDisplayInfo.affiliation})" rendered="#{!empty roleAssignment.assigneeDisplayInfo.affiliation}"/>
                                 </p:column>
                                 <p:column width="98%" headerText="#{bundle['dataverse.permissionsFiles.files.tabHeader.fileName']}" rendered="#{!empty manageFilePermissionsPage.selectedRoleAssignee}">
-                                    <h:outputText value="#{roleAssignment.definitionPoint.displayName}"/>
+                                    <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.deleted']}) " rendered="#{roleAssignment.deleted}"/>
```
mheppler: I think the "(Deleted)" text should be after the file name, not before, so you can easily scan the file names, since that is the value the table rows are sorted on. The same could be said about the file directory path that is added to this column (e.g "one/001.txt", "two/002.txt" "three/003.txt" shouldn't be the order in the table, if sorting by file path).

src/main/webapp/permissions-manage-files.xhtml
```diff
@@ -314,7 +321,9 @@
                                     <p:ajax event="toggleSelect"  update=":rolesPermissionsForm:selectionCounter" />
 
                                     <p:column selectionMode="multiple" style="width:2%;text-align:center"/>
-                                    <p:column headerText="#{bundle['dataverse.permissionsFiles.assignDialog.fileName']}">
+                                    <p:column width="98%"  headerText="#{bundle['dataverse.permissionsFiles.assignDialog.fileName']}">
+                                        <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.deleted']}) " rendered="#{file.deleted}"/>
+                                        <h:outputText rendered="#{!empty file.directoryLabel}"  value="#{file.directoryLabel}/"/>
```
mheppler: See comment above at line 236 about the file.directoryLabel, in the Grant File Access popup, looking out of order because the table rows are sorted by File Name. Same applies here to the Restricted Files table.

src/main/webapp/permissions-manage-files.xhtml
sekmiller: I think we need to leave it where it ism because including deleted files also effects the role assignments table and if the check box is hidden in the restricted files panel it will be cumbersome to update the role assignments table.

src/main/webapp/permissions-manage-files.xhtml
sekmiller: Yup. moved Deleted to after File Name

src/main/webapp/permissions-manage-files.xhtml
sekmiller: The second column for Draft was not something I introduced here. I think it was done previously for the "same file name" issue that we talked about last week

src/main/webapp/permissions-manage-files.xhtml
mheppler: In that case I would suggest that each table gets its own checkbox, rather than one checkbox, which I would say is just as hidden away at the top of the page. If we did mockups before development, the checkboxes would be directly above each of the restricted file tables, next to the file count.

djbrooke: @mheppler To reiterate what I said in Slack, it would be good if the two of you sat down together to discuss this and other feedback regarding this issue.

mheppler: @djbrooke Right. I sent him a Slack msg offering to discuss.

djbrooke: @mheppler perfect, thanks

mheppler: Thank you.

sekmiller:
@mheppler as discussed I added the "include deleted files" check box to both panels. (Please see lines 53 and 165 - for additional formatting improvements.) Also, as discussed, I'll leave it to you to add the file path and re-style the name/path.

Thanks!

mheppler:
Talked to @sekmiller, talked to @sbarbosadataverse, going to try another approach. Rather than continuing with my effort to add more "Include Deleted Files" checkbox toggles to each of the tables in the panels and popups on the Restricted File Permissions pg, I am going to try to have only one checkbox toggle above the Restricted Files panel datatable, and remove the deleted files from ever displaying in the Grant File Access popup datatable.

As Sonia pointed out, "there is no need to grant access to a deleted file", so there is no need to display deleted files in the Grant File Access popup and no need for a checkbox toggle.

Here is what the Grant File Access popup looks like with the changes as of now, which include my attempt to add a checkbox toggle to the table.

Screen Shot 2020-02-19 at 4 26 19 PM

landreev:
@sekmiller I moved it into qa; one nitpicking thing though:  there seems to be an extra semicolon typo-ed in line 153 of ManageFilePermissionsPage.java

mheppler:
Latest and greatest UI.

Screen Shot 2020-02-21 at 8 48 12 AM

Issue #6525, opened 2020-01-15
weld-osgi-bundle-2.2.10.SP1.jar has moved.


donsizemore:
just an issue to accompany a pull request to point Glassfish 4.1 docs at the new location:

```
[ec2-user@ip-172-30-0-146 ~]$ wget http://central.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/2.2.10.SP1/weld-osgi-bundle-2.2.10.SP1-glassfish4.jar
--2020-01-15 20:09:52--  http://central.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/2.2.10.SP1/weld-osgi-bundle-2.2.10.SP1-glassfish4.jar
Resolving central.maven.org (central.maven.org)... 151.101.52.209
Connecting to central.maven.org (central.maven.org)|151.101.52.209|:80... connected.
HTTP request sent, awaiting response... 501 HTTPS Required
2020-01-15 20:09:52 ERROR 501: HTTPS Required.

[ec2-user@ip-172-30-0-146 ~]$ wget https://central.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/2.2.10.SP1/weld-osgi-bundle-2.2.10.SP1-glassfish4.jar
--2020-01-15 20:09:58--  https://central.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/2.2.10.SP1/weld-osgi-bundle-2.2.10.SP1-glassfish4.jar
Resolving central.maven.org (central.maven.org)... 151.101.52.209
Connecting to central.maven.org (central.maven.org)|151.101.52.209|:443... connected.
The certificate's owner does not match hostname ‘central.maven.org’
```

new location should/might be https://repo1.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/2.2.10.SP1/weld-osgi-bundle-2.2.10.SP1.jar

pull request forthcoming.

PR #6663, opened 2020-02-20
6510 repeated ingests

### LINKED ISSUE #6510: Ingest: Multiple (repeated) ingests of tabular data files
Back in 2018 we have enabled ingest on tab-delimited text files. (This was done as part of the CSV ingest improvement; by reusing the same parsing code, but with the TAB for the delimiter character). 

Apparently, we now have a condition where a successfully ingested file gets picked up for ingest AGAIN - because the content type is tab-delimited, and tab-delimited files are now ingestable... this of course should never happen, because the file already has a datatable object associated with it. But apparently it does occasionally, and an ingested file gets ingested again, corrupting the tab file and the saved original in the process. 

I have a list. There are relatively few of these cases, but this is still very annoying and I believe we should treat it as urgent.

landreev:
No application code in this PR. Only documentation and a diagnostic tool to be sent out to the remote installations, to check if their databases have been affected by these issues.

The release note in this PR **supercedes** the release note added in #6612 (removed in this PR). 
Same for the "pre-release note" - we are only sending out ONE note/one diagnostics script to the remote installations; which checks for both this issue, #6510 and #6525  

Closes #6510

testing the diagnostic script against the copy of the prod. db on vm5 should confirm that neither issue is present; since our database has already been cleaned up. 
If we have an older (2+ months) snapshot sitting around, a more thorough test could be to run it against that, to confirm that it actually finds the issues, when present. 

**Additional documentation**:


djbrooke:
Thanks @sekmiller for the review. I have some small text changes that I'll discuss with @landreev and then we can send this out. I'll assign to myself for now and Leonid and I can talk after standup.

PR #6670, opened 2020-02-21
Iqss/6666 metadata export failure

### LINKED ISSUE #6666: Multistore: Metadata Export fails - tries to save in dataverse.files.directory, instead of the store
The various export format cache files representing dataset metadata are normally written to the storage identifier location upon dataset publish. This is not happening for file type storage when the name of the file storage identifier is anything other than "file", ie. "file1", etc.

This does work for S3 and does work when storage identifier is named "file".

qqmyers:
**What this PR does / why we need it**: Proposed fix for two issues that affect creation of metadata export files that were exposed after the multi-store PR. Should also fix an as-yet-unnoticed issue of the same type that affects storing a dataset logo.

Closes #6666 

One of the problems (0 byte exports for file stores) looks like it has existed for a long time, which is odd (unless everyone is on S3). The other is a consequence of multi-store (nominally multi-store works as advertised, but there was code creating metadata exports and dataset logos that made a no-longer-valid assumption (noted in the code).

Using a file store (nominally one with an id other than 'file' to match the reporting conditions), verify that metadata export files are created and are non-zero length/have real content.

Using the same config, store/view a dataset logo

Using any store, create a dataset, change the default store for that dataset (either the global default, or the store used by the enclosing dataverse) and publish to generate metadata files. Verify that they are stored in the same store as the dataset rather than in the current default store.

This PR catches exporter exceptions slightly differently and it is now possible for one exporter to fail without stopping other metadata export files from being created.

No changes required. If this doesn't make v4.20 (with multi-store), it's possible that some cached aux files will be stored in the wrong place and might need to be removed. (Perhaps developers should be alerted.)

**Additional documentation**:


djbrooke:
Thanks @qqmyers, we'll review

landreev:
Thank you!
Haven't had a chance to test it yet; but the changes are making sense. As I said earlier, I expected that "file" to still be hard-coded there...
But this part: 
> One of the problems (0 byte exports for file stores) looks like it has existed for a long time

How long are we talking about? The lines in ExportService you mentioned are indeed very old... But we do know that metadata exports are working - as of the recent release at least (all/most of our test boxes use the filesystem for storage; and somebody with an installation outside of harvard would have noticed too, I think...)

OK, I'll leave a comment in the orig. issue.

qqmyers:
@landreev - I've seen commits made on PRs, but there may be some trick - maybe direct edits at github but not pushes from local repos. As a work-around, if you can make a PR to that branch, I can go accept it.

PR #6686, opened 2020-02-25
4964 harvesting issues

### LINKED ISSUE #4964: Harvesting - Broken dataset title links for non-Dataverse/OAI-PMH repositories
When Dataverse harvests from some non-Dataverse sources (e.g. ICPSR, DataCite), clicking on the dataset link doesn't take users to the source's dataset page.

You can see an example in [this dataverse on Harvard Dataverse](https://dataverse.harvard.edu/dataverse/srda_harvested/), where a set of metadata records from DataCite was harvested. Clicking on the dataset title link (https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.6141/tw-srda-af010014-1) take you to DataCite's 404 Not Found page (https://oai.datacite.org/dataset.xhtml?persistentId=doi:10.6141/tw-srda-af010014-1). Clicking on the DOI link (https://doi.org/10.6141/tw-srda-af010014-1) in the citation box takes you to the source's dataset page.

(On Harvard Dataverse's "[ICPSR Harvested Dataverse](https://dataverse.harvard.edu/dataverse/icpsr)", clicking on the dataset titles correctly takes you to the sources' dataset pages. I'm not sure how those dataset records were collected into that ICPSR Harvested Dataverse. But they're out of sync with the records that ICPSR makes available over OAI-PMH, and this bug is prevent Harvard Dataverse from updating the datasets it's harvested from ICPSR.)

Could we investigate what's going on?

These are two github issues that might be related:
https://github.com/IQSS/dataverse/issues/4831
https://github.com/IQSS/dataverse/issues/4707

landreev:
This addresses a few harvesting issues from #4964:
- Most important one: the create/edit dialog on the harvesting clients page is now going to actively push the user to select a sensible "archive type" for the remote repository. It no longer defaults to "Dataverse", and it's not going to allow the user to leave it blank. Dataverse needs to know this "archive type" in order to be able to generate redirect links for harvested objects.
- For datasets harvested from "Generic OAI Archives", much improved redirect links. Prod. example: https://dataverse.harvard.edu/dataverse/srda_harvested - clicking on the titles in the search cards is currently redirecting to a bland OAI page on datacite.org. With the fix in this PR, it'll change to redirecting to the proper study pages  via the DOI resolver. 
- Some improved labels/user messages. ("DC" is dropped from the "Generic OAI archive" label, per Julian's suggestion). 
- A somewhat limited fix for the exotic, but annoying problem discovered as we were working on the issue: (bear with me!)
The SRDA dataverse (link above) is configured to harvest from oai.datacite.org. The archive has ~2200 sets total; and it takes *forever* to retrieve the whole list. The site only gives them out in chunks of 50; and as of yesterday it was taking between 40-50+ seconds for each chunk (!). So it would take ~30 min. to retrieve the whole list. In its current form harvesting client setup page needs the full set list before it can proceed to the next step and offer the user a choice of which set to harvest... Obviously, we can't make the user wait 30 min.; and something times out anyway before it's done. This means it is impossible to set up harvesting from oai.datacite.org, and impossible to edit an already configured client. (Julian must have set it up originally back when there were either fewer sets, or it wasn't taking as long to retrieve them). 
The best I could think of to work around this, I'm now checking on the progress, and if it's taken longer than a minute to retrieve the first 100 sets, I proceed to the next step, show the user the truncated list of sets, and warn them about what happened:
Screen Shot 2020-02-25 at 6 22 58 PM

They may still not be able to select the set they want; but they may be able to set up the client with the default set, for example, and then configure the client with the set they want via the API; or they may be able to edit the schedule of an existing client... better than what's going on now. 

If anyone can think of a better solution, please open a new issue for it. But I don't think it's worth investing too much energy into this. I'm not aware of another OAI archive where this is a problem. 

Closes #4964

**Additional documentation**:


src/main/java/propertyFiles/Bundle.properties
```diff
@@ -503,6 +504,7 @@ harvestclients.newClientDialog.step4=Step 4 of 4 - Display
 harvestclients.newClientDialog.harvestingStyle=Archive Type
 harvestclients.newClientDialog.harvestingStyle.tip=Type of remote archive.
 harvestclients.newClientDialog.harvestingStyle.helptext=Select the archive type that best describes this remote server in order to properly apply formatting rules and styles to the harvested metadata as they are shown in the search results. Note that improperly selecting the type of the remote archive can result in incomplete entries in the search results, and a failure to redirect the user to the archival source of the data.
+harvestclients.newClientDialog.harvestingStyle.required=Please select one of the values from the menu
```
pdurbin:
```suggestion
harvestclients.newClientDialog.harvestingStyle.required=Please select one of the values from the menu.
```

I didn't run the code to see how the GUI looks but perhaps this sentence should end with a period? Apologies if I'm mistaken and some other string is concatenated or something.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -482,6 +482,7 @@ harvestclients.newClientDialog.oaiSets.tip=Harvesting sets offered by this OAI s
 harvestclients.newClientDialog.oaiSets.noset=None
 harvestclients.newClientDialog.oaiSets.helptext=Selecting "none" will harvest the default set, as defined by the server. Often this will be the entire body of content across all sub-sets.
 harvestclients.newClientDialog.oaiSets.helptext.noset=This OAI server does not support named sets. The entire body of content offered by the server will be harvested.
+harvestclients.newClientDialog.oaiSets.listTruncated=Please note that the remote server was taking too long to return the full list of available OAI sets, so the list was truncated! Please select a set from the current list (or select the "no set" option), and try again later if you need to change it.
```
pdurbin: My only thought here is that we could replace the exclamation mark with a period. This is sort of a "voice" thing. I'm not sure what the voice for the Dataverse software is or should be.

pdurbin:
> If anyone can think of a better solution

Nope. At least, I can't think of anything that doesn't significantly increase the scope of this issue. Seems like a great workaround. Kudos to @landreev .

src/main/java/propertyFiles/Bundle.properties
landreev: PLEASE NOTE THAT THE REMOTE SERVER WAS TAKING TOO LONG!!!!


caps_on

src/main/java/propertyFiles/Bundle.properties
landreev: ok, done.

src/main/java/propertyFiles/Bundle.properties
landreev: ok, done.

PR #6704, opened 2020-02-27
6665 index dv api fails

### LINKED ISSUE #6665: Index: Cannot api index a particular dataverse in develop, no errors, hangs, affects/prevents batch index too.
At some point the develop branch acquired a regression where indexing via api a particular dataverse hangs, with no errors. Index all is also impacted when it hits that dv -no errors, logging halts but indexing continues very slowly and seemingly out of order -no further dv's are indexed and datasets and files are indexed but extremely slowly with no logs.

Additionally, indexing that single dv appears to continue to consume 100% cpu even though apparently hung.

This works fine with current production build and does not appear to be related to the recent solr upgrade.

From the Solr  upgrade pr where we first encountered this https://github.com/IQSS/dataverse/pull/6631

So it looks like it is not the solrj client version, nor solr server but something in the develop branch related to indexing a particular dataverse we have in production, Murray Research Archive. I'm not sure yet what might be the problem but 1. the dv fails to index with no error, 2. batch index is impacted in a weird way, still partly functions, 3. logging is impacted and does not help with understanding the problem. I think 2 and 3 are preexisting and 1 might be related to dv metadata changes in this branch?

Effectively I can reproduce part of the behavior by trying to index the last attempted dv in the logs:
 curl http://localhost:8080/api/admin/index/dataverses/10

sekmiller:
This makes the indexing of linked dataverses more efficient by modifying the solr docs of the owned datasets instead of doing a full re-index.  When the datasets were fully re-indexed dataverses with large numbers of datasets would fail to re-index.

Closes #6665 

verify that a dataverse with a large number of datasets (mra - for one) can be re-indexed via the api. Also verify that linking dataverses also display the datasets of the linked dataverse.

no
none
**Additional documentation**:


kcondon:
@sekmiller It looks like batch index performance became a lot slower: from 18hours to 6days, estimated. Will do further testing to see whether it is a completion rate issue versus a memory/resource problem, versus a few problem datasets.

Issue #39, opened 2014-07-09
Dataset Title: "Enter a Title for your dataset..." being bold and not a watermark


eaquigley:
---

Author Name: **Elizabeth Quigley** (@eaquigley)
Original Redmine Issue: 3446, https://redmine.hmdc.harvard.edu/issues/3446
Original Date: 2014-01-28
Original Assignee: Elda Sotiri

---

"Enter a Title for your dataset..." being bold and not a watermark in the title field caused users to skip over that field when filling out the form. Needs to be changed to be a watermark or not there at all. 

---

Related issue(s): #27
Redmine related issue(s): [3434](https://redmine.hmdc.harvard.edu/issues/3434)

---

PR #6726, opened 2020-03-04
Iqss/6725 analytics bug

### LINKED ISSUE #6725: analytics tracking broken for download button
Hitting download shows the following in the browser javascript console:
Uncaught TypeError: Cannot read property 'indexOf' of undefined
    at stripId (dataset.xhtml?persistentId=doi:10.18738/T8/F3QE0H:135)
    at getFileId (dataset.xhtml?persistentId=doi:10.18738/T8/F3QE0H:128)
    at HTMLAnchorElement.<anonymous> (dataset.xhtml?persistentId=doi:10.18738/T8/F3QE0H:18)
    at HTMLDocument.dispatch (jquery.js.xhtml?ln=primefaces&v=6.2:2)
    at HTMLDocument.v.handle (jquery.js.xhtml?ln=primefaces&v=6.2:2)

I think this code used to work, but it could be a typo in the example file in the documentation as well. In either case, changing the selector allows the right file identifier to be found so that the download can be tracked in google analytics. PR to follow.

qqmyers:
**What this PR does / why we need it**: Fixes a bug in the analytics-code.html example file in the docs, which breaks button-level tracking for Google analytics.

Closes #6725 

The issue shows a javascript console error that should go away when the fix is applied.

Up to you - I created one. To fix the bug, installations have to reinstall the analytics-code.html file or modify their existing one. It does look to me like this has been broken since ~v.417 - hard to tell if it was a bug/typo in the code or if a UI change broke things at some point. (Pretty sure it worked to start but I can't say when the UI change might have happened.) That said, it only affects installs that have deployed a custom analytics file though.

**Additional documentation**:

### DIFF
--- a/doc/release-notes/6725-analytics-bug.md
+++ b/doc/release-notes/6725-analytics-bug.md
@@ -1,0 +1,3 @@
+# Google Analytics Download Tracking Bug
+
+The button tracking capability discussed in the installation guide (http://guides.dataverse.org/en/4.20/installation/config.html#id88) relies on an analytics-code.html file that must be configured using the :WebAnalyticsCode setting. The example file provided in the installation guide is no longer compatible with recent Dataverse releases (>v4.16). Installations using this feature should update their analytics-code.html file by following the installation instructions using the updated example file. (Alternately, sites can modify their existing files to include the one-line change made in the example file at line 120.)
--- a/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html
+++ b/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html
@@ -117,7 +117,7 @@
       var row = target.parents('tr')[0];
       if(row != null) {
         //finds the file id/DOI in the Dataset page
-        label = $(row).find('td.col-file-metadata > a').attr('href');
+        label = $(row).find('td.col-file-metadata  a').attr('href');
       } else {
         //finds the file id/DOI in the file page
         label = $('#fileForm').attr('action');


mheppler:
@qqmyers looking at the latest and greatest version of this example code in the `develop` and `master` branches, I do not see a typo/missing character.

https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html

Is it possible this may only be an issue on your fork?

qqmyers:
The issue is in line 120 - the '>' character there only finds direct children which is not the case. It should be removed.

mheppler:
D'oh. Apologies. I was seeing your changes backwards, while looking at our analytics code and wonder how we already had the direct child selector... 🤦‍♂.

But, while I was in there, I did do a little more digging into the dataset pg source code for the files table, and looking at those `td.col-file-metadata` table columns. I think that rather than relying on the selector to through that entire block of code, we could keep the original direct child selector and use a different container.

```
<div class="file-metadata-block">
    <a href="/file.xhtml?persistentId=doi:10.7910/DVN/XXXXXX/XXXXXX&amp;version=2.0">
        Sample.txt
    </a>
...
```

I haven't tested it, but I believe we could use `div.file-metadata-block > a` to achieve the same selection of the file id/DOI in the dataset page.

qqmyers:
@mheppler - that looks fine too. If you change it, the release notes should change to match.

mheppler:
@qqmyers PR #39 submitted. Release notes did not specify the change, so it did not need revising.

PR #6735, opened 2020-03-10
6734 default locale

### LINKED ISSUE #6734: As a developer, I want BundleUtil.getDefaultLocale honor my environment
With the merge of #6051 for #5473 for v4.16, a new method `BundleUtil.getDefaultLocale()` has been introduced, used for retrieval of localized strings.

https://github.com/IQSS/dataverse/blob/7717fdc28cbe4dfd314d9c66255bce2e0eb7d98d/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java#L143-L153

This new method will override any locale configuration as soon as the environment variable "LANG" is set and the method is used. For #6694 I was trying to reuse the existing i18n infrastructure to avoid multiple places to run to, thus using this method.

Problems occur now during testing. As I am on a machine with `LANG=de_DE.UTF-8` set, I cannot test with `Locale=en_US`. The pinpointing in beloved `pom.xml` is helpless, as this method wins over defaults.

We should remove the code and get back to forwarding default system behavior, already respecting the `LANG` environment variable. This way we can still pinpoint things, but override as necessary.

PR forthcoming.

Also pinging @JayanthyChengan as the original author of this method.

poikilotherm:
Please see my description for #6734 

Closes #6734

Nothing.

This is currently used in `BundleUtil:147`, `BundleUtil:156` and `SystemConfig:506`.
It should be enough to enable a second language and test setting the default via `-Duser.language` and/or `LANG` (for the JVM, which is a bit tricky when you don't want to mess with your profile).

Nope.

Nope. Actually nothing should change, this is mostly dev related.

**Additional documentation**:
Nada.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
@@ -159,19 +159,20 @@ public class BundleUtil {
         }
         return getStringFromBundleNoMissingCheck(key, null, bundle);
     }
-
+    
+    /**
+     * Return JVM default locale.
+     *
+     * For now, this simply forwards default system behaviour.
+     * That means on JDK8 the system property user.language will be set on startup
+     * from environment variables like LANG or via Maven arguments (which is important for testing).
+     * (See also pom.xml for an example how we pinpoint this for reproducible tests!)
+     * (You should also be aware that good IDEs are honoring settings from pom.xml.)
+     *
+     * Nonetheless, someday we might want to have more influence on how this is determined, thus this wrapper.
+     * @return Dataverse default locale
+     */
     public static Locale getDefaultLocale() {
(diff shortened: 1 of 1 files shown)


djbrooke:
Thanks @poikilotherm, we'll review this. Does this have any impact on the optimization added in https://github.com/IQSS/dataverse/pull/6428/files ?

poikilotherm:
The default is retrieved in the code you mentioned, so it is related. But as it's just about retrieving the Locale, I'm very certain this is not affecting performance in any way.

Re-reading the change you mentioned, one might think about refactoring the getCurrentLocale method, too. I'll keep the scope narrowed to a small change for now, poke me if you want me to look into that.

djbrooke:
@sekmiller I think it's fine to move along if it looks good to you.

PR #6747, opened 2020-03-18
6697 delete user api

### LINKED ISSUE #6697: Update Delete User API Endpoint
#4475 proposes adding a delete button to the dashboard. We want this eventually, but there's some value that can be provided by updating our existing Delete User API to handle a few of the more simple cases. Specifically:

- Test users 
- Users that haven't done anything that prevents their deletion in Dataverse

Over the past few months I've received deletion requests, and when I go to @scolapasta he has to run a few DB queries to determine if the user can be deleted or not. I'd like to be able to delete the user myself using an API, and, if the user cannot be easily deleted, I want to receive messaging that the user can't be deleted.

sekmiller:
Adds testing to the delete user api with a message back to the admin describing the reason the user may not be deleted - or actually doing the deletion

Closes #6697 

**Suggestions on how to test this**: test that users who have done various things created/contributed to ds, has a role, has an open access request, etc. to verify messaging
Details:
Created DV/DS: cannot be deleted.
Edited DS version:  cannot  be deleted.
Downloaded a file: cannot  be deleted if there's a guestbook record.
Has a role defined: cannot  be deleted.

Has an open access request: will be deleted along with request.
Has a notification: will be deleted along with notification
Has a World Map Token: will be deleted along with token
Is a group member: will be deleted along with group membership.

No, but coding done with the idea that "Delete User" will be added to Admin Dashboard
No
**Additional documentation**:
added to admin native api doc


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2806,6 +2806,21 @@ Make User a SuperUser
...
+
+Deletes an ``AuthenticatedUser`` whose ``identifier`` (without the ``@`` sign) is passed. ::
+
+    DELETE http://$SERVER/api/admin/authenticatedUsers/$identifier
+    
+Deletes an ``AuthenticatedUser`` whose ``id``  is passed. ::
+
+    DELETE http://$SERVER/api/admin/authenticatedUsers/id/$id
+    
+Note: If the user has performed certain actions such as creating or contributing to a Dataset they may not be deleted.
```
scolapasta: we should probably add downloading, and no reason (I think) to say "may".

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -511,6 +533,69 @@ public AuthenticatedUser lookupUser( String apiToken ) {
...
+        }
+
+        if (!explicitGroupService.findGroups(au).isEmpty()) {
+            if (!retVal.isEmpty()) {
+                retVal += "; ";
+            }
+            retVal += BundleUtil.getStringFromBundle("admin.api.deleteUser.failure.groupMember");
+        }
+
+        if (userHasPendingAccessRequests(au)) {
```
scolapasta: Currently, once an access request is acted on, we delete, so I think it would be fair to just also delete these requests when you delete the user (instead of failing). My one question is that there has been talk of storing these for historical purposes. If/when we do, we may want to then either a) not allow delete of user or b) anonymize the requests. Or c) still delete them, of course, I just think we'll need to think of that case.

But that said, it makes sense to change that delete to be cascading now.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -511,6 +533,69 @@ public AuthenticatedUser lookupUser( String apiToken ) {
...
+        }
+
+        if (!datasetVersionService.getDatasetVersionUsersByAuthenticatedUser(au).isEmpty()) {
+            if (!retVal.isEmpty()) {
+                retVal += "; ";
+            }
+            retVal += BundleUtil.getStringFromBundle("admin.api.deleteUser.failure.versionUser");
+        }
+
+        if (!explicitGroupService.findGroups(au).isEmpty()) {
```
scolapasta: For delete of user, I think it's be fine to also just remove them from groups at the time.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -511,6 +533,69 @@ public AuthenticatedUser lookupUser( String apiToken ) {
...
+    method which checks for reasons that a user may not be deleted
+    -has created dvObjects
+    -has roles
+    -has guestbook records
+    -etc.
+    An empty string is returned if the user is 'deletable'
+    */
+    
+    public String getDeleteUserErrorMessages(AuthenticatedUser au) {
+        String retVal = "";
```
scolapasta: it's probably simpler code to add them to a list, then do a join to concatenate them at the end.

doc/sphinx-guides/source/api/native-api.rst
sekmiller: Changes "may" to "cannot" and added "downloaded a file"

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
sekmiller: Deleting requests, explicit group memberships and world map tokens in a newly created method that is run if there are no other records that would prevent a deletion (created dataset, downloaded a file etc.)

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
sekmiller: done

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
sekmiller: done

kcondon:
@sekmiller Can you refresh this from develop? I'm seeing a bug that I believe was fixed earlier in S3 branch

kcondon:
Adding a user to a group does not update view on save, though user is added. This is not how it works on demo in 4.19 nor on dataverse-internal running a different dev build.

Issue #6748, opened 2020-03-18
Internationalization DCT - Drop down configure button


lubitchv:
Drop Down "Configure" button in Dataverse shows entries of external tools that are taken from the database.  Therefore the name of tools are only shown in English. For Data Curation there is a need to show the name in different languages.

djbrooke:
@lubitchv thanks. Is this something you're planning to work on? If so I'll plan to move it to Community Dev.

lubitchv:
@djbrooke Yes, I am working on this now.

djbrooke:
Thanks @lubitchv!!

While I have you, I just wanted to make sure that you saw my note in https://github.com/IQSS/dataverse/pull/6577#issuecomment-598733687 about the indexing we had to pull out of the upcoming release.

lubitchv:
Yes, @djbrooke . I saw it. I will think how one can optimize it although I do not have solution yet.

PR #6755, opened 2020-03-20
6748 configure dropdown int

lubitchv:
**What this PR does / why we need it**: Internationalization of DCT

**Which issue(s) this PR closes**: #6748 

**How to test**: 
In foreign bundles, for example in Bundle_fr.properties add

`externaltools.dct.displayname=Data Curation Tool French`
`externaltools.dct.description=Data Curation Tool for curation of variables French`
`externaltools.explorer.displayname=Data Explorer French`

Deploy new code.

Into manifest json of DCT [https://github.com/scholarsportal/Dataverse-Data-Curation-Tool/blob/master/DataCuration_v4.17up.json](https://github.com/scholarsportal/Dataverse-Data-Curation-Tool/blob/master/DataCuration_v4.17up.json) add field `toolName: "dct"`
Into manifest json of DataExplorer  https://github.com/scholarsportal/Dataverse-Data-Explorer/blob/master/dataExplorer_v4.17up.json add field  `toolName: "explorer"`.
Run `curl -X POST -H 'Content-type: application/json' --upload-file manifest.json http://localhost:8080/api/admin/externalTools` for corresponding DataCuration_v4.17up.json and dataExplorer_v4.17up.json instead of manifest.json

Go to ingested tab file, switching the langage should change the "Configure" and "Explore" dropdown box. For "Explore" one will need to have at least two explorer tools to have a dropdown box. For "Configure" were will be a popup window with corresponding description after selecting the tool. There is no popup for explorer.


src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
```diff
@@ -328,4 +330,21 @@ public String toString() {
...
+        }
+    }
+
+    public String getDisplayNameLang() {
+        if (this.getDisplayName().equals(BundleUtil.getStringFromDefaultBundleEng("externaltools.dct.displayname"))) {
+            return (BundleUtil.getStringFromBundle("externaltools.dct.displayname"));
+        } else {
+            return this.getDisplayName();
+        }
+    }
```
qqmyers: Granted there's only one config tool I know of, but won't this call all config tools "Data Curation Tool"? Can the string name be parametrized per tool/by tool name?

qqmyers:
This looks like a one-off solution for DCT (which is admittedly the only config tool I'm aware of). 

Could the checks be to see if externaltool.\<toolid\>.diplayName exists rather than hardcoded?

That might make the check for english go away - tools could continue to have a default name in their json registration file and if the correct translation exists, it could be used instead.

djbrooke:
Thanks @qqmyers for the feedback. We'd indeed want this to handle other configure tools in the future. @lubitchv I'll move this back to community dev for now.

lubitchv:
@qqmyers I am not sure that proposed solution is better, since in that case for any potential additional external tool that is not DCT, it will show DCT display name.

@djbrooke  The question is how do we differentiate between different configure tools. In my solution it was by display name in database compare it to name in englsh Bundle. We can add it to any new configure tool. This is the easiest way. Another solution is to add field in json but we still will need to find a proper translation and it is still comparison. But if anyone has better ideas I am open to them.

lubitchv:
@qqmyers @djbrooke Ok, maybe contracting bundle name form display name will ease the solution. This was your suggestion @qqmyers? Then we can check if such name exist in foreign bundle.

qqmyers:
@lubitchv - yes. I think using the display name could work - removing spaces, etc., but it probably wouldn't be bad to add an id into the tool definition instead (if someone changes their default json to a French displayname, using the displayname to look for other translations in the properties files would fail...)

src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
```diff
@@ -174,4 +174,12 @@ public static Locale getDefaultLocale() {
...
         return new Locale("en");
     }
+
+    public static String getStringFromDefaultBundleEng(String key )  {
+        ResourceBundle bundle = getResourceBundle(defaultBundleFile, new Locale("en"));
+        if (bundle == null) {
+            return null;
+        }
+        return getStringFromBundleNoMissingCheck(key, null, bundle);
+    }
```
qqmyers: With the update to add toolName (which is great!), I think this is no longer needed. Is that right?

lubitchv:
I added new column toolName into externatool table. This is an optional field in manifest json for both configure and explore type of tools. toolName for Data curation should be "dct" and for Data Explorer should be "explorer". The displayName and descriptionName are taken from Bundle which property is build from toolName.

src/main/java/edu/harvard/iq/dataverse/util/BundleUtil.java
lubitchv: @qqmyers You are right. I just removed it.

djbrooke:
@lubitchv @qqmyers is this ready for code review (or was this the code review? 😄)

lubitchv:
@djbrooke Yes, I think it is ready for code review.

djbrooke:
Thanks @lubitchv, we'll review it. Can you add some more details about how to test this?

qqmyers:
@djbrooke - yeah, my informal review is complete, so just having someone at IQSS take a look is good. For testing, since this is now general, it would be useful to check that previewers work OK as is, and that you can internationalize the display name/description by adding a toolname for one and adding something to the bundle file. (Any explore tool I guess - my point is just to check explore and config tool categories.) I guess dataset-level tools should also be tested to see if they break. (I don't think @lubitchv has changed the xhtml where they get displayed, but otherwise this change would work for them too).

djbrooke:
@qqmyers great, thanks!

lubitchv:
@djbrooke I added some comments for testing.

djbrooke:
@lubitchv @qqmyers thanks for the discussion here. I talked briefly with @TaniaSchlatter and @scolapasta just now. If I'm understanding the implementation correctly, should we update http://guides.dataverse.org/en/latest/api/external-tools.html with:

- documentation about toolname
- documentation about updating bundle.properties in the core project if you'd like your tool to be able to be localized

lubitchv:
@djbrooke Yes, I will add these to the documentation.

djbrooke:
Thanks @lubitchv!! On to QA.

kcondon:
@lubitchv Would you refresh from develop as we have just updated the version. Thanks!

lubitchv:
I updated the branch from develop and renamed V4.19.0.3__6748-configure-dropdown-toolname.sql to V4.20.0.2__6748-configure-dropdown-toolname.sql

kcondon:
@lubitchv This works in terms of display language but am seeing some unexpected behavior:
if I have the original version of data explore and data curation tools defined, as well as the new ones, I see both sets in drop down when french is selected but only one set when english is selected. So, I removed all tools and then added only the new curation and explore tools. French works fine but when I add english, the configure drop down for curation tool is blank, whereas the  behavior for explore, with no other explore tool defined, is not to present a list but only the button without a list.

PR #6774, opened 2020-03-26
2734 preserve orig filename

### LINKED ISSUE #2734: Add originalFileName field to json
As per discussion at https://groups.google.com/forum/#!topic/dataverse-community/zsC4yltISS0: when a bundle is retrieved through the data access api, the json metadata contain two fields relating to the original file from which the .tab format is derived: "originalFileFormat" and "originalFormatLabel". Unless the bundle is unzipped and the package contents extracted, the filename of the original file has to be inferred from these fields. This is causing a few issues when we ingest content from Dataverse into Archivematica, because Archivematica needs to know the name of the original file before the unpackaging micro-service takes place.

sekmiller:
This PR provides a method for storing the original file name for ingested tabular files. It also adds the original file name to the json printer for data files. For files that were uploaded previously the son printer will infer the original file name from the file metadata and a conversion of the extension based on the original file format field on the data table.

Closes #2734 

None

Verify that ingested tabular files json export includes the original file name even if the file was ingested previously and thus does not have the original file name saved to the dataTable table.

No

No

**Additional documentation**:
There is a sql script that adds the new field to the dataTable table.


src/main/java/edu/harvard/iq/dataverse/DataFile.java
```diff
@@ -398,6 +399,29 @@ public Long getOriginalFileSize() {
...
+
+    private String getDerivedOriginalFileName() {
+        FileMetadata fm = getFileMetadata();
+        String filename = fm.getLabel();
+        String originalExtension = FileUtil.generateOriginalExtension(getOriginalFileFormat());
+        String extensionToRemove = StringUtil.substringIncludingLast(filename, ".");
+        if (StringUtil.nonEmpty(extensionToRemove)) {
+            String newFileName = filename.replace(extensionToRemove, originalExtension);
+            return newFileName;
+        }
```
landreev: @sekmiller Am I reading this correctly - that if the current filename has no extension ("foo"), this code will return "foo"? - shouldn't it still return "foo.<ORIG_EXT>"? 
(this shouldn't really happen - since ingested files should always have the ".tab" extension - but still)

sekmiller: Good points. I will take another pass at this.

sekmiller: @landreev I put in your suggestion on using replaceAll and relying on the fact that a tabular file should always end in ".tab ". It occurs to me that any editing of the file label to anything other than foo.tab will break this (including your "no extension" case). How worried should we be about this, and any suggestions to code against it?

sekmiller: Answered my own question while out for a run: so the replaceAll("tab$" which I put in should fix 100% of the cases. Since we know it won't I can modify the previous "look for an extension to replace" code by adding a $ to the replace string so that inadvertent internal instances of the extension are not replaced, and if there's no extension found just append the extension derived from the file type. All of this points out that even after all this we cannot be certain that we're returning anything like the actual original file name.

matthew-a-dunlap:
Was just talking about this with @donsizemore , we're excited to see this as part of Dataverse 5!

PR #6790, opened 2020-04-03
6558 validate files on publish

### LINKED ISSUE #6558: Validate the physical files when publishing a dataset
This is a proposal/an idea to consider. 

Analyzing the issues we've had with users' physical files over the years, most were caused by something that went wrong with the files still being in the Draft/unpublished state. 

This is because we only attempt/allow to delete the physical file before the datafile gets published. And that's when something can go wrong - most commonly, an attempted delete or ingest would remove or corrupt the physical file, while preserving the datafile entry in the database. Then the user publishes the version, without realizing the physical file is missing. 

This makes me think we should consider going through the files as they are being published, and validating/confirming that they are still present, that the checksums add up, etc. 
Somewhat similarly to how we attempt to register global ids for datafiles on publish. 

Doing so would prevent most, if not all possible situations when a physical file is missing from a published dataset.

landreev:
See the issue; and the release note: https://github.com/IQSS/dataverse/blob/6558-validate-files-on-publish/doc/release-notes/6558-validate-files-on-publish.md

Closes #6558 

Create a dataset, upload some files, mess up one of the files in the dataset directory or in the S3 bucket... try to publish.
Try it with both the synchronous and asynchronous ("too many files") publishing modes. 
There is a new UI error message notifying the user about the failure to publish, explaining to them that some of their files do not pass validation, and telling them that they must contact support in order to fix it, before they can make another attempt to publish.
It uses the standard "red" error banner system at the top of the page.

yes, above.
**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2011,48 +2037,42 @@ private String init(boolean initFull) {
...
                         BundleUtil.getStringFromBundle("dataset.pidRegister.workflow.inprogress"));
             }
+            if (dataset.isLockedFor(DatasetLock.Reason.FileValidationFailed)) {
+                // the dataset is locked, because one or more datafiles in it 
+                // failed validation during an attempt to publish it. 
+                if (FacesContext.getCurrentInstance().getExternalContext().getFlash().get("errorMsg") == null) {
+                    JH.addMessage(FacesMessage.SEVERITY_ERROR, BundleUtil.getStringFromBundle("dataset.publish.file.validation.error.message"),
+                            BundleUtil.getStringFromBundle("dataset.publish.file.validation.error.contactSupport"));
+                }
+                /* and now that we've shown the message to the user - remove the lock? */
```
sekmiller, line 2062: Because we want the admin to intercede here we probably don't want to remove the lock, but should we update the  edit data button so that if there is a file validation failure the user cannot add files? We would still allow them to make other edits?

sekmiller: Or maybe use the file validation failure lock to disable the publish button?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -225,6 +240,98 @@ private void updateParentDataversesSubjectsField(Dataset savedDataset, CommandCo
         }
     }
 
+    private void validateDataFiles(Dataset dataset, CommandContext ctxt) throws CommandException {
+        try {
+            for (DataFile dataFile : dataset.getFiles()) {
+                // TODO: Should we validate all the files in the dataset, or only 
+                // the files that haven't been published previously?
+                logger.log(Level.FINE, "validating DataFile {0}", dataFile.getId());
```
sekmiller, line 248: I think it would be desirable to attempt to validate all of the files before throwing an exception. It would make the admins job easier if they knew more than one file failed validation.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
landreev: No, we don't want to remove that lock automatically - that was an idea I was entertaining during the development. I should remove that comment.
As for the second comment - that's exactly what that "failed validation" lock does. It disables the publish button, and will prevent an execution of the publish command, if called via API. But it shouldn't prevent the contributor from making changes. But yes, this should be documented clearly.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
landreev: Thanks for the suggestion. I decided to quit on the first invalid file, so that it's faster. And then I tell the admin in the troubleshooting instruction that they need to verify all the files in the dataset.
I just had this idea: I should probably leave the finalizePublication command as is, stopping on the first file that fails - but then I should provide a one step API for the admin, that will validate all the files in the dataset (using the existing code), and spell out to them which ones need to be fixed - ?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: OK. I missed that the lock disables publish because it's if it's locked for any reason, not just file validation failure.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
landreev: OK, this is done.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
landreev: The way that page is dealing with/checking for various types of locks is a little scary.
Added a quick note to the "dataset management"/"publish dataset" section of the user guide.

kcondon:
@landreev Turning off validation does not seem to stop validation.

landreev:
>  Turning off validation does not seem to stop validation.

OK, will investigate and fix.

landreev:
@kcondon 
In item 3.:
> No logging present on develop branch for same scenario. Update: I have also seen this fail but with index error only at end.

So does the dataset get published or not, w/ develop branch on dvn-build? (not 100% how to read "same scenario" - so checking...)
As for the "logging", I'm going to assume that those were whitehat scan messages - is that safe?

kcondon:
@landreev Yes, demo works

landreev:
Checked in fixes, should be ready to re-test.

landreev:
@kcondon
>     4. Minor doc issue, preexisting but seems like we should fix since we now reference it 2x, :PIDAsynchRegFileCount does not have it's own entry in config/settings, only mentioned inline as a setting in file pid and now file validation settings.

It was really "preexisting", in that apparently it wasn't documented at all. I added that mention of it, to both the file reg. and the validation options sections. I wasn't sure it needed its own section, but sure, I just added its own paragraph.

landreev:
There's one more thing - whether we want to validate every file, every time, where I was soliciting opinions on how to handle it. I'm leaning towards skipping it for minor versions; but validating every file, new and old on every major version. But this hasn't been implemented yet.

landreev:
> I'm leaning towards skipping it for minor versions; but validating every file, new and old on every major version. 

OK, this has been checked in as well.

Issue #6554, opened 2020-01-22
Adjust how new variable metadata is added to DDI XML exports so that the exports are valid against DDI schema


jggautier:
The Data Curation Tool adds new metadata elements to the DDI XML that Dataverse produces. The variable elements universe and qstn (and sub-elements inside qstn like literal question, interview instructions, and post question) are in this example:

```
<var ID="v12560" name="var1" intrvl="discrete">
	<location fileid="f9842"/>
	<labl level="variable">var1</labl>
	<universe>This is the universe.</universe>
	<sumStat type="mean">5.5</sumStat>
	<varFormat type="numeric"/>
	<notes subject="Universal Numeric Fingerprint" level="variable" type="Dataverse:UNF">UNF:6:8OuUohmPouAgl2L3IrfN2A==</notes>
	<notes>
		<![CDATA[ These are the notes for var1, v12560. ]]>
	</notes>
	<qstn>
		<qstnLit>Is this a literal question?</qstnLit>
		<ivuInstr>These are instructions.</ivuInstr>
		<postQTxt>Is this the post question?</postQTxt>
	</qstn>
</var>
```

The DDI schema definition (view-source:https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd) says that the `<qstn>` element and its sub elements are out of order here, and should come before `<universe>`. 

<img width="629" alt="Screen Shot 2020-01-21 at 9 10 03 PM" src="https://user-images.githubusercontent.com/18374574/72859643-97eced00-3c92-11ea-86ad-5d4b9a1f52a5.png">

So it should be:
```
<var ID="v12560" name="var1" intrvl="discrete">
	<location fileid="f9842"/>
	<labl level="variable">var1</labl>
	<qstn>
		<qstnLit>Is this a literal question?</qstnLit>
		<ivuInstr>These are instructions.</ivuInstr>
		<postQTxt>Is this the post question?</postQTxt>
	</qstn>
	<universe>This is the universe.</universe>
	<sumStat type="mean">5.5</sumStat>
	<varFormat type="numeric"/>
	<notes subject="Universal Numeric Fingerprint" level="variable" type="Dataverse:UNF">UNF:6:8OuUohmPouAgl2L3IrfN2A==</notes>
	<notes>
		<![CDATA[ These are the notes for var1, v12560. ]]>
	</notes>
</var>
```
[...]

PR #6794, opened 2020-04-03
6554 var ddi validation

lubitchv:
**What this PR does / why we need it**: Fixes validation error for variable level metadata (order of qstn and universe).

**Which issue(s) this PR closes**: #6554 

Closes # #6554 

**Suggestions on how to test this**: Use Data Curation Tool for adding var metadata for questions and universe. Save to dataverse, publish the dataset. Validate the generated cached xml file using DDI validator, for example  [ https://www.freeformatter.com/xml-validator-xsd.html ]( https://www.freeformatter.com/xml-validator-xsd.html ) There will be errors, since this PR does not fix all the errors for DDI validation, but only for question and universe. Should be no errors for qstn and for universe.


djbrooke:
Thanks @lubitchv, we'll review this.

Issue #6772, opened 2020-03-26
Typeahead widget on Link Dataset popup is slow and doesn't display all options as expected


TaniaSchlatter:
The type ahead feature on the Linking popup is slow to respond to user input. Also, if a user types quickly, not all results that include what has been typed in are shown. This is problematic because it looks like not all results are returned.

All items that include "Tania Schlatter Dataverse": 
<img width="1190" alt="Screen Shot 2020-03-26 at 2 18 31 PM" src="https://user-images.githubusercontent.com/15386424/77683192-6e8c5e00-6f6e-11ea-9c63-74e871c394f5.png">

What is shown when I type "Tania Schlatter Dataverse". "Tania Schlatter Dataverse" is not shown:
<img width="1195" alt="Screen Shot 2020-03-26 at 2 18 42 PM" src="https://user-images.githubusercontent.com/15386424/77683201-70eeb800-6f6e-11ea-9374-89f3e226311c.png">

The typeahead for super users is more responsive and accurate:
<img width="898" alt="Screen Shot 2020-03-26 at 2 20 32 PM" src="https://user-images.githubusercontent.com/15386424/77683200-70eeb800-6f6e-11ea-9cb4-9dd27c334a00.png">

<img width="1167" alt="Screen Shot 2020-03-26 at 2 20 38 PM" src="https://user-images.githubusercontent.com/15386424/77683197-70562180-6f6e-11ea-8a2f-e664a4657bdd.png">

TaniaSchlatter:
Possibly related to #6754

TaniaSchlatter:
In the design meeting we discussed addressing this by removing the typeahead, and providing more robust functionality for finding Dataverses you can add to in a popup. 

The idea is that Harvard or other open installation users, who can deposit to Root, see the option to change the default Dataverse and add a template in a new block, below. "Edit" will open a popup (mockup to come). 

<img width="1027" alt="Create_dataset_revised2" src="https://user-images.githubusercontent.com/15386424/78172045-637c7680-7423-11ea-8b32-6bb5452e3c4b.png">

We need further discussion to figure out what happens for installations who's users can not add to the Root dataverse. We discussed that the Add Data button should not appear in the header for installations who's users who can not add to Root (related to #6708).

There are 4 user states that relate to the Add Data button in the header. We need to figure out how those states relate to and might affect this proposed approach.

1. User can not add to Root = no "Add Data" button in header.
[...]

PR #6804, opened 2020-04-09
6634 primefaces8

### LINKED ISSUE #6557: Cannot upload files without extensions
### Background:
*  Political Analysis is working with authors and with Code Ocean (whom I work for) to make authors' code reproducible & executable as part of the publication process.
* Authors upload their materials to Code Ocean, e.g. https://doi.org/10.24433/CO.4189072.v1, and then those materials are exported to the Political Analysis Dataverse (https://dataverse.harvard.edu/dataverse/pan) 
* We currently do this manually but the goal is to have API integration that automates the export and uploading process.

### Problem:
* Code Ocean's 'master script' (the thing the system executes when a reader or author presses 'Reproducible Run') is just called `run`, no file extension. It's a shell script comprising something like:

```
#!/usr/bin/env bash
set -ex

# This is the master script for the capsule. When you click "Reproducible Run", the code in this file will execute.
Rscript "Run_all.R"
```

* It's a bash script, but we dropped the extension at some point.

* Dataverse does not seem to allow us to upload files without file extensions.
* When I drag and drop `run` (or something called `newfile` with arbitrary text in it) via the Dataverse Interface, my browser (Firefox) scrolls to the top of the page, and when I scroll down, the file has not been uploaded.
* When I tried a few weeks ago (on Chrome) to upload all the materials from https://doi.org/10.24433/CO.4189072.v1, I received a message that `run` and `Run_all.R` are duplicates and only `Run_all.R` was successfully uploaded.

### The short-term solution:
* We can rename all the `run` scripts `run.sh` before they are uploaded to Dataverse.

### The problem with that short-term solution, long-term:
* An additional manual step in the process makes API integration harder because it's one more thing that could go wrong in an unsupervised setting (what if users already have a `run.sh` script uploaded, for instance). 

### The ideal solution (from our POV):
* Dataverse allows files that do not have file extensions to be uploaded.

### LINKED ISSUE #6634: Upgrade to PrimeFaces 8
When it's [available](https://github.com/primefaces/primefaces/releases), we should upgrade. Have not seen any [updates from PrimeFaces](https://www.primefaces.org/blog/).

Blocked/related(?) issues:
- Cannot upload files without extensions #6557
- java.lang.OutOfMemoryError: GC overhead limit exceeded #4902 

More details on the changes from 7.0 in the [Migration Guide](https://github.com/primefaces/primefaces/wiki/Migration-Guide).

#### Breaking Changes
* PrimeFaces is now only compatible with Java8+
* Removed long time deprecated image attribute on buttons
* Facelet function `p:component` has been removed in favor of: `p:resolveFirstComponentWithId('myId', view).clientId`
* Facelet function `p:widgetVar` have been removed in favor of: `p:resolveWidgetVar('myTable', view)`
* StreamedContent API
  * `DefaultStreamedContent`: Constructors and setters have been deprecated in favor of `DefaultStreamedContent.Builder`
  * `ByteArrayContent` has been removed in favor of `DefaultStreamedContent.Builder#stream(() -> new ByteArrayInputStream(...))`
  * `LazyDefaultStreamedContent` has been removed, instead use `DefaultStreamedContent` with a `Supplier` which will always "lazy-creates" the InputStream: `DefaultStreamedContent.Builder#stream(() -> new FileInputStream(...))`
* DataTable
  * Methods `PrimeFaces#clearTableStates()` and `PrimeFaces#clearDataListStates()` have been deprecated in favor of `PrimeFaces#multiViewState()`
  * DataTable: `getFilters` / `setFilters` has been removed in favor of `getFilterBy` / `setFilterBy`
  * `LazyDataModel#load` signatures has been changed; sortBy is now a Map instead of List, filterBy Map value is now a `FilterMeta` object instead the filter value. Please note that there is a change between 8.0RC1 and 8.0RC2.
* Menu
  * `DynamicMenuModel` has been removed, use `BaseMenuModel` instead.
* FileUpload API
  * Models have been moved from `org.primefaces.model` to `org.primefaces.model.file`
  * `UploadedFile#getContents` has been renamed to `getContent`
  * `UploadedFile#getInputstream` has been renamed to `getInputStream`
  * `FileUpload#fileUploadListener` has been renamed into `listener`
  * In case of using multiple feature in a simple mode, use `UploadedFiles` model to get all files in a single request.
* DataExporter API
  * `org.primefaces.component.export.Exporter` interface allows you to implement your own exporter using `exporter` attribute
  * Removal of `repeat` attribute: used as a workaround in case of tables contained in an iterator component _(e.g `ui:repeat`, `p:tabView` etc.)_. No longer necessary, instead, define every ids component you wish to export (without specifying component index)
  * Apache POI minimum version 4.1.1 to use Excel export
* TextEditor for security reasons (https://github.com/primefaces/primefaces/issues/5163) now requires the OWASP Sanitizer library by default but you can choose to opt-out of using the library by setting the attribute `secure="false"`.
* InputMask for security reasons (https://github.com/primefaces/primefaces/issues/5105) validates the mask used on the server side. If you would like to disable that set validateMask="false".  Especially if you use a custom mask with `$.mask.definitions` for example: https://github.com/primefaces/primefaces/issues/3234#issuecomment-361221532 
* `p:watermark`has been reimplemented with a simple HTML placeholder attribute. Therefore `PrimeFaces.cleanWatermarks()` and `PrimeFaces.showWatermarks()` has been removed.
* InputNumber: upgraded using [Autonumeric](http://autonumeric.org/)
  * `emptyValue`: `sign` is now `always`
  * `symbolPosition`: `prefix` is now `p`, `suffix` is now `s`
  * `decimalPlaces`: now defaults to 2
* TimeLine: updated to the newest [vis](https://visjs.org/) timeline version.
  * `TimelineEvent` uses `java.time.LocalDateTime` instead of `java.util.Date`
  * `TimelineEvent` has now 3 more properties to fine tunning editable: editableTime, editableGroup and editableRemove.
  * `TimelineGroup` has now styleClass property similar to the one in TimelineEvent.
  * Properties deprecated in favor of others: `axisOnTop`, `groupsChangeable`, `groupsWidth`, `groupMinHeight`, `snapEvents`, `timeChangeable`.
  * Properties deprecated with NO replacement: `dragAreaWith` (use css classes instead), `unselectable`, `groupsOnRight`, `animate`, `animateZoom`, `showButtonNew`, `showNavigation`, `browserTimeZone`.
  * New properties: `maxHeight`, `orientationAxis`, `orientationItem`, `editableAdd`, `editableTime`, `editableGroup`, `editableRemove`, `editableOverrideItems`, `eventHorizontalMargin`, `eventVerticalMargin`, `groupStyle`, `snap`, `clickToUse`, `clientTimeZone`.
  * New `extender` property to specify a javascript function to extend configuration of vis.js timeline component similar to extender property in `<p:chart />` component.
* Schedule upgraded to FullCalendar 4.x
  * `ScheduleEvent` uses `java.time.LocalDateTime` instead of `java.util.Date`
  * Removed deprecated attributes
  * Localization comes with FullCalendar: does not require/use PF-localization
  * New view names (`dayGridMonth`, `dayGridWeek`, `dayGridDay`, `timeGridWeek`, `timeGridDay`, `listYear`,` listMonth`, `listWeek`, `listDay`): old ones are translated by Schedule component
  * `ScheduleEntryMoveEvent`: new properties `yearDelta` and `monthDelta`
  * `ScheduleEntryResizeEvent`: new properties `yearDelta` and `monthDelta`: splitted properties between start and end (up to PF 7.0 only end)
  * Property `slotLabelFormat` was removed (only available in commercial version of FullCalendar - see https://fullcalendar.io/docs/slotLabelFormat)
  * `ScheduleEvent`: new properties `groupId` and `overlapAllowed`
* Calendar and DatePicker: Without a value-binding it converts OOTB to java `java.time.LocalDate`. (up to PrimeFaces 7.0:  `java.util.Date`) This may be relevant when you use Calendar or DatePicker as a filter for eg DataTable. You can add something like `<f:convertDateTime type="date" />` to your Calendar or DatePicker to get a java.util.Date.

#### Others
* Fluent builders
  * Menu: instanciate menu items using builders (e.g `DefaultMenuItem#builder()`)
  * Timeline: instanciate timeline events using builders (e.g `TimelineEvent#builder()`)
  * Schedule: instanciate schedule events using builders (e.g `DefaultScheduleEvent#builder()`)
  * DefaultStreamedContent: instanciate `DefaultStreamedContent` using builders (e.g `DefaultStreamedContent#builder()`)
  * All constructors with arguments are deprecated
* Calendar and DatePicker offer built-in support for `java.time.LocalDate`, `java.time.LocalDateTime` and `java.time.LocalTime` (no converter needed anymore)
* Many events (eg SelectEvent, UnselectEvent, RowEditEvent, etc.) now use generics
* Cropper: change libraries from JCrop to Cropper.js
* PanelGrid: new layout="flex" based on PrimeFlex/FlexGrid
* Menu
  * `MenuModel#addElement()` has been deprecated, add element using `MenuModel#getElements()#add()` instead.

scolapasta:
Upgrades Primefaces to the latest v8.

Closes #6634 
Closes #6557

changes were surgical and are due to changes in primefaces APIs or moving some popup code

Smoke / regression tests; no new functionality. (though upload component now allows files without extension, so that should be tested in order to close #6557)

no (except in allowing users to upload files without extensions)

Might be worth mentioning the upgrade, but nothing special needs to be done.

**Additional documentation**:


kcondon:
Issues found so far:
1. Uploading 3 files to s3 direct causes file upload to not complete, no errors, progress bars stuck.
2. Host Dataverse type ahead/lookup logic not working after lookup failure/correction on create dataset page. Does not try again.

scolapasta:
For #2, since it also seems to happen on dev, doesn't seem to be PF8 related, specifically. Regardless, we've decided to remove this component completely for a better designed solution (#6772). So I don't think it's worth the effort to fix as it will be gone soon - that issue is in the Up Next column and should be complete before the next release.

I'll copy over the report to that issue, just so we can track, it there if need be.

kcondon:
@scolapasta I'm going to merge this since all functional stuff is working but there is that one extra logging issue, which I'll open as a separate ticket.

PR #6809, opened 2020-04-10
6761 payara installer update

### LINKED ISSUE #6761: Payara Installer Updates
Update the installer to install and configure payara 5.201, rather than glassfish 4. Since Payara is based on Glassfish, it should be possible to keep the same configurations from jvm-options.

landreev:

These changes make the new Python3 installer work with Payara. (v.5.201+ required)

Closes #6761

**Additional documentation**:


poikilotherm:
Instead of configuring domain1, you might think about using the production domain, which does a lot of tweaks for you already.

https://docs.payara.fish/documentation/payara-server/production-ready-domain.html

landreev:
> Instead of configuring domain1, you might think about using the production domain, which does a lot of tweaks for you already.

Cool, I'll take a look.

scripts/installer/install.py
```diff
@@ -229,12 +230,12 @@
...
    # 1a. check to see if warfile is available
    warfile = "dataverse.war"
-   warfileVersion = None
+   warfileVersion = ""
    if not os.path.isfile(warfile):
       # get dataverse version from pom.xml
       tree = ET.ElementTree(file='../../pom.xml')
       for elem in tree.iter("*"):
-         if elem.tag == '{http://maven.apache.org/POM/4.0.0}version':
+         if elem.tag == '{https://maven.apache.org/POM/4.0.0}version':
```
pdurbin, line 238: I'm just noting the conversation about this change at https://github.com/IQSS/dataverse/pull/6812#issuecomment-612100332

conf/jhove/jhove.conf
```diff
@@ -3,7 +3,7 @@
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xmlns="http://hul.harvard.edu/ois/xml/ns/jhove/jhoveConfig"
  xsi:schemaLocation="http://hul.harvard.edu/ois/xml/ns/jhove/jhoveConfig
-                     file:///usr/local/glassfish4/glassfish/domains/domain1/config/jhoveConfig.xsd">
+                     file:///usr/local/payara5/glassfish/domains/domain1/config/jhoveConfig.xsd">
```
donsizemore, line 6: This has been fine under Glassfish4 and is fine under Payara5, but installation-main sez

> The installer places files called jhove.conf and jhoveConfig.xsd into the directory /usr/local/glassfish4/glassfish/domains/domain1/config by default and makes adjustments to the jhove.conf file based on the directory into which you chose to install Glassfish.

We might a) update that to payara5 and b) I don't see logic in the script to adjust the path in jhove.conf; we might note in the docs that the file would need to be edited?

donsizemore:
I just ran these updated scripts successfully using stock python-3.6 on CentOS 8. server.log looks clean as best I can tell.

conf/jhove/jhove.conf
landreev: There's no code in the script that adjusts that path, no. But there's **the next best thing** in there, a comment:
```
# @todo: The JHOVE conf file has an absolute PATH of the JHOVE config schema file (uh, yeah...)
# - so it may need to be readjusted, if glassfish lives somewhere other than /usr/local/glassfish4
# (replicate from the old installer)
```
(it looks like it just slipped through the cracks. will add.)

scripts/installer/default.config
```diff
@@ -1,7 +1,7 @@
 [glassfish]
 HOST_DNS_ADDRESS = localhost
-GLASSFISH_USER = glassfish
-GLASSFISH_DIRECTORY = /usr/local/glassfish4
+GLASSFISH_USER = dataverse
+GLASSFISH_DIRECTORY = /usr/local/payara5
```
pdurbin, line 4: I almost touched `doc/sphinx-guides/source/_static/util/default.config` as part of the doc issue I'm working on but stopped after seeing the edit above. There are many copies of `default.config` (screenshot below) and we should figure out where/when to update them.

Screen Shot 2020-04-13 at 2 36 41 PM

landreev: Note that these other copies are a different config file. Must be an older style docker config? - the format is different/incompatible with what the python installer is using. 

I should probably rename the new, configparser-style file.

landreev:
We are going to stay with domain1, at least during this dev. cycle. Among other things, for the sake of keeping newly-built Payara instances as similar to the ones running on domains migrated from existing Glassfish instances as possible.
We have adopted the more important options listed in https://docs.payara.fish/documentation/payara-server/production-ready-domain.html.
The goal of this PR, at this point in the release cycle, is to have an installer that a developer can use to create a dev. environment. For which domain1 appears to be optimal.  We may  want to revisit whether to use `domains/production` as the template for a production installation; once we are closer to releasing D5 into the world where it will be used in prod. environment.

poikilotherm:
Thanks for the info @landreev. Let me add a note that dataverse-kubernetes images use `production` only, so I'm testing it with that domain.

landreev:
@poikilotherm Great, thanks. Let's keep comparing notes on this. 
I definitely see advantages in eventually going with this production domain template. Mostly because the "production" seem to be the one they are planning on maintaining/updating. 
But then it doesn't seem to be the optimal configuration for a dev. environment. So it sounds like we'll probably have to maintain 2 different setting templates, for dev. and prod. use. Not to mention changing every line in the guides where we say "domains/domain1" now...
I'm ok with all this, but maybe not in this iteration.

scripts/installer/install.py
```diff
@@ -616,7 +631,7 @@
 print("as the following two JVM options:")
 print("\t<jvm-options>-Ddoi.username=...</jvm-options>")
 print("\t<jvm-options>-Ddoi.password=...</jvm-options>")
-print("and restart glassfish")
+print("and restart payara")
 print("If this is a production Dataverse and you are planning to register datasets as ")
 print("\"real\", non-test DOIs or Handles, consult the \"Persistent Identifiers and Publishing Datasets\"")
 print("section of the Installataion guide, on how to configure your Dataverse with the proper registration")
```
pdurbin, line 637:
```suggestion
print("section of the Installation guide, on how to configure your Dataverse with the proper registration")
```

pdurbin:
> production domain template

Just a heads up that @donsizemore is looking into it for dataverse-ansible: https://github.com/IQSS/dataverse-ansible/issues/176

PR #6842, opened 2020-04-20
#4285 implement sorting in Manage Users table.

### LINKED ISSUE #4285: Sort and filter functionality on User Administration
Manage Users table lists all user accounts in table form.
If sort functionality and filter by fields or either of them can be available, it would make the user management much easier.

pkiraly:
**What this PR does / why we need it**: It makes the Manage Users table sortable by clicking the column headings

**Which issue(s) this PR closes**: #4285

Closes #4285

**Special notes for your reviewer**: It is a user interface change in the xhtml file.

**Suggestions on how to test this**:  Right now I can not imagine other ways to test than build the war, log in as abministrator, go to Manage Users and click on the column headings.

**Does this PR introduce a user interface change?**: It makes the Manage Users table sortable by clicking the column headings

**Is there a release notes update needed for this change?**: It is a minor change, however as a site administrator I alaways wanted to know who are the new users, and I was not able to get than other than issue SQL query from the command line. Now sort users by user ID gives me an easy way to check new users.

**Additional documentation**: No

### DIFF
--- a/src/main/webapp/dashboard-users.xhtml
+++ b/src/main/webapp/dashboard-users.xhtml
@@ -47,26 +47,26 @@
                     </ui:include>
 
                     <p:dataTable id="userTable" value="#{DashboardUsersPage.userList}" var="user" widgetVar="userTable">
-                        <p:column style="width:60px" class="text-center" headerText="#{bundle['dashboard.list_users.tbl_header.userId']}">
+                        <p:column style="width:60px" class="text-center" headerText="#{bundle['dashboard.list_users.tbl_header.userId']}" sortBy="#{user.id}">
                             <h:outputText value="#{user.id}" />
                         </p:column>
-                        <p:column headerText="#{bundle['dashboard.list_users.tbl_header.userIdentifier']}">
+                        <p:column headerText="#{bundle['dashboard.list_users.tbl_header.userIdentifier']}" sortBy="#{user.userIdentifier}">
                             <span class="glyphicon glyphicon-user text-danger" jsf:rendered="#{user.superuser}" style="margin-right:6px;"></span>
                             <h:outputText value="#{user.userIdentifier}" class="#{user.superuser ? 'text-danger' : ''}" />
                         </p:column>
-                        <p:column headerText="#{bundle['dashboard.list_users.tbl_header.name']}">
+                        <p:column headerText="#{bundle['dashboard.list_users.tbl_header.name']}" sortBy="#{user.lastName}">
                             <h:outputText value="#{user.firstName} #{user.lastName}" />
                         </p:column>
-                        <p:column width="20%" headerText="#{bundle['dashboard.list_users.tbl_header.email']}">
+                        <p:column width="20%" headerText="#{bundle['dashboard.list_users.tbl_header.email']}" sortBy="#{user.email}">
                             <h:outputText value="#{user.email}" />
                         </p:column>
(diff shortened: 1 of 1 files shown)


djbrooke:
Thanks @pkiraly we'll take a look. We have some key people out this week so it may be a bit delayed. 

Any idea how this performs on large user lists? We have ~30k users in Harvard Dataverse.

pkiraly:
@djbrooke No, I do not know how it will behave with such a large number. But now I check what happens in the browser. In the background sort action triggers an Ajax call, it sends the following dataa to the Java application

```
curl 'https://.../dashboard-users.xhtml' \
  -H 'User-Agent: ...' \
  -H 'Accept: application/xml, text/xml, */*; q=0.01' \
  -H 'Accept-Language: en-US,hu;q=0.7,en;q=0.3' \
  --compressed \
   -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'Faces-Request: partial/ajax' \
  -H 'X-Requested-With: XMLHttpRequest' \
  -H 'Origin: https://...' \
  -H 'DNT: 1' \
  -H 'Connection: keep-alive' \
  -H 'Referer: https://.../dashboard-users.xhtml' \
  -H 'Cookie: ...' \
  --data 'javax.faces.partial.ajax=true
             &javax.faces.source=dashboardUsersForm%3AuserTable
             &javax.faces.partial.execute=dashboardUsersForm%3AuserTable
             &javax.faces.partial.render=dashboardUsersForm%3AuserTable
             &dashboardUsersForm%3AuserTable=dashboardUsersForm%3AuserTable
             &dashboardUsersForm%3AuserTable_sorting=true
             &dashboardUsersForm%3AuserTable_skipChildren=true
             &dashboardUsersForm%3AuserTable_encodeFeature=true
             &dashboardUsersForm%3AuserTable_sortKey=dashboardUsersForm%3AuserTable%3Aj_idt240
             &dashboardUsersForm%3AuserTable_sortDir=1
             &dashboardUsersForm=dashboardUsersForm
             &dashboardUsersForm%3AselectedPage=1
             &dashboardUsersForm%3AsearchFiles=
             &javax.faces.ViewState=-1703367417622560877%3A-8325075942681282905'
```
So the heavy lifting happens on the backend, I guess it simply runs a SQL query, and I do not think that PSQL would have a problem with sorting 30K records. 

Guess however does not provide a good evidence: if you happen to have a tip how to generate 50K fake users, I could follow that and can test it against them.

pdurbin:
> if you happen to have a tip how to generate 50K fake users

It would be time consuming and you'd need to write a script but it's possible to create users via API. Here are the docs: http://guides.dataverse.org/en/4.20/api/native-api.html#create-a-builtin-user

pkiraly:
Thanks @pdurbin, I'll write script to generate fake users, and report the results

pdurbin:
@pkiraly my other thought was that if you're ok (for now) with an API-based solution, maybe you could use the "list users" API since it has "createdTime" in the output: http://guides.dataverse.org/en/4.20/api/native-api.html#list-users

pkiraly:
I've generated some thousand users, and it turned out that the sorting doesn't do what I expected: it sorts only the current page, not the whole database table.

So I close this PR as an unsuccessful approach. Unfortunately I am not familiar with Primefaces to imagine an alternative solution.

pkiraly:

PR #6847, opened 2020-04-21
5716 edit file api perms

### LINKED ISSUE #5716: Update DataFile creation/editing API permissions  to superusers only
Currently (via api) normal user can edit md5s and other file information that should be generated by Dataverse itself. Also users cana DB entry without a file on disk. This functionality is still needed for super-users but should be disabled otherwise.
   - Current known locations are import & create dataset
      - Import should already be superuser only
   - Error if a normal user uploads json that violates this
   - Add docs to make clear for superusers that they can create detached files

This was originally part of #5674 but broken off as it was not in scope.

sekmiller:
This PR restricts the use of the import dataset and create dataset (with files) to accounts with superuser permissions. This is needed to strictly limit the instances in which  file metadata/datafile records are created in the absence of a physical file.

Closes #5716 

verify that a non-super user may still use the create dataset api as long as files are not included.

No

**Additional documentation**:


sekmiller:
@landreev I thought there was going to be more to it than that, but after writing the test I realized that the update version has never been used to add files so I left it at that.

landreev:
> I thought there was going to be more to it than that, but after writing the test I realized that the update version has never been used to add files so I left it at that.

Makes sense. And it's still a useful test.

PR #6860, opened 2020-04-28
Destroy Dataset with a Mapped DataFile

### LINKED ISSUE #4093: Destroy Dataset: Cannot destroy a dataset that has a mapped shape file.
This is currently in production, v4.7.1

Server log error:

Local Exception Stack: 
Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.5.2.v20140319-9ad6abd): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: org.postgresql.util.PSQLException: ERROR: update or delete on table "dvobject" violates foreign key constraint "fk_maplayermetadata_datafile_id" on table "maplayermetadata"
  Detail: Key (id)=(3047705) is still referenced from table "maplayermetadata".
Error Code: 0
Call: DELETE FROM DVOBJECT WHERE (ID = ?)
        bind => [1 parameter bound]
Query: DeleteObjectQuery([DataFile id:3047705 name:null])
        at org.eclipse.persistence.exceptions.DatabaseException.sqlException(DatabaseException.java:340)
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.processExceptionForCommError(DatabaseAccessor.java:1611)
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeDirectNoSelect(DatabaseAccessor.java:898)
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeNoSelect(DatabaseAccessor.java:962)
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.basicExecuteCall(DatabaseAccessor.java:631)
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeCall(DatabaseAccessor.java:558)
        at org.eclipse.persistence.internal.sessions.AbstractSession.basicExecuteCall(AbstractSession.java:2002)
        at org.eclipse.persistence.sessions.server.ClientSession.executeCall(ClientSession.java:298)
        at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:242)
        at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.deleteObject(DatasourceCallQueryMechanism.java:203)
        at org.eclipse.persistence.internal.queries.StatementQueryMechanism.deleteObject(StatementQueryMechanism.java:104)
        at org.eclipse.persistence.queries.DeleteObjectQuery.executeDatabaseQuery(DeleteObjectQuery.java:218)
        at org.eclipse.persistence.queries.DatabaseQuery.execute(DatabaseQuery.java:899)
        at org.eclipse.persistence.queries.DatabaseQuery.executeInUnitOfWork(DatabaseQuery.java:798)
        at org.eclipse.persistence.queries.ObjectLevelModifyQuery.executeInUnitOfWorkObjectLevelModifyQuery(ObjectLevelModifyQuery.java:108)
        at org.eclipse.persistence.queries.DeleteObjectQuery.executeInUnitOfWorkObjectLevelModifyQuery(DeleteObjectQuery.java:119)
        at org.eclipse.persistence.queries.ObjectLevelModifyQuery.executeInUnitOfWork(ObjectLevelModifyQuery.java:85

landreev:

Self-explanatory; see issue. 

Closes #4093

It implements cascade delete on both the map layer and tokens; as in the original PR 2.5 yrs ago. In addition, the code is added to the DeleteDataFileCommand, that deletes any data associated with the layer that may exist on the WorldMap side. 

Deploy on demo, destroy all those old, as of now un-destroyable datasets. 

no 
no
**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteDataFileCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteDataFileCommand.java
@@ -24,6 +24,9 @@ import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.harvard.iq.dataverse.GlobalIdServiceBean;
+import edu.harvard.iq.dataverse.UserNotification;
+import java.sql.Timestamp;
+import java.util.Date;
 
 /**
  * Deletes a data file, both DB entity and filesystem object.
@@ -209,6 +212,31 @@ public class DeleteDataFileCommand extends AbstractVoidCommand {
             }
         } catch (Exception e) {
             logger.log(Level.WARNING, "Identifier deletion was not successfull:", e.getMessage());
+        }
+        
+        // If there is a Map Layer associated with this file, we may need to 
+        // try and remove the layer data on the WorldMap side. 
+        if (ctxt.mapLayerMetadata().findMetadataByDatafile(doomed) != null) {
+            // (We need an AuthenticatedUser in order to produce a WorldMap token!)
+            String id = getUser().getIdentifier();
+            id = id.startsWith("@") ? id.substring(1) : id;
+            AuthenticatedUser authenticatedUser = ctxt.authentication().getAuthenticatedUser(id);
+            try {
+                ctxt.mapLayerMetadata().deleteMapLayerFromWorldMap(doomed, authenticatedUser);
+
+                // We have the dedicatd command DeleteMapLayerMetadataCommand, but 
+                // there's no need to use it explicitly, since the Dataverse-side
+                // MapLayerMetadata entity will be deleted by the database 
+                // cascade on the DataFile. -- L.A. Apr. 2020
+
+            } catch (Exception ex) {
+                // We are not going to treat it as a fatal condition and bail out, 
+                // but we will send a notification to the user, warning them 
+                // there may still be some data associated with the mapped layer, 
+                // on the WorldMap side, un-deleted:
+                ctxt.notifications().sendNotification(authenticatedUser, new Timestamp(new Date().getTime()), UserNotification.Type.MAPLAYERDELETEFAILED, doomed.getFileMetadata().getId());
+            }
+
         }
         DataFile doomedAndMerged = ctxt.em().merge(doomed);
         ctxt.em().remove(doomedAndMerged);
--- a/src/main/java/edu/harvard/iq/dataverse/DataFile.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataFile.java
@@ -19,6 +19,7 @@ import edu.harvard.iq.dataverse.util.BundleUtil;
 import edu.harvard.iq.dataverse.util.FileUtil;
 import edu.harvard.iq.dataverse.util.ShapefileHandler;
 import edu.harvard.iq.dataverse.util.StringUtil;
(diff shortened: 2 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteDataFileCommand.java
```diff
@@ -210,6 +213,31 @@ public FileVisitResult postVisitDirectory(final Path dir, final IOException e)
...
+            String id = getUser().getIdentifier();
+            id = id.startsWith("@") ? id.substring(1) : id;
+            AuthenticatedUser authenticatedUser = ctxt.authentication().getAuthenticatedUser(id);
+            try {
+                ctxt.mapLayerMetadata().deleteMapLayerFromWorldMap(doomed, authenticatedUser);
+
+                // We have the dedicatd command DeleteMapLayerMetadataCommand, but 
+                // there's no need to use it explicitly, since the Dataverse-side
+                // MapLayerMetadata entity will be deleted by the database 
+                // cascade on the DataFile. -- L.A. Apr. 2020
```
pdurbin, line 230: I had forgotten that there's a DeleteMapLayerMetadataCommand but this comment makes me wonder why we don't call it. I get that we don't **need** to call it thanks to the cascade, but would calling the command cause problems? Does it do more than we want? It seems to clean up thumbnails, for example. Also, there's tiny typo above: dedicatd.

pdurbin: I'm still curious about these questions but there's no need to hold this up from QA so I moved it over. This pull request adds a lot of value.

djbrooke:
Let's hold off on moving to QA pending the question/typo above

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteDataFileCommand.java
landreev: Sorry, missed the original question 4 hours ago. I guess the answer is, the delete layer command is a more surgical tool; it's for a case where we just need to delete the layer, without touching the file. 
Calling it in the context of trying to wipe everything, including the datafile and all its associated physical files, would not hurt, strictly speaking - but I couldn't think of why we would want to call it either; seeing how the cascade is moire reliable? It does remove the map thumbnails - but then we run a full physical file cleanup after DeleteDataFileCommand, that will delete all aux files associated with the datafile, thumbnails and otherwise. 
And then again there is the fact the command doesn't do everything we need, either. It does not attempt to go out and remove whatever possible data may be sitting on the WorldMap side. I could only *assume* that it was by design. 
My attitude with this PR was to check in something ASAP that works and doesn't break things. 
I generally sympathize with the goals of making things cleaner; calling commands instead of database queries and such. But in the case of WorldMap, the whole is a big hack - isn't it? Our attempt to "think about it some more" to see if we were doing everything right resulted in a 2.5 year delay, and a whole bunch of these un-destructible datasets still sitting in production...

pdurbin: Yes, agreed. I don't think this will break anything and it definitely adds value. Thanks for the explanation.

PR #6893, opened 2020-05-04
6574 filenames

### LINKED ISSUE #6574: File Upload - Allow files of the same name within different directories of a dataset
Within in a dataset, we currently add "-1" to any files uploaded that have the same name. We'll need to change this as we integrate with more reproducibility tools and other systems, where scripts, code, etc/ are expecting the names of files to not change. 

Needs some thought, and may make sense to discuss at the same time as #4813.

---

Status (by @pdurbin). When "fixed" is indicated, it means in the branch below.

- Bugs described below by @mheppler have been reproduced, specifically:
  - From the "edit metadata for files" page, you can give files the same name. Fixed.
  - If you delete some files while uploading, the counting is not reset so you'll get "-3" when it could have been "-2", for example.
- Bugs in the API have been found:
  - It was possible to give rename files to have the same name. Fixed.
  - It was possible to "move" files to directory with a file that has the same name, creating a duplicate. Fixed.
- API tests have been written:
  - uploadTwoFilesWithSameNameSameDirectory
  - uploadTwoFilesWithSameNameDifferentDirectories
  - renameFileToSameName
  - moveFileToDirectoryContainingSameFileName

`6574-filenames` is the branch where fixes are being pushed. Here's the "compare" view: https://github.com/IQSS/dataverse/compare/6574-filenames

pdurbin:

Several bugs were found while investigating filenames:

- From the "edit metadata for files" page, you could give files the same name.
- Via API, it was possible to give rename files to have the same name.
- Via API, it was possible to "move" files to directory with a file that has the same name, creating a duplicate.

Closes #6574

Methods are in IngestUtil because that's where similar methods are found. Previous methods were focused on returning a unique filename such as README-1.md. The new methods are focused on identifying duplicates.

Try exercising the bugs above on develop vs. this branch. The API to use is this one, passing "label" and "directoryLabel" in the JSON: http://guides.dataverse.org/en/4.20/api/native-api.html#updating-file-metadata

Yes, if you try to rename a file to create a naming conflict, you will see an error like this:

Screen Shot 2020-05-01 at 5 10 27 PM

I don't believe so.

**Additional documentation**:

None.


src/main/java/edu/harvard/iq/dataverse/ingest/IngestUtil.java
```diff
@@ -100,6 +101,80 @@ public static String duplicateFilenameCheck(FileMetadata fileMetadata, Set<Strin
...
+     * @param directoryLabel The new directoryLabel (file path) that is being
+     * proposed. Can be null.
+     * @param fileMetadatas The list fileMetadatas to be compared against,
+     * probably from a draft.
+     * @param dataFile The file that is being updated with a new name or path.
+     * @return true if there is a conflict, false otherwise.
+     */
+    public static boolean conflictsWithExistingFilenames(String label, String directoryLabel, List<FileMetadata> fileMetadatas, DataFile dataFile) {
+        List<String> filePathsAndNames = new ArrayList<>();
+        for (FileMetadata fileMetadata : fileMetadatas) {
```
qqmyers, line 120: This loop is findDuplicateFilenames() ?

sekmiller, line 120: would it be possible to combine this with findDuplicateFilenames so that the UI and api are using the same method? You're passing in the filemetadata list from the dataset version which is the param for findDuplicateFilenames

pdurbin: @sekmiller good idea. Fixed in 8d6a5c8e6.

pdurbin:
I addressed code review in 8d6a5c8e6 (reduced duplicate code). I also added the new test class to the API test suite in dc77ac19f and merged the latest from develop.

PR #6901, opened 2020-05-08
ability to reserve DOIs through DataCite #5093

### LINKED ISSUE #5093: Implement ability to reserve DOIs through DataCite
Datacite has added the ability to reserve DOIs pre publish (as EZID did). This ticket will track adding the ability to reserve those DOIs when we create datasets, as well as a one time ability to reserve DOIs for existing drafts that have only reserved their DOIs on the dataverse side.

pdurbin:

We want to reserve PIDs from DataCite when we create datasets.

Closes #5093

- Most of the infrastructure was already there. It hinges on changing registerWhenPublished from true to false.
- We are now using both the old MDS API and the new REST API.

- Configure for DataCite
- Test all the documented APIs

No.

No.

**Additional documentation**:

No.


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1038,6 +1038,20 @@ See also these related database settings below:
...
+
+doi.baseurlstringnext
++++++++++++++++++++++
+
+Dataverse uses multiple APIs from DataCite:
+
+- The DataCite MDS API is older, XML-based, and configured using :ref:`doi.baseurlstring`.
+- The DataCite REST API is newer, JSON-based, and configured using ``doi.baseurlstringnext``.
+
+In production, ``doi.baseurlstringnext`` should be set to ``https://api.datacite.org``
```
qqmyers, line 1051: Since this is specific to DataCite, it might be better to name it for that - doi.dataciteapibaseurl or something. (FWIW - I could swear that at some point I discussed adding a second option versus swapping mds for api in baseurlstring with someone and decided against having a second option, but I can't find code nor conversation...)

kcondon:
Issues found:
X 1. API doc is missing protocol (doi:) from PID in examples.
2. Add release note on adding jvm-option
3. Get info api fails when PID exists with "error" but no other information, nothing in log. Curl method worked.

Issue #6682, opened 2020-02-25
As a developer, I want a button in the README saying the percentage code coverage for the API test suite (integration tests)


pdurbin:
If you look in the README today, you'll see three buttons having to do with automated testing and code coverage that I most recently tweaked in pull request #6272 and that look like this:

![Screen Shot 2020-02-25 at 2 17 27 PM](https://user-images.githubusercontent.com/21006/75279704-0e24c800-57da-11ea-84ca-7b0494e6b445.png)

If we convert these buttons into a table, here's how it would look:

|  | Unit Tests | Integration Tests |
| --- | ------------- | ------------- |
| Status shown (pass vs. fail)  | DONE  | DONE |
| Code coverage measured | DONE  | TODO |

Now that @donsizemore just gave me a taste at http://irclog.iq.harvard.edu/dataverse/2020-02-25#i_119256 of his new tricks, we can see that code coverage in the "develop" branch as of this writing (commit 7227c38) is 41% for **integration tests** (the API test suite), much higher than the 19% we see above for **unit tests**. Here's a screenshot where we see that number at https://jenkins.dataverse.org/job/IQSS-dataverse-develop/ws/target/coverage-it/index.html if you scroll to the bottom:

![Screen Shot 2020-02-25 at 2 16 24 PM](https://user-images.githubusercontent.com/21006/75280254-1b8e8200-57db-11ea-9605-de5f2099805c.png)

I took a quick look at https://jenkins.dataverse.org/job/IQSS-dataverse-develop/badge/ but it's non-obvious to me how to surface that 41% number in a 4th button. The definition of done for this issue is figuring this out. I'm breaking this out, small chunk style, from #6124 which has other details.

The 4th button would look something like this:

![Screen Shot 2020-02-25 at 2 38 01 PM](https://user-images.githubusercontent.com/21006/75281033-84c2c500-57dc-11ea-80b1-f344503c6f04.png)

PR #6907, opened 2020-05-11
6881, 6882 delete on cancel and queue handle upload requests

### LINKED ISSUE #6881: Cancel in create dataset doesn't delete temp files
While hitting 'Cancel' rather than 'Save' when editing datasets properly removes temp files, doing so on the dataset creation page does not. Looking at the code, the 'Cancel' method in DatasetPage just calls returnToLatestVersion which clears the newfiles list but does not delete any files. In contrast, the EditDatafilesPage cancel goes through two loops to delete the newfiles (those in the lower pane) and the uploadedFiles (those in process/still in the upper list). (Actually, in mode SINGLE or SINGLE_REPLACE, EditDatafilesPage doesn't try to delete any files - shouldn't it delete the one upload?).

### LINKED ISSUE #6882: File Upload: Direct S3 intermittently fails on file upload during dataset creation on payara 5.
Uploading files on dataset creation when using S3 has had some issues but many have been fixed. This is a placeholder for remaining ones that appeared to be specific to Payara5.

The basic use case(s) are to upload 3 smaller files while creating a dataset in a dataverse configured to use direct s3 upload. The behavior is it will work sometimes and fail sometimes but in a couple different ways: hung on all files 100% progress, 0% progress, or some 100, some 0.

Try running the same test several times in a row, noting UI behavior, browser log, console and network, and server.log

Have tried as a new user, both in root dv and in a sub dv.

qqmyers:
**What this PR does / why we need it**: It adds logic to delete of temp files when a dataset creation is cancelled and, for direct uploads, avoids parallel calls to handleExternalUploads which seems to cause intermittent failures in Payara. 

Closes #6881
Closes #6882 

**Special notes for your reviewer**: The #6882 fix is just with the new handlingUpload flag and the wait loop to make sure only one is sent at a time. I made 1 PR since both involve using the await sleep call and since I've been testing them together (probably good for formal testing as well).

**Suggestions on how to test this**:For #6682, I think the problem was caused by a second call to handleExternalUpload() occurring while the first was updating the bean state. I think this causes the NPE in the Payara/JSF code, but there is no detail to confirm that. So, I think the only way to test is to just watch for the recurrence of the NPE listed in the issue.
For 6881, the general idea is to hit create dataset, add some files, and hit cancel - the temp files should go away. Since direct and normal uploads  work differently, both should be tested. Similarly, since new temp files are in two lists - basically corresponding to the initial upload list and the list after uploads complete (with the description, path, etc.) - testing when there are files in both lists would be good. There is a difference between direct and normal in how I've fixed this. With normal, cancel is called immediately and all temp files are purged - uploads in progress might throw an exception in the log (since cancelling sends you from the dataset page back to the parent dataverse page, but I think the logic still removes the temp files properly. With directupload,  I've opted to, once the cancel button is hit, to wait for any uploads to S3 to finish so that Dataverse can be told about them and then delete them as with normal upload. If that takes more than ~1 second, I'm disabling the save button and changing cancel to be disabled and say 'cancel in progress'. (A different design might avoid this, but it would involve being able to cancel S3 uploads in progress and detecting whether they indeed were cancelled or finished (and left a file)). In both cases, the user should be redirected to the parent dataverse page and no temp files should remain.

**Does this PR introduce a user interface change?**: button label change and disabling during cancel operation - see the discussion above about direct uploads being cancelled and waiting for more than 1 second. (Potentially could always do this - having someone hit save after cancel would always be bad, but only cancelling direct upload is likely to take enough time that someone might try, but we could always disable save/cancel after one is hit if that's preferable.) 

**Is there a release notes update needed for this change?**: Don't think so - things work basically the same for the user, but temp file cleanup is better.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -115,6 +115,7 @@ import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.lang.StringEscapeUtils;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.apache.commons.io.IOUtils;
 
 import org.primefaces.component.tabview.TabView;
@@ -246,6 +247,9 @@ public class DatasetPage implements java.io.Serializable {
     private Long versionId;
     private int selectedTabIndex;
     private List<DataFile> newFiles = new ArrayList<>();
+    private List<DataFile> uploadedFiles = new ArrayList<>();
+    private MutableBoolean uploadInProgress = new MutableBoolean(false);
+
     private DatasetVersion workingVersion;
     private DatasetVersion clone;
     private int releaseRadio = 1;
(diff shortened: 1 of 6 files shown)


kcondon:
@qqmyers Can't upload files on create to any storage, local or s3, cannot upload files to local or indirect s3 after dataset creation but can on direct s3.

PR #6924, opened 2020-05-20
4813 allow duplicate files

### LINKED ISSUE #4813: File Upload - allow files with same MD5 (or other checksum) in a dataset
If someone uploads a file in a dataset that Dataverse notices already has a file with the same content  (both files have the same MD5), Dataverse shows an error and doesn't allow the "duplicate" file to be uploaded.

<kbd></kbd>

Issues with this feature have been discussed in another github issue (https://github.com/IQSS/dataverse/issues/2955, closed when File Replace was released in Dataverse 4.6.1), in Dataverse's Google Group [here](https://groups.google.com/forum/#!searchin/dataverse-community/duplicate$20file|sort:date/dataverse-community/21TFtn7bBJc/D0_QI6ejBQAJ) and [here](https://groups.google.com/forum/#!searchin/dataverse-community/duplicate$20file|sort:date/dataverse-community/FLnm8-60sOs/gNByf3c4CAAJ), and in [a recent Dataverse support ticket](https://help.hmdc.harvard.edu/Ticket/Display.html?id=263576), where a depositor wrote that "for uploading shape files for two different polygons but the same projection, it might be nice to be able to upload both at the same time." For this researcher, a common workaround, uploading the file in different double-zipped archived files (7-Zip, tar file, etc) won't work because the journal policy doesn't allow depositors to upload archived files.

### LINKED ISSUE #6468: File Replace - "...same file" validation error msg not displayed correctly
On Dataverse 4.18.1, when I try to replace a file with another file that has the same content (same md5), I'm not immediately getting the error message that "You cannot replace a file with the exact same file.":

<kbd>filereplacebug</kbd>

There's no messaging that the replace didn't work or why.

But if I then try to replace the file with a file that **does not** have the same content, the replace succeeds, and the error message that should have appeared when the replace failed the first time appears here:

<kbd>filereplacebug2</kbd>

So it's like the error message for failed attempts only appears once a successful attempt as been made.

sekmiller:
This will allow users to upload multiple files with the same checksum value to a given dataset. On upload either via the interface or api the user will be warned that a duplicate file exists in the dataset (with the current path/label of the duplicate file.) At that point they can directly delete the newly uploaded file - if the upload is via the UI.

#4813 - allow files with the same MD5/checksum to exist in the same dataset

Closes #4813
Closes #6468 

Really wanted to take a stick of dynamite to AddReplaceFileHelper, but ended up working with it as it exists. Also fixed an issue with the editFileMetadata api where if you weren't updating the file's label you'd get an duplicate file name error. This was causing a failure in the .testForceReplaceAndUpdate Test

various scenarios uploading a duplicate file including replace
[document outlining upload use cases and expected messaging](https://docs.google.com/document/d/14w6kyJVB8H2F2_AvEAHoNObTSlVDxu1wRYnZC-yN_yo/edit)

Introduces a popup on upload of a duplicate file, which warns the user and allows them to immediately delete the newly uploaded file

We could note that duplicate files within a dataset are now allowed as a new feature

**Additional documentation**:


sekmiller:
I updated the messaging based on discussion during the Design Meeting.

djbrooke:
@TaniaSchlatter @mheppler I'm going to move this back to In Progress based on the discussions in the design meeting. Thanks for the review.

@TaniaSchlatter - if there are text changes needed, I'm happy to make them directly on the branch as you guide. Let me know if this would be helpful.

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -147,6 +146,18 @@ File Handling
 
 Certain file types in Dataverse are supported by additional functionality, which can include downloading in different formats, previews, file-level metadata preservation, file-level data citation; and exploration through data visualization and analysis. See the sections below for information about special functionality for specific file types.
 
+Duplicate Files
+===============
+
+Beginning with Dataverse 5.0, the way Dataverse handles duplicate files (filename and checksums) is changing to be more flexible. Specifically:
+
+- Files with the same checksum can be included in a dataset, even if the files are in the same directory.
+- Files with the same filename can be included in a dataset as long as the files are in different directory.
```
mheppler, line 155: This should read "...in different directories."

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -147,6 +146,20 @@ File Handling
...
+.. _duplicate-files:
+
+Duplicate Files
+===============
+
+Beginning with Dataverse 5.0, the way Dataverse handles duplicate files (filename and checksums) is changing to be more flexible. Specifically:
+
+- Files with the same checksum can be included in a dataset, even if the files are in the same directory.
+- Files with the same filename can be included in a dataset as long as the files are in different directories.
+- If a user attempts to add a file to a directory where a file already exists with that directory/filename combination, Dataverse will adjust the file path and names by adding "-1" or "-2" as applicable. This change will be visible in the list of files being uploaded.
```
scolapasta, line 158: is it worth being more explicit here about upload vs edit? i.e. related to the next bullet about changing the directory, is also changing the name after upload. (so the suggestion is to add that there, and change add to upload here)

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
```diff
@@ -869,9 +869,57 @@ private String getBundleString(String msgName){
     public void deleteFilesCompleted(){
         
     }
+    
+    public void deleteMarkedAsDuplicateFiles() {
```
scolapasta, line 873: does this need to be a separate method? or could it call delete files after setting the selected files from the files with same checksum? Or maybe better yet, have a private delete method that takes a list, have this call this with the list of files with same checksum and have the other (regular) delete call the private method with the list of selected files. (you also have to be careful to differentiate between delete of newly uploaded vs pre-existing, but I think in the end it would be cleaner (and less duplicate) code. Let me know what you think of this suggestion.

src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
```diff
@@ -472,6 +472,22 @@ public void setSelected(boolean selected) {
...
     
+    /*
+    For use during file upload so that the user may delete 
+    files that have already been uploaded to the current dataset version
+    */
+    
+    @Transient
+    private boolean markedAsDuplicate;
+
+    public boolean isMarkedAsDuplicate() {
```
scolapasta, line 483: is this for checksums? if so, do we need, since we already have it on the datafile object?

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -378,9 +378,16 @@ public Response updateFileMetadata(@FormDataParam("jsonData") String jsonData,
                     // on *new* datafiles, that haven't been saved in the database yet;
                     // but it should never be the case in the context of this API) 
                     // -- L.A. Mar. 2020
+                    //SEK 5/2020 - we can't use checksum because it
```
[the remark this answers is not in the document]
scolapasta: did you check if we can just remove?

src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
sekmiller: We actually do need it because there's a separate list of file metadata that feed the uploaded files table and if the user deletes the duplicates we have to remove them too. I was also using it to write the success message, but since were no longer using names there I did some code simplification there.

scolapasta: But can't you call fileMetaData.datafile.isMarkedAsDuplicate()? or something like that?

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
sekmiller: Makes sense. i'll take a look. I know we have different success messages - that were introduced yesterday, but it might not be too bad to consolidate.

sekmiller: Was able to use only one method for delete. Ran into a weird situation - real edge case - where if you uploaded a dupe, allowed it to stay, uploaded it again, decided no once is enough so deleted the new one it would also delete the first dupe that was "ok" - so fixed that, allowing the first dupe to stay.

doc/sphinx-guides/source/user/dataset-management.rst
sekmiller: I made a change to the doc here, please see that it helps to add some clarity.

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
scolapasta: Turns out the one method way (and checking if selected to determine which delete is wanted) is problematic, so we do still need two public methods. BUT they just need to "collect" the files and then call a private delete which still centralizes the code. @sekmiller's aware of the change needed, and will move to QA once that is all set.

kcondon:
So, I tested all the rules and used cases as best I could. I think they all work except on upload when paths are involved for duplicates that have not yet been saved.
1. Upload file2.txt twice, edit paths to be c,d. Cannot save, says duplicate filenames.
This works if we upload a zip with dupe filenames in different paths.

sekmiller:
I was able to upload duplicate files and edit their paths and successfully save them
Screen Shot 2020-08-03 at 9 45 58 AM

PR #6931, opened 2020-05-22
add delay for direct S3 upload of tiny files

### LINKED ISSUE #6930: File upload: Direct S3 upload has intermittent failures when uploading larger number of files.
Direct S3 file upload was recently tested using 100 small files, uploaded simultaneously. While smaller numbers of files appears to be reliable, the larger 100 file set has shown intermittent failures. Attached is a file containing 3 tests and their results.
[directs3100fileuploaderr.txt](https://github.com/IQSS/dataverse/files/4670035/directs3100fileuploaderr.txt)

qqmyers:
**What this PR does / why we need it**: slows down reporting uploads of time files to avoid concurrency issue(s) in Dataverse (see issue)

Closes #6930 

**Special notes for your reviewer**: the delay is currently 500 msec per file, could be made longer if the problem persists. In my testing 200 msec was too short (still saw issues sometimes) so it can't get shorter than that. FWIW, prior updates should have made sure that there is only one call from the fileupload.js to Dataverse at a time, so I'm assuming that there's some processing in Dataverse going on after the HTTP call returns, during which a new call still causes problems. If there are other ways to handle this, I'd be interested to know what else might be done. (This should be effective, but it would be nice not to just have an arbitrary delay.)

**Suggestions on how to test this**: same as for previous tests - with < 1K files (26 bytes is fine)

### DIFF
--- a/src/main/webapp/resources/js/fileupload.js
+++ b/src/main/webapp/resources/js/fileupload.js
@@ -49,7 +49,7 @@ function setupDirectUpload(enabled) {
     var callback = function(mutations) {
       mutations.forEach(function(mutation) {
         for(i=0; i<mutation.addedNodes.length;i++) {
-          //Add a listener on any replacedment file 'select' widget
+          //Add a listener on any replacement file 'select' widget
           if(mutation.addedNodes[i].id == 'datasetForm:fileUpload_input') {
             fileInput=mutation.addedNodes[i];
             mutation.addedNodes[i].addEventListener('change', function(event) {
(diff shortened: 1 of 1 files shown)


kcondon:
Still seeing null ptr/index out of bounds on 25-100 small file uploads, sometimes completes with one file missing, sometimes halts part way. Had experimented with bumping up timeout to 2000ms but it did not solve the problem.

qqmyers:
@kcondon - ready for another/last try - I found a place where the code wasn't waiting before requesting new upload URLs, which could then cause the concurrency/state issue as before. I can definitely see a slow-down in the rate for small files (where the upload and hash times are negligible). Given that the other changes to stop concurrent things seemed to reduce the occurrences for me, I'm hopeful that this fix will hit the ones you're seeing.

PR #6960, opened 2020-06-04
Update codebook2-0.xsl

### LINKED ISSUE #6959: Bug in DDI Codebook Export XSLT
https://github.com/IQSS/dataverse/blob/66ed398483679081cd723a807ad85f35c2221097/src/main/resources/edu/harvard/iq/dataverse/codebook2-0.xsl#L180

"ICPSR" is hardcoded at Line 180. From the document format, I think the distributor must be the intended information for that place, since it's paired with the DOI. It appears on the first line of the formatted Codebook output.

I have a pull request coming with a fix to the xsl. I'm afraid I may have mucked up the order of doing this, but I did my best! It's my first Dataverse pull request.

johnhuck:
I have replaced a hardcoded value ("ICPSR") with an XPath acquired value for the distributor name. I have made a presumption that the distributor name is the appropriate element to present here, since it is paired with the DOI.

This fixes the XSLT that prints the first line of the Codebook output.

This fixes a bug

Closes #6959

This is my first Dataverse pull request, so let me know if I am missing a step or otherwise blundering about. I'm not sure if I've gone about this correctly at all.

Apply the stylesheet to a DDI document? That's what I did. Or perform a metadata export (Codebook Export) and check the first line.

No

No

**Additional documentation**:

### DIFF
--- a/src/main/resources/edu/harvard/iq/dataverse/codebook2-0.xsl
+++ b/src/main/resources/edu/harvard/iq/dataverse/codebook2-0.xsl
@@ -177,7 +177,7 @@ font-size: 10pt;
             </head>
             <body bgcolor="#ffffff" lang="en-US">
                 <h1>
-                    <xsl:value-of select="stdyDscr/citation/titlStmt/titl"/> (ICPSR <xsl:value-of
+                    <xsl:value-of select="stdyDscr/citation/titlStmt/titl"/> (<xsl:value-of select="stdyDscr/citation/distStmt/distrbtr"/><xsl:text> </xsl:text><xsl:value-of
                         select="stdyDscr/citation/titlStmt/IDNo"/>)<xsl:if test="stdyDscr/citation/titlStmt/altTitl">
                         <br/>(<xsl:value-of select="stdyDscr/citation/titlStmt/altTitl"/>)</xsl:if>
                 </h1>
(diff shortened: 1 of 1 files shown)


djbrooke:
Hey @johnhuck, thanks for the bug report and the PR! We'll review it!

cc: @jggautier

johnhuck:
Today I find myself still pondering whether Distributor is the right information to replace with here. Hard to know what the original intention behind the stylesheet was. But since "ICPSR" was hard-coded, it must represent something invariable, so author/creator seems less likely. If there were any doubts about this, though, the "ICPSR" string could simply be snipped out of the original stylesheet, leaving the DOI presented on its own in the brackets.

djbrooke:
@jggautier it would be good if you could provide some thoughts and possibly direction here. Let me know if you need any help.

@johnhuck I'll move this back to community dev for now. It looks like there may be a build failure as well, but we'll hear from Julian about the proper implementation first. :)

jggautier:
> Today I find myself still pondering whether Distributor is the right information to replace with here. Hard to know what the original intention behind the stylesheet was. But since "ICPSR" was hard-coded, it must represent something invariable, so author/creator seems less likely. If there were any doubts about this, though, the "ICPSR" string could simply be snipped out of the original stylesheet, leaving the DOI presented on its own in the brackets.

Hi @johnhuck. I was just about to email you and basically ask the same question. Let's continue the conversation in the GitHub issue.

djbrooke:
Thanks @jggautier! 🙏

johnhuck:
I will submit a fresh PR based on discussion in https://github.com/IQSS/dataverse/issues/6959

PR #6966, opened 2020-06-05
reorder links to get firefox to use larger favicon

### LINKED ISSUE #6965: Due to order and Firefox bug, Firefox gets wrong favicon
https://bugzilla.mozilla.org/show_bug.cgi?id=751712 causes Firefox to get the 16 pixel favicon for Dataverse instead of the 32 pixel one. Simple PR to follow.

qqmyers:
https://bugzilla.mozilla.org/show_bug.cgi?id=751712

**What this PR does / why we need it**: works-around a firefox bug to display better favicon

Closes #6965

**Suggestions on how to test this**: Firefox only as far as I know. The code change is just to put the 32 pixel favicon line in the source after the 16 pixel one. So you can check the source. That should also change the favicon itself but that could potentially be cached in your browser.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: just shows better favicon

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dataverse_template.xhtml
+++ b/src/main/webapp/dataverse_template.xhtml
@@ -25,8 +25,8 @@
 	    <ui:insert name="jsonld_header"></ui:insert>
         </f:facet>
         <link rel="apple-touch-icon" sizes="180x180" href="#{resource['images/fav/apple-touch-icon.png']}"/>
-        <link rel="icon" type="image/png" sizes="32x32" href="#{resource['images/fav/favicon-32x32.png']}"/>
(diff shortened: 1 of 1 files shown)


mheppler:
@qqmyers would you be so kind as to provide some screenshots? Have you confirmed this change does not effect other browsers? I am looking at the favicon in my browser tabs, and I can't tell if it's one size or another. The default Dataverse logo favicon used on the Harvard Dataverse (4.20) looks sharp.

Screen Shot 2020-06-05 at 4 13 43 PM

qqmyers:
image
That's from @rdemgenski who saw the effect, only on Firefox and confirmed that the PR fixed the issue.

PR #6968, opened 2020-06-05
allow changes to file metadata via API #6962

### LINKED ISSUE #6962: Unable to change file metadata via API if not also changing label or directoryLabel
I'm aware of #6846 but it was opened before pull request #6893 was merged. It was failing with a 500 error.

As of 97e7a60 it's failing with a 400 error and "Filename already exists at data.tsv" which makes me suspicious of pull request #6893. Here's where I added an extra pretty print to see that error:

Screen Shot 2020-06-05 at 3 24 24 PM

pdurbin:

A regression was introduced in pull request #6893 whereby modifying file metadata API failed with "Filename already exists". This bug is in develop, not production.

Closes #6962

None.

- Try updating file metadata via API in develop vs this pull request. Docs: http://guides.dataverse.org/en/4.20/api/native-api.html#updating-file-metadata
- Ensure the FilesIT is passing

No.

No.

**Additional documentation**:

None.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Files.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Files.java
@@ -392,6 +392,11 @@ public class Files extends AbstractApiBean {
                 javax.json.JsonObject jsonObject = jsonReader.readObject();
                 String label = jsonObject.getString("label", null);
                 String directoryLabel = jsonObject.getString("directoryLabel", null);
+                // If the user is trying to change the label/directoryLabel or not.
+                boolean labelChange = true;
+                if (label == null && directoryLabel == null) {
+                    labelChange = false;
+                }
                 String path = "";
                 if (directoryLabel != null) {
                     path = directoryLabel + "/";
@@ -399,7 +404,7 @@ public class Files extends AbstractApiBean {
                 if (label == null) {
                     label = df.getFileMetadata().getLabel();
                 }
-                if (IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
+                if (labelChange && IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
                     String incomingPathPlusFileName = path + label;
                     return error(BAD_REQUEST, BundleUtil.getStringFromBundle("files.api.metadata.update.duplicateFile", Arrays.asList(incomingPathPlusFileName)));
                 }
--- a/src/main/java/edu/harvard/iq/dataverse/api/Files.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Files.java
@@ -399,14 +399,14 @@ public class Files extends AbstractApiBean {
                 }
                 String oldLabel = df.getFileMetadata().getLabel();
                 String oldDirectoryLabel = df.getFileMetadata().getDirectoryLabel();
-                String oldPathPlusName = oldDirectoryLabel + oldLabel;
+                String oldPathPlusName = oldDirectoryLabel + "/" + oldLabel;
                 if (directoryLabel == null) {
                     directoryLabel = oldDirectoryLabel;
                 }
                 if (label == null) {
                     label = oldLabel;
                 }
-                String incomingPathPlusName = directoryLabel + label;
+                String incomingPathPlusName = directoryLabel + "/" + label;
                 if (oldPathPlusName.equals(incomingPathPlusName)) {
                     labelChange = false;
                 }
--- a/src/main/java/edu/harvard/iq/dataverse/api/Files.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Files.java
@@ -390,31 +390,20 @@ public class Files extends AbstractApiBean {
 
                 JsonReader jsonReader = Json.createReader(new StringReader(jsonData));
                 javax.json.JsonObject jsonObject = jsonReader.readObject();
-                String label = jsonObject.getString("label", null);
-                String directoryLabel = jsonObject.getString("directoryLabel", null);
-                // If the user is trying to change the label/directoryLabel or not.
-                boolean labelChange = true;
-                String oldLabel = df.getFileMetadata().getLabel();
-                String oldDirectoryLabel = df.getFileMetadata().getDirectoryLabel();
-                String oldPathPlusName = oldDirectoryLabel + "/" + oldLabel;
-                if (directoryLabel == null) {
-                    directoryLabel = oldDirectoryLabel;
-                }
-                if (label == null) {
-                    label = oldLabel;
-                }
-                String incomingPathPlusName = directoryLabel + "/" + label;
-                if (oldPathPlusName.equals(incomingPathPlusName)) {
-                    labelChange = false;
-                }
-                logger.fine("For file id " + df.getId() + " user is trying to change the label: " + labelChange);
-                if (labelChange && IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
-                    String pathPlusFilename = "";
-                    if (directoryLabel != null) {
-                        pathPlusFilename = directoryLabel + "/" + label;
-                    } else {
-                        pathPlusFilename = label;
+                String incomingLabel = jsonObject.getString("label", null);
+                String incomingDirectoryLabel = jsonObject.getString("directoryLabel", null);
+                String existingLabel = df.getFileMetadata().getLabel();
+                String existingDirectoryLabel = df.getFileMetadata().getDirectoryLabel();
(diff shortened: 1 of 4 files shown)


pdurbin:
I'm putting this back in code review after fixing a bug @scolapasta found. Now an API user can pass in the existing label or directoryLabel and we still allow edits to go through.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -392,14 +392,26 @@ public Response updateFileMetadata(@FormDataParam("jsonData") String jsonData,
...
                 String directoryLabel = jsonObject.getString("directoryLabel", null);
+                // If the user is trying to change the label/directoryLabel or not.
+                boolean labelChange = true;
+                if (label == null && directoryLabel == null) {
+                    labelChange = false;
+                }
+                String oldLabel = df.getFileMetadata().getLabel();
+                String oldDirectoryLabel = df.getFileMetadata().getDirectoryLabel();
+                String oldPathPlusName = oldDirectoryLabel + oldLabel;
+                String incomingPathPlusName = directoryLabel + label;
```
scolapasta, line 403: does this do weird things if either one is null? i.e I would expect if you only passed on in, and it was the same, then there should be no labelChange, but I think the code for the equals would be false.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -392,14 +392,26 @@ public Response updateFileMetadata(@FormDataParam("jsonData") String jsonData,
...
+                }
                 String path = "";
                 if (directoryLabel != null) {
                     path = directoryLabel + "/";
                 }
                 if (label == null) {
-                    label = df.getFileMetadata().getLabel();
+                    label = oldLabel;
                 }
-                if (IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
```
scolapasta, line 402: if directoryLabel is not passed in, do we want path to be ""? or keep the old value (as you do with label)?

pdurbin: `path` is only used in the error message...

- "Filename already exists at README.md"
- "Filename already exists at code/README.md"

... and I'm happy with how it works. So yes, I think leaving it as "" (empty string) is fine.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
pdurbin: I figured what matters in terms of if there was a change or not is if the "path + filename" is different, so that's what I'm comparing. No, weird things do not happen if one is null, according to my testing.

pdurbin:
@scolapasta I responded to your comments.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
scolapasta: Sorry, maybe what I wrote wasn't clear, so let me use an example:
oldDirectory = "directory" and old label = "label"

If I only path in a label as "label" (i.e. no directory), I would expect it to allow it, as I am not changing anything. But the label check would compare "directorylabel" with "nulllabel" and so would run the check when it shouldn't.

Also if (for some reason) for new values you passed directory = "director" and label = "ylabel", the check would find them equal and so not check names, even though you are changing the names. 

So the check also needs to have a "/" in there.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
scolapasta: Sorry path confused me here, what I'm getting it is:

Why don't we have have a 
`
                if (directoryLabel == null) {
                    directoryLabel = oldDrectorylLabel;
                }
`

like with label, otherwise null gets passed into the directory+name check, no? (and if we did this, we could really get rid of path altogether and just use directoryLabel + /. And this way if directorLabel were null, we would get a more accurate message anyway:

"Filename already exists at code/README.md", when "Filename already exists at README.md" isn't accurate?

src/main/java/edu/harvard/iq/dataverse/api/Files.java
pdurbin: Good catch. I fixed both of these.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
pdurbin: I got rid of the extra path variable and made sure the messages are accurate.

pdurbin:
I responded to code review with a couple commits and comments.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -392,16 +392,32 @@ public Response updateFileMetadata(@FormDataParam("jsonData") String jsonData,
...
+                }
+                String incomingPathPlusName = directoryLabel + "/" + label;
+                if (oldPathPlusName.equals(incomingPathPlusName)) {
+                    labelChange = false;
                 }
-                if (IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
-                    String incomingPathPlusFileName = path + label;
-                    return error(BAD_REQUEST, BundleUtil.getStringFromBundle("files.api.metadata.update.duplicateFile", Arrays.asList(incomingPathPlusFileName)));
+                if (labelChange && IngestUtil.conflictsWithExistingFilenames(label, directoryLabel, fmdList, df)) {
+                    String pathPlusFilename = "";
```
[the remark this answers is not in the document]
pdurbin: Because if we do (I just tried), we get "null/" as part of the output. Tests fail with messages like this:

```
Expected: Filename already exists at README.md
  Actual: Filename already exists at null/README.md
```

That is to say, that code (below) is there to make sure there's no "null/" in the output, like above.

    String pathPlusFilename = "";
    if (directoryLabel != null) {
        pathPlusFilename = directoryLabel + "/" + label;
    } else {
        pathPlusFilename = label;
    }

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -392,16 +392,32 @@ public Response updateFileMetadata(@FormDataParam("jsonData") String jsonData,
                 javax.json.JsonObject jsonObject = jsonReader.readObject();
                 String label = jsonObject.getString("label", null);
                 String directoryLabel = jsonObject.getString("directoryLabel", null);
-                String path = "";
-                if (directoryLabel != null) {
-                    path = directoryLabel + "/";
+                // If the user is trying to change the label/directoryLabel or not.
+                boolean labelChange = true;
+                if (label == null && directoryLabel == null) {
```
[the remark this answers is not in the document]
pdurbin: Good idea. Removed in 491caff84.

pdurbin:
I removed the code we don't need (good catch) and left a comment about the output.

pdurbin:
Not to self, while old directory is null, pass in "" as the directory.

src/test/java/edu/harvard/iq/dataverse/api/DuplicateFilesIT.java
```diff
@@ -419,4 +420,77 @@ public void existingDirectoryPassLabelChangeDescription() throws IOException {
...
+     * directory as “” (because you don’t realize you can just not pass it).
+     * when it check and compares, the old directory is null. so will that mean
+     * labelChange = true and it will fail even though you didn’t really change
+     * the directory?
+     *
+     * While "labelChange" does end up being true,
+     * IngestUtil.conflictsWithExistingFilenames returns false, so the change is
+     * allowed to go through. The description is allowed to be changed and the
+     * directoryLabel remains null even though the user passed in an empty
+     * string, which is what we want.
```
scolapasta, line 436: The issue here is that if it gets to the conflictswithexisting stage and that returns false, I think we have a problem:

if conflictswithexisting filenames returns false, it’s because it’s treating ‘’ and null differently, since the name is the same). So if you did the same thing, but actually changed the name of the file to a different file name that does exist? Wouldn’t conflictswithexisting still return false then and allow you to rename to an already existing file.

The use case is you have two files:
• the existing values are null for directory for both, names are file1 and file2
• call the api to edit file1 with “” as directoryLabel and label as “file2"
• I’m concerned that labelChange would be true and conflictswithexisting would be false and that would allow you to have the same null/file2 for both.

scolapasta: I'll add, one solution could be to not store null for directoryLabel ever. We could either store "" or even "/". This would also help in adding a unique constraint, since unique constraints are problematic with nulls.

pdurbin:
I attempted to simplify the code in 2debed2. DuplicatesIT and FilesIT both pass.

PR #6973, opened 2020-06-09
Optional use of header to get user IP address

### LINKED ISSUE #6456: Feature Request: re-enable IP group usage with X-Forwarded-For header
This is a successor to #1368 and a follow up to #2826.
Mentioning @pdurbin, @landreev, @djbrooke, @scolapasta, @michbarsinai 

----

## Context
When running on Kubernetes, you are most likely not going to use Apache + AJP as your reverse proxy. This will either be [nginx](https://github.com/kubernetes/ingress-nginx) or [traefik](https://docs.traefik.io/providers/kubernetes-ingress) on private clouds, or other ingress controllers like [Amazon ALB](https://docs.aws.amazon.com/de_de/eks/latest/userguide/alb-ingress.html) on public clouds.

Running on Kubernetes, the application server will never be exposed directly in production setups. (If so, this is a bad idea). Thus you have no option to use IP groups at all in these setups. (For a production grade Docker deployment this should be true, too)

## Proposed solution
1. Create a database setting `:EnableProxyIPGroups`, defaulting to `off`
2. Modify [`DataverseRequest` constructor](http://github.com/poikilotherm/dataverse/blob/4779395ffbc7df59e34eddc8c4878c4e86c11c59/src/main/java/edu/harvard/iq/dataverse/engine/command/DataverseRequest.java#L20-L35) to enable processing [`X-Forwarded-For` header](https://en.wikipedia.org/wiki/X-Forwarded-For) when setting enabled.
  - Maybe add safeguards about trusted proxy addresses (living in the header value)
  - Maybe add safeguards that proxy shall have a RFC1918 address when not specified as trusted IP
3. Describe setting in docs. Add BIG warning about possible abuse and precautions. Hint to relevant doc pieces of proxies about secured usage. (All ingress controllers I know filter this header on incoming requests unless otherwise configured)

## Benefits
Bring back easy groups for advanced setups, otherwise not available. IP groups are very easy to use in many cases.

## Possible drawbacks and risks
As outlined in #1368 and #2826, this header is always a security risk. Because of that, it should only be used when you know what you do, have proper documentation etc.

## Estimated complexity/size/time
Should be fairly easy to do, low hanging fruit. Safeguards are a bit tricky.

## Resources
I would offer a contribution. Usual Code Review / QA process.

### LINKED ISSUE #6970: Enable geo-location for MDC when behind a load balancer
@landreev recently discovered that when Dataverse is run behind load balancers (specifically on AWS), it picks up the IP address of the load balancer internally (in DataverseRequest.java) and that is eventually what's written in the MDC count log, which then breaks MDC's geo-location capability (based on using Geomind's IP database to assign the access/download to a give country/state). 
Fortunately, it looks like AWS send an additional x-forwarded-for header by default that includes the user's IP address. There are other headers that sometimes can be used to send the IP address, i.e.             
            "X-Forwarded-For",
            "Proxy-Client-IP",
            "WL-Proxy-Client-IP",
            "HTTP_X_FORWARDED_FOR",
            "HTTP_X_FORWARDED",
            "HTTP_X_CLUSTER_CLIENT_IP",
            "HTTP_CLIENT_IP",
            "HTTP_FORWARDED_FOR",
            "HTTP_FORWARDED",
            "HTTP_VIA",
            "REMOTE_ADDR" 

Since this may be different for different installations, we've discussed making a configurable option to select which header to check to get the IP address.
We've done a quick test at QDR picking up the x-forwarded-for header using AWS LBs and have verified that the mdc logs get the user ip. I'm working to submit a PR that would add configurability.

Note: By changing the logic in DataverseRequest.java, this will affect anyplace else the requestor's IP is used - seems like that shouldn't be the LB address in any case (i.e. it's not just an MDC issue).

Also note: AFAIK, the IP address is only used in MDC to get the geo-location - it's other functionality to filter robots, stop duplicates, etc. should not be affected/will work with logs prior to this PR even when behind LBs.

qqmyers:
**What this PR does / why we need it**: Allows optional specification of an HTTP header as the source of user ip addresses. This is needed to get valid ip addresses for MDC and IP based access control when running Dataverse behind a proxy/load balancer

Closes #6970 

**Special notes for your reviewer**:  There are security implications if the ip can be spoofed. The header is explicitly not used/ignored if it reports a localhost (127.0.0.1) address

**Suggestions on how to test this**: This should work with/without a header set on a 'normal' test instance (since the header should be null, and should change the reported IP (e.g. as reported in the mdc logs) on a server behind an AWS load balancer when you use 'X-Forwarded-For' from the IP of the load balancer to the correct user IP address. 

**Is there a release notes update needed for this change?**: Yes? Since it affects MDC reporting.

**Additional documentation**:

### DIFF
--- a/doc/release-notes/6970-ip-addresses-behind-proxy
+++ b/doc/release-notes/6970-ip-addresses-behind-proxy
@@ -1,0 +1,6 @@
+### Tracking users' IP addresses behind an address-masking proxy
+
+It is now possible to collect real user IP addresses in MDC logs 
+and/or set up an IP group on a system running behind a proxy/load balancer 
+that hides the addresses of incoming requests. See "Recording User IP 
+Addresses" in the Configuration section of the Installation Guide.
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -1188,10 +1188,11 @@ dataverse.useripaddresssourceheader
 **Make sure** to read the section about the :ref:`Security Implications 
 <user-ip-addresses-proxy-security>` of using this option earlier in the guide!
 
-If set, specifies an HTTP Header such as X-Forwarded-For to use to retrieve the user's IP address. Useful in cases 
-such as running Dataverse behind load balancers where the default option of getting the Remote Address from the servlet isn't correct 
-(e.g. it would be the load balancer IP address). Note that unless your installation always sets the header you configure here, this 
-could be used as a way to spoof the user's address. Allowed values are: 
+If set, specifies an HTTP Header such as X-Forwarded-For to use to retrieve the user's IP address. For example:
+
+``./asadmin create-jvm-options '-Ddataverse.useripaddresssourceheader=X-Forwarded-For'``
+
+This setting is useful in cases such as running Dataverse behind load balancers where the default option of getting the Remote Address from the servlet isn't correct (e.g. it would be the load balancer IP address). Note that unless your installation always sets the header you configure here, this could be used as a way to spoof the user's address. Allowed values are: 
 
 .. code::
 
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -39,6 +39,12 @@ Forcing HTTPS
 
 To avoid having your users send credentials in the clear, it's strongly recommended to force all web traffic to go through HTTPS (port 443) rather than HTTP (port 80). The ease with which one can install a valid SSL cert into Apache compared with the same operation in Payara might be a compelling enough reason to front Payara with Apache. In addition, Apache can be configured to rewrite HTTP to HTTPS with rules such as those found at https://wiki.apache.org/httpd/RewriteHTTPToHTTPS or in the section on :doc:`shibboleth`.
 
+Recording User IP Addresses
++++++++++++++++++++++++++++
+
+By default, Dataverse captures the IP address from which requests originate. This is used for multiple purposes including blocking access to the admin API and Make Data Count reporting. When Dataverse is configured behind a proxy such as a load balancer, the default setting may not capture the correct IP address. ...(Leonid to expand...) 
+
+
 .. _PrivacyConsiderations:
 
 Privacy Considerations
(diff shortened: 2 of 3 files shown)


qqmyers:
Sure. That would be helpful.

 FWIW @landreev and I are looking into issues of multiple values/multiple headers right now.

poikilotherm:
> Sure. That would be helpful.
> 
> FWIW @landreev and I are looking into issues of multiple values/multiple headers right now.

Great. It's quite late already over here, so please let me add my review tomorrow (while you are asleep 😄).

landreev:
@poikilotherm 
> @qqmyers would you like me to provide a review for this? I see some potential for optimization and have opinions 😉

We absolutely want to hear your feedback. 
But just in case one of your questions was why it was being implemented as a JVM, and not a database option - the answer is, it would take an impractical amount of code changes, because of how that DataverseRequest object is created and initialized throughout the application.

qqmyers:
@poikilotherm - based on a side discussion, I've updated the code to handle cases where there may be multiple headers and/or headers with several IP addresses. Not sure how that affects your existing comments but hopefully you can still review. 

We believe that the ordering is always such that the last value is closest to Dataverse. This code should cover the single proxy/load balancer case and enable IP group support with, for example, a campus proxy and load balancer, although in the latter case, MDC gets the campus proxy IP address (since that's what would be needed for IP groups.). If you think the logic is confused or the code isn't clear, we can hop into chat, etc.

poikilotherm:
I was up for hacking away on this in spare time. It's late, so just a first draft: #6977

qqmyers:
Just updated to fix the bug identified (not getting the lastIndexOf(',') in the header string and did some minor cleanup per the comments. If we want this for 5.0 or for back-porting prior to #6977 , I think the code is ready (awaiting documentation).

landreev:
@qqmyers 
> Just updated to fix the bug identified (not getting the lastIndexOf(',') in the header string and did some minor cleanup per the comments. If we want this for 5.0 or for back-porting prior to #6977 , I think the code is ready (awaiting documentation).

Thank you, much appreciated.
I looked at #6977 the other day. I like the idea (will comment there as well). But seeing how the goal appears to be to try and cut the v5 release literally in the next few days, I would much like to use this PR. So I will proceed checking in the doc additions as discussed, and then move it along.

landreev:
I have removed the "closes" annotation for #6456. So that the issue could be used to track a potential filter implementation PR (DRAFT PR #6977)

kcondon:
@qqmyers There are some doc formatting issues: the first explanatory section is one big wall of text with no breaks. Leonid thinks the line breaks need to be formatted differently. The second involves the jvm-option section. It would be good to provide an example command line and the actual argument option list is malformatted -the first option is normally formatted but has a couple extra quotes before the name and then the remaining options are formatted as a quoted pasted section of text, indented with quoted bar.

landreev:
> @qqmyers There are some doc formatting issues: ...

yeah, my blurb is a mess. will address.

kcondon:
Functionality checks out, blocked when not behind lb,  header address added to mdc log
just waiting for doc clean up to merge. Can view built docs here:


http://guides.dataverse.org/en/6970_use_header_for_ip_address/installation/config.html#dataverse-useripaddresssourceheader

http://guides.dataverse.org/en/6970_use_header_for_ip_address/installation/config.html#recording-user-ip-addresses

landreev:
@kcondon 
I made a PR with some cosmetic fixes (https://github.com/QualitativeDataRepository/dataverse/pull/42; I don't have a write on @qqmyers ' repo)

landreev:
@qqmyers did merging my diff overwrite the asadmin example command you had added earlier? (`./asadmin create-jvm-options '-Ddataverse.useripaddresssourceheader=X-Forwarded-For'`) - I'm not seeing it anymore... 
(I fixed a lot of other, unrelated junk in that file)

Issue #6694, opened 2020-02-26
Refactor authentication provider JSON configuration


poikilotherm:
To configure an authentication provider in Dataverse, you need to send a JSON document to the Admin API. [Example for OAuth2](http://guides.dataverse.org/en/4.19/installation/oauth2.html#dataverse-side).

Within the document, the real options for your provider are encoded in a string:
```json
"factoryData":"type: github | userEndpoint: NONE | clientId: FIXME | clientSecret: FIXME",
```

This is not very convenient in terms of validation, readability and extending. With upcoming changes to provide more configuration options for an authentication provider like what to override, what can be trusted etc etc, this should be changed to sth. more sophisticated.

The different providers are also very inconsistent in what you can do with them. For example, OAuth2 providers can have a customizable title, while the Shibboleth provider cannot.

I propose to switch to something like this:

```json
{
    "id":"github",
    "type":"github",
    "enabled":true
    "title": "GitHub",
    "description": {
       "en_US": "This is Github.",
       "de_DE": "Das ist Github."
    },
    "image": "<base64 encoded logo>"
    "options": {
       "client_id": "my-client-id",
       "client_secret": "my-client-secret"
    }
}
```

[**Please find a complete JSON Schema (in draft-07 notation) for details, unambiguity and precision in my gist.**](https://gist.github.com/poikilotherm/a9390d095afa9f139251df3e8b1bab38) It can be used to play for example on https://jsonschemalint.com or any other *draft-07* compliant editor/validator.

**Notes:**
1. You might have noticed, that this includes a necessary change to `AuthenticationServiceBean` to change the concept of `AuthenticationProviderFactory` to provide a factory for every provider we offer instead of grouping the OAuth2 ones. The benefit is the reduced complexity and increased unambiguity.
2. Also removed is the `subtitle` option. It has never been used and should be replaced with `description`. That could offer a better approach for things like those we do for ORCID, but in a more general fashion and more flexible.

pdurbin:
@poikilotherm yes, I've always felt `factoryData` is kind of gross with its pipes and such. A full JSON doc, as you've proposed, would be much cleaner. Awesome that you're thinking along the lines of a JSON schema. 🎉 

I guess I have a few questions:

- Why change the UI at all? Can we leave it alone? And just have this refactoring affect the backend?
- Would you store the information in the database differently? I assume so. If so, would you write SQL upgrade script to migrate the old settings to the new format?
[...]

Issue #3438, opened 2016-10-27
Email Verification: 2 issues; refresh un-verifies, verify email button should always be available.


kcondon:
Users are starting to use this feature and one reported an issue where they had confirmed their email but the account page said not verified and they did not know what to do, see RT 242896

I found a couple things that may impact this:
1. If I'm on the account info page where there is a verify email button, I press it and then after clicking the confirmation button I refresh that page, it asks to resubmit the form and then it "unverifies" me and sends another email.
1. It seems the verify email button does not appear until after the initial confirmation email timeout, default 24hrs. In cases where there is an issue, like this, it may be useful to always display the button.

pdurbin:
I  just noticed this bug at https://github.com/IQSS/dataverse/issues/3405#issuecomment-256977123

bsilverstein95:
Assigning this issue to myself, will work to recreate both of these and go from there. May need to consult the team regarding best practice with changing the Verify Email button's behavior (keeping it visible after confirmation vs having it disappear).

mheppler:
Proposed solution:
- Always show "Verify Email" button on Account pg
- User clicks button, backend logic...
  - checks for and clears expired tokens
  - checks for active tokens
  - if none, send email with new token
  - if active token found, triggers confirmation popup telling them to check their email

chewsw:
May I check if there are any updates on the missing email verification button? I've encountered cases where the email verification button do not appear even after the 24 hours deadline has passed.

pdurbin:
@chewsw hi, no one is working on this bug. If someone out there in the community would like to work on it, I'm happy to point out where in the code to start looking.

TaniaSchlatter:
Notes from discussion.

Situations of use:

- people don't click the link in the welcome email, or the link in the email expires. 
- @donsizemore wants to see/track unvalidated accounts?

Conventional, general account creation flow is that users don't have an account until/unless an account is verified via email. This convention prevents security issues and reduces fake/dead accounts. Users should not be able to do anything without a verified account.

TaniaSchlatter:
@poikilotherm I edited the notes above after further discussion with @pdurbin. Are you thumbs up for having the request account verification email, or for questioning the purpose of unverified accounts?

TaniaSchlatter:
Adding a screenshot of the current state:

[...]

PR #6974, opened 2020-06-10
6936 Implement email domain based groups

### LINKED ISSUE #6936: As an administrator, I would like to create groups of people based on their email address domain
Until some other (better) approaches to auto map user attributes from authentication providers to groups are ready, I would like to introduce a very simple group provider based on mail addresses.

The idea is to take the domain part of a users mail (like `@fz-juelich.de` for me) and place everyone with that domain in a group. What domains are mapped should be configurable.

This is a very simple and basic approach to grant permissions to people coming from your home institution, while externals might access and use your dataverses, too. 

Should of course be able to handle more than one group in multi tenant installations...
And obviously will only do this if email is verified and confirmed... ;-)

If this is of interest for more people out there, I'm happy to contribute this as an upstream patch.

(This is another step to make Jülich DATA ready for real traffic quickly...)

poikilotherm:
Next to explicit groups, IP groups and Shibboleth groups, there is another valuable source of information to which association a user belongs to: a verified email.

Based on the domain part of the email, it's very easy to create groups based on affiliations. This might be handy for installations like the one from Leuven, where different universities share one instance (multi-tenancy). Our use case @ FZJ is to be able to open up for collaboration while automatically granting some rights to staff.

As this is independent from Shibboleth, this might be good for other instances like WZB (ping @RightInTwo), that create users via API.

Closes #6936 

This is not only introducing a new group provider, but also refactors some other places in the codebase. Some noteable places:

- Add a simpler verification check function for email addresses and make use of it. It's much easier to understand now and easier to extend (looking at #6694).
-  Extend the usage of `ExceptionMapper`s in REST API endpoints. There are a lot of places with lots of boilerplate and catch all like try-catch code. This makes the code more verbose and violates [DRY principle](https://dzone.com/articles/software-design-principles-dry-and-kiss). Obviously I didn't touch too much of other code to keep the scope of this PR as small as possible. (But didn't want to encourage the use of catch-alls any further...)

- Deploy, but don't yet publish the root dataverse.
- Create a new mail domain group
- Add the group as member or similar to the root dataverse
- Create a user with a group-matching mail address
- Verify user email
- Get access to the root dataverse
- Delete the permission
- Verify permission is gone

Nope.

It should be mentioned this is a new option to create groups of people.

**Additional documentation**:

Nada.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/authorization/providers/builtin/DataverseUserPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/authorization/providers/builtin/DataverseUserPage.java
@@ -546,6 +546,7 @@ public class DataverseUserPage implements java.io.Serializable {
     
     /**
      * Determines whether the button to send a verification email appears on user page
+     * TODO: cant this be refactored to use confirmEmailService.hasVerifiedEmail(currentUser) ?
      * @return 
      */ 
     public boolean showVerifyEmailButton() {
@@ -557,12 +558,11 @@ public class DataverseUserPage implements java.io.Serializable {
     }
 
     public boolean isEmailIsVerified() {
-
-        return currentUser.getEmailConfirmed() != null && confirmEmailService.findSingleConfirmEmailDataByUser(currentUser) == null;
+        return confirmEmailService.hasVerifiedEmail(currentUser);
     }
 
     public boolean isEmailNotVerified() {
-        return currentUser.getEmailConfirmed() == null || confirmEmailService.findSingleConfirmEmailDataByUser(currentUser) != null;
+        return !confirmEmailService.hasVerifiedEmail(currentUser);
     }
 
     public boolean isEmailGrandfathered() {
--- a/doc/sphinx-guides/source/admin/index.rst
+++ b/doc/sphinx-guides/source/admin/index.rst
@@ -26,6 +26,7 @@ This guide documents the functionality only available to superusers (such as "da
    dataverses-datasets
    solr-search-index
    ip-groups
+   mail-groups
    monitoring
    reporting-tools-and-queries
    maintenance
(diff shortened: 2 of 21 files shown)


poikilotherm:
This is done I think. Can somebody like do a review? Thx!

landreev:
> This is done I think. Can somebody like do a review? Thx!
> BTW it works already on our production instance 😄

Will do, thank you!

src/main/java/edu/harvard/iq/dataverse/authorization/providers/builtin/DataverseUserPage.java
```diff
@@ -546,6 +546,7 @@ public void sendConfirmEmail() {
     
     /**
      * Determines whether the button to send a verification email appears on user page
+     * TODO: cant this be refactored to use confirmEmailService.hasVerifiedEmail(currentUser) ?
```
landreev, line 549: Somebody else here has pointed out that there is a connection with issue #3438; in a sense that, if we rely on users having verified emails for the purposes of group membership, not being able to verify it becomes more of an issue. 
Since you have been using these email groups in production, have you done anything to address #3438? (one possible solutions appear to be to just always show this button below - ?)

poikilotherm: Nope, I didn't. I was going to lean on auth provider support for this. Is this still an issue? It's 4 years old and a lot happened since then. If we encounter these problems again, I could take a look at this...

landreev: It appears that it has never been fixed precisely because it was not much of a serious issue for anyone (I didn't have any recollection of it). And form what I'm reading, that condition only happens if a user tries to refresh the page at the wrong time during the verification... The fact that you haven't heard many complaints about it is somewhat reassuring.
We'll address #3438, now that we have an extra reason to do so. I asked in part because I was wondering if the refactoring you suggested in the comment above might by itself solve it - but no, I don't think so.

poikilotherm:
Yeeeeehaw! :cowboy_hat_face:  Thanks for merging @kcondon

PR #6977, opened 2020-06-11
6970 proxy client ip

### LINKED ISSUE #6456: Feature Request: re-enable IP group usage with X-Forwarded-For header

### LINKED ISSUE #6970: Enable geo-location for MDC when behind a load balancer

poikilotherm:
## :exclamation: THIS IS A DRAFT PR
1. This is just a proof of concept, not yet doing any real work due to stubs.
2. It lacks documentation.
3. It lacks more tests.

**:repeat: This is an alternative solution based on feedback to #6973**

Allows optional specification of an HTTP header as the source of user ip addresses. This is needed to get valid ip addresses for MDC and IP based access control when running Dataverse behind a proxy/load balancer

**Which issue(s) this PR closes:**

Closes #6970
Closes #6456

Nada right now.

- Deploy Dataverse with a reverse proxy
- Configure the system property to match your proxies headers to forward your IP
- Tune log level to fine and watch it log your IP.
- Maybe add an IP group to test if the mangling works as expected.

Nope.

Maybe it should be mentioned.

**Additional documentation**:

Not yet done. Needs big fat warnings in the admin guide.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/authorization/ReverseProxyFilter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/authorization/ReverseProxyFilter.java
@@ -1,0 +1,114 @@
+package edu.harvard.iq.dataverse.authorization;
+
+import edu.harvard.iq.dataverse.util.SystemConfig;
+import org.apache.commons.lang3.StringUtils;
+
+import javax.inject.Inject;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebFilter;
+import javax.servlet.http.HttpFilter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+@WebFilter( filterName = "reverse-proxy", urlPatterns = "/*")
+public class ReverseProxyFilter extends HttpFilter {
+    
+    @Inject
+    SystemConfig systemConfig;
+    
+    String headerField;
+    
+    private static final Logger logger = Logger.getLogger(ReverseProxyFilter.class.getName());
+    private static final String X_FORWARDED_FOR = "X-Forwarded-For"; // de-facto standard
+    private static final String RFC7239_FORWARDED = "Forwarded";
+    private static HashSet<String> ALLOWED_HEADERS = new HashSet<String>(Arrays.asList(
+        X_FORWARDED_FOR,
+        RFC7239_FORWARDED,
+        "Proxy-Client-IP",
+        "WL-Proxy-Client-IP",
+        "HTTP_X_FORWARDED_FOR",
+        "HTTP_X_FORWARDED",
+        "HTTP_X_CLUSTER_CLIENT_IP",
+        "HTTP_CLIENT_IP",
+        "HTTP_FORWARDED_FOR",
+        "HTTP_FORWARDED",
+        "HTTP_VIA",
+        "REMOTE_ADDR" ));
+    
+    @Override
+    protected void doFilter(HttpServletRequest req, HttpServletResponse res, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest proxy = req;
+        
+        // process headers and try to find an ip if a header field is present (=configured)
+        if (headerField != null) {
+            String proxiedIp;
+            
+            // get header(s) matching the field and process if not empty (= not present in request)
+            List<String> headers = Collections.list(req.getHeaders(headerField));
+            logger.log(Level.FINER, "Header(s) \""+headerField+"\": "+StringUtils.join(", "));
+            
+            if ( ! headers.isEmpty()) {
+                // special parsing for X-Forwarded-For de-facto standard
+                if (headerField.equals(X_FORWARDED_FOR)) {
+                    proxiedIp = parseXForwardedFor(headers);
+                }
+                // special parsing for RFC 7239 "Forwarded" syntax
+                else if (headerField.equals(RFC7239_FORWARDED)) {
+                    proxiedIp = parseRFC7239Forwarded(headers);
+                }
+                // otherwise get last header (if multiple), which should have been added by the proxy the client hits
+                // *first*, thus being the real client (and not an intermediate proxy). we assume headers are in order.
+                else {
+                    proxiedIp = validateIp(headers.get(headers.size()-1));
+                }
+                
+                // the ip has been found and verified/validated, let's create a usefull proxied request, allowing
+                // to use getRemoteAddr() and retrieve clients instead of proxies ip.
+                if (proxiedIp != null) {
+                    logger.log(Level.FINE, "Real client IP: \""+proxiedIp+"\"");
+                    proxy = new ProxyRequestWrapper(req, proxiedIp);
+                }
+            }
+        }
+        
+        super.doFilter(proxy, res, chain);
+    }
+    
+    String parseXForwardedFor(List<String> headers) {
+       return null;
+    }
+    
+    String parseRFC7239Forwarded(List<String> headers) {
+        return null;
+    }
+    
+    String validateIp(String ip) {
+        return null;
+    }
+    
+    
+    @Override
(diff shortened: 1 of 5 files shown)


landreev:
Thank you @poikilotherm !
I will look at it first thing in the morning (our morning :) .

src/main/java/edu/harvard/iq/dataverse/authorization/ReverseProxyFilter.java
```diff
@@ -0,0 +1,114 @@
...
+            if ( ! headers.isEmpty()) {
+                // special parsing for X-Forwarded-For de-facto standard
+                if (headerField.equals(X_FORWARDED_FOR)) {
+                    proxiedIp = parseXForwardedFor(headers);
+                }
+                // special parsing for RFC 7239 "Forwarded" syntax
+                else if (headerField.equals(RFC7239_FORWARDED)) {
+                    proxiedIp = parseRFC7239Forwarded(headers);
+                }
+                // otherwise get last header (if multiple), which should have been added by the proxy the client hits
```
poikilotherm, line 70: I'm not so sure about this. Copy-cat from #6973. Shouldn't it be the first header that is added by the first proxy? I have never seen this in real life anyway... :sweat_smile:

qqmyers: It depends on what you're supporting. The global order is that left to right, first to last goes from closer to the user to closer to the server. So taking the first value, when there's no spoofing, should be the client's IP. That would be what you want for MDC, but not what you want for IP groups, where the fact that you're coming through a campus proxy is the key information. With spoofing, a client can add the first/leftmost value, so taking that will get you spoofed. The only way to assure the first/left-most value is good is if you explicitly can trust all the intermediate proxies. 

For the implementation in 6973, to avoid having to come up with a trusted white list of intermediates, we chose to only support the case of at most a single proxy beyond the LB. In those cases, the last/rightmost value is either the users' IP reported by the LB - good for MDC and IP groups, or it is the proxy IP - good for IP groups and ~OK for MDC (hides the true location, but you'd get the geo-location of the proxy, e.g. the campus for most folks.

I think the 'right' way to do this is with a whitelist, reporting the first/left-most trusted value as the IP address and returning the full list of proxy IPs to process for evaluating IP groups. Given the amount of work for that, and the likelihood that very few places will have long proxy chains - (do we know of anyone with more than an LB and a campus proxy?) we chose the strategy describe here.

src/main/java/edu/harvard/iq/dataverse/authorization/ReverseProxyFilter.java
```diff
@@ -0,0 +1,114 @@
...
+    
+    String validateIp(String ip) {
+        return null;
+    }
+    
+    
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        String field = systemConfig.getUserIpAddressSourceHeaderField();
+        if (! ALLOWED_HEADERS.contains(field)) {
```
qqmyers, line 104: comparison should be case insensitive - I'd updated that in the other PR.

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -1056,4 +1056,8 @@ public boolean isDatafileValidationOnPublishEnabled() {
 	public boolean directUploadEnabled(Dataset dataset) {
     	return Boolean.getBoolean("dataverse.files." + dataset.getDataverseContext().getEffectiveStorageDriverId() + ".upload-redirect");
 	}
+	
+	public String getUserIpAddressSourceHeaderField() {
+        return System.getProperty("dataverse.useripaddresssourceheader", null);
```
qqmyers, line 1061: null is the default default so I think you can use System.getProperty("dataverse.useripaddresssourceheader");

src/main/java/edu/harvard/iq/dataverse/authorization/groups/impl/ipaddress/ip/IpAddress.java
```diff
@@ -20,6 +27,22 @@ public static IpAddress valueOf( String s ) {
...
+     * @param s IP address as string representation
+     * @param def A default value to return if s is null or fails to parse.
+     * @return An object to interact with an IP, either the parsed one or the default value.
+     * @throws IllegalArgumentException if def is null.
+     */
+    public static IpAddress valueOf(String s, IpAddress def) {
+        if (def == null) { throw new IllegalArgumentException("Default IP address may not be null."); }
+        try {
+            return valueOf(s);
+        } catch (IllegalArgumentException ex) {
```
qqmyers, line 41: ToDo - Change all the code using IpAddress that expected to receive this - a quick search shows IpAddress is used in workflows and JSON parsing in addition to the DataverseRequest class.

src/main/java/edu/harvard/iq/dataverse/authorization/ReverseProxyFilter.java
qqmyers: I guess a meta question here - given that a filter is independent of Dataverse, does an open source one already exist that we can use? I saw one that just grabs the first value, which has the issue discussed above, but perhaps there is one that handles a whitelist of trusted proxies, etc.?

poikilotherm:
Not relevant anymore

PR #6983, opened 2020-06-15
allow newer java versions (LTS and current), correct interactive logic

### LINKED ISSUE #6982: install.py requires Java 1.8
install.py requires Java 1.8 at https://github.com/IQSS/dataverse/blob/develop/scripts/installer/install.py#L273

At the risk of duplicating #4259 and #5116 I'd like to submit a small PR to remove this check from install.py, or better to simply check for 1.8+

As of 6e2f42597c115cc95e9e5a5790a7ab48c6879fa0 dataverse-ansible ensures that Payara, Solr and Maven launch with openjdk-11 LTS, but I'm dying at this check in install.py

donsizemore:

In testing/using dataverse-ansible, @pallinger correctly pointed out that 1) install.py requires openjdk 1.8, and 2) somehow when openjdk-1.8.0 is specified install.py carries on in ansible without complaint, but when openjdk-11 is specified it hangs waiting for a password, when --noninteractive is passed.

This PR modifies install.py to allow OpenJDK 1.8 (long-term RedHat support), 11 (OpenJDK LTS) and 14 (current OpenJDK, not recommended), and corrects the noninteractive password logic per @landreev 

Closes #6982 

**Special notes for your reviewer**: none come to mind

**Suggestions on how to test this**: re-run install.py with and without --noninteractive, while openjdks 1.8.0, 11 and maybe 14 are set as default

**Additional documentation**: none.

This can definitely wait until after the DCM.


scripts/installer/install.py
```diff
@@ -269,8 +269,8 @@
    # 1d. check java version
    java_version = subprocess.check_output(["java", "-version"], stderr=subprocess.STDOUT).decode()
    print("Found java version "+java_version)
-   if not re.search("1.8", java_version):
-      sys.exit("Dataverse requires Java 1.8. Please install it, or make sure it's in your PATH, and try again")
+   if not re.search('(1.8|11|14)', java_version):
+      sys.exit("Dataverse requires Java 1.8, 11 or 14. Please make sure it's in your PATH, and try again.")
```
pdurbin, line 273: Supporting Java 14 feels a little arbitrary. It's not an LTS release. It just happens to be the current release. My concern is that if we really want to support the latest release of Java, we're going to have to edit this line every six months. I'd be happier with just putting the next LTS release in there, Java 17, which is due out September 2021. Or we could remove Java 14.

Here's a table from Wikipedia that shows all the recent Java releases:

Screen Shot 2020-06-15 at 2 10 42 PM

pdurbin: @donsizemore we talked about this a bit in Slack. Can you please remove Java 14 and only support 8 and 11? Thanks!

PR #6994, opened 2020-06-17
IQSS/6993-use-single-AWS-client-per-store

### LINKED ISSUE #6993: Use single AWS client per S3 storage driver
In working on #6763, where I wanted to access AWS for a give store statically (to asynchronously cancel a multipart upload that isn't associated with an S3AccessIO instance yet - see issue/PR there for more), I realized that the S3 drivers are initializing a new AWS client for each S3AccessIO instance, despite the fact that the info (per driver) doesn't change. Since it should be a useful efficiency improvement on it's own and is testable independent of the overall multipart upload changes, I'm creating a separate issue/PR for it.

qqmyers:
**What this PR does / why we need it**:  Refactors the S3AccessIO class to avoid creating a new AWS client for each instance, and instead keeping track of one AWS client per store. 

Closes #6993 

**Suggestions on how to test this**: This should do nothing :-) - behavior should be the same (except lower memory use perhaps). Since it's S3-related, testing should focus on getting/putting files to S3. Nominally, since the change is related to keeping track of S3 stores correctly, testing should probably use a couple stores and verify that read/writes are still happening to the correct store/bucket.

**Additional documentation**:


qqmyers:
@landreev - I'll be curious as well w.r.t. any memory/performance improvement - could be fun with 1K+ files in a dataset.

In any case - note/reminder for testing - #6995 depends on this PR (includes it) so this should probably be tested/merged first.

Issue #6684, opened 2020-02-25
Layout and styling of action buttons on the dataset and file page


TaniaSchlatter:
Related issues: #3341 #3404 #3815 #6072 #6605 #6604 #6085 #6450 

Design to address the addition of action buttons at the dataset level (scaleability), improve the visibility of edit button, and prepare for responsive design. 

Changes to the dataset action buttons apply to the File page action buttons 

Implementing a new design will provide the opportunity to address accessibility issues found on buttons

Link to mockups: https://sketch.cloud/s/mWGQJ/a/4KZgLZ

Dynamic "run in..." link under the Download/Access button functionality is out of scope for this effort

Design to dos:
- [x] inventory: list of all button options, conditions for appearance, and behavior of states for both dataset and file actions – [version 1](https://docs.google.com/spreadsheets/d/1guVKxR00vS6ICBuJSkvwDefrK9vw_-SJDgp9mi4Jxr4/edit#gid=823971545) | [version 2](https://docs.google.com/spreadsheets/d/12TNSplVTxwBnNm3cj3MwpQW_4djTn9_G9hA01Cz_xq0/edit#gid=183143698)
- [x] review the inventory and designs to determine if all situations/configurations are planned for, or needed additions
- [ ] mock up any common configurations and/or states that are missing if needed 
- [ ] map buttons to bootstrap styles (create custom bootstrap theme if needed)
- [ ] update style guide with new styles

mheppler:
Lots and lots has been moving and changing in order to deliver the improved button style and responsiveness of page layout. Here is a detailed outline of completed improvements and remaining to-do's.

**TEMPLATE**

_Breadcrumbs_
- [x] removed highlighted/bold current page
- [ ] added render logic to hide on root datasets

**FILE PG**

_Responsive layout_
- [x] moved citation, button, metrics, deaccession reason
- [x] improved responsive grid col widths
- [x] added Bootstrap 3 layout order hack (https://stackoverflow.com/a/24834574)

_File Name header_
- [x] increased font size 

_Ingest Warning icon_
- [x] moved next to file name, removed from file
- [x] download-buttons fragment

_Restricted Access/Lock icon_
- [x] moved next to version, publication status labels

_Publication Status labels_
- [x] removed colors, switched to default

_Citations_
- [x] added new button styles, responsive col widths
- [x] CSS word-wrap: break-word for xs viewport + DOI link

_Action buttons_
- [x] added new layout, larger, added "File" to text labels
[...]

PR #7008, opened 2020-06-22
6938 analytics new buttons

### LINKED ISSUE #6938: Analytics - Event tracking code not capture all the btn clicks
Here are is the current list of known analytics event tracking issues for the new action button UI (#6684) on the file and dataset pg. Changes to fix these will require release notes for sysadmins to update their installation's analytics code if they use event tracking code provided in the Installation Guide.

**FILE PG**
- [x] btn-explore, access file btn
- btn-compute ??
- btn-preview ??

**DATASET PG**
- [x] btn-download w/dropdown file table header
- [x] btn-download w/dropdown file table row (tabular)
- [x] btn-explore, access dataset btn
- btn-compute ??
- btn-preview ??

Will review with @qqmyers on any changes as QDR is also currently running the analytics event tracking code.

mheppler:

With the new button UI delivered in  #6684, the analytics functionality was reviewed. This fixes some issues as a result of the new UI, as well as some bugs that were previously not reported. Updates to the sample code provided in the Installation Guide.

Closes #6938 Analytics - Event tracking code not capture all the btn clicks 

These fixes should work for both 4.20 and 5.0 UI's.

Because this analytics code is used by installations that run 4.20, it should be tested on both the new 5.0 btn UI, as well as the previous 4.20 UI.

N/A

Yes, sysadmin who have < 4.20 installations should be notified they can update their existing analytics code to track click events that were being missed.

**Additional documentation**:

N/A


mheppler:
Need to coordinate with @qqmyers and @djbrooke to confirm 4.20 fixes, and write appropriate release notes before getting this to QA.

djbrooke:
@mheppler @qqmyers for the release notes, how do y'all want to handle it? I'm not sure I have enough info to write them from scratch but I can get on a call to discuss if it would be helpful

djbrooke:
@mheppler @qqmyers nm, just re-read. I can pull something together using the notes from 4.16 and 4.20 for your review. Moving back to dev for now.

mheppler:
@kcondon wanted to outline some code clean up in my last commit here. 

- [x] fix publish btn styling on datatset pg, round the squared off border on right side of btn
- [x] fix download btn-group accessibility, add aria-label="File Actions" text to bundle
- [x] remove jim's downloadFile script (#6957) out of file-download-button-fragment.xhtml, move under download-selected btn in dataTable header in filesFragment.xhtml, so the javascript only appears once in source code

That last bullet is the major piece. In the original changes to change large bulk downloads from GET's to POST's, the javascript snippet that was added to the file-download-button-fragment.xhtml resulted in the same javascript snippet being added multiple times to the source code for every file table row, while it only needs to be added to the page once. I moved that snippet to the table header code in filesFragment, so it is grouped under the button that requires it, and it only displays in the source code once. Also followed up with @qqmyers to confirm that the behavior I was seeing in my browser console was correct for a bulk download of 1,500 files. (See screenshot.)

Screen Shot 2020-06-24 at 2 29 18 PM

mheppler:
>The only note I have is that the 'category' reported to Google for the download for Tab files is the label for the specific option rather than the generic 'Download' - probably useful to have the added info.

@qqmyers, after looking at my test Google Analytics dashboard, I am leaning towards useful. Previously the only download click events tracked for the single-click, non-tabular files, had "Download" for both Event Category and Event Action. Providing this additional information in the category still allows the "Download" totals to be filtered in reports under Event Action.

That said, I am happy to learn more about how these events and reports are used for your installations, and if there are further improvements that can be made. We are just starting to think about how we drill into them ourselves, as we've only collected a couple months of analytics from our installation.

kcondon:
@qqmyers gave a good tip for testing: check the comments in the analytics code for which to btns to check:
grep "//" analytics-code.html
<script async="async" src="https://www.googletagmanager.com/gtag/js?id=UA-61753334-3"></script>
  //<![CDATA[
    // Download button
    // Request Access button
    // Explore button
    // Compute button
    // Preview button
    // Cite button
    // Export Metadata button
    // Contact button
    // Share button
      //Special case - the Download button that downloads all selected files in the dataset
        //finds the file id/DOI in the Dataset page
        //finds the file id/DOI in the file page
  //]]>

kcondon:
@mheppler All of the above works except the share button, which is not yet hooked up.

PR #7015, opened 2020-06-24
only iterate over current files in latest published version

### LINKED ISSUE #7014: CuratePublishedDataCommand fails when files have been deleted
The current code iterates through all files in the dataset, not the latest version and, when it finds one that has no filemetadata for the current version, as part of checking to make sure files haven't been added/deleted between the last published version and the draft, flags it as an error. PR to follow.

qqmyers:
**What this PR does / why we need it**: Fixes the ability to re-publish without updating the version on datasets where files have been deleted in prior versions.

Closes #7014 

**Special notes for your reviewer**: a flaw in my original logic

**Suggestions on how to test this**: This class is used for the admin only 'Update Current Version' option under the publish menu that is available when there are metadata only changes. That option should work as before for all datasets. The fix is that, prior to this fix, if you have a dataset in, for example, v2 with fewer/different files than v1 had, when you make a metadata update (update the description) for v2 and try to 'Update Current Version' it would fail and after this PR it should work.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
@@ -85,18 +85,26 @@ public class CuratePublishedDatasetVersionCommand extends AbstractDatasetCommand
         Dataset tempDataset = ctxt.em().merge(getDataset());
 
         // Look for file metadata changes and update published metadata if needed
-        for (DataFile dataFile : tempDataset.getFiles()) {
-            List<FileMetadata> fmdList = dataFile.getFileMetadatas();
+        List<FileMetadata> pubFmds = updateVersion.getFileMetadatas();
+        int pubFileCount = pubFmds.size();
+        int newFileCount = tempDataset.getEditVersion().getFileMetadatas().size();
+        if (pubFileCount != newFileCount) {
+            logger.severe("Draft version of dataset: " + tempDataset.getId() + " has: " + newFileCount + " while last published version has " + pubFileCount);
+            throw new IllegalCommandException("Different number of files in draft version", this);
+        }
+        for (FileMetadata publishedFmd : pubFmds) {
+            DataFile dataFile = publishedFmd.getDataFile();
             FileMetadata draftFmd = dataFile.getLatestFileMetadata();
-            FileMetadata publishedFmd = null;
-            for (FileMetadata fmd : fmdList) {
-                if (fmd.getDatasetVersion().equals(updateVersion)) {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
```diff
@@ -85,18 +85,26 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
 
         // Look for file metadata changes and update published metadata if needed
-        for (DataFile dataFile : tempDataset.getFiles()) {
-            List<FileMetadata> fmdList = dataFile.getFileMetadatas();
+        List<FileMetadata> pubFmds = updateVersion.getFileMetadatas();
+        int pubFileCount = pubFmds.size();
+        int newFileCount = tempDataset.getEditVersion().getFileMetadatas().size();
+        if (pubFileCount != newFileCount) {
+            logger.severe("Draft version of dataset: " + tempDataset.getId() + " has: " + newFileCount + " while last published version has " + pubFileCount);
+            throw new IllegalCommandException("Different number of files in draft version", this);
```
pdurbin, line 93: English that users can see in the UI (including errors) should be moved to a bundle.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
```diff
@@ -85,18 +85,26 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
-            }
             boolean metadataUpdated = false;
-            if (draftFmd != null && publishedFmd != null) {
+            if (draftFmd == null || draftFmd.getDatasetVersion().equals(updateVersion)) {
+                if (draftFmd == null) {
+                    logger.severe("Unable to find latest FMD for file id: " + dataFile.getId());
+                } else {
+                    logger.severe("No filemetadata for file id: " + dataFile.getId() + " in draft version");
+                }
+                throw new IllegalCommandException("Cannot change files in the dataset", this);
```
pdurbin, line 105: English should go in the bundle. In addition, since we have the file id, maybe it would be worth exposing it to the user in the error message, if it will help them troubleshoot the problem.

qqmyers:
@pdurbin - yes the comparison of file counts is just a check. Nominally the UI won't show the button to allow this if they are different but the API doesn't have a separate check.

pdurbin:
@qqmyers belt and suspenders. Perfect.

kcondon:
There is currently a bug with the publish version popup, #7051 , so I tested using the API and it worked.

Issue #6915, opened 2020-05-16
As a repository admin, I would like to disable email confirmation for known domains


RightInTwo:
For our own domain, we synchronize (custom sync, no Shib, no OIDC) the accounts in Dataverse with our institutional accounts. To still use the API to create these users, it would be great to have a switch in the json that says "this email address is valid" and therefore skip the confirmation mail (or at least use a different text for it - just a "welcome" instead a "please verify your email") or set the "email confirmed date" as another way to disable confirmation.

Are there any workarounds that make do with just the API? I'm thinking of just patching our installation and hard-coding the valid domain into authorization/AuthenticationServiceBean.java ->
createAuthenticatedUser(...) with
`authenticatedUser.setEmailConfirmed(emailConfirmedNow);` if the email address ends in our TLD.

"Welcome" emails would be sent separately.

Any ideas?

donsizemore:
related to #5663

pdurbin:
If you're saying you'd like to add another field to the JSON at http://guides.dataverse.org/en/4.20/api/native-api.html#create-a-builtin-user I don't think that would be a huge deal. Here's what's there right now:

```
{
  "firstName": "Lisa",
  "lastName": "Simpson",
  "userName": "lsimpson",
  "affiliation": "Springfield",
  "position": "Student",
  "email": "lsimpson@mailinator.com"
}
```

Maybe something like this would be added:

```
  "emailConfirmed": true,
```

Changes would have to be made to the back end to support this, of course, but I don't think it would be a huge effort.

By the way, there are a couple methods in Admin.java that are used for testing (in ConfirmEmailIT.java):

- startConfirmEmailProcess
- getConfirmEmailToken

scolapasta:
What if this was an attribute of the authentication provider itself? In general, in would be true for shib and false for anything else, but @RightInTwo's installation could set it to true for builtinusers?

RightInTwo:
@scolapasta Hi Gustavo! There would need to be a way of defining which domains are trustworthy. Users from outside the institute would need to validate their addresses just as usual.

scolapasta:
Oh right, I misread and thought you meant all your built in users. But you're saying just your domain. Then I guess this would not work, because someone could just type in the domain that is trustworthy without an actual account (the whole reason for the confirmation in the first place).

RightInTwo:
@scolapasta True, thanks. Self-signup would need to be disallowed for that domain.

[...]

PR #7033, opened 2020-06-29
6915-feature-builtin-users-sendEmailNotification-new-optionalParameter

RightInTwo:
Allow to disable sending an email for confirmation after creating a user.

Closes none, but solves part of #6915 

We don't have a dev server, so there was no local testing yet.

Create and edit users with the builtin-users API with and without explicitly submitting PathParam with 
@POST
@Path("{password}/{key}/{sendEmailNotification}")

and as QueryParam with

@POST
...and &sendEmailNotification=...

NO

could be a new small feature to mention

**Additional documentation**:
see doc/sphinx-guides/source/api/native-api.rst

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/BuiltinUsers.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/BuiltinUsers.java
@@ -84,8 +84,8 @@ public class BuiltinUsers extends AbstractApiBean {
     //and use the values to create BuiltinUser/AuthenticatedUser.
     //--MAD 4.9.3
     @POST
-    public Response save(BuiltinUser user, @QueryParam("password") String password, @QueryParam("key") String key) {
-        return internalSave(user, password, key);
+    public Response save(BuiltinUser user, @QueryParam("password") String password, @QueryParam("key") String key, @DefaultValue("true") @QueryParam("sendEmailNotification") String sendEmailNotification) {
+        return internalSave(user, password, key, sendEmailNotification);
     }
 
     /**
@@ -105,9 +105,10 @@ public class BuiltinUsers extends AbstractApiBean {
         return internalSave(user, password, key);
     }
     
-    //TODO: detailed description 
     /**
-     * Created this for #6915....
+     * Created this new endpoint to resolve issue #6915, optionally preventing 
+     * the email notification to the new user on account creation by adding 
+     * "false" as the third path parameter.
      *
      * @param user
      * @param password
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/api/BuiltinUsers.java
```diff
@@ -105,7 +105,28 @@ public Response create(BuiltinUser user, @PathParam("password") String password,
...
+     * Created this for #6915...
+     *
+     * @param user
+     * @param password
+     * @param key
+     * @param sendEmailNotification
+     * @return
+     */
+    @POST
+    @Path("{password}/{key}/{sendEmailNotification}")
```
pdurbin, line 119: This new endpoint should be documented. Here's a link to the current docs in this area: http://guides.dataverse.org/en/4.20/api/native-api.html#create-a-builtin-user

The file to edit is doc/sphinx-guides/source/api/native-api.rst

djbrooke:
@RightInTwo, moving this back to Community Dev. I saw your note about it not being ready yet, but it looks like you already have one review suggestion! :)

src/main/java/edu/harvard/iq/dataverse/api/BuiltinUsers.java
```diff
@@ -105,7 +108,29 @@ public Response create(BuiltinUser user, @PathParam("password") String password,
...
+    public Response create(BuiltinUser user, @PathParam("password") String password, @PathParam("key") String key, @PathParam("sendEmailNotification") Boolean sendEmailNotification) {
```
pdurbin, line 124: sendEmailNotification appears to be a path parameter rather than a query parameter, as documented above. @RightInTwo can you provide a full curl example?

RightInTwo: Hi @pdurbin, we added both ways (query and path), just as the create() function is implemented for both currently. The docs at http://guides.dataverse.org/en/latest/api/native-api.html#assign-default-role-to-user-creating-a-dataset-in-a-dataverse seem to omit the path variant completely, which is why we left it out as well.

pdurbin: Oh, I get it. The query parameter version is called "save". Thanks.

RightInTwo:
- [ ] Improve docs (path vs. query params)
- [ ] Improve code comments

RightInTwo:
Wonderful! Thank you @pdurbin, @kcondon and @djbrooke!

PR #7044, opened 2020-07-01
fix mimetype of error pages

### LINKED ISSUE #7043: Error pages are application/xhtml+xml
Unlike Dataverse's normal pages which are delivered as mimetype text/html despite their .xhtml extensions, the error pages (403.xhtml, 404.xhtml, and 500.xhtml) are sent as application/xhtml+xml. This doesn't directly affect how they work, but this breaks some uses of javascript in those pages, i.e. use of jquery to dynamically add content in a custom header or footer breaks as doing something like $('#header').html('something'); doesn't work - the html() method isn't allowed/supported on xhtml+xml. 

QDR has this issue - I don't know if there are others affected, and I haven't checked whether other scripts, such as the google analytics are affected on these pages. 

A quick fix I've found is to add the .xhtml extension to the mime-map in web.xml. I haven't seen any side effects. I'll submit a PR with that change, but if there are better ways to handle this, let me know.

qqmyers:
**What this PR does / why we need it**: Gives 403, 404, and 500 error pages the text/html mimetype which allows javascript/jquery to work correctly in any scripts embedded, for example, in custom headers or footers

Closes #7043 

**Special notes for your reviewer**: suggestions welcome

**Suggestions on how to test this**: Check the mimetype in the browser dev console/network list and verify that the 403.xhtml, 404.xhtml, and 500.xhtml pages have the text/html mimetype as normal, e,g, dataset.xhtml, pages do. Looks like you can go directly to those pages, rather than having to create the relevant error condition to see them.

**Additional documentation**:


mheppler:
@qqmyers did some reading up on mime-types and XHTML. Found this [W3 reference](https://www.w3.org/2003/01/xhtml-mimetype/) that suggests the "short answer is application/xhtml+xml". Curious if you tried it as opposed to the `text/html` setting.

> The second technique has probably the best coverage of existing browsers while keeping the new MIME-type for those able to understand it; the downside is that it can be quite tricky to implement on your site, depending on how much access you have to its settings.

qqmyers:
@mheppler - application/xhtml+xml is what these started with. To see the issue, compare https://data.qdr.syr.edu/ and https://data.qdr.syr.edu/bob - the former has our header, the latter doesn't. You can see the xhtml+xml type in the browser console.

The W3 ref above notes that there can be differences, such as with rules for the <br> tag, so it looks like there could be side effects of switching to text/html, but this is what all the other pages use. (and all of them, normal and error, say their type is text/html in their internal metadata).

mheppler:
Ah, thank you for the clarification. I missed the part where that is the same type the other pages are using. I'll still ping @scolapasta and request his blessing on the PR.

pdurbin:
@qqmyers I'm making a swing through pull requests in code review and I'm confused about this one. When I go to the 404.xhtml page on a server that's running develop, it's already showing the "text/html" MIME type you want. Here's some output from `curl -i`:

```
$ curl -i http://ec2-3-235-63-95.compute-1.amazonaws.com/404.xhtml
HTTP/1.1 200 OK
Date: Thu, 09 Jul 2020 18:43:36 GMT
Server: Apache/2.4.37 (centos) OpenSSL/1.1.1c
Set-Cookie: JSESSIONID=4e3e12329c3ac101437cf32379b9; Path=/; HttpOnly
Content-Type: text/html;charset=UTF-8
Transfer-Encoding: chunked

<?xml version='1.0' encoding='UTF-8' ?>
<!DOCTYPE html>
...
```

From the issue it looks like I'm supposed to expect "application/xhtml+xml" as the MIME type for the 404.xhtml page on the develop branch. Do you know why I'm not seeing it? I also tried with Firefox. Same result as curl.

qqmyers:
@pdurbin - I don't know why this would be. My only guesses would be that without this PR, Dataverse uses the platform default which is somehow different for QDR, or there's something related to adding custom headers/footers (less likely I'd think). QDR is  a fork, but I don't know of any way the changes I know of would affect the mimetype. I guess this can be tabled unless someone else can reproduce it somewhere.

pdurbin:
@qqmyers thanks. I'm fine with tabling it for now. I just moved it to "Community Dev" so we don't see it in our "Code Review" column. We've had a lot of pull requests lately so it's been a full column.

Sounds like I was testing correctly, which I'm glad to hear. Obviously, if anything changes, please let us know!

qqmyers:
@pdurbin - even more fun, I can't reproduce this on QDR's dev machine (and don't want to restart the prod machine to see if I can make it happen there again). The only difference between dev and prod would be outside Dataverse (load balancer, machine config, etc.)

pdurbin:
I did have the thought that I always Apache in the mix on servers. I don't know if MIME types are passed through or converted or what.

qqmyers:
can't reproduce issue, so this shouldn't be needed.

luddaniel:
> @qqmyers I'm making a swing through pull requests in code review and I'm confused about this one. When I go to the 404.xhtml page on a server that's running develop, it's already showing the "text/html" MIME type you want. Here's some output from `curl -i`:
> 
> ```
> $ curl -i http://ec2-3-235-63-95.compute-1.amazonaws.com/404.xhtml
> HTTP/1.1 200 OK
> Date: Thu, 09 Jul 2020 18:43:36 GMT
> Server: Apache/2.4.37 (centos) OpenSSL/1.1.1c
> Set-Cookie: JSESSIONID=4e3e12329c3ac101437cf32379b9; Path=/; HttpOnly
> Content-Type: text/html;charset=UTF-8
> Transfer-Encoding: chunked
> 
> <?xml version='1.0' encoding='UTF-8' ?>
> <!DOCTYPE html>
> ...
> ```
> 
> From the issue it looks like I'm supposed to expect "application/xhtml+xml" as the MIME type for the 404.xhtml page on the develop branch. Do you know why I'm not seeing it? I also tried with Firefox. Same result as curl.

Hello everyone, our team is currently experiencing the same issue reported (using Dataverse v5.0) and we believe that the pull request is relevant. The issue is related to Payara default configuration and pages with http status codes in error.

The test done is fact incorrect because you have requested a page with HTTP code 200.
The simpliest way to test is to request an invalid url ```https://<hosted_dataverse>/this-is-incorrect-url``` this result in ```Status Code``` equals to ```404 Not Found``` and the ```Content-Type``` is ```application/xhtml+xml;charset=UTF-8```. 

Sadly, ```xhtml+xml``` content type will not work with many javascript code and html code where most of working page of Dataverse uses ```Content-Type: text/html;charset=UTF-8```. This is a big problem with you use customized header and footer with javascript for cookies managment, CDN...

For now, we edited the file *payara5/glassfish/domains/datainra/config/default-web.xml* with 
```
<mime-mapping>
    <extension>xhtml</extension>
    <mime-type>text/html</mime-type>
  </mime-mapping>
```
It seems to work nicely.
Hope it helps, I'll be happy to add more details if necessary.

pdurbin:
@luddaniel thanks for your detailed explanation. I tried your this-is-incorrect-url test and got the same results you did.

@qqmyers what do you think? Should we reopen #7043 which already mentions javascript not working, etc.?

qqmyers:
Yep - sounds good.

poikilotherm:
1. I see that https://github.com/payara/Payara/blob/72eac25f0121bc8c021a8e9ad26963b085192611/appserver/admin/gf_template/src/main/resources/config/default-web.xml#L1155 hasn't been touched in a while.
2. I see that sending XHTML as `text/html` is a [dirty hack per W3C](https://www.w3.org/2003/01/xhtml-mimetype/)
3. I see that the [W3C verifier has a lot to complain about for Dataverse demo](https://validator.w3.org/check?uri=https%3A%2F%2Fdemo.dataverse.org%2F&charset=utf-8&doctype=XHTML+1.0+Strict&group=1&No200=1&user-agent=W3C_Validator%2F1.3+http%3A%2F%2Fvalidator.w3.org%2Fservices)
4. I see that `<!DOCTYPE html>` is valid for HTML5 only, as it's missing the DTD link (https://www.w3schools.com/tags/tag_doctype.asp)
5. We claim to use XHTML (xmlns of `<html>`), but send "Content-Type: text/html" via [this `<meta>` tag](http://github.com/IQSS/dataverse/blob/9c1e6f0c94bfec59e8950d12827bfca1789dc7ce/src/main/webapp/dataverse_template.xhtml#L19-L19)
See also the [error messages that set me on the right trail.](https://validator.w3.org/check?uri=https%3A%2F%2Fdemo.dataverse.org%2Finvalid&charset=utf-8&doctype=HTML+4.01+Strict&group=1&No200=1&user-agent=W3C_Validator%2F1.3+http%3A%2F)
6. Accessing https://demo.dataverse.org/404.xhtml has MIME-Type `text/html` per using `dataverse_template.xhtml`
7. Triggering a real 404 [makes the servlet send the file directly](http://github.com/IQSS/dataverse/blob/9c1e6f0c94bfec59e8950d12827bfca1789dc7ce/src/main/webapp/WEB-INF/web.xml#L7-L7), likely overriding the MIME-Type per default MIME config in Payara.

So we should fix a few more urgent things:
1. Switch from XHTML to HTML5, as we are using Bootstrap. (See [boostrap docs](https://getbootstrap.com/docs/4.3/getting-started/introduction/#starter-template) and also [this commit comment in our codebase](https://github.com/IQSS/dataverse/commit/ac8edd6df2ac94999bab540e30053001bd1f481b)
2. Refactor all XHTML->HTML5 files to use `.html` instead, fixing the MIME type issue for the error pages
3. Remove the "Content-Type" meta tag.
4. Insert the `<meta charset>` tag per https://www.w3.org/International/questions/qa-html-encoding-declarations.en (needs to be at the very top, < 1024 byte)

It would be a good thing to do in the same go:
1. Fix all the XML namespaces in our files, as some have changed e. g. for JSF
2. Make the validators more happy, also by introducing regular checks via Maven plugins https://github.com/validator/maven-plugin

Do you want me to create a small PR for the more urgent parts?

poikilotherm:
OK while some of [my older comment](https://github.com/IQSS/dataverse/pull/7044#issuecomment-825651203) remains somewhat valid / relevant, after some more digging into the JSF/XHTML/Servlet stuff, this could be a bug in Payara.

As we do have the routing set up in `web.xml` to make the error pages render via JSF (and it's rendering), it looks like the surrounding servlet container is still setting the MIME type for XHTML, maybe misleaded as the file exists. JSF should set this to `text/html` as it does for any other case.

Maybe we should create a small reproducer and test with different Payara versions and other app servers? It would make it much easier to argument for a possible bug in Payara.

pdurbin (approved):
I haven't tested this but it's a four line change we can back out if there's any trouble. Approved.

Issue #4529, opened 2018-03-21
Data Access API: Download by Dataset


SamiSousa:
Is there any method of downloading all the files of a dataset using the Data Access API? Something like using he global_id of the dataset to download all files in a zip, similar to the bundle download. Thanks!

pdurbin:
@SamiSousa thanks for the suggestion.

For a little more context, see our conversation in IRC: http://irclog.iq.harvard.edu/dataverse/2018-03-20#i_64759

pdurbin:
@SamiSousa it was nice meeting you this afternoon at BU! I couldn't remember if you had opened any issues but found this one. Like I was saying, with 800+ issues I sometimes reach out to the person who originally opened the issue to see if they're still interested. I got the impression that you may or may not be interested in this issue long term, after your class is over, which is totally fine.

We did discuss this issue during backlog grooming in late March but we're worried about performance implications primary, even though the code to address this issue is probably straightforward.

Also, is this issue on topic enough to link your team's video from here? I ask because I just checked my email and didn't get your message yet. We have pretty aggressive spam filtering enabled and I get a summary email at the end of the day that may give me the opportunity to see your email and have it delivered to my inbox. Thanks!

pdurbin:
@SamiSousa nevermind! I clicked "release and send" in the antispam tool...

![screenshot from 2018-05-01 17-46-55](https://user-images.githubusercontent.com/21006/39496119-2b2d73c8-4d6b-11e8-9dda-c632a4c99cd9.png)


... and now I have a link to your video and code:

- https://www.youtube.com/watch?v=6G86wwgJHnc
- https://github.com/dataverse-broker/dataverse-broker

SamiSousa:
Great meeting you too Phil! In the project, we ended up using the Search and Data Access APIs to list files and download individual files, so this specific feature isn't a high priority request from me. Hope this helps!

pdurbin:
@SamiSousa no problem. Thanks for clarifying. I just sent a message about your project to the Dataverse community at https://groups.google.com/d/msg/dataverse-community/P4llZSssZ2Q/zvhGltLpAQAJ and you are welcome to make sure I didn't misrepresent your project at all. The video is really interesting! Thanks for sharing!

pdurbin:
@SamiSousa questions are coming in already! Please see https://github.com/dataverse-broker/dataverse-broker/issues/46 . Thanks!

pdurbin:
[...]

Issue #6972, opened 2020-06-09
Download: 404 error for pdf downloads in projects with ingested tabular files 


rdemgenski:
When selecting a single pdf file on a data project page and downloading it from the upper download option button, it (1) gives the option of "original format" and "archival format" when it should not; and (2) selecting original format leads to 404 download error

(1) only happens for projects that include ingested files

(2) error occurs only for non-ingested files

(3) error happens when selecting original format (works when selecting .tab even though that doesn't exist)

Issue replicated on demo: https://demo.dataverse.org/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2FNDP9TB

Erroneous download option:
![image](https://user-images.githubusercontent.com/42038889/84171580-f865a680-aa48-11ea-8bd9-ce4e7e0e3b73.png)

Error message:

{"status":"ERROR","code":404,"message":"'/api/v1/access/datafile/1113532' datafile access error: requested optional service (image scaling, format conversion, etc.) is not supported on this datafile."}

Issue first discovered on QDR DV v. 4.20-qdr

(also reported in #7505)

djbrooke:
Thanks @rdemgenski for the report. I was able to confirm this on demo.dataverse.org.

landreev:
OK, we've just run into this issue all over again, while testing PR #7047. 

So there are  a couple of issues at play:
It looks like we ALWAYS present the "original vs. tabular" choice whenever there are any tab. files in the (entire) dataset. Even if the selected checkboxes are not tabular; even if there's only one checkbox, for a non-tabular file, that is selected. (even if there are no tab. files in this given version either!) 
So we do want to replace this with a check on only the files that are actually selected via the checkboxes. 

In most situations, It's not a huge deal of an issue - but could be confusing. It only becomes a real issue, with the failing download, when there's only one, non-tabular file selected; or multiple files are selected, but only one is left once the restricted-to-the-user files are stripped off the list. Because then it's redirected to the single file download API; and it doesn't like when a format that does not apply to the specified file is requested. Note that this latter case will still be a possibility, even after the fix described above is applied. So we should also address that, by dropping the "format=orig" parameter from the final /api/access/datafile call, if the one remaining file is not tabular.  

(should be trivial enough fixes though)

PR #7047, opened 2020-07-01
Add "download all" buttons (including size of dataset) to dataset page

### LINKED ISSUE #6118: Make download all files feature more prominent
As a user / file downloader I'd like to have the download all files feature to be more prominent. Currently, this option only appears after ticking the box "1 to X Files of Y Files" on the top of the file table. A possible solution would be to describe the selection boxes more explicitly and have two selection boxes, one box to select all the files in the file table currently displayed on the page, and another selection box to select all the files in the dataset; see the following illustration:

image

For datasets with less than the minimal number of displayed files on one page (10 files?) the two selection boxes would obviously select the same amount of files, i.e. all files in the dataset.

### LINKED ISSUE #6400: Display the total size of a dataset on the dataset page
DataverseNL users (per @LauraHuisintveld in the community mailing list - https://groups.google.com/d/msg/dataverse-community/IN-ap31FcDU/H9S4C6poBQAJ) would like the total size of a dataset in on the dataset page. This is useful information when downloading all the files of a dataset at once.

This information is not currently included in the dataset redesign.
Somewhat related to:
- https://github.com/IQSS/dataverse/issues/5588 Download all error reporting
- https://github.com/IQSS/dataverse/issues/6118 Make download all files more prominent

Questions:
- What is our current state? We display file sizes. We capture the total byte size of the files.
- What can we know about download size? When can we know it?
- How might we display the size? Mockup options for position and formatting, depending on what we know about the size and how we know it. 

Ideas:
The goal is to reduce incomplete downloads (the manifest solution), and move the warning earlier in the download experience. 

1. @mheppler reminds that there my be previous mockups with a [smart download button that keeps track of size of files selected](https://iqssharvard.mybalsamiq.com/projects/dataset5000redux-dataverse50/Files%20-%20Dataset%20-%20Depositor), and/or number of files selected, and [warns if/when "too large"](https://iqssharvard.mybalsamiq.com/projects/filesallthings-enduser-dataverse42andbeyond/Dataset%20-%20Files%20-%20Download%20Limit%20Error)
2. alternate is to provide info – show total on the page (top of table, for example) and warn if "too large"
3. alternate is to enable the download for larger size data, whenever a zip is generated. Related: #6093. Depends on AWS, which not all installations are on.

pdurbin:

This pull request adds "download all" buttons at the dataset level and shows the size of the dataset.

- Closes #6118  Make download all files feature more prominent
- Closes #6400 Display the total size of a dataset on the dataset page

(this should probably be removed, since the issue has been resolved? - At least I'm going to mark it as resolved. -- L.A.)
Riddle me this... why do I get a different "size of dataset" number that what I see for the file? Here's a screenshot:

Screen Shot 2020-07-01 at 3 47 34 PM

Also, there has been discussion of adding a "download all files" API but @djbrooke and I have decided to split this off into a separate chunk, represented by #4529.

- Test both tabular and non tabular files.
- Test datasets with and without terms of use.
- Test datasets with and without a guestbook.
- Test restricted files.

I don't see any mockups linked from #6118 but I believe they are our there somewhere. In addition to the screenshot above, here's one for a dataset with no tabular files:

Screen Shot 2020-07-01 at 3 54 03 PM

I think there's already a release note about general dataset redesign improvements? I can add another note if we want.

**Additional documentation**:

I didn't touch the User Guide because my understanding is that we are going to update it all at once after the UI changes.


src/main/webapp/dataset.xhtml
```diff
@@ -143,17 +139,29 @@
...
-                                                    <ui:remove>
                                                     <!-- DOWNLOAD -->
-                                                    <!-- TO-DO #3513 DOWNLOAD ALL LINK -->
                                                     <ui:fragment rendered="#{DatasetPage.canDownloadFiles()}">
                                                         <li class="dropdown-header">#{bundle['dataset.accessBtn.header.download']} <span class="glyphicon glyphicon-download-alt"/></li>
-                                                        <li class="disabled">
-                                                            <span class="ui-commandlink ui-widget ui-state-disabled btn-download">Download (Placeholder)</span>
+                                                        <!-- NORMAL DOWNLOAD BUTTON (NO TABULAR FILES) -->
+                                                        <li jsf:rendered="#{!DatasetPage.isHasTabular()}">
+                                                            <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadArchival()}">
```
mheppler, line 147: Links are missing the btn-download style class included in the placeholder, and required by the analytics code.

src/main/webapp/dataset.xhtml
```diff
@@ -143,17 +139,29 @@
...
+                                                        </li>
+                                                        <!-- DOWNLOAD ORIGINAL BUTTON (TABULAR FILES PRESENT) -->
+                                                        <li jsf:rendered="#{DatasetPage.isHasTabular()}">
+                                                            <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadOriginal()}">
+                                                                #{bundle.downloadOriginal} (ZIP, #{DatasetPage.sizeOfDataset})
+                                                            </p:commandLink>
+                                                        </li>
+                                                        <!-- DOWNLOAD ARCHIVAL FILES (TABULAR FILES PRESENT) -->
+                                                        <li jsf:rendered="#{DatasetPage.isHasTabular()}">
+                                                            <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadArchival()}">
```
mheppler, line 159: "ZIP" file type labels should be moved to the bundle.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2886,8 +2889,35 @@ public void setSelectedDownloadableFiles(List<FileMetadata> selectedDownloadable
...
+        try {
+            long bytes = commandEngine.submit(cmd);
+            return FileSizeChecker.bytesToHumanReadable(bytes);
+        } catch (CommandException ex) {
+            return "";
+        }
+    }
+
+    public void validateAllFilesForDownloadArchival() {
+        selectAllFiles();
```
mheppler, line 2904: I assume this `selectAllFiles()` is what's responsible for the odd UI behavior... when you select "Download" from the Access Dataset dropdown, you get the ZIP download, but now all the files are selected in the files table... that shouldn't be necessary... there has to be a way to get the ZIP without the unexpected UI behavior.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2886,8 +2889,35 @@ public void setSelectedDownloadableFiles(List<FileMetadata> selectedDownloadable
...
         this.selectedNonDownloadableFiles = selectedNonDownloadableFiles;
     }
-    
-            
+
+    public String getSizeOfDataset() {
+        GetDatasetStorageSizeCommand cmd = new GetDatasetStorageSizeCommand(dvRequestService.getDataverseRequest(), dataset, false, GetDatasetStorageSizeCommand.Mode.DOWNLOAD, workingVersion);
+        try {
+            long bytes = commandEngine.submit(cmd);
+            return FileSizeChecker.bytesToHumanReadable(bytes);
```
mheppler, line 2897: Bug in the download size displayed. My locally tested datasets are saying the download will be "300 B", but the resulting ZIP is 11 KB, or 11,000 B.

src/main/webapp/dataset.xhtml
mheppler: There is already a `application/zip=ZIP Archive` entry in the MimeTypeDisplay.properties, not sure if it would be possible to pull from that file, or if we need a new entry in Bundle.properties.

mheppler:
Kicked the tires on the analytics a lil more. Built the branch, added style classes for `btn-download`, and confirmed that worked. Also realized that the explore like was missing the `btn-explore` style class, so the latest from develop needs to be merged as that style class was added in the recently merged **6938 analytics new buttons #7008** PR.

One additional detail that will need to be tweaked in the analytics code is to resolve how the new Access Dataset > Download clicks are recorded in the Google Analytics reports. The way the code works, it pulls the link text to report what button was clicked, so now that the file format and size was added to the new download link text, it makes into the report, which is going to mean you can not group those events by category (see screenshot).

Screen Shot 2020-07-01 at 11 49 27 PM

src/main/webapp/dataset.xhtml
pdurbin: Fixed in 054f249a4.

src/main/webapp/dataset.xhtml
pdurbin: Fixed in 140ba41e4.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
pdurbin: Yes, this what I noted when saying "riddle me this" in the description of this pull request (with a screenshot), that there's a mismatch in size. GetDatasetStorageSizeCommand needs discussion in tech hours or #dv-tech (heads up @scolapasta ) or a smaller group. I haven't looked into the implementation but the thing I was wondering about is how we estimate the size of a zip. Text files compress really well. Video files don't. I doubt the number will ever be 100% accurate. But yes, it seems pretty far off right now.

scolapasta: We can discuss next tech hours, sure. But isn't the inaccuracy in the wrong direction?

i.e. the ip should be smaller than the size from GetDatasetStorageSizeCommand? I could imagine a scenario where it would be large (due to zip headers), but in that case, I would expect it to be close in size.

But maybe this is "close" in this scenario.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
pdurbin: I tried adding `clearSelection()` to the end like this...

```
     public void validateAllFilesForDownloadArchival() {
         selectAllFiles();
         boolean guestbookRequired = isDownloadPopupRequired();
         boolean downloadOriginal = false;
         validateFilesForDownload(guestbookRequired, downloadOriginal);
+        clearSelection();
     }
```

... but now it thinks no files are selected:

Screen Shot 2020-07-02 at 1 08 07 PM

So yes, I agree a fix would be nice but I'm not sure what that fix is.

scolapasta: Wouldn't the fix be to use a different method than what the one that cares about selected files? i.e

change current method to get selected files and get a list, path that to new private method that takes core of current method but uses this new list.

Create a new method that generates a list and calls new that new private method.

You'd of course have to make sure all the checks for permissions are happening correctly, but I'd expect that to already be in the backend side (i.e. we already shouldn't be relying in the front end to only pass in the permissible values)

pdurbin: @scolapasta @mheppler please see 427c88356. This pull request is now much bigger but it no longer uses `selectedFiles` which is what the checkboxes are connected to.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
pdurbin: As discussed in tech hours, the problem seems to stem from double counting tabular files (counting both the size of the original file and the archival version). I just pushed a fix in 19577769b.

pdurbin:
I'm ready for more code review. I addresses the checkbox issue and am now differentiating between sizes for original files vs archival. Here's a screenshot:

Screen Shot 2020-07-07 at 5 18 29 PM

djbrooke:
Two more changes we want to make here, it was added to another issue by mistake. We believe the change to the publish button is already done.

86605833-3507bd80-bf75-11ea-935b-bd58d2e3e382

kcondon:
@pdurbin Any resolution to the file size download all mismatch? I'm seeing the zip download is larger than the individual files.

pdurbin:
@kcondon I thought the size estimate was better as of 1957776 (before that, tabular files were being counted twice). Can you please post a screenshot of the problem you're seeing?

kcondon:
So there are at least two issues:
1. the download size for both tab and original are off but the tab is off by a lot.
2. the actual size of the zip file is considerably smaller than what's presented in the download, so I am guessing it refers to the uncompressed size?

kcondon:
image

kcondon:
This is a screen shot using FF, saving the zip to windows, downloading the original format:
image

landreev:
> So there are at least two issues:
> 
>     1. the download size for both tab and original are off but the tab is off by a lot.
> 
>     2. the actual size of the zip file is considerably smaller than what's presented in the download, so I am guessing it refers to the uncompressed size?
> 
> 
> Otherwise download all appears to work

Yes on 2. - that's not really an issue; the size the user is seeing in the download dialog is the compressed size of the zip; we want to report the sizes of the actual files. 

As for 1., looking at the sizes in the pull down menu in the example above: 
the size for the "original" format appears to be correct (62.3KB); but for the "tabular" it does appear to be off - actually, it looks like it's off by exactly the size of the original; i.e., it keeps counting both format sizes, instead of counting just the tabular size.

landreev:
After some testing and discussing it with Kevin - just another thought: 
maybe we want to add the word "uncompressed" there, to avoid confusion.
As in "(62.3KB Uncompressed)" instead of just "(62.3KB)"?

just an idea

mheppler:
> maybe we want to add the word "uncompressed" there, to avoid confusion "(62.3KB Uncompressed)"

Thank you for the review, @landreev. I thinking something along the same lines. Will review some options with @TaniaSchlatter and put suggested revisions here in this pull request.

pdurbin:
In 7e79f1b I addressed the problem tabular download options showing for versions of datasets that no longer have tabular files (because they were deleted and then the dataset was republished).

In 115353c I took another swing at getting GetDatasetStorageSizeCommand to behave properly and I thought I had it working but while testing I found further problems and weirdness so I ultimately decided to write a small command that does only the job I want. It seems to work fine so I switched over to it. See c13a5cc. I'm not doing any error checking within that small method (getDownloadSize in DatasetUtil) but I'm happy to add some if we're worried about nulls in the database for the size of files.

I'm moving this to code review.

@djbrooke and @TaniaSchlatter I noticed you're still assigned. I deployed the code here if you'd like to check it out but there are no UI changes: http://ec2-54-160-66-157.compute-1.amazonaws.com

mheppler:
Told Phil that I trusted his work, but I still went and confirmed the download sizes that I had poked at early. Looking much, much better. Didn't test anything larger than the 2.1 MB download here, but it be worth testing if a 210 MB download is off by 20 MB and if that is worth trying to fix.

Screen Shot 2020-07-16 at 5 00 27 PM

landreev:
@mheppler @pdurbin The numbers above look right to me. What looks like a discrepancy - the fact that you're seeing "2,252,470 bytes (2.3 MB on disk)" in the Apple info box, but "2.1 MB" on the Dataverse side appears to be due to the fact that Apple is interpreting "1 MB" as "1,000,000 bytes"; and Dataverse is (more correctly!) using the "1MB = 1024 * 1024 bytes" definition. What we are doing is more correct. But what Apple is doing is somewhat OK too - both definitions are widely used (unfortunately). And real life users are used to seeing  discrepancies between the 2 definitions by now. 

Can I move it to QA?

kcondon:
@landreev One thing I learned from my networking class is the 1,000,000 bytes value is used to count network traffic and the 1024 style is for disk storage. I think. Will need to reference my class notes ;)

kcondon:
@pdurbin @landreev So, it looks pretty good -I'm fine with file sizes. However, I did notice one issue:
the old style download all button still presents orig and archival formats when tab file is removed in the current version. While testing shows this exists in 4.20, what is different is when you choose download original, it throws an error on download:
{"status":"ERROR","code":404,"message":"'/api/v1/access/datafile/53792' datafile access error: requested optional service (image scaling, format conversion, etc.) is not supported on this datafile."}

So, kind of a weird issue where part of it preexists but is actually worse now. Maybe the same fix that was applied to the access download all would work here too?

kcondon:
@pdurbin @landreev @mheppler I've found a weird, preexisting bug with old download all that is technically out of scope for this issue but I'm wondering whether the same fix for hiding tab download options when tab deleted in current version might be easy to add to it?
The bug is when you have deleted tab file in existing version, have a restricted file with terms of use,  not just terms of access enabled, do not yet have access to restricted file, and choose download all but original version only. Kind of narrow but throws ugly white error screen and nothing is downloaded: {"status":"ERROR","code":404,"message":"'/api/v1/access/datafile/53792' datafile access error: requested optional service (image scaling, format conversion, etc.) is not supported on this datafile."}

landreev:
@kcondon What's "old download all"? Do you mean selecting all the checkboxes and clicking "download" in the upper right corner of the files table? 
I'm quite confused by the condition you are describing. If it's something pre-existing, I'm surprised nobody has noticed before.

landreev:
@pdurbin @kcondon @mheppler
I'm assuming this is the test dataset in question: https://dataverse-internal.iq.harvard.edu/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2FSJSEZX&version=3.1

Yeah, I can see the "original" and "tabular" options under the old style multi-download button. I assumed earlier, that the way it used to work, there was never a real "download all", i.e. it was always a download of the specified list of files from the list of the selected checkboxes... and that way the above condition would never happen, in a version that has no tabular files. But we must have added an optimization at some point - I now have some vague memories of it - for that "select all" checkbox, that actually bypasses the checkboxes and goes to the complete list of files when "all" is selected... So yes, it is probably the same logic error - searching for tabular files in the entire dataset; instead of the current version. And could probably be fixed just as easily. 
Whether it should be fixed as part of this PR, seeing how it is pre-existing and in a different code section - I don't have a strong opinion.

landreev:
@pdurbin @kcondon @mheppler

Oh, and here's the reason you get that weird API error: because one of the files is restricted, and there are 2 files total, in the end there is only 1 file in the list - and Dataverse is smart enough to then determine that this is NOT a multiple file download case; so it redirects it to the single file download API. apparently it is not smart enough to then drop the "format=original" parameter - and the single download API treats that as a fatal error, when a format that's not applicable to the current file is requested. 
This is annoying - but is indeed a fairly edge case....

kcondon:
@landreev @pdurbin @mheppler Leonid, thanks for confirming! I don't believe it is technically in scope but the question I was asking is whether it would be easy to adopt the solution for not showing download original/archive on versions that do not have tabular files, like Phil did already. If not, I'll just open a new issue.

landreev:
@kcondon

>  @pdurbin @mheppler Leonid, thanks for confirming! I don't believe it is technically in scope but the question I was asking is whether it would be easy to adopt the solution for not showing download original/archive on versions that do not have tabular files, like Phil did already. If not, I'll just open a new issue.

I feel like it's worth fixing the "don't show the format choice when there are no tab. files left in the version" part. The fix would be the exact. same approach; we'll just need to find where to apply it in the old code.

landreev:
OK, it looks like this situation is NOT even unique to the "select all checkboxes" scenario. In the old setup, it really looks like we ALWAYS present  the "original vs. tabular" choice whenever there are any tab. files in the dataset. Even if the selected checkboxes are not tabular; even if there's only one checkbox, for a non-tabular file, that is selected (we have an open issue - #6972 for that). 

It really looks like we want to replace this with a check on only the files that are actually selected. It's not a huge deal of an issue - but could be confusing. The fact that we've been talking about it for a while now illustrates that point.

landreev:
As I just mentioned to @kcondon and @djbrooke , I'm totally fine with merging this PR; and then using #6972 to fix the pre-existing conditions described above. Any objections?

kcondon:
Going once, twice...

Issue #5345, opened 2018-11-27
Proposal: make persistent EJB timers non-persistent or not rely on database


poikilotherm:
This is related to #5292 and a story of that epic.
As [`DataverseTimerServiceBean.java`](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/timer/DataverseTimerServiceBean.java) has been mostly written and maintained by @landreev, this most certainly is for him...

### Context
Currently, in [`glassfish-setup.sh`](https://github.com/IQSS/dataverse/blob/0045cd6368e0d3fee8fd7e37ec87d7d1233ae30d/scripts/installer/glassfish-setup.sh#L127) you need to setup a database connection for storage of persistent EJB timers.

Reusing the database for Dataverse for EJB timers is alright, but it is not easy to depict when heading for containers. It is much easier for container approaches to use application scoped JDBC connections, but those seem not to be reusable for EJB timers.

There seemed to be some issues in the past, too. #3672, #3789 and more.

### Proposal
*Just remove the persistent nature of timers.*

* They are created on each startup on the timing server ([see docs](http://guides.dataverse.org/en/latest/admin/timers.html)).
* In a single node setup, when the server is down, IMHO there is no need to have the timers persistent in a database. No one will execute them. Executing them after the server comes back seems with HarvestingClient and ExportClient as current use cases not very important.
* In a multi node setup, when the timing server is down, you need to change it to another server, so they get executed again. This will result in broken persistent timers anyway, as they had been created with the name of the dead timing server.

When persistance is needed in a multi node setup in the future, there is Hazelcast in Payara which supports storing timers as long as at least one cluster member is online...

Comments are very much appreciated. :-)

poikilotherm:
Following @pdurbin suggestion about DDD (doc driven development), I request a review of my plans in #5371 by @landreev and potentially others.

Cheers :smile: :beer:

landreev:
@poikilotherm 
As a quick note, yes, this makes perfect sense. 

At present, we are not benefiting from the fact that the timers are stored in the database, at all. It's kind of the other way around: you can say that we go to the extra trouble of making sure that our timers are NOT persisted. In a multi-node cluster, we specifically try to run all the timers on just one of the nodes (for various reasons); AND we clear and re-create all the timers on startup. 

So yes, it will be cleaner, not to bother saving them in the database in the first place. 

[...]

PR #7048, opened 2020-07-01
6819 - Include PostgreSQL JDBC driver in WAR

### LINKED ISSUE #6819: Include Postgres JDBC driver in pom.xml
(I looked everywhere, but I could not find an existing issue for this, thus creating a new. This is related to my old epic #5292 and had a commit for it, see df2b8cc)

Modern app servers are capable to load the JDBC driver from the deployed WAR when setting up database connections.

This makes manual installation of the Postgres driver obsolete and enables usage during integration tests with Testcontainers.

The addition to `pom.xml` is very simple:
```
<dependency>
    <groupId>org.postgresql</groupId>
    <artifactId>postgresql</artifactId>
    <version>42.2.12</version>
</dependency>
```

~~In addition, instead of letting the installer configure the database connection in `domain.xml`, we can provide the connection details either from annotations (Java EE standard compliant, should be preferred) or provide them from `glassfish-resources.xml` (which also support variables substitution from domain, environment or alias).~~

Note: the above has been skipped as beyond scope for now. Small steps.

Obviously, this affects a lot of places. Docs (#6593), installer (#6761), container images, ...
It makes updating and testing much easier: there is only ONE place to change the version, ideally by using a Maven property.

I already tested this a while ago, see links below.
What I don't remember: it's not clear if the timers database can be used for this, too. IIRC this is why I created #5345...

See also:
- payara/docker-payaraserver-full#69
- https://github.com/poikilotherm/dataverse/blob/65a88d14f22679d2405b09a50757e6bc5ebdf08c/pom.xml#L709-L721
- https://github.com/poikilotherm/dataverse/commit/e1b3f429d6a18e19ab3563d5b3be0cbb58de31a9#diff-d705b5d323debe3ee2bb19639ba12b48

poikilotherm:

With Payara 5, we don't have to deal with pushing the Postgres driver JAR into the right places. Instead it's much easier to have a single place to run to: a version used in pom.xml, shipped with the WAR, used consistently on the app server deployment.

Closes #6819

Nada. Maybe the docs changes about the timer issues. Would be very strange to see those again.

Follow release notes docs and guide docs. Deploy to Payara. Look for the EJB timers being created in the DB. Test as usual. This should work as before. Looking at the Flyway Schema History should also reveal that the migrations have been applied as usual.

Nope.

:battery: included.

**Additional documentation**:
Nada.


tests/jenkins/groupvars.yml
```diff
@@ -89,7 +89,6 @@ dataverse:
   postgres:
     reporpm: https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
     version: 9.6
-    jdbcurl: https://jdbc.postgresql.org/download/postgresql-42.2.12.jar
```
pdurbin, line 92: Pinging @donsizemore to see if he has an opinion on this jdbcurl line being removed.

scripts/installer/install.py
```diff
@@ -452,15 +446,6 @@
...
 
-# 4b1. PostgresQL driver:
-pg_driver_jarpath = "pgdriver/"+pgJdbcDriver
-
-try:
-   copy2(pg_driver_jarpath, gfJarPath)
-   print("Copied "+pgJdbcDriver+" into "+gfJarPath)
-except:
-   print("Couldn't copy "+pgJdbcDriver+" into "+gfJarPath+". Check its permissions?")
-
```
pdurbin, line 463: Mostly a question for @landreev , I suppose, but should we continue to maintain the old Perl installer and make equivalent changes there? @poikilotherm I'm talking about `scripts/installer/install`.

tests/jenkins/groupvars.yml
donsizemore: @pdurbin fine by me.

pdurbin:
After chatting with @landreev I pulled this out of QA so he has time to comment.

@poikilotherm one thing that you could perhaps comment on... What is the short term or long term vision for this change? I saw stuff about "modern" and Testcontainers and configuration via annotations. Is that where we're going? Thanks. 😄

landreev:
This may not be an issue necessarily, I'm just curious to understand how this works. If the driver is part of the application war file, this means it is only available while the application is deployed. But the database is defined on the domain level (`<jdbc-connection-pool datasource-classname="org.postgresql.ds.PGPoolingDataSource" name="dvnDbPoo
l" res-type="javax.sql.DataSource">`). Does that really mean that you can define a connection pool, and specify a datasource driver that's not actually available to the application server?

On our production servers we define this data source with the extra "validation-required" attributes: `validation-table-name="setting" is-connection-validation-required="true"` - so we are telling the application server to keep pinging the database to confirm that it's alive. What's going to happen when the application is undeployed? 

What about the ejb timer - in our setup, it uses the main PG database. But the timer is an application of its own, and it appears to still be there (deployed under `.../domain1/applications/ejb-timer-service-app`) even after the main application is undeployed? Granted, we do not need the timer when the application is not running; and it may be automatically deactivated when there are no applications deployed (?). 

I just want to know for sure that Payara is not going to be confused and crash after `asadmin undeploy dataverse`.

scripts/installer/install.py
landreev: Yes, I'd like to keep scripts/installer/install working, at least for a short while.

poikilotherm:
I did not test what happens when you undeploy, as this no scenario for me.

As long as the app will be deployed, the domain level and timer database connections seem to work just fine.

landreev:
> Does that really mean that you can define a connection pool, and specify a datasource driver that's not actually available to the application server?

The answer appears to be "yes" - I was able to do things like `asadmin create-jdbc-connection-pool ... --datasourceclassname org.postgresql.ds.PGPoolingDataSource` without the PostgreSQL driver being present. 

> As long as the app will be deployed, the domain level and timer database connections seem to work just fine.

Sure, but undeploying and redeploying is part of normal operations. So we  need know for sure it's not going to result in Payara crashing every time, especially with the database monitoring in the configuration. Should also be a straightforward thing to test; but it is an extra thing to test. I'm wondering if it would be ok to delay testing and merging this PR until after dataverse v5 is released? I don't have a strong opinion on this, but will ask others on the team.

poikilotherm:
Hi Leonid @landreev, my mid term goal for all of this is moving the database definitions from the installer/server level to the application. 
See https://javaee.github.io/javaee-spec/javadocs/javax/annotation/sql/DataSourceDefinition.html and 

Not just the installer, but ideally all resource definitions. It will be configurable, so no one has to be afraid of lost control. 
See https://docs.payara.fish/community/docs/5.201/documentation/payara-server/server-configuration/var-substitution/types-of-variables.html

Once this is done, there will be no problem with a dangling connection, as it is available as long as the app is deployed. :wink:

I felt like this would be beyond scope, but if you think this should be done in the same go, I'd be happy to add more. We can keep it to the DB connection for now.

IMHO this would be a major step forward in terms of modern deployment strategies.

scripts/installer/install.py
poikilotherm: Done in 7ed1de7

poikilotherm:
@landreev and @pdurbin I changed this to a draft PR again to avoid accidents.

Please note that I added two commits that add the necessary bits for defining the database within the app.
It's not very configurable yet and lacks some special like the validation @landreev mentioned, but I wanted to give you a quick demo. This works flawless when testing on Payara on K8s (mainly to use the static definitions of server etc).

Some notes about this. Please find some literature about how to do this and advantages/disadvantages here:
- http://dplatz.de/blog/2018/self-contained-jee-app.html
- https://docs.payara.fish/community/docs/5.2020.2/documentation/payara-server/advanced-jdbc/advanced-connection-pool-properties.html
- https://blog.payara.fish/pooling-for-databases-on-the-same-database-server

And please note that I switched from `PGPoolingDataSource` because it has been deprecated since 2017. See https://github.com/pgjdbc/pgjdbc/issues/729. A pool is auto-created by the annotated definition. FWIW: bigger pooling like those mentioned in the driver upstream issue and deprecation warning don't seem necessary. They are also mentioned in the Payara blog article mentioned above.

poikilotherm:
Hi @landreev,
some time passed and I wonder if there is a chance to push this forward again?
Are there remaining concerns with this?
Thx!

(It would be great to have this done in October to be counted for my Hacktoberfest shirt... :wink: )

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.inject.Singleton;
+import javax.sql.DataSource;
+
+@Singleton
+@DataSourceDefinition(
+    name = "java:global/jdbc/dataverse",
+    // Using PGXADataSource instead of deprecated PGPoolingDataSource
+    className = "org.postgresql.xa.PGXADataSource",
+    user = "${MPCONFIG=dataverse.db.user}",
+    // Has to be using ALIAS for at least Payara 5.2020.3, as MPCONFIG does not seem to read properly
```
poikilotherm, line 15: This has been solved as of Payara 5.2020.4.
We should use the fixed version of Payara and implement this in a clean way instead of relying on the workaround.
MicroProfile Config support needs to be adressed ASAP IMHO.

donsizemore: FWIW integration tests on 5.2020.5 produce the same results as 5.2020.2:

Screen Shot 2020-10-23 at 10 43 06

GitHub won't let me attach the HTML report, even zipped, but I can e-mail or Slack it if anyone likes.

landreev: Could you please clarify the "solved as of Payara 5.2020.4" part? Does that mean the PGXADataSource is just not supported in earlier versions?

landreev: OK,  got it,  the alias part.

poikilotherm: @landreev this is just Github highlighting stuff that is in the context. The comment is about L15-17, see https://github.com/poikilotherm/dataverse/blob/2e6b6f9f4815e8c53fe59787b2b273aac3ddc24a/src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java#L15-L17

poikilotherm:
@landreev you mentioned `validation-table-name="setting" is-connection-validation-required="true"` above.
Should we move those to `persistance.xml` and make it configurable?

See also https://github.com/payara/Payara-Server-Documentation/blob/master/documentation/payara-server/advanced-jdbc/advanced-connection-pool-properties.adoc

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
landreev: Unless I'm missing something, it looks like this PR leaves "asadmin ... create-jdbc-connection-pool" command in the installer? Should it be removed?

landreev: > @landreev you mentioned `validation-table-name="setting" is-connection-validation-required="true"` above.
> Should we introduce those to `persistance.xml` and make it configurable?
> 
> See also https://github.com/payara/Payara-Server-Documentation/blob/master/documentation/payara-server/advanced-jdbc/advanced-connection-pool-properties.adoc

I'm not sure about this part. It looks like we never felt these validation settings were important enough, as to recommend them to everybody... So maybe we shouldn't eve worry about it too much.

landreev:
> @landreev you mentioned `validation-table-name="setting" is-connection-validation-required="true"` above.
> Should we introduce those to `persistance.xml` and make it configurable?
> 
> See also https://github.com/payara/Payara-Server-Documentation/blob/master/documentation/payara-server/advanced-jdbc/advanced-connection-pool-properties.adoc

I'm not sure about this part. It looks like we never felt these validation settings were important enough, as to recommend them to everybody... So maybe we shouldn't eve worry about it too much.

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
landreev: And what about any existing installations - do they need to explicitly remove their existing `<jdbc-connection-pool ... ` sections from their configuration? Or is it ok to leave them in place?

poikilotherm:
This is ready for review.

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
poikilotherm: It would be ok to leave them on place. I added some docs how to remove them anyway.

landreev (approved):
Looks good. Thank you for adding the release note and extra information to the guide.

kcondon:
@poikilotherm Install works, timer works, was following upgrade instructions in release notes and could not delete existing pool. I think there is a typo, dvnDbpool should be dvnDbPool but even after fixing command line, getting error with resource in use (payara running?)

./asadmin delete-jdbc-connection-pool dvnDbPool
remote failure: Some jdbc resources or persistence manager factory resources are referencing this connection pool. Use cascade option to delete them as well.


```
    <jdbc-connection-pool datasource-classname="org.postgresql.ds.PGPoolingDataSource" name="dvnDbPool" res-type="javax.sql.DataSource">
      <property name="PortNumber" value="5432"></property>
      <property name="create" value="true"></property>
      <property name="ServerName" value="localhost"></property>
      <property name="databaseName" value="dvndbsimple"></property>
      <property name="User" value="dvnapp"></property>
      <property name="password" value="${ALIAS=db_password_alias}"></property>
    </jdbc-connection-pool>
    <jdbc-resource pool-name="dvnDbPool" jndi-name="jdbc/VDCNetDS"></jdbc-resource>
```

poikilotherm:
Thx @kcondon. Could your try that cascade option?

kcondon:
@poikilotherm Thanks, did not find cascade at first, thought it was postgresql. Tried it and it worked, so maybe update release note?

./asadmin delete-jdbc-connection-pool --cascade=true dvnDbPool

doc/release-notes/6819-update-db-conn.md
```diff
@@ -0,0 +1,37 @@
...
+rm /tmp/password.txt
+
+asadmin set configs.config.server-config.ejb-container.ejb-timer-service.timer-datasource=java:global/jdbc/dataverse
+```
+
+If you are using a PostgreSQL server on `localhost:5432`, you can omit `dataverse.db.host` and `dataverse.db.port`.
+
+You are safe to delete the old alias and DB pool afterwards:
+```
+asadmin delete-jdbc-connection-pool dvnDBpool
```
poikilotherm, line 35:
```suggestion
asadmin delete-jdbc-connection-pool --cascade=true dvnDBpool
```

poikilotherm:
I added a one click to commit suggestion comment, but I can't merge before CI isn't done. That'll take a while, EC2 is slow :wink:

doc/release-notes/6819-update-db-conn.md
poikilotherm: @kcondon just click commit 🥳

poikilotherm:
@kcondon looks like I mixed up pending CI with pending review in the GitHub app UI.
Committed the change as requested.

kcondon:
@poikilotherm Now running into an issue while manually deploying:
./asadmin deploy /tmp/dataverse-5.1.1.war
remote failure: Error occurred during deployment: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.6.payara-p1): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLException: Error in allocating a connection. Cause: Access denied to execute this method : setURL
Error Code: 0. Please see server.log for more details.
Command deploy failed.

poikilotherm:
Did you upgrade Payara?

https://github.com/payara/Payara/issues/4864

kcondon:
@poikilotherm I think I have found it but not sure yet why:
[2020-11-06T17:26:52.236+0000] [Payara 5.2020] [SEVERE] [] [javax.enterprise.resource.resourceadapter.com.sun.gjc.util] [tid: _ThreadID=89 _ThreadName=admin-thread-pool::admin-listener(1)] [timeMillis: 1604683612236] [levelValue: 1000] [[
  RAR5103 : Error setting java bean value : [dvndbsimpledvndbsimple]]]


the db name is dvndbsimple but it is listed twice. probably something I did, will need to look over where it is configured.

poikilotherm:
@kcondon not sure if your last comment was a question, a note to self or a hint I should chime in...?

kcondon:
@poikilotherm Apologies, was distracted but will dive back in.
i have stack traces in the server log when it fails. There are a couple. I provided a small part of the first one above. It was all of the above: an observation, something I needed to look at, and a question to you about whether you recognize how it might have happened. I looked and my entries appeared correct. I will look once more for duplicates but it aside from command line cut and paste, not sure how it would have happened? Don't want to keep you up any later so can follow up on Monday if needed.

kcondon:
@poikilotherm summarizing current state:
1. clean install with your branch worked fine with payara .2, assuming it also works with .5
2. initial upgrade process showed it needed cascade added to deleting connection pool from config
3. subsequent deployment of war file resulted in perm error allocating a connection. rechecked config and it looked ok. it was pointed out I should be running .5 rather than .2
4. updated to .5 using a simplified process only using domain.xml from .2, in the process discovering what may have caused perm issue: perhaps updating payara files as root rather than as unprivileged user.
5. after ensuring develop branch runs fine on .5 installation, repeated upgrade instructions
6. deployment of new branch failed with Error occurred during deployment: JNDI lookup failed for the resource: Name: [java:module/env/edu.harvard.iq.dataverse.flyway.StartupFlywayMigrator/dataSource]
server.log showed many entries of type: cannot be cast to org.glassfish.hk2.classmodel.reflect.impl.AnnotationTypeImpl
7. web search shows post by payara dev to set this jvm-option to suppress error: -Ddeployment.resource.validation=false https://github.com/payara/Payara/issues/2387 
8. redployment error now shows: Error occurred during deployment: Exception while preparing the app : Invalid resource : jdbc/VDCNetDS__pm. server.log shows more cannot be cast to org.glassfish.hk2.classmodel.reflect.impl.AnnotationTypeImpl (I had iterated here, deleted old postgres driver from lib dir just in case, no change that I could tell)
9. web search shows what appears to be a bug in .5: https://github.com/payara/Payara/issues/4952

kcondon:
@poikilotherm Passing back for comments or resolution on issues.

poikilotherm:
*Issues found by QA (@kcondon):*

- [X] update release notes, opinions may vary here, but note that upgrading to .5 payara is required for this release and that it should be done before adjusting db conn pool and deploying war file and they should confirm the existing deployed app works with the new payara config before proceeding. *(joined the files, emphasized and updated to be more like other release notes on upgrading.)*
- [x] update release notes, in the delete db pool command, need to change dvnDBpool to be dvnDbPool
- [x] update release notes, remove part about omitting host and port if localhost:5432 since on my box at least one cannot deploy without them, gets invalid siteUrl error. *(fixed by adding missing properties file for MP Config)*
- [x] cannot restart app server and have application load up although it is still shown as deployed by asadmin. *(Tried really hard to reproduce. No luck. Used local installation.)*
- [ ] was not able to get harvest client timer to fire off but this may have been complicated by the restart issue above. can try another approach but so far no luck.
- [x] have noticed deploying and undeploying seem a bit slower. not a major concern but @pdurbin had opened an issue about making it faster. *(Both ways of deployment had reproducable times of 32.5 secs +-1.0 sec. Ran both 5 times each.)*

@kcondon can we try this again...?

poikilotherm:
@donsizemore as you mentioned it today on IRC: do you want us to update to PostgreSQL Client 42.2.18?

From what I can see from the [Changelog](https://jdbc.postgresql.org/documentation/changelog.html#version_42.2.18) between .12 and .18, there shouldn't be anything biting back when pushed.

pdurbin:
I'm keeping this pull request in code review for now but I expect some movement after @poikilotherm @scolapasta @landreev @qqmyers and I discuss it after the community call on Tuesday.

My take (after discussing with @poikilotherm yesterday) is that this pull request and the 5345-nonpersistent-ejb branch linked from #5345 could be split up into four smaller pull requests that could move across the board in this order:

- Upgrade to a newer version of Payara.
- Change EJB timers to be non-persistent #5345.
- Switch to the `@DataSourceDefinition` annotation introduced in Java EE 6, configurable with a microprofile-config.properties file using MicroProfile.
- Move the postgres driver into the war (and out of the installer).

Even if we don't split up this pull request, I don't believe we want to merge it as-is because there are problems with undeploying (having to do with timers using the database).

poikilotherm:
Closing in favor to #7416, #7419, #7420 and a yet-to-be-done PR about the `@DataSourceDefinition`, see #7418

PR #7049, opened 2020-07-01
Delete scripts/setup/asadmin-setup.sh

poikilotherm:

Apparently, there are no usages of this script anymore. It uses Glassfish 4 and ancient Postgres 9.3. As it's the only file left in the directory, better remove it to avoid confusion.

Opened this PR upon request by @pdurbin on IRC _without_ opening an issue before.

Do you know of this script being actively used today?

Nada.

Nada.

Nada.

**Additional documentation**:
Nada.


pdurbin:
@michbarsinai thanks for weighing in. This script isn't actively used and would need to be updated to keep it in line with the scripts we do use.

PR #7053, opened 2020-07-04
7052 Fix infinite IN PROGRESS and DELETE IN PROGRESS labels

### LINKED ISSUE #7052: Harvesting run infinitely stays at IN PROGRESS and DELETE IN PROGRESS
When the issue occurs: When a harvesting client is running/deleting and on the server dataverse gets rebooted. Since the status in the database is still running/deleting, the user can't rerun or delete the client because the buttons are disabled.
Which page it occurs on: Root>Dashboard
To whom it occurs: People who harvest
Version: 4.20

JingMa87:
**What this PR does / why we need it**: Fixes infinite IN PROGRESS and DELETE IN PROGRESS labels

**Which issue(s) this PR closes**: 7052

Closes #7052 

**Special notes for your reviewer**: People will be happy about this

**Suggestions on how to test this**: Run a harvesting client and delete another one at the same time, then restart Glassfish immediately when you see the IN PROGRESS and DELETE IN PROGRESS labels. After you're done restarting you'll see the RUN FAILED AND DELETE FAILED labels instead of the old ones and you can edit the client again. So you can rerun or delete like normally.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Yes

image

**Is there a release notes update needed for this change?**: Yes, it's a noticeable change.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/harvest/client/ClientHarvestRun.java
+++ b/src/main/java/edu/harvard/iq/dataverse/harvest/client/ClientHarvestRun.java
@@ -5,16 +5,9 @@
  */
 package edu.harvard.iq.dataverse.harvest.client;
 
+import javax.persistence.*;
 import java.io.Serializable;
 import java.util.Date;
-import javax.persistence.Entity;
-import javax.persistence.GeneratedValue;
-import javax.persistence.GenerationType;
-import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
-import javax.persistence.Temporal;
-import javax.persistence.TemporalType;
 
 /**
  *
@@ -40,12 +33,13 @@ public class ClientHarvestRun implements Serializable {
         this.id = id;
     }
 
-    public enum RunResultType { SUCCESS, FAILURE, INPROGRESS };
+    public enum RunResultType { SUCCESS, RUN_FAILED, RUN_IN_PROGRESS, DELETE_FAILED };
     
     private static String RESULT_LABEL_SUCCESS = "SUCCESS";
-    private static String RESULT_LABEL_FAILURE = "FAILED";
-    private static String RESULT_LABEL_INPROGRESS = "IN PROGRESS";
-    private static String RESULT_DELETE_IN_PROGRESS = "DELETE IN PROGRESS";
+    private static String RESULT_LABEL_RUN_FAILED = "RUN FAILED";
+    private static String RESULT_LABEL_RUN_IN_PROGRESS = "RUN IN PROGRESS";
+    private static String RESULT_LABEL_DELETE_IN_PROGRESS = "DELETE IN PROGRESS";
+    private static String RESULT_LABEL_DELETE_FAILED = "DELETE FAILED";
     
     @ManyToOne
     @JoinColumn(nullable = false)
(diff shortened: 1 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/harvest/client/ClientHarvestRun.java
```diff
@@ -5,16 +5,9 @@
  */
 package edu.harvard.iq.dataverse.harvest.client;
 
+import javax.persistence.*;
```
pdurbin, line 8: I'm surprised it isn't already written at http://guides.dataverse.org/en/4.20/developers/coding-style.html but we tend not to use wildcard imports like this.

src/main/java/edu/harvard/iq/dataverse/harvest/client/ResetHarvestInProgressBean.java
```diff
@@ -0,0 +1,23 @@
+package edu.harvard.iq.dataverse.harvest.client;
+
+import javax.annotation.PostConstruct;
+import javax.ejb.EJB;
+import javax.ejb.Singleton;
+import javax.ejb.Startup;
+
+@Singleton
```
pdurbin, line 8: Maybe I'm just being paranoid but we've had trouble with Singleton beans recently. See https://github.com/IQSS/dataverse.harvard.edu/issues/73

JingMa87: I read the issue you linked very closely since the last thing I want to do is introduce new errors. In the issue there's an authentication bean that's a singleton and when the bean gets held up by a process, no other clients can use it to authenticate. This is bad. However, in my use case the singleton is only used during the startup of Dataverse to reset any hanging harvesting clients. After that there's no use for it, so a singleton bean seems the right approach here. Moreover, in Dataverse there's many singleton beans with an additional @Startup tag to initialize a certain process and those don't seem to be causing issues either. So I'm fairly confident that a singleton is the way to go here.

src/main/java/edu/harvard/iq/dataverse/harvest/client/ClientHarvestRun.java
```diff
@@ -40,12 +40,13 @@ public void setId(Long id) {
         this.id = id;
     }
 
-    public enum RunResultType { SUCCESS, FAILURE, INPROGRESS };
+    public enum RunResultType { SUCCESS, RUN_FAILED, RUN_IN_PROGRESS, DELETE_FAILED };
```
landreev, line 43: Not a big deal, but what is the value of adding this status condition for a failure to delete a client? It has some value for a dataverse admin, to know that the last attempt to harvest from a certain server resulted in a failure. Is it really useful to know that an attempt to delete it failed? - should the startup check simply remove the "delete in progress" flag quietly instead?
After all, seeing how the client is still there makes it somewhat clear that the attempt to get rid of it didn't work out, so they should try again?

I may be missing some situation where it could actually be useful - so I'm open to hearing it.

src/main/java/edu/harvard/iq/dataverse/harvest/client/ResetHarvestInProgressBean.java
landreev: I'm not worried about it being a singleton - seeing how it has nothing in it but the `@PostConstruct` code. 
But there's another reason why this is not going to work, specifically for us here at Harvard. We have 2 production servers. So if one of the two is being restarted, it'll reset the flags even though the actions in question may still be in progress on the other node. 

It's easy to address **for the most part**: in a multi-server situation only one node is designated to do harvesting, and other timer-run jobs. This "master node" is the one that has the JVM option `-Ddataverse.timerServer=true`. And you can use `if (systemConfig.isTimerServer())` to check if this node is the timer node in your code, before proceeding to clear the flags. 

HOWEVER, there is an exception to the rule - when a harvesting run is started not by the timer, but manually, through the harvesting client page and/or API, the job can end up running on either of the multiple nodes. 

Admittedly,  this will not affect many otheaverse installations. (I'm actually not aware of another production installation at this point that is using a multiple server setup). But this would affect us, so I am a bit hesitant about doing this. 
And I'm not really positive as to what  the best reliable solution is. That ClientHarvestRun object would need an extra entry identifying the node actually running the harvest, maybe?

landreev: Another solution would be to change how the manually-run harvests are handled. I.e., instead of actually running the harvest it could create a one-time timer entry for 5 sec. from now...

landreev: I'm not 100% sure if being able to have these harvest-related  flags cleared automatically is worth adding any complicated code to address the multiple server scenario... 
So yet another, much simpler, solution would be to implement it as above - but with an option to disable this mechanism; so on startup the service above would check the setting before proceeding - `if (enableClearingHarvestJobsOnStartup) { ... }`. It will be enabled by default; so it would work for most installations. But then in the guide it will say "make sure to disable this mechanism if you are running multiple servers!". Somewhere in the same part of the guide where we mention that `dataverse.timerServer`setting.

src/main/java/edu/harvard/iq/dataverse/harvest/client/ClientHarvestRun.java
JingMa87: @landreev I guess handling it quietly would be ok. So instead of DELETE FAILED or FAILED we just show nothing?

src/main/java/edu/harvard/iq/dataverse/harvest/client/ResetHarvestInProgressBean.java
JingMa87: @landreev My plan is to add a sane default for a configurable DB key "EnableClearingHarvestJobsOnStartup" using the SettingsServiceBean, do you have any objections?

pdurbin:
I'm not sure of the status of this PR.  @JingMa87 recently explained that he switched to another job: https://github.com/IQSS/dataverse/issues/7412#issuecomment-1273464175

@janvanmansum has been helping explain which issues and PRs are a priority for DANS.

For this one, @landreev recently added the ability to stop a harvest in progress...

- #9187

... which seems related, at least. So I'm not sure if we need this PR or if it should be rewritten (the branch is gone from GitHub, by the way).

pdurbin:
@JingMa87 I'm closing this PR but if you're still interested in it, please reply! Thanks!

PR #7057, opened 2020-07-06
7056 Skip unknown types in a dataset which cause the whole dataset to…

### LINKED ISSUE #7056: Unknown types in a dataset causes the whole dataset to fail from being harvested
When the issue occurs: When harvesting using dataverse_json and there's custom types like "titulo" or "mraCollection"
Which page it occurs on: Harvesting Dashboard
To whom it occurs: People who harvest using dataverse_json
Version: 4.20

image

JingMa87:
… fail from being harvested

**What this PR does / why we need it**: Makes datasets with unknown types harvestable using dataverse_json

**Which issue(s) this PR closes**: 7056

Closes #7056 

**Special notes for your reviewer**: Good fix

**Suggestions on how to test this**: Harvest the Princeton_Authored_Datasets OAI set from the harvard repo using dataverse_json. Without this fix you will get 267 successful harvest and 2 fails. With the fix you will harvest all 269 datasets. The controversial datasets are https://dataverse.harvard.edu/api/datasets/export?exporter=dataverse_json&persistentId=doi:10.7910/DVN/B6OJKG and https://dataverse.harvard.edu/api/datasets/export?exporter=dataverse_json&persistentId=doi:10.7910/DVN/I5O6OS.

**Is there a release notes update needed for this change?**: I think it's an important change for harvesting from another dataverse.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
@@ -1,23 +1,7 @@
 package edu.harvard.iq.dataverse.util.json;
 
 import com.google.gson.Gson;
-import edu.harvard.iq.dataverse.ControlledVocabularyValue;
-import edu.harvard.iq.dataverse.DataFile;
-import edu.harvard.iq.dataverse.DataFileCategory;
-import edu.harvard.iq.dataverse.Dataset;
-import edu.harvard.iq.dataverse.DatasetField;
-import edu.harvard.iq.dataverse.DatasetFieldConstant;
-import edu.harvard.iq.dataverse.DatasetFieldCompoundValue;
-import edu.harvard.iq.dataverse.DatasetFieldServiceBean;
-import edu.harvard.iq.dataverse.DatasetFieldType;
-import edu.harvard.iq.dataverse.DatasetFieldValue;
-import edu.harvard.iq.dataverse.DatasetVersion;
-import edu.harvard.iq.dataverse.Dataverse;
-import edu.harvard.iq.dataverse.DataverseContact;
-import edu.harvard.iq.dataverse.DataverseTheme;
-import edu.harvard.iq.dataverse.FileMetadata;
-import edu.harvard.iq.dataverse.MetadataBlockServiceBean;
-import edu.harvard.iq.dataverse.TermsOfUseAndAccess;
+import edu.harvard.iq.dataverse.*;
 import edu.harvard.iq.dataverse.TermsOfUseAndAccess.License;
 import edu.harvard.iq.dataverse.api.Util;
 import edu.harvard.iq.dataverse.api.dto.FieldDTO;
@@ -32,14 +16,14 @@ import edu.harvard.iq.dataverse.workflow.Workflow;
 import edu.harvard.iq.dataverse.workflow.step.WorkflowStepData;
 import org.apache.commons.validator.routines.DomainValidator;
 
+import javax.json.*;
+import javax.json.JsonValue.ValueType;
 import java.io.StringReader;
 import java.sql.Timestamp;
 import java.text.ParseException;
 import java.util.*;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import javax.json.*;
-import javax.json.JsonValue.ValueType;
 
 /**
  * Parses JSON objects into domain objects.
@@ -394,7 +378,10 @@ public class JsonParser {
             List<DatasetField> fields = new LinkedList<>();
             for (JsonObject fieldJson : fieldsArray.getValuesAs(JsonObject.class)) {
                 try {
-                    fields.add(parseField(fieldJson, testType));
+                    DatasetField field = parseField(fieldJson, testType);
+                    if (field != null) {
+                        fields.add(field);
+                    }
                 } catch (CompoundVocabularyException ex) {
                     DatasetFieldType fieldType = datasetFieldSvc.findByNameOpt(fieldJson.getString("typeName", ""));
                     if (lenient && (DatasetFieldConstant.geographicCoverage).equals(fieldType.getName())) {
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
```diff
@@ -1,23 +1,7 @@
...
-import edu.harvard.iq.dataverse.DatasetFieldType;
-import edu.harvard.iq.dataverse.DatasetFieldValue;
-import edu.harvard.iq.dataverse.DatasetVersion;
-import edu.harvard.iq.dataverse.Dataverse;
-import edu.harvard.iq.dataverse.DataverseContact;
-import edu.harvard.iq.dataverse.DataverseTheme;
-import edu.harvard.iq.dataverse.FileMetadata;
-import edu.harvard.iq.dataverse.MetadataBlockServiceBean;
-import edu.harvard.iq.dataverse.TermsOfUseAndAccess;
+import edu.harvard.iq.dataverse.*;
```
pdurbin, line 4: I mentioned this in a review on another pull request but we favor using the full import statement instead of wildcards.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
```diff
@@ -566,7 +553,8 @@ public DatasetField parseField(JsonObject json, Boolean testType) throws JsonPar
     
 
         if (type == null) {
-            throw new JsonParseException("Can't find type '" + json.getString("typeName", "") + "'");
+            logger.info("Can't find type '" + json.getString("typeName", "") + "'");
```
pdurbin, line 556: Is this going to happen a lot? If so, should it be reduced from "info" to "fine" so it doesn't clutter the logs?

jggautier:
> We've never promoted using the dataverse_json format for harvesting (I believe DDI is preferred) but this should make it more possible to use it.

Maybe not in the guides but I've promoted it to reduce information loss and most of the harvesting Harvard Dataverse does from other Dataverse installations is using dataverse_json. I would think that DDI Codebook would be preferred if all metadata could be mapped to it, but that's probably going to be less and less true with increasing support for other standards.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
JingMa87: I'll reverse the change, my Intellij automatically does this. Super nice feature though otherwise.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
JingMa87: @pdurbin This'll happen potentially multiple thousands of times when doing a harvesting run on a big OAI set for the first time. So I'll just change it to "fine".

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
pdurbin: Yeah, I've heard that about Intellij. Thanks for reversing the change.

kcondon:
@JingMa87 I am still seeing the two failures.

JingMa87:
> @JingMa87 I am still seeing the two failures.
> 
> ```
> <record>
>   <date>2020-07-10T16:28:10</date>
>   <millis>1594398490946</millis>
>   <sequence>8513</sequence>
>   <logger>edu.harvard.iq.dataverse.harvest.client.HarvesterServiceBean.kc0605202020-07-10T16-27-22</logger>
>   <level>SEVERE</level>
>   <class>edu.harvard.iq.dataverse.harvest.client.HarvesterServiceBean</class>
>   <method>logGetRecordException</method>
>   <thread>187</thread>
>   <message>Exception processing getRecord(), oaiUrl=https://dataverse.harvard.edu/oai, identifier=doi:10.7910/DVN/B6OJKG, edu.harvard.iq.dataverse.api.imports.ImportException, Failed to import harvested dataset: class edu.harvard.iq.dataverse.engine.command.exception.CommandException (Command [DatasetCreate dataset:20721] failed: Exception thrown from bean: javax.ejb.EJBTransactionRolledbackException: Exception thrown from bean: org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at http://localhost:8983/solr/collection1: ERROR: [doc=dataset_20721] unknown field 'mraCollection')</message>
> </record>
> ```

@kcondon I can't reproduce the error. I've looked into the DatasetCreate Command but without the actual error happening it's like looking for a needle in a haystack. Are you on the latest version with the latest changes? What happens if you re-run the harvest, does it work the second time?

kcondon:
@JingMa87 I have run it again and still see the failures. Running it again just results in 0 harvested, 0 failed. Looks like failures do are not harvested again.

JingMa87:
@kcondon I pushed all my changes to the branch already. What if you delete the harvesting client, add it again and re-run, do you still get the issue?

kcondon:
Yes, I have done that several times, every time I test. You could add debug code and I could run a harvest to see what it thinks it is doing.
________________________________

JingMa87:
Does this query in the database return something for you? If so, this row is not supposed to be there and stops the field type mraCollection from being filtered out by my code change. Then, it gets passed to Solr and the error happens.
`SELECT * FROM public.datasetfieldtype where name = 'mraCollection';`

I've looked into adding debugging code, but the class where the error happens (CreateHarvestedDatasetCommand) is a whole big mess of polymorphism, generics and interfaces gone wrong. When I use the debugger, I can't inspect the execute() function of that Command where the error happens so I'm relegating to reading the code. Hopefully this helps.

kcondon:
@JingMa87 Yes, that sql query returns a record. Why do you say it should not be there? Also, wouldn't you have seen the same record in your own test? I understand about the code complexity. I was thinking about a lower tech debugging of adding debug logging if that helps.

JingMa87:
@kcondon I'll make a new ticket for clients that do have the field in the DB and will try to fix that in the coming days.

PR #7062, opened 2020-07-07
4597 Fix invalid identify response.

### LINKED ISSUE #4597: OAI PMH: Invalid Identify-Response
There seems to be a problem with the OAI-PMH-Identify-Response.

Different OAI-PMH validators report an invalidity of the
Dataverse OAI-Identify response:

For example:
- http://validator.oaipmh.com/
- http://oval.base-search.net/

OAI-PMH URL: https://heidata.uni-heidelberg.de/oai

ERROR: Identify response well-formed but invalid: Element 'XOAIDescription': This element is not expected. Expected is ( ##other{http://www.openarchives.org/OAI/2.0/}* )., line 1

Our Dataverse version is 4.8.2

JingMa87:
**What this PR does / why we need it**: Changes the Identify response for a OAI PMH request so it's XSD valid.

**Which issue(s) this PR closes**: 4597

Closes #4597 

**Special notes for your reviewer**: It's looking good

**Suggestions on how to test this**: Example call: https://dataverse.harvard.edu/oai?verb=Identify

**Is there a release notes update needed for this change?**: Perhaps

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/XdataProvider.java
+++ b/src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/XdataProvider.java
@@ -6,6 +6,7 @@ import com.lyncode.xoai.dataprovider.exceptions.*;
 import com.lyncode.xoai.dataprovider.handlers.*;
 import com.lyncode.xoai.exceptions.InvalidResumptionTokenException;
 import com.lyncode.xoai.dataprovider.model.Context;
+import com.lyncode.xoai.model.oaipmh.Identify;
 import com.lyncode.xoai.model.oaipmh.OAIPMH;
 import com.lyncode.xoai.model.oaipmh.Request;
 import com.lyncode.xoai.dataprovider.parameters.OAICompiledRequest;
@@ -76,7 +77,9 @@ public class XdataProvider {
 
             switch (request.getVerbType()) {
                 case Identify:
-                    response.withVerb(identifyHandler.handle(parameters));
+                    Identify identify = identifyHandler.handle(parameters);
+                    identify.getDescriptions().clear(); // We don't want to use the default description
+                    response.withVerb(identify);
                     break;
                 case ListSets:
                     response.withVerb(listSetsHandler.handle(parameters));


src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -193,10 +193,48 @@ private RepositoryConfiguration createRepositoryConfiguration() {
...
+            Element oaiIdentifier = doc.createElement("oai-identifier");
+            oaiIdentifier.setAttribute("xmlns", "http://www.openarchives.org/OAI/2.0/oai-identifier");
+            oaiIdentifier.setAttribute("xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance");
+            oaiIdentifier.setAttribute("xsi:schemaLocation", "http://www.openarchives.org/OAI/2.0/oai-identifier http://www.openarchives.org/OAI/2.0/oai-identifier.xsd");
+            doc.appendChild(oaiIdentifier);
+            Element scheme = doc.createElement("scheme");
+            scheme.appendChild(doc.createTextNode("oai"));
+            oaiIdentifier.appendChild(scheme);
+            Element repositoryIdentifier = doc.createElement("repositoryIdentifier");
+            repositoryIdentifier.appendChild(doc.createTextNode("dataverse.org"));
```
pdurbin, line 216: In #4597 @JingMa87 and I have been discussing making "repositoryIdentifier" (and possibly other identifiers configurable).

JingMa87:
@pdurbin The coverage decreased, do you need a test for all the functions people add?

pdurbin:
@JingMa87 no, that's not necessary. I see you made some changes. Are you ready for more code review?

JingMa87:
@pdurbin Go ahead.

src/main/webapp/editFilesFragment.xhtml
```diff
@@ -282,7 +282,7 @@
                     
                     <p:commandButton id="updateEditDataFilesButtonsForUpload" action="#{EditDatafilesPage.uploadFinished()}" update="datasetForm:editDataFilesButtons" rendered="#{showFileButtonUpdate}" style="display:none"/>
                     <p:commandButton id="updateEditDataFilesButtonsForDelete" action="#{EditDatafilesPage.deleteFilesCompleted()}" update="datasetForm:editDataFilesButtons" rendered="#{showFileButtonUpdate}" style="display:none"/>
-                    <p:commandButton id="AllUploadsFinished" action="#{EditDatafilesPage.uploadFinished()}" update="@form,datasetForm:fileUpload,datasetForm:dropBoxUserButton,datasetForm:uploadMessage,datasetForm:rsyncPanel,datasetForm:filesCounts,datasetForm:filesTable" oncomplete="javascript:bind_bsui_components();javascript:uploadWidgetDropMsg();" style="display:none"/>
```
pdurbin, line 285: Why is this change to `@form` in this pull request?

src/main/webapp/dataset.xhtml
```diff
@@ -1589,7 +1589,7 @@
                         </ui:fragment>
                         <div class="button-block">
                             <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" 
-                                             onclick="PF('publishDataset').hide();PF('blockDatasetForm').hide();" action="#{DatasetPage.releaseDataset}" immediate="true"/>
```
pdurbin, line 1592: Why is this change to `immediate="true"` true` in this pull request?

doc/release-notes/4597-fix-invalid-identify-response.md
```diff
@@ -0,0 +1 @@
+Added the following database options: Scheme, RepositoryIdentifier, Delimiter, SampleIdentifier.
```
pdurbin, line 1: Can you please change all these (in the code too, of course) to start with "Harvesting"? That is:

:HarvestingScheme, :HarvestingRepositoryIdentifier, :HarvestingDelimiter, :HarvestingSampleIdentifier

Otherwise, they're too generic, especially :Scheme or :Delimiter.

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -180,6 +180,11 @@ private RepositoryConfiguration createRepositoryConfiguration() {
         String dataverseName = dataverseService.findRootDataverse().getName();
         String repositoryName = StringUtils.isEmpty(dataverseName) || "Root".equals(dataverseName) ? "Test Dataverse OAI Archive" : dataverseName + " Dataverse OAI Archive";
         InternetAddress internetAddress = MailUtil.parseSystemAddress(settingsService.getValueForKey(SettingsServiceBean.Key.SystemEmail));
+        String scheme = settingsService.getValueForKey(SettingsServiceBean.Key.Scheme, "oai");
+        String repositoryIdentifier = settingsService.getValueForKey(SettingsServiceBean.Key.RepositoryIdentifier, "dataverse.org");
+        String delimiter = settingsService.getValueForKey(SettingsServiceBean.Key.Delimiter, ":");
+        String sampleIdentifier = settingsService.getValueForKey(SettingsServiceBean.Key.SampleIdentifier, "doi:10.7910/DVN/1HE30F");
```
pdurbin, line 186: In http://oval.base-search.net this DOI is giving an error:

 ERROR: Identify response well-formed but invalid: Element '{http://www.openarchives.org/OAI/2.0/oai-identifier}sampleIdentifier': [facet 'pattern'] The value 'doi:10.7910/DVN/1HE30F' is not accepted by the pattern 'oai:[a-zA-Z][a-zA-Z0-9\-]*(\.[a-zA-Z][a-zA-Z0-9\-]*)+:[a-zA-Z0-9\-_\.!~\*'\(\);/\?:@&=\+$,%]+'., line 5

In context:

Screen Shot 2020-07-21 at 12 23 12 PM

pdurbin:
As of 7ebdb42 one validator was looking good:

## before

valBefore

## after

valAfter

But another one was still showing errors:

## before

ovalBefore

## after

ovalAfter

If it's possible to fix all of these, great! If not, no problem, we're happy for any improvement! Thanks!

doc/release-notes/4597-fix-invalid-identify-response.md
pdurbin: Whoops, I jumped the gun. Can you please have these start with "Oiapmh" instead? Like this:

:OaipmhScheme, :OaipmhRepositoryIdentifier, :OaipmhDelimiter, :OaipmhSampleIdentifier

Also, can you please document these in doc/sphinx-guides/source/installation/config.rst? Thanks!

src/main/webapp/editFilesFragment.xhtml
JingMa87: Good point, I was testing some new functionality but inadvertently pushed these changes so I'll revert them. Same applies to `immediate="true"`.

JingMa87:
@pdurbin The reason for the error is that the sampleIdentifier has to start with "oai", which is actually not correct for Dataverse since a typical identifier starts with "doi". If this is a problem, we don't necessarily have to use these elements like scheme and sampleIdentifier. The parent `<description>` element is both optional and extensible and the content depends on the XSD schema you add as an attribute to the child element, which in our case is the `<oai-identifier>` element with a schema from the OAI website: http://www.openarchives.org/OAI/2.0/oai-identifier.xsd. So if we don't want to add a sample identifier which is not representative for Dataverse, we can leave out the description element all together since the minOccurs of that element is 0 and we'd still pass all validation tests.

pdurbin:
@JingMa87 I love the idea of removing the description element and having fewer database options.

JingMa87:
@pdurbin You can review the changes now. That XOAI library is really not user friendly tbh. :frowning:

kcondon:
@pdurbin Would you mind checking with Leonid on the validity of that approach? I think that should happen before it gets to QA.

pdurbin:
@kcondon sure. Can do. For now I parked this in code review.

landreev:
That `<description>` element isn't used for anything; so it should be safe to remove. We weren't putting it there on purpose either, it was just the default behavior of the XOAI library. 

I'm surprised this was causing the response to be invalid though; I thought the spec basically said there could be any free style xml inside that `<description>...</description>` block, as long as all the tags were properly closed, etc. (And I do remember the our site was passing the OAI validator, after we implemented it). But if any specific software doesn't like it as is, sure, let's remove it.

pdurbin:
Thanks for the comment from @landreev and the 👍 from @jggautier. I'm moving this to QA.

JingMa87:
@landreev I don't know if they changed it in the meantime but the OAI-PMH XSD says that the content of the description element should have an XSD which determines the child elements (http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd).

image

PR #7063, opened 2020-07-08
Update account.rst

### LINKED ISSUE #7055: Addition to User Guide
The section Notifications in the Dataverse User Guide (cf. [http://guides.dataverse.org/en/latest/user/account.html](http://guides.dataverse.org/en/latest/user/account.html)) says that as a user you will receive notification when "[a]nother Dataverse user has requested access to a restricted file in one of your datasets". I suggest you add that this is only true if the user has the Curator role. A user with only the Contributor role will not get this notification. Instead the curator of the dataverse where the dataset/file has been published will get a notification about requested access to a restricted file. In our installation, the curator will then contact the depositor to check whether access should be granted. It makes sense that only users with the Curator role get this notification as the Curator role is needed to grant access to a restricted file.

philippconzett:
Closes #7055

This PR adds information to the existing guide. Please feel free to adjust the formulation.

### DIFF
--- a/doc/sphinx-guides/source/user/account.rst
+++ b/doc/sphinx-guides/source/user/account.rst
@@ -160,7 +160,7 @@ You will receive a notification when:
 
 - You've created your account
 - You've created a dataverse or added a dataset
-- Another Dataverse user has requested access to restricted files in a dataset that you published (If instead you submitted the dataset for review, the curators of the dataverse that contains your dataset will get a notification about requests to access your restricted files.)
+- Another Dataverse user has requested access to restricted files in a dataset that you published (if you submitted your dataset for review and it was published by a curator, the curators of the dataverse that contains your dataset will get a notification about requests to access your restricted files.)
 - A file in one of your datasets has finished the ingest process
 
 Notifications will only be emailed one time even if you haven't read the notification on the Dataverse site.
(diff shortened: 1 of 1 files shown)


djbrooke:
Thanks @philippconzett !

@jggautier - would you be OK reviewing this PR and either making changes or working with @philippconzett on this?

jggautier:
> Another Dataverse user has requested access to a restricted file in one of your datasets (Note! A user with only the Contributor role will not get this notification. Instead the curator of the dataverse where the dataset/file has been published will get a notification about requested access to a restricted file.)

Hi @philippconzett. I rewrote a little to try to keep the section's "you" language and the perspective and emphasize that the user gets the notification if she published the dataset but not if she submitted it for review.

Another Dataverse user has requested access to restricted files in a dataset that you published (if you submitted the dataset for review and weren't able to publish it, the curators of the dataverse that contains your dataset will get a notification about requests to access your restricted files.)

How does that sound? If it sounds good to you, could you make the change in your PR?

philippconzett:
Thanks, @jggautier! I think the request access to restricted file feature is only available for published datasets, so I think we should change the text in brackets slightly. Here's my suggestion:

Another Dataverse user has requested access to restricted files in a dataset that you published (if your dataset was (reviewed and) published by a curator, the curators of the dataverse that contains your dataset will get a notification about requests to access your restricted files.)

jggautier:
@philippconzett, I see your point. Could you take a look at the latest changes?

@djbrooke, once the update is finished, do I move this to the QA column?

djbrooke:
@jggautier if you are @philippconzett are OK with it feel free to move it on. Thank you to you both!

PR #7064, opened 2020-07-08
7054 payara bom

### LINKED ISSUE #7054: Dependency management: start using Payara BOM
I came across this very interesting article in the Payara blog: https://blog.payara.fish/align-dependencies-with-payara-bom

As we have few dependencies with a comment about "keep version aligned to Payara release", we should start using the BOM to ease dependency management.

PR forthcoming.

poikilotherm:
A lot of libraries used by us (and again by those libs) are provided by the Payara application server.

There is no need to bundle those in our WAR (making it smaller) and it's always a good idea to keep things inline with each other, to avoid conflicts between libraries. The newly released Payara BOM is making this significantly easier for basic libraries as Jackson, Jakarta, etc.

Also, our application server maintains a few patched versions of libraries. A good example is the JSF patch recently removed (see #7036), now being bundled with the upstream installation. This change to `pom.xml` includes the repository of these patched versions, so we use them locally and during tests, too.

Closes #7054 

The Jackson BOM has been removed due to the fact that the Payara BOM already provides that one.

Nada. Usual things. I didn't move much (the deps cleanup is still to come), so this should not have a bigger impact.

Nada.

Nada.

**Additional documentation**:
https://blog.payara.fish/align-dependencies-with-payara-bom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,11 +26,8 @@
         <skipUnitTests>false</skipUnitTests>
     
         <jakartaee-api.version>8.0.0</jakartaee-api.version>
-        <!-- Make this correspond to the version used in Payara: 5.201 uses (patched) 2.3.14 -->
-        <mojarra.version>2.3.14</mojarra.version>
+        <payara.version>5.2020.2</payara.version>
         <aws.version>1.11.762</aws.version>
-        <!-- Should always correspond with Payara bundled version: 5.201 uses 2.10.2 -->
-        <jackson.version>2.10.2</jackson.version>
         <commons.logging.version>1.2</commons.logging.version>
         <httpcomponents.client.version>4.5.5</httpcomponents.client.version>
         <junit.version>4.12</junit.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -26,11 +26,8 @@
         <skipUnitTests>false</skipUnitTests>
     
         <jakartaee-api.version>8.0.0</jakartaee-api.version>
-        <!-- Make this correspond to the version used in Payara: 5.201 uses (patched) 2.3.14 -->
-        <mojarra.version>2.3.14</mojarra.version>
+        <payara.version>5.2020.2</payara.version>
```
pdurbin, line 29: Does this mean that this new "payara.version" property needs to be bumped every time we want to move to a new version of Payara?

pom.xml
```diff
@@ -90,24 +87,35 @@
             <name>Local repository for hosting jars not available from network repositories.</name>
             <url>file://${project.basedir}/local_lib</url>
         </repository>
+        <repository>
+            <id>payara-patched-externals</id>
```
pdurbin: Does this further tie us to Payara? Will it be harder to move to Liberty or Wildfly?

pom.xml
```diff
@@ -63,6 +60,17 @@
     <!--Maven checks for dependendies from these repos in the order shown in the pom.xml
         This isn't well documented and seems to change between maven versions -MAD 4.9.4 -->
     <repositories>
+        <repository>
+            <id>payara-patched-externals</id>
```
pdurbin, line 64: Does this further tie us to Payara? Would it be harder to switch to Liberty or Wildfly, for example?

pom.xml
```diff
@@ -26,11 +26,8 @@
         <skipUnitTests>false</skipUnitTests>
     
         <jakartaee-api.version>8.0.0</jakartaee-api.version>
-        <!-- Make this correspond to the version used in Payara: 5.201 uses (patched) 2.3.14 -->
-        <mojarra.version>2.3.14</mojarra.version>
```
pdurbin, line 30: Does the version of Mojarra no longer matter?

pom.xml
poikilotherm: It should be moved. This is not a must, but a strong should.

This hasn't happened in the past with GF 4.1, as that appserver hadn't any updates. It won't be much of a hassle to do this and it makes testing with newer appserver versions easier.

pom.xml
poikilotherm: Yes. But we are pretty bound to Pasta anyway, as we rely on their patched lib releases. As long as those don't hit upstream, it would lead to troubles when deploying to Wildfly/Liberty/... anyway.

pom.xml
poikilotherm: It does, but it is already included in the BOM. It should be kept the same as the appserver anyway, so that updating is now obsolete.

Instead, we change the appserver version in `payara.version`.

pom.xml
poikilotherm: By the way, as the article outlined, too: you would do this with the Wildfly/Jetty/Liberty BOM when using those appservers.

You might think about creating different maven profiles when multi appserver compatibility becomes a feature.

pom.xml
pdurbin: Whoops. I asked this twice. Here's the answer from @poikilotherm from https://github.com/IQSS/dataverse/pull/7064#discussion_r452391625

"Yes. But we are pretty bound to Payara anyway, as we rely on their patched lib releases. As long as those don't hit upstream, it would lead to troubles when deploying to Wildfly/Liberty/... anyway."

pom.xml
pdurbin: Ok. Maybe after we've done a few of these Payara upgrades (pull request #7036 was the last one), we'll get into a nice pattern of which files need to be touched (similar to how we know which files to touch when making a release, and even document them). So heads up here to @donsizemore (who made the last bump) that a new line is being added.

PR #7085, opened 2020-07-14
Fix for #7060 direct download causes RedirectException

### LINKED ISSUE #7060: direct download causes RedirectException
On the current develop branch, setting -Ddataverse.files.s3.download-redirect=true and trying to download a file gives me:

{
"status": "ERROR",
"code": 500,
"type": "RedirectionException",
"message": "Internal server error. More details available at the server logs.",
"incidentId": "113fbb33-f4e0-402e-b9a9-5417c3c9822b"
}

If I remove that jvm option, I can download the file, so I think the underlying s3 store is configured correctly. WHen I use:
bin/asadmin set-log-levels edu.harvard.iq.dataverse.api.DownloadInstanceWriter=FINE
the presigned download URL that shows in the log is actually correct/works, so it looks like something has changed in how the redirect is processed in Dataverse. 

FWIW: I used the ansible scripts to create an EC2 instance and configured s3 manually after that.

qqmyers:
**What this PR does / why we need it**: Allows download redirects

Closes #7060 direct download causes RedirectException

**Special notes for your reviewer**: Don't know why this is needed now and wasn't needed in 4.20

**Suggestions on how to test this**: Use S3, turn on download redirects - it should work (and won't without this PR)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/BundleDownloadInstanceWriter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/BundleDownloadInstanceWriter.java
@@ -12,12 +12,12 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.IOException;
 
+import javax.ws.rs.InternalServerErrorException;
+import javax.ws.rs.NotFoundException;
 import javax.ws.rs.WebApplicationException;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
-import javax.ws.rs.core.Response;
-
 import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.ext.Provider;
 
@@ -196,10 +196,10 @@ public class BundleDownloadInstanceWriter implements MessageBodyWriter<BundleDow
                 }
             }
         } catch (IOException ioex) {
-            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+            throw new InternalServerErrorException();
         }
 
-        throw new WebApplicationException(Response.Status.NOT_FOUND);
+        throw new NotFoundException();
 
     }
 
--- a/src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
@@ -244,7 +244,7 @@ public class DownloadInstanceWriter implements MessageBodyWriter<DownloadInstanc
                         }
                         
                         if (redirect_url_str == null) {
-                            throw new WebApplicationException(new ServiceUnavailableException());
+                            throw new ServiceUnavailableException();
                         }
                         
                         logger.fine("Data Access API: direct S3 url: "+redirect_url_str);
@@ -274,7 +274,7 @@ public class DownloadInstanceWriter implements MessageBodyWriter<DownloadInstanc
                             logger.fine("Issuing redirect to the file location on S3.");
                             throw new RedirectionException(response);
                         }
-                        throw new WebApplicationException(new ServiceUnavailableException());
+                        throw new ServiceUnavailableException();
                     }
                 }
                 
@@ -368,7 +368,7 @@ public class DownloadInstanceWriter implements MessageBodyWriter<DownloadInstanc
             }
         }
         
-        throw new WebApplicationException(Response.Status.NOT_FOUND);
+        throw new NotFoundException();
 
     }
     
--- a/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/InternalServerErrorExceptionHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/InternalServerErrorExceptionHandler.java
@@ -1,0 +1,39 @@
+package edu.harvard.iq.dataverse.api.errorhandlers;
+
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.json.Json;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.InternalServerErrorException;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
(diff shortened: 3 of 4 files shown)


pdurbin:
@qqmyers pull request #6974 is the one I mentioned in tech hours just now.

"ThrowableHandler" has this as a comment: "Produces a generic 500 message for the API, being a fallback handler for not specially treated exceptions."

In addition a "JsonParseExceptionHandler" was added.

I have no idea if this is what changed things or not.

qqmyers:
@pdurbin - ah - you've got it. There was no ThrowableHandler in 4.20! So that one is probably cutting off the default handling of the RedirectionException. I'm not sure what the need for it is other than to add logging info for other errors - perhaps it should be more selective? In any case, I think this makes it clearer why this PR works/is needed. Could be a new issue to question whether the ThrowableHandler is needed? Or should we try to resolve now?

scolapasta:
Good find @pdurbin.

@qqmyers I'd suggest we resolve this now - if we don't beed it we don't need this PR. If we do, this PR is fine.

qqmyers:
@poikilotherm - looks like your addition of src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/ThrowableHandler.java  in #6936 broke the handling for the RedirectException used in direct S3 downloads. Can you shed some light on if/why it's needed? Is it just to provide additional logging versus the default? The PR here adds a handler specifically for redirects, but it's only needed if ThrowableHandler blocks the default handling, so we're trying to figure out whether it's better to remove ThrowableHandler, to constrain it to not catch RedirectionExceptions somehow, or to use this PR. Any thoughts?

poikilotherm:
**tl;dr:** My vote: 1) Add the exception handler in this PR, completing the gang. :people_holding_hands:  2) Refactor some other places to avoid throwing unspecific `WebApplicationException`s. :v:  3) Get everyone happy. :smile_cat: 

:bug: :bug: :bug: :bug: :bug: :bug: :bug: 

Hey folks, let me shed some light on this. :bulb: 

It's a good common practice to catch unknown exceptions, log them properly and present a meaningful error message to the user. This is what `ThrowableHandler` does:
- Avoid sending appserver default HTML error pages when users expect JSON errors from a JSON REST API.
- Log exceptions with details about when and where this happened, what was the Request URL, etc etc.
- Avoid sending possible sensible information over the wire (you can't know).
- Use proper HTML statuses instead of firing meaningless exception messages to a user that has no idea what's going on.

I took this as reason to scan the complete codescan for used JAX RS exceptions: :mag: 
```
rg -NI "javax.ws.rs.*Exception;" src/main | sort | uniq -c
      3 import javax.ws.rs.BadRequestException;
      2 import javax.ws.rs.ForbiddenException;
      1 import javax.ws.rs.NotAllowedException;
      5 import javax.ws.rs.NotFoundException;
      1 import javax.ws.rs.RedirectionException;
      4 import javax.ws.rs.ServiceUnavailableException;
      5 import javax.ws.rs.WebApplicationException;
```

All of those are subclasses of `javax.ws.rs.WebApplicationException`. Except for `RedirectException` and `WebApplicationException` all already feature a proper error handler in package `api.errorhandlers`.

Let's take a look at the usages of the generic top class: :mag: :mag: 
```
src/main/java/edu/harvard/iq/dataverse/api/BundleDownloadInstanceWriter.java
199:            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
202:        throw new WebApplicationException(Response.Status.NOT_FOUND);

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
85:                    //throw new WebApplicationException(Response.Status.SERVICE_UNAVAILABLE);
225:                        //throw new WebApplicationException(Response.Status.SERVICE_UNAVAILABLE);
247:                            throw new WebApplicationException(new ServiceUnavailableException());
277:                        throw new WebApplicationException(new ServiceUnavailableException());
371:        throw new WebApplicationException(Response.Status.NOT_FOUND);
```

Those occasions look like they could use a proper refactoring to their more specific exception variants. :recycle: 
I'd be happy to create a PR against @qqmyers branch if you don't want to do this yourself. :wink: 

I vote for keeping `ThrowableHandler` and taking care of exceptions. :+1:

qqmyers:
@poikilotherm - 

My understanding is that the framework already supplies exception handlers for these cases and one only has to override them if you need non-standard results. We don't for RedirectionException - that appears to be the way you are supposed to cause a redirect. Similarly, the docs for WebApplicationException say it does what you want - returns a 500 error with no additional info 'leaked'. Or it uses the status and or whole response you send if you want to specify those.

That said, I'm not opposed to wrapping unexpected Throwables to improve logging about them, but my guess is that catching them is also breaking some of the WebApplicationExceptions you note above (and any others), e.g. the WebApplicationException(new ServiceUnavailableException()); is throwing a 500 error now instead of 503. (The same would be true for any  NotAcceptableException, NotAuthorizedException, NotSupportedException, but I don't see any of those in the code). So the refactoring is not optional (for anything returning a non-500 response) and developers need to be aware that they can't use the other three classes.

Perhaps a useful follow-up would be to refactor the ExceptionHandler classes so they all extend from a base(s) that standardize the internal logging and support more of the default behaviors (a good student project?)

qqmyers:
I've updated the branch but I don't see the PR automatically updating as it usually does. Anyone have ideas?

Regardless - I added an InternalServerErrorExceptionHandler and removed the generic WebApplicationExceptions that were no longer correctly handled. I have not tested all the underlying error conditions. My guess is that the return codes will now be correct but the exact message returned for errors could be different from what it was, which won't be a problem unless we have tests checking that or are worried about that type of change to the API.

djbrooke:
Hey @poikilotherm - we think this is good to go for Dataverse 5. Please let us know if you have any additional comment. If not, we'll plan to move this ahead on the board and get it merged. Thanks!

landreev:
Just confirming that the extra redirect from under api/REST in my zipper branch is throwing the same exception under recent Payara versions.

qqmyers:
@landreev - yes. It catches all RedirectionExceptions and triggers an actual redirect to the URL in the response you put in the exception.

Issue #2911, opened 2016-02-03
File API download bypasses terms of use and guestbook


scolapasta:
Currently, when you download a file through the UI, all logic for creating a GuestbookResponse row is down before hitting the API to download the file. 

If you download the file directly from the API, you don't create a row here, so the count does not go up. Also this bypasses the terms of use and guestbook completely. We need to make sure a ) a row gets created, so counts are accurate, b) that we determine how we want to handle the bypassing of the terms of use (via a token?) rather than just acting like they don't exist.

djbrooke:
Updating this to cover terms of use and not increasing download count, which is covered in #3331.

solhm:
We are looking forward to this functionality because we are facing some issue related to copyright regarding organizations harvesting our dataverse using the API. Since they are getting a direct download link to the files and they put them on their websites, users are downloading them without any knowledge of or agreement to the terms of use.

pdurbin:
@solhm thanks for your comment. I just brought up "File API download bypasses terms of use" with @djbrooke @scolapasta and @sekmiller while discussing #3758.

pdurbin:
@alejandratenorio brought up this issue today and we've been discussing it at https://dataverse.zulipchat.com/#narrow/stream/379856-security/topic/No.20Restricted.20Files.20.2F.20Access.20conditions/near/427955246

alejandratenorio:
Hi all,

Possibly CIMMYT could collaborate on this. As @pdurbin suggested, we would like to have a proposal validated by you before any development. 
We think that the file download could work as follows, (it's our proposal v. 0.2):

**These are our assumptions:**
- You can create an API token only if you have a user on a Dataverse instance. At least we have each user's last name, first name and email address. Desirably the Affiliation.
- Using the API, anyone can download files with no access restrictions.
- If someone uses an API Token, we could know the user associated to that token, do not? 
- As a user, when you request access to a restricted datafile you must accept the Terms of Access for Restricted Files.

**File download:** 
**Terms of use:**
_As a user, when you request access to a restricted datafile you must accept the Terms of Access for Restricted Files and its Terms of use._

- If its dataset has no Terms of use & the datafile has no access restrictions:
o	No changes.

- If its dataset has no Terms of use & the datafile has access restrictions:
o	No changes, an API token is required. 

- If its dataset has Terms of use & the datafile has no access restrictions:
o	_An API token is required because we must be sure that a user accepts the terms of use._
[...]

PR #7086, opened 2020-07-14
add API to download all files by dataset #4529

### LINKED ISSUE #4529: Data Access API: Download by Dataset

pdurbin:

API users have long wanted a way to download all the files in a dataset using its persistent identifier (DOI or Handle).

Closes #4529

- ~~One thought I had just now is that there's no way to specify original format vs archival format. Should this be added?~~ I added docs and tests for `format=original` in 44cc815.
- I made no attempt to address #2911 which is about how terms and guestbook acceptance can by bypassed via API. In my testing, I was able to bypass a guestbook with required fields for name and email.
- I assume that Make Data Count is covered deeper in the API than I went (by existing APIs I'm calling into, I mean).

Follow the API docs. Try variations with restricted files, tabular files, guestbook/terms.

No.

Yes, included.

**Additional documentation**:

Included.

### DIFF
--- a/conf/docker-aio/run-test-suite.sh
+++ b/conf/docker-aio/run-test-suite.sh
@@ -8,4 +8,4 @@ fi
 
 # Please note the "dataverse.test.baseurl" is set to run for "all-in-one" Docker environment.
 # TODO: Rather than hard-coding the list of "IT" classes here, add a profile to pom.xml.
-mvn test -Dtest=DataversesIT,DatasetsIT,SwordIT,AdminIT,BuiltinUsersIT,UsersIT,UtilIT,ConfirmEmailIT,FileMetadataIT,FilesIT,SearchIT,InReviewWorkflowIT,HarvestingServerIT,MoveIT,MakeDataCountApiIT,FileTypeDetectionIT,EditDDIIT,ExternalToolsIT,AccessIT,DuplicateFilesIT -Ddataverse.test.baseurl=$dvurl
+mvn test -Dtest=DataversesIT,DatasetsIT,SwordIT,AdminIT,BuiltinUsersIT,UsersIT,UtilIT,ConfirmEmailIT,FileMetadataIT,FilesIT,SearchIT,InReviewWorkflowIT,HarvestingServerIT,MoveIT,MakeDataCountApiIT,FileTypeDetectionIT,EditDDIIT,ExternalToolsIT,AccessIT,DuplicateFilesIT,DownloadFilesIT -Ddataverse.test.baseurl=$dvurl
--- a/doc/release-notes/4529-download-by-dataset.md
+++ b/doc/release-notes/4529-download-by-dataset.md
@@ -1,0 +1,6 @@
+In previous versions of Dataverse, downloading all files from a dataset via API was a two step process:
+
+- Find all the database id of the files.
+- Download all the files, using those ids (comma-separated).
+
+Now you can download all files from a dataset (assuming you have access to them) via API by passing the dataset persistent ID (PID such as DOI or Handle) or the dataset's database id. Versions are also supported like with the "download metadata" API you can pass :draft, :latest, :latest-published, or numbers (1.1, 2.0).
--- a/doc/sphinx-guides/source/api/dataaccess.rst
+++ b/doc/sphinx-guides/source/api/dataaccess.rst
@@ -14,6 +14,8 @@ Downloading All Files in a Dataset
 
 The "download all" API downloads as many files as possible from a dataset as a zipped bundle.
 
+By default, tabular files are downloaded in their "archival" form (tab-separated values). To download the original files (Stata, for example), add ``format=original`` as a query parameter.
+
 There are a number of reasons why not all of the files can be downloaded:
 
 - Some of the files are restricted and your API token doesn't have access (you will still get the unrestricted files).
--- a/doc/sphinx-guides/source/api/dataaccess.rst
+++ b/doc/sphinx-guides/source/api/dataaccess.rst
@@ -7,6 +7,73 @@ More advanced features of the Access API include format-specific transformations
 .. contents:: |toctitle|
    :local:
 
+.. _download-all-api:
+
+Downloading All Files in a Dataset
+----------------------------------
+
+The "download all" API downloads as many files as possible from a dataset as a zipped bundle.
+
+There are a number of reasons why not all of the files can be downloaded:
+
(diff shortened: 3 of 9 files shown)


pdurbin:
Ok, I'm ready for more review. I updated the description above to reflect that I added docs and tests for `format=original` (no code changes required).

I also did a quick test of a dataset with a guestbook. I made no attempt to address #2911 which is about how guestbooks and terms can be bypassed via API. The UI says name and email are required (screenshot below) but the API continues to let you have the files.

Screen Shot 2020-07-15 at 4 19 07 PM

scolapasta:
I'm in favor of: /api/access/dataset/

landreev:
I'm not requesting this as a change, in this PR, but just want to put this on record: there is a potential optimization that can be added - instead of doing a full isAccessAuthorized() lookup for every file, it is possible to cache some permission information (for ex. - you don't really need to figure out whether this user has the permission to view this unpublished dataset every time from scratch). Which could in certain cases result in non-trivial savings... I remember this because *we used to do that*, in /api/access/datafiles/... Until we realized that even though that API call was originally created for downloading multiple files from the dataset page, arbitrary file ids spanning multiple datasets could be supplied to it directly - so extrapolating any file permission info, under the assumption that all the other files are within the same dataset was obviously wrong and dangerous. But that code is still in github history, and now that we have a call where staying within the same dataset is guaranteed, we could potentially revive it. 
It's not 100% clear to me whether it's worth bothering with - so I'm going to open a new issue for it; and we could decide later on whether we want to touch this.

pdurbin:
Ok, in 4da877b I renamed the endpoint from "downloadAll" to "dataset". Back to code review.

landreev:
Just a quick note on " how guestbooks and terms can be bypassed via API" (#2911) mentioned above. This is completely out of scope for this issue, obviously. #2911 is a very known limitation of the API; it's not something we overlooked or forgot about - it was a decision that the management made at the very beginning of v.4, to allow machine access to files that otherwise require terms of use and/or guestbooks. v3.0 was using the other extreme - it was not possible to download a file under terms of use from a script, without some screen scraping and faking the checkbox click on the form. 
We should finally address this. But there is no particularly easy or quick way to address it. For example, while it is not difficult to make the API accept some extra JSON to populate the fields of a guestbook entry, it is not clear if that would be useful at all. Once you provide a way to fill these entries automatically, it does in a sense defeat the purpose of a guestbook, that was to collect information from human users. So this would take some research, to create a system that's actually useful. 

One thing we could implement relatively quickly and easily, would be to make it configurable: to have a switch, system-wide and/or scoped within dataverse/datasets, for allowing this API bypass of the terms of used. If set, it the API would return 403 when called on a file that has terms of use and/or required guestbook. If not, the current behavior would continue (but maybe the guestbook record should be explicitly marked as "machine access", for MDC, and other, purposes). The current setup, to allow the API to bypass terms of use completely, was implemented after some deliberations that applied to the requirements of the *Harvard* dataverse; but we know by now that other instances are not happy with it - so at least giving them an easy way to disable it will be helpful. 

(I'll copy this into the original issue)

kcondon:
@pdurbin issues found so far:
1. Can't seem to get format=original to work, downloads a file with part of command line whose contents says "error".

pdurbin:
> ThrowableHandler

This was recently changed in #7085

qqmyers:

kcondon:
@qqmyers Should it be a separate ticket?

qqmyers:
@kcondon -yeah - not related to this PR that I can see. Also - is this 406 error from a test? If so, it might also make sense to fix the test (unless we really wanted to test 406 instead of seeing if the dcterms metadata can be downloaded.

kcondon:
@qqmyers I'm not sure what's causing it. It appears to happen on it's own.

qqmyers:
FWIW - datasets/export can produce "application/xml", "application/json", or "application/html". My guess is that if someone makes an API call requesting something other than one of those, it's a 406.

kcondon:
@pdurbin This works but since it only ever seems to say "error" when something goes wrong, maybe that could be looked at?

poikilotherm:
@kcondon I extended the error handling stuff in #7136. Do you wish for me to add a more meaningfull error messages for error 406?

kcondon:
@poikilotherm Is there something more meaningful? If it doesn't add value, then no. If it provides context or potentially a direction for action, then yes. I trust your judgement. In the case for errors in this pr, it literally just says error, when there are clearly different causes and I am error prone when first using an API. :(

pdurbin:
@kcondon is right. Only "ERROR" is seen like this...

```
{
    "status": "ERROR"
}
```

... if a guest user tries to download from a dataset with multiple unpublished files.

Additionally, I can get this error (I'm using PIDs instead of database IDs)...

```
{
    "status": "ERROR",
    "code": 403,
    "message": "'/api/v1/access/dataset/%3ApersistentId' you are not authorized to access this object via this api endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org."
}
```

... if the guest user (no API token) attempts to use the new "download by dataset" endpoint on a dataset with a single restricted file.

Let me see what I can do.

pdurbin:
@kcondon et al, good catch on the error handling. I wasn't using the WrappedResponse object properly but as of fc7df593d I am. This means that instead of always sending BAD_REQUEST (400), you should see UNAUTHORIZED (401) as appropriate. You should also see more detailed messages like these:

- User :guest is not permitted to perform requested action.
- Bad api key 

I updated the tests to reflect these and others.

I didn't attempt to do anything with the "you are not authorized to access this object via this api endpoint" message because it's pre-existing and I'm trying to be mindful of scope. If I call into the older "datafiles" (plural) API on which this new "download by dataset" API is build, you get a similar message. Here's an attempt to use that API by the guest user to download a single restricted file:

curl http://localhost:8080/api/access/datafiles/4216

```
{
  "status": "ERROR",
  "code": 403,
  "message": "'/api/v1/access/datafiles/4216' you are not authorized to access this object via this api endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org."
}
```

Furthermore, this error at least conveys "not authorized" and gives FORBIDDEN (403) so I think it's ok.

I'm moving this back to code review. I'm aware that there's more conversation above about error handling as well as in #7134 and #7136 but I imagine it'll be handled in separate issues and pull requests.

PR #7100, opened 2020-07-17
make ec2-create-instance.sh downloadable #7099

### LINKED ISSUE #7099: ec2 script can't be downloaded from the docs
While reviewing pull request #7093 and testing it locally, I noticed a failure in develop (I'm on c9ec15862):

/Users/pdurbin/NetBeansProjects/dataverse/doc/sphinx-guides/source/developers/deployment.rst:89: WARNING: download file not readable: /Users/pdurbin/NetBeansProjects/dataverse/doc/sphinx-guides/source/developers/https:/raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/ec2/ec2-create-instance.sh

pdurbin:

Users can't download ec2-create-instance.sh without it.

Closes #7099

`:download:` only works on local files, which is why I'm changing it.

Build the docs. Make sure you can download the script.

No.

No.

**Additional documentation**:

None.

### DIFF
--- a/doc/sphinx-guides/source/developers/deployment.rst
+++ b/doc/sphinx-guides/source/developers/deployment.rst
@@ -82,7 +82,9 @@ Download and Run the "Create Instance" Script
 
 Once you have done the configuration above, you are ready to try running the "ec2-create-instance.sh" script to spin up Dataverse in AWS.
 
-Download :download:`ec2-create-instance.sh<https://raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/ec2/ec2-create-instance.sh>` and put it somewhere reasonable. For the purpose of these instructions we'll assume it's in the "Downloads" directory in your home directory.
+Download `ec2-create-instance.sh`_ and put it somewhere reasonable. For the purpose of these instructions we'll assume it's in the "Downloads" directory in your home directory.
+
+.. _ec2-create-instance.sh: https://raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/ec2/ec2-create-instance.sh
 
 To run it with default values you just need the script, but you may also want a current copy of the ansible `group vars <https://raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/defaults/main.yml>`_ file.
 


doc/sphinx-guides/source/developers/deployment.rst
```diff
@@ -82,7 +82,9 @@ Download and Run the "Create Instance" Script
 
 Once you have done the configuration above, you are ready to try running the "ec2-create-instance.sh" script to spin up Dataverse in AWS.
 
-Download :download:`ec2-create-instance.sh<https://raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/ec2/ec2-create-instance.sh>` and put it somewhere reasonable. For the purpose of these instructions we'll assume it's in the "Downloads" directory in your home directory.
+Download `ec2-create-instance.sh`_ and put it somewhere reasonable. For the purpose of these instructions we'll assume it's in the "Downloads" directory in your home directory.
```
mheppler, line 85: I suspect that rather than changing the format of the download reference, the warning could be resolved by adding a space to the original source code, between the link text and URL brackets. Like so, `:download:`ec2-create-instance.sh <https://...`.

pdurbin: @mheppler nope. I just tried adding a space and I still got `WARNING: download file not readable`.

I'm pretty sure `:download:` only works on local files. Please see https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#referencing-downloadable-files

Now, there is an inline syntax that also works. It would look like this (you're right that a space is needed in this case):

    Download `ec2-create-instance.sh <https://raw.githubusercontent.com/GlobalDataverseCommunityConsortium/dataverse-ansible/master/ec2/ec2-create-instance.sh>`_ and put it somewhere reasonable. For the purpose of these instructions we'll assume it's in the "Downloads" directory in your home directory.

I went with the non-inline syntax because the link is long. I find it more readable.

mheppler: Sorry, @pdurbin. Missed that it was not a local file. Changing it text link that opens an external page is totally kosher. 👍

PR #6060, opened 2019-07-26
2471 destroy command should unregister doi

sekmiller:
## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Thanks!

## Related Issues

- closes #2471 Destroy Command should unregister DOI

## Pull Request Checklist

- [x] Unit [tests][x] none
- [x] Integration [tests][]: None
- [x] Deployment requirements,  None
- [x] [Documentation][docs] None
- [x] Merged latest from "develop" [branch][x] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

Issue #2471, opened 2015-08-19
Destroy command should unregister dataset from DOI / persistent ID provider


scolapasta:

bencomp:
Could you please explain the term "unregister"? The PID provider usually has rules for this and may not even support taking a PID out of the registry.
To my knowledge the best way to take a PID out of the universe is to update it to redirect to a page (perhaps a generic one) explaining what happened to the dataset. This is similar to what happens when you deaccession a dataset.

pdurbin:
@scolapasta @sekmiller @kcondon do you know if this was fixed as part of pull request #3826?

djbrooke:
- we believe we have this in place, but the Datacite API may have not existed at the time. 
- we would want this to work for handles, and EZID-generated DOIs as well 
- we'd want this to work for files as well (this is implemented in the code as a try-catch already and we may want to refactor)

sekmiller:
Neither EZID nor DataCite allow for the deletion of PIDs once they have been made public. EZID allows reserved (non-published) PIDs to be deleted. The current implementation of DataCite does not register PIDs until publication.

So for EZID PIDs, when a dataset is destroyed its PID's status is set of "unavailable, withdrawn by author" and its target url is changed from the Dataverse dataset page to an EZID PID page where the "unavailable" status is displayed. 

For DataCite PIDs, on dataset destroy the status (and for any published files is set to "registered". A "registered" status removes the PID from any DataCite searches.  Also, the target url remains pointing at Dataverse so that anyone with the url will get a 404 - not found error.  (This was deemed appropriate as the dataset is truly gone and not just deaccessioned.)

Handles for destroyed datasets are fully deleted.

poikilotherm:
A quick note from our onboarding session with Datacite: they provide a tombstone mechanism, which should be preferred over a 404.

sekmiller:
Is there documentation of DataCite's tombstone mechanism?

landreev:
If the documentation for this is not readily google-able, we can always ask them. Martin Fenner has been very willing to answer any technical questions (he reads support@datacite.org). 

I don't have a strong opinion though on whether this is something we need - specifically, in the context of this issue, as part of the "Destroy dataset" command. This is kind of a nuclear option, reserved for special situations. And I know in the past some of the people who wanted the functionality were actually interested in erasing all traces of the dataset.

sekmiller:
@landreev the DeleteIdentifier method was already implemented in the HandlenetServiceBean.

PR #7101, opened 2020-07-17
restore code from #2471, changes to merge that with new pid api

### LINKED ISSUE #7078: Changes to DOI Updating upon Dataset deletion were reverted
As part of the Community call discussion last night, I looked into what happens to the DOI state for a dataset that is destroyed in Dataverse. It looks like the current develop branch will not do anything to the DOI (at least for DataCite) if the dataset being destroyed has been published. With pointers from @djbrooke, I realized that the changes in #6060 to address #2471 have been reverted by #6901 from @pdurbin . #2471/#6060 sets many of the metadata fields to :unav and switches 'findable' DataCite DOIs to registered to stop them from being indexed. If that's still the intended behavior, I think this code has to be restored/adapted to work with #6901 (perhaps this was temporarily dropped to simplify getting the DOI reservation work in place?).

qqmyers:
**What this PR does / why we need it**: restores PIDS being invalidated when a dataset is destroyed

Closes #7078

**Special notes for your reviewer**: Tried to make minimal changes. #6901 created a separate delete method in PidUtils which I was able to move back into the DataCiteServiecBean code so that the Pids delete api isn't DataCIte specific any more. I did leave the API as only delete PIDs for unpublished Datasets though (with the assumption that, for DataCite, they have draft DOIs). 

**Suggestions on how to test this**: The API should work as before (except delete would also work for EZID and Handles (to the extent that their delete code is correct). The change this PR should make is the same as for #6060 - publish a dataset, and destroy it via the API and you should see the DataCite DOI change from findable to registered and have it's metadata replaced with :unav entries. (4.20 should already work this way, the 5.0 dev branch no longer did).

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
@@ -16,7 +16,9 @@ import java.util.logging.Logger;
 import javax.ejb.EJB;
 import javax.ejb.Stateless;
 
+import org.apache.commons.httpclient.HttpException;
 import org.apache.commons.httpclient.HttpStatus;
+
 
 /**
  *
@@ -182,7 +184,7 @@ public class DOIDataCiteServiceBean extends AbstractGlobalIdServiceBean {
      * Deletes a DOI if it is in DRAFT/RESERVED state or removes metadata and changes it from PUBLIC/FINDABLE to REGISTERED.
      */
     @Override
-    public void deleteIdentifier(DvObject dvObject) throws Exception {
+    public void deleteIdentifier(DvObject dvObject) throws IOException, HttpException {
         logger.log(Level.FINE,"deleteIdentifier");
         String identifier = getIdentifier(dvObject);
         Map<String, String> doiMetadata = new HashMap<>();
@@ -198,14 +200,10 @@ public class DOIDataCiteServiceBean extends AbstractGlobalIdServiceBean {
                 case RESERVED:
                 case DRAFT:    
                     logger.log(Level.INFO, "Delete status is reserved..");
-                    try {
-                    	//service only removes the identifier from the cache (since it was written before DOIs could be registered in draft state)
-                        doiDataCiteRegisterService.deleteIdentifier(identifier);
-                        //So we call the deleteDraftIdentifier method below until things are refactored
-                        deleteDraftIdentifier(dvObject);
-                    } catch (Exception e) {
-                        logger.log(Level.WARNING, "delete failed: " + e.getMessage(), e);
-                    }
+                  	//service only removes the identifier from the cache (since it was written before DOIs could be registered in draft state)
+                    doiDataCiteRegisterService.deleteIdentifier(identifier);
+                    //So we call the deleteDraftIdentifier method below until things are refactored
+                    deleteDraftIdentifier(dvObject);
                     break;
 
                 case PUBLIC:
@@ -248,7 +246,8 @@ public class DOIDataCiteServiceBean extends AbstractGlobalIdServiceBean {
             connection.setRequestProperty("Authorization", basicAuth);
             int status = connection.getResponseCode();
             if(status!=HttpStatus.SC_NO_CONTENT) {
-            	throw new IOException(BundleUtil.getStringFromBundle("pids.deletePid.failureExpected", Arrays.asList(doi.asString(), Integer.toString(status))));
+            	logger.warning("Incorrect Response Status from DataCite: " + status + " : " + connection.getResponseMessage());
+            	throw new HttpException("Status: " + status);
             }
             logger.fine("deleteDoi status for " + doi.asString() + ": " + status);
     }
--- a/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
@@ -187,14 +187,11 @@ public class DOIDataCiteServiceBean extends AbstractGlobalIdServiceBean {
     public void deleteIdentifier(DvObject dvObject) throws IOException, HttpException {
         logger.log(Level.FINE,"deleteIdentifier");
         String identifier = getIdentifier(dvObject);
-        Map<String, String> doiMetadata = new HashMap<>();
-        try {
-            doiMetadata = doiDataCiteRegisterService.getMetadata(identifier);
-        } catch (Exception e) {
-            logger.log(Level.WARNING, "deleteIdentifier: get matadata failed. " + e.getMessage(), e);
-        }
-
-        String idStatus = doiMetadata.get("_status");
+        //ToDo - PidUtils currently has a DataCite API call that would get the status at DataCite for this identifier - that could be more accurate than assuming based on whether the dvObject has been published
+        String idStatus = DRAFT;
+        if(dvObject.isReleased()) {
+        	idStatus = PUBLIC;
+        } 
         if ( idStatus != null ) {
             switch ( idStatus ) {
                 case RESERVED:
--- a/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
@@ -1,8 +1,13 @@
 package edu.harvard.iq.dataverse;
 
-import edu.harvard.iq.dataverse.pidproviders.PidUtil;
+import edu.harvard.iq.dataverse.util.BundleUtil;
+
 import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -10,6 +15,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.ejb.EJB;
 import javax.ejb.Stateless;
+
+import org.apache.commons.httpclient.HttpStatus;
 
 /**
  *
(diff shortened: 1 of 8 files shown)


src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
```diff
@@ -251,15 +251,6 @@ public boolean isDoiInstallation() {
         }
     }
 
-    public boolean isDataCiteInstallation() {
```
pdurbin, line 254: There's something weird going on. I expected isDataCiteInstallation to be in use by dataset.xhtml but it's not, in either this branch or the develop branch. It was added in 28f3222 to grey out the publish button if a DOI hasn't been registered. But now this logic is... gone? Do we still want it? Flagging @landreev .

src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
```diff
@@ -171,33 +178,79 @@ public void deleteRecordFromCache(Dataset datasetIn){
...
+        GlobalId doi = dvObject.getGlobalId();
+        /**
+         * Deletes the DOI from DataCite if it can. Returns 204 if PID was deleted
+         * (only possible for "draft" DOIs), 405 (method not allowed) if the DOI
+         * wasn't deleted (because it's in "findable" state, for example, 404 if the
+         * DOI wasn't found, and possibly other status codes such as 500 if DataCite
+         * is down.
+         */
 
-    private boolean updateIdentifierStatus(DvObject dvObject, String statusIn) {
```
pdurbin, line 188: @qqmyers I don't think we talked about this `updateIdentifierStatus` method. Was it simply unused?

src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
```diff
@@ -171,33 +178,79 @@ public void deleteRecordFromCache(Dataset datasetIn){
...
+        }
+    }
+        
     /**
      * Deletes DOI from the DataCite side, if possible. Only "draft" DOIs can be
      * deleted.
      */
-    @Override
-    public void deleteIdentifier(DvObject dvObject) throws IOException {
+    private void deleteDraftIdentifier(DvObject dvObject) throws IOException {
```
pdurbin, line 225: What happens if this deleteDraftIdentifier method is called on a non-DataCite installation (Handle or EZID)?

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidUtil.java
```diff
@@ -83,26 +83,4 @@ private static String acceptOnlyDoi(String persistentId) {
...
     }
-
-    /**
-     * Deletes the DOI from DataCite if it can. Returns 204 if PID was deleted
-     * (only possible for "draft" DOIs), 405 (method not allowed) if the DOI
-     * wasn't deleted (because it's in "findable" state, for example, 404 if the
-     * DOI wasn't found, and possibly other status codes such as 500 if DataCite
-     * is down.
-     */
-    public static int deleteDoi(String persistentId, String baseUrl, String username, String password) throws IOException {
```
pdurbin, line 94: The test in PidUtilTest that calls this method should be deleted:

```
-------------------------------------------------------------
COMPILATION ERROR : 
-------------------------------------------------------------
edu/harvard/iq/dataverse/pidproviders/PidUtilTest.java:[35,29] cannot find symbol
  symbol:   method deleteDoi(java.lang.String,java.lang.String,java.lang.String,java.lang.String)
  location: class edu.harvard.iq.dataverse.pidproviders.PidUtil
1 error
```

See also https://travis-ci.org/github/IQSS/dataverse/builds/710110797

src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
qqmyers: It's private in the DataCite code. THe code here could be put in the public deleteIdentifier method - just decided to keep it separate since it will hopefully become part of the doiDataCiteRegisterService.deleteIdentifier(identifier); call on line 204 at some point.

src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
qqmyers: I only deleted this method because I couldn't find it being used. If it is, we can restore.

src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
pdurbin: Ah, right, I forgot we're in the DataCite-specific bean. Thanks.

src/main/java/edu/harvard/iq/dataverse/DOIDataCiteServiceBean.java
qqmyers: Yeah - looks like this private method exists in the EZID and Handle code, called by the public publicizeIdentifier() method but the DataCite code just have the same code directly in the publicizeIdentifier method.

src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
pdurbin: @qqmyers you're right that it isn't being used but that's only because something got lost on the xhtml side. Please restore it. For more details, please see #7102.

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidUtil.java
pdurbin: Fixed.

src/main/java/edu/harvard/iq/dataverse/api/Pids.java
```diff
@@ -104,6 +104,12 @@ public Response reservePid(@PathParam("id") String idSupplied) {
     public Response deletePid(@PathParam("id") String idSupplied) {
         try {
             Dataset dataset = findDatasetOrDie(idSupplied);
+            //Restrict to never-published datasets (that should have draft/nonpublic pids). The underlying code will invalidate
+            //pids that have been made public by a pid-specific method, but it's not clear that invalidating such a pid via an api that doesn't
+            //destroy the dataset is a good idea.
+            if(dataset.isReleased()) {
+            	return badRequest("Not allowed for Datasets that have been published.");
```
pdurbin, line 111: This "isReleased" check is making me question if we should remove this "delete PID" API endpoint altogether (as well as the DeletePidCommand and docs, of course).

I added it because I was looking at DataCite APIs from a CRUD perspective. How can I create a PID, read it, update it, delete it. I was learning how to do these things with curl and I'm aware that sometimes sysadmins have a need to operate on the DataCite API directly to fix certain things. My thought was that we could make it easier by giving them APIs in Dataverse to operate on DataCite, rather than re-documenting the DataCite API or simply sending them to those docs.

The other thing I was thinking when adding this Delete PID API is that I wanted a way to blank out `GlobalIdCreateTime` and `IdentifierRegistered` so I could test the "register PID" API (which I still think is a useful, necessary API). This is much more of a dev concern than prod, however.

scolapasta: I would agree that having an API just for dev or test should either be removed or moved to a different area of the API (I think we have admin/test? we could also add admin/dev?)

qqmyers: Moving this, or all of these calls to admin, and then potentially removing the !isReleased() restriction could make sense - if anyone has been destroying datasets with earlier versions of Dataverse, a way to sync up with current practice might be helpful.

qqmyers: is there a best practice about only having isSuperUser() restrictions for admin api calls? (I seem to remember that argument from something I did, but not sure if it was a general practice.)

scolapasta: they don't have isSuperUser() in the code, because they don't require a token - things like indexing, etc. They are protected by security policy in that they can only be run from localhost. (they also don't use commands, since commands require user / permission checks).

qqmyers: There are actually a bunch in Admin.java (some from me) but maybe they're not needed. My question was more in the other direction and taking a look, I see there are lots of non-admin api calls that have a superuser restriction. (If there weren't it might add to the argument that the Pids api should be admin.)

scolapasta: I guess I should have said - api calls in admin *should be* for things that don't require a user, and are protected by policy. There are definitely a lot of non admin calls that do have a superuser restriction. Things like Destroy. We want superusers, who don't have localhost access, to be able to call these.

pdurbin: As I mentioned to @qqmyers today, I found the "delete PID" API useful today in development. Also, it's used in the testCannotPublishUntilReserved test in PidsIT.

kcondon:
@qqmyers @scolapasta @pdurbin I guess I could use some help with this. What am I testing, how do I test it, what are the correct results. Cannot test EZID any longer so will accept it as a risk. Handles, not easily tested either so will focus on continued, normal operation of Datacite. I can use Fabrica to see the draft status before publish(thanks Jim) but both the older datacite api and phils pid info api shows it as 404, both before publish and after delete.  Also, Phil, wasn't sure what you were asking me -was a pid api useful?

qqmyers:
The main thing this PR does is to restore functionality for updating the DataCite DOI when you destroy a Dataset. You should just be able to look in Fabrica to see the DataCite side: publish Dataset, make sure you can see the DOI, it has metadata, and it's in the 'findable' state. Then destroy the dataset via the API and verify that the DataCite DOI is in the 'registered' state, that the metadata (such as authors) say :unav, there's no title, pubyear is 9999, etc. 

The questions about the PID api/whether they're needed/useful are really a separable issue - the change to it in this PR is a ~no-op - the original pids delete DOI api only called the delete endpoint (so it would remove a draft DOI, but not change a findable one) and the code here just adds a check to make sure that the api only works on a draft DOI and then calls a more general delete method that would work on both draft (deleting it) and findable (making it registered/blanking the metata) DOIs. I guess the thing that could be tested is to verify that the delete call only works on a draft DOI and doesn't change/thows an error if you use it on a findable DOI/the DOI from a published dataset.

PR #7103, opened 2020-07-21
Deadlock fix and index cleanup #6865

### LINKED ISSUE #6865: Internal Exception: org.postgresql.util.PSQLException: ERROR: deadlock detected
Sometimes, but not all the time, when I run the API test suite on my laptop, some of the API tests fail with a 500 response and when I look at server.log I see errors like this:

`Internal Exception: org.postgresql.util.PSQLException: ERROR: deadlock detected`

Here's an example of a failing test but the tests vary:

[ERROR]   FilesIT.testReplaceFileBadJson:898 Expected status code <200> doesn't match actual status code <500>.

Here are a couple more examples from last week, both from the same run:

[ERROR] FileMetadataIT.tearDownDataverse:96 Expected status code <200> doesn't match actual status code <500>.

[ERROR] FilesIT.testReplaceFileBadJson:898 Expected status code <200> doesn't match actual status code <500>.

In tech hours today we discussed how these local failures are probably related to failures we've seen on https://jenkins.dataverse.org

Here's my server.log (renamed so GitHub will take it) from today's run: 
[server.log.txt](https://github.com/IQSS/dataverse/files/4548577/server.log.txt)

scolapasta:
It cleans up some indexing code - it extracts the index call that was in the DataverseServiceBean save into separate helper methods that can be called my commands in their onSuccess methods; it also removes the indexing of role related permissions during execute (since the dataverse is indexed in the onSuccess), which appeared to be the cause of the deadlocks we were getting in the tests

Closes #6865 

The Commands that call the save are listed here and whether I added on success call. Please review these decisions carefully, in case I missed something.

CreateDataverseCommand - already had an onSuccess index call - so it was already doubling up
PublishDataverseCommand - already had a call to index the permissions on success
RenameDataverseCommand- added a call to re index object on success
UpdateDataverseCommand - already had a call to index the object and all child datasets (unsure why it needed, but kept same logic it had)
UpdatePermissionRootCommand - added a call to re index object and permissions on success (honestly wasn't sure when this command is needed so went the safe root)

MoveDataverseCommand - indexing is already happening in execute - should be moved to onSuccess, but out of scope for this change

UpdateDataverseDefaultContributorRoleCommand - doesn't need indexing 
UpdateDataverseGuestbookCommand - doesn't need indexing
UpdateDataverseGuestbookRootCommand - doesn't need indexing
UpdateDataverseTemplateRootCommand - doesn't need indexing
UpdateDataverseThemeCommand - doesn't need indexing

The straightforward thing to do would be test where each of the above commands are used, as they are the ones that will be affected by these changes. In particular that roles permissions are indexed correctly for CreateDataverseCommand.

no

no

**Additional documentation**:


scolapasta:
In trying to merge with debug branch, i added some commits I didn't want here, so am closing and restarting with a fresh PR.

PR #7105, opened 2020-07-21
6865 no more deadlocks

### LINKED ISSUE #6865: Internal Exception: org.postgresql.util.PSQLException: ERROR: deadlock detected

scolapasta:
It cleans up some indexing code - it extracts the index call that was in the DataverseServiceBean save into separate helper methods that can be called my commands in their onSuccess methods; it also removes the indexing of role related permissions during execute (since the dataverse is indexed in the onSuccess), which appeared to be the cause of the deadlocks we were getting in the tests

Closes #6865 

The Commands that call the save are listed here and whether I added on success call. Please review these decisions carefully, in case I missed something.

CreateDataverseCommand - already had an onSuccess index call - so it was already doubling up
PublishDataverseCommand - already had a call to index the permissions on success
RenameDataverseCommand- added a call to re index object on success
UpdateDataverseCommand - already had a call to index the object and all child datasets (unsure why it needed, but kept same logic it had)
UpdatePermissionRootCommand - added a call to re index object and permissions on success (honestly wasn't sure when this command is needed so went the safe root)

MoveDataverseCommand - indexing is already happening in execute - should be moved to onSuccess, but out of scope for this change

UpdateDataverseDefaultContributorRoleCommand - doesn't need indexing 
UpdateDataverseGuestbookCommand - doesn't need indexing
UpdateDataverseGuestbookRootCommand - doesn't need indexing
UpdateDataverseTemplateRootCommand - doesn't need indexing
UpdateDataverseThemeCommand - doesn't need indexing

The straightforward thing to do would be test where each of the above commands are used, as they are the ones that will be affected by these changes. In particular that roles permissions are indexed correctly for CreateDataverseCommand.

no

no

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
@@ -16,6 +16,7 @@ import edu.harvard.iq.dataverse.batch.util.LoggingUtil;
 import edu.harvard.iq.dataverse.dataaccess.ImageThumbConverter;
 import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
 import edu.harvard.iq.dataverse.search.IndexServiceBean;
+import edu.harvard.iq.dataverse.search.SolrIndexServiceBean;
 import edu.harvard.iq.dataverse.search.SolrSearchResult;
 import edu.harvard.iq.dataverse.util.SystemConfig;
 import java.io.File;
@@ -35,14 +36,11 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.json.Json;
 import javax.json.JsonArrayBuilder;
-import javax.json.JsonObject;
-import javax.json.JsonObjectBuilder;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
 import javax.persistence.NonUniqueResultException;
 import javax.persistence.PersistenceContext;
 import javax.persistence.TypedQuery;
-import javax.ws.rs.core.Response;
 import org.apache.solr.client.solrj.SolrServerException;
 
 /**
@@ -56,6 +54,9 @@ public class DataverseServiceBean implements java.io.Serializable {
     private static final Logger logger = Logger.getLogger(DataverseServiceBean.class.getCanonicalName());
     @EJB
     IndexServiceBean indexService;
+    
+    @EJB
+    SolrIndexServiceBean solrIndexService; 
 
     @EJB
     AuthenticationServiceBean authService;
@@ -91,21 +92,33 @@ public class DataverseServiceBean implements java.io.Serializable {
        
         dataverse.setModificationTime(new Timestamp(new Date().getTime()));
         Dataverse savedDataverse = em.merge(dataverse);
+        return savedDataverse;
+    }
+    
+    public boolean index(Dataverse dataverse) {
+        return index(dataverse, false);
+
+    }
+        
+    public boolean index(Dataverse dataverse, boolean indexPermissions) {    
         /**
          * @todo check the result to see if indexing was successful or not
          * added logging of exceptions 
          */
         try {
-            Future<String> indexingResult = indexService.indexDataverse(savedDataverse);
+            indexService.indexDataverse(dataverse);
+            if (indexPermissions) {
+                solrIndexService.indexPermissionsOnSelfAndChildren(dataverse);
+            }
         } catch (IOException | SolrServerException e) {
-            String failureLogText = "Post-save indexing failed. You can kickoff a re-index of this dataverse with: \r\n curl http://localhost:8080/api/admin/index/dataverses/" + savedDataverse.getId().toString();
+            String failureLogText = "Post-save indexing failed. You can kickoff a re-index of this dataverse with: \r\n curl http://localhost:8080/api/admin/index/dataverses/" + dataverse.getId().toString();
             failureLogText += "\r\n" + e.getLocalizedMessage();
-            LoggingUtil.writeOnSuccessFailureLog(null, failureLogText, savedDataverse);
+            LoggingUtil.writeOnSuccessFailureLog(null, failureLogText, dataverse);
+            return false;
         }
 
-//        logger.log(Level.INFO, "during dataverse save, indexing result was: {0}", indexingResult);
-        return savedDataverse;
-    }
+        return true;
+    }    
 
     public Dataverse find(Object pk) {
         return em.find(Dataverse.class, pk);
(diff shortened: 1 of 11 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDataverseCommand.java
```diff
@@ -168,17 +169,8 @@ public Dataverse execute(CommandContext ctxt) throws CommandException {
...
-        } catch (IOException | SolrServerException e){           
-            Dataverse dv = (Dataverse) r;            
-            String failureLogText = "Indexing failed. You can kickoff a re-index of this dataverse with: \r\n curl http://localhost:8080/api/admin/index/dataverses/" + dv.getId().toString();
-            failureLogText += "\r\n" + e.getLocalizedMessage();
-            LoggingUtil.writeOnSuccessFailureLog(this, failureLogText,  dv);
-            return false;
-        }
-        return true;
+    public boolean onSuccess(CommandContext ctxt, Object r) {  
+        return ctxt.dataverses().index((Dataverse) r);
```
sekmiller, line 173: The index dataverse here does not index permissions. Should you be using index((Dataverse) r, true)?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseCommand.java
```diff
@@ -90,20 +90,19 @@ public Dataverse execute(CommandContext ctxt) throws CommandException {
...
-
-        List<Dataset> datasets = ctxt.datasets().findByOwnerId(result.getId());
+        
+        // first kick of async index of datasets
+        // TODO: is this actually needed? Is there a better way to handle
         try {
-            Future<String> indResponse = ctxt.index().indexDataverse(result);
+            Dataverse result = (Dataverse) r;
+            List<Dataset> datasets = ctxt.datasets().findByOwnerId(result.getId());
             ctxt.index().asyncIndexDatasetList(datasets, true);
```
sekmiller, line 99: reindexing datasets may not actually be needed here. (based on the instances where it is called ManageTemplates, Guestbook, etc.)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdatePermissionRootCommand.java
```diff
@@ -37,6 +37,13 @@ public Dataverse execute( final CommandContext ctxt) throws CommandException {
 		}
 	}
 
+    @Override
+    public boolean onSuccess(CommandContext ctxt, Object r) {  
+        return ctxt.dataverses().index((Dataverse) r,true);
+    }        
+
+
+    //TODO: Review this as this will never be an instance of Dataset, will it?
```
sekmiller, line 46: This will never be a dataset so we can simplify the code here

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -91,21 +92,33 @@ public Dataverse save(Dataverse dataverse) {
...
+            indexService.indexDataverse(dataverse);
+            if (indexPermissions) {
+                solrIndexService.indexPermissionsOnSelfAndChildren(dataverse);
+            }
         } catch (IOException | SolrServerException e) {
-            String failureLogText = "Post-save indexing failed. You can kickoff a re-index of this dataverse with: \r\n curl http://localhost:8080/api/admin/index/dataverses/" + savedDataverse.getId().toString();
+            String failureLogText = "Post-save indexing failed. You can kickoff a re-index of this dataverse with: \r\n curl http://localhost:8080/api/admin/index/dataverses/" + dataverse.getId().toString();
             failureLogText += "\r\n" + e.getLocalizedMessage();
-            LoggingUtil.writeOnSuccessFailureLog(null, failureLogText, savedDataverse);
+            LoggingUtil.writeOnSuccessFailureLog(null, failureLogText, dataverse);
```
sekmiller, line 116: Kind of a nit, but there are cases where this isn't being called by onSuccess so the failure message so maybe add to the general log in those cases.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDataverseCommand.java
scolapasta: Actually, the index without the boolean (of false) does index the call indePermissionForOneDVObject. 

indexPermissions or (index with true) adds a call to indexPermissionOnSelfAndChildren, but that shouldn't be needed in the case of creating a dataverse.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseCommand.java
scolapasta: I think you're right, but I wanted to change as little as possible while surgically fixing the deadlocks. So left this code as it was, but added the TODO.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdatePermissionRootCommand.java
scolapasta: Similar to last, left as todo and out of scope.

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
scolapasta: This is a brand new method and all the calls I added are in onSuccess methods. Additionally as a tech design decision, we've said we want all such indexing to be onSuccess, no? So I think it's ok.

Note, that currently not all indexing is actually on success (see MoveDataverse), but changing that was out of scope for this issue).

PR #7107, opened 2020-07-22
5467 Return useful error message for DDI upload.

### LINKED ISSUE #5467: As a depositor, I want to know why the API to create datasets using DDI xml failed
When I use the API to import a dataset with a DDI xml file (see [section in API guide](http://guides.dataverse.org/en/latest/api/native-api.html#import-a-dataset-into-a-dataverse-with-a-ddi-file)) and the file isn't an xml file or (more likely to happen) if the xml isn't well-formed I get a message that doesn't tell me why the import failed.

The error looks like the html of the 500 server error page.
screen shot 2019-01-14 at 4 31 05 pm

JingMa87:
**What this PR does / why we need it**: Makes the error message more useful when uploading an invalid DDI XML

**Which issue(s) this PR closes**: 5467

Closes #5467 

**Special notes for your reviewer**: In Dataverses.java I removed the exception from the method signature and handled it in a try catch, where I return a useful message. In ImportDDIServiceBean I added a try catch with a specific error message for the `<codebook>` element. In ImportServiceBean I removed the message "XMLStreamException" which isn't useful.

**Suggestions on how to test this**: Upload a pdf or invalid DDI XML and you will see a useful error message. If you upload a valid DDI XML starting from the `<codebook>` element, you will get a success message. If you upload a DDI XML where you remove the slashes from the identifier in the <IDNo> elements, you will get the invalid DOI error (make sure to do it for all occurrences of the identifier). I added the command and screenshots below. The success message is just to verify all is still working.

COMMAND
```
curl -H X-Dataverse-key:xxxxxxxxxxxx -X POST http://localhost:8080/api/dataverses/root/datasets/:importddi --upload-file test.xml
```

SUCCESS
image

FAIL
image

BAD DOI
image

**Is there a release notes update needed for this change?**: It's not a big change so no

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
@@ -102,7 +102,7 @@ import java.text.MessageFormat;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.Optional;
-import javax.persistence.NoResultException;
+import javax.xml.stream.XMLStreamException;
 
 /**
  * A REST API for dataverses.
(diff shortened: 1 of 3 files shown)


TaniaSchlatter:
Please add what the messages that are displayed are (Success, Fail, Bad DOI)

JingMa87:
@TaniaSchlatter You can find the messages in the screenshots in the description.

TaniaSchlatter:
@JingMa87 Right - API only. Thanks!

PR #6098, opened 2019-08-16
4933 remove subset

mheppler:
## Related Issues

- closes #4933 Data subset: Downloading data subset is not working for basic file in 4.9.2 

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [x] [Documentation][docs] completed
- [x] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #7110, opened 2020-07-22
7079 guides download all and analytics

### LINKED ISSUE #7079: Access Dataset + Download "All Files" - Update Guides and Analytics
With the new dataset level download option under the new Access Dataset btn (#7047) the User Guide should be reviewed and updated to outline this new feature.

There should also be updates to the analytics event tracking sample code provided in the Installation Guide, as the download format and size included in the dropdown option text is sent to the Google Analytics reports. That text should be stripped in order to keep the category and values consistent for reporting purposes. Inclusion of varying file sizes in the reports will prohibit those events from being grouped and properly counted.

Screen Shot 2020-07-01 at 3 47 34 PM

Screen Shot 2020-07-14 at 9 38 09 AM

mheppler:

- Adds reference to new Access Dataset btn and download all files feature to the User Guide.
- Adds function to strip download size of ZIP from the link text sent to analytics reports when tracking clicks of said new download all files links under Access Dataset btn, to the sample code provided in the Installation Guide.

Closes #7079 Access Dataset + Download "All Files" - Update Guides and Analytics

N/A

Need to configure analytics, using new sample code provided in Installation Guide, should see "Original Format ZIP" and "Archival Format (.tab) ZIP" recorded in the analytics reports.

N/A

Need release notes notifying sysadmins to update their analytics code if configured already, so the file size "(5.1 MB)" is not included in the Event Category recorded in the Google Analytics reports.

**Additional documentation**:

N/A


doc/sphinx-guides/source/user/find-use-data.rst
```diff
@@ -117,9 +119,6 @@ Ingested files can be downloaded in several different ways.
 
 - The variable metadata for the file in DDI format
 
-- A subset of the columns of the data
```
pdurbin, line 120: Ah, I assume we forgot to remove this line about subset as part of #6098.

kcondon:
@mheppler Quick question, should https://raw.githubusercontent.com/IQSS/dataverse/2ac69d7503d12a1013530683db4dbcbb9b8bb969/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html be downloadable from the guides? I see analytics-code.html highlighted but not downloadable.

PR #7111, opened 2020-07-22
adding support for changing sessionId at login

### LINKED ISSUE #3254: Sign Up Url - Session ID
This is part of a series of issues raised by NTU libraries as discussed with @pdurbin

Refer to other issue - #3253 

The jsession ID that is generated when the user clicks on "Sign Up" url. The appscan report says that the session ID generated before the user log in can be modified or altered by an middle man. So the session ID must be generated after the sign up is completed. This was flagged as Medium vulnerability.

When testing the url with and without the jsessionid the effect is the same. Is this session ID in anyway required for the sign up. Can this be disabled?

/dataverseuser.xhtml?editMode=CREATE&redirectPage=%2Fdataverseuser.xhtml

/dataverseuser.xhtml;jsessionid=2aaf45f07480a9153e53bffbd921?editMode=CREATE&redirectPage=%2Fdataverse.xhtml

qqmyers:
**What this PR does / why we need it**: Changes the sessionId at login

Closes #3254 

**Special notes for your reviewer**: The code changes here work for the standard login (e.g. as dataverseAdmin) but haven't been checked with other login types. @landreev was going to look for other places where the same sessionId change might be needed for alternate login methods. (FWIW: It does look like this login method calls multiple auth providers, so it may cover all or at least several.) I put the code to change the Id in a util class to make that easier. Nominally this PR doesn't fully close #3254 until it covers a all the login mechanisms. (edit: I was indeed wondering earlier whether this needed to be applied in several places; similar to how we call configureSessionTimeout() from all the different auth provider impls.; but that does not appear to be necessary here: when increasing the session timeout, we wanted to be sure we were doing it only once the user has successfully authenticated/logged in;  in this case, we are perfectly fine with issuing a new session before we try to authenticate - L.A.)

**Suggestions on how to test this**: Open the dev console and watch the jsessionId - cookie (usually) or in the URL -when you succeed in logging in, it should change, and the login should succeed as normal (i.e. no changes except to the session id)
Please test with different auth providers - Shib, etc. - in addition to the builtin provider. 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseSession.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseSession.java
@@ -6,6 +6,7 @@ import edu.harvard.iq.dataverse.actionlogging.ActionLogRecord;
 import edu.harvard.iq.dataverse.actionlogging.ActionLogServiceBean;
 import edu.harvard.iq.dataverse.authorization.users.GuestUser;
 import edu.harvard.iq.dataverse.authorization.users.User;
+import edu.harvard.iq.dataverse.util.SessionUtil;
 import edu.harvard.iq.dataverse.util.SystemConfig;
 import java.io.IOException;
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


kcondon:
This works for built in -the session id changes upon log in, but doesn't appear to change when using oauth (orcid). Will also test shib after initial investigation.

PR #7116, opened 2020-07-22
global update of doi:mdcbaseurlstring and doi.baseurlstringnext

### LINKED ISSUE #7113: doi.mdcbaseurlstring and doi.baseurlstringnext are the same thing
The URL for the DataCite REST api (i.e. api.test.datacite.org or api.datacite.org in production) is currently (in develop) recorded via two jvm options (those in the title). The former (mdc...) was in 4.20 and is configured in Ansible. I'm not sure either name is great, but using the former would be backward compatible. If we can decide what single name to use, it should be easy to change the relevant code before we release 5.0

qqmyers:
to doi.dataciterestapiurlstring

**What this PR does / why we need it**: Consolidates the existing jvm option for configuring Make Data Count queries to DataCite and the new one for the /pids API/reserving DOIs into a single option with a neutral (not-mdc-centric) name

Closes #7113 

**Special notes for your reviewer**: If there are issues with the name, this PR may need a few global search/replaces, but I hope I've caught all the places a change is needed. 

**Suggestions on how to test this**: Set the jvm option incorrectly and verify that the /pids?persistentId=<your DOI> and the /admin/makeDataCount/{id}/updateCitationsForDataset API calls fail (showing that both are picking up the new jvm option) and then set it to https://api.test.datacite.org to verify that these calls work. To test the defaults, you can make sure old doi.mdcbaseurlstring jvm option is set. The changes to the install scripts should result in this new jvm option being set instead of the old doi.mdcbaseurlstring option.

**Is there a release notes update needed for this change?**: yes - we're changing the name of/replacing a jvm option defined since 4.18

**Additional documentation**:


qqmyers:
Heads-up: @donsizemore , @poikilotherm may want to watch this for the effect on ansible/k8 installs.

src/main/java/edu/harvard/iq/dataverse/api/Pids.java
```diff
@@ -45,7 +45,7 @@ public Response getPid(@QueryParam("persistentId") String persistentId) {
         } catch (WrappedResponse ex) {
             return error(Response.Status.FORBIDDEN, BundleUtil.getStringFromBundle("api.errors.invalidApiToken"));
         }
-        String baseUrl = System.getProperty("doi.baseurlstringnext");
+        String baseUrl = systemConfig.getDataCiteRestApiUrlString();
```
sekmiller, line 48: So this API endpoint only supports DataCite?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeletePidCommand.java
```diff
@@ -47,7 +47,7 @@ protected void executeImpl(CommandContext ctxt) throws CommandException {
         GlobalIdServiceBean idServiceBean = GlobalIdServiceBean.getBean(protocol, ctxt);
         try {
             // idServiceBean.deleteIdentifier(dataset); // didn't work
-            String baseUrl = System.getProperty("doi.baseurlstringnext");
+            String baseUrl = ctxt.systemConfig().getDataCiteRestApiUrlString();
```
sekmiller, line 50: Again, surprised that this command only supports DataCite, but it was like this previously.

src/main/java/edu/harvard/iq/dataverse/api/Pids.java
qqmyers: Good catch. Yes, and I think the docs say that. However, the other places that were DataCite-specific had been resolved in earlier PRs that were merged into dev. I just resolved the conflicts in this PR and you can see DoiDataCiteServiceBean now gets the new URL via the SystemConfig method and the DeletePIDCommand is no longer DataCite specific.

sekmiller: Great. Will pass to QA.

PR #7120, opened 2020-07-23
7117 metadata field type uri_no

### LINKED ISSUE #7117: Allow more types (schemes) of URLs (URIs) in metadata fields
**Disclaimer:** I'm very aware of #6030 but this can't wait for us. If IQSS is unhappy with this, it can reside in our mini-fork.

**tl;dr**: Add a new metadata field `uri_no` representing absolute, non-opaque URIs, being the correct term for "URL". Using `url` type is not sufficient as HTTP/S only.

**Context**:
For Jülich DATA, we want our contributors to provide URLs to data or at least documentation of whereabouts, when they don't or can't upload the data. (Which is our major use case...)

Lots of our data references will reside on windows network shares (`smb://foo.bar/share/folder`) and other obscure places (think `rsync://...`, `ipfs://...`, `s3://`, `gpfs://...`, `git+xxx://`, `http://`, `ftp://...`, ...). Thus we need a broader support for likely any kind of URL to come no matter if a browser understands it.

Our use case is also described at https://jugit.fz-juelich.de/fdm/schemas/-/issues/2 and will be documented in depth in our guide.

**Technical Background**:
Please keep in mind that a ["URL" (uniform resource locator)](https://en.wikipedia.org/wiki/URL) is only a colloquial term. It's a common practice to use it, but strictly speaking, URLs are a subset of URIs (uniform resource identifiers), defined in [RFC 3986](https://tools.ietf.org/html/rfc3986).

URIs become URLs by adding an "authority" part - usually meaning a network resource - which is anything after the schema (e.g. `http:`), like `//dataverse.org` and before the "path", like `/login.xhtml`. 

URIs without an authority are "opaque" (leaving out some other cases), URLs are always non-opaque.
Some good examples for commonly known opaque URIs: `tel:+13930303`, `isbn:1292-92219-1212`, `mailto:john.doe@foo.bar`.
That could be even more formalised into URN (uniform resource name): `urn:isbn:1292-92219-1212`

In Java both concepts exist as `java.net.URL` and `java.net.URI`. The key difference is that a URL object in Java always has to be backed by a scheme handler, as the API promises you can open a stream for it.

**Problem**:
Currently, when using a `url` typed metadata field in a (custom) metadata block, this will only support URLs with `http, https, file` and `jar` scheme. (See topic protocol handlers in [Java Docs](https://docs.oracle.com/javase/8/docs/api/java/net/URL.html#URL-java.lang.String-java.lang.String-int-java.lang.String-))

Also, the current field type `url` might be used to implicitly rely on being HTTP/S only. There are a lot of places in upstream metadata blocks where placeholder tell people they should provide a "full URL, starting with http://".

image
image

**Suggestion**:
Instead of changing the current `url` type and any logic beyond that, I propose to add a new type `uri_no`, meaning "non-opaque URI" being an alias for URLs. This will exclude URNs and URIs without the authority part, leaving any kind of URL as allowed usage for the field.

This is a surprisingly small change, PR forthcoming. It hasn't much UI impact, it's more on the metadata side of things.

poikilotherm:

This PR adds a new metadata field type `uri_no`, providing support for any kind of URL instead of HTTP/S only. It's really small and more about the why can be found in the issue (which is again very small)

Closes #7117

Nada.

Deploy with this PR applied. Create a metadata schema with the new field type in use. Deploy schema, activate in a Dataverse. Create a Dataset and try to insert any kind of URL.

No. Or at least only very very minimal, as URL support existed before, but was limited to HTTP/S.

I don't think so. Maybe. Don't know. It's a small thing.

**Additional documentation**:

Nada.

### DIFF
--- a/doc/sphinx-guides/source/admin/metadatacustomization.rst
+++ b/doc/sphinx-guides/source/admin/metadatacustomization.rst
@@ -197,6 +197,7 @@ Each of the three main sections own sets of properties:
 |                       | is meant to contain.  | | \• text              |
 |                       |                       | | \• textbox           |
 |                       |                       | | \• url               |
+|                       |                       | | \• uri_no            |
 |                       |                       | | \• int               |
 |                       |                       | | \• float             |
 |                       |                       | | \• See below for     |
@@ -418,7 +419,18 @@ FieldType definitions
 |                                   | Management page in the User Guide.|
 +-----------------------------------+-----------------------------------+
 | url                               | If not empty, field must contain  |
-|                                   | a valid URL.                      |
+|                                   | a valid URL, restricted to schemes|
+|                                   | ``http``, ``https``,              |
+|                                   | ``file``, ``jar``. (A limitation  |
+|                                   | of `Java URL class`_.)            |
++-----------------------------------+-----------------------------------+
+| uri_no                            | If not empty, field must contain  |
+|                                   | a valid, non-opaque, absolute URI.|
+|                                   | (Meaning an URL.)                 |
+|                                   | See also :rfc:`3986`,             |
+|                                   | `Good practices for URIs`_,       |
+|                                   | `W3C Axioms`_,                    |
+|                                   | `W3C URI Clarification`_.         |
 +-----------------------------------+-----------------------------------+
 | int                               | An integer value destined for a   |
 |                                   | numeric field.                    |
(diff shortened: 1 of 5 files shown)


djbrooke:
Hey @poikilotherm, thanks for the PR here. I'm going to assign myself and also @jggautier to this issue so we can catch up next week. I understand that you need this and that it's small but I'm hesitant to add this to the application with the flexible metadata work and TRSA work underway. If we change this we'd need to potentially do some mapping/migration when we add a more robust solution.

jggautier:
@djbrooke and I haven't discussed this yet but I just wanted to write before more time passes that I agree with @poikilotherm that there's merit to the existing "url" fieldtype, which restricts possible metadata values to http:// and https://. For all of the metadata fields that use this fieldtype, including the "Alternative URL" field, acceptable values should start with http:// and https://

It sounds like the only reason to create a new fieldtype that accepts all types of URLs is to support the need to publish the metadata of datasets whose data files are only in some storage outside of the Dataverse installation's file server(s):

- If so, then I think the main thing to consider is how this might conflicts or could exist with the TRSA plans, like @djbrooke wrote. 
- If not, then what other reasons would Dataverse need a new fieldtype for all types of URLs?

poikilotherm:
Thank you @djbrooke and @jggautier for discussing this in our call today.
As you folks don't see Dataverse as a metadata registry only (which is one of our main use cases), I'm going to close this pull request. The code will live on in our fork. If anyone is interested for a use case, feel welcome to reach out.

Issue #6980, opened 2020-06-11
Reserving Datacite DOIs asynchronously


landreev:
This may have slipped through the cracks in the original PR #6901: We need either one of the following: An automated way to reserve Datacite dois for draft datasets for which the identifiers have not been reserved yet, after the fact, i.e. asynchronously. Or a release note explaining that this needs to be done.
This is for the dois that for one reason or another have failed to get reserved on create (starting v5). And, most importantly, for any draft datasets in Datacite-using Dataverses that have been created before v5. Since starting v5 they will not be publishable, until reserved. 

We have discussed having an automatic timer job for this. And it would not be hard to implement. But it is not necessary, strictly speaking - since we have the API to list the unreserved DOIs, and the API to reserve them. So it is something that an admin could run once in a while, or script together a cron job to do. However, if we are not providing a timer, we at least need to add a release note explaining that any Datacite-using installation needs to do this for their existing drafts once v5 is deployed. 

Apologies if I'm missing something and this is in fact explained somewhere.

djbrooke:
Additional context from now-closed-as-duplicate #7019:

In #5093/#6901 we added support for reserving DOIs with Datacite.

There may be some steps to get DOIs into a reserved state, and we want to make sure that all drafts remain publishable. We also don't want Dataverse to have old drafts marked as reserved when they are not actually reserved.
There should be a step about running an API to mark drafts as reserved (whether a single time or on a regular cron)

djbrooke:
@scolapasta Thanks for discussing and estimating - is the plan to handle this through a release note or through code?

scolapasta:
Release note.

landreev:
Another thing we can recommend - and should definitely do in our prod. - would be to add a few extra words to the warning banner about the upcoming upgrade; something along the lines of "This Dataverse will be down between ... and ... for a planned upgrade. Please note that users may experience temporary problems publishing previously unpublished drafts immediately after the upgrade". 
We can keep a similar warning up after the upgrade as well, until all the unpublished DOIs are reserved.

landreev:
I've tested the case where we have (in v4.20) an unpublished draft with the null GlobalIdCreateTime, that DOES exist on the Datacite side - what happens when we run the new reserve API on it? 
[...]

PR #7121, opened 2020-07-23
add message when DataCite DOI has not been reserved #7102

### LINKED ISSUE #7102: DOI registration is not confirmed by Datacite
Now that pull request #6901 has been merged, a DOI is reserved from DataCite when a dataset is created.

However, what if something goes wrong? What if the DOI is not reserved on the DataCite side? In this case we want to prevent publishing.

As part of the pull request above, the following commit implemented this:

- 28f3222 grey out publish button if DataCite and no PID create time

However, the code on the xhtml side was removed in a flurry of merge conflict resolution (where render logic was refactored).

#6980 is related in that the button will stay greyed out until someone does something to reserve the DOI.

pdurbin:

Here's the problem scenario:

- DataCite was down when the dataset was created, meaning that the DOI was never reserved on the DataCite side.
- When the user tries to publish, they get a somewhat confusing error: "Cannot publish dataset because its persistent identifier has not been reserved." (screenshot below).

Screen Shot 2020-07-23 at 11 25 56 AM

To help explain what's going on, this pull request add a message to the top of the dataset page:

Screen Shot 2020-07-23 at 11 17 42 AM

Closes #7102

- ~~I copied and pasted the text that was given to me but I would prefer to use the word "reserved" rather than "registered" because we already use the former throughout our documentation and code.~~ Fixed in fee17726d.
- ~~I used "Dataset Locked" as the short message before the long explanation but we can certainly change this to whatever we want.~~ Changed to "Dataset DOI Not Reserved" in fee17726d.
- I made use of the the "displayLockInfo" infrastructure even though this message isn't a true DatasetLock.

The message should only appear for DataCite installations. In addition, the message should only appear for datasets with a null value for GlobalIdCreateTime (dvobject table), which is what we use to indicate if a DOI is reserved. It's the time the PID was created on the remote system.

Yes, please see screenshots, above.

No, I think it falls under the umbrella of Dataverse 5 allowing DOIs to be reserved.

**Additional documentation**:

None.

### DIFF
--- a/src/main/java/propertyFiles/Bundle.properties
+++ b/src/main/java/propertyFiles/Bundle.properties
@@ -1322,8 +1322,8 @@ dataset.locked.ingest.message=The tabular data files uploaded are being processe
 dataset.unlocked.ingest.message=The tabular files have been ingested.
 dataset.locked.editInProgress.message=Edit In Progress
 dataset.locked.editInProgress.message.details=Additional edits cannot be made at this time. Contact {0} if this status persists.
-dataset.locked.pidNotReserved.message=Dataset Locked
-dataset.locked.pidNotReserved.message.details=The DOI displayed in the citation for this dataset has not yet been registered with DataCite. Please do not share this DOI until it has been registered.
+dataset.locked.pidNotReserved.message=Dataset DOI Not Reserved
+dataset.locked.pidNotReserved.message.details=The DOI displayed in the citation for this dataset has not yet been reserved with DataCite. Please do not share this DOI until it has been reserved.
 dataset.publish.error=This dataset may not be published due to an error when contacting the <a href=\{1} target=\"_blank\"/> {0} </a> Service. Please try again.
 dataset.publish.error.doi=This dataset may not be published because the DOI update failed.
 dataset.publish.file.validation.error.message=Failed to Publish Dataset
--- a/src/main/java/propertyFiles/Bundle.properties
+++ b/src/main/java/propertyFiles/Bundle.properties
@@ -1322,6 +1322,8 @@ dataset.locked.ingest.message=The tabular data files uploaded are being processe
 dataset.unlocked.ingest.message=The tabular files have been ingested.
 dataset.locked.editInProgress.message=Edit In Progress
 dataset.locked.editInProgress.message.details=Additional edits cannot be made at this time. Contact {0} if this status persists.
+dataset.locked.pidNotReserved.message=Dataset Locked
+dataset.locked.pidNotReserved.message.details=The DOI displayed in the citation for this dataset has not yet been registered with DataCite. Please do not share this DOI until it has been registered.
 dataset.publish.error=This dataset may not be published due to an error when contacting the <a href=\{1} target=\"_blank\"/> {0} </a> Service. Please try again.
 dataset.publish.error.doi=This dataset may not be published because the DOI update failed.
 dataset.publish.file.validation.error.message=Failed to Publish Dataset
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -2128,6 +2128,15 @@ public class DatasetPage implements java.io.Serializable {
                     BundleUtil.getStringFromBundle("dataset.locked.ingest.message"));
             lockedDueToIngestVar = true;
         }
+
+        // With DataCite, we try to reserve the DOI when the dataset is created. Sometimes this
+        // fails because DataCite is down. We show the message below to set expectations that the
+        // "Publish" button won't work until the DOI has been reserved using the "Reserve PID" API.
+        if (settingsWrapper.isDataCiteInstallation() && dataset.getGlobalIdCreateTime() == null) {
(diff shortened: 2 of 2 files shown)


src/main/java/propertyFiles/Bundle.properties
```diff
@@ -1322,6 +1322,8 @@ dataset.locked.ingest.message=The tabular data files uploaded are being processe
 dataset.unlocked.ingest.message=The tabular files have been ingested.
 dataset.locked.editInProgress.message=Edit In Progress
 dataset.locked.editInProgress.message.details=Additional edits cannot be made at this time. Contact {0} if this status persists.
+dataset.locked.pidNotReserved.message=Dataset Locked
+dataset.locked.pidNotReserved.message.details=The DOI displayed in the citation for this dataset has not yet been registered with DataCite. Please do not share this DOI until it has been registered.
```
mheppler, line 1326: I know we don't want to keep going back and forth on this, but... this msg reads like we need additional success and warning issues if the DOI was finally registered or ultimately changed. This came to mind reading the comment from @landreev in Slack in regards to issue #6980.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -1322,6 +1322,8 @@ dataset.locked.ingest.message=The tabular data files uploaded are being processe
 dataset.unlocked.ingest.message=The tabular files have been ingested.
 dataset.locked.editInProgress.message=Edit In Progress
 dataset.locked.editInProgress.message.details=Additional edits cannot be made at this time. Contact {0} if this status persists.
+dataset.locked.pidNotReserved.message=Dataset Locked
```
mheppler, line 1325: I don't think we need to add another "Dataset Locked" bundle value here. There is already `dataset.locked.message`.

pdurbin:
In design standup I got some guidance from @TaniaSchlatter and @djbrooke on wording.

src/main/java/propertyFiles/Bundle.properties
pdurbin: I put the key in as a placeholder but just changed it to "Dataset DOI Not Reserved" in fee17726d.

src/main/java/propertyFiles/Bundle.properties
pdurbin: I don't feel strongly about the wording and don't mind if someone changes it.

djbrooke:
Wording as of https://github.com/IQSS/dataverse/commit/fee17726d996145ad16c1110cc34139a97c1d945 looks good, thanks @pdurbin

kcondon:
@pdurbin I'm seeing the doi not reserved lock banner on the create dataset page. should it be there? a little confusing imho.

pdurbin:
No, it shouldn't be on the create page.

kcondon:
Well you are checking on a null global create time, right? create page has null too. maybe needs to be excluded?

pdurbin:
Yeah, there's probably some kind of MODE!=CREATE we can check.

kcondon:
Also, getting something weird happening:
1. make incorrect datacite url in domain.xml
2. create and attempt to publish dataset, fails to reserve, check.
3. fix url, restart try to publish, persists in saying could not reserve. It should now be able to but it doesn't. Why not?
log says:
```

[2020-07-29T19:48:00.263+0000] [Payara 5.201] [SEVERE] [] [edu.harvard.iq.dataverse.DatasetPage] [tid: _ThreadID=90 _ThreadName=http-thread-pool::jk-connector(3)] [timeMillis: 1596052080263] [levelValue: 1000] [[
  Cannot publish dataset because its persistent identifier has not been reserved.]]
```

pdurbin:
In 1c1ae2c0e I just added the logic to not show the "DOI not reserved" message on create.

For your other scenario, the plan reserve DOIs regularly and there's a new script in pull request #7142 for this. Basically, if DataCite is down, we simply can't reserve DOIs during that period. So maybe the script runs nightly or so to reserve them.

pdurbin:
@kcondon yes, thanks for the discussion. I have very little practical experience with EZID and have only glanced at that code, so I'll take your word for it as to how it behaved.

It looks like there's a new issue and pull request where the discussion can continue:

- issue #7146: Backward compatibility/error recovery for PID reservation during publish 
- pull request #7147: Reserve PID at publication if needed

(Or we can keep talking here if that's better.)

kcondon:
@pdurbin Fair enough, though your old officemate Stephen created the EZID implementation and several others on the team were familiar with that functionality and I'd thought the original discussion about implementing DataCite's version was to copy what EZID did and that was my expectation going into this. I really just want a clear idea of what things do and that are logical so I have that for now and hopefully future discussions will proceed on that basis. Thanks.

Fernand0S:
Hello, i have problem publications article in Dataverse. The configurations /usr/local/glassfish4/glassfish/domain/domain1/config/domain.xml 

<jvm-options>-Ddoi.username=</jvm-options> and <jvm-options>-Ddoi.password=</jvm-options> for use of Handle is not Doi, modified strings but error persistent.


image

pdurbin:
@Fernand0S hi, please email support@dataverse.org about your problem to open a support ticket. Thanks.

Fernand0S:
Philip, I solved problem. reply you with email.

PR #7124, opened 2020-07-23
tsv ingest fix for directupload

### LINKED ISSUE #7122: ingest of tsv fails in direct upload
Trying direct upload of .tsv file via direct upload to s3 failed to trigger ingest. Investigating showed two issues:

- My browser did not assign a mimetype to .tsv files during upload so it was being processed as application-octet-stream. Adding code to at least detect the mimetype via fileExtension (a partial solution to 6762) helped, but assigned the mimetype text/tab-separated-values to the file.
- Looking into that it appears that the changes in #6517 to reserve the text/tab-separated-values type for ingested .tab files did not catch the assignment of tsv to that mimetype in META-INF/mime.types. Changing the assignment there to use text/tsv as the mimetype allows the .tsv file to be ingested.

PR to follow.

qqmyers:
**What this PR does / why we need it**: enables ingest of .tsv files (and potentially other types) with direct upload when the browser does not supply a mimetype. The changes here invoke code to determine the mimetype from the extension and then update the mapping of tsv files to the currently required text/tsv mimetype.

Closes #7122

**Suggestions on how to test this**: setup for direct upload and try uploading a small tsv - it should ingest to produce a .tab file. Also turn on fine logging for the FileUtil class and look for "Determined type: text/tsv".  - If you see this, it indicates that you are in the test case where your browser doesn't supply a mimetype for .tsv files. If you don't see this, you may need to remove any mapping for the .tsv extension from your test browser/machine (or I can slack/demo against the test server from my browser that doesn't provide a mimetype).

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1091,8 +1091,13 @@ public static String generateOriginalExtension(String fileType) {
...
 			} 
 		} else {
-			//Remote file, trust supplier
-			finalType = suppliedContentType;
+			if(suppliedContentType==FileUtil.MIME_TYPE_UNDETERMINED_DEFAULT) {
+				finalType=determineFileTypeByExtension(fileName);
+				logger.fine("Determined type: " + finalType);
+			} else {
+			  //Remote file, trust supplier
+			  finalType = suppliedContentType;
```
sekmiller, line 1099: Do we care if the suppliedContentType is blank or null?

qqmyers: That should already be being caught (e.g. in the EditDatafilesPage.handleExternalUpload method) but worth a safety check here probably. And we shouldn't replace the default with a null/blank response if determineFileTypeByExtension returns null/blank.

sekmiller: If you make that change, I'll put this into QA. Thanks!

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1126,7 +1106,42 @@ public static String generateOriginalExtension(String fileType) {
     }   // end createDataFiles
     
 
-    private static File saveInputStreamInTempFile(InputStream inputStream, Long fileSizeLimit)
+	private static boolean useRecognizedType(String suppliedContentType, String recognizedType) {
```
qqmyers, line 1109: just moved and made to return a boolean, no logic changes.

PR #7045, opened 2020-07-01
#6823 remove form update on file upload

sekmiller:
**What this PR does / why we need it**:
The page refresh that happens when a file is uploaded to a dataset in Create mode is annoying to users. This fix will leave the focus on the file upload section of the page

**Which issue(s) this PR closes**:

Closes #6823 Add Dataset after file upload browser jumps to top of page

**Special notes for your reviewer**:

**Suggestions on how to test this**:
Upload files via native on Dataset Create. Also run under dual mode with dcm and upload under native on DS Create,

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

Issue #6823, opened 2020-04-15
Add Dataset - After upload file complete, browser jumps to top of page


jggautier:
Seen on Demo Dataverse and Harvard Dataverse running Dataverse version 4.20, on the newest versions of Chrome and Firefox:

On the dataset create page, the file upload section is at the bottom of the page. When uploading a file in a new dataset, once the file is uploaded, the dataset create page jumps to the top or near the top. Depositors have to scroll back down to click the Save Dataset button.

Is this jumping behavior by design? Since the file upload section is at the bottom of the page, isn't file uploading the last step before saving the draft dataset? Or is jumping to the top encouraging people to review the metadata they've entered while scrolling back down to save the dataset?

djbrooke:
I don't think this is by design. :) That's weird. IMHO we should fix. Will move this to the board.

donsizemore:
FWIW I'm seeing this on Payara5 as well (primefaces8 branch just now)

mheppler:
Took a peek into this, and I believe the reloading of the entire form at Line 285 of editFilesFragment.xhtml is causing the page to jump.

```
<p:commandButton id="AllUploadsFinished" 
    action="#{EditDatafilesPage.uploadFinished()}"
    update="@form,datasetForm:fileUpload,datasetForm:dropBoxUserButton,datasetForm:uploadMessage,datasetForm:rsyncPanel,datasetForm:filesCounts,datasetForm:filesTable" 
    oncomplete="javascript:bind_bsui_components();javascript:uploadWidgetDropMsg();" 
    style="display:none"/>
```

That `@form` in the hidden button update was added by @sekmiller in 8dffc46bbdc9e8782efcdb82c8c974411ca39385 as part of the dual mode work in issue #4610. It looks to be a pretty convoluted series of elements listed in all those various hidden commandButton update attributes, so I'm just gonna leave this in Up Next and ask Stephen to take a peek in more detail when he has a chance.

PR #7126, opened 2020-07-23
post 6823 edit file form update fixes

### LINKED ISSUE #7123: Page update issues after #6823
The solution for #6823 - the page jumping to the top of the dataset creation form involved removing the update of @form.  However, direct uploads and the switch of the 'Done' button in the edit datafiles form to the 'Save/Cancel' options relied on that update. A fix for the direct uploads just involves updating more of the editFilesFragment.html form. However,  to resolve the problem of the 'Done' button not updating involves updating an element that only exists in the editdatafiles.xhtml page and not in the dataset.xhtml form (i.e. when uploading files in create mode.) I was able to fix that by adding an empty element to the dataset page so that the id exists for all uses of the editFilesFragment form.

I'll post a PR. It would be useful to have some feedback on whether the two elements I picked are the best ones to update and whether there's a better option than adding an empty element to help resolve the second problem.

qqmyers:
**What this PR does / why we need it**: fixes bugs introduced in by #7045

Closes #7123

**Special notes for your reviewer**: I picked the contentOfHttpPanel to fix the direct upload issue (when the select files widget is updated, the directupload script that is outside that (above in the code) needs to run again and wasn't. Any other element outside the script could work. For the fix of the button updates, I changes the name of the element from filesButtons to completionButtons primarily because an empty element with that name is now in the dataset.xhtml but the name could stay as it was.(I originally put the empty panel around the button bar in the dataset form (in which case a name change would be more helpful) but I updating those buttons (in dataset create mode) isn't needed.)

**Suggestions on how to test this**: Current production - using direct upload and uploading a file leaves the 'select new files' box with upload and cancel buttons showing. The button at the bottom of the page remains as 'Done' rather than updating to 'Save/Cancel' . (Same behavior after the first upload to a file store, but the 'Done' button updates to Save/Cancel after a second upload of additional file(s)). With the fix, direct-upload and normal upload to s3 or files (testing just one of these two should be OK) in dataset create and edit mode should always work - direct upload shows a normal select files box after a single file is uploaded and the 'Done' button changes in edit for all cases after one upload, and the dataset create form still doesn't scroll to the top of the page (just keeps the select files box in view)).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: only restores missing functionality from prior to #7045.

**Additional documentation**:

### DIFF
--- a/src/main/webapp/editFilesFragment.xhtml
+++ b/src/main/webapp/editFilesFragment.xhtml
@@ -132,7 +132,7 @@
                                   disabled="#{lockedFromEdits ||  !(datasetPage || EditDatafilesPage.showFileUploadComponent()) }"
                                   listener="#{EditDatafilesPage.handleFileUpload}" 
                                   process="filesTable" 
-                                  update=":datasetForm:filesTable, @([id$=filesButtons])"
+                                  update=":datasetForm:filesTable, @([id$=completionButtons])"
                                   label="#{bundle['file.selectToAddBtn']}"
                                   oncomplete="javascript:bind_bsui_components();uploadFinished(PF('fileUploadWidget'));"
                                   onstart="javascript:uploadWidgetDropRemoveMsg();uploadStarted();"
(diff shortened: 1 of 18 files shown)


src/main/webapp/editFilesFragment.xhtml
```diff
@@ -282,7 +282,7 @@
...
                     <p:commandButton id="updateEditDataFilesButtonsForDelete" action="#{EditDatafilesPage.deleteFilesCompleted()}" update="datasetForm:editDataFilesButtons" rendered="#{showFileButtonUpdate}" style="display:none"/>
-                    <p:commandButton id="AllUploadsFinished" action="#{EditDatafilesPage.uploadFinished()}" update="datasetForm:fileUpload,datasetForm:dropBoxUserButton,datasetForm:uploadMessage,datasetForm:rsyncPanel,datasetForm:filesCounts,datasetForm:filesTable" oncomplete="javascript:bind_bsui_components();javascript:uploadWidgetDropMsg();" style="display:none"/> 
+                    <p:commandButton id="AllUploadsFinished" action="#{EditDatafilesPage.uploadFinished()}" update="datasetForm:completionButtons,datasetForm:contentOfHttpPanel,datasetForm:fileUpload,datasetForm:dropBoxUserButton,datasetForm:uploadMessage,datasetForm:rsyncPanel,datasetForm:filesCounts,datasetForm:filesTable" oncomplete="javascript:bind_bsui_components();javascript:uploadWidgetDropMsg();" style="display:none"/>
```
mheppler, line 285: Does the addition of the contentOfHttpPanel to the update here make the fileUpload component redundant?

src/main/webapp/filesFragment.xhtml
```diff
@@ -117,7 +117,7 @@
                     </div>
                 </div>
                 <div class="col-xs-7 text-right">
-                    <p:outputPanel id="filesButtons">
+                    <p:outputPanel id="completionButtons">
```
mheppler, line 120: Isn't this a duplicate ID of the new completionButtons outputPanel added to the dataset?

qqmyers: ~yes, but the editFileFragment is either embedded in the dataset.xhtml or filesFragement.xhtml, but not both, so I think there is only ever one element with this id. Having them the same is really the point - as the AllUploadsFinished command tries to update 'completionButtons' and fails if it can't find it, which is what happened when I tried this without having the empty element in dataset.xhtml with this name. If there's  a way to update an element only if it exists, that should work instead.

src/main/webapp/editFilesFragment.xhtml
qqmyers: maybe? It was there when @form was in it too, so I wasn't sure - if it was redundant then too then maybe it can be removed. Some of the other elements could also be redundant - I didn't check.

qqmyers:
@sekmiller - adding the buttons to the update was needed to fix the issue of updating the button from 'Done' to Save/Cancel - they're outside the contentOfHttpPanel. Since they only existed in filesFragment and didn't exist when you're in dataset create mode, I also had to add a panel with the same name in dataset.xhtml. The name change itself is gratuitious and could be reverted - it doesn't hurt to have an empty panel in dataset.xhtml with id filesButtons (just thought that was an odd name when I originally put the panel around the dataset create button panel (which doesn't have anything to do with filesButtons). So - if it helps to change the name back, it would be easy to do.

sekmiller:
@qqmyers I thought that using " update=":datasetForm:filesTable, @([id$=filesButtons])" meant that it didn't matter if the id occurred in the dataset page in create mode, and it seemed to work fine in 6823, unless it was something that I missed due to direct upload. Either way I think it might be helpful to change the name back to filesButtons to help avoid merge conflicts as long as it's easy. Thanks!

qqmyers:
@sekmiller - I'm confused - there was no @([id$=filesButtons]) before or after #6823? Before it, the @form refreshed the buttons without specifying their id (but it also refreshed the whole dataset creation page moving you back to the top). If @([id$=filesButtons]) will refresh the buttons only if that id exists, then that would mean there's no need for that element in dataset.xhtml and I can remove that as well. Is that the case?

sekmiller:
I believe so, yes.

qqmyers:
@sekmiller - I see that the @([id$=filesButtons]) was/is used in the file input's update method, so per your suggesting, I've just used that in the command as well. Thanks!

PR #7136, opened 2020-07-28
7134 refactor errorhandlers

### LINKED ISSUE #7134: Error Handling: Multiple server.log errors indicating http 406 Not Acceptable
Seeing this in the logs, without tying it to a specific user action, though I am testing the download all api:

[2020-07-22T08:52:30.214+0000] [Payara 5.201] [SEVERE] [] [edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler] [tid: _ThreadID=92 _ThreadName=http-thread-pool::jk-connector(5)] [timeMillis: 1595407950214] [levelValue: 1000] [[
Uncaught REST API exception:
Incident: 39be6b43-cbbf-4dc9-a19a-484f413872a3
URL: https://dataverse-internal.iq.harvard.edu/api/v1/datasets/export?exporter=dcterms&persistentId=doi%3A10.70122%2FFK2%2FEVQ2OH
Method: GET
javax.ws.rs.NotAcceptableException: HTTP 406 Not Acceptable

@qqmyers Has made the following comments:

poikilotherm:

- This refactors the API error handlers to be more relaxed about non-special-treatment Jersey exceptions.
- It also removes boilerplate and duplicated code from the package.
- It takes care of a standardized response format for errors (which should be adapted in other places to be consistent throughout our beloved API).
- It streamlines the logging of exceptions by enabling debugging and proper parsing in log analysis.

Closes #7134 

I do hope the scope is not to big. I did not change the remaining two exception handlers for JSON and constraints to keep it smaller. Happy to change those if you think that makes sense.

One might also argue that special messages for Error 403 and 405 are not of much use. Happy to delete those.

I could also switch the Jersey Error 500 message to be the same as in ThrowableHandler.

The 302 and 307 were added because sometimes you might want to see a proper message via API, for example when you don't auto-follow redirects with `curl` et al.

I'm not sure if we should add a comment about the logging to the docs.

After deployment:
- Try debugging verbosity: `asadmin set-log-levels edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler=FINEST` and go for a 404 error
- Not sure how to generate a 500 error...
- A 404, 405 and 406 are easy to test
- A 302/307 is not commonly used in Dataverse. The only place is for ZIP downloads. So to test this, you'll have to enable those. I created a unit test for this during development, but deleted it as it wasn't of that much use...
- No idea how to trigger a 503 error in Jersey...

Ah API isn't classic UI, so let's say no.

Nope.

**Additional documentation**:

None.


qqmyers:
@poikilotherm - FWIW: redirects happen when you use direct downloads with S3. They are a normal part of operation so it seems odd to respond to the user with any form of incident id, etc.

poikilotherm:
@qqmyers I could create a simple filter that removes the JSON entity for such cases like 302/307 redirects, although it should be configurable. When you try to debug things with `curl`, it might be beneficial to see a more verbose output. On the other hand: there are the logs... I see valid scenarios for both cases.

More people post their opinions where to go from here?

qqmyers:
The underlying issue (w.r.t. 5.0) here is that something that should be a 406 response is getting returned as a 500 error. 

This PR, as is, will change the error responses for many of the API calls - in a good way/to give more info to humans/standardize a json error body to work with, but that seems like a significant API change that could impact external tools (hopefully not - I know the DVUploader mostly deals with the response status and doesn't do much more than print any message, and with no standardization across the API, perhaps other tools haven't done much to parse the error messages either). I don't know how big a deal this is in practice but if people are concerned, it might not be something to push into 5.0.

Conversely, the underlying refactoring, to catch all WebApplicationExceptions before they get to the ThrowableHandler could avoid errors we haven't yet discovered (other error codes that shouldn't be converted to 500 errors). (And removes a lot of duplicate code - also good.) 

How much work would it be to avoid changing the external API responses for now / split the PR into a 5.0 fix and a post-5.0 API change? Is that a possibility?

poikilotherm:
OK let's look at the details of responses:

Feature | Before PR | With PR
---------- | ------------- | ------------
All handlers except `RedirectionException` return JSON | :+1:  | :+1: 
All handlers have proper error code in Response + HTTP header | :no_entry: | :+1: 
`status == ERROR` | :+1: | :+1: 
`code == XXX`, XXX = `int` | :+1: | :+1: 
Responses with `500` have JSON field `incidentId` | :+1: | :+1: 
`message` field has some special messages for some cases like 403 etc | :+1: | 🤏🏼 (same logic, but now without URI included)

So the most important fields like `status`, `code` and `incidentId` stayed stable. (It's debatable if it makes sense to extend the usage of `incidentId`, but that doesn't make it unstable.)

The new fields for `method`, `url`, `errorType` are debatable, but extensions, thus not making things unstable. Happy to filter those to be included in cases of HTTP status codes > 400. (< 400 is not really an error)

I'm very confident no one will rely on parsing a message to machine actionable. If so, 5.0 is a major release allowing for breaking changes.

I'm with you in terms of this needs polishing to be perfect, but I wouldn't consider it unstable. And even if - it's a release with breaking changes anyway. Happy to add stuff to API docs if you think this is missing. I also agree that streamlining the API responses becomes more important. There are a lot of places that could use a dust-off in the API code in terms of responses...

qqmyers:
Thanks for the detail - less concerning given that the changes are mostly to add fields instead of changing them (minor note: I'm not sure things like 406 were json prior to ThrowableHandler though, but that's not a run-time issue - if you get that you fix your code).
I think it would be worthwhile to filter for <400 but it might be better to wait for other opinions on that. 

As for docs, I haven't checked what's there now but it would probably be good to have info about the standardized response format you're implementing.

qqmyers:
@poikilotherm - had some discussion - can you go ahead and filter for <400 and document the standard json format for responses to finish this off?

poikilotherm:
Sure. But not today :smile: It's late over here :wink:

poikilotherm:
@qqmyers I just pushed my latest code change. I created a new factory/decorator pattern based response formatter, much like the `javax.ws.rs.core.Response.ResponseBuilder`, but enhanced with fancy JSON. (I didn't want to inherit from the abstract base class for a complete custom ResponseBuilder, which would be a lot overhead for this.)

If you think this is feasible, I would provide developer guide updates on this. There are a million places inside the API code where things would need to be changed to this new builder to streamline things, but this is definitely beyond scope for this issue.

src/main/java/edu/harvard/iq/dataverse/api/util/JSONResponseBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class JSONResponseBuilder {
```
qqmyers, line 18: Since this is only for errors, could/should it be JSONErrorResponseBuilder?

src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+                if ( (ex.getMessage()+"").toLowerCase().startsWith("tabular data required")) {
+                    jrb.message(BundleUtil.getStringFromBundle("access.api.exception.metadata.not.available.for.nontabular.file"));
+                } else {
+                    jrb.message("Bad Request. The API request cannot be completed with the parameters supplied. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.");
+                    jrb.request(request);
+                }
+                break;
+            // Forbidden
+            case 403:
+                jrb.message("Not authorized to access this object via this API endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.");
```
qqmyers, line 56: per my review comment of changing the AbstractApiBean.error method - would it make sense to move these status-specific messages to the JSON builder? That way they'd apply to calls from error() as well versus only happening on the exception route.

src/main/java/edu/harvard/iq/dataverse/api/util/JSONResponseBuilder.java
poikilotherm: I would like to keep this open minded. For now its errors only, but it would be great not only to streamline error responses, but success stories, too. Like seen in `ok()`, ... at https://github.com/IQSS/dataverse/blob/352b17f09beac5e0a9f44eb8a06cae7b5f6d2905/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L652-L747

src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
poikilotherm: Yes, that makes sense as far as I can foresee now.

But for the moment I would leave them there until it becomes clearer how the API response stuff will be done in the future.
(If you ask me, the errors should be handled completely by throwing JAX-RS or POJO exceptions, so there is a single place to run to and look at, avoiding to many places where stuff is wrapped in JSON sauce.)

poikilotherm:
@qqmyers If you feel like it, you might take a look at the freshly baked docs. I would love feedback on readability and comprehensibility...

poikilotherm:
Alright @pdurbin @qqmyers what do you think? Is this ready for QA?

qqmyers:
Not sure what procedure is, but I think the code itself is ready for QA - my comments were only about documentation which could be handled in parallel.

djbrooke:
Thanks @qqmyers for checking. Let's take the time to get this into good shape before moving it on to QA. @poikilotherm I'll assign back to you.

poikilotherm:
Guys, it's exhausting writing about this. Can we please schedule a call for this and talk about what we want, where to go from here and what is within scope of this issue and what's beyond? ("Small steps", to quote @pdurbin)

From the git log, I'd say we ping @pdurbin, @sekmiller, @michbarsinai, @landreev, @matthew-a-dunlap and @qqmyers for a schedule.
image

doc/sphinx-guides/source/developers/api.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Easy to follow code examples and explanations:
+
+- https://www.baeldung.com/jersey-rest-api-with-spring#using-exceptionmapper (never mind this is about Spring, the technology is the same!)
+- https://www.baeldung.com/jersey-bean-validation#custom-exception-handler
+
+For Dataverse, all API exception handlers live in package `edu.harvard.iq.dataverse.api.errorhandlers`_.
+*Remember, this does not handle exceptions popping via web UI, thus not served by Jersey JAX-RS!*
+
+1. ``ThrowableHandler`` catches any Java exception thrown and uncatched in business logic (either by accident or on purpose)
```
donsizemore, line 103: really minor tweak: `uncatched` might better be expressed as `uncaught` or `not caught`

scolapasta:
> Guys, it's exhausting writing about this. Can we please schedule a call for this and talk about what we want, where to go from here and what is within scope of this issue and what's beyond? ("Small steps", to quote @pdurbin)
> 
> From the git log, I'd say we ping @pdurbin, @sekmiller, @michbarsinai, @landreev, @matthew-a-dunlap and @qqmyers for a schedule.

@poikilotherm We discussed briefly at tech hours - in order to get this into Dataverse 5, could you separate this into two PRs: one the code, which he can review now and merge; the other, the discussion of developer guidance, which is out of scope for Dataverse 5. Thanks.

poikilotherm:
I close this in favor of the new code PR #7170 and new docs/refactoring issue #7171 and #7172

PR #7142, opened 2020-07-29
6980 datacite reservation upgrade

### LINKED ISSUE #6980: Reserving Datacite DOIs asynchronously

landreev:

 This is a documentation-only PR, the upgrade instructions for installations using Datacite - the DOIs for any unpublished drafts will need to be reserved. 
See the discussion in the issue #6980 for more information. 

Closes #6980 

The good news is that we don't need to do anything special for any DOIs that are "unpublished" in the Dataverse, but somehow already exist on the Datacite side. It's not even necessary to check if any such cases exist before running the batch reserve job as described in the instruction. 

Can be tested on any system running the develop branch (in its current state), and configured with a Datacite test account (not "fake"!). If it has any datasets created before the "reserve DOIs" PR was merged, they will be reported by the "unreserved" API, and they can be reserved as described in the instruction. If there are no such legacy unreserved DOIs, the condition can be emulated by setting the GlobalIdCreateTime to null in the dvobject table for some draft dataset(s). 

**Additional documentation**:


doc/release-notes/6980-datacite-reservation-upgrade
```diff
@@ -0,0 +1,28 @@
+# Dataverse installatins using Datacite: upgrade action required.
+
+For Dataverses using Datacite v5.0 changes the process of registering the Global Identifier (DOI) for the dataset. Instead of registering it when the dataset is published for the first time, Dataverse will try to "reserve" the DOI when it's created (by registering it as a "draft", using Datacite terminology). When the user publishes the dataset, the DOI will be publicized as well (by switching the registration status to "findable").
```
poikilotherm, line 3:
```suggestion
For installations using Dataverse v5.x, the process of registering the Global Identifier (DOI) for the dataset changes. Instead of registering it when the dataset is published for the first time, Dataverse will try to "reserve" the DOI when it's created (by registering it as a "draft", using Datacite terminology). When the user publishes the dataset, the DOI will be publicized as well (by switching the registration status to "findable"). 
```

poikilotherm: And is this really true for all types of PIDs? I couldn't tell from a quick view at the code, but if not, maybe this should be more Datacite-using specific?

landreev: Hmm. I can change "dataverses" to "installations" if you prefer; but otherwise your proposed commit is changing what  I am actually trying to say, somewhat. We want it to say "dataverses (or installations) **using Datacite**" - because this is Datacite-specific, and I do want to repeat it in this paragraph. That sentence may need a comma - but I don't think it should go where you are  adding it. I meant to say something along the lines of 
"For (Dataverse) installations (that are) using Datacite, (the upgrade to) v5.0 will change the process of ..." 

May not have been my best sentence ever, I'll try improve it.  But not sure I like your version either, sorry.

poikilotherm: That's fine. I was just concerned by a non-existing "DataCite 5.0" :smile:

doc/release-notes/6980-datacite-reservation-upgrade
```diff
@@ -0,0 +1,28 @@
...
+   jq '.data.count[].pid' | tr -d '"' | 
+   while read doi
+   do
+      curl -s -H "X-Dataverse-key:$API_TOKEN" -X POST http://localhost:8080/api/pids/:persistentId/reserve?persistentId=$doi
+   done
+```
+
+If you have a large number of unpublished drafts in the database, reserving them one by one may take some time. So if you want to be extra nice to your users, you may warn them ahead of time that some may have trouble publishing their drafts immediately after the upgrade. 
+
+Going forward, once all the DOIs have been reserved for the legacy drafts, some newly-created datasets may still end up with unreserved DOIs. (For example, if Datacite service was unavailable at the time of the datast creation; we don't expect it to happen too often, but it is not impossible). To address this preemptively we recommend that you set up a cronjob to run the script above (perhaps with some extra diagnostics added) weekly or even nightly.
```
pdurbin, line 27:
```suggestion
Going forward, once all the DOIs have been reserved for the legacy drafts, some newly-created datasets may still end up with unreserved DOIs. (For example, if Datacite service was unavailable at the time of the dataset creation; we don't expect it to happen too often, but it is not impossible). To address this preemptively we recommend that you set up a cronjob to run the script above (perhaps with some extra diagnostics added) weekly or even nightly. 
```

doc/release-notes/6980-datacite-reservation-upgrade
```diff
@@ -0,0 +1,28 @@
+## Dataverse installations using Datacite: upgrade action required.
+
+For installations that are using Datacite, Dataverse v5.0 introduces a change in the process of registering the Global Identifier (DOI) for a dataset. Instead of registering it when the dataset is published for the first time, Dataverse will try to "reserve" the DOI when it's created (by registering it as a "draft", using Datacite terminology). When the user publishes the dataset, the DOI will be publicized as well (by switching the registration status to "findable").
```
pdurbin, line 3:
```suggestion
For installations that are using Datacite, Dataverse v5.0 introduces a change in the process of registering the Persistent Identifier (DOI) for a dataset. Instead of registering it when the dataset is published for the first time, Dataverse will try to "reserve" the DOI when it's created (by registering it as a "draft", using Datacite terminology). When the user publishes the dataset, the DOI will be publicized as well (by switching the registration status to "findable"). 
```

I know in the code we call it GlobalId but everywhere else we call it a Persistent Identifier, a PID.

poikilotherm:
Maybe better use Markdown as release notes on GitHub are using Markdown instead of Restructured Text. Sphinx offer support for rendering Markdown files, using this in in dv-k8s...

doc/release-notes/6980-datacite-reservation-upgrade
qqmyers: A missing period? using Datacite. v5.x changes....

landreev: @qqmyers it should be properly punctuated in the current version, yes.

landreev:
Yes, I always get conufsed, about markdown vs. rst... But what is it that it currently doesn't like - the "tripple back tick"? I thought that was md... Will check!

qqmyers:
Could just be the lack of a file extension? There's a mix of .md and no extension in the doc/release-notes dir and the .md ones are rendering

landreev:
> Could just be the lack of a file extension? There's a mix of .md and no extension in the doc/release-notes dir and the .md ones are rendering

D'oh! 
(thank you @qqmyers)

landreev:
(let's put this PR on hold, until #7147 is finalized)

doc/release-notes/6980-datacite-reservation-upgrade.md
```diff
@@ -0,0 +1,26 @@
...
+
+For installations that are using Datacite, Dataverse v5.0 introduces a change in the process of registering the Persistent Identifier (DOI) for a dataset. Instead of registering it when the dataset is published for the first time, Dataverse will try to "reserve" the DOI when it's created (by registering it as a "draft", using Datacite terminology). When the user publishes the dataset, the DOI will be publicized as well (by switching the registration status to "findable"). 
+
+This approach makes the process of publishing datasets simpler and less error-prone. If the DOI has not been reserved (for example, if Datacite was down when the dataset was created), the author can still attempt to publish. But it does help to ensure that the DOI is guaranteed to be available beforehand. New APIs have been provided for finding any unreserved Datacite-issued DOIs in your Dataverse, and for reserving them (see below). If you are upgrading an installation that uses Datacite, we specifically recommend that you reserve the DOIs for all your pre-existing unpublished drafts as soon as Dataverse v5.0 is deployed, since none of them were registered at create time.
```
pdurbin, line 5: The sentence starting with "But" feels unnecessary.

```suggestion
This approach makes the process of publishing datasets simpler and less error-prone. If the DOI has not been reserved (for example, if Datacite was down when the dataset was created), the author can still attempt to publish. New APIs have been provided for finding any unreserved Datacite-issued DOIs in your Dataverse, and for reserving them (see below). If you are upgrading an installation that uses Datacite, we specifically recommend that you reserve the DOIs for all your pre-existing unpublished drafts as soon as Dataverse v5.0 is deployed, since none of them were registered at create time.
```

doc/release-notes/6980-datacite-reservation-upgrade.md
```diff
@@ -0,0 +1,26 @@
+## Dataverse installations using Datacite: upgrade action required.
```
pdurbin, line 1:
```suggestion
## Dataverse installations using Datacite: upgrade action recommended.
```

doc/release-notes/6980-datacite-reservation-upgrade.md
```diff
@@ -0,0 +1,26 @@
...
+   curl -s -H "X-Dataverse-key:$API_TOKEN" http://localhost:8080/api/pids/unreserved |
+   # the API outputs JSON; note the use of jq to parse it:
+   jq '.data.count[].pid' | tr -d '"' | 
+   while read doi
+   do
+      curl -s -H "X-Dataverse-key:$API_TOKEN" -X POST http://localhost:8080/api/pids/:persistentId/reserve?persistentId=$doi
+   done
+``` 
+
+Going forward, once all the DOIs have been reserved for the legacy drafts, you may still get an occasional dataset with an unreserved identfier. Datacite service instability would be a potential cause. There is no reason to expect that to happen often, but it is not impossible. You may consider running the script above (perhaps with some extra diagnostics added) regularly, from a cron job or otherwise, to address this preemptively.
```
pdurbin, line 26:
```suggestion
Going forward, once all the DOIs have been reserved for the legacy drafts, you may still get an occasional dataset with an unreserved identifier. Datacite service instability would be a potential cause. There is no reason to expect that to happen often, but it is not impossible. You may consider running the script above (perhaps with some extra diagnostics added) regularly, from a cron job or otherwise, to address this preemptively.
```

doc/release-notes/6980-datacite-reservation-upgrade.md
landreev: But, without the sentence starting with "But", doesn't it become completely unclear why they need to bother at all? 
Or do you feel it's explained enough by the first sentence ("simpler and less-error prone")?

pdurbin: Yeah, I was thinking it's explained enough.

Issue #2225, opened 2015-06-01
Publish Dataset: When DataCite is down, can't publish but no message to user, other issues.


kcondon:
[2015-06-01T12:12:14.983-0400] [glassfish 4.1] [INFO] [] [edu.harvard.iq.dvn.core.index.DOIEZIdServiceBean] [tid: _ThreadID=53 _ThreadName=jk-connector(4)] [timeMillis: 1433175134983] [levelValue: 800] [[
  modifyMetadata failed]]

[2015-06-01T12:12:14.984-0400] [glassfish 4.1] [INFO] [] [edu.harvard.iq.dvn.core.index.DOIEZIdServiceBean] [tid: _ThreadID=53 _ThreadName=jk-connector(4)] [timeMillis: 1433175134984] [levelValue: 800] [[
  String edu.ucsb.nceas.ezid.EZIDException: internal server error]]

[2015-06-01T12:12:14.984-0400] [glassfish 4.1] [INFO] [] [edu.harvard.iq.dvn.core.index.DOIEZIdServiceBean] [tid: _ThreadID=53 _ThreadName=jk-connector(4)] [timeMillis: 1433175134984] [levelValue: 800] [[
  localized message internal server error]]

[2015-06-01T12:12:14.985-0400] [glassfish 4.1] [INFO] [] [edu.harvard.iq.dvn.core.index.DOIEZIdServiceBean] [tid: _ThreadID=53 _ThreadName=jk-connector(4)] [timeMillis: 1433175134985] [levelValue: 800] [[
  cause null]]

[2015-06-01T12:12:14.985-0400] [glassfish 4.1] [INFO] [] [edu.harvard.iq.dvn.core.index.DOIEZIdServiceBean] [tid: _ThreadID=53 _ThreadName=jk-connector(4)] [timeMillis: 1433175134985] [levelValue: 800] [[
  message internal server error]]

[2015-06-01T12:12:19.496-0400] [glassfish 4.1] [WARNING] [AS-EJB-00056] [javax.enterprise.ejb.container] [tid: _ThreadID=50 _ThreadName=jk-connector(1)] [timeMillis: 1433175139496] [levelValue: 900] [[
  A system exception occurred during an invocation on EJB EjbDataverseEngine, method: public java.lang.Object edu.harvard.iq.dataverse.EjbDataverseEngine.submit(edu.harvard.iq.dataverse.engine.command.Command) throws edu.harvard.iq.dataverse.engine.command.exception.CommandException]]

[...]

PR #7147, opened 2020-07-29
Reserve PID at publication if needed

### LINKED ISSUE #7146: Backward compatibility/error recovery for PID reservation during publish
The preferred way of addressing the change in 5.0 to making DataCite DOIs reservable at dataset creation time is for an admin to initial (and possibly periodically) run a script to identify unreserved DOIs and reserve them. However, it would be useful to support reservation at publication time if publishing is attempted prior to admin action (either at the 5.0 transition or after a DataCite outage). Creating this issue and PR so we can consider whether this makes sense.

qqmyers:
**What this PR does / why we need it**: Reserves a PID if needed  at publication time (using a reserve at creation-time PID provider but none has been created, so either a backward-compatibility issue or after a PID provider outage)

Closes #7146

**Special notes for your reviewer**: Would slightly change the guides documentation if adopted

**Suggestions on how to test this**: Try publishing without having reserved DOIs at DataCite

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
```diff
@@ -75,7 +74,17 @@ public PublishDatasetResult execute(CommandContext ctxt) throws CommandException
         boolean reservingPidsSupported = !idServiceBean.registerWhenPublished();
         if (reservingPidsSupported) {
             if (theDataset.getGlobalIdCreateTime() == null) {
-                throw new IllegalCommandException(BundleUtil.getStringFromBundle("publishDatasetCommand.pidNotReserved"), this);
+            	//Backward compatibility and recovery from PID provider outage - if we should have reserved at creation time but didn't, try here as well to keep moving. If we still can't reserve (service down, connflict, etc.) then bail
+            	try {
+                    String returnString = idServiceBean.createIdentifier(this.getDataset());
```
landreev, line 79: If we are going back to trying to register an unreserved DOI during publishing, it should probably happen in the other, asynchronous part of the workflow - in FinalizeDatasetPublicationCommand. Otherwise we can again end up in a situation where Datacite decides to hang for a while, and this is happening while the dataset is still unlocked, etc.
And in the finalize command we already have this code:
```
if ( theDataset.getGlobalIdCreateTime() == null ) {
   try {
      registerExternalIdentifier(theDataset, ctxt);
   }
   ...
```
but of course that registerExternalIdentifier method will need to be modified, because in its current form it would get a new pid assigned to the dataset, if the old one already exists on the Datacite end...

qqmyers: Does DataCite hang/delay an http call (versus fail)? If so I can move this/integrate with registerExternalIdentifier to avoid waiting for the call to succeed/fail. It seems like going async and not being able to respond to the user when the first step fails (before file validation, file DOI registrations) except with the async notification isn't great either, but if there are cases where reserving the dataset DOI might induce people to click again, I guess it makes sense.

landreev: Well, it is a network call - so it can hang... I believe we have seen registration calls actually taking a long time and/or timing out, during the worst of their service troubles.
But yes, one nice thing about the current implementation is that we can exit right away with a clear message, before the dataset gets locked and the command goes async.
(we could alternatively try and leave it in the first part of the command - but make the call with a forced short timeout?)

qqmyers: OK - updated registerExternalIdentifier and moved things to Finalize. As far as I can tell, the httpClient should be limiting the wait for a response to one minute (system default) but that could be lowered in the DataCIteRESTfullClient (easiest to set once for all calls since there's one shared client) and I think the whole call to registerExternalIdentifier could move to Publish then with a cut/paste.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -155,32 +155,30 @@ protected void tidyUpFields(DatasetVersion dsv) {
...
             if ( globalIdServiceBean != null ) {
                 if (globalIdServiceBean instanceof FakePidProviderServiceBean) {
-                    try {
-                        globalIdServiceBean.createIdentifier(theDataset);
-                    } catch (Throwable ex) {
-                        logger.warning("Problem running createIdentifier for FakePidProvider: " + ex);
-                    }
-                    theDataset.setGlobalIdCreateTime(getTimestamp());
-                    theDataset.setIdentifierRegistered(true);
-                    return;
```
qqmyers, line 170: I think this was just to avoid the retry attempts - same set of calls otherwise.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -155,32 +155,30 @@ protected void tidyUpFields(DatasetVersion dsv) {
...
+                        if(retry) {
+                            do  {
+                                theDataset.setIdentifier(ctxt.datasets().generateDatasetIdentifier(theDataset, globalIdServiceBean));
+                                logger.log(Level.INFO, "Attempting to register external identifier for dataset {0} (trying: {1}).",
+                                    new Object[]{theDataset.getId(), theDataset.getIdentifier()});
+                                attempts++;
+                            } while (globalIdServiceBean.alreadyExists(theDataset) && attempts <= FOOLPROOF_RETRIAL_ATTEMPTS_LIMIT);
                         }
-
-                        if (globalIdServiceBean.alreadyExists(theDataset)) {
```
qqmyers, line 183: Similar - extra call to alreadyExists since attempts will be over the limit if the while call above doesn't find alreadyExists()==false by the end.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -83,9 +83,20 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
+		 *   case it will retry if collisions are found 
+		 * - In the case where the PID Provider
+		 *   does support reservations, but the Dataset's PID hasn't been reserved (was
+		 *   created prior to v5.0 and admin hasn't run reserve API, or PID Provider was
+		 *   offline at creation time), in which case no retries will be attempted.
+		 */
+        GlobalIdServiceBean idServiceBean = GlobalIdServiceBean.getBean(getDataset().getProtocol(), ctxt);
         if ( theDataset.getGlobalIdCreateTime() == null ) {
-            registerExternalIdentifier(theDataset, ctxt);
+            registerExternalIdentifier(theDataset, ctxt, idServiceBean.registerWhenPublished());
```
qqmyers, line 99: Independent of the larger discussion about the PR, if we keep the new boolean flag, it could always be set to false here - PID providers that do not register draft DOIs would then try once to create a DOI at publication time and fail if the DOI is already in use, versus retrying and causing all of the files to be 'lost' (at the path for the old/conflicted DOI instead of one found in a retry). Retries would then only occur in the create command (for providers that can reserve).

landreev: This does make sense - maybe we should just hard-code it to false, when the method is called from the publication command. It's not exactly clear what the point would be, in keeping the current behavior for handles. 
?

pdurbin:
Should this issue be tested as well? Hat tip to @mheppler who found it. Feels related.

-  Publish Dataset: When DataCite is down, can't publish but no message to user, other issues. #2225

qqmyers:
At a minimum the recent changes would make that issue obsolete. I think between the several PRs to reserve PIDs, add notification, etc. most/all of the recommendations there are implemented.

PR #7152, opened 2020-07-31
Add timer for saved searches #6910

### LINKED ISSUE #6910: Add a timer for running saved searches
We currently have a way to define a saved search that gets associated with a dataverse (anything that matches that search gets linked to that dataverse) and an API to rerun the search. However, this API currently needs to be called manually*.

(*) when saved search functionality was added, we did not yet have a Timer Service, and we had planned on just using a cron job to automatically call the api

We should add a Timer, similar to Export Service or Harvesting, that runs these saved searches automatically.

rtreacy:
The timer method wraps the makeLinksForAllSavedSearches in a no-param method returning void with the @Schedule annotation.
The timer is sheduled to run once a week on Sunday at 12:30 AM. If we decide a different schedule is better in general, we
should adjust thiis in the code before release. In special cases the schedule can be modified using and ejb-jar.xml file
of the form(for ex Tuesday 2:30 PM) :

<?xml version="1.0" encoding="UTF-8"?>
<ejb-jar xmlns = "http://java.sun.com/xml/ns/javaee"
         version = "3.1"
         xmlns:xsi = "http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation = "http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/ejb-jar_3_1.xsd">
    <enterprise-beans>
        <session>
            <ejb-name>SavedSearchServiceBean</ejb-name>
            <ejb-class>edu.harvard.iq.dataverse.search.savedsearch.SavedSearchServiceBean</ejb-class>
            <session-type>Stateless</session-type>
            <timer>
                <schedule>
                    <minute>30</minute>
                    <hour>14</hour>
                    <day-of-week>Tue</day-of-week>
                </schedule>
                <timeout-method>
                    <method-name>makeLinksForAllSavedSearchesTimer</method-name>
                </timeout-method>
            </timer>
        </session>
    </enterprise-beans>
</ejb-jar>

Closes #6910 

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/_static/admin/ejb-jar.xml
+++ b/doc/sphinx-guides/source/_static/admin/ejb-jar.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ejb-jar xmlns = "http://java.sun.com/xml/ns/javaee" 
+         version = "3.2"
+         xmlns:xsi = "http://www.w3.org/2001/XMLSchema-instance" 
+         xsi:schemaLocation = "http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/ejb-jar_3_1.xsd">
+    <enterprise-beans>
+        <session>
+            <ejb-name>SavedSearchServiceBean</ejb-name>
+            <ejb-class>edu.harvard.iq.dataverse.search.savedsearch.SavedSearchServiceBean</ejb-class>
+            <session-type>Stateless</session-type>
+            <timer>
+                <schedule>
+                    <minute>30</minute>
+                    <hour>14</hour>
+                    <day-of-week>2</day-of-week>
+                </schedule>
+                <timeout-method>
+                    <method-name>makeLinksForAllSavedSearchesTimer</method-name>
+                </timeout-method>
+            </timer>
+        </session>
+    </enterprise-beans>
+</ejb-jar>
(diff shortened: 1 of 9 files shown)


qqmyers:
Is it true that you have to rebuild, or can you just drop the ejb file into the dir after the war is unpacked? (and restart payara I assume). That would be something many installations could handle.

pdurbin:
@qqmyers I don't know but I think what we have here is sufficient for our needs. I expect most installations won't bother changing it. Here are some resources I looked at:

- Using the Timer Service: https://eclipse-ee4j.github.io/jakartaee-tutorial/ejb-basicexamples005.html
- https://stackoverflow.com/questions/8320055/possible-to-change-ejb-parameter-at-runtime-for-schedule-annotation
- https://stackoverflow.com/questions/15973661/dynamic-parameters-for-schedule-method-in-an-ejb-3-x

kcondon:
The precompiled Sunday 12:30am time works as expected. Did not test the configurable time, need to compile it. Dropping it in an already deployed app doesn't work.

pdurbin:
> Dropping it in an already deployed app doesn't work.

Thanks for testing this, @kcondon

rtreacy:
If putting in documentation about reconfiguring is going to cause people to reconfigure because we say it can be done, I think we should remove the documentation

Issue #7140, opened 2020-07-28
Add Archiver to store to Google Cloud


qqmyers:
QDR is using an Archiver class that can send archival bags to Google cloud storage (normal or cold), but this code hasn't been merged into the community version because of a library conflict between the Google API and Glassfish 4. That is hopefully resolved with Payara which would make this work the same way the Duracloud/Dspace and local archiver classes work. I'm adding the issue as a reminder to try/do this at some point. If it's a post 5.0 priority for anyone, leave a note.

PR #7163, opened 2020-08-05
Remove guava library which is reported as potential security issue. D…

rtreacy:
…ependency is apparently very old and not necessary any longer. Code compiles cleanly when the dependency is removed from the project

Closes #

**Additional documentation**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,12 +262,6 @@
             <version>${flyway.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>16.0.1</version>
-            <type>jar</type>
-        </dependency>
-        <dependency>
             <groupId>jakarta.platform</groupId>
             <artifactId>jakarta.jakartaee-api</artifactId>
             <version>${jakartaee-api.version}</version>


rtreacy:
Need to take another look at this. Removing the dependency from the pom causes a later version of the library to be pulled in, but it still has the reported memory problem.

rtreacy:
One source of the dependency is the use of org.everit.json.schema.* in edu.harvard.iq.dataverse.provenance.ProvInvestigator
It was introduce in 2018. Can this json functionality be implemented with standARD Java EE json?

qqmyers:
FWIW: I think the google cloud library I've used in #7140 uses guava and if I recall correctly, it needs a newer version ((v>20)  than the one in Glassfish (in the /modules subdir). I'm not sure if payara also uses guava as glassfish did, but that's something to check.

pdurbin:
> One source of the dependency is the use of org.everit.json.schema.* in edu.harvard.iq.dataverse.provenance.ProvInvestigator
> It was introduce in 2018. Can this json functionality be implemented with standARD Java EE json?

The functionality is validating JSON files against a JSON Schema. I don't know if Java/Jakarta EE does that or not. Good question.

A quick search isn't promising:

- JSON Schema itself isn't standardized: https://blogs.oracle.com/theaquarium/json-schema
- JSON Schema support as a non-goal for JSON-B: https://jakarta.ee/specifications/jsonb/2.0/jsonb-spec-2.0-RC2.html

poikilotherm:
FWIW @pdurbin JSON Schema is about to become an IETF standard some day. See https://json-schema.org/

rtreacy:
btw, it turned out that there were 5 libraries we use that are dependent on guava. Instead of removing it, there is now a pull request that upgrades the guava version to 29.0-jre

PR #7164, opened 2020-08-06
7046 dataverse 5 release notes

### LINKED ISSUE #7046: Dataverse 5 Release Notes
It's that time again. As we approach the release for Dataverse 5, we'll start pulling together the release notes.

djbrooke:

The release notes for Dataverse 5!

Closes #7046

No longer a draft PR. Thanks for all the feedback so far!

Bring it into your favorite text editor and take a look. 

N/A 

N/A

**Additional documentation**:

N/A


doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+### New JVM Options and Database Settings
+
+Several new JVM options and DB Settings have been added in this release. More documentation about each of these settings can be found in the Configuration section of the [Installation Guide](http://guides.dataverse.org/en/5.0/installation/config.rst).
+
+#### New JVM Options
+
+- :doi.baseurlstringnext: Set with a value of "https://api.datacite.org" for production environments and "https://api.test.datacite.org" for test environments. Must be set if you are using DataCite as your DOI provider.
+- :dataverse.useripaddresssourceheader: If set, specifies an HTTP Header such as X-Forwarded-For to use to retrieve the user's IP address. This setting is useful in cases such as running Dataverse behind load balancers where the default option of getting the Remote Address from the servlet isn't correct (e.g. it would be the load balancer IP address). Note that unless your installation always sets the header you configure here, this could be used as a way to spoof the user's address.
```
pdurbin, line 159: Just noting that this sounds a little scary:

"Note that unless your installation always sets the header you configure here, this could be used as a way to spoof the user's address."

Maybe we could end with "If you don't need this setting, don't set it."

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+- :FileValidationOnPublishEnabled: Toggles validation of the physical files in the dataset when it's published, by recalculating the checksums and comparing against the values stored in the DataFile table. By default this setting is absent and Dataverse assumes it to be true. If enabled, the validation will be performed asynchronously, similarly to how we handle assigning persistent identifiers to datafiles, with the dataset locked for the duration of the publishing process.
```
pdurbin, line 160: :FileValidationOnPublishEnabled is a database setting, not a JVM option, so it should be moved down there.

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+Many updates have been made to address common Harvesting failures. You may see Harvests complete more often and have a higher success rate on a dataset-by-dataset basis.
+
+### New JVM Options and Database Settings
+
+Several new JVM options and DB Settings have been added in this release. More documentation about each of these settings can be found in the Configuration section of the [Installation Guide](http://guides.dataverse.org/en/5.0/installation/config.rst).
+
+#### New JVM Options
+
+- :doi.baseurlstringnext: Set with a value of "https://api.datacite.org" for production environments and "https://api.test.datacite.org" for test environments. Must be set if you are using DataCite as your DOI provider.
```
qqmyers, line 158: @djbrooke - FWIW:  #7116 changes this but I see it didn't get tagged as 5.0 - as is there are two options that both point to the same URL:  doi.mdcbaseurlstring and :doi.baseurlstringnext.

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+- :FileValidationOnPublishEnabled: Toggles validation of the physical files in the dataset when it's published, by recalculating the checksums and comparing against the values stored in the DataFile table. By default this setting is absent and Dataverse assumes it to be true. If enabled, the validation will be performed asynchronously, similarly to how we handle assigning persistent identifiers to datafiles, with the dataset locked for the duration of the publishing process.
+- :http.request-timeout-seconds: To facilitate large file upload and download, the Dataverse installer bumps the Payara **server-config.network-config.protocols.protocol.http-listener-1.http.request-timeout-seconds** setting from its default 900 seconds (15 minutes) to 1800 (30 minutes).
+
+#### New Database Settings
+
+- :CustomZipDownloadServiceUrl: If defined, this is the URL of the zipping service outside the main Application Service where zip downloads should be directed (instead of /api/access/datafiles/).
```
pdurbin, line 165: "Application Service" sounds weird to me. Maybe "application server"? Perhaps the idea is that there can be multiple application servers forming a service and the zipper can live outside all of that. Maybe just say "the main application"? I don't know.

doc/release-notes/5.0-release-notes.md
qqmyers, line 160: This is a setting not a jvm option

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+- :ShibAttributeCharacterSetConversionEnabled: By default, all attributes received from Shibboleth are converted from ISO-8859-1 to UTF-8. You can disable this behavior by setting to false.
+- :ChronologicalDateFacets: Facets with Date/Year are sorted chronologically by default, with the most recent value first. To have them sorted by number of hits, e.g. with the year with the most results first, set this to false.
+- :NavbarGuidesUrl: Set to a fully-qualified URL which will be used for the "User Guide" link in the navbar.
+
+### Custom Analytics Code Changes
+
+You should update your custom analytics code to implement necessary changes for tracking updated dataset and file buttons. There was also a fix to the analytics code that will now properly track downloads for tabular files.
+
+We have updated the documentation and sample analytics code snippet provided in [Installation Guide > Configuration > Web Analytics Code](http://guides.dataverse.org/en/5.0/installation/config.html#web-analytics-code) to reflect the changes implemented in this version (#6938/#6684).
```
pdurbin, line 174: After reading the first line I'm wondering how to update my custom analytics code.

The second line gives the link but perhaps these could be woven together a bit more smoothly.

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Tracking users' IP addresses behind an address-masking proxy
```
pdurbin, line 176: This is the other non Title Case I mentioned above.

doc/release-notes/5.0-release-notes.md
qqmyers: Or 'see guide for proper use and details about the security concern.'

qqmyers, line 159: THese jvm options also don't get a leading ':' like settings do they?

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+The buttons available on the Dataset and File pages have been redesigned. This change is to provide more scalability for future expanded options for data access and exploration, and to provide a consistent experience between the two pages. The dataset and file pages have also be redesigned to be more responsive and function better across multiple devices.
+
+This is an important step in the incremental process of the Dataset and File Redesign, following the release of on-page previews, filtering and sorting options, tree view, and other enhancements. Additional features in support of these redesign efforts will follow in later 5.x releases.
+
+### Payara 5
+
+A major upgrade of the application server provides security updates, access to new features like MicroProfile Config API, and will enable eagerly anticipated upgrades to other core technologies.
+
+Note that moving from Glassfish to Payara will be required as part of the move to Dataverse 5. If your installation upgrades to Dataverse 5 without upgrading to Payara, the first step of any troubleshooting will be to suggest an upgrade to Payara.
```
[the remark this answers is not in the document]
djbrooke: Will remove

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+Please note that this is a major release and these are long release notes. We offer no apologies. :)
+
+## Release Highlights
+
+### Continued Dataset and File Redesign: Dataset and File Button Redesign, Responsive Layout
+
+The buttons available on the Dataset and File pages have been redesigned. This change is to provide more scalability for future expanded options for data access and exploration, and to provide a consistent experience between the two pages. The dataset and file pages have also be redesigned to be more responsive and function better across multiple devices.
+
+This is an important step in the incremental process of the Dataset and File Redesign, following the release of on-page previews, filtering and sorting options, tree view, and other enhancements. Additional features in support of these redesign efforts will follow in later 5.x releases.
```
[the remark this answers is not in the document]
djbrooke: Yes

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Payara 5
+
+A major upgrade of the application server provides security updates, access to new features like MicroProfile Config API, and will enable eagerly anticipated upgrades to other core technologies.
```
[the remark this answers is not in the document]
djbrooke: I was trying to channel @poikilotherm here!

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,313 @@
...
+
+Primefaces, the open source UI framework upon which the Dataverse front end is built, has been updated to the most recent version. This provides security updates and bug fixes and will also allow Dataverse developers to take advantage of new features and enhancements.
+
+### Download Dataset
+
+Users can now more easily Download all files in Dataset. If this causes server instability, it's suggested that Dataverse Installation Administrators take advantage of the new Standalone Zipper Service described below.
+
+#### Download All Option on the Dataset Page
+
+In previous versions of Dataverse, downloading all files from a dataset meant several clicks to select files and initiate the download. The Dataset Page now includes a Download All option for both the original and archival formats of the files in a dataset under the "Access Dataset" button.
```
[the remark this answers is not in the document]
djbrooke: "Download Dataset" is a higher level heading and is an umbrella for both the Download All through the UI and the Download all through the API. I wanted to get in a plug for the zipper service here as well

doc/release-notes/5.0-release-notes.md
djbrooke: Only in @djbrooke's release note world. I'll update.

djbrooke:
@pdurbin and @qqmyers thanks for the excellent feedback. I made a large percentage of the changes that you suggested. These release notes are indeed a 🐻 . 

@qqmyers I'll respond to your comment about doi.baseurlstringnext above.

doc/release-notes/5.0-release-notes.md
djbrooke: @qqmyers @scolapasta what is the preferred resolution here? Should #7116 be tagged as 5.0 or no?

qqmyers: I would recommend that - to avoid having duplicate jvm options in 5.0 that would then disappear. 

If #7116 is a concern because it changes build scripts, a more minimalist approach would just be to replace uses of the newly created doi.baseurlstringnext with doi:mdcbasurlstring (and maybe note that mdcbaseurlstring must be set/is no longer optional/just for mdc) and then use an update to #7116 to give that jvm option a more neutral/correct name.

djbrooke: Thanks @qqmyers I'll tag it as 5.0 and we can accept any feedback in that PR. 

cc: @pdurbin as I saw you brought this up on a Slack thread as well.

doc/release-notes/5.0-release-notes.md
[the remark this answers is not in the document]
pdurbin: Ah, umbrella term and a higher level heading. Got it.

doc/release-notes/5.0-release-notes.md
pdurbin: Ah, I didn't realize there's more detail in the guides about the security concern. Sure, mentioning this would be good. Or we could remove the scary line and let people discover it in the guides. Less is more.

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,321 @@
...
+- If the directory or name of an existing or newly uploaded file is edited in such a way that would create a directory/filename combination that already exists, Dataverse will display an error.
+- If a user attempts to replace a file with another file that has the same checksum, an error message will be displayed and the file will not be able to be replaced.
+- If a user attempts to replace a file with a file that has the same checksum as a different file in the dataset, a warning will be displayed.
+- Files without extensions can now be uploaded through the UI.
+
+### Pre-Publish DOI Reservation with DataCite
+
+Dataverse installations using DataCite will have DOIs reserved as "draft" with DataCite when the Dataset is created, instead of the DOI reservation taking place locally at create time and the reservation and switch to "findable" taking place with DataCite at publish time. This allows the DOI to reserved earlier in the publication process.
```
landreev, line 63: crap, I've typed, and lost a comment here twice in a row... there may be something wrong with this paragraph: it appears to be saying "dataverse will be reserving the DOI at create time, instead of reserving it at create time"? - maybe not (it says "locally" in the second part?) - but there may be simply TMI in it. I would leave the technical details (about "findable" and "draft" etc.) in the upgrade instruction further down; but only say 
"Dataverse installations using DataCite will be able to reserve the persistent identifiers for datasets ahead of publishing time." - or something along these lines?

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,321 @@
...
+
+ Replace the -XX:PermSize and -XX:MaxPermSize JVM options with -XX:MetaspaceSize and -XX:MaxMetaspaceSize.
+
+   <jvm-options>-XX:MetaspaceSize=256m</jvm-options>
+   <jvm-options>-XX:MaxMetaspaceSize=512m</jvm-options>
+
+Set both Xmx and Xms at startup to avoid runtime re-allocation. Your Xmx value should likely be higher:
+
+   <jvm-options>-Xmx2048m</jvm-options>
+   <jvm-options>-Xms2048m</jvm-options>
```
landreev, line 260: Not sure why it was needed, mentioning the Xmx adn Xms lines to the upgrade instruction. I mean, it's an existing installation - they must have these settings in their config already. I can see how this can confuse somebody. 
I would simply remove this paragraph and the last two jvm-options lines.

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,321 @@
...
+15. Update Astrophysics Metadata Block (if used)
+
+   `wget https://github.com/IQSS/dataverse/releases/download/5.0/astrophysics.tsv`
+   `curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @astrophysics.tsv -H "Content-type: text/tab-separated-values"`
+
+16. (Recommended) Run ReExportall to update JSON Exports  
+
+   <http://guides.dataverse.org/en/5.0/admin/metadataexport.html?highlight=export#batch-exports-through-the-api>
+
+17. For installations using DataCite, pre-register DOIs:
```
landreev, line 302: Since there is already a detailed section for this step, above - does it need to be here, in the upgrade-to-payara instruction - probably not?

doc/release-notes/5.0-release-notes.md
```diff
@@ -0,0 +1,321 @@
...
+   curl -s -H "X-Dataverse-key:$API_TOKEN" http://localhost:8080/api/pids/unreserved |
+   # the API outputs JSON; note the use of jq to parse it:
+   jq '.data.count[].pid' | tr -d '"' | 
+   while read doi
+   do
+      curl -s -H "X-Dataverse-key:$API_TOKEN" -X POST http://localhost:8080/api/pids/:persistentId/reserve?persistentId=$doi
+   done
+```
+
+18. Set up Standalone Zipper Service if desired. Please consult the scripts/zipdownload/README.md in the Dataverse 5 source tree for more information.
```
landreev, line 321: similarly, this is already explained in its own section above - not sure it needs to be repeated in the upgrade-to-payara instruction - ?

donsizemore:
@landreev on Xmx and Xms - perhaps the paragraph should've begun with "Adjust" or "Review" rather than set. The step was included a recommendation in the performance tuning document but I'm fine with having it removed, especially if it will cause confusion.

doc/release-notes/5.0-release-notes.md
djbrooke: Updated text in https://github.com/IQSS/dataverse/commit/2ec68b34c4671203fc1360ff88d0de95989d61d6

I agree that at this point in the doc (release highlights) we can be less technical.

doc/release-notes/5.0-release-notes.md
djbrooke: @landreev I removed this in https://github.com/IQSS/dataverse/commit/c593f89bff1e4ba3e5988bc9dd81708ff739f0a1

heads up to @donsizemore in case this was included for some other reason we're not seeing

doc/release-notes/5.0-release-notes.md
djbrooke: We've done this previously, that is, added it in the notes above and called out and individual step.

I added a (Recommended for installations using DataCite) at the start of the instruction. I also added a note for Add doi.dataciterestapiurlstring

Let me know what you think following the updates in https://github.com/IQSS/dataverse/commit/abc4d9bc136d027862dab7b8b72381b89e340d9a

doc/release-notes/5.0-release-notes.md
djbrooke: I added the experimental note in https://github.com/IQSS/dataverse/commit/abc4d9bc136d027862dab7b8b72381b89e340d9a. I think it's fine to add this in two different instruction formats and it matches what we've done in previous notes but if you feel strongly I can revisit. I don't do the upgrades myself so if this sort of thing is not helpful to mention multiple times I'm fine to adjust it.

PR #7165, opened 2020-08-06
7159 fix restassured etc

### LINKED ISSUE #7159: Fix RestAssured tests broken by #6918
The PR required extra RestAssured code to accommodate the changed publish behavior. It wasn't included so it broke the tests. 
This issue is to handle the followup PR to fix it. 
Also, per discussion with @scolapasta, addressing the long Sleep statement in the Dataset service bean, by moving the execution of the finalize command to the onSuccess() part of the PublishDataset command.

landreev:
The issue was opened initially because I forgot to check in some RestAssured code (UtilIT.java, etc.) that was needed for the tests to run. (The new publish mechanism involves waiting for the dataset to unlock, if locks are present; so that was missing). 
But we also discovered some non-test related issues that needed to be addressed: 
There was a massive sleep statement in the service bean before calling the async. portion of the command that we forgot about. In this PR I added an onSuccess() to the PublishDataset command; the second command call is made from there, and this appears to make any sleep delay unnecessary (per suggestion from @scolapasta)
Due to some weirdness in how the onSuccess() of the second command is executed, I also realized that even after all the changes in #6918 there was still a possibility for the dataset to become unlocked before the publication was finalized. (If the indexing and/or export are taking a long time, for ex.). This was addressed by moving the unlocking code to onSuccess(). 
Further simplified the RestAssured changes (only UtilIT.java is modified).
Modifed PublishDatasetResult to treat "publish workflow" and "asynchronous publishing in progress" as distinct states.

Closes #7159

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
@@ -788,21 +788,22 @@ public class DatasetServiceBean implements java.io.Serializable {
      * Any failure notifications to users should be sent from inside the command.
      */
     @Asynchronous
+    @TransactionAttribute(TransactionAttributeType.SUPPORTS)
     public void callFinalizePublishCommandAsynchronously(Long datasetId, CommandContext ctxt, DataverseRequest request, boolean isPidPrePublished) {
 
         // Since we are calling the next command asynchronously anyway - sleep here 
         // for a few seconds, just in case, to make sure the database update of 
         // the dataset initiated by the PublishDatasetCommand has finished, 
         // to avoid any concurrency/optimistic lock issues. 
-        // Aug. 2020/v5.0: It appears to be working consistently without any 
+        // Aug. 2020/v5.0: It MAY be working consistently without any 
         // sleep here, after the call the method has been moved to the onSuccess()
         // portion of the PublishDatasetCommand. I'm going to leave the 1 second
-        // sleep commented-out below for now: -- L.A.
-        /*try {
+        // sleep below, for just in case reasons: -- L.A.
+        try {
             Thread.sleep(1000);
         } catch (Exception ex) {
-            logger.warning("Failed to sleep for 1 second.");
-        }*/
+            logger.warning("Failed to sleep for a second.");
+        }
         logger.fine("Running FinalizeDatasetPublicationCommand, asynchronously");
         Dataset theDataset = find(datasetId);
         try {
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
@@ -780,21 +780,36 @@ public class DatasetServiceBean implements java.io.Serializable {
         return workflowComment;
     }
     
+    /**
+     * This method used to throw CommandException, which was pretty pointless 
+     * seeing how it's called asynchronously. As of v5.0 any CommanExceptiom 
+     * thrown by the FinalizeDatasetPublicationCommand below will be caught 
+     * and we'll log it as a warning - which is the best we can do at this point.
+     * Any failure notifications to users should be sent from inside the command.
+     */
     @Asynchronous
-    public void callFinalizePublishCommandAsynchronously(Long datasetId, CommandContext ctxt, DataverseRequest request, boolean isPidPrePublished) throws CommandException {
+    public void callFinalizePublishCommandAsynchronously(Long datasetId, CommandContext ctxt, DataverseRequest request, boolean isPidPrePublished) {
 
         // Since we are calling the next command asynchronously anyway - sleep here 
         // for a few seconds, just in case, to make sure the database update of 
         // the dataset initiated by the PublishDatasetCommand has finished, 
         // to avoid any concurrency/optimistic lock issues. 
-        try {
-            Thread.sleep(15000);
+        // Aug. 2020/v5.0: It appears to be working consistently without any 
+        // sleep here, after the call the method has been moved to the onSuccess()
+        // portion of the PublishDatasetCommand. I'm going to leave the 1 second
+        // sleep commented-out below for now: -- L.A.
+        /*try {
+            Thread.sleep(1000);
         } catch (Exception ex) {
-            logger.warning("Failed to sleep for 15 seconds.");
-        }
+            logger.warning("Failed to sleep for 1 second.");
+        }*/
         logger.fine("Running FinalizeDatasetPublicationCommand, asynchronously");
         Dataset theDataset = find(datasetId);
-        commandEngine.submit(new FinalizeDatasetPublicationCommand(theDataset, request, isPidPrePublished));
+        try {
+            commandEngine.submit(new FinalizeDatasetPublicationCommand(theDataset, request, isPidPrePublished));
+        } catch (CommandException cex) {
+            logger.warning("CommandException caught when executing the asynchronous portion of the Dataset Publication Command.");
+        }
     }
     
     /*
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -1020,7 +1020,7 @@ public class Datasets extends AbstractApiBean {
             PublishDatasetResult res = execCommand(new PublishDatasetCommand(ds,
                         createDataverseRequest(user),
                     isMinor));
-            return res.isCompleted() ? ok(json(res.getDataset())) : accepted(json(res.getDataset()));
+            return res.isWorkflow() ? accepted(json(res.getDataset())) : ok(json(res.getDataset()));
             }
         } catch (WrappedResponse ex) {
             return ex.getResponse();
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.logging.Logger;
 import static java.util.stream.Collectors.joining;
+import static edu.harvard.iq.dataverse.engine.command.impl.PublishDatasetResult.Status;
 
 /**
  * Kick-off a dataset publication process. The process may complete immediately, 
@@ -91,7 +92,7 @@ public class PublishDatasetCommand extends AbstractPublishDatasetCommand<Publish
             theDataset = ctxt.em().merge(theDataset);
             ctxt.em().flush();
             ctxt.workflows().start(prePubWf.get(), buildContext(theDataset, TriggerType.PrePublishDataset, datasetExternallyReleased));
-            return new PublishDatasetResult(theDataset, false);
+            return new PublishDatasetResult(theDataset, Status.Workflow);
             
         } else{
             // We will skip trying to register the global identifiers for datafiles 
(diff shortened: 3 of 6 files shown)


qqmyers:
Wasn't the point (one of the points) of having onsuccess to reduce the time during which things were locked/the apparent time needed to complete? The move to unlock in onsuccess of FinalizePub would undo that. That may not be a bad thing - the dataset doesn't act fully published until the index is updated and metadata exports are available - but it does seem like a change. And, if FullTextIndexing is on, the lock may stay around longer than is reasonable (whatever that is). (If I'm waiting to make sure my publication worked and didn't fail and I have to wait for full-text indexing to complete, even though publish itself has been done for a while - is that what we'd want?)

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -780,21 +780,36 @@ public WorkflowComment addWorkflowComment(WorkflowComment workflowComment) {
...
+            logger.warning("Failed to sleep for 1 second.");
+        }*/
         logger.fine("Running FinalizeDatasetPublicationCommand, asynchronously");
         Dataset theDataset = find(datasetId);
-        commandEngine.submit(new FinalizeDatasetPublicationCommand(theDataset, request, isPidPrePublished));
+        try {
+            commandEngine.submit(new FinalizeDatasetPublicationCommand(theDataset, request, isPidPrePublished));
+        } catch (CommandException cex) {
+            logger.warning("CommandException caught when executing the asynchronous portion of the Dataset Publication Command.");
+        }
```
qqmyers, line 812: if this submit fails, isn't something beyond logging required?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1020,7 +1020,7 @@ public Response publishDataset(@PathParam("id") String id, @QueryParam("type") S
             PublishDatasetResult res = execCommand(new PublishDatasetCommand(ds,
                         createDataverseRequest(user),
                     isMinor));
-            return res.isCompleted() ? ok(json(res.getDataset())) : accepted(json(res.getDataset()));
+            return res.isWorkflow() ? accepted(json(res.getDataset())) : ok(json(res.getDataset()));
```
qqmyers, line 1023: does this result in a 200 when the dataset is still inprogress/publishing not yet finalized? Seems like 202 is the right code for that (as it was) and the test should be watching for a 202?

landreev:
@qqmyers  

The primary reason this had to be done was that there is something wrong with how the execute() and onSuccess() are executed in the Finalize command specifically. (Apparently on account of how it's called from the dataset service - @scolapasta can explain it better). Normally execute() would be in its own transaction; then, by the time it's done the db update of the dataset is finalized, and the dataset/version appear published; then onSuccess() kicks in and does its thing.  With Finalize, the 2 appear to run in the *same* transaction - so the dataset does not get updated until onSuccess is done. (this must be happening with other commands throughout the application; and we'll  probably need to address it). The locks on the other hand were removed in transactions, at the end of execute(), so they would disappear before onSuccess() was done - meaning, while the dataset was still appearing as draft. I honestly wasn't thinking much about why we were indexing and exporting that way... moving the unlocking was just the easiest way to fix it. 

Alternatively, once the transaction situation is fixed, the unlock can be moved back. Gustavo had a suggestion of a transaction attribute in a certain place that could fix it (for this specific command call). The question is, do you believe this is important enough to address right now, or can we live with it in v5.0? 


> Wasn't the point (one of the points) of having onsuccess to reduce the time during which things were locked/the apparent time needed to complete?  ... That may not be a bad thing - the dataset doesn't act fully published until the index is updated and metadata exports are available - but it does seem like a change. ...

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
landreev: Not sure what else could be done here.  I added a comment there; we weren't even logging before - that method was throwing CommandException - which was pointless combining with Async. 
The main idea is that the finalize command should send any notifications, or do any other failure handling internally.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
landreev: With the async publication in progress it returns a 200; the presence of the "finalizing" lock is the indication of it being in progress. 
To me it seemed like the 202 return code was specifically added for the workflow situation; and there may be some third party code that relies on it in that context. When the async publication was added we just used that mechanism that was there for the workflow.  But using the same code for this scenario seemed wrong. 
And most importantly, yes, this allowed me not to have to modify much more RestAssured code.

landreev:
I'm going to try @scolapasta's suggestion for a fix, for callFinalizePublishCommandAsynchronously(). Just want to confirm first that the tests are passing in the standard environment.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: I understand the practicality and don't think it's a big issue in terms of any effect, but I think 202 is the right code for anything asynchronous that could fail after you get your response. If its not worth addressing now, I'd suggest a ToDo/note in the code.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -2232,12 +2251,27 @@ static Boolean sleepForSearch(String searchPart, String apiToken,  String subTre
...
     
-    
-    
+    // backward compatibility version of the method that takes long for the id:
     static Response checkDatasetLocks(long datasetId, String lockType, String apiToken) {
+        String datasetIdAsString = String.valueOf(datasetId);
+        return checkDatasetLocks(datasetIdAsString, lockType, apiToken);
+    }
+    
+    static Response checkDatasetLocks(String idOrPersistentId, String lockType, String apiToken) {
```
qqmyers, line 2260: Broad question - now that finalize is always asynchronous, it looks like these tests check that the publish call succeeds and the lock goes away, but is there anything to check that the async publish succeeded rather than failed? (E.g. test that a new version exists after the lock?).

landreev: That would be the proper way to handle it, yes.
It didn't look like we ever check anything other than the return http code of that publish call. But you are right, even that was more than what we are doing now. 
In many tests the success of the publish call is tested indirectly, because what the test does requires the dataset to be published. But I agree it would be prudent, to check on the version numbers before and after - or perform some other explicit checks.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
landreev: So maybe the same code 202, but something in the body to differentiate from the "workflow" state. As it was implemented, it was returning the code 202 and `{"status":"WORKFLOW...}` in both cases - which was wrong.
It feels like this warrants its own issue - I'll open it.

landreev:
@qqmyers @scolapasta 
Adding the TransactionAttributeType.SUPPORTS to the dataset service method appears to have fixed it with the finalize command. I have moved the unlocking code back; and ran more testing. 
Once again, we'll have to look for similar situations elsewhere where commands are executed from service beans. Or even from the API beans that (for some reason?) have the `@Stateless` annotation.

scolapasta:
So is this ready for review again?

And in this scenario is there still a chance it gets unlocked before the execute transaction completes? It's probably ok for now, but I wonder if we need to still somehow be after the execute transaction completes (but not onsuccess because we need it to happen even if there's a failure.

landreev:
Seeing how it's the very last thing in execute() - I don't think so? I mean, not in any way that could encourage or enable a user to try to publish the dataset again, or to edit it etc. while the original transaction is still being executed; in a practical situation. 

> And in this scenario is there still a chance it gets unlocked before the execute transaction completes? It's probably ok for now, but I wonder if we need to still somehow be after the execute transaction completes (but not onsuccess because we need it to happen even if there's a failure.

landreev:
> So is this ready for review again?

I think so. I'm going to open a new issue for the other thing that was brought up. Changing the return code of the publish API; and the handling of it in RestAssured, with some added checks on the success of the operation. That's 2 things really - but I believe they should be handled in the same issue.

landreev:
Of course one test failed on the last full jenkins run: 
```
edu.harvard.iq.dataverse.api.FileMetadataIT.testJsonParserWithDirectoryLabels
Multiple Failures (2 failures)
	java.lang.AssertionError: Expected status code <201> doesn't match actual status code <500>.

	java.lang.AssertionError: Expected status code <200> doesn't match actual status code <404>.
```
is this anything we've seen before, as an intermittent thing?

landreev:
Needless to say, it passes on my own system consistently. 
If I need to insert a 1 second sleep in there, I will.

scolapasta:
I have not seen those before. If you'd like I can spin up an instance of two to see if it's consistent.

landreev:
We can just keep running it at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/view/change-requests/job/PR-7165

The test doesn't try to publish any datasets. 

> I have not seen those before. If you'd like I can spin up an instance of two to see if it's consistent.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -780,21 +780,37 @@ public WorkflowComment addWorkflowComment(WorkflowComment workflowComment) {
...
         // for a few seconds, just in case, to make sure the database update of 
         // the dataset initiated by the PublishDatasetCommand has finished, 
         // to avoid any concurrency/optimistic lock issues. 
+        // Aug. 2020/v5.0: It MAY be working consistently without any 
+        // sleep here, after the call the method has been moved to the onSuccess()
+        // portion of the PublishDatasetCommand. I'm going to leave the 1 second
+        // sleep below, for just in case reasons: -- L.A.
         try {
-            Thread.sleep(15000);
+            Thread.sleep(1000);
```
scolapasta, line 803: Did you try it without this? I'd love to get this removed, if we don't actually need it.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
```diff
@@ -196,6 +199,24 @@ private void verifyCommandArguments() throws IllegalCommandException {
...
+    @Override
+    public boolean onSuccess(CommandContext ctxt, Object r) {
+        Dataset dataset = null;
+        try{
+            dataset = (Dataset) r;
+        } catch (ClassCastException e){
+            dataset  = ((PublishDatasetResult) r).getDataset();
+        }
+
+        if (dataset != null) {
```
scolapasta, line 214: I guess this is just to be extra safe, but would this ever be false? (since we're already in OnSuccess, that means the execute succeeded, no?

qqmyers:
@landreev - not sure about the 500 error but the 404 could be from things now being asynchronous - the new dataset isn't guaranteed to be there by the next call anymore so could get a 404.

landreev:
> @landreev - not sure about the 500 error but the 404 could be from things now being asynchronous - the new dataset isn't guaranteed to be there by the next call anymore so could get a 404, and it could be a race - sometimes it works.

@qqmyers Except none of the datasets used in FilemetadataIT appear to be getting published. Unless it's done in some hidden indirect way that I'm not seeing... but I don't think so. (?)

Well, it worked on the next try, after the last commit. Run number 3 is blue again.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
landreev: I guess in case something else still goes wrong, in some other way? - idk.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
landreev: Can I leave it in place? 
I initially removed it completely; but I'm just afraid to take it out. Publishing appears to be working (at least I haven't seen it fail yet). But I believe if it happens instantly/very fast, on small datasets w/ fast registration, etc., it can confuse the autorefresh on the page.

qqmyers:
Good point - I was thinking create but that's not async/wasn't changed. Your change of the log text must have done it!

PR #7170, opened 2020-08-10
7134 errorhandlers refactored - code only

### LINKED ISSUE #7134: Error Handling: Multiple server.log errors indicating http 406 Not Acceptable

poikilotherm:
NOTE: this is a successor to PR #7136 as requested by @scolapasta, containing the code only.

- This refactors the API error handlers to be more relaxed about non-special-treatment Jersey exceptions.
- It also removes boilerplate and duplicated code from the package.
- It takes care of a standardized response format for errors (which should be adapted in other places to be consistent throughout our beloved API).
- It streamlines the logging of exceptions by enabling debugging and proper parsing in log analysis.

Closes #7134

I do hope the scope is not to big. I did not change the remaining two exception handlers for JSON and constraints to keep it smaller. Happy to change those if you think that makes sense.

One might also argue that special messages for Error 403 and 405 are not of much use. Happy to delete those.

I could also switch the Jersey Error 500 message to be the same as in ThrowableHandler.

The 302 and 307 were added because sometimes you might want to see a proper message via API, for example when you don't auto-follow redirects with `curl` et al.

I'm not sure if we should add a comment about the logging to the docs.

After deployment:
- Try debugging verbosity: `asadmin set-log-levels edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler=FINEST` and go for a 404 error
- Not sure how to generate a 500 error...
- A 404, 405 and 406 are easy to test
- A 302/307 is not commonly used in Dataverse. The only place is for ZIP downloads. So to test this, you'll have to enable those. I created a unit test for this during development, but deleted it as it wasn't of that much use...
- No idea how to trigger a 503 error in Jersey...

Ah API isn't classic UI, so let's say no.

Nope.

**Additional documentation**:

None.


kcondon:
@poikilotherm I'm still seeing the 406 errors:
[2020-08-10T20:00:42.722+0000] [Payara 5.201] [SEVERE] [] [org.glassfish.jersey.server.ServerRuntime$Responder] [tid: _ThreadID=97 _ThreadName=http-thread-pool::jk-conn
ector(2)] [timeMillis: 1597089642722] [levelValue: 1000] [[
  An exception has been thrown from an exception mapper class edu.harvard.iq.dataverse.api.errorhandlers.WebApplicationExceptionHandler.
        at org.glassfish.json.JsonObjectBuilderImpl.validateValue(JsonObjectBuilderImpl.java:198)
        at org.glassfish.json.JsonObjectBuilderImpl.add(JsonObjectBuilderImpl.java:66)
        at edu.harvard.iq.dataverse.api.util.JsonResponseBuilder.requestContentType(JsonResponseBuilder.java:118)
        at edu.harvard.iq.dataverse.api.errorhandlers.WebApplicationExceptionHandler.toResponse(WebApplicationExceptionHandler.java:76)
        at edu.harvard.iq.dataverse.api.errorhandlers.WebApplicationExceptionHandler.toResponse(WebApplicationExceptionHandler.java:24)
        at org.glassfish.jersey.server.ServerRuntime$Responder.mapException(ServerRuntime.java:528)
        at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:403)
        at org.glassfish.jersey.server.ServerRuntime$1.run(ServerRuntime.java:261)
        at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
        at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
        at org.glassfish.jersey.internal.Errors.process(Errors.java:244)

Not sure what is causing it but might be in export since I had just published the dataset. I was also just creating a new one but had not saved yet, just opened the edit form.

kcondon:
@poikilotherm In the spirit of not following instructions and acting randomly, I have started creating a dataset in the UI, modified the host dv so it is invalid and not completed any required fields other than title, then saved. Big stack traces in log, ui kind of stuck in grayed out view. On Demo, form validation catches it and presents error to user. Not sure whether it is related to these changes or not.

kcondon:
@poikilotherm After discussing with Jim, I understand these error handlers are for REST api only and that the UI, though it may call same underlying code, does not follow the same error handling path. So, I will focus on API errors here and maybe open a new issue for UI behavior I'm seeing.

poikilotherm:
Thx @qqmyers for handling this. Was a bit late yesterday for me... :wink:

I'm fine with returning a JSON null value, although it's debatable if one would rather exclude the field from the response. Let's stick with it as is now :+1: 

@kcondon are any problems remaining with the 406 errors or all set now? (I assume as you merged, but wanted to make sure everything is fine now...)

kcondon:
@poikilotherm Everything is set for now, thanks.

Issue #6766, opened 2020-03-25
Extend the metrics API with  functionality to get subverse specific metrics


PaulBoon:
Our clients want to get metrics only for their 'institutional' subverse, like total number of datasets, files and storage size within their subverse.  With the current API this is not possible because the metrics is only for the whole dataverse instance including all it's verses.

PaulBoon:
At DANS we have implemented this in our fork and are going to make this into a PR; code is at: 
https://github.com/DANS-KNAW/dataverse/blob/dans-develop/src/main/java/edu/harvard/iq/dataverse/api/MetricsDans.java
and 
https://github.com/DANS-KNAW/dataverse/blob/dans-develop/src/main/java/edu/harvard/iq/dataverse/metrics/MetricsDansServiceBean.java

djbrooke:
@PaulBoon thanks. We'll be happy to take a look at a PR but I believe that the Dataverse size API under /dataverses can be passed a specific Dataverse (http://guides.dataverse.org/en/latest/api/native-api.html#report-the-data-file-size-of-a-dataverse). I'll defer to @scolapasta or another developer about the desired implementation here, but I don't think we'd want one API under /metrics and another under /dataverses.

Also, are there any UI changes here?

PaulBoon:
The extension API has a lot of endpoints with return json with much more than a simple count. 
I know this code needs some reverse engineering to produce decent documentation, but the basic url contains the alias of dataverse you want metrics on
https://$SERVER/api/info/metrics-dans/{topLevelDvAlias}/
and then it is followed by: 

- dataverses
- dataverses/addedOverTime
- dataverses/byCategory
- datasets
- datasets/bySubject
- datasets/addedOverTime
- files
- tree
- dataverses-report
- datasets-report
- downloads

It is based on the notion that we have a  hierarchy of sub-verses.

djbrooke:
Thanks @PaulBoon, I understand you're adding a lot of additional functionality. Unless there are objections from the technical team, we'll just plan to wait on the PR.

donsizemore:
@PaulBoon thank you for this. our most common requests are for file download count by year and number of datasets published per year. Is the latter included in datasets-report? If not, I can open another issue after your PR.

qqmyers:
Re: 1 above - and the MakeDataCount metrics are different again, providing a count for a given month specified in the call, but nothing cumulative.

qqmyers:
[...]

PR #7178, opened 2020-08-11
IQSS/7177 enhance metrics api

### LINKED ISSUE #7177: Metrics API enhancement
Multiple metrics-related issues (such as #6766, #3313, #3527) envision per-dataset metrics. The existing metrics APIs (api/info/metrics and /admin/makeDataCount) do not yet support this and differ in other ways (e.g. how a time-series is reported, error reporting, published or draft datasets, whether objects or sql queries are used).

As part of a QDR effort to implement metrics reporting, I've started trying to standardize aspects of these APIs and, drawing on the code from DANS mentioned in #6766, to add per-dataverse capabilities. I'm submitting a draft PR to make the current state of this work visible to the community and am looking for any/all feedback as to how this can be adapted/extended to address related metrics issues and other community needs.

qqmyers:
**What this PR does / why we need it**: Enhances and standardizes the design of metrics api endpoints

Closes #7177

**Additional documentation**:


src/main/resources/db/migration/V5.0.0.1__7177-newmetrics.sql
```diff
@@ -0,0 +1,13 @@
+
+alter table metric add column if not exists dataverse_id bigint;
+
+DO $$
+BEGIN
+
+  BEGIN
+    alter table metric add constraint fk_metric_dataverse_id FOREIGN KEY (dataverse_id) REFERENCES dataverse(id);
+  EXCEPTION
+    WHEN duplicate_object THEN RAISE NOTICE 'Table constraint foo.bar already exists';
```
pdurbin, line 10: I assume this isn't supposed to say "foo.bar" but I doubt it's a big deal.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1901,4 +1903,16 @@ public static boolean isFileAlreadyUploaded(DataFile dataFile, Map checksumMapNe
         return false;
     }
 
+    public static String jsonToCSV(JsonArray jsonArray, String... headers) {
```
pdurbin, line 1906: This method doesn't handle arbitrary JSON. Maybe rename to flatJsonToCsv? And add some docs on when to use it?

src/main/java/edu/harvard/iq/dataverse/api/Metrics.java
```diff
@@ -26,404 +41,836 @@
...
             }
         }
-        
+
     }
-    
+
+    // Throws a WebApplicationException if alias is not null and Dataverse can't be
+    // found
+    private Dataverse findDataverseOrDieIfNotFound(String alias) {
```
pdurbin, line 850: I'm used to the "orDie" methods working including the original findDataverseOrDie to work in certain way, to work on unpublished things, to throw a WrappedResponse when it can't be found. Since this method only returns published dataverses maybe it could be renamed? This is just a passing thought. I can also live with it with the current name.

pdurbin (commented):
Overall, I'm very excited to see continued development of the Metrics API and new vizualizations being built on top of it. I see that useful, real-world metrics from the Dataverse fork maintained by DANS (#6766) are being folded in. It's nice that multiple formats are supported (CSV/JSON) and I understand that an effort has been made to make metrics-related APIs more consistent with each other.

My two biggest concerns are backward incompatibility and documentation. I don't know if there is backward incompatibility or not. There were no existing tests, so it's hard to tell, but I'm seeing both "toMonth" and "monthly" in the code, for example and if anyone's application is going to break because something has changed, we should absolutely add a release note to this pull request to enumerate all the breakages. Even better would be to remain backward compatible to avoid any breakage in the first places.

Regarding documentation, I'm seeing exciting new API endpoints like "tree" but nothing about it in the API Guide. There seems to be a new choice between CSV and JSON. Great. Please document it for the users, for QA, etc.

Should this pull request close #6766? Does it contain all or most of the metrics code DANS wrote?

I'd like to note that the pull request itself is large so I'm sure I'm missing things. Also, I'm not a SQL expert so if there are doubts about the queries, someone else should review it.

See also my review on the related pull request for dataverse-metrics: https://github.com/IQSS/dataverse-metrics/pull/52

qqmyers:
Thanks @pdurbin !

The only backward incompatibility that has been found is on the few calls that can now produce csv or json - one must now explicitly request application/json to get it whereas that was the default/only option before. The recent updates to the dataset-metrics app handle this there. (Needed to get QDR metrics for the global count because QDR has these endpoints deployed). The reason that I flipped the default was so that getting the metrics URLs in a browser would get spread-sheet ready data rather than json to convert, figuring that API users could specify json (if they weren't doing so already). If that bit of convenience is not worth the inconsistency, the default could be flipped. (Note that API callers who do request application/json will be compatible with earlier Dataverse versions since they only serve json.)

That, and any other endpoints can certainly be documented before this is merged, but I wanted to make sure there aren't significant changes request before trying to bring docs up-to-date.

src/main/java/edu/harvard/iq/dataverse/api/Metrics.java
qqmyers: Not sure I understand why this doesn't look like an orDie method - if the alias requested doesn't exist, it throws an exception (a WebApplicationException rather than WrappedResponse to use the updated error handling that's been added). Is it what is returned that is an issue or something else? Or just findPublishedDataverseOrDie ?

src/main/resources/db/migration/V5.0.0.1__7177-newmetrics.sql
qqmyers: :-) - caught at cut/paste !

djbrooke:
Thanks @qqmyers for responding to feedback and getting this into a state where it can reviewed again. I see the note above about adding documentation about the backward incompatible changes once the code is set - I'm supportive of that and will be happy to help. Sounds like the next step is additional code review.

qqmyers:
@djbrooke - thanks! FWIW: the discussion about documentation was prior to https://github.com/IQSS/dataverse/pull/7178/commits/4caed91f766582012a47a5308f1cfd906c501108 . Assuming that we also have release notes for this, adding a note that "users of the current metrics API endpoints should specifically request an 'Content-Type:application/json' response to avoid getting the new CSV format." could cover the backward compatibility issue.

djbrooke:
Thx @qqmyers - I'll add this and move to QA.

pdurbin (changes requested):
I believe I've found five things that need fixing:

- flyway script rename (error below)
- guides won't build (error below)
- guides look odd (screenshot below)
- missing release note (especially about backward compatibility)
- merge latest from develop (after 5.4 is released)

## Flyway script rename

Currently, the app won't deploy due to this error

Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 5.0.0.1
Offenders:
-> /usr/local/payara5/glassfish/domains/domain1/applications/dataverse-5.3/WEB-INF/classes/db/migration/V5.0.0.1__6872-assign-storage-drivers-to-datasets.sql (SQL)
-> /usr/local/payara5/glassfish/domains/domain1/applications/dataverse-5.3/WEB-INF/classes/db/migration/V5.0.0.1__7177-newmetrics.sql (SQL)

## Guides won't build

Warning, treated as error:
/Users/pdurbin/github/iqss/dataverse/doc/sphinx-guides/source/api/metrics.rst:65: WARNING: Bullet list ends without a blank line; unexpected unindent.

## Guides look odd

Maybe this is just me but this looks a bit odd with the larger type and indentation:

Screen Shot 2021-03-31 at 1 52 55 PM


## Missing release note

Danny offered to add this. It's especially important to know the backward incompatible change that requires the metrics app to be updated: https://github.com/IQSS/dataverse-metrics/pull/52

## Merge latest from develop

This should be done after 5.4 is updated in pom.xml and elsewhere.

qqmyers:
This is up-to-date w.r.t. 5.4 including the merge and updating the flyway script. I found an extraneous bit of text in the guide file that was causing the build error. I'm not sure if that also fixes the text size issue. If not, it's beyond my .rst knowledge to adjust the size - if anyone can point to an example of how to structure such a list, I'd be happy to apply it.

W.r.t. to backward compatibility - I just added a 6 line PR in the dataverse-metrics repo that adds the now required Accept:application/json header that would be needed to keep the an existing metrics UI deployment working with Dataverse instances with the PR here. There's also the larger PR over in dataverse-metrics that creates all the new graphs that this PR supports from the Dataverse side which includes the same fix. Depending on what gets merged there (and into what release) by the time this PR is released, the release notes could say something like requiring an update to 0.2.8 to continue retrieving global metrics and updating to 0.3.0 to enable the new per Dataverse collection metrics UI for the local installation.

pdurbin:
@qqmyers you're probably in the best position to add a release note. If you could do that it would be most appreciated.

pdurbin:
> it's beyond my .rst knowledge to adjust the size

Yeah, me too. I just pushed a "good enough" fix in 1e4a4cf49 after getting the go ahead from @qqmyers . To make the text size reasonable, I converted the sub items to a bulleted list as in the "Total" example below.

Screen Shot 2021-05-10 at 11 27 31 AM

Longer term we should probably try to figure out what's going on so people who are trying to write docs like this don't need bullets. I looked at https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#indentation but couldn't get anything better working.

pdurbin (approved):
I haven't run the code myself but this seems to be ready for QA.

PR #7180, opened 2020-08-11
#7176 add testing for null host DV on DS Create

### LINKED ISSUE #7176: Create Dataset: Form validation not working on save, throws stack trace, gray screen
This works in 4.20, doesn't work in current dev.
Some code investigation (thanks jim) indicates it likely changed here: https://github.com/IQSS/dataverse/pull/6909

Steps to reproduce:
1. create dataset, delete host dv, enter title, leaving rest of required fields empty, save
2. notice null ptr errors in server log, UI has gray overlay on dataset form, no validation errors. in 4.20 on demo, throws normal missing entry validation errors.

sekmiller:
For recent button updates we moved some of the render logic to the DatasetPage bean. This logic assumes that the dataset has an owner (host dataverse). if on create the user decides to null out the host dataverse this logic breaks and the page fails to render. This PR updates the render logic to guard against a null host dataverse.

Closes #7176 Create Dataset Validation fails etc.

No
No

**Additional documentation**:
None

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -1270,6 +1270,9 @@ public class DatasetPage implements java.io.Serializable {
         return permissionsWrapper.canUpdateDataset(dvRequestService.getDataverseRequest(), this.dataset);
     }
     public boolean canPublishDataverse() {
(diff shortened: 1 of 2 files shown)


sekmiller:
OK. I'll try to handle it that way.

sekmiller:
Abandoning in favor of another approach

PR #7181, opened 2020-08-11
The sample ejb-jar.xml file had a bad version number that would preve…

### LINKED ISSUE #7175: Tweak timer doc
Although I still think using ejb-jar.xml deployment descriptors to reconfigure automatic (@Schedule) timers should be a rare one-off, because of the complexity it re-introduces in Java EE, now that it's out in public documentation, I want to make sure the docs are as accurate and clear as possible, while we're thinking of it.

There's an additional way to besides compiling the app to inject an edited ejb-jar.xml file into the application.

Also when I was assembling the documentation, I inserted the wrong sample file, which had a version problem, which would prevent the dataverse war file from deploying if the ejb-jar.xml file was inserted. I discovered this while re-testing some things after some issues had come up with the merge

rtreacy:
…nt deployment of the application if the ejb-jar file was used to modify the saved searchers timer schedule.

I updated the timer.rst file to document a procedure for inserting an edited ejb-jar.xml file into a war, rather than doing a full build of the application

**What this PR does / why we need it**:The sample file cannot be used with the wrong version number

**Which issue(s) this PR closes**:Bad version number in sample ejb-jar.xml; document procedure for updating the war file to include the ejb-jar.xml outside of a build environment

Closes #7175 

**Suggestions on how to test this**: You can test the procedure documented at the end of the timer.rst file. The description of this procedure starts with "Alternatively, you can insert an ejb-jar.xml ..."

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:no

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/admin/timers.rst
+++ b/doc/sphinx-guides/source/admin/timers.rst
@@ -69,25 +69,25 @@ This job is automatically scheduled to run once a week at 12:30AM local time on 
 
 * Alternatively, you can insert an ejb-jar.xml file into a provided dataverse war file without building the application.
 
-    * Check if there is already an ejb-jar.xml file in the war file 
+  * Check if there is already an ejb-jar.xml file in the war file 
 
-        * jar tvf $DATAVERSE-WAR-FILENAME | grep ejb-jar.xml
+    * jar tvf $DATAVERSE-WAR-FILENAME | grep ejb-jar.xml
 
-            * if the response includes " WEB-INF/ejb-jar.xml", you will need to extract the ejb-jar.xml file for editing
+      * if the response includes " WEB-INF/ejb-jar.xml", you will need to extract the ejb-jar.xml file for editing
 
-                * jar xvf $DATAVERSE-WAR-FILENAME WEB-INF/ejb-jar.xml 
+        * jar xvf $DATAVERSE-WAR-FILENAME WEB-INF/ejb-jar.xml 
 
-                    * edit the extracted WEB-INF/ejb-jar.xml, following the :download:`sample file <../_static/admin/ejb-jar.xml>` provided.
+          * edit the extracted WEB-INF/ejb-jar.xml, following the :download:`sample file <../_static/admin/ejb-jar.xml>` provided.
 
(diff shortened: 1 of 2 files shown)


doc/sphinx-guides/source/admin/timers.rst
```diff
@@ -59,14 +59,37 @@ This job is automatically scheduled to run once a week at 12:30AM local time on
...
 
   * The provided parameters in the sample file are <minute>, <hour>, and <dayOfWeek>; additional parameters are available
 
     * For a complete reference for calendar expressions that can be used to schedule Timer services see: https://docs.oracle.com/javaee/7/tutorial/ejb-basicexamples004.htm
 
 * Build and deploy the application
 
+* Alternatively, you can insert an ejb-jar.xml file into a provided dataverse war file without building the application.
+
+    * Check if there is already an ejb-jar.xml file in the war file (as I'm writing, the dataverse war file is named dataverse-4.20.war)
```
scolapasta, line 72: Since this will be for the Dataverse 5 release, we should change this to that - I think you can also jsut say e.g. instead of "as I'm writing"

rtreacy: like 'Check if there is already an ejb-jar.xml file in the war file, e.g dataverse-5.war'   and same wherever referenced in the commands?

kcondon:
Confirmed the instructions work, thanks Bob. Passing over to Danny/Mike for doc formatting clean up.

djbrooke:
Thanks @kcondon and @rtreacy... @mheppler do you want to take the first stab or should I?

rtreacy:
One thing I noticed when I looked back at what Kevin showed me, is that aside from the enlarged text, which I have no idea where that comes from, even after reading the basic sphinx documentation, there also a logical formatting error, which I could fix, but since you guys have it now, maybe you could take care of.

The last 3 lines of my instructions, eg 
Insert...
    jar...
Deploy...
should each be moved to the left, so that Insert and Deploy are at the same indentation as Check

mheppler:
@kcondon @djbrooke @rtreacy -- fixed the bulleted list formatting, too many indent spaces.

Screen Shot 2020-08-12 at 2 26 26 PM

Issue #5615, opened 2019-03-07
Add "Add Data" to navbar, plus Host Dataverse input to create dataverse/dataset forms


TaniaSchlatter:
One way to have people create more dataverses is to make it easier to see how to do it. People who get to dataverse from google and bypass the home page or search results page don't see a clear way to add data. 

One solution: adding a button to the header on every page will help make it clear how to add to Harvard Dataverse. 

**_Adding a button. The button has a dropdown to select a new dataverse or new dataset, as current Add Data functionality from the buttons on the home page:_**
![add_data_header_loggedin](https://user-images.githubusercontent.com/15386424/53977593-5b92bb80-40d7-11e9-8be4-e45ea96d132a.jpg)

If users are not logged in, the Log in/Sign in popup with a link to sign up (issue #6170) is shown. 
Proposed flow:
![Add_data_header_flow](https://user-images.githubusercontent.com/15386424/65626305-46aa4680-df9b-11e9-89d9-0fe8af968e31.jpg)

**_Removing the "Add Data" button from the search results page, and showing it on the header for not logged in users:_** 
![add_data_header_search_nologin](https://user-images.githubusercontent.com/15386424/53977675-8ed54a80-40d7-11e9-8834-aa7b18e75938.png)

TaniaSchlatter:
Adding the ability to add data from the header in an expected way means that users need to be able to select the dataverse and dataset that they want to  deposit data to or create a dataverse in.

djbrooke:
@TaniaSchlatter I moved this over to the design column on the board. Feel free to move back to inbox if this is not the correct status.

TaniaSchlatter:
Related issue #5493

TaniaSchlatter:
Docs that break down:

- [Current flow for adding data and linking](https://docs.google.com/presentation/d/1hMOoMmM0wRc1VGXQkEJ_HoAnkF_lzwVL6gecMHVvJbg/edit?usp=sharing)

- [Additional ideas for adding data and linking](https://docs.google.com/presentation/d/1Vj6LSYu-s8_nPMhlGKYvGhp_4KMrITMKjlWmtYUvoyo/edit?usp=sharing)

pdurbin:
Related: As the creator of a dataverse I'd like to know how to make an "Add Data" button appear for non-logged in users to encourage user interaction #5718

djbrooke:
- Consider a configuration option for whether or not it displays on the custom homepage
[...]

PR #7197, opened 2020-08-13
6798 create dv validation

### LINKED ISSUE #6798: Create Dataverse - Inline form validation errors missing proper formatting
If you click ~add then remove (with the new email address blank) on the email section~ Save Changes with empty required fields on the dataverse create page you get a unformatted validation error msgs. Also, as you can see in the screenshot below, the Dataverse Name validation error msg is missing altogether.

Screen Shot 2020-07-09 at 12 08 56 PM

sekmiller:
**What this PR does / why we need it**: Fixes the display of validation messages when required fields are left blank on Dataverse Create (presumably Edit as well)

Closes #6798 Create Dataverse inline form validations lack proper formatting.

No
None

**Additional documentation**:


src/main/webapp/dataverse.xhtml
```diff
@@ -110,7 +110,8 @@
                                           data-toggle="tooltip" data-placement="auto right" data-original-title="#{bundle['dataverse.name.title']}"></span>
                                 </h:outputLabel>
                                 <div class="form-col-container">
-                                    <p:inputText id="name" styleClass="form-control" value="#{DataversePage.dataverse.name}" required="true"/>
+                                    <p:inputText id="name" styleClass="form-control" value="#{DataversePage.dataverse.name}" />
```
scolapasta, line 113: How was this not there before??

src/main/webapp/dataverse.xhtml
```diff
@@ -135,7 +136,7 @@
                                 <div class="form-col-container">
                                     <div class="input-group">
                                         <div class="input-group-addon"><span class="small">#{systemConfig.dataverseSiteUrl}/dataverse/</span></div>
-                                        <p:inputText id="identifier" styleClass="form-control" validator="#{DataversePage.validateAlias}" value="#{DataversePage.dataverse.alias}" required="true"/>
+                                        <p:inputText id="identifier" styleClass="form-control" validator="#{DataversePage.validateAlias}" value="#{DataversePage.dataverse.alias}" />
```
scolapasta, line 139: Why remove required on this (and other) fields?

src/main/webapp/dataverse.xhtml
mheppler: These required attributes were adding in 4.18 for issue #5615 in an attempt to improve accessibility of the forms.

src/main/webapp/dataverse.xhtml
sekmiller: The required is handled by the @NotBlank bean validation on the fields in Dataverse. All of them are annotated with NotBlank

src/main/webapp/dataverse.xhtml
sekmiller: I can take the approach of trying to get the required message from the ValidationMessages.properties file and see if that works.

sekmiller: Looks like it was just a miss. No one leaves the name blank? or something.

mheppler: The attributes do not appear necessary for the validation to function, but it is in order to put an accessibility friendly `aria-required="true"` in the input component code, as the red asterisk visual indicator is not sufficient.

scolapasta: @mheppler So do all the others ones need to have it added back?

mheppler: I believe so. I tried using `pt:aria-required="true"` in the p:input component to get the attribute in there another way, but it was just ignored.

sekmiller: I returned all of the required='true' where applicable. I am now getting the requiredMessage from the Bundle because the ValidationMessages.properties are only used by the Bean Validations

src/main/webapp/dataverse.xhtml
sekmiller: again, returned the required=true based on Mike's need of them for accessibility.

scolapasta (approved):

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -2116,6 +2116,7 @@ dataverse.search.user=Only authenticated users can save a search.
 dataverse.alias=alias
 dataverse.alias.taken=This Alias is already taken.
 
+
```
mheppler, line 2119: @sekmiller Looks like the only change here is an extra return. We should resort that change.

src/main/java/ValidationMessages.properties
```diff
@@ -47,3 +48,5 @@ password.current=Please enter your current password.
 password.validate=Password reset page default email message.
 
 guestbook.name=Enter a name for the guestbook
+
```
mheppler, line 51: @sekmiller These extra returns at the bottom are not needed either.

sekmiller: Ok. will do.

src/main/java/propertyFiles/Bundle.properties
sekmiller: yup.

PR #7211, opened 2020-08-19
Add API methods for reporting and cleaning up inconsistencies between the database and Solr Index

### LINKED ISSUE #4225: Indexing: Create API endpoint(s) that allows an admin to detect/delete stale solr record
Although probably rare, see #4205, there may be orphaned solr records that need to be removed outside of Dataverse, directly from the Solr API. It would be more convenient if an admin could use the familiar Dataverse API to pass the identifier of a stale dataset record to be removed.

ekraffmiller:
Updated existing API method for detecting orphan objects in the database, and in Solr. Added a new API method for clearing all orphans. 
(#4225)

There was an existing Index API method, called status, which reported all the inconsistencies between the database and Solr Index. I updated this method to report the object ids, rather than the count of objects, and also implemented the "exist in Solr but not in the database" type of orphan. 

Updated the code to use a Solr query cursor, to improve performance & memory management.

Closes #4225 

I tested this in my local environment by creating a dataverse and dataset, shutting down solr, then deleting the data when solr was shutdown.


ekraffmiller:
Ok, will update the docs, and look into adding testing for the IndexServiceBean (since it failed the coveralls check), even though we decided not to do an integration test.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -3061,6 +3061,20 @@ Note that if you are attempting to validate a very large number of datasets in y
...
+
+Get a list of all database objects that are missing in Solr, and Solr documents that are missing in the database::
+
+  curl $SERVER_URL/api/admin/index/status
+
+Remove all Solr documents that are orphaned (ie not associated with objects in the database)::
+
+  curl $SERVER_URL/api/admin/index/clear-orphans
+
+
```
pdurbin, line 3077: These docs look fine but I would prefer seeing them on this page: http://guides.dataverse.org/en/latest/admin/solr-search-index.html

It's nice having a single "clear-orphans" endpoint but is it fast enough? Will there be demand for a "clear dataset orphan"? I'm fine with putting "clear-orphans" out into the world and getting feedback about if more granularity is needed.

doc/release-notes/4225-stale-solr-records.md
```diff
@@ -0,0 +1,7 @@
+## Major Use Cases
+
+- Administrators will be able to use new APIs to keep the Solr index and the DB in sync, allowing easier resolution of an issue that would occasionally cause search results to not load. 
+
+## Notes for Dataverse Installation Administrators
+
+New API endpoints... point to Guides
```
pdurbin, line 7: We should fill this in before it goes to QA. If my suggestion below is followed, we would point to http://guides.dataverse.org/en/latest/admin/solr-search-index.html

src/main/java/edu/harvard/iq/dataverse/api/Index.java
```diff
@@ -384,15 +384,16 @@ public Response indexStatus() {
         JsonObjectBuilder contentInDatabaseButStaleInOrMissingFromSolr = getContentInDatabaseButStaleInOrMissingFromSolr();
 
         JsonObjectBuilder contentInSolrButNotDatabase;
+        JsonObjectBuilder permissionsInSolrButNotDatabase;
         try {
             contentInSolrButNotDatabase = getContentInSolrButNotDatabase();
-        } catch (SearchException ex) {
+            permissionsInSolrButNotDatabase = getPermissionsInSolrButNotDatabase();
+ } catch (SearchException ex) {
```
pdurbin, line 391:
```suggestion
        } catch (SearchException ex) {
```

src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
```diff
@@ -532,12 +528,15 @@ public JsonObjectBuilder deleteAllFromSolrAndResetIndexTimes() throws SolrServer
     }
 
     /**
-     * @todo Do we want to report the root dataverse (id 1, often) in
-     * permissionsInDatabaseButMissingFromSolr?
+     * 
      *
      * @return A list of dvobject ids that should have their permissions
      * re-indexed Solr was down when a permission was added. The permission
```
pdurbin, line 534:
```suggestion
     * re-indexed because Solr was down when a permission was added. The permission
```

src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
```diff
@@ -547,7 +546,7 @@ public JsonObjectBuilder deleteAllFromSolrAndResetIndexTimes() throws SolrServer
             Timestamp permissionModificationTime = dvObject.getPermissionModificationTime();
             Timestamp permissionIndexTime = dvObject.getPermissionIndexTime();
             if (permissionIndexTime == null) {
-                if (dvObject.getId() != rootDvId) {
+                if (dvObject.getId() != rootDvId && !dvObject.isInstanceofDataFile()) {
```
pdurbin, line 549: As a possible performance optimization, since we're skipping data files anyway, perhaps instead of the `dvObjectService.findAll()` above we could get just a list of dataverses and datasets. I don't know if it would help with performance or not, or if performance is even a problem.

src/test/java/edu/harvard/iq/dataverse/api/IndexIT.java
```diff
@@ -0,0 +1,122 @@
...
+                .statusCode(OK.getStatusCode());
+
+        Response deleteDataverseResponse = UtilIT.deleteDataverse(dataverseAlias, apiToken);
+        deleteDataverseResponse.prettyPrint();
+        deleteDataverseResponse.then().assertThat()
+                .statusCode(OK.getStatusCode());
+
+        Response deleteUserResponse = UtilIT.deleteUser(username);
+        deleteUserResponse.prettyPrint();
+        assertEquals(200, deleteUserResponse.getStatusCode());
```
pdurbin, line 110: I go back and forth on cleanup like this. One one hand it's nice not to junk up your dev environment with test data. On the other hand, if the delete fails and the whole test suite fails as a result, it's frustrating. One thing I sometimes do is put the delete calls in but don't assert that they succeed. No need to changes anything here. It's just something that popped in my head.

pdurbin (changes requested):
I added a few comments but the only thing I think we should definitely change is where the docs go. I think they should be in the Solr page next to related commands.

doc/sphinx-guides/source/api/native-api.rst
ekraffmiller: moving the documentation to search-index.html.

src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
ekraffmiller: talked about this and decided to see how performance is in production before making a change

src/test/java/edu/harvard/iq/dataverse/api/IndexIT.java
ekraffmiller: removing the assert tests from the delete calls.

pdurbin (approved):
Looks good as of abcc9d8

I haven't tried the code myself but in addition to testing basic functionality, I'd suggest running a "clear-orphans" on a production database to see how long it takes.

pdurbin:
@kcondon heads up the @ekraffmiller and I have been talking about the test failures on build 16 of this branch. Since build 15 passed and build 16 was simply merging the latest from develop, I don't think this branch adds any breakage. Here are the build 15 test results: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/view/change-requests/job/PR-7211/15/testReport/

I did note that one of the breakages in build 16, InReviewWorkflowIT.testCuratorSendsCommentsToAuthor, also appears in develop build 579: https://jenkins.dataverse.org/job/IQSS-dataverse-develop/579/testReport/

kcondon:
@ekraffmiller Passing back to review status feedback on longer running jobs.

pdurbin (approved):
The async code is in place so we're ready for more QA. I also made a tiny tweak to the docs to make the curl command consistent with other commands on the page.

PR #7214, opened 2020-08-21
5050 Parse all dc identifier elements and allow identifiers that don'…

### LINKED ISSUE #5050: Harvesting zenodo client fail
When I try to add zenodo (https://zenodo.org/oai2d) as harvesting client on dataverse 4.8.6, 4.9.1 and 4.9.2, I get the error :

"Failed to find a global identifier in the OAI_DC XML record"

JingMa87:
…t have "doi" or "hdl" in them.

**What this PR does / why we need it**: Allows dataverse to harvest more datasets

**Which issue(s) this PR closes**: 5050

Closes #5050 

**Special notes for your reviewer**: Small change, big impact!

**Suggestions on how to test this**: Harvest the server https://api.figshare.com/v2/oai with prefix oai_dc and set portal_895. Without the fix they would all fail but now they're all harvested.

**Is there a release notes update needed for this change?**: Yes

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
@@ -366,24 +366,35 @@ public class ImportGenericServiceBean {
     }
     
     private String getOtherIdFromDTO(DatasetVersionDTO datasetVersionDTO) {
+        List<String> otherIds = new ArrayList<>();
         for (Map.Entry<String, MetadataBlockDTO> entry : datasetVersionDTO.getMetadataBlocks().entrySet()) {
             String key = entry.getKey();
             MetadataBlockDTO value = entry.getValue();
             if ("citation".equals(key)) {
                 for (FieldDTO fieldDTO : value.getFields()) {
                     if (DatasetFieldConstant.otherId.equals(fieldDTO.getTypeName())) {
-                        String otherId = "";
                         for (HashSet<FieldDTO> foo : fieldDTO.getMultipleCompound()) {
                             for (FieldDTO next : foo) {
                                 if (DatasetFieldConstant.otherIdValue.equals(next.getTypeName())) {
-                                    otherId =  next.getSinglePrimitive();
+                                    otherIds.add(next.getSinglePrimitive());
                                 }
-                            }
-                            if (!otherId.isEmpty()){
-                                return otherId;
                             }
                         }
                     }
+                }
+            }
+        }
+        if (!otherIds.isEmpty()) {
+            // We prefer doi or hdl identifiers like "doi:10.7910/DVN/1HE30F"
+            for (String otherId : otherIds) {
+                if (otherId.contains(GlobalId.DOI_PROTOCOL) || otherId.contains(GlobalId.HDL_PROTOCOL)) {
+                    return otherId;
+                }
+            }
+            // But identifiers without hdl or doi like "10.6084/m9.figshare.12725075.v1" are also allowed
+            for (String otherId : otherIds) {
+                if (otherId.startsWith("10.") && otherId.contains("/")) {
+                    return GlobalId.HDL_PROTOCOL + ":" + otherId;
                 }
             }
         }
--- a/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
@@ -375,6 +375,7 @@ public class ImportGenericServiceBean {
             if ("citation".equals(key)) {
                 for (FieldDTO fieldDTO : value.getFields()) {
                     if (DatasetFieldConstant.otherId.equals(fieldDTO.getTypeName())) {
+                        String otherId = "";
                         for (HashSet<FieldDTO> foo : fieldDTO.getMultipleCompound()) {
                             for (FieldDTO next : foo) {
                                 if (DatasetFieldConstant.otherIdValue.equals(next.getTypeName())) {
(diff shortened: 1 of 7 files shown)


JingMa87:
@pdurbin @jggautier Ready for review!

jggautier:
Works great from what I can tell! Spun up the branch and just saw that records from sets from Zenodo and Figshare were all harvested successfully.

Tried harvesting ICPSR and about a third of the 10k records failed. Can't tell why, but I think another GitHub issue should be opened for this.

JingMa87:
@jggautier Great to hear! In the case of EASY there's forthcoming issues with the controlled vocabulary. The EASY datasets have dc language values of "en", fr", "de", but those are not allowed in Dataverse which supports values like "English", "German". What is the reason for this restriction actually?

jggautier:
@JingMa87 I think the reason for the restriction is that Dataverse hasn't implemented a method for adding to the metadata exports the 2-3 letter code of the language that the depositor chooses from the Language controlled vocabulary. Dataverse simply adds to the metadata export the value that the depositor sees in the metadata form:

<kbd>Screen Shot 2020-08-31 at 11 43 11 AM</kbd>

I think we agree that it would be preferable if Dataverse shows the depositor the full language, e.g. "English", but uses the corresponding ISO codes in the metadata exports, e.g. `<dc:language>en</dc:language>` for the OAI-PMH feed and `<dcterms:language>en</dcterms:language>` for the DC Terms metadata available through the metadata export dropdown and the API.

Then the other way around, on import, Dataverse will need to know that `<dc:language>en</dc:language>` means that the Language value it should display on the dataset page UI is "English" (for repositories in English)

Does internalization need to be considered? E.g. when dataset metadata is imported into an installation with "Spanish" internalization, `<dcterms:language>en</dcterms:language>` should be displayed on the dataset page as the localized value "Inglés"?

I know the community's discussed this before, but I can't find a GitHub issue about it. Could you create a GitHub issue (if you're not already planning to)?

JingMa87:
@jggautier Here's the issue: https://github.com/IQSS/dataverse/issues/7243. It shouldn't have an impact on this PR.

djbrooke:
@jggautier @JingMa87 - is this ready for Code Review? Apologies for the delay.

jggautier:
I think it is!

JingMa87:
@djbrooke Yes it is!

src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
```diff
@@ -366,27 +368,42 @@ private FieldDTO makeDTO(DatasetFieldType dataverseFieldType, FieldDTO value, St
...
-                            }
                         }
                     }
                 }
             }
         }
+        if (!otherIds.isEmpty()) {
+            // We prefer doi or hdl identifiers like "doi:10.7910/DVN/1HE30F"
+            for (String otherId : otherIds) {
+                if (otherId.contains(GlobalId.DOI_PROTOCOL) || otherId.contains(GlobalId.HDL_PROTOCOL)) {
```
landreev, line 392: Should this be `if (otherId.startsWith(GlobalId.DOI_PROTOCOL) ...` instead? 
Or maybe even `if (otherId.toLowerCase().startsWith(GlobalId.DOI_PROTOCOL) ...`?

landreev (changes requested):
I have made one small change request. Looks good otherwise!
Sorry again for the delay with this PR; we ended up ignoring some PRs and issues because of all the v5.0 related tasks.

landreev:
@jggautier 
>  ... Spun up the branch 
>  ... 
> Harvested ICPSR and about a third of the 10k records failed. Can't tell why, but I think another GitHub issue should be opened for this.

Just noticed the ICPSR part. ICPSR would be one archive from which we absolutely want to harvest DDI (and not DC). So we may not necessarily care why their DC records are failing to import (?).

src/main/java/edu/harvard/iq/dataverse/api/imports/ImportGenericServiceBean.java
JingMa87: @landreev The identifier could also be "https://doi.org/10.7910/DVN/1HE30F". I made the identifier into lowercase just in case.

landreev: @JingMa87 
Looking at your last commits, it's still looking like your code is doing "if ... contains()"... I still think it should be "if ... startsWith()" instead. Or it will just assume that any identifier that happens to contain the characters "hdl" is a handle, no? 
And yes, it's a good idea to check for the "https://doi.org/10.7910/DVN/1HE30F" form as well. Please note that we also have GlobalId.DOI_RESOLVER_URL and GlobalId.HDL_RESOLVER_URL defined. So maybe add .startsWith() for these too?

JingMa87: Agreed, I pushed the changes.

landreev: I'm really sorry to be difficult, I know you have other things to work on - but do we really want to *permanently* convert to lower case? 
I only suggested it for the test... to be able to catch both "hdl:..." and "HDL:..." - but I'm not sure even that is necessary... 
I don't think we want to save "doi:10.7910/DVN/1HE30F" as "doi:10.7910/dvn/1he30f" - ?
Let's just remove the `otherId = otherId.toLowerCase();` line. 
Thank you!

JingMa87: No worries! I made the changes.

landreev (approved):

Issue #3577, opened 2017-01-18
Documentation: When content is wide enough to require horizontal scroll, content scrolls but left nav outline does not.


kcondon:
Example: Go to Shibboleth section or R/Two Ravens sections, scroll down to where a line is long enough to enable horizontal scrollbar and then scroll horizontally, notice content scrolls on top of nav outline block on left. If you scroll horizontally when such a line is not present (headers seem to allow horizontal scrolling due to white space), then both content and nav outline block scroll as a single piece.

http://guides.dataverse.org/en/latest/installation/r-rapache-tworavens.html
http://guides.dataverse.org/en/latest/installation/shibboleth.html

This exists in 4.6 but was found when testing Oauth in 4.6.1.

pdurbin:
@dlmurphy and I just talked about this. Here's how it looks:

![screen shot 2017-06-06 at 2 35 01 pm](https://user-images.githubusercontent.com/21006/26845703-5e212576-4ac5-11e7-82a6-881d7f783241.png)

The cause is a long line of code. To trigger the bug, scroll horizontally.

dlmurphy:
This issue came up today because @bsilverstein95 was trying to use a literal to show a line of code that was long enough to extend the page width. It was impossible for a reader to see the entire length of the line of code without breaking the page, but including any line breaks in that code would break it when the reader copied and pasted it.

mheppler:
Found this writing documentation for #6919, so I found a quick fix and shot off a PR.

The problem was a result of the `::caption::` option for code-block. Remove that option by moving the caption into the page content text, and the scrolling horizontal over the sidenav bug goes away.

**Admin Guide**
- Mail Domain Groups
    - [x] code-block caption, wrote file name into text
**Installation Guide**
- OpenID Connect Login Options
    - [x] code-block caption, wrote file name into text

PR #7221, opened 2020-08-24
Fixed guides scrolling horizontal layout bug

mheppler:

Removes ``:caption::`` option from code-block in two pages to fix a horizontal scrolling over sidenav layout bug. Moved caption to context text so users can still reference it.

Closes *#3577 Documentation: When content is wide enough to require horizontal scroll, content scrolls but left nav outline does not.*

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/admin/mail-groups.rst
+++ b/doc/sphinx-guides/source/admin/mail-groups.rst
@@ -33,11 +33,9 @@ To list just that Mail Domain Group, you can include the alias in the curl comma
 Creating a Mail Domain Group
 ----------------------------
 
-Mail Domain Groups can be created with a simple JSON file:
+Mail Domain Groups can be created with a simple JSON file such as domainGroup1.json:
 
 .. code-block:: json
-   :caption: domainGroup1.json
-   :name: domainGroup1.json
 
     {
       "name": "Users from @example.org",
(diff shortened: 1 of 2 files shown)


mheppler:
Removed ref to removed caption. Didn't notice that warning when building locally for whatever reason.

pdurbin (approved):
Looking good as of e26d4ea

@poikilotherm I guess we shouldn't use captions.

mheppler:
@pdurbin @poikilotherm, didn't dig into how we can fix the use of captions. Just saw that captions caused bad horizontal scrolling issues, so I moved that context into the main content text.

There is another issue related the Sphinx version update #7233 that might magically resolve this simply by using the latest and greatest version of Sphinx... OR... it might require us to use not only the latest version of Sphinx to build the guides BUT also an updated version of the [theme template files](https://github.com/ryan-roemer/sphinx-bootstrap-theme) to resolve this layout issue.

pdurbin:
@mheppler sure. Seems like a good fix. At first I was skeptical about it closing #3577 since pages are mentioned in there (Shib, Two Ravens) that you didn't touch. I have no idea why those pages are ok now. I guess we just need to keep avoiding horizontal scroll for now. And yeah, like you said, we're way behind on Sphinx so maybe an upgrade will help.

Issue #18, opened 2014-07-01
Hardcoded harvard email server, installation freezes


brainstorm:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 03:49 min
[INFO] Finished at: 2014-07-01T00:19:54-08:00
[INFO] Final Memory: 25M/124M
[INFO] ------------------------------------------------------------------------
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Use of uninitialized value $user_entry in string ne at ./install line 229.
Could not establish connection to mail.hmdc.harvard.edu,
the address you provided for your Mail server.
Please select a valid mail server, and try again.
```

The block of text above keeps repeating indefinitely.

PR #7222, opened 2020-08-24
Security18 hibernate validator vulnerability

rtreacy:
**What this PR does / why we need it**:an attacker to escalate permissions and access private values and create invalid instances - see CVE-2017-7536. It is reported to be fixed in versions 5.2.5.Final and greater

**Which issue(s) this PR closes**:Investigate and Address Possible Vulnerability in org.hibernate:hibernate-validator #18

Closes  #IQSS/dataverse-security#18

**Suggestions on how to test this**:The library upgrade is reported to be sufficient to eliminate the security vulnerability. However, it required code changes due to changes in the api and a dependency change due to a change in the initialization behavior of the library
The issue you were seeing earlier may have been due to a specified library conflicting with provide library in Payara. I had thought this was addressed due to a comment by Oliver, but somehow that change may not have gotten into the pull request. I believe it is up to date now and the hibernate validator pom entry references the Payara provided version. org.hibernate.validator.internal.util.CollectionHelper is provided by Payara

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:no

**Additional documentation**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -294,9 +294,9 @@
             <version>1.7</version> <!-- Or 1.8-SNAPSHOT -->
         </dependency>
         <dependency>
-            <groupId>org.hibernate</groupId>
+            <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>6.1.5.Final</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.glassfish</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -296,7 +296,12 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.0.3.Final</version>
+            <version>6.1.5.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.el</artifactId>
+            <version>3.0.1-b11</version>
         </dependency>
         <dependency>
             <groupId>commons-lang</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -296,7 +296,7 @@
         <dependency>
             <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <scope>provided</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.glassfish</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -296,13 +296,12 @@
         <dependency>
             <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <scope>compile</scope>
         </dependency>
-        <dependency>
+<!--       <dependency>
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.el</artifactId>
             <scope>provided</scope>
-        </dependency>
+        </dependency> -->
         <dependency>
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -297,11 +297,11 @@
             <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
         </dependency>
-<!--       <dependency>
+      <dependency>
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.el</artifactId>
             <scope>provided</scope>
-        </dependency> -->
+        </dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -294,9 +294,14 @@
...
-            <groupId>org.hibernate</groupId>
+            <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.0.3.Final</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>jakarta.el</artifactId>
+            <scope>provided</scope>
```
poikilotherm, line 304: I doubt this is necessary. There isn't a single use of this package as far as I know from the codebase. Using EL in JSF usually doesn't imply adding it to the direct dependencies. Your commit message states it had been added because of the validator dependency, but the Payara BOM already should take care of that. Please try to remove and test.

pom.xml
```diff
@@ -294,9 +294,14 @@
             <version>1.7</version> <!-- Or 1.8-SNAPSHOT -->
         </dependency>
         <dependency>
-            <groupId>org.hibernate</groupId>
+            <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.0.3.Final</version>
+            <scope>compile</scope>
```
poikilotherm, line 299: This should remain `provided` or be deleted, as `compile` is the default.

src/test/java/edu/harvard/iq/dataverse/URLValidatorTest.java
```diff
@@ -35,15 +39,15 @@ public void testIsValidWithUnspecifiedContext() {
...
+        ConstraintValidatorContext context = new ConstraintValidatorContextImpl(validatorFactory.getClockProvider(), PathImpl.createPathFromString(""),null, null);
 
         assertEquals(true, new URLValidator().isValid(value, context));
     }
 
     @Test
     public void testIsValidWithContextButInvalidURL() {
         String value = "cnn.com";
-        ConstraintValidatorContext context = new ConstraintValidatorContextImpl(null, PathImpl.createPathFromString(""), null);
+        ConstraintValidatorContext context = new ConstraintValidatorContextImpl(validatorFactory.getClockProvider(), PathImpl.createPathFromString(""),null, null);
```
poikilotherm, line 50: The complete test class should be refactored to use a JUnit5 `@ParameterizedTest`, usable for both unit tests.
The context object should be mocked. Creating a real object is of no use here and at least on my machine creates errors. Happy to create a PR against this PR.

pom.xml
poikilotherm: IMHO you should take a look at this:

image

We're using an **internal** validator class in our code and I'm not sure if this is a supported use case. This should be refactored and use [Java 8 Collections API](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html).

poikilotherm: BTW @kcondon I noticed you're testing on Payara 5.201. It's not important for this issue (both 5.201 up to 5.2020.4 use Hibernate Validator 6.1.2 Final), but we should make sure that testers, QA etc all use the same version referenced in `pom.xml`.

poikilotherm (changes requested):

pom.xml
rtreacy: Tested this again without jakarta.el - brings back the problem it was addressing
-------------------------------------------------------------------------------
Test set: edu.harvard.iq.dataverse.ingest.IngestUtilTest
-------------------------------------------------------------------------------
Tests run: 12, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.121 s <<< FAILURE! - in edu.harvard.iq.dataverse.ingest.IngestUtilTest
testDirectoryLabels  Time elapsed: 0.103 s  <<< ERROR!
javax.validation.ValidationException: HV000183: Unable to initialize 'javax.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
	at edu.harvard.iq.dataverse.ingest.IngestUtilTest.testDirectoryLabels(IngestUtilTest.java:473)

pom.xml
rtreacy: @poikilotherm - I suspected that this was the reason I couldn't use provided, but refactoring that code is a separate issue IMHO

kcondon:
@poikilotherm I was not aware that this was tied to a specific version of Payara. I think as a policy, there needs to be more clarity on what is supported. After all, in the wild, people will use whatever generation they happen to be on, within some reasonable range, and expect it to work, without any additional instructions from us, that is.

PR #7266, opened 2020-09-14
File extensions missing for some original downloads

### LINKED ISSUE #7265: Filename extension dropped when downloading some saved originals
`TL;DR:`
For "legacy" saved originals (created before v5.0) of certain types, the extension is missing on download. For example, saved originals of type `application/x-stata-14` are downloading without the `.dta` extension. This is happening because some types are missing from the list in `FileUtil.generateOriginalExtension()`. 

Extra info: 

This must have happened in the PR that added saving the entire filename of the original, uningested file, as uploaded by the user (instead of saving just its type). For legacy files the download code is supposed to be doing what we were doing before - generate the extension based on the type. As part of that PR an attempt was made to consolidate some code previously duplicated in different parts of the application. The specific fragment of code related to this issue is the following in `dataaccess.StoredOriginalFile.java` (now commented out):

```
    /* 
            DataFile.getOriginalFileName() method replaces this code
    private static String generateOriginalExtension(String fileType) {

        if (fileType.equalsIgnoreCase("application/x-spss-sav")) {
            return ".sav";
        } else if (fileType.equalsIgnoreCase("application/x-spss-por")) {
            return ".por";
        } else if (fileType.equalsIgnoreCase("application/x-stata") || fileType.equalsIgnoreCase("application/x-stata-13") || fileType.equalsIgnoreCase("application/x-stata-14") || fileType.equalsIgnoreCase("application/x-stata-15")) {
            return ".dta";
        } else if (fileType.equalsIgnoreCase("application/x-dvn-csvspss-zip")) {
            return ".zip";
        } else if (fileType.equalsIgnoreCase("application/x-dvn-tabddi-zip")) {
            return ".zip";
        } else if (fileType.equalsIgnoreCase("application/x-rlang-transport")) {
            return ".RData";
        } else if (fileType.equalsIgnoreCase("text/csv") || fileType.equalsIgnoreCase("text/comma-separated-values")) {
            return ".csv";
        } else if (fileType.equalsIgnoreCase("text/tsv") || fileType.equalsIgnoreCase("text/tab-separated-values")) {
            return ".tsv";
        } else if (fileType.equalsIgnoreCase("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")) {
            return ".xlsx";
        }
        logger.severe(fileType + " does not have an associated file extension");
        return "";
    } */
```

However, if you look at `DataFile.getOriginalFileName()` mentioned above, it calls `FileUtil.generateOriginalExtension()` which tries to do the same thing as the above, BUT doesn't support all of the types above: 

```
    public static String generateOriginalExtension(String fileType) {

	if (fileType.equalsIgnoreCase("application/x-spss-sav")) {
            return ".sav";
	} else if (fileType.equalsIgnoreCase("application/x-spss-por")) {
            return ".por";
        } else if (fileType.equalsIgnoreCase("application/x-stata")) {
            return ".dta";
        } else if (fileType.equalsIgnoreCase( "application/x-rlang-transport")) {
            return ".RData";
        } else if (fileType.equalsIgnoreCase("text/csv")) {
            return ".csv";
        } else if (fileType.equalsIgnoreCase( "application/vnd.openxmlformats-officedocument.\
spreadsheetml.sheet")) {
            return ".xlsx";
        }

        return "";
    }
```

hence no extensions for saved stata files of types other than "application/stata". 
The fix would be to copy-and-paste the  code from StoredOriginalFile.java to FileUtil.java.

landreev:

Closes #7265

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
@@ -695,21 +695,27 @@ public class FileUtil implements java.io.Serializable  {
     }
 
     public static String generateOriginalExtension(String fileType) {
-
         if (fileType.equalsIgnoreCase("application/x-spss-sav")) {
             return ".sav";
         } else if (fileType.equalsIgnoreCase("application/x-spss-por")) {
-            return ".por";
-        } else if (fileType.equalsIgnoreCase("application/x-stata")) {
+            return ".por";    
+        // in addition to "application/x-stata" we want to support 
+        // "application/x-stata-13" ... etc.:
+        } else if (fileType.toLowerCase().startsWith("application/x-stata")) {
             return ".dta";
-        } else if (fileType.equalsIgnoreCase( "application/x-rlang-transport")) {
+        } else if (fileType.equalsIgnoreCase("application/x-dvn-csvspss-zip")) {
+            return ".zip";
+        } else if (fileType.equalsIgnoreCase("application/x-dvn-tabddi-zip")) {
+            return ".zip";
+        } else if (fileType.equalsIgnoreCase("application/x-rlang-transport")) {
             return ".RData";
-        } else if (fileType.equalsIgnoreCase("text/csv")) {
+        } else if (fileType.equalsIgnoreCase("text/csv") || fileType.equalsIgnoreCase("text/comma-separated-values")) {
             return ".csv";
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
This looks fine so I'm approving it. My only thought is that we could also delete the commented out code in StoredOriginalFile.java but there's no rush.

PR #7272, opened 2020-09-16
Configuring file stores (storage drivers) for individual datasets

### LINKED ISSUE #6872: Direct S3 Upload mechanism for individual datasets (API based)
As of 4.20 we now have the mechanism for uploading files directly to an S3 bucket. As of now it can be enabled on an individual store, that can be assigned to a dataverse. 
We want to add a finer-grain mechanism for enabling it for individual datasets and/or users. That would help a use case of a one off giant file upload, without forcing direct upload on an entire dataverse.

One approach that was discussed was to leave the GUI upload controls as they are now. But add this extra control granularity for API uploads only. Most users who need to upload an occasional 30GB file will likely be happy to use the DvUploader tool. This solution would have zero UI impact (enabling direct upload for a dataset or user would also be done via API).

This should be fairly straightforward. If we chose to just go with enabling it for datasets, and not bother with users, the current code for the most part already supports enabling it for any dvobject. 

From @qqmyers:

> Could just split the flag into 'direct upload allowed' and 'use in UI'.

landreev:
Even though the title of the issue specifically mentions "direct S3 upload", the PR adds something more general - an ability to designate a file store for a specific dataset. The file store in question does not have to be S3. But enabling direct S3 upload is the main use case behind this PR. For example, if we want to enable  direct upload for a specific dataset in prod., without opening it for everybody, we will achieve it by 
1. Creating another S3-type store, pointing to the same storage bucket (for ex., "s3direct")
2. Enable direct upload on the above. 
3. Configure "s3direct" to be the designated storage driver for the dataset in question, using the API added in this PR: 
`curl -H "X-Dataverse-key: XXX" -X PUT -d s3direct http://localhost:8080/api/datasets/NNNN/storageDriver`

The words "API based" in the issue title refer to the fact that a file store can be configured for a dataset via API only; i.e. there's no GUI for that (like we have for dataverses). But once a direct upload-enabled store has been assigned to a dataset, the uploads will work both via the dataset page, or the API (using the DVUploader utility); just like when it's enabled dataverse-wide. 

Closes #6872

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/admin/dataverses-datasets.rst
+++ b/doc/sphinx-guides/source/admin/dataverses-datasets.rst
@@ -59,6 +59,8 @@ The available drivers can be listed with::
 
     curl -H "X-Dataverse-key: $API_TOKEN" http://$SERVER/api/admin/dataverse/storageDrivers
     
+(Individual datasets can be configured to use specific file stores as well. See the "Datasets" section below.)
+
 
 Datasets
 --------
@@ -129,4 +131,24 @@ Diagnose Constraint Violations Issues in Datasets
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 To identify invalid data values in specific datasets (if, for example, an attempt to edit a dataset results in a ConstraintViolationException in the server log), or to check all the datasets in the Dataverse for constraint violations, see :ref:`Dataset Validation <dataset-validation-api>` in the :doc:`/api/native-api` section of the User Guide.
+
+Configure a Dataset to store all new files in a specific file store
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Configure a dataset to use a specific file store (this API can only be used by a superuser) ::
+ 
+    curl -H "X-Dataverse-key: $API_TOKEN" -X PUT -d $storageDriverLabel http://$SERVER/api/datasets/$dataset-id/storageDriver
+    
+The current driver can be seen using::
+
+    curl http://$SERVER/api/datasets/$dataset-id/storageDriver
+
+It can be reset to the default store as follows (only a superuser can do this) ::
+
+    curl -H "X-Dataverse-key: $API_TOKEN" -X DELETE http://$SERVER/api/datasets/$dataset-id/storageDriver
+    
+The available drivers can be listed with::
+
+    curl -H "X-Dataverse-key: $API_TOKEN" http://$SERVER/api/admin/dataverse/storageDrivers
+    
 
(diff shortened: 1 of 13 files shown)


qqmyers (approved):
Looks good to me. The only ~issue I see is that there's no UI indicator that a Dataset isn't inheriting the storageDriver from it's Dataverse, so it may not be obvious when a Dataset has been switched. No reason that can be a future issue though.

landreev:
@qqmyers 
> Looks good to me. The only ~issue I see is that there's no UI indicator that a Dataset isn't inheriting the storageDriver from it's Dataverse, so it may not be obvious when a Dataset has been switched. No reason that can be a future issue though.

We don't otherwise have a UI indicator showing which store the files are going to, when it's inherited from the dataverse... do we?
Thanks for bringing this up - made me check and realize that there was a whole bunch of places where the upload size limit inherited from the dataverse would be used, instead of the dataset's own, etc...

qqmyers:
Good catch! I was just thinking that an admin can check (or change) the Dataverse setting under General Information, but that value may not be what the dataset has.

kcondon:
1. get current driver fails silently with bad id: {} and 500 error in server log. Does indicate bad id if drop last /
2. get available drivers seems to double the driver names:
curl -H "X-Dataverse-key: xxxx-yyyy-zzz" http://localhost:8080/api/admin/dataverse/storageDrivers
{"status":"OK","data":{"s31":"s31","s32":"s32","file2":"file2","file1":"file1"}}'

landreev:
>     2. get available drivers seems to double the driver names:
>        curl -H "X-Dataverse-key: xxxx-yyyy-zzz" http://localhost:8080/api/admin/dataverse/storageDrivers
>        {"status":"OK","data":{"s31":"s31","s32":"s32","file2":"file2","file1":"file1"}}'

Please note that this is not part of my PR - this is an existing API; that was merged some months ago. 
It doesn't "double" the driver names. These are "id" and "label" pairs describing each driver. We've been using the same value for both in all the practical use cases (I'm not sure why anyone would want to use different values? Something more descriptive for the label - idk); but they *can* be different.

qqmyers:
FWIW: TDL has s3 with label "TDL" and s3tacc with label "TACC". Harvard might want labels like 'Normal' and 'Large Data', etc.

landreev:
> FWIW: TDL has s3 with label "TDL" and s3tacc with label "TACC". Harvard might want labels like 'Normal' and 'Large Data', etc.

I was thinking something along these lines - there may be a situation where we need to spell out the name of the institution that owns a specific storage location (or even the name of the grant that pays for it)... We may need/want to display this kind of info on the page later on...

landreev:
>     1. get current driver fails silently with bad id: {} and 500 error in server log. 

True; just checked in a fix. Should be printing the same "no such dataset" error message as the PUT and DELETE versions now.

PR #7274, opened 2020-09-17
Catch abnormal EC2 exit and fail build

### LINKED ISSUE #7268: As someone doing code review I want to trust that tests are actually passing
Our new continuous integration system is fantastic but there are still some kinks to work out.

Today I was doing a code review and the system told me that all tests are passing but when I dug into it, I learned that the EC2 instance was never completely configured so the API test suite didn't run (screenshots below).

What we want instead is that any failure will mark the build as failed so we aren't misled when doing code review.

Screen Shot 2020-09-15 at 10 40 09 AM
Screen Shot 2020-09-15 at 10 40 17 AM
Screen Shot 2020-09-15 at 10 40 39 AM

donsizemore:
**What this PR does / why we need it**: AWS sometimes dies "connection closed by remote host", which doesn't trigger an exit failure so the Jenkins run continues. This checks for a semaphore created in EC2 after the Ansible run, and fails the build if not found.

Closes #7268

**Suggestions on how to test this**: submit a PR, kill EC2 instance in mid-run

### DIFF
--- a/tests/jenkins/ec2/Jenkinsfile
+++ b/tests/jenkins/ec2/Jenkinsfile
@@ -41,6 +41,13 @@ pipeline {
 	  sourcePattern: 'src/main/java',
 	  exclusionPattern: 'src/test*']
 	)
+        script {
+          if (fileExists('./ansible_complete')) {
+            sh '/bin/rm ./ansible_complete'
+          } else {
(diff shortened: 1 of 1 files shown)


mheppler (approved):
👍 Looks like it delivers as advertised.

PR #7277, opened 2020-09-21
sort compound field children by display order

### LINKED ISSUE #3685: Metadata fields on the "edit metadata" pages of some datasets are displayed out of order
This issue collects several instances of metadata fields within compound metadata fields being displayed out of order on edit dataset metadata pages.

When author metadata fields were added in the citation metadata block of [this dataset](https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/EYMZZX&version=DRAFT) (it's unpublished, so you'll need admin access to see it), the order of the fields are different than normal ("Name," Affiliation," on the first row, "Identifier Scheme," and "Identifier" on the second row).

screen shot 2017-03-10 at 10 38 06 am

Unfortunately, I'm not able to reproduce this bug, either in other existing or new datasets or even in this dataset: If you press the plus sign to add another author for that dataset, the order of fields will be correct.

qqmyers:
**What this PR does / why we need it**: Sorts children of compound dataset fields by the specified displayOrder which orders them as intended in the metadata summary display

Closes #3685 

**Special notes for your reviewer**:  This most directly resolves the issue identified in https://groups.google.com/g/dataverse-community/c/EwtjRM4BhHM, but @jggautier has pointed out that there may be a connection to #3685. My guess for #3685 would be that the child fields are currently being sorted based on datasetfieldtype ids AND that there are duplicate datasetfieldtypes in the db affecting some datasets. If that's true, this will 'fix' those datasets (assuming the displayorder is correct for any duplicate fields), but it would probably be good to check the db to see if there are duplicates to resolve (or other issue). In either case, the change her may fix #3685 from a practical standpoint, and it will address the community issue, but there may be additional db work to resolve the underlying cause of #3685.

**Suggestions on how to test this**: To see the effect, you need datasetfieldtypes where the ids are not in the displayorder. ON a fresh install, I caused this by:
 ```
update datasetfieldtype set displayorder=10 where id=12;
 update datasetfieldtype set displayorder=11 where id=11;
```
which reverses the order of the author identifier and identifier scheme fields. 
Without this PR, the display is unchanged (identifier scheme before the identifier). With it, the identifier comes first, both in the metadata edit display and the metadata summary (e.g. "Admin, Dataverse (Dataverse.org) 1234 - ORCID:" . (Note that the '-' before 'ORCID' and the ':' after it are part of the field formatting and don't really make sense if you reverse the order from ' - ORCID: 1234' - this isn't a problem with the PR, just an artifact from this minimal test setup.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: only change is to correct the ordering of compound field children.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
@@ -9,6 +9,7 @@ import edu.harvard.iq.dataverse.util.BundleUtil;
 import edu.harvard.iq.dataverse.util.MarkupChecker;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -95,9 +96,10 @@ public class DatasetFieldCompoundValue implements Serializable {
     public List<DatasetField> getChildDatasetFields() {
         return childDatasetFields;
     }
-
+    
     public void setChildDatasetFields(List<DatasetField> childDatasetFields) {
         this.childDatasetFields = childDatasetFields;
+        this.childDatasetFields.sort(DatasetField.DisplayOrder);
     }
 
     @Override


pdurbin (approved):
One-line fix and the tests passed. Approved. 😄

kcondon:
Jim looking into whether changes are taking place after testing not showing changes.

qqmyers:
After discussing with @kcondon - it looks like this (compound field ordering in the summary display) is already fixed somehow in v 5.0. So far, I have not been able to find where an alternate fix was made for 5.0, either in the java or .xhtml parts, but I've confirmed that swapping the displayorder as noted in the test procedure above shows the error in 4.20 and not in 5.0. So - closing.

PR #7289, opened 2020-09-30
7282 datacite error handling

### LINKED ISSUE #7282: PID registration errors not handled properly
FInalizeDatasetPublicationCommand was supposed to stop if an attempt to register a datacite pid failed, then unlock the dataset and send a notification to the user saying "datacite issue was encountered, please try again later". This was tested and worked as designed with at least some datacite errors. But we are definitely seeing cases where it does not. Where a datacite failure results in an exit that's not "clean" - that leaves the dataset locked, and without a notification to the user.

I do remember there was an issue worked on in 5.0 in which the handling of datacite failures in the registration service bean was specifically addressed. It may be entirely possible to address this without going into the service bean, by just catching exceptions differently inside the command proper - it's just not obvious to me.

If you look at the exception stack trace below, it starts with the DataCiteRESTfullClient receiving an error code. But the way it was supposed to be handled, idServiceBean would simply return "false" on encountering a failure. As in lines 344-345 of the finalize command:

```
//A false return value indicates a failure in calling the service
if (!idServiceBean.publicizeIdentifier(df)) {
```

but instead the method appears to be throwing an uncaught runtime. Still, I would expect the following lines in the command (357-363) to properly handle the exception:

```
} catch (Throwable e) {
   // Send failure notification to the user: 
   notifyUsersDatasetPublishStatus(ctxt, dataset, UserNotification.Type.PUBLISHFAILED_PIDREG);    
   ctxt.datasets().removeDatasetLocks(dataset, DatasetLock.Reason.finalizePublication);
   throw new CommandException(BundleUtil.getStringFromBundle("dataset.publish.error", args), this);
}
```

but instead it appears to fail (with an ejb exception?) in the notifyUsers... part. 
[stacktrace.txt](https://github.com/IQSS/dataverse/files/5277777/stacktrace.txt)

landreev:

This is a very limited scope PR, it fixes a specific issue, an example of which is attached to the issue as a stacktrace. A condition caused by *intermittent*  DataCite errors; such that during the publicaiton of a dataset the first call to DataCite succeeds, but the second one doesn't. When that happens, in its current state the application exits the publication command without unlocking the dataset and notifying the user. 

I wanted to check in more diagnostic logging, not immediately related to the issue above, but decided to handle that separately.

Closes #7282 

I tested fairly thoroughly by hard-coding exceptions in all possible places. I still cannot think of a way to test it in QA. Other than setting up our own fake datacite service, and programming it to serve the first (/metadata), but not the second (/doi) of the 2 api calls made during a dataset registration. 

The best QA for it would probably be just confirming that datasets can still be published under normal circumstances.

**Additional documentation**:


qqmyers (approved):
Makes sense to me. I agree there's no reason to process an UnsupportedEncodingException differently (even though it means a config error rather than something that can occur randomly at run time). Overall, looks like this parallels the other work to get rid of the RuntimeException so the transaction won't roll back and adds exception-specific info to help with debug - as intended.

scolapasta (approved):
overall looks good; may make sense to remove the commented  UnsupportedEncodingException  catch code, rather than leaving it in uncommented. (I get leaving it in, but I could also see it never getting cleaned up).

landreev:
> @scolapasta: overall looks good; may make sense to remove the commented UnsupportedEncodingException catch code, rather than leaving it in uncommented. (I get leaving it in, but I could also see it never getting cleaned up).

OK, mostly I just wanted the reviewers to see my comment there. But will remove it now.

PR #7308, opened 2020-10-07
Increase s3 scaling

qqmyers:
**What this PR does / why we need it**: To fix issues with exhausting the aws client http pool, this increases the pool size and eliminates some redundant checks on the bucket

Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
+++ b/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
@@ -1044,10 +1044,7 @@ public class S3AccessIO<T extends DvObject> extends StorageIO<T> {
     		AmazonS3ClientBuilder s3CB = AmazonS3ClientBuilder.standard();
 
     		ClientConfiguration cc = new ClientConfiguration();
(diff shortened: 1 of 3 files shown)


landreev (approved):

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1043,6 +1043,14 @@ private static AmazonS3 getClient(String driverId) {
     		// get a standard client, using the standard way of configuration the credentials, etc.
     		AmazonS3ClientBuilder s3CB = AmazonS3ClientBuilder.standard();
 
+    		ClientConfiguration cc = new ClientConfiguration();
+    		Integer poolSize = Integer.getInteger("dataverse.files." + driverId + ".connection-pool-size");
```
scolapasta, line 1047: this can be simplified a little:
public static Integer getInteger(String nm, int val) where val is the default (so no null check needed)

scolapasta (approved):
I added a suggestion for one (small) simplification.

PR #7313, opened 2020-10-07
Iqss/7309 s3 resource leaks

### LINKED ISSUE #7309: Potential S3 resource leaks exposed by 5.1
The switch in v5.1 to use a single AWS client appears to have exposed some additional resource leaks /open connections to S3. One source of these may be some uses of Java Channels where the Channel is not closed. This issue is a placeholder for work to fix those leaks, check for others, and potentially to add some monitoring/retry capabilities to avoid situations where S3 connections fail (causing downloads as well as display of thumbnails, etc. to fail).
(Note #7308 increases the default pool size and makes it configurable. Setting the pool size to a larger value should help.)

qqmyers:
**What this PR does / why we need it**:  Closes more streams/channels that may hold S3 http connections open and exhaust the aws s3 client's pool.

Closes #7309 

Tried to use the new try-with-resources format where that looked simple and just used Apache's IOUtils.closeQuietly() for others (it just closes if the object isn't null and eats any exception). If I understand correctly, using try-with-resources will still throw any exception (unless you add a catch) so I think the only effect is to assure the streams/channels are closed.

One place where this PR isn't complete is in the tabulardata/impl/plugins classes. There are other classes there that do the same thing as DTAFileReaderSpi in creating a stream that doesn't look like it gets closed, and in having methods that have a stream param. In this latter case, I didn't find where these methods were getting called, so I didn't want to just close the streams in case that was already handled by the caller (or the stream is reused, etc.). So - some discussion about what could/should happen there would help.

**Suggestions on how to test this**: If we can reliably exhaust the pool by setting a low value and using some of the functionality affected by this PR (e.g. thumbnails created for pdfs), then applying this PR should stop that problem. 
Testing should also cover making sure that things still work, e.g. that I haven't accidentally closed streams/channels before they're used. 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: No.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
@@ -67,6 +67,8 @@ import edu.harvard.iq.dataverse.ingest.tabulardata.impl.plugins.sav.SAVFileReade
 import edu.harvard.iq.dataverse.ingest.tabulardata.impl.plugins.por.PORFileReader;
 import edu.harvard.iq.dataverse.ingest.tabulardata.impl.plugins.por.PORFileReaderSpi;
 import edu.harvard.iq.dataverse.util.*;
+
+import org.apache.commons.io.IOUtils;
 //import edu.harvard.iq.dvn.unf.*;
 import org.dataverse.unf.*;
 import java.io.BufferedInputStream;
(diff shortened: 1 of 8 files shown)


src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/dta/DTAFileReaderSpi.java
```diff
@@ -131,6 +131,7 @@ public boolean canDecodeInput(Object source) throws IOException {
 
     @Override
     public boolean canDecodeInput(BufferedInputStream stream) throws IOException {
+    	//who closes this stream?
```
qqmyers, line 134: guessing this method could/should close it but I didn't see where it was getting called.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -1056,12 +1061,11 @@ private BufferedInputStream openFile(DataFile dataFile) throws IOException {
         if (storageIO.isLocalFile()) {
             inputStream = new BufferedInputStream(storageIO.getInputStream());
         } else {
-            ReadableByteChannel dataFileChannel = storageIO.getReadChannel();
-            File tempFile = File.createTempFile("tempIngestSourceFile", ".tmp");
-            FileChannel tempIngestSourceChannel = new FileOutputStream(tempFile).getChannel();
-            
-            tempIngestSourceChannel.transferFrom(dataFileChannel, 0, storageIO.getSize());
-            
+        	File tempFile = File.createTempFile("tempIngestSourceFile", ".tmp");
```
qqmyers, line 1064: FWIW - this private method doesn't get used according to eclipse - do we need it?

landreev: It really appears an unused leftover method.

src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/dta/DTAFileReaderSpi.java
landreev: We are not calling these methods at this time. They are part of a builtin infrastructure for automatically matching ingest plugins to specific file types... We are not using it now, but we may go back to it.  
It would be the responsibility of the code that calls these methods to close the streams.

landreev (approved):
Awesome!

PR #2964, opened 2016-02-23
#24 add data cite doi support

sekmiller:
Updated Documentation for Installers Guide included

PR #7314, opened 2020-10-07
Remove unused dependency in response to Dependabot notification

djbrooke:

Addresses a security issue raised by Dependabot by removing an unused dependency from pom.xml. Thanks @qqmyers for the alley oop here. :)

None, but will allow us to respond to the Dependabot alert.

I don't spend much time in pom.xml so please make sure I did not mess up formatting.

Since it's unused, it shouldn't have any impact?

N/A

N/A

**Additional documentation**:

N/A


pom.xml
```diff
@@ -440,11 +440,6 @@
             <artifactId>slf4j-log4j12</artifactId>
             <version>1.6.1</version>
         </dependency -->
-        <dependency>
```
qqmyers, line 443: FWIW: This was added in #2964 but I don't see any actual use of it in that PR. Haven't checked the other pom additions from it, but they aren't dependabotted either.

pdurbin (approved):
The API tests passed. I think we're good.

PR #7320, opened 2020-10-09
persist debug=true in session, toggle with query parameter #7273

### LINKED ISSUE #7273: Add "debug=true" to session, so we can use it more robustly
Currently, we have the ability (as a superuser) to add a parameter of debug=true to a URL and for 1 or two pages (the dataverse page for example), spit out some extra info in the UI. However if you then click (for example on "next page", the debug goes away.

This is to set the debug into the session so you don't have to add it to the url each time. We could even consider adding a UI toggle if we want, but that's not as important.

We can also in the future consider adding more uses of debug = true if we find other places it could be useful.

pdurbin:

Showing debug information on a page is useful but currently requires you to add debug=true as a query parameter each time you load the page. The only page this works on is the dataverse page (and this pull request doesn't change that).

This pull request allows you to add debug=true as a query parameter once and have the boolean remain true for your session.

Closes #7273

I didn't touch the existing logic in SearchIncludeFragment:

- debug is only true if you are a superuser
- ~~another way for debug to be true is the undocumented ":Debug" database setting.~~ 2020-10-14 update: we decided to remove this in a4582f0.

- Log in as a superuser.
- Go to http://localhost:8080/?debug=true (or the equivalent siteUrl) to set enable debugging for your session.
- Navigate around, observing that the extra debugging information is only available on on the dataverse page and relates to Solr (screenshot below) and is only available to superusers.
- Go to http://localhost:8080/?debug=false to disable debugging.
- ~~Play around with the undocumented ":Debug" database setting, setting it to true and false and observe that it acts the same was as the debug query parameter.~~

How the debugging on the dataverse page looks hasn't changed but here's what it looks like:

Screen Shot 2020-10-09 at 10 53 08 AM

Maybe? I'm happy to add one.

**Additional documentation**:

None. I'm happy to add some. I documented the boolean within the code, at least.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
@@ -1018,8 +1018,7 @@ public class SearchIncludeFragment implements java.io.Serializable {
     }
 
     public Boolean getDebug() {
-        return (session.isDebug() && session.getUser().isSuperuser())
-                || settingsWrapper.isTrueForKey(":Debug", false);
+        return session.isDebug();
     }
 
     public void setDebug(Boolean debug) {
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
@@ -130,7 +130,7 @@ public class SearchIncludeFragment implements java.io.Serializable {
     Map<String, String> staticSolrFieldFriendlyNamesBySolrField = new HashMap<>();
     private boolean solrIsDown = false;
     private Map<String, Integer> numberOfFacets = new HashMap<>();
-    private boolean debug = false;
+    private Boolean debug;
 //    private boolean showUnpublished;
     List<String> filterQueriesDebug = new ArrayList<>();
 //    private Map<String, String> friendlyName = new HashMap<>();
@@ -1017,13 +1017,15 @@ public class SearchIncludeFragment implements java.io.Serializable {
         this.rootDv = rootDv;
     }
 
-    public boolean isDebug() {
-        return (debug && session.getUser().isSuperuser())
+    public Boolean getDebug() {
+        return (session.isDebug() && session.getUser().isSuperuser())
                 || settingsWrapper.isTrueForKey(":Debug", false);
     }
 
-    public void setDebug(boolean debug) {
-        this.debug = debug;
+    public void setDebug(Boolean debug) {
+        if (debug != null) {
+            session.setDebug(debug);
+        }
     }
 
     public List<String> getFilterQueriesDebug() {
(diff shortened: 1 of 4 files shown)


sekmiller (approved):

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
```diff
@@ -1017,13 +1017,15 @@ public void setRootDv(boolean rootDv) {
         this.rootDv = rootDv;
     }
 
-    public boolean isDebug() {
-        return (debug && session.getUser().isSuperuser())
+    public Boolean getDebug() {
```
scolapasta, line 1020: why not still isDebug?

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
```diff
@@ -1017,13 +1017,15 @@ public void setRootDv(boolean rootDv) {
...
+        return (session.isDebug() && session.getUser().isSuperuser())
```
scolapasta, line 1021: Should all this logic go to the session? i.e. to use in other future pages it would be easier to not have to repeat the superuser check or the settings check?

scolapasta (commented):
A couple of small comments; neither are critical, but I think it might be cleaner.

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
pdurbin: Because with isDebug I get this error:

`/dataverse.xhtml @47,87 value="#{SearchIncludeFragment.debug}": The class 'edu.harvard.iq.dataverse.search.SearchIncludeFragment' does not have a readable property 'debug'.`
`javax.el.PropertyNotFoundException: /dataverse.xhtml @47,87 value="#{SearchIncludeFragment.debug}": The class 'edu.harvard.iq.dataverse.search.SearchIncludeFragment' does not have a readable property 'debug'.`

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
pdurbin: Not a bad idea. Consolidated in a4582f0 (where I also removed that undocumented :Debug database setting, as we discussed).

In ce78897 I made another change we discussed in Slack which is to remove the getters and setters from the search code and to rely on the getters and setters in the session code. In that commit I also attempted to explain why I'm using a Boolean with the viewParam. In short, I'm using null to mean "don't change the state".

pdurbin:
I'm moving this back to code review. Here are the new commits:

a4582f05e consolidate logic, remove :Debug database setting #7273
ce78897f7 switch to getters and setters in session bean #7273

I also left a comment at https://github.com/IQSS/dataverse/pull/7320#discussion_r504946017

In short, we are centralizing logic in the session bean and taking it out of a search-related bean. I also removed the one use of the undocumented :Debug database setting.

scolapasta (approved):

Issue #5379, opened 2018-12-07
As a developer, I need documentation on "to rebase or not" in feature branches.


poikilotherm:
Dear devs, QA and anybody else who it may concern,

I would be very glad if some documentation about how to handle a diverged `develop` branch in my feature branch when creating or updating a PR before merge.

Seeing a lot of "merge" commits in the codebase and personally not being a fan of those if they are caused by devs merging in `develop` instead of rebasing and cleaning up conflicts in their commits, I would really appreciate some guidance on this.

Enforcing a standard or rule on this is load on QA, but IMHO it is worth it. Open for discussion :-) Again, I would really appreciate feedback, as I am currently rebasing, which has its advantages and disadvantages (just like merging has).

pdurbin:
@poikilotherm thanks for opening this issue. For now I'll just encourage you to read through #3418, especially the part where I mention my desire for a "clean git history" but you have to understand that we're still relatively new to git and many of us used SVN for years and years. Only recently have we started using branches and pull requests per feature. We are slowly evolving and maturing and feedback like this is healthy for us. Thanks.

poikilotherm:
Looks like this is nothing too many people care about. :shrug: Closing for cleanup.

pdurbin:
I just thought I'd mention that @adaybujeda for example likes to rebase his pull requests. This is different that what we do internally (we always just merge develop into our branches). So we're starting to gain experience with both approaches. I like the idea of a clean git history but I'm fine with merging or rebasing. It's all good. 😄

pdurbin:
From @adaybujeda at https://github.com/IQSS/dataverse/pull/7325#discussion_r940290338

"For PRs, I tend to squash all my changes into a single commit. Usually, when new changes/updates need to be made to the pr, I will rebase the branch, reset my commit and make a new one with the latest changes/updates. I find it easier to review and to reason about the changes this way."

The context is that we're thinking about removing a change from a PR and we don't want that removal to apply to another PR down the road.

Overall, I like the idea that squashing and rebasing would provide a clean git history. There's an image of this here: https://blog.carbonfive.com/always-squash-and-rebase-your-git-commits/

One concern I have with squashing and rebasing is that it's harder for the reviewer to see what changed after the reviewer gives feedback and the developer changes something. There's no diff to look at. And the reviewer probably doesn't remember exactly how the old code looked.

[...]

PR #5359, opened 2018-11-30
4949 download package s3

matthew-a-dunlap:
Note: There are still two changes incoming for this PR.
1. Documentation around the `:DownloadMethod` changes and a .rst page on how to set up dcm s3.
2. A bump to the dcm version number once this pr is merged https://github.com/sbgrid/data-capture-module/pull/35

## Related Issues

- connects to #4949 

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #7325, opened 2020-10-13
Simple support for remote web stores

### LINKED ISSUE #7324: Simple support for remote web stores
Based on use cases from Odum's TRSA (see #5213) and recent work to support remote uploads, I'm suggesting a mechanism extending the current StorageIO mechanism that would allow Dataverse management of a file at a remote URL 'as though' it were in S3. Recognizing that there are many potential remote stores that might fit this model, I've created [a design document](https://docs.google.com/document/d/1rDhL2QBY2NhVqan3Mwp11BJ0O9naiFEhvnlbKVDQkBc/edit?usp=sharing) and a proof-of-concept implementation (#7325) to share with the community.

The basic concept is to treat the file URL as read-only and to retrieve it's bytes or provide a download redirect in the same way that the S3 store manages file access. However, since the remote web store is assumed ~read-only in this design, any/all derived files (thumbnails, ingested versions, provenance files, etc.) are managed by an underlying S3 or File store.

This mechanism works for public URLs. I've also suggested/implemented a URL presigning mechanism, roughly analogous to what S3 stores use to provide secure download URLs, that could be used by a remote store to verify that Dataverse was the source of the request (which would only be made if the user is allowed access per Dataverse's configured controls) and only allow access when Dataverse has 'pre-approved' the request. (The code includes java code to sign and validate these requests - validation implemented as module/filter for common web servers could simplify what needs to be done at the remote store.)

I've created this issue, design doc, and draft PR to encourage discussion and get feedback. Does this mechanism, or extensions of it, support use cases from other community members? Are there alternative designs that could be made general and fit well into Dataverse's architecture? Are there concerns about the mechanism proposed?

qqmyers:
**What this PR does / why we need it**: Enables a class of remote stores with a general mechanism

Closes #7324 

**Suggestions on how to test this**: The API should allow testing with any web server. Create a file at a remote URL and verify that Dataverse works (file is downloadable, normal aux files can be created). To test security, the signed URL can be retrieved from the remote server and 'hand' validated (removing the "token" param and running a SHA512 hash on the remainder on an online site to compare with the token.)

EDIT: Also, see the comment below, with the asadmin commands etc. for setting up a sample HTTP store and creating a working Datafile entry with a storage identifier pointing to a remote URL. 
There is more information in the issue, and in the design document linked there - https://docs.google.com/document/d/1rDhL2QBY2NhVqan3Mwp11BJ0O9naiFEhvnlbKVDQkBc/edit?usp=sharing - on how this "HTTP overlay"  works, and how it stores extra files, etc.  (L.A.)

Also - since this PR includes a refactoring of classes related to external tools (due to some code sharing with the[ OpenDP work](https://github.com/IQSS/dataverse/tree/7715-signed-urls-for-external-tools) which leverages the same URL signing introduced here, some basic regression testing of external tool config/launch should be done. Essentially, the code to create the parameters needed by a previewer (or other tool) on its command line has been moved to a new base class. If configuring a previewer and seeing the preview works, it would confirm that move hasn't accidentally broken anything. (This change should not make any changes to how external tools work, it is just internal refactoring.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: API only at this point.

**Additional documentation**:


qqmyers:
Verified this now works with file backing store. For example, using a default ansible install of this branch,
```
cd /usr/local/payara5/glassfish
bin/asadmin create-jvm-options -Ddataverse.files.trsa.type=http
bin/asadmin create-jvm-options -Ddataverse.files.trsa.label=trsa
bin/asadmin create-jvm-options -Ddataverse.files.trsa.secretkey=12345
bin/asadmin create-jvm-options -Ddataverse.files.trsa.baseStore=file
bin/asadmin create-jvm-options -Ddataverse.files.trsa.url-expiration-minutes=120
bin/asadmin create-jvm-options '-Ddataverse.files.trsa.baseUrl=https\://qdr.syr.edu'
service payara restart
```

Create a dataset, get your api key and submit a 'file', e.g.
`curl -H X-Dataverse-key:affc67b6-c57d-430f-b1a7-18b518460817 -X POST -F 'jsonData={"description":"My description.","directoryLabel":"data/subdir1","categories":["Data"], "restrict":"false", "storageIdentifier":"trsa://themes/custom/qdr/images/CoreTrustSeal-logo-transparent.png", "checksumType":"MD5", "md5Hash":"509EF88AFA907EAF2C17C1C8D8FDE77E", "label" : "testlogo.png", "fileName":"testlogo.png","mimeType":"image/png"}' "http://localhost:8080/api/datasets/:persistentId/add?persistentId=doi:10.5072/FK2/LLR3PR"`

The main difference from submitting a file is that instead of a file, you send fileName, mimeType, storageIdentifier, and md5Hash values in the jsonData parameter (which is what the S3 direct upload does as well.)

In the db, the storageIdentifier is expanded to : 
`trsa://1752915917f-637c086e157e//themes/custom/qdr/images/CoreTrustSeal-logo-transparent.png `
which includes the storageIdentifier used by the underlying file store.

normal download works, and setting:
`bin/asadmin create-jvm-options -Ddataverse.files.trsa.download-redirect=true`
and restarting results in a redirect to:
`https://qdr.syr.edu/themes/custom/qdr/images/CoreTrustSeal-logo-transparent.png?until=2020-10-15T00:58:32.487&method=GET&token=a4d2115eb0b4a6e762455e45e3ed62ffd0e88bcccf1a9d5142f39d34b98b48d194ff561f5b6cec39592032adbc6ece736202008966202fd3f63e1c660951be09`
Removing the token and rehashing the URL up to and including "&token=" appended with the secret key results in the same SHA512 token. Using it for a GET and prior to the time shown, would represent a valid URL. (The extra params are ignored at QDR as the URL is public and they haven't implemented a validator. - It was just a convenient URL to use).

Testing aux files and S3 next.

qqmyers:
Aux files work with a file store:
image
- I think this means the current PR works as intended for file backing stores if anyone would like to try it out.
Still looking for S3 issues _ I expect there are some.

qqmyers:
S3 is also working now. That's everything that was originally defined.

landreev (commented):
Sorry it took me a while to figure out what's going on this PR. I originally assumed it would be a very bare bones read-only storageIO - but it does all this cool extra stuff for the base store overlay and aux. files, etc. Very nice. 
I'm ready to pass it to Kevin (just added a few words to the QA section). The only question I wanted to ask: Would it make sense to add support for basic http auth? (Would be very easy; username/password could be configured as extra jvm store options). That way a non-public url could be imported. The Dataverse then assumes the responsibility for managing access authorization etc... But I understand if the answer is we don't want to support that (I guess you can say that access to anything restricted is a TRSA case - ?)

qqmyers:
I think the only concern with basic auth would be that implementing direct-download would have to expose the auth info/give permanent access. If that were disabled as an option, so download only involved the Dataverse server using the basic authinfo and then passing the data on to the user, I think it could work. 

I guess the question from that is whether there are use cases where that makes sense (versus - if you're using http for large data, having the download route through Dataverse would defeat the purpose) - perhaps some of the more general harvest/present data managed in other systems use cases could use this.

In any case, I don't think it's a big technical issue to implement it. If there's someone who needs it now, I think it could be added now. If not, since I don't think it would change the way things work for public/signed resources, it could probably be added later in another PR.

landreev:
I definitely didn't mean including basic auth direct download redirects... It would only be for internal access (for non-direct downloads, for thumbnail generations, etc.). Direct downloads, if enabled, would require the user obtaining the authorization with the remote source; or the remote source trusting the tokens issued by the dataverse... 
But, regardless, I'm happy with not addressing this in the PR.

landreev (approved):
Very cool.

landreev:
I have moved the PR. 
I keep thinking how this idea of an overlaying one storage type with another, "base" storageIO could be useful a standard mechanism... Something you could enable with other non-local storage types. 
(for example, as an option for keeping the files on s3, but storing regenerate-able aux files, thumbnails and such, on the local filesystem)

kcondon:
Tested with local file and s3 direct and indirect base stores, using Jim's examples, and verified algorithm worked using Jim's method of sha512 hashing the url with token removed and secret appended.

djbrooke:
Thanks @kcondon. This one will sit for a little while we move forward discussions on TRSAs.

qqmyers:
Fixed a bug found at Odum w.r.t. using an overlay store and having a dataset with a separate logo.

Also refactored code and updated to support #7901

qqmyers:
Reported from Odum - the ingest process uses the backupAsAux method to move the original file to a new location (currently same store with .orig appended to the name) which doesn't work for an overlay store as the overlay isn't writable. A one-off fix would be to make the overlay store do a virtual switch (return the original remote file URL when the .orig aux file is requested and allow the attempt to write a .tab file replacement against the base store and then return that when the main file is requested going forward.) A better solution might be to flip the logic to not move the original at all, but that would require renaming all ingested files in storage, etc. Alternately, we could just disallow ingest for overlay files entirely, though it might still be nice to allow remote ingest and for the remote store to provide the tab and .orig formats, with all other aux files going to the base store.) As the Data Commons project intends to use this for remotely hosted files available via Globus, it would be good to get consensus on what the best approach for handling ingest with the overlay store would be. (Nominally, with Globus used for very large files, ingest will probably be turned off anyway, so a natural choice could be to turn off ingest. However, if one of the other options is better for the community, it might make sense to go in that direction (and then rely on ingest size limits to avoid ever ingesting in the Globus case).

akio-sone:
The line 489 of fileFragment.xhtml of the commit [edc9152](https://github.com/IQSS/dataverse/pull/7325/commits/edc915254c915cc7ec6e8f8e8ce75fd3b9658053), column 61(after toString()), ";", causes a parsing error `com.sun.el.parser.ParseException: Encountered "}" at line 1, column 61.` and fails to render the Dataset page when a link to the Dataset page is clicked; Netbeans 13.0 editor also red-flags this semi-colon as an error.  After removing this semi-colon this parsing error disappears.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -663,6 +665,34 @@ Migrating from Local Storage to S3
...
 Is currently documented on the :doc:`/developers/deployment` page.
 
+Trusted Remote Storage
+++++++++++++++++++++++
+
+In addition to having the type "remote" and requiring a label, Trusted Remote Stores are defined in terms of a baseURL - all files managed by this store must be at a path starting with this URL, and a baseStore - a file, s3, or swift store that can be used to store additional ancillary dataset files (e.g. metadata exports, thumbnails, auxiliary files, etc.).
+These and other available options are described in the table below.
+
+Remote stores can range from being a static trusted website to a sophisticated service managing access requests and logging activity
+and/or managing access to a secure enclave. For specific remote stores, consult their documentation when configuring the remote store in Dataverse.
```
pdurbin, line 675:
```suggestion
and/or managing access to a secure enclave. For specific remote stores, consult their documentation when configuring the remote store in your Dataverse installation.
```

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -663,6 +665,34 @@ Migrating from Local Storage to S3
...
+    dataverse.files.<id>.baseUrl                 <?>                 **Required** All files must have URLs of the form <baseUrl>/*               (none)
+    dataverse.files.<id>.baseStore               <?>                 **Required** The id of a base store (of type file, s3, or swift)            (none)
+    dataverse.files.<id>.download-redirect       ``true``/``false``  Enable direct download (should usually be true).                            ``false``
+    dataverse.files.<id>.secretKey               <?>                 A key used to sign download requests sent to the remote store. Optional.    (none)
+    dataverse.files.<id>.url-expiration-minutes  <?>                 If direct downloads and using signing: time until links expire. Optional.   60
+    dataverse.files.<id>.remote-store-name       <?>                 A short name used in the UI to indicate where a file is located. Optional   (none)
+    dataverse.files.<id>.remote-store-url        <?>                 A url to an info page about the remote store used in the UI. Optional.      (none)
```
pdurbin, line 691: A couple of these descriptions don't end in a period.

pdurbin, line 691: I can live with it but it seems weird that we have both kebab-case and camelCase in this table of JVM options. If we can pick one, great.

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
```diff
@@ -3044,7 +3048,7 @@ public void saveAdvancedOptions() {
     }
 
     public boolean rsyncUploadSupported() {
-        // ToDo - rsync was written before multiple store support and currently is hardcoded to use the "s3" store. 
+        // ToDo - rsync was written before multiple store support and currently is hardcoded to use the DataAccess.S3 store.
```
pdurbin, line 3051: I'm a little confused by this comment. We built rsync support for (and with) SBGrid and they use the "file" store. Later we added the ability to use rsync/DCM with the "s3" store but I don't think any installation is actually running in that mode.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2371,40 +2371,40 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
         String newFilename = null;
         String newFileContentType = null;
         String newStorageIdentifier = null;
-		if (null == contentDispositionHeader) {
```
pdurbin, line 2374: Thank you for replacing all these tabs with spaces!

src/main/java/edu/harvard/iq/dataverse/dataaccess/DataAccess.java
```diff
@@ -42,9 +44,16 @@ public DataAccess() {
 
     };
 
+    public static final String FILE = "file";
+    public static final String S3 = "s3";
+    static final String SWIFT = "swift";
+    static final String REMOTE = "remote";
```
pdurbin, line 50: At some point, should we consider grouping these into an enum?

src/main/java/edu/harvard/iq/dataverse/dataaccess/RemoteOverlayAccessIO.java
```diff
@@ -0,0 +1,561 @@
...
+            .setCookieSpec(CookieSpecs.STANDARD).setExpectContinueEnabled(true).build();
+    private static boolean trustCerts = false;
+    private int httpConcurrency = 4;
+
+    public RemoteOverlayAccessIO(T dvObject, DataAccessRequest req, String driverId) throws IOException {
+        super(dvObject, req, driverId);
+        this.setIsLocalFile(false);
+        configureStores(req, driverId, null);
+        logger.fine("Parsing storageidentifier: " + dvObject.getStorageIdentifier());
+        // TODO: validate the storage location supplied
```
pdurbin, line 83: Is this still a valid TODO?

src/main/java/edu/harvard/iq/dataverse/dataaccess/RemoteOverlayAccessIO.java
```diff
@@ -0,0 +1,561 @@
...
+        urlPath = dvObject.getStorageIdentifier().substring(dvObject.getStorageIdentifier().lastIndexOf("//") + 2);
+        logger.fine("Base URL: " + urlPath);
+    }
+
+    public RemoteOverlayAccessIO(String storageLocation, String driverId) throws IOException {
+        super(null, null, driverId);
+        this.setIsLocalFile(false);
+        configureStores(null, driverId, storageLocation);
+
+        // TODO: validate the storage location supplied
```
pdurbin, line 93: Is this also still a valid TODO?

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
```diff
@@ -457,12 +457,12 @@ public static List<DatasetField> getDatasetSummaryFields(DatasetVersion datasetV
     }
     
     public static boolean isAppropriateStorageDriver(Dataset dataset){
-        // ToDo - rsync was written before multiple store support and currently is hardcoded to use the "s3" store. 
+        // ToDo - rsync was written before multiple store support and currently is hardcoded to use the DataAccess.S3 store.
```
pdurbin, line 460: Please see my comment above about rsync and S3.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
```diff
@@ -276,63 +276,6 @@ public JsonObjectBuilder toJson() {
         return jab;
     }
 
-    public enum ReservedWord {
```
pdurbin, line 279: Why are we moving ReservedWord around? What does it have to do with remote stores? (I'll know by the end of my initial scan of the code, I hope.)

src/test/java/edu/harvard/iq/dataverse/util/UrlTokenUtilTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertEquals("42 test en", urlTokenUtil.replaceTokensWithValues("{fileId} test {localeCode}"));
+        assertEquals("42 test en", urlTokenUtil.replaceTokensWithValues("{fileId} test {localeCode}"));
+        
+        assertEquals("https://librascholar.org/api/files/42/metadata?key=" + apiToken.getTokenString(), urlTokenUtil.replaceTokensWithValues("{siteUrl}/api/files/{fileId}/metadata?key={apiToken}"));
+        
+        URLTokenUtil urlTokenUtil2 = new URLTokenUtil(ds, apiToken, "en");
+        assertEquals("https://librascholar.org/api/datasets/50?key=" + apiToken.getTokenString(), urlTokenUtil2.replaceTokensWithValues("{siteUrl}/api/datasets/{datasetId}?key={apiToken}"));
+        assertEquals("https://librascholar.org/api/datasets/:persistentId/?persistentId=doi:10.5072/FK2ABCDEF&key=" + apiToken.getTokenString(), urlTokenUtil2.replaceTokensWithValues("{siteUrl}/api/datasets/:persistentId/?persistentId={datasetPid}&key={apiToken}"));
+    }
+}
```
pdurbin, line 50: I'm happy to see this test about URL tokens but it is possible to have some tests of the remote web store?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -663,6 +665,34 @@ Migrating from Local Storage to S3
...
+    
+    ===========================================  ==================  ==========================================================================  =============
+
```
pdurbin, line 694: The docs are looking good overall but I'm slightly twitchy about so many variants on what we're calling this thing:

- trusted remove services
- trusted remote web-accessible system
- trusted remote storage
- trusted remote stores
- remote store

Maybe these are all legit. Maybe we can consolidate a bit. Not sure. I still dream of a glossary (#4274). 😄

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -238,13 +238,15 @@ As for the "Remote only" authentication mode, it means that:
 - ``:DefaultAuthProvider`` has been set to use the desired authentication provider
```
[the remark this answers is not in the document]
qqmyers: #8894

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
qqmyers: This note dates from https://github.com/IQSS/dataverse/pull/6490 which was based on seeing the 's3' hardcoded entries from https://github.com/IQSS/dataverse/pull/5359/files in the code. Since I didn't know how to run rsync, I just maintained the dependence on s3 as the name and added the note.
At this point, I can't easily tell if the rsyncUploadSupported method is only called with the S3 variant (in which case the note is ~true but only w.r.t. to the S3 variant of rsync/DCM) or if it would also be called when using a file store (in which case the DatasetUtil.isAppropriateStorageDriver is probably incorrectly returning false and might either need to add the 'file' store or any store of type 'file' if there isn't a similar hardcode of the name 'file' somewhere.
In any case - this PR is just changing s3 from a string to a constant, so I'd suggest a separate issue if this should be pursued.

src/main/java/edu/harvard/iq/dataverse/dataaccess/DataAccess.java
qqmyers: Probably makes sense - I'll try to remember as I look into the larger refactoring I brought up in a recent tech hour.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
qqmyers: It isn't directly. Since the URL signing is getting re-used in OpenDP, I was trying to keep in sync with what I gave Bob to avoid any inconsistencies wr.t. to the URlSigner/URLToken classes. It could probably be removed from this PR if someone can show me how to do that without adding a commit that will remove it from Bob's branch when he merges with develop after this PR.

src/test/java/edu/harvard/iq/dataverse/util/UrlTokenUtilTest.java
qqmyers: Hmm - there should be some of the basic URLSigning as well. I'll look.

doc/sphinx-guides/source/installation/config.rst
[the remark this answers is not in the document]
pdurbin: That works. Thanks.

doc/sphinx-guides/source/installation/config.rst
pdurbin: Standardized to kebab-case. Thanks.

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
pdurbin: Right. Looks like the comment was added in 6c506ab5 as part of #6488.

Let's not worry about this comment.

Obviously, rsync/DCM support is becoming a bit of a maintenance burden. If anyone is still using it, we should probably decide whether to breathe new life into it or rip it out.

Like you said, we can revisit all this in a future issue and PR.

src/main/java/edu/harvard/iq/dataverse/dataaccess/DataAccess.java
pdurbin: Sounds good. Future refactoring works for me. I just wanted to bring it up while I was looking at it.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
pdurbin: I resolved the comment/conversation above with a note. Resolving this one too.

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
pdurbin: My concern has to do with external tools. How much risk is there that we are breaking something in this PR? How much QA should we do? If we keep this change, should the "how to test" for this PR be updated to explain what to test?

With regard to git stuff, we could perhaps ask @adaybujeda for advice. His PRs often have a single commit. I assume he's squashing his commits (something I haven't played with) and rebasing. My dumb way of doing this (getting down to a single commit) is to create a new branch.

Let's first decide if we want the change in or out. Do we want to mess with the external tools code? Then let's decide what if anything to do in git for this branch/PR.

doc/sphinx-guides/source/installation/config.rst
pdurbin: I see some consolidation of terminology was done. Thanks!

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
pdurbin: @qqmyers and I decide to leave the external tools change in (it's just an automated refactoring, he said). He updated "how to test" to explain the external tools need to be regression tested.

The debate about rebasing continues in #5379 😄 ... but I'm resolving this thread.

src/test/java/edu/harvard/iq/dataverse/util/UrlTokenUtilTest.java
pdurbin: It looks like we have UrlSignerUtilTest now. Thanks.

Also, I added RemoteStoreIT but it's really just for developers to run. I don't think our testing setup allows JVM options to be configured, etc to have this test be part of the API test suite.

pdurbin:
@qqmyers and I have been talking a lot over Slack and he's been fixing things.

To summarize the state of things from my perspective:

- When `dataverse.files.<id>.download-redirect` is set to "true", we get a 500 error and a broken image on the file listing on the dataset page (screenshot below). My understanding is that enabling download redirect like this will often be desired so the remote file is not streamed through Dataverse.
- When the store is not configured, attempting to use the store when adding a file now fails, as desired. The message doesn't say that the store isn't configured, but it's something. This is how the message reads: `"status":"ERROR","message":"You must upload a file or provide a storageidentifier, filename, and mimetype."}`

Screen Shot 2022-08-09 at 10 01 02 AM

qqmyers:
FWIW: The first issue in https://github.com/IQSS/dataverse/pull/7325#issuecomment-1209450686 was due to the direct download setting of the remote store being assumed to be the same for the base store and aux files like thumbs. In PR #8891, this error was caught - ~fixed as part of enabling Globus downloads. I've gone ahead and made the check for direct download to now be aux object specific (in anticipation of remote stores like TRSA that can provide their own thumbs, etc.) as a way to fix the issue here and to be more flexible in #8891.

akio-sone:
This is a question concerning the publication of a Dataset: 

(1) does PR provide a safe framework for a mixed(remote/local)-storage Dataset, i.e., one of its DataFiles is local whereas the remaining ones are remote; then (2A) if this mixing is allowed, the following check in `FinalizeDatasetPublicationCommand#execute()` would be problematic because of the call of its `validateDataFiles()` method that tests whether a file locally exists. (2B) if this mixing is **NOT** allowed, i.e., all remote or all local, for a Dataset whose DataFiles are not local, the above DataFile-check would be problematic. 

To avoid this local-file check, is there a method that flags whether a Dataset has a remote DataFile or is there a method that tells whether a DataFile is not local?  While I have checked Dataset and DataFile classes, I could not find a ready-to-use method that can play such a role: for Dataset,  getStorageDriverId() method,  and for DataFile, `getStorageIO()` => `StorageIO#isLocalFile()`, these existing methods can be used for creating such convenience methods, though.

qqmyers:
I don't see any check that the file is local. Validation does require getting the file bytes though and, as of this PR, if a remote store does not allow Dataverse to get the bytes, validation, if enabled, would fail. The settings related to FileValidationOnPublish, and/or the settings to limit validation by file size or aggregate dataset size would have to be used to avoid that for now. File uploaded to the base store of a remote store would not have a problem validating those is the same as for any file or S3 store. 

If I've missed something, let me know.

One thought for a future improvement is to allow a store to be marked as 'not readable by Dataverse' which would make it simpler to turn off ingest, thumbnail creation, file type determination by content, validation, full-text-indexing, etc. just for such stores.

akio-sone:
>The settings related to FileValidationOnPublish, and/or the settings to limit validation by file size or aggregate dataset size would have to be used to avoid that for now. 

Since `FileValidationOnPublishEnabled` is set to `true` by factory default, it would be better to add a note about this issue to RSO-related sections within User Guides and release note of version 5.12.

pdurbin:
> When `dataverse.files.<id>.download-redirect` is set to "true", we get a 500 error and a broken image on the file listing on the dataset page (screenshot below). My understanding is that enabling download redirect like this will often be desired so the remote file is not streamed through Dataverse.

This is fixed. The 500 error is gone and the image looks good:

Screen Shot 2022-08-10 at 12 07 47 PM

> When the store is not configured, attempting to use the store when adding a file now fails, as desired. The message doesn't say that the store isn't configured, but it's something.

The message now says a VALID storageIdentifier is required, so it's better.

pdurbin (approved):
I'm approving this because I'm happy with where the branch is. @qqmyers fixed some final issues that were bothering me.

Also, the tests are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-7325/122/testReport/

I'm going to add one more commit to delete some cruft from the tests I added but (again) the run above passed so we're good in terms of all API tests passing.

pdurbin:
> it would be better to add a note

@akio-sone thanks, I talked to @qqmyers and he plans to add a note.

PR #7333, opened 2020-10-15
Fix for the json LD export error in the dataset page

### LINKED ISSUE #7310: Dataset page: null pointer trying to export LD Json metadata
Dataset page inserts "LD Json" (aka "schema.org") metadata fragment, for published datasets. This format is normally already exported and cached; so we  never noticed the issue below: 

If the fragment is not cached, export is called to generate it on the fly. And then it fails, with a null pointer below. Note that this is uniquely specific to calling this export from the page while it's loading. Requesting the format using the API works and caches the format, fixing the issue. 

This is likely caused by the fact that during the initial loading of the page we are using some optimized, minimalist way of initializing the dataset. So it loads without all the entities fully instantiated from the database yet. (The null pointer from DataFile.java:456 suggests the FileMetadata objects are not yet attached to the DataFiles). 

Depending on how much effort this would take to fix, we may just choose to leave it as is - but intercept the exception in a way that does not kill the page. So basically just live with the fact that for some published datasets this metadata fragment is not going to be included until it's exported elsewhere.

landreev:
A fix for the optimized/hacky method that quickly populates all the DataFiles in a Dataset with custom queries.  This addresses the json LD export issue found in #7310.

(I'm going to open a general issue for adding a section to the dev. guide - a checklist for anytime we modify certain entities. For example, custom queries in this method may need to be modified again if we add any more fields to the datafile-level entities. For most other entities, the json export and import methods may need to be modified. There may be more such places?)

Closes #7310

See the issue on how to reproduce. 
The error was thrown specifically when retrieving the original name of an ingested tabular file. So a super prudent way to test this would be with both a file ingested since 5.0 (for which the original filename is saved in the database), and for a file ingested earlier. In a new database the latter can be emulated by a direct db update: 
`UPDATE datatable SET originalfilename=NULL WHERE datafile_id=...;`.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
@@ -578,7 +578,7 @@ public class DataFileServiceBean implements java.io.Serializable {
         
         int i = 0; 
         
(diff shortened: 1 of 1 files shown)


qqmyers (approved):
Looks good.

PR #7334, opened 2020-10-16
IQSS/7068 Reserve File Pids

### LINKED ISSUE #7068: Reserve file-level DOIs?
At QDR, we're thinking of enabling reserving DOIs for files prior to publication. 

**User Story:** As a researcher, I want to refer to a data file (think a specific interview or document) in a manuscript. I _don't_ want to publish the data project until the manuscript is published and I want to avoid having to replace all file placeholders with DOIs prior to publication.

**Implementation considerations:** Given that files are often deleted and DOIs never published, we don't want to reserve those DOIs with datacite but rather reserve them internally, the way project DOIs are currently reserved in Dataverse. 
Given the fact that files can be deleted (and, in fact, commonly are deleted and replaced in our curation process) we might want to consider making this available via a manual option rather than always (which may lead to misunderstandings and non-existing DOIs ending up in citations); on the other hand, auto-generating them is a cleaner solution with no additional UI requirement.

**Question:** Would this be something that IQSS would consider for the core dataverse code and if so, what are the main concerns that we'd need to address? (I didn't go back to look through the file DOI discussion to see what were the considerations there, so may just require a link).

CC @qqmyers

qqmyers:
**What this PR does / why we need it**: Reserves file PIDs at creation time, if enabled, analogous to how datasets are now handled.

Closes #7068 

**Suggestions on how to test this**: Should be able to see file PIDs/DOIs after file creation in the datacite console (e.g. doi.test.datacite.org). Deleting a file, or the dataset, prior to publication should remove the draft PID. Nominally this code calls whatever PID provider you use, so it should work with other providers capable of reserving PIDs. I'm not sure it is necessary to test them though - if the PID provider works, and this code works with DOIs, it should also work for those others.

**Additional documentation**:


djbrooke:
Adding this to code review, because I think it is desirable and non-controversial functionality for those who have turned on File PIDs. 

I'd be interested in thoughts from the Harvard Dataverse team as we've seen some issues with long running jobs around PID registration, and I wonder if this would potentially make it worse since there's now two steps (reservation and flipping to public) instead of one.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -221,4 +223,79 @@ public void setDataset(Dataset dataset) {
...
+
+    protected void registerFilePidsIfNeeded(Dataset theDataset, CommandContext ctxt, boolean b) throws CommandException {
+     // Register file PIDs if needed
+        String protocol = theDataset.getProtocol();
+        String authority = theDataset.getAuthority();
+        GlobalIdServiceBean idServiceBean = GlobalIdServiceBean.getBean(protocol, ctxt);
+        String currentGlobalIdProtocol = ctxt.settings().getValueForKey(SettingsServiceBean.Key.Protocol, "");
+        String currentGlobalAuthority = ctxt.settings().getValueForKey(SettingsServiceBean.Key.Authority, "");
+        String dataFilePIDFormat = ctxt.settings().getValueForKey(SettingsServiceBean.Key.DataFilePIDFormat, "DEPENDENT");
+        boolean shouldRegister = ctxt.systemConfig().isFilePIDsEnabled()                                  // We use file PIDs
```
pdurbin, line 235: I thinking about how we iterate though all the files below. Couldn't we short circuit that iteration and simply return early if file level PIDs are not enabled?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -221,4 +223,79 @@ public void setDataset(Dataset dataset) {
...
+                && !idServiceBean.registerWhenPublished()                                                 // The provider can pre-register
+                &&((currentGlobalIdProtocol.equals(protocol) && currentGlobalAuthority.equals(authority)) // the dataset PID is a protocol/authority Dataverse can create new PIDs in
+                        || dataFilePIDFormat.equals("INDEPENDENT"));                                      // or the files can use a different protocol/authority
+        logger.fine("IsFilePIDsEnabled: " + ctxt.systemConfig().isFilePIDsEnabled());
+        logger.fine("RegWhenPub: " +  !idServiceBean.registerWhenPublished());
+        logger.fine("OK provider: " + ((currentGlobalIdProtocol.equals(protocol) && currentGlobalAuthority.equals(authority)) // the dataset PID is a protocol/authority Dataverse can create new PIDs in
+                        || dataFilePIDFormat.equals("INDEPENDENT")));
```
pdurbin, line 242: I know it's more code but rather than having hardcoded strings like "INDEPENDENT" my preference is to reference the place where the string comes from like this: SystemConfig.DataFilePIDFormat.INDEPENDENT.toString()

That way if you do a "find usages" in your editor, it easily comes up. And if we ever change the string (which is doubtful), it's easier to find.

This might just be a me think. Please feel free to get opinions from others.

pdurbin (changes requested):
Overall, this looks fine. I'm only choosing "request changes" because I think we need a release note. I'm also suggesting a couple code changes but I don't see either one as showstoppers.

djbrooke:
Thanks @qqmyers and @pdurbin for talking with me about this. I made release note additions and API doc edits because I thought we recommended reservation for file PIDs similar to how we did reservations for Dataset PIDs when we moved to the new pre-publish verification workflow in 5.0 (https://github.com/IQSS/dataverse/blob/develop/doc/release-notes/5.0-release-notes.md#dataverse-installations-using-datacite-upgrade-action-recommended). I backed these changes out as we don't need to reserve PIDs for existing files and the API does not currently support this. This does not result in a failure during publishing, it just means that those File PIDs will be registered at that time.

pdurbin:
@djbrooke the release note looks good.

@kcondon I'd suggest testing with many files to see how it performs.

kcondon:
Tested, functions as expected with caveat of increasing proxy timeout for large number of files.

djbrooke:
Since this potentially introduces a long wait earlier in the dataset creation workflow, I'm going to put this on ice until we dig into #7393.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1238,14 +1238,6 @@ public List<Long> selectFilesWithMissingOriginalSizes() {
...
     
 
-    /**
-     * Check that a identifier entered by the user is unique (not currently used
-     * for any other study in this Dataverse Network). Also check for duplicate
-     * in the remote PID service if needed
-     * @param datafileId
-     * @param storageLocation
-     * @return  {@code true} iff the global identifier is unique.
-     */
```
qqmyers, line 1248: Not related to the PR - this just doesn't fit the method below which doesn't involve PIDs

landreev:
Looks good overall. 
Could you please sync the branch and resolve the conflict(s). 

What is the back story, was there a specific reason why it got stuck and sat for 4 years? Was it just because of the concerns about file uploads becoming slower?

qqmyers:
As far as I know, performance was the only potential issue - could still be an issue (although QDR uses it), but easier to manage overall since file PIDs can be enabled/disabled as desired now.

landreev (approved):
(re-)approving, in order to move the PR along into "ready for QA".

PR #7336, opened 2020-10-16
Iqss/7335 change name when replacement name conflicts with other files

### LINKED ISSUE #7335: Check for duplicate names when replacing a file and choosing the name of another
Per https://github.com/IQSS/dataverse/issues/7223#issuecomment-690582649 , the earlier #7250 fix to allow replacing a file and using the same filename as the original did not check for the case when the name of a new replacement file has the same name as some other file in the dataset, in which case the name still should change (e.g. have "_1" appended).

qqmyers:
**What this PR does / why we need it**: Prevents duplicate names if a replacement file has the same name as some other file (on it isn't replacing) in the dataset

Closes #7335 

**Suggestions on how to test this**: There's a new test and it passes! The new part is if you replace file1 and give it the name of file2, you'll see the normal "_1" appended. If you replace file1 and use the same name as file 1, the name won't change.

**Additional documentation**:


sekmiller (approved):

sekmiller:
This branch needs to be updated with the latest from dev before it can go to QA, but that is all it needs

djbrooke:
@qqmyers assigning to you to update from dev

Issue #5847, opened 2019-05-15
Deleting a file from the file page gives no feedback. Did it work?


pdurbin:
Tested on https://demo.dataverse.org running v. 4.13 build 1309-a68b59d if I upload a file and then try to delete it from the file page, I get no feedback. Here are screenshots:

## Click Delete from file page

![Screen Shot 2019-05-15 at 12 06 05 PM](https://user-images.githubusercontent.com/21006/57791102-10fe7280-770a-11e9-9ed9-379dc565d43c.png)

## Confirm delete

![Screen Shot 2019-05-15 at 12 06 11 PM](https://user-images.githubusercontent.com/21006/57791103-10fe7280-770a-11e9-8084-84ccdb1da0be.png)

## Bug: no feedback. Did it work? Was the file deleted? Why am I still staring at it?

![Screen Shot 2019-05-15 at 12 06 18 PM](https://user-images.githubusercontent.com/21006/57791104-11970900-770a-11e9-9108-35cc54bc8e59.png)

I don't have access to server.log on demo but I'm going to suspect it's the same stack trace as in #5842.

djbrooke:
Ah, this bug only happens in certain cases (don't they all?). Good news, file is deleted, bad news, as the title says, is no feedback. 

Steps that I took to reproduce this:

- Created a new test dataset in demo
- Added a file
- Went to the file page
- Deleted the file 
- No feedback, no navigation to the dataset page

Steps where this worked as expected:

- Created a new test dataset in demo
- Added a file
- Published the dataset
- Went to the file page
- Deleted the file 
- Taken back to the dataset page with a green message that the file was successfully deleted

sekmiller:
In addition to the lack of feedback the under the covers manifestation of this bug is that files that have been deleted via the file page have not had their physical files deleted. It's difficult to assess how widespread this may be or how to find these physical files and delete them manually. (or if it's worth going through the exercise.)

sekmiller:
@qqmyers  when we were testing this issue another issue was discovered wherein if you attempt to delete a published file from a draft version of a dataset via the file page, the delete fails and the dataset is left in "locked for edit in progress" state.  The problem is when you try the second merge at around line 179 (UpdateDatasetVersionCommand) you get a java.lang.IllegalArgumentException: Cannot merge an entity that has been removed: edu.harvard.iq.dvn.core.study.FileMetadata 

[...]

Issue #3913, opened 2017-06-15
Is a delete file endpoint available?


leeper:
The native API currently allows file update (and now file replace) operations but it doesn't appear to allow file delete. This means that because replace only works on *published* files, it is not possible to update the file contents of a draft dataset without going through the GUI to manually delete the staged file or to delete the entire draft (all files and metadata) through the API. It would be great to be able to delete both unpublished/draft files and to delete files generally from a version (in the way that is currently possible through the GUI).

pdurbin:
@leeper well, as I mentioned in person today, you could use SWORD to delete a file but you're right that there's no "native" API to delete a file. See also #3419 and for all I know #2464 is still a bug. #2222 is also related.

pdurbin:
@leeper thanks for your feedback. I copied your comment into #3419 where the work would be done and am closing this issue.

pdurbin:
On second thought, I like the way this issue is worded and how it's from a real user rather than me so I'm re-opening it in favor of #3419, which I just closed as a duplicate.

qqmyers:
FWIW: #7149 is about adding the ability to replace files in draft versions, so it addresses the original use case above. (working on a PR)

janvanmansum:
I am interested in the last use case "deleting files from a (draft) version similar to the UI". Would it make sense to have a new API endpoint similar to "replace" but which deletes the file from the current draft (and creates a draft if the last version is a published version)?

I imagine calling that endpoint would look like this for persistent IDs. 

```
export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
export SERVER_URL=https://demo.dataverse.org
export PERSISTENT_ID=doi:10.5072/FK2/AAA000

curl -H "X-Dataverse-key:$API_TOKEN" -X POST  "$SERVER_URL/api/files/:persistentId/delete?persistentId=$PERSISTENT_ID"
```

And for database IDs:
```
export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
export SERVER_URL=https://demo.dataverse.org
export ID=5

curl -H "X-Dataverse-key:$API_TOKEN" -X POST  "$SERVER_URL/api/files/$ID/delete"
```

pdurbin:
[...]

PR #7337, opened 2020-10-16
Allow replacing a draft file

### LINKED ISSUE #4380: As a researcher, when I replace a file, I want the file metadata of the replaced file to be used for the new file, so that I don't have to re-enter the file metadata
I've come across ~~two~~ a possible issue~~s~~ with file replace
~~I used the "file replace" feature to replace a file with a new one with the same name. The new file was given the "-1" suffix once it is uploaded.  Logically this file will be replacing the previous file and won't coexist with the original file, so the "-1" suffix isn't necessary.~~

When replacing a file the original file's metadata doesn't not carryover. I would expect the files description and tags to be the same as the original file since I'm replacing the file. There is also no warning that the metadata and tags will have to be reentered. 

I feel like if I have to reenter the metadata and manually rename the replaced file, I would have just deleted the original file and uploaded the new file.

### LINKED ISSUE #7149: Support replacing files while in draft
If I understand correctly, Dataverse's replace file capabilities are only possible post-publication and not within a draft Dataset. 

QDR's curation workflow involves replacing files in the draft Dataset (e.g. to remove PII) and, in connection with #7068, there is interest in being able to replace a file in the draft version/prior to publication without changing the PID reserved for the Datafile. I'm considering modifying the existing replace mechanism to work in a draft Dataset, probably deleting the old file unless it is used in a  previously published version. Is there interest in this in the community? Any design guidance/ alternative mechanisms/ other feedback?

(Looking at other issues - might be able to look into #4380 at the same time.)

qqmyers:
**What this PR does / why we need it**: Enables replacing a file in draft, keeping the file metadata except label by default (user can then edit before saving if desired). (Note - not using the old file name is both because it is confusing and because one of the use cases is to, for example, replace a doc file with a pdf-a version that needs a different extension to get proper mimetyping).

Closes #7149
Closes #4380

**Special notes for your reviewer**:  Also ties to #3913. There was some discussion about using a different name instead of 'replace' for this since the effect is slightly different for draft files when there is no prior released version (the new file gets the DOI of the old file because the old one is fully deleted, versus the new file getting a new DOI). Thinking through this further, it seems to be weirder to me to have ~replace on a draft file get called two different things depending on whether a released version exists, and it seems fairly natural to me for the DOI to be reused when it can. I also don't know what else I would call this. If there's still interest in making this a separate call and there's a name suggestion, I can update the PR.

**Suggestions on how to test this**: Replace of published files should work as before, but it should be possible to replace a file in a draft version as well. When doing so, metadata including description, tags, categories, path, on the file should remain by default (and be editable prior to hitting save as normal). Replacing a file that is only available in draft should delete the replaced file entry in the db and remove the physical file file from storage. The new file should retain the DOI of the one removed. Replacing a file in draft that has previously been released should not remove the old file (still available in prior versions) and it should get a new DOI.

Should work in both UI and API.

Should also text that https://github.com/IQSS/dataverse/issues/5847#issuecomment-497716243 doesn't reappear as this PR reverts one of the changes made in that issue (hopefully compensating to address the problem in a different way).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: replace option available for draft files, file metadata is retained from replaced file by default.

**Additional documentation**:


qqmyers:
OK some notes on integration test failures:

A couple were related to the PR code (a bug and a test that expected an error when trying to replace a draft file which now succeeds). The rest appear to be unrelated to the PR code (which was fun to debug since other branches appear to pass the tests - in some cases that appears to be timing and in others because build failures lead to IT tests not running without the green/red pass/fail indicator turning red.)

A couple appear to be intermittent and probably related to timing: a second call to update a dataset fails when the same API call works manually. E.g. test_006_ReplaceFileGood  and testForceReplaceAndUpdate in a recent build. I suspect that these need the same type of UtilIT.sleepForLock() calls that some of the ingest methods have, though my guess would be that the delay would be shorter. (testCuratorSendsCommentsToAuthor also failed that build with a wait for lock error so things were definitely slow).

Two others appear to be cases where the code and tests had ~matching minor errors (in 5.1.1 but the code in question is several years old so probably before that as well). In testAddFileBadJson and testReplaceFileBadJson, sending some types of bad json (such as a simple string in the tests) resulted in a 200/OK status and the file being added when it presumably should have failed due to bad json. The code was catching, logging, and ignoring a ClassCastException to do that. Other types of bad json, such as "{}:string" did fail but weren't tested. In the updated code, I stopped catching the ClassCastException in OptionalFileParams and then caught both that and the possible json parsing exception in the same clause in the calling classes to generate a 400 error (now with a Bundle message!). The behavior of the code and tests should match.

qqmyers:
OK, after significantly more work this week, I think this is ready to go. To get the new and existing functionality to all work, I had to dig back into the merge issues brought up in #5847 and I think I understand more, if not all of the problems that could occur in that code/ all of the different input states (in terms of whether filemetadata, datafiles, datasetversions are persisted, merged, not merged coming into the code). The rewrite of the file delete code has more commenting that can hopefully guide anyone making changes after this.

To get all of the tests to work, I had to do some further tweaking to the sleepForLock code to handle null for the lock type. In doing that I realized that when the replace api call fails due to a lock it appears to still send an OK, but fails the test due to differences in what's in the "data" part of the response. 

There, and a few other places, I've added a ToDo - if any of those are worth adding an issue for, they could probably be small ones.

djbrooke:
Thanks @qqmyers - moving to code review

src/main/webapp/file.xhtml
```diff
@@ -267,7 +267,7 @@
                                                                 </li>
                                                             </ui:fragment>     
                                                         </ui:fragment>
-                                                        <ui:fragment rendered="#{FilePage.fileMetadata.dataFile.released and FilePage.draftReplacementFile == false and !FilePage.fileMetadata.dataFile.filePackage}">                                        
+                                                        <ui:fragment rendered="#{FilePage.draftReplacementFile == false and !FilePage.fileMetadata.dataFile.filePackage}">                                        
                                                         <li><!-- start: replace file link -->
```
sekmiller, line 271: We're getting multiple replace file links here because of the change in render. Once a  released file has been replaced it can be replaced again, so taking out the "isReleased" gets two Replace links to render

qqmyers: Good catch. The problem is the next lines though - the second Replace button that would trigger the warning dialog to not replace a file that's already being replaced in the draft version was showing and it shouldn't have. For now, I've just adjusted the logic there (only show the dialog if there is a draft replacement for the file, which is not true for the replacing file.) 

The logic is simpler too. FWIW, I went back about 4 years in github to see if I could find out why the current terms were in there and didn't find a clear reason, so perhaps someone else might have a better sense and be able to see if there's anything my change misses?

However, I'm also not sure that dialog is needed anymore - a file that has been replaced in the latest version now shows a dialog when you click the 'Edit' button itself, so you never get to the individual menu items like replace and delete and therefore can't trigger the dialogs related to them. If it's true that there is no way to get to those items for a file that has been replaced/deleted in the current version, they (those replace and delete items and the dialogs they trigger) should probably be deleted (new issue?).

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
```diff
@@ -165,38 +167,77 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
+                    }
+                } else {
+                    // Not sure if this is needed now that there is a dataset merge above, but we
+                    // need to assure it is on the context
+                    fmd = ctxt.em().merge(fmd);
+                }
+                // There are two datafile cases as well - the file has been released, so we're
+                // jsut removing it from the current draft version or it is only in the draft
+                // version and we completely remove the file.
                 if (!fmd.getDataFile().isReleased()) {
```
sekmiller, line 217: Getting a null pointer here when I try to delete a file via the EditDataFilesPage from the Dataset Page. Will investigate further.

qqmyers: @sekmiller - The new commits should fix this - the test of the two dataset versions in line 208 needed to use ! .equals() instead of !=, and then an fmd merge was needed (removing the else around line 212.

src/main/java/edu/harvard/iq/dataverse/util/FileMetadataUtil.java
```diff
@@ -0,0 +1,92 @@
...
+            collection.remove(fmToDelete);
+        } else {
+            Iterator<FileMetadata> fmit = collection.iterator();
+            while (fmit.hasNext()) {
+                FileMetadata fmd = fmit.next();
+                // If not, we can remove based on a match based on the id of the related
+                // datafile
+                if (fmToDelete.getDataFile().getStorageIdentifier().equals(fmd.getDataFile().getStorageIdentifier())) {
+                    // and a match on the datasetversion
+                    if (fmToDelete.getDatasetVersion().getId() == null) {
```
sekmiller, line 48: Getting a null pointer here when I try the uncommon, but previously supported, action of uploading a number of files, but then deleting one before saving the rest to the draft version

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
sekmiller: Ok. i'm able to delete a file from a draft version. Thanks!

src/main/java/edu/harvard/iq/dataverse/util/FileMetadataUtil.java
qqmyers: Resolved - the new util method assumed that an fmd would always have an associated datasetversion which is not true for this case. - BTW - thanks for the thorough testing!

sekmiller (commented):
Looks good overall. Thanks Jim, for addressing the issues I've found. Gustavo asked me to keep it out of QA for now pending the merging of kebabs and any notes coming from design.

djbrooke:
@TaniaSchlatter at long last I've gotten the ability to spin up EC2 instances on my laptop, and I've stood up this branch on AWS. It works as I expect. A few weeks ago I had flagged this for UI/UX review as it enables to workflow to replace draft files and to retain metadata on replace, which @qqmyers has outlined above.

If you'd like to review let me know. I don't want to keep the instance up and incur cost, so I'll spin it down for now, but let me know when you're ready to take a look and I'll spin up another instance. I'll need ~20 min to get the instance going so just give me a little warning and we can step through it together or you can do so independently.

TaniaSchlatter:
Reviewed the AWS instance. It would be consistent to have Replace in the kebab on the file table on the dataset page, as well as under "Edit" on the file page. 
replace_draft_file_pages

djbrooke:
Thanks @TaniaSchlatter - that's covered in #7376 so we'll implement this with the current workflows and get the consistency on entry points across dataset kebab/file page in #7376.

djbrooke:
After some discussion, I'm going to move this to QA.

We'll reopen #7343 to add replace as a small, uncontroversial chunk towards the larger effort in #7376, which requires some research and decisions.

PR #7339, opened 2020-10-20
default guides BaseUrl to https

### LINKED ISSUE #7327: Use https for guides.dataverse.org
Right now the dataverse guides site is on http (not on https). 
Some users are concerned  about the ‘non secure’ connection that browsers indicate.
It would be better if we also have the guides on https. Although technically these are static pages and https is not needed, this is a usability thing and user perception is important.

donsizemore:
**What this PR does / why we need it**: e-mail message sent containing links to http://guides.dataverse.org are receiving a higher spam score and are being flagged as such. This PR updates the SystemConfig.java and SettingsWrapper.java saneDefault values to use https. There are many other hard-coded references to specific pages on the guides, but those should be picked up by the redirect: `RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [END,NE,R=permanent]`

Closes #7327

**Suggestions on how to test this**: build branch, click "User Guide" on homepage. Since most everything appears to use settingsWrapper.guidesBaseUrl I'm assuming this should update the URL used in outgoing e-mail messages as well.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
+++ b/src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
@@ -125,7 +125,7 @@ public class SettingsWrapper implements java.io.Serializable {
         if (true)
 
             if (guidesBaseUrl == null) {
-            String saneDefault = "http://guides.dataverse.org";
+            String saneDefault = "https://guides.dataverse.org";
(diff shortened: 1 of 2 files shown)


scolapasta (approved):
Seems simple enough.

PR #7341, opened 2020-10-20
#7097 move dv indexing to on success in Finalize Dataset Publication Command

### LINKED ISSUE #7097: in FinalizeDatasetPublicationCommand, move dataverse indexing (due to subjects) to onSuccess
In FinalizeDatasetPublicationCommand, we moved the indexing of the dataset to the onSuccess method. However, we still index the dataverse (if subjects were added) in the execute.

So what could happen is that subjects are added, the dataverse is indexed, then the transaction fails, and the index is misaligned with what the actual metadata has for dataverse subjects.

sekmiller:
As it is currently written it's possible for the Finalize Dataset Publication Command to fail in such a way that the dataverse owners have been reindexed fond subjects while the transaction to save those subjects fails. Moving the re-index to OnSuccess will prevent that from happening.

Closes #7097 in FinalizeDatasetPublicationCommand move reindex dv to OnSuccess

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
@@ -32,6 +32,7 @@ import edu.harvard.iq.dataverse.GlobalIdServiceBean;
 import edu.harvard.iq.dataverse.batch.util.LoggingUtil;
 import edu.harvard.iq.dataverse.engine.command.Command;
 import edu.harvard.iq.dataverse.util.FileUtil;
+import java.util.ArrayList;
 import java.util.concurrent.Future;
 import org.apache.solr.client.solrj.SolrServerException;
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -224,6 +227,20 @@ public boolean onSuccess(CommandContext ctxt, Object r) {
             LoggingUtil.writeOnSuccessFailureLog(this, failureLogText,  dataset);
             retVal = false;
         }
+        
+        //re-indexing dataverses that have additional subjects
+        if (!dataversesToIndex.isEmpty()){
+            for (Dataverse dv : dataversesToIndex) {
+                try {
+                    Future<String> indexString = ctxt.index().indexDataverse(dv);
+                } catch (IOException | SolrServerException e) {
```
landreev, line 236: Maybe a stupid question - but is it wrong to just "catch (Exception e)" in a situation like this - where we don't care why it failed/what kind of an exception it was? Does catching any exception make it more reliable potentially... or are there reasons why it should be avoided?

landreev (commented):
Looks good overall.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
sekmiller: In general, but maybe not in this case, if we get an exception that is unexpected we want it to fail loudly so that we can address it in another way if necessary.

landreev: OK, makes sense.

landreev (approved):
Looks good.

PR #7345, opened 2020-10-21
7307 get user by token

### LINKED ISSUE #7307: API/OpenDP: Retrieve User Information
## API: Retrieve User Information
- Via a user’s apiToken, retrieve the user’s data
- Output may be the same as the existing “List Single User” API endpoint in the Dataverse Native API:
  - http://guides.dataverse.org/en/latest/api/native-api.html#id146
   ```json
  {
    "authenticationProviderId": "builtin",
    "persistentUserId": "dataverseAdmin",
    "position": "Admin",
    "id": 1,
    "identifier": "@dataverseAdmin",
    "displayName": "Dataverse Admin",
    "firstName": "Dataverse",
    "lastName": "Admin",
    "email": "dataverse@mailinator.com",
    "superuser": true,
    "affiliation": "Dataverse.org"
  }
  ```
- Or a subset of these fields: `persistentUserId`, `firstName`, `lastName`, `email`

Note: I am separating this out from #7158:
 - This is [Part 3 of #7158](https://github.com/IQSS/dataverse/issues/7158#issuecomment-673489082)
 - Required for the OpenDP MVP

sekmiller:
Adds an endpoint to the admin API to list a single user by passing that user's token. Required for the OpenDP MVP

Closes #7307 

**Additional documentation**: In addition to adding the doc to the List Single User section, I also made a slight edit to the adjacent "Create Authenticated User". It was missing a header and had a typo.


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2875,8 +2875,15 @@ Sample output using "dataverseAdmin" as the ``identifier``::
       "superuser": true,
       "affiliation": "Dataverse.org"
     }
+    
+Alternatively you may pass a user's api ``token`` to retrieve the same output::
+   
+    GET http://$SERVER/api/admin/authenticatedUsers/token/$token
```
scolapasta, line 2881: Unless I'm missing something, he admin api is blocked off (rightly so), so external tools would not be able to call this. Do we have other user APIs, if so this should go there. (if not, we should create something like "/api/users/"*)

* I've never been a fan of the plural, but we should be consistent with "datasets", "dataverses", etc.

Also, we should be getting the token from the header, like we do from other APIs.

scolapasta (changes requested):

doc/sphinx-guides/source/api/native-api.rst
sekmiller: My bad, I assumed that since it was getting the same info it should have the same permissions required. I will change/fix.

qqmyers: I think as it is, the only token you can get info about is the one you already have, and it really functions like a /api/users/me call. Does anyone need to use one apitoken to check other ones? If not, and if this call should allow a header, then perhaps it shouldn't also have a token path parameter (just the header or &key= query param) to avoid scanners.

scolapasta: Yes, I left that part out, but by getting the token from the header, it doesn't need to have a token path parameter.

pdurbin: For documentation purposes, we already have a "Get All Notifications by User" endpoint and it probably makes sense to put the new docs about getting your user info near these older docs, or vice versa: https://guides.dataverse.org/en/5.1.1/api/native-api.html#get-all-notifications-by-user

sekmiller: As discussed, I moved the endpoint to the api/users path. I also moved the doc as suggested by Phil. that whole native api doc needs some reorganization work especially with respect to user related end points.

pdurbin:
I just noticed these docs as well: https://guides.dataverse.org/en/5.1.1/api/native-api.html#users-token-management

They're for actions also under "api/users" (for managing API tokens). @sekmiller should we put move these under the new "User Information" section that's being added in your pull request? Or do some doc clean up in some future pull request?

@scolapasta and @qqmyers have all your questions been addressed? I just looked at the latest code and it seems to be in pretty good shape.

pdurbin (approved):
The code seems fine and @sekmiller and I decided against further consolidating the /api/users docs at this time. Approved.

Issue #6937, opened 2020-05-26
Add byte offset functionality to StorageIO and Access API


landreev:
We want to be able to support the standard HTTP "bytes=" parameter. For example, call `/api/access/datafile/<id>?bytes=8192-8208` and get just the 16 bytes at the 8K offset, instead of the entire file. 

(it's "working" for some people in our own prod. Dataverse - see #5957 - but only because we redirect to S3, which does understand the parameter). 

To implement it, we should add some form of "open at offset specified" functionality to the StorageIO structure. Then adding the handling of the extra parameter to the Access API should be trivial. 

Should be straightforward overall.

djbrooke:
- Small modification of AccessIO
- per @djbrooke, This is more defensive than anything - we wouldn't want an installation to have files downloaded many times when this feature could be used (saving on costs and allowing for better stability)

djbrooke:
- from @qqmyers - this will allow us to do better file ID for large files
- this will continue to support the Tectonic use case on Harvard Dataverse - and we're concerned about that same use case happening for another installation, where there'd be cost/stability implications

PR #7350, opened 2020-10-22
7275 upload auxiliary files

### LINKED ISSUE #7275: Add API to (relatively) arbitrarily upload auxiliary files
The use case, as relevant to the OpenDP implementation: Summary stats with added differential privacy "noise" will be generated by the OpenDP software for a specific Datafile, outside of Dataverse; then deposited into Dataverse, for later retrieval. This issue is for implementing the APIs, for the deposit and later retrieval of these fragments.  (There will be multiple physical files in different formats - xml, json; and possibly multiple versions of such diff private metadata sets for a single datafile. On the implementation layer all these individual fragments will be saved as "auxiliary files" using the standard Dataverse StorageIO system). 

More generally, this is an API for accepting, storing and serving some metadata that Dataverse cannot produce itself (unlike, for example, image thumbnails or DDI XML describing the data variables that Dataverse knows how to generate from Datafiles). So it must be deposited by an external application before Dataverse can serve it. 

There is no *direct* UI impact. The deposit happens automatically, the remote OpenDP application performs it without the human user being directly involved. (The user does not need to know anything about this API). But once the diff. private metadata has been deposited for a Datafile, Dataverse will be showing extra options (on the dataset and/or file pages) - most likely an extra explore option. In the old PSI demo implementation the user would also see an option to download the deposited diff. private metadata (in Json format) in the normal download pulldown menu. (I'm assuming this issue is just for the API. And any such UI logic will be handled separately). Dataverse does NOT do anything with the actual diff. private metadata deposited (at least as currently defined)! It only knows how to store it for later retrieval and how to serve it on demand; and knows that certain extra things can be done by outside applications for datafiles that have diff. private metadata saved. (for example, Dataverse may know to add an extra redirect link to a diff. private data explore viewer for such a datafile; but all the "exploring" etc. will happen outside the Dataverse application). 

(end use case description -- L.A.)

In planning how to handle the auxiliary files for upload and considering other possible future tools (e.g. a mapping tool or a time series graphing tool) that would require auxiliary metadata, it seems clear that designing a flexible system to upload auxiliary files would be useful.

Note we currently already have two types of auxiliary files: those that can already be recreated solely by dataverse and those that cannot. An example fo the first would be thumbnails for images. An example of the second is the original file on a tabular download.

This issue concerns only the latter, as the files that config tools would deposit are not recreate-able (without the use of the tool).

ekraffmiller:
What this PR does / why we need it: This is a generic API for uploading and downloading Auxiliary files. Needed for OpenDP use case

Which issue(s) this PR closes: 7275

Closes #7275 

Suggestions on how to test this: Integration test in AccessIT

Does this PR introduce a user interface change? If mockups are available, please link/include them here: no UI change

Is there a release notes update needed for this change?:

Additional documentation: documentation added to Dataverse Developer doc


src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -227,6 +227,18 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
...
+                    String auxTag = di.getAuxiliaryFile().getFormatTag(); 
+                    String auxVersion = di.getAuxiliaryFile().getFormatVersion();
+                    if (auxVersion != null) {
+                        auxTag = auxTag + "_" + auxVersion;
+                    }
+                    long auxFileSize = di.getAuxiliaryFile().getFileSize();
+                    InputStreamIO auxStreamIO = new InputStreamIO(storageIO.getAuxFileAsInputStream(auxTag), auxFileSize);
+                    auxStreamIO.setFileName(storageIO.getFileName() + "." + auxTag);
+                    auxStreamIO.setMimeType(di.getAuxiliaryFile().getContentType());
+                    storageIO = auxStreamIO;
```
qqmyers, line 240: storageIO.open called above creates the inputStream for the main file. If it's not going to be used, it needs to be closed (storageIO.getInpputStream().close() - StorageIO doesn't implement Closeable yet/have a close() method - which could be useful someday).

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -0,0 +1,114 @@
...
+            // If that is successful (does not throw exception),
+            // then save to db.
+            // If the db fails for any reason, then rollback
+            // by removing the auxfile from storage.
+            storageIO = dataFile.getStorageIO();
+            AuxiliaryFile auxFile = new AuxiliaryFile();
+            storageIO.saveInputStreamAsAux(fileInputStream, auxExtension);
+            auxFile.setChecksum(FileUtil.calculateChecksum(storageIO.getAuxFileAsInputStream(auxExtension), systemConfig.getFileFixityChecksumAlgorithm()));
+
+            Tika tika = new Tika();
```
qqmyers, line 73: So far Tika is only used for full-text indexing and content-type detection is done differently for Datafiles. Could/should these both be done the same way? (not necessarily in this PR)

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -0,0 +1,114 @@
...
+        try {
+            // Save to storage first.
+            // If that is successful (does not throw exception),
+            // then save to db.
+            // If the db fails for any reason, then rollback
+            // by removing the auxfile from storage.
+            storageIO = dataFile.getStorageIO();
+            AuxiliaryFile auxFile = new AuxiliaryFile();
+            storageIO.saveInputStreamAsAux(fileInputStream, auxExtension);
+            auxFile.setChecksum(FileUtil.calculateChecksum(storageIO.getAuxFileAsInputStream(auxExtension), systemConfig.getFileFixityChecksumAlgorithm()));
```
qqmyers, line 71: It's not too hard to wrap the fileInputStream and calculate the has during the original write (versus having to retrieve the file and scan it a second time. (The util.Bag code has examples.)

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1084,6 +1113,65 @@ private String getWebappImageResource(String imageName) {
...
+     * @return 
+     */
+    @Path("datafile/{fileId}/metadata/{formatTag}/{formatVersion}")
+    @POST
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+
+    public Response saveAuxiliaryFileWithVersion(@PathParam("fileId") Long fileId,
+            @PathParam("formatTag") String formatTag,
+            @PathParam("formatVersion") String formatVersion,
+            @FormDataParam("origin") String origin,
```
qqmyers, line 1135: FWIW: the normal file add uses a single jsonData param rather than multiple FormDataParams. Not sure if that's intended to be consistent across the API.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
landreev: Thanks, I missed that. Fixed. 
It's not pretty, what goes on in that DownloadInstance/DownloadInstanceWriter setup, btw. I am opening an issue for cleaning up/simplifying/rewriting it.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
ekraffmiller: Can you explain more what you had it mind? I looked at BagGenerator and BagValidation map, and can't find an example of this.  I could see how you could combine the logic that reads the stream to calculate the checksum, with the read from stream that happens in StorageIO, but I don't see an example of where that's happening.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
ekraffmiller: for Datafiles, JHove is used, which works on a file as opposed to an inputStream.  So to use JHove, I think I would have to save the aux file to a temp location, unless I'm missing something. Is this something we want to do?

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
qqmyers: Sorry - it would help if I pointed to the right place. It's actually the archiving commands that do this. E.g. [the Google Cloud one](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GoogleCloudSubmitToArchiveCommand.java#L78 ). They complicate things with threads, but the basic idea is that you wrap your stream with a java.security.DigestInputStream and, once you've finished reading the bytes, you can get the digest for the algorithm you set from that stream.

ekraffmiller: ok, that makes sense, thank you

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
qqmyers: FWIW: Requiring a temp file for mime detection is a problem for Datafiles too, so it may make sense to use Tika instead of JHOVE there as well. (#6937 would make it possible to get ranges of bytes from S3 to help avoid having to get a whole file).

ekraffmiller: ok, should we create a separate ticket for changing mime detection to Tika for Datafiles? (Is this something that is usually talked about at dv_tech hour?)

src/main/java/edu/harvard/iq/dataverse/api/Access.java
ekraffmiller: I agree with the desire for consistency, but the parameters listed in Auxiliary files are required, whereas for depositing a Datafile, the jsonData parameter contains only optional params.  So I think it's probably better to keep the Auxiliary API as is so that the required parameters are more obvious.

pdurbin (changes requested):
Overall, this looks good. Below I'll make some high-level comments and summarize feedback I left on specific lines.

The pull request template wasn't used so I'm not sure if this pull request is meant to close #7275 or not. This should be clarified.

The branch is quite behind develop so the latest need to be merged in.

There are no docs. An update to the API Guide should be made with full curl examples of how to use these new endpoints. Also, a release note fragment should be added to describe the feature.

~There seems to be a security/privacy issue with how isPublic is implemented: You can download an aux file from a dataset draft. I don't know if this is by design or not but this will probably be unexpected behavior.~ Update: I re-tested this and got `..."code": 403, "message": "Not authorized to access this object via this API endpoint...`

It's clear that this pull request is a small chunk along the way to integration between OpenDP and Dataverse, which is fine. I see that there's an issue at #7400 to worry more about the UI implications of these new aux files.

Since the focus in this pull request is on APIs I'm wondering if we should do a little more. Specifically, here are some use cases I can anticipate people wanting:

- I want to be able to delete an aux file.
- I want to be able to list all the aux files for a data file.
- For one of my aux files, I want to be able to check its content type and other information.

@qqmyers noticed that filetype detection is done through a different method (Tika) that than our normal methods. I'm ok with QA testing it as-is but we should probably make these consistent at some point.

I'm still a little troubled that we have API endpoints for files under both "access" and "files". We need to keep all endpoints working, of course, but eventually we should come up with guidance on what should go where. I prefer "files" but I understand that "access" was there first.

Again, overall, looks good.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
ekraffmiller: Discussed this during tech hours, we agreed that it would be good to eventually switch from JHove to Tika in other parts of the code for consistency

src/main/java/edu/harvard/iq/dataverse/DataFile.java
```diff
@@ -281,6 +284,14 @@ public String getDuplicateFilename() {
...
         this.duplicateFilename = duplicateFilename;
     }
+
+    public List<AuxiliaryFile> getAuxiliaryFiles() {
+        return auxiliaryFiles;
+    }
+
+    public void setAuxiliaryFiles(List<AuxiliaryFile> auxiliaryFiles) {
+        this.auxiliaryFiles = auxiliaryFiles;
+    }
```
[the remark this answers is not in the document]
ekraffmiller: We decided to keep the focus on this ticket to deposit and download APIs, other API calls can be added in the future.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
landreev: I'm about to approve the PR. There's just one thing I regret not having thought about earlier, when we were having this discussion, above: WHY are we trying to detect mime types at all? As opposed to just making the uploading client supply it, as another parameter? 
We try to detect mime types on "normal" uploaded files - but that's because users upload  arbitrary files... This API is for something that's structured; at least in the immediate use case scenario. I.e. that preprocessed summary stats fragment will always be in JSON, the diff. private DDI will always be XML etc. etc. We don't expect any variety there... 
I don't see this as  a problem, having this detection code in place. It may come in handy for other cases. And, even if we decide that these aux uploads should, or can supply the mime type as a parameter - we could use this detection as an extra validation step. 
Still, I can't help but feel I should've said "let's not even worry about it, let's just add a parameter instead", early on.

landreev (approved):

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
landreev: I'm about to approve the PR. There's just one thing I regret not having thought about earlier, when we were having this discussion, above: WHY are we trying to detect mime types at all? As opposed to just making the uploading client supply it, as another parameter?
We try to detect mime types on "normal" uploaded files - but that's because users upload arbitrary files... This API is for something that's structured; at least in the immediate use case scenario. I.e. that preprocessed summary stats fragment will always be in JSON, the diff. private DDI will always be XML etc. etc. We don't expect any variety there...
I don't see this as a problem, having this detection code in place. It may come in handy for other cases. And, even if we decide that these aux uploads should, or can supply the mime type as a parameter - we could use this detection as an extra validation step.
Still, I can't help but feel I should've said "let's not even worry about it, let's just add a parameter instead", early on.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1084,6 +1113,64 @@ private String getWebappImageResource(String imageName) {
...
+
+        if (!dataFile.isTabularData()) {
+            return error(BAD_REQUEST, "Not a tabular DataFile (db id=" + fileId + ")");
+        }
+         
+
+        AuxiliaryFile saved = auxiliaryFileService.processAuxiliaryFile(fileInputStream, dataFile, formatTag, formatVersion, origin, isPublic);
+      
+        if (saved!=null) {
+            return created(saved.getId().toString(),JsonPrinter.json(saved));
```
pdurbin, line 1164: This use of `created` with an id seems to be out of step with all other uses. Notice below that we always return the path to the object just created so it can be further manipulated:

Screen Shot 2020-11-24 at 11 50 22 AM

Since this API is experimental, we could just switch it from `created` to `ok` (both "add file" and "add prov" use `ok` and this code itself used `ok` until I started asking about returning an id in a JSON response which lead to `created` being added in c1e03a0) or we could put a proper path in, like `created` expects. From the test in AccessIT, the file can be downloaded from a path like this:

    .get("/api/access/datafile/" + tabFile1Id + "/metadata/dpJSON/v1")

pdurbin (commented):
I made a minor comment about `created` vs `ok`.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
pdurbin: @landreev and I talked this out in Slack and decided to go with switching from created to ok in 471dd25 (which is how this pull request started.

pdurbin (approved):
@landreev and I decided to make the "add" response 200/OK. I also merged the latest from develop.

kcondon:
@ekraffmiller Getting 404 endpoint does not exist error. I'm sure it is my fault but could use some help.

Issue #7256, opened 2020-09-09
scripts for rebuilding a dev environment (with sample data)


pdurbin:
Yesterday in tech hours we talked about things to help developers and I brought up what I think of as the "rebuild with sample data" scripts I've been using for years and years. (I stopped using them only recently because we switched to Payara.)

There are lots of reasons developers might want to rebuild from time to time:

- You've been running integration tests and all your "nice" datasets are buried beneath them.
- You're ready for a clean start with known (or no) data.
- You're having trouble with database upgrade scripts.

By rebuild I don't mean that every single dependency is removed. In fact, the application server (Glassfish or Payara) is largely untouched in my scripts. That said, a number of major changes are made:

- The database is dropped.
- The data files are deleted.
- Solr is cleared out.

These scripts eventually evolved into the ones we used on the phoenix server for many years. This server was rebuilt as described above on every run. The scripts can be found at https://github.com/IQSS/dataverse/tree/develop/scripts/deploy/phoenix.dataverse.org

After "rebuild" has run, the "post" scripts gets executed, and starts with some setup...

- Run setup-all.sh.
- Run SQL scripts (reference data and create sequence).
- Set DOI provider to FAKE.

... and then continues on to load some sample data. As I mentioned on the call, these scripts create some "birds and trees" users, dataverses, and datasets. (Even though our sample data repo is newer, it doesn't create users.) The Spruce Goose dataset (screenshot below) might be familiar.

When estimating this issue, here are some questions to consider:

- Do we want the "birds and trees" data? Or would we rather have the [sample data][]? Or both?
- Should we defer worrying about sample data until a later issue?
- Should we consider adding a rebuild or reinstallation of Payara as part of this? Or should we stick to the model above?
- Should we use the bash scripts above as a starting point? Or should this be a (dangerous!) feature of the installer?

[sample data]: https://github.com/IQSS/dataverse-sample-data

<img width="1184" alt="Screen Shot 2020-09-09 at 2 20 18 PM" src="https://user-images.githubusercontent.com/21006/92640953-0e833d00-f2ac-11ea-9448-0745640e87c7.png">

poikilotherm:
Some ideas:
1. Move all SQL stuff to a Flyway baseline and/or migration. #5871
2. Make DOI providers configurable via MicroProfile Config API (just like everything else) #7000
3. Move all from setup.sh to bootstrapping code #5361 while keeping it configurable with #7000
[...]

Issue #5361, opened 2018-12-03
Bootstrapping config on first deployment


poikilotherm:
This is a story for #5292 and it depends on introduction of Flyway in #5344 / #5349.

----

While installing Dataverse, you need to create a few database entries regarding admin user, root dataverse, roles, metadata blocks and more. Details can be found in `setup-all.sh` and `reference_data.sql`.

This is currently done by the install script. This story is about a proposal to change this. Positive sideffect: this could be a first step towards a [web based installer as @pdurbin would love to see](http://irclog.iq.harvard.edu/dataverse/2018-11-30#i_80571).

### Load initial data after migrations happened
As soon as the database has been created by JPA and Flyway, we can hook a bootstrap loader into the `BaseFlywayCallback.afterMigration()` routine. [Example](https://thoughts-on-java.org/java-based-database-migrations-callbacks-flyway)

After checking that the database has no root dataverse present, it could load the data currently loaded via REST API with internal calls to the appropriate services, based on the same data structures (mostly JSON files).

A switch offers skipping bootstraping from code. It has to be enabled to let bootstrapping happen for maximum backward compatibility.

### Benefits from this and things to keep in mind
* Development and Test installations, either Docker or Appserver based might not need to use an installer as long as default values are sufficient.
* Adding a switch via a default `false` system property allows skipping, so the installer as it is now can stay as is.
* Current setup of metadatablocks via loading from JSON files can be reused.
* Setting things via the API is more future proof than a database dump, as the API is more flexible. Migrations for a fixture seem ineffective.

As always: I would really appreciate comments and input.

poikilotherm:
Some more explaining context why this came up in the first place:

While hacking on #5292, I missed an easy option to run the `setup-all.sh` script after the WAR has been deployed. That is not easy to achieve without application support and ideas like an init container or other stuff seemed a rather ugly workaround.

Bootstrapping from within the application was the natural thing to come up with next. This can of course be done independently from Flyway (simple @StartUp EJB), but it seemed more natural to make this happen AFTER all migrations passed.

pdurbin:
Closing because we now have configbaker.

PR #7355, opened 2020-10-23
7256 purge referencedata

poikilotherm:
This is a first step to ease deployments by making parts of the installer unnecessary.

One step during install is loading the SQL script `reference_data.sql`. I'd replace it by Flyway migrations & callback actions.

That way, the database related script is streamlined with the other database operations plus bootstrapping gets easier (no more installing Python Postgres nor `psql` in containers necessary).

**Which issue(s) this PR ~~closes~~ relates to**:

Related to #7256 (closed, but that set the trigger)
Related to #5361 (this PR is only a partial solution, so don't close)

Nada.

Deploy a fresh installation and check if indexes and initial data are still present.

Nope.

I don't think so.

**Additional documentation**:
Nope.


poikilotherm:
Hey @landreev @sekmiller @pdurbin I'd love some chatter on this, as you guys crafted the original script.

poikilotherm:
I think this is ready for review. It's narrow scoped. More stuff about bootstrapping to come later.

scolapasta (commented):
could you explain the names of the two new files?

landreev:
I noticed that you are adding a couple of comments to the V1__flyway_schema_baseline.sql. I'm just curious, why is it NOT causing a problem? - Is that because Flyway purposefully does not save the checksum for the baseline file?

poikilotherm:
@scolapasta 
Actually it's only one new with a to date unknown filename schema: `afterMigrate__1-upsert-referenceData.sql`. 

This is simply a script executed every time after a successfull migration happened (so this will be executed everytime a new DB migration happens). Please see [flyway callback docs](https://flywaydb.org/documentation/concepts/callbacks#sql-callbacks) for more details.

It will not cause trouble because precautions are in place to only insert data if the DB is empty (during bootstrap). In the future, more scripts can be added with incremented numbers.

The name of `V5.1.1.3__7256-purge-referencedata.sql` should be updated to reflect the recent 5.2 release, the files were created before that.

@landreev 
No checksum magic here. The file content is simply ignored AFAIK when the migration is already applied. At least it has been ignored when I tried to insert indexes with it... (see fc3dbb6)

scolapasta:
@poikilotherm ah gotcha. the 5.1.1.3 confused me. Can you update? And I had not head the term "upsert" before, so didn't know what that meant (now I do :)

poikilotherm:
Sure!
So you folks think this is the way to go? Should we talk about it some more? Or just update and push to QA then?

scolapasta:
Seems reasonable enough to me and I know @landreev took a look, as well. @poikilotherm if you feel it needs any more discussion, we can include as part of the Tuesday meeting.

poikilotherm:
@scolapasta I updated the files.

I'll take the liberty to move to QA, as y'all were happy with code review. Please revert if any concerns arise with that.

poikilotherm:
Thanks for merging @kcondon! 🥳

PR #7363, opened 2020-10-27
add script for rebuilding dev environment #7256

### LINKED ISSUE #7256: scripts for rebuilding a dev environment (with sample data)

pdurbin:

Sometimes developers benefit from dropping their database. This script does that and the subsequent setup so that API test can run.

Closes #7256

As mentioned in the issue and tech hours today, I've been using a variation of this script for years. That script was also in use on the phoenix server for years.

- Set up a dev environment, create a dataset or two.
- Run the script.
- The dev environment should be clear and empty (no datasets, etc.).

No.

Probably not worth mentioning.

**Additional documentation**:

Just the included dev guide update.

### DIFF
--- a/scripts/dev/dev-rebuild.sh
+++ b/scripts/dev/dev-rebuild.sh
@@ -22,7 +22,7 @@ echo "Deleting ALL DATA FILES uploaded to Dataverse..."
 # TODO: Make this configurable.
 rm -rf $PAYARA_DIR/glassfish/domains/domain1/files
 
-echo "Terminating database settings so we can drop the database..."
+echo "Terminating database sessions so we can drop the database..."
 psql -U postgres -c "
 SELECT pg_terminate_backend(pg_stat_activity.pid)
 FROM pg_stat_activity
--- a/scripts/dev/dev-rebuild.sh
+++ b/scripts/dev/dev-rebuild.sh
@@ -19,6 +19,7 @@ echo "Deleting \"generated\" directory..."
 rm -rf $PAYARA_DIR/glassfish/domains/domain1/generated
 
 echo "Deleting ALL DATA FILES uploaded to Dataverse..." 
+# TODO: Make this configurable.
 rm -rf $PAYARA_DIR/glassfish/domains/domain1/files
 
 echo "Terminating database settings so we can drop the database..."
--- a/scripts/dev/dev-rebuild.sh
+++ b/scripts/dev/dev-rebuild.sh
@@ -1,0 +1,71 @@
+#!/bin/sh
+PAYARA_DIR=/usr/local/payara5
+ASADMIN=$PAYARA_DIR/glassfish/bin/asadmin
+DB_NAME=dvndb
+DB_USER=dvnapp
+
+echo "Checking if there is a war file to undeploy..."
+LIST_APP=$($ASADMIN list-applications -t)
+OLD_WAR=$(echo $LIST_APP | awk '{print $1}')
+NEW_WAR=target/dataverse*.war
+if [ ! -z $OLD_WAR ]; then
+  $ASADMIN undeploy $OLD_WAR
+fi
+
+echo "Stopping app server..."
+$ASADMIN stop-domain
+
+echo "Deleting \"generated\" directory..."
+rm -rf $PAYARA_DIR/glassfish/domains/domain1/generated
+
+echo "Deleting ALL DATA FILES uploaded to Dataverse..." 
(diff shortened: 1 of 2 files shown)


scripts/dev/dev-rebuild.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+echo "Stopping app server..."
+$ASADMIN stop-domain
+
+echo "Deleting \"generated\" directory..."
+rm -rf $PAYARA_DIR/glassfish/domains/domain1/generated
+
+echo "Deleting ALL DATA FILES uploaded to Dataverse..." 
+rm -rf $PAYARA_DIR/glassfish/domains/domain1/files
```
scolapasta, line 22: since you can change this location, can this be determined from the jvm option instead of hardcoded?

scripts/dev/dev-rebuild.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo "Terminating database settings so we can drop the database..."
```
scolapasta, line 24: just because I'm curious, what does this do?

scripts/dev/dev-rebuild.sh
pdurbin: I had that thought, that perhaps some devs keep their data files somewhere other than default (domain1/files). I figured I'd wait until there's demand to make this configurable. We could either add a variable to the top called something like DATA_FILES_DIR or yes, perhaps pull it out from `asadmin list-jvm-options`.

scolapasta: OK, I'm fine with pushing that to later. Could you a) add that as a todo comment and b) add a note in the docs to point out that files would have to be in the default location?

scripts/dev/dev-rebuild.sh
pdurbin: I'm assuming you mean the pg_terminate_backend stuff. It's just a SQL statement I found online somewhere. I didn't dig into it. In practice, it's sometimes necessary to terminate all sessions because you've got a session or two left over from pgAdmin or a terminal. In my experience, you can drop a database with open sessions.

scolapasta: Ah, makes sense. I was confused by "Terminating database settings". I think that's a typo and should be "sessions".

scripts/dev/dev-rebuild.sh
pdurbin: Sure, fixed in ecde956db.

scripts/dev/dev-rebuild.sh
pdurbin: Oh, a typo. I meant sessions, not settings. Fixed in d3778f360.

scolapasta (approved):

PR #7378, opened 2020-11-02
Handle non-ascii chars in search

### LINKED ISSUE #820: Search: Accented characters, ampersands, negative numbers and other special characters
This was suggested by Eleni during discussions around search issues with accented characters. 

The suggestion is to allow searching both using the original accented characters and without, in those cases where a user may not have access to accented keyboard, etc.

qqmyers:
**What this PR does / why we need it**: adds the ASCIIFoldingFilterFactory which should allow for matches when accented characters are used - in the text (the index side) or when a search term is added (the filter itself means an unaccented char in a search query works while the preserveOriginal == true means that search for the accented char also works).

Closes #820

**Special notes for your reviewer**: This doesn't handle all the cases in the title of #820, but there's a comment suggesting that things like & and negative numbers are/were addressed in other issues. So - guessing this closes #820 but not sure.
Also note the concern in the issue that this approach can cause false positives: e.g. "cañon" will be found searching for "canon" which isn't right if you intended to search only for the English word canon.

There are 2 new search cases that should work with this change:
an accented char in some text field should be matched by an ascii search .e.g. "cañon" will be found searching for "canon"
an accented search will match an ascii field.,e.g.  "canon" will be found searching for "cañon"

Should work for words with all such chars, eg. the list in the issue: (á, à, â, ç, é, è, ê, ë, í, ó, ö, ú, ù, û, ü…).

**Additional documentation**:


pdurbin (changes requested):
This is just a config file so it's fine to go to QA but a release note should be added to explain what to do, how to put the file into place, if the Solr index needs to be cleared or not, etc.

qqmyers:
To start by answering the questions: this should require both the standard 'copy schema.xml' to the solr config dir (same as with any release dealing with solr). The other solr files don't need to be updated. To get the fix, one then has to run a solr re-index - might recommend the incremental one to avoid down-time.

mheppler:
@qqmyers @pdurbin @scolapasta ...

Do the changes to Solr in _7373 solr upgrade #7645_ make the debate about the release notes and schema and re-index here moot?

qqmyers:
Good call - since copying the schema.xml from the Dataverse sw repo and reindexing are part of a fresh install, the instructions for that PR will cover what's needed here. (We could warn not to copy the local ones from solr 7.x (since there's a change in schema.xml due to this PR), but there are lots of things one shouldn't do...)

conf/solr/7.7.2/schema.xml
```diff
@@ -38,7 +38,7 @@
     catchall "text" field, and use that for searching.
 -->
 
-<schema name="default-config" version="1.6">
+<schema name="default-config" version="1.7">
```
mheppler, line 41: When PR #7645 gets merged (in QA), this PR will need to be updated, as this edit is made to the soon-to-be older 7.7.2 schema.

doc/release-notes/820-non-ascii-chars-in-search.md
```diff
@@ -0,0 +1,8 @@
+### Search with non-ascii characters
+
+Many languages include characters that have close analogs in ascii, e.g. (á, à, â, ç, é, è, ê, ë, í, ó, ö, ú, ù, û, ü…). This release changes the default Solr configuration to allow search to match words based on these associations, e.g. a search for Mercè would match the word Merce in a Dataset, and vice versa. This should generally be helpful, but can result in false positives.,e.g. "canon" will be found searching for "cañon". 
+
+## Upgrade Instructions
+
+1. You will need to replace or modify your `schema.xml` and restart solr. Re-indexing is required to get full-functionality from this change - the standard instructions for an incremental reindex could be added here.
```
mheppler, line 7: When PR #7645 gets merged (in QA), this PR will need to be updated, as these release notes should be merged with the release notes included in that Solr update PR.

mheppler: Here are the release notes from the Solr upgrade PR.

> ### Solr Update
> 
> With this release we upgrade to the latest available stable release in the Solr 8.x branch. We recommend a fresh installation of Solr 8.8.1 (the index will be empty)
> followed by an "index all".
> 
> Before you start the "index all", Dataverse will appear to be empty because
> the search results come from Solr. As indexing progresses, partial results will 
> appear until indexing is complete.
> 
> 
> See http://guides.dataverse.org/installation/prerequisites.html#installing-solr
> 
> [for the additional upgrade steps section]
> 
> Run the script updateSchemaMDB.sh to generate updated solr schema files and preserve any other custom fields in your Solr configuration.
> For example: (modify the path names as needed)
> cd /usr/local/solr-8.8.1/server/solr/collection1/conf
> wget https://github.com/IQSS/dataverse/releases/download/v5.4/updateSchemaMDB.sh
> chmod +x updateSchemaMDB.sh
> ./updateSchemaMDB.sh -t .
> 
> See http://guides.dataverse.org/en/5.4/admin/metadatacustomization.html?highlight=updateschemamdb for more information.

poikilotherm:
Please be aware of #7662 for this (maybe we can address this in the same go here?)

doc/release-notes/820-non-ascii-chars-in-search.md
mheppler: @djbrooke in tech hours, @scolapasta suggested moving this PR forward with the note to REMOVE this "Upload Instructions" section, as it a duplicate of the steps already included in the Solr upgrade PR #7645.

scolapasta (approved):

doc/release-notes/820-non-ascii-chars-in-search.md
djbrooke: @mheppler great, thanks, I'll leave a note in this PR at the top of the notes so I don't forget, in case something slips and we don't get this into the release we expect to.

PR #7384, opened 2020-11-03
Iqss/7383 file pid reservation api call

### LINKED ISSUE #7383: With file PIDs able to be pre-registered, an api to do that for existing files would be useful
When pre-registration of DataCite  DOIs for Datasets was implemented, a PID reservation API call was added. Now that file PIDs can also be pre-registered, it would be useful to have a way to update existing files.

qqmyers:
**What this PR does / why we need it**: extends the reserve pid api for datasets to also reserve any unreserved files at the same time. Also does some cleanup and improvement of error messages.

Closes #7383 

**Special notes for your reviewer**: This PR extends #7334/ includes the changes from that PR.

**Suggestions on how to test this**: enable file PIDs and test being able to reserve file PIDs via the API(either use existing files, or misconfigure DataCite so that PID reservation fails during file creation).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: just updates to error messages

**Is there a release notes update needed for this change?**: could restore the recommendation to pre-reserve PIDs

**Additional documentation**:


djbrooke:
Since this depends on #7334, I'm going to put this on ice until we dig into #7393.

qqmyers:
Changes here are now included in #7334  - so closing this PR.

PR #7385, opened 2020-11-03
file tags searchable from advanced search and dataset page #7194

### LINKED ISSUE #7194: Add File tags to advanced search, update dataset page to allow file tag searching (with proper syntax)
Currently you can type fileTags:XXXXX and dataverse will search the index for XXXXX in the fileTags. But we should add this as a field in Advanced Search, so the user does not need to know this syntax.

(additionally we need to make sure it works this way from the dataset page, as in some quick tests fileTags:XXXXX did not return results as expected)

pdurbin:

File tags are absent from the Advanced Search page. This pull request adds it.

When you type file tags into the search box for the file listing on the dataset page, you don't get hits on the files with those tags. (Only file name and file description are searched.) This pull request fixes this.

None. (This PR adds part of what is #7194; the 2nd part has expanded into a broader conversation, so we'll leave the issue open for that)

I followed the same pattern as other search fields. I didn't re-write the search code.

- Add some file tags to files.
- Search for them from the Advanced Search page.
- Search for them from the file listing on the dataset page.

Yes.

Screenshot of "File Tags" from the Advanced Search page:

Screen Shot 2020-11-03 at 3 10 41 PM

Screenshot of searching for a tag called "Foo Bar" on the file listing on the dataset page:

Screen Shot 2020-11-03 at 3 47 18 PM

Yes, included.

**Additional documentation**:

None.


scolapasta (approved):
Looks good; I edited the PR conversation to reflect that this does not close #7194, as that has spawned into a new discussion. But this clearly adds value for the first part, so moving it to QA.

kcondon:
@pdurbin Would you refresh this from develop?

pdurbin:
@kcondon sure, done in 86568b1.

PR #6023, opened 2019-07-15
Added CDATA syntax to javascript in analytics code sample file in Ins…

mheppler:
## Related Issues

- closes #5967 Broken 403 Error pg

## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [x] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [x] [Documentation][docs] completed
- [x] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

Issue #5967, opened 2019-06-24
Broken 403 Error pg


mheppler:
I have been seeing this horrible yellow error pg in my localhost (Firefox, develop v 4.15) a lot lately, while I believe most of them are of the 500 variety, I am creating this issue specifically for the 403, because that's easier for me to reproduce.

![Screen Shot 2019-06-24 at 4 19 32 PM](https://user-images.githubusercontent.com/687227/60049291-01286580-969c-11e9-978a-c9198a8b3380.png)

Navigate to an unpublished dataset as an admin in one browser, copy and paste the URL into another browser, when prompted log in as a user without access permissions, and boom, ugly error pages.

The Google Chrome version ain't much better.

![Screen Shot 2019-06-24 at 4 21 05 PM](https://user-images.githubusercontent.com/687227/60049344-18675300-969c-11e9-9894-e6f9c0d5a283.png)

The 404 pg works as expected, for what it's worth.

This behavior also appears on demo (v 4.14 build 1326-57a392d). However, dataverse-internal (v. 4.15 build 1359-631db06) is properly displaying the 403 error pg in both Firefox and Chrome

mheppler:
Randomly found this error in my server log when clicking on a dataverse from the featured carousel. Not sure why it popped up. My browser loaded the target dataverse pg with no problems.

```
Warning:   org.apache.catalina.core.StandardHostValve@7ec925bd: Exception processing ErrorPage[errorCode=404, location=/404.xhtml]
java.io.IOException: Broken pipe
	at sun.nio.ch.FileDispatcherImpl.write0(Native Method)
	at sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47)
	at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:93)
	at sun.nio.ch.IOUtil.write(IOUtil.java:51)
	at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:471)
	at org.glassfish.grizzly.nio.transport.TCPNIOUtils.flushByteBuffer(TCPNIOUtils.java:149)
	at org.glassfish.grizzly.nio.transport.TCPNIOUtils.writeCompositeBuffer(TCPNIOUtils.java:87)
	at org.glassfish.grizzly.nio.transport.TCPNIOAsyncQueueWriter.write0(TCPNIOAsyncQueueWriter.java:112)
[...]

PR #7392, opened 2020-11-04
get dialog to close on Save Changes for groups #6792

### LINKED ISSUE #6792: Manage Groups - Dialog popup does not close on save, action saved by system
On the Manage Groups pg, when editing an existing group in the Dataverse Group popup, the Save Changes btn silently saves the action, but does not close the popup. Checking the console, there is this XML parsing error when you click Save Changes.

```
XML Parsing Error: mismatched tag. Expected: </update>.
Location: .../manage-groups.xhtml?dataverseId=####
Line Number 150, Column 3:
```

There was also a console error `ReferenceError: handle_dropdown_popup_scroll is not defined` on the Manage Groups pg when it loads (#5653), but that does not appear to be related.

@kcondon initially discovered this on a test server, but could not reproduce it on another test server. One major difference between these two machines was the analytics button click event tracking code (http://guides.dataverse.org/en/4.20/installation/config.html#tracking-button-clicks) that was set up on the machine throwing the errors.

This bug is not on demo (4.19, no analytics), but it is on production (also 4.19) where the analytics event tracking code was added last month.

pdurbin:

When you click "Save Changes" on the groups page, the dialog doesn't close.

Closes #6792

This reverts pull request #6023 (f40d653 specifically) so I'm concerned #5967 will resurface. Mostly I'm just putting this pull request out there for discussion, so see if others have ideas about how to fix this.

Edit a group and click Save Changes. The dialog should close.

No, apart from a dialog closing properly. It looks like this:

Screen Shot 2020-11-04 at 4 53 56 PM

Yes but I haven't added one yet because I'm not sure this is the fix we want.

**Additional documentation**:

None.


pdurbin:
As an alternative fix, I tried moving the dialog down (diff below) but it didn't help. The only way I've gotten "Save Changes" to work properly is with the (admittedly controversial!) CDATA removal in e0fc1f8 (this pull request).

```
beamish:dataverse pdurbin$ git diff -U5
diff --git a/src/main/webapp/manage-groups.xhtml b/src/main/webapp/manage-groups.xhtml
index 706c8f7ea..91c94d807 100644
--- a/src/main/webapp/manage-groups.xhtml
+++ b/src/main/webapp/manage-groups.xhtml
@@ -94,10 +94,12 @@
                         <div class="button-block">
                             <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.deleteGroup()}" update="@all"/>
                             <p:commandButton styleClass="btn btn-link" value="#{bundle.cancel}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.unrenderDeletePopup()}" update="@all"/>
                         </div>
                     </p:dialog>
+                    <!-- Explicit Group Edit Popup -->
+                    <ui:include src="explicitGroup-new-dialog.xhtml"/>
                     <p:dialog id="viewGroup" styleClass="largePopUp" header="#{bundle['dataverse.manageGroups.tab.action.btn.view.dialog.header']}" widgetVar="viewGroup" modal="true">
                         <p:focus context="viewGroup"/>
                         <div class="form-horizontal">
                             <div class="form-group">
                                 <label for="select_GroupName" class="col-sm-3 control-label">
@@ -176,12 +178,10 @@
                             <button class="btn btn-link" onclick="PF('viewGroup').hide()" type="button">
                                 #{bundle.cancel}
                             </button>
                         </div>
                     </p:dialog>
-                    <!-- Explicit Group Edit Popup -->
-                    <ui:include src="explicitGroup-new-dialog.xhtml"/>
                 </h:form>
             </ui:define>
         </ui:composition>
     </h:body>
 </html>
```

qqmyers:
Given that the problem appears to be linked to having the script in the .html, which is included in the .xhtml and causing issues due to the > and & chars in the scripts (as far as we know), maybe pulling it out to a separate .js file and linking that would resolve this and avoid the earlier issue?

src/main/webapp/manage-groups.xhtml
```diff
@@ -92,8 +92,8 @@
                             <span class="glyphicon glyphicon-warning-sign text-warning"/> <span class="text-warning">#{bundle['dataverse.manageGroups.tab.action.btn.delete.dialog.tip']}</span>
                         </p>
                         <div class="button-block">
-                            <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.deleteGroup()}" update="@all"/>
-                            <p:commandButton styleClass="btn btn-link" value="#{bundle.cancel}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.unrenderDeletePopup()}" update="@all"/>
+                            <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.deleteGroup()}" update="manageGroupsForm"/>
```
scolapasta, line 95: I think this will also need the messagePanel.

scolapasta, line 95: I think this will also need the messagePanel.

src/main/webapp/manage-groups.xhtml
```diff
@@ -171,7 +171,7 @@
                             <p:commandButton id="saveGroup" styleClass="btn btn-default"
                                            actionListener="#{manageGroupsPage.saveExplicitGroup}"
                                            value="#{bundle.saveChanges}"
-                                           update="@all">
```
scolapasta, line 174: I think this will also need the messagePanel.

src/main/webapp/manage-groups.xhtml
```diff
@@ -92,8 +92,8 @@
                             <span class="glyphicon glyphicon-warning-sign text-warning"/> <span class="text-warning">#{bundle['dataverse.manageGroups.tab.action.btn.delete.dialog.tip']}</span>
                         </p>
                         <div class="button-block">
-                            <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.deleteGroup()}" update="@all"/>
-                            <p:commandButton styleClass="btn btn-link" value="#{bundle.cancel}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.unrenderDeletePopup()}" update="@all"/>
+                            <p:commandButton styleClass="btn btn-default" value="#{bundle.continue}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.deleteGroup()}" update="manageGroupsForm"/>
+                            <p:commandButton styleClass="btn btn-link" value="#{bundle.cancel}" onclick="PF('deleteConfirmation').hide()" action="#{manageGroupsPage.unrenderDeletePopup()}" update="manageGroupsForm"/>
```
scolapasta, line 96: this is not directly related but a good time for cleanup, I think - other cancel buttons, just hide popups instead of calling the backing bean. In this case, there seems to be a rendered that is set and unset and feels like it could go away. See for example how the view group dialog on this very same page does not have a rendered attribute.

scolapasta (changes requested):
Looks good, though I think incomplete. (I also added a cleanup request that I think would make sense to consider for this PR. Lastly, do you want to add a release note? I don't it needs one as it's a simple bug fix, but that's up to you.

src/main/webapp/manage-groups.xhtml
pdurbin: Fixed in 79301e6.

pdurbin: Fixed in 79301e6.

src/main/webapp/manage-groups.xhtml
pdurbin: Fixed in 79301e6.

src/main/webapp/manage-groups.xhtml
pdurbin: Fixed in 79301e6.

pdurbin:
Ok, after addressing some concerns in 79301e6, I'm ready for more code review.

scolapasta (approved):
Looks good!

PR #7410, opened 2020-11-13
don't always return OK from get PID API #7187

### LINKED ISSUE #7187: Native API: Get PID info endpoint always returns ok
The get pid info endpoint always returns ok, even when the datacite rest api url is incorrect:

curl -H "X-Dataverse-key:XYZ
" http://localhost:8080/api/pids?persistentId=doi:10.70122/FK2/ABCDEF

{"status":"OK","data":{"response":"crapi.test.datacite.org"}

pdurbin:

The "get PID" API always returns OK (200) even if there's a problem.

Closes #7187

The test case is a hostname that isn't in DNS. I changed the JSON format for the error.

Before:

```
{
  "status": "OK",
  "data": {
    "response": "junk.test.datacite.org"
  }
}
```

After:

```
{
  "status": "ERROR",
  "message": "Problem getting status code from https://junk.test.datacite.org. Is it in DNS? Is doi.dataciterestapiurlstring configured properly?"
}

```

Change doi.dataciterestapiurlstring to a hostname that isn't in DNS, then try the "get PID" endpoint.

No.

No.

**Additional documentation**:

None.


src/main/java/edu/harvard/iq/dataverse/api/Pids.java
```diff
@@ -48,8 +48,12 @@ public Response getPid(@QueryParam("persistentId") String persistentId) {
         String baseUrl = systemConfig.getDataCiteRestApiUrlString();
         String username = System.getProperty("doi.username");
         String password = System.getProperty("doi.password");
-        JsonObjectBuilder result = PidUtil.queryDoi(persistentId, baseUrl, username, password);
-        return ok(result);
+        try {
+            JsonObjectBuilder result = PidUtil.queryDoi(persistentId, baseUrl, username, password);
+            return ok(result);
+        } catch (Exception ex) {
+            return error(Response.Status.BAD_REQUEST, ex.getLocalizedMessage());
```
qqmyers, line 55: Since this is a config problem, BAD_REQUEST is probably not the right response.

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidUtil.java
```diff
@@ -45,7 +46,7 @@ public static JsonObjectBuilder queryDoi(String persistentId, String baseUrl, St
             status = connection.getResponseCode();
         } catch (IOException ex) {
             // Hostname not in DNS, for example.
-            return Json.createObjectBuilder().add("response", ex.getLocalizedMessage());
+            throw new RuntimeException(BundleUtil.getStringFromBundle("pids.datacite.errors.noResponseCode", Arrays.asList(baseUrl)));
```
qqmyers, line 49: This stops the specific error in the issue from resulting in an OK response, but the other exceptions in this class are still being caught and returned as OK. One option for handling these would be to just throw subclasses of WebApplicationException directly from this class (see, for example, the table on [this page](https://dennis-xlc.gitbooks.io/restful-java-with-jax-rs-2-0-2rd-edition/content/en/part1/chapter7/exception_handling.html) for the available exceptions. The edu.harvard.iq.dataverse.api.errorhandlers.WebApplicationExceptionHandler will handle them from there. Each type of error has a default message and any InternalServerErrorExceptions get additional logging.

You can also just let any Throwable be caught by the ThrowableHandler - which are returned as 500 errors.

Neither of these let you pass the exception message on to the user though, so you'd still have to catch exceptions for that in Pids.java if you want to include that in the response (but you could still throw the WebApplicationExceptions to also be able to pass the error type/status code.)

src/main/java/edu/harvard/iq/dataverse/api/Pids.java
pdurbin: @qqmyers yeah, 500 probably makes more sense. Fixed in 7c55e2bd3.

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidUtil.java
pdurbin: @qqmyers good ideas. How does 7c55e2bd3 look?

qqmyers: That works - a god way to be able to send a custom message with a 500. I made another commit to try and catch/assign status codes to other types of errors that can occur. See what you think/feel free to reject/edit further.

qqmyers: https://github.com/IQSS/dataverse/pull/7410/files?w=1 makes it easier to see the changes since indenting changed overall...

pdurbin: @qqmyers as of the commit you added I don't think the case of a PID that can't be found makes sense. Below you can see that when a PID exists, you get a nice "findable" message but when you add some characters to the end of the PID such that it doesn't exist, you get a message saying "API endpoint does not exist on this server" which doesn't make sense:
```
beamish:dataverse pdurbin$ export PID=doi:10.70122/FK2/9BXT5O
beamish:dataverse pdurbin$ 
beamish:dataverse pdurbin$ curl -s -H "X-Dataverse-key:$API_TOKEN" $SERVER_URL/api/pids?persistentId=$PID | jq .
{
  "status": "OK",
  "data": {
    "id": "10.70122/fk2/9bxt5o",
    "state": "findable"
  }
}
beamish:dataverse pdurbin$ export PID=doi:10.70122/FK2/9BXT5Ozzz
beamish:dataverse pdurbin$ 
beamish:dataverse pdurbin$ curl -s -H "X-Dataverse-key:$API_TOKEN" $SERVER_URL/api/pids?persistentId=$PID | jq .
{
  "status": "ERROR",
  "code": 404,
  "message": "API endpoint does not exist on this server. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.",
  "requestUrl": "http://localhost:8080/api/v1/pids?persistentId=doi:10.70122/FK2/9BXT5Ozzz",
  "requestMethod": "GET"
}
beamish:dataverse pdurbin$ 
```

qqmyers: Hmm - 404 seems OK, but the message should probably be updated. I'll change it to "The resource does not exist in this server". If you have a better idea, go ahead and change it. If you think it needs a custom response, I think the way you did the 500 error would work. (FWIW: See https://stackoverflow.com/questions/9930695/rest-api-404-bad-uri-or-missing-resource for a discussion of best practice. I haven't looked too far beyond that, but it seems reasonable.)

pdurbin: @qqmyers honestly, I was happier back with my original commit for this pull request (116379e). We should probably have a call because there's a lot to unpack. For now I'll move this out of code review and put both of us on it.

qqmyers: The main concern I have for the PR is to get all of the possible exceptions that can occur to result in errors rather than OKs. If you want to cover those a different way, I don't have a problem - just thought it would be easier to show code than trying to add comments. In any case, happy to discuss to help resolve things.

djbrooke: @qqmyers @pdurbin thanks for the discussion here. If you're both around next week can you take a few minutes to catch up and figure out the next steps to keep this moving?

pdurbin: Sure, some of this is process:

- Please don't change my code without asking.
- Let's not change our "this API doesn't exist" message for an issue about PIDs. Let's please consider this separately.

Some if it is technical:

- If a PID doesn't exist on the DataCite side, I'm not sure we should return a 404 from the Dataverse side. What we should definitely return is an informative message. My intention with this API is to allow Dataverse admins to get information about the state of the world on the DataCite side without having to use DataCite's API directly.

djbrooke: Sounds good, happy to join in also if there's some question of scope.

pdurbin: @qqmyers I know we haven't had a chance to talk yet but I'm putting this back in code review. Maybe you can take a look and ping me on Slack. Apart from merging the latest from develop...

In 00cf232c7 I reverted the change to the "endpoint doesn't exist" message. I'm open to revising this message as a separate effort.

In c359cbe5f I'm not exposing more detail to the Dataverse API user that the the 404 is coming from DataCite and that we're merely passing the information along. Here's how it looks:

```
curl -s -H "X-Dataverse-key:$API_TOKEN" $SERVER_URL/api/pids?persistentId=$PID -i
HTTP/1.1 404 Not Found
Server: Payara Server  5.2020.6 #badassfish
X-Powered-By: Servlet/4.0 JSP/2.3 (Payara Server  5.2020.6 #badassfish Java/AdoptOpenJDK/1.8)
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: PUT, GET, POST, DELETE, OPTIONS
Access-Control-Allow-Headers: Content-Type, X-Dataverse-Key
Content-Type: application/json;charset=UTF-8
Content-Length: 95
X-Frame-Options: SAMEORIGIN

{"status":"ERROR","message":"404 (NOT FOUND) from DataCite for DOI doi:10.70122/FK2/9BXT5Ozzz"}
```

qqmyers (approved):

Issue #6497, opened 2020-01-09
Semantic Metadata API calls


qqmyers:
The current API for getting, editing, deleting dataset metadata requires understanding Dataverse's internal metadata representation and does not map local terms to global identifiers as is now supported in metadatablocks. In contrast, the OAI_ORE export exposes metadata in a way that is more repository-agnostic / globally-interpretable (see excerpt below). 

The RDA grant to provide a standards-based archival format that can be exported/imported will need to import based on the OAI_ORE information. To do that, having a way to add/edit metadata, just given the information in the OAI_ORE file would be useful. (Nominally adding information about Dataverse's internal representation to the OAI_ORE would make it possible to format things as required by the current API, but part of the goal is to be interoperable with Bags/OAI_ORE files produced by other repositories which would not have this info.)

I propose to add new add/edit/delete API calls for dataset metadata that would have relatively flat JSON-LD payloads to support this work that I hope will be generally useful/easier to use than the current calls and that I hope can be retained as further work to update Dataverse's metadata capabilities (which I understand may replace/supplement metadatablocks) moves forward.

The general concept would be to allow upload in a format similar to the following, which is just a JSON-LD excerpt from an example dataset's ORE map. FWIW - I haven't fully scrubbed this example to remove things like the dataset version number which is ~read-only or the includedInDataCatalog entry that represents the dataset's location in a dataverse rather than metadata, but I wanted to show the general concept that there is just a set of key/value pairs where there keys are then defined in the `@context` in terms of a global identifier that can be used by Dataverse or other repository to map the item to the correct internal field. (I.e. the initial implementation of these api calls would look at the global term and then look it up in the metadata block definition to identify how to store it, essentially reversing the logic used to scan through metadata fields and look at the block definition to find the global term that is used to produce the ORE map to start with.) Unless/until further work is done, e.g. to allow Dataverse to accept new metadata terms dynamically, these api calls could return an error if the metadata terms/values do not match with configured metadata blocks. Similarly, they could co-exist with the existing APIs since they don't change anything w.r.t. the internal representation. 

Given the relatively small size and short timescale for the RDA grant work, I'd appreciate any comments/concerns/suggested alternative approaches, etc. asap. (FWIW: I plan to start working on some of the logic to match a globally-defined term to an internal dataverse field to start - seems like this is going to be needed in some form to handle externally produced Bags, regardless of whether the metadata is eventually submitted through the existing apis or the ones proposed here.)

```
{
[...]

Issue #5899, opened 2019-05-31
Allow terms metadata to be changed with Native API


jggautier:
Some users want to be able to change Terms of Use metadata for multiple datasets in bulk (without need to open and edit each dataset's metadata). Users can use the Native API to add or edit metadata in the metadata fields that are in Dataverse's metadata blocks, but as of 4.19 the Native API doesn't let users do the same for Terms of Use and Terms of Access fields (in the Terms tab). In fact, if you use [the "Update Metadata"](http://guides.dataverse.org/en/latest/api/native-api.html#update-metadata-for-a-dataset) endpoint to create or edit a dataset draft, which will overwrite any existing dataset metadata, all values in the Terms of Use and Terms of Access fields are also removed.

It would also be helpful to document in some brief way which fields can and can't be edited using the Native API.

(This is following discussion from the related issue at https://github.com/IQSS/dataverse/issues/2585)

pdurbin:
@jggautier thanks for creating this issue! 🎉 

> gives the impression that using the Native API to change Terms of Use has been possible since 4.8.6

I wish but #4431 is about being able to see the API Terms of Use from https://dataverse.harvard.edu/api/info/apiTermsOfUse for example. Here's the setting: http://guides.dataverse.org/en/4.14/installation/config.html#apitermsofuse

jggautier:
Adding the workaround that Phil suggested in #2585:

> Also, as we discussed, a workaround for now, in the absence of being able to update terms via the native API might be:
> 
> - serialize your dataset as native dataverse JSON format.
> - change the terms of use via SWORD
> - reupload your metadata from the JSON backup you made
> 
> The reason for backing up and re-applying the native JSON format is that with SWORD your metadata will be overwritten with very few (Dublin Core) fields. At least, that's how I believe it works. It's been a while since I've tested this.

A Dataverse user is trying this workaround now and working through issues (https://help.hmdc.harvard.edu/Ticket/Display.html?id=276348)

djbrooke:
- We want the granular option to edit terms, specifically "Terms of Use" and "Restricted Files + Terms of Access" (see screenshot)
- Add docs for what works and what does not work
- We'd want the workflow to be the same as what's in the UI, in that the option to provide terms is only available to those users deselecting CC0 and creating a new dataset version
- Do we want to just accept terms JSON that people send and change it off CC0 (do we do it silently or do we provide info)? - we want to provide info per @djbrooke 

[...]

Issue #4259, opened 2017-11-03
Java 11 Upgrade (no free security fixes for Oracle Java 8 after January 2019)


matthew-a-dunlap:
At some point we should upgrade to Java 9 to gain the power of new java technology and remain secure.

Outcome of the #4217 spike.

matthew-a-dunlap:
[Info from #4217]

Java 9: Looks pretty good, I seem to have it mostly running after a few hours today, but we probably don't want to upgrade yet.

Details:

Tested on Oracle JDK 9.0.1+11 on osx 10.13
Need Netbeans 9 for Java 9 to compile easily. There is only a developer build of Netbeans 9 and no release date (but active development). May be able to work around this with digging into maven but I did not do this.
Builds w/o tests after cleaning up a few new ambiguous references (from new classes brought into Java 9)
Tests have 1 error, the root cause seems to be that MimetypesFileTypeMap() getContentType is not returning right for png. Looking like implementation weirdness from Java 9, for now I'm just leaving it broken.
Starts with more sql noise than I remember, but its been a while since I deployed from command line (see java 9 sql error.txt).
Most of the simple dataverse interactions work! Create/delete dataset/dataverse works. Create user works. Uploading data DOES NOT work (see java 9 file upload error.txt).
TravisCI works fine switching to Java9.
Glassfish 4.1 itself is unable to detect the JDK version for Java 9. Glassfish needs to be set to explicitly use Java 8.

pdurbin:
I'm glad we now have a issue dedicated to Java 9. I like working in small chunks. Thanks @matthew-a-dunlap 

Here's the comment I left on #4217.

> Java 9

I don't believe it makes sense to upgrade to Java 9 at this time. It's only been out for a few weeks so Linux vendors haven't had time to package it as `java-1.9.0-openjdk`. We want the same easy upgrade that we had from Java 7 to Java 8 where we simply change a single digit in a `yum` command like we did in 1abbd77 for #2151.

[...]

Issue #3068, opened 2016-04-07
"Compact"/Make complete the native API output


raprasad:
This is probably worth a separate FRD and includes (certainly not limited to) items such as:
- For files, include API access links
- Making sure the API is complete.  e.g Files should include file tags: #3067 
  - The original native API was written well over 18 months ago
- Compacting the API/Removing extranneous information -- see example in the next comment

This is not a "massive" project but modifications on the existing code.

raprasad:
Example of current API output for dataset metadata and a "cleaned up version."
Cleaning available via a quick test script: https://github.com/IQSS/json-schema-test/blob/master/filemetadata/api_test/metadata_transformer.py

## Current version (12,172 bytes)

```javascript
"metadataBlocks": {
            "citation": {
                "displayName": "Citation Metadata", 
                "fields": [
                    {
                        "typeName": "title", 
                        "multiple": false, 
                        "typeClass": "primitive", 
                        "value": "North Carolina Vital Statistics -- Birth/Infant Deaths 1976"
                    }, 
                    {
                        "typeName": "author", 
                        "multiple": true, 
                        "typeClass": "compound", 
                        "value": [
                            {
                                "authorName": {
                                    "typeName": "authorName", 
                                    "multiple": false, 
                                    "typeClass": "primitive", 
                                    "value": "State Center for Health Statistics"
                                }
                            }
                        ]
                    }, 
                    {
                        "typeName": "datasetContact", 
                        "multiple": true, 
                        "typeClass": "compound", 
                        "value": [
                            {
                                "datasetContactName": {
                                    "typeName": "datasetContactName", 
                                    "multiple": false, 
                                    "typeClass": "primitive", 
                                    "value": "David Sheaves"
                                }, 
                                "datasetContactEmail": {
                                    "typeName": "datasetContactEmail", 
                                    "multiple": false, 
                                    "typeClass": "primitive", 
                                    "value": "david_sheaves@unc.edu"
                                }
                            }
                        ]
                    }, 
                    {
                        "typeName": "dsDescription", 
                        "multiple": true, 
                        "typeClass": "compound", 
                        "value": [
                            {
                                "dsDescriptionValue": {
                                    "typeName": "dsDescriptionValue", 
                                    "multiple": false, 
                                    "typeClass": "primitive", 
[...]

PR #7414, opened 2020-11-16
Iqss/6497 semantic api

qqmyers:
**What this PR does / why we need it**: This adds an API call that allows metadata to be added using json-ld - in the same format as the OAI-ORE export, and without having to retrieve the existing metadata and having to understand the way Dataverse is storing things internally. It's one step in supporting dataset migration via RDA Bags (see #6497) that might be useful independent of additional api calls to migrate a dataset. For example, it allows submission of new terms and conditions.

Closes #

**Suggestions on how to test this**: see examples in comments below

**Additional documentation**:


qqmyers:
FWIW: The compile issue here is due to the titianium json-ld library which requires Java 11+.  So this is dependent upon #4259.

janvanmansum:
@qqmyers I am not sure but the problem reported by Jenkins may be the same problem that I ran into. If a pre-publication workflow is set, the `successMsg` never gets initialized, leading to a stack-trace like this:

```
[#|2021-04-26T20:07:23.679+0200|SEVERE|Payara 5.2020|edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler|_ThreadID=90;_ThreadName=http-thread-pool::jk-connector(5);
_TimeMillis=1619460443679;_LevelValue=1000;|
        at org.glassfish.json.JsonObjectBuilderImpl.validateValue(JsonObjectBuilderImpl.java:198)
        at org.glassfish.json.JsonObjectBuilderImpl.add(JsonObjectBuilderImpl.java:66)
        at edu.harvard.iq.dataverse.api.Datasets.publishMigratedDataset(Datasets.java:1294)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
```

You probably want a `successMsg` for the workflow branch of this if-statement: https://github.com/GlobalDataverseCommunityConsortium/dataverse/blob/e584d6fa0b10ad5eb1bd275abe445b554470693c/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1271

qqmyers:
@janvanmansum   - Thanks - updated to follow the same conventions as the publish command - 202/Accepted if there's a workflow and 200/OK if not. (Unlike the UpdateCurrentVersion code I copied form, there isn't really a need for a specific successMsg since there is only one possibility per use case (with or without a pre-pub workflow).)

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
...
+an experimental API has been created that provides a json-ld alternative the the v1.0 API calls to get/set/delete Dataset metadata in the :doc:`/api/native-api`.
+
+You may prefer to work with this API if you are building a tool to import from a Bag/OAI-ORE source or already work with json-ld representations of metadata, or if you prefer the flatter json-ld representation to Dataverse software's json representation (which includes structure related to the metadata blocks involved and the type/multiplicity of the metadata fields.) 
+You may not want to use this API if you need stability and backward compatibility (the 'experimental' designation for this API implies that community feedback is desired and that, in future Dataverse software versions, the API may be modified based on that feedback).
+
+Note: The examples use the 'application/ld+json' mimetype. For compatibility reasons, the APIs also be used with mimetype "application/json-ld"
```
pdurbin, line 10: What's the story or history on "application/ld+json" vs. "application/json-ld"? Why support both?

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
...
+  export VERSION='1.0'
+  export SERVER_URL=https://demo.dataverse.org
+ 
+  Example: Change the Dataset title 
+ 
+    curl -X PUT -H X-Dataverse-key:$API_TOKEN -H 'Content-Type: application/ld+json' -d '{"Title": "Submit menu test", "@context":{"Title": "http://purl.org/dc/terms/title"}}' "$SERVER_URL/api/datasets/$DATASET_ID/metadata?replace=true"
+
+  Example 2: Add a description using the DATASET PID
+
+    curl -X PUT -H X-Dataverse-key:$API_TOKEN -H 'Content-Type: application/ld+json' -d '{"citation:Description": {"dsDescription:Text": "New description"}, "@context":{"citation": "https://dataverse.org/schema/citation/","dsDescription": "https://dataverse.org/schema/citation/dsDescription#"}}' "$SERVER_URL/api/datasets/:persistentId/metadata?persistentId=$DATASET_PID"
```
pdurbin, line 55: Should we have URLs like https://dataverse.org/schema/citation/dsDescription actually work?

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export DATASET_ID='12345'
+  export DATASET_PID='doi:10.5072/FK2A1B2C3'
+  export VERSION='1.0'
+  export SERVER_URL=https://demo.dataverse.org
+ 
+  Example: Delete the TermsOfUseAndAccess 'restrictions' value 'No restrictions' for the latest version using the DATASET PID
```
pdurbin, line 73: This sounds like a fix for this issue:

-  Allow terms metadata to be changed with Native API #5899

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
...
+
+    curl -X PUT -H X-Dataverse-key:$API_TOKEN -H 'Content-Type: application/ld+json' -d '{"https://dataverse.org/schema/core#restrictions":"No restrictions"}' "$SERVER_URL/api/datasets/:persistentId/metadata/delete?persistentId=$DATASET_PID"
+
+Note, this example uses the term URI directly rather than adding an '@context' element. You can use either form in any of these API calls.
```
pdurbin, line 77: It's good to mention this "either form" factoid here but we should also mention it above somewhere in the main intro.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
...
+Create a Dataset
+----------------
+
+Specifying the Content-Type as application/ld+json with the existing /api/dataverses/{id}/datasets API call (see :ref:`create-dataset-command`) supports using the same metadata format when creating a Dataset.
+
+With curl, this is done by adding the following header:
+
+.. code-block:: bash
+
+  -H 'Content-Type: application/ld+json'
```
pdurbin, line 91: I'm confused about how this just sort of ends. Can we have a downloadable JSON-LD file to use as an example when creating a dataset?

scripts/api/data/metadatablocks/migration.tsv
```diff
@@ -0,0 +1,5 @@
+#metadataBlock	name	dataverseAlias	displayName	blockURI
```
pdurbin, line 1: Do we need this migration.tsv? It isn't loaded by default on new installations?

scripts/search/tests/data/dataset-finch1.jsonld
```diff
@@ -0,0 +1,26 @@
+
+{
+  "http://purl.org/dc/terms/title": "Darwin's Finches",
+  "http://purl.org/dc/terms/subject": "Medicine, Health and Life Sciences",
+  "http://purl.org/dc/terms/creator": {
+      "https://dataverse.org/schema/citation/author#Name": "Finch, Fiona",
+      "https://dataverse.org/schema/citation/author#Affiliation": "Birds Inc."
+  },
```
pdurbin, line 8: How do I specify multiple authors?

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
```diff
@@ -168,7 +169,8 @@ public Response getByName(@PathParam("name") String name) {
                     .add("parentAllowsMultiples", parentAllowsMultiplesDisplay)
                     .add("solrFieldSearchable", solrFieldSearchable)
                     .add("solrFieldFacetable", solrFieldFacetable)
-                    .add("isRequired", isRequired));
+                    .add("isRequired", isRequired)
+                    .add("uri", uri));
```
pdurbin, line 173: I'm happy to see someone updating this old API. 😄

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -651,6 +651,96 @@ public Response updateDraftVersion( String jsonBody, @PathParam("id") String id,
...
+            logger.log(Level.SEVERE, "Error parsing dataset json. Json: {0}", jsonLDBody);
+            return error(Status.BAD_REQUEST, "Error parsing Json: " + jpe.getMessage());
+        }
+    }
+    
+    @PUT
+    @Path("{id}/metadata/delete")
+    @Consumes("application/ld+json, application/json-ld")
+    public Response deleteMetadata(String jsonLDBody, @PathParam("id") String id) {
+        logger.info("In delteMetadata");
```
pdurbin, line 717: This should be deleted or set to fine.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -651,6 +651,96 @@ public Response updateDraftVersion( String jsonBody, @PathParam("id") String id,
...
+        try {
+            Dataset ds = findDatasetOrDie(id);
+            DataverseRequest req = createDataverseRequest(findUserOrDie());
+            DatasetVersion dsv = ds.getEditVersion();
+            boolean updateDraft = ds.getLatestVersion().isDraft();
+            dsv = JSONLDUtil.deleteDatasetVersionMDFromJsonLD(dsv, jsonLDBody, metadataBlockService, datasetFieldSvc);
+            logger.info("Updating ver");
```
pdurbin, line 724: Same. Delete or set to fine.

src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -816,7 +824,7 @@ private String generateInfoFile() {
         } else {
             info.append(
                     // FixMe - handle description having subfields better
-                    WordUtils.wrap(getSingleValue(aggregation.getAsJsonObject(descriptionTerm.getLabel()),
+                    WordUtils.wrap(getSingleValue(aggregation.get(descriptionTerm.getLabel()),
```
pdurbin, line 827: Should this be getAsString?

src/main/java/edu/harvard/iq/dataverse/util/json/JsonLDTerm.java
```diff
@@ -48,6 +45,8 @@
     public static JsonLDTerm totalSize = JsonLDTerm.DVCore("totalSize");
     public static JsonLDTerm fileCount = JsonLDTerm.DVCore("fileCount");
     public static JsonLDTerm maxFileSize = JsonLDTerm.DVCore("maxFileSize");
+    
+    public static JsonLDTerm metadataOnOrig = JsonLDTerm.DVCore("metadataOnOrig");
```
pdurbin, line 49: Ah, it seems like maybe all installations do need the new .tsv file. If so, it should be added to the script that adds them all during installation.

pdurbin (commented):
Overall, this is looking really good and I'm excited to have a smaller, more compact JSON format for creating a dataset. That was the dream of #3068 (which focused on output, initially).

I left comments and questions throughout. I'd be happy to get on a call to discuss them. I haven't run the code locally yet but I might like to do that too.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -2164,4 +2157,111 @@ public void testRestrictFileExportDdi() throws IOException {
...
+        jsonLDString = getData(response.getBody().asString());
+        JsonObject jsonLDObject = JSONLDUtil.decontextualizeJsonLD(jsonLDString);
+        assertEquals("New Title", jsonLDObject.getString("http://purl.org/dc/terms/title"));
+
+        // Add an additional description (which is multi-valued and compound)
+        // Also add new terms of use (single value so would fail with replace false if a
+        // value existed)
+        String newDescription = "{\"citation:Description\": {\"dsDescription:Text\": \"New description\", \"https://dataverse.org/schema/core#termsOfUse\": \"New terms\"} \"@context\":{\"citation\": \"https://dataverse.org/schema/citation/\",\"dsDescription\": \"https://dataverse.org/schema/citation/dsDescription#\"}}";
+        response = UtilIT.updateDatasetJsonLDMetadata(datasetId, apiToken, newDescription, false);
+        response.then().assertThat().statusCode(OK.getStatusCode());
```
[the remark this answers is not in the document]
qqmyers: Some typos in the json-ld -should be fixed

conf/solr/8.8.1/schema_dv_mdb_copies.xml
```diff
@@ -0,0 +1,162 @@
+<schema>
```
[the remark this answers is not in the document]
qqmyers: No reason - it's gone.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
+Dataset Semantic Metadata API
+=============================
+
+The OAI_ORE metadata export format represents Dataset metadata using json-ld (see the :doc:`/admin/metadataexport` section). As part of an RDA-supported effort to allow import of Datasets exported as Bags with an included OAI_ORE metadata file,
```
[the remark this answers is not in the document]
qqmyers: It did support development - via a small GDCC grant in the first part of 2020.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,91 @@
+Dataset Semantic Metadata API
+=============================
+
+The OAI_ORE metadata export format represents Dataset metadata using json-ld (see the :doc:`/admin/metadataexport` section). As part of an RDA-supported effort to allow import of Datasets exported as Bags with an included OAI_ORE metadata file, 
+an experimental API has been created that provides a json-ld alternative the the v1.0 API calls to get/set/delete Dataset metadata in the :doc:`/api/native-api`.
```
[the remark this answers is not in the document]
qqmyers: It's still all v1.0.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: I had json-ld to start, which is what DANS has coded to and what the DVUploader can support now (DVUploader can, at least as a proof-of-concept, read archived Bags and recreate a dataset using this PR and the related migrate API PR - that was part of the RDA proposal). As far as I can tell,  ld+json is newer/more of a standard so I added support for that. I kept the older one just to avoid having to immediately update those other components.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: That would be useful. These URLs were defined a while back with the changes to add URLs to the metadata blocks and export them in the OAI-ORE/Bag structures. It would probably require some discussion of what should actually be available from such links and where that info would come from (given that new metadata blocks can define new URLs).

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: Yes - this in particular is Sciences PO's interest in this. DANS is using it for their overall migration effort. It wasn't clear to me though whether this should close that issue or not.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: This is basically standard JSON-LD info - @contexts can always be replaced by a no-@context form that has the same meaning. The good thing is that the code in the PR normalizes whatever JSON-LD it gets, so any valid form (with/without @context, using a @context different than the one Dataverse uses in the OAI_ORE exports) will work. That's nominally implicit in saying the payloads are JSON-LD though.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: I added an example.

scripts/api/data/metadatablocks/migration.tsv
qqmyers: It is not required and is not auto-loaded. It's a bit of a complex situation: The RDA grant had, as one goal, the idea of being able to import Bags created by other repos. An expectation there is that the other repo (a Dataverse with different blocks or some other repo software entirely) could have metadata that doesn't fit into the receiving Dataverse's schema (including blocks). So - the metadataOnOrig field is a json structured field that can store any/all metadata that doesn't match a known field. So a transfer can be done without losing metadata. That said, putting metadata in this field makes it less useful than if there were a matching field. The current code will ignore metadata that doesn't match if this block is not installed and will use it if it is. I can explain that, or remove the code from the PR. For migration cases, I think even this limited functionality could be useful, but I suspect that most people using the 'experimental' API here won't want to enable it.

scripts/search/tests/data/dataset-finch1.jsonld
qqmyers: An array of objects with the name/affiliation keys as the value of the creator key. This is the same way the OAI-ORE export shows multiple authors.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
qqmyers: This is mostly a separate thing. Slava pointed out that this API didn't include the URI for a term (which was added with the OAI-ORE/Bag work from QDR). Since this was a one line change related to semantics, I added it here but the rest of the changes do not depend on this API having the extra field.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: removed

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: removed

src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
qqmyers: Description is a multi-valued field so this could be a JsonObject or JsonArray. The added logic here handles both cases. Nominally this is also separable from the API itself - just caught in testing along the way. Could be moved to a separate PR.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonLDTerm.java
qqmyers: As noted above, this is optional/not needed by default. If we decide to remove the functionality, this line would also go.

qqmyers:
@pdurbin - thanks for reviewing! I think I've addressed the straight-forward issues. The one thing that is tbd would be to decide on whether to leave the metadataOnOrig functionality in or not. It could probably be pulled out for now and treated as another PR (that could sit for a while) if there's concern about having it in the code now. (FWIW: I think DANS has created a custom metadata block for the metadata they had in their old system that doesn't match and therefore isn't relying on the metadataOnOrig functionality.)

If I missed something, let me know. I'm also happy to just talk through things as well.

doc/sphinx-guides/source/_static/api/dataset-create.jsonld
```diff
@@ -0,0 +1,15 @@
...
+  "http://purl.org/dc/terms/subject": "Medicine, Health and Life Sciences",
+  "http://purl.org/dc/terms/creator": {
+      "https://dataverse.org/schema/citation/author#Name": "Finch, Fiona",
+      "https://dataverse.org/schema/citation/author#Affiliation": "Birds Inc."
+  },
+  "https://dataverse.org/schema/citation/Contact": {
+    "https://dataverse.org/schema/citation/datasetContact#E-mail": "finch@mailinator.com",
+    "https://dataverse.org/schema/citation/datasetContact#Name": "Finch, Fiona"
+  },
+  "https://dataverse.org/schema/citation/Description": {
```
pdurbin, line 12: The capitalization seems inconsistent.

Contact and Description (title case) but also author, datasetContact and dsDescription (camel case)?

What are the rules?

pdurbin (changes requested):
Overall, the changes are looking good. A few things on my mind:

- As suggested by @qqmyers, I vote we remove the metadataOnOrig code and tsv if we really don't need it. It's the kind of thing we can add in later with more discussion, if necessary. If we decided to keep it, it definitely needs to be documented.
- In the guides, I'd be happy to have a little more of an intro to JSON-LD and examples of how to create a dataset with one author vs. multiple, for example. We can live without this though, so please consider it a "nice to have".
- Under "Suggestions on how to test this" it says "see examples in comments below" but I'm confused about what this means. Perhaps it means the examples in the guides. I'd suggest clarifying this.
- I'm still slightly weirded out that we have URLs like https://dataverse.org/schema/citation/Description in examples but they're 404s.
- I left a comment about the case being used in the JSON-LD examples. At first glance it doesn't seem consistent.

doc/sphinx-guides/source/_static/api/dataset-create.jsonld
qqmyers: These come from the citation.tsv block with the pattern <block name>/<field title> for single fields and <block name>/<field title>/<child field name> being used. That choice was made back when OAI_ORE/BagIt/archiving was introduced. The use if title was an attempt to use URIs that mirrored what users see in the UI. I'm less sure why I used name for child fields - not sure if there was a conflict or if it was an issue with originally just trying a flatter <blockname>/<childfield title> and realizing that there are multiple child fields with the title 'Name' for example.

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
qqmyers: It would be great to have some review of these 'internal terms' both to see if there are external terms that make sense (I did try to avoid them when I did the original archiving work but some terms either didn't have a match at all or matched in meaning but didn't allow the values Dataverse has (either multiple versus single values or things like contact having to be a person or requiring a phone number in the value when Dataverse doesn't have that, etc. - basically term range issues). Doing that and doing something more formal to give the remaining real ones live URLs with some info about that term (deciding whether the definitions are community wide (with a single community URL) or are instance specific (with URLs related to the specific machine), etc.) would be good. Perhaps at least creating an issue would be a good step?

scripts/search/tests/data/dataset-finch1.jsonld
qqmyers: Having a human-readable term label (that probably isn't globally unique/machine interpretable) while still supporting strict machine readability adds some complexity, but its pretty straight forward to either never use an @context or to always use a standard/static one.

Regardless, it wouldn't be hard to add documentation over time.

qqmyers:
I've removed the metadataOnOrig logic (and created https://github.com/GlobalDataverseCommunityConsortium/dataverse/tree/RDA-with_metadataOnOrig to keep that code around as a proof-of-concept for RDA/if/when needed).

src/main/java/edu/harvard/iq/dataverse/util/json/JsonLDTerm.java
pdurbin: Removed in a5a745d.

doc/sphinx-guides/source/_static/api/dataset-create.jsonld
pdurbin: Ok. Sounds like a preexisting condition to me. 😄 It would be nice to have more consistency but oh well. I assume we don't want to revisit decisions made during BagIt export.

pdurbin (approved):
I still haven't run the code myself but this seems ready for QA. Thanks for the pull request and for hanging there with changes I asked for, @qqmyers!

PR #7415, opened 2020-11-16
prevent double linking of datasets and dataverses #7259

### LINKED ISSUE #7259: Linking Dataset API called twice will create two rows in db
When calling the link dataset API, if you call it twice for the same dataset - dataverse combo, it will not fail the 2nd time, but will say success and add a second (duplicate) row to the db. You then can't use the unlink, because it throws a 500 (probably because query doesn't return single result).

2nd row needs to then be deleted directly in the db.

This should be caught in the command. (and also should be checked for Dataverse links, as well)

pdurbin:

Users have been able to link a dataset to the same dataverse twice (or more) via API. This prevents it.

Update: in testing, it was found that it's also possible to link a dataverse to the same dataverse more than once. A fix has been added for this too.

Closes #7259

~I tacked a test on an existing test method but I'm happy to create a dedicated test method if that's preferred.~ Update: a new LinkIT test class was added.

- Try to link the same dataset to the same dataverse via API multiple times.
- Try to link the same dataverse to the same dataverse via API multiple times.

No.

Probably not worth it. Just a bug fix.

**Additional documentation**:

None.


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/LinkDatasetCommand.java
```diff
@@ -50,6 +52,13 @@ public DatasetLinkingDataverse execute(CommandContext ctxt) throws CommandExcept
         if (linkedDataset.getOwner().getOwners().contains(linkingDataverse)) {
             throw new IllegalCommandException(BundleUtil.getStringFromBundle("dataset.link.not.to.parent.dataverse"), this);
         }
+        List<Dataverse> existingLinksToDataverses = new ArrayList<>();
+        for (DatasetLinkingDataverse datasetLinkingDataverse : linkedDataset.getDatasetLinkingDataverses()) {
```
scolapasta, line 56: will this be efficient enough, instead of adding a find method specific for an existing link?

src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
```diff
@@ -254,6 +254,12 @@ public void testMoveLinkedDataset() {
         linkDataset.then().assertThat()
                 .statusCode(OK.getStatusCode());
 
+        // A dataset cannot be linked to the same dataverse again.
```
scolapasta, line 257: is this test in the move tests? I see what you mean then - it works, but I wonder if we won't lose track that it's being tested here.

scolapasta (commented):
Added a few comments / questions. Also, do we know if the same issue exists for linking dataverses? if so, should we add similar logic to LinkDataverseCommand?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/LinkDatasetCommand.java
pdurbin: @scolapasta and I just discussed this and decided that we'd see how it goes in QA and optimize later if need be.

src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
pdurbin: Tests moved to their own class in 6bccd4f.

pdurbin:
> Also, do we know if the same issue exists for linking dataverses? if so, should we add similar logic to LinkDataverseCommand?

Yes, the same issue exists. Fixed in 5d4a3dc.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/LinkDatasetCommand.java
pdurbin: Good news, I found an alreadyLinked method and switched to it in 109a3d3a1.

pdurbin:
I'm going to go ahead and put this into code review even though I might add another test for #7430 (as discussed in Slack). Here are the changes since the last code review:

- Tests have been moved to a dedicated class called LinkIT.
- In fact the same problem happens with linking dataverses. Fixed and I updated the description of this pull request.
- The potential performance problem noticed in the last review was fixed by switching to an "alreadyLinked" method.

pdurbin:
I moved it back to in progress to reflect that I'm still hacking on tests.

pdurbin:
In 403a966 I wrote some tests that demonstrate that #7430 is a real issue. Hopefully they will be a good starting point when we pick that issue up. I'm moving this back to code review.

scolapasta (approved):

Issue #7418, opened 2020-11-17
Move database configuration into application


poikilotherm:
Currently, the database connection is handled on the application server level.

As outlined in #6819 and #7048, this isn't necessary due to Java EE 7 `@DataSourceDefinition`.
Using MicroProfile Config API #7000, it is a piece of cake to configure.

Let's go for this. Requires #6819 and #7417 to be merged first.

PR #7416, opened 2020-11-17
5345 Make EJB timers nonpersistent

poikilotherm:
**What this PR does / why we need it**:
As discussed in #5345, #7048, on IRC, Slack and Zoom plus [this Google doc](https://docs.google.com/document/d/1MU8Qo7EYmy75LaiwabzdID-Co3QriEe0i-YxV98hnSA/edit), we need to make the EJB timers non-persistent for enabling application-level database connection definitions.

**Which issue(s) this PR closes**:

Closes #5345

**Special notes for your reviewer**:
- It might me an idea to add a Flyway migration dropping the unused EJB timer table from the DB.

**Suggestions on how to test this**:
- Remember to reconfigure the EJB timer database (see release note)
- Check if timers still work.
- Check if timers still work on multi-instance installation without conflicting runs.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Nope.

**Is there a release notes update needed for this change?**:
Yeah, a small one about removing the EJB timer database configuration setting on the app server. :battery: included.

**Additional documentation**:
This is a quick and dirty solution, not very sophisticated for future enhancements. At some point in the future, we need to come back to the ideas of the Google Doc.

PR #7420, opened 2020-11-17
6819 jdbc driver war

### LINKED ISSUE #6819: Include Postgres JDBC driver in pom.xml

poikilotherm:
Move the JDBC driver to the WAR, loading it from there instead of from the Glassfish libs.
No more manual installation, keeping all deps tight together via Maven POM. A huge benefit for testing and containers.

Closes #6819
Should be done after #7416

None?

Try usual database operations. If tested before #7416, check EJB timers.

Nope.

Nope. Those will be necessary for #7418.

**Additional documentation**:
Nada.


scolapasta (approved):
looks good (and as discussed)

poikilotherm:
🥳🥳🥳🥳

Thanks for merging @kcondon !

PR #7434, opened 2020-11-20
7263 dismissable banner

### LINKED ISSUE #7263: Inform users of Terms of Use Update, Dismissible Banner
We have a minor terms of use update coming for dataverse.harvard.edu. We've been directed that we can just inform users of this in the application, and this option is attractive to me as it does not require us to build a workflow for re-acceptance (covered in #4978). 

My first thought is a dismissable banner with a quick message and a link to the TOU that, once dismissed, does not appear again. We'd need to track the dismissal somehow. This feature may also be useful for other messages that we want to show once, such as a release announcement or some announcement about an upcoming community meeting.

We have StatusMessageHeader and StatusMessageText available to us (http://guides.dataverse.org/en/latest/installation/config.html#id174) but don't think either of these are what we want (as they are messages turned off and on by admins), but we do want to use that banner pattern:

Screen Shot 2020-09-11 at 2 23 25 PM

Thanks @TaniaSchlatter for the screenshot and also for stating this a different way: "...having 2 types of system alert banners that look and appear the same but have two different purposes; persistent until turned off by the admin, vs persistent until turned off by the user"

sekmiller:
Adds flexibility to the posting of messages to the application ui including the ability to target messages to logged in users. 

Closes #7263 Inform users of updates to TOU 

I added a view action of initBannerMessages to every page alongside initBreadcrumbs. Is there a better way to do that?
(Updated to move init functionality to get method)

These have been previewed/reviewed by the design team

This should be part of release notes especially since StatusMessage from the settings table will not work
**Additional documentation**:


scolapasta:
@sekmiller 

>I added a view action of initBannerMessages to every page alongside initBreadcrumbs. Is there a better way to do that?

My concern here is that we'd have to remember to add this for any new page. The same could probably be set for breadcrumbs, too)

I think if the only way method that is being called is getBanners (is this true?) then you can just add something there to initialize.

sekmiller:
@scolapasta - that makes sense. the advantage is that banners should be consistent across all pages while breadcrumbs really do need to be changed on just about every page.

scolapasta:
True! I had forgotten that point (and to be clear, my consistent and change, you mean what is passed to the "get" method in order to generate them. i.e. for breadcrumbs, it does make sense for each page to define how its breadcrumbs are defined.

pdurbin (changes requested):
Overall this is looking good but after playing around with it I noticed that if you delete a banner message all the history is deleted as well. That means that if I dismissed an important notice about a terms of service change, the admin will have no way of going back to see if or when I dismissed the notice. I explained this to @djbrooke and @TaniaSchlatter this morning and they said they'd like to keep the history.

sekmiller:
Added "active" as a boolean to the Banner Message table so that when the admin uses the DELETE api the active boolean is set to "false". This allows us to retain the history of users' dismissals of messages while hiding out of date messages.

sekmiller:
added deactivate as a separate endpoint and noted it in the doc.

pdurbin (approved):
Now that the "deactivate" idea is here (retaining history of who dismissed what when), this is ready for QA.

PR #7444, opened 2020-11-30
document default zip upload files limit, provide curl example

### LINKED ISSUE #7442: ZipUploadFilesLimit defaults to 1000
per
https://github.com/IQSS/dataverse/blob/390cccda99b128589032382102f3be6cddabf61a/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java#L95

but not
https://guides.dataverse.org/en/latest/installation/config.html#zipuploadfileslimit

we should document the default. PR forthcoming.

donsizemore:
**What this PR does / why we need it**: the ZipUploadFilesLimit setting defaults to 1,000 but this isn't documented in the configuration guide.

Closes #7442 

**Suggestions on how to test this**: just a doc change

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -1650,7 +1650,7 @@ For example, if you want your installation of Dataverse to not attempt to ingest
 :ZipUploadFilesLimit
 ++++++++++++++++++++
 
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
Looks good (thanks for the curl example!) and I confirmed that the default limit is 1000:

https://github.com/IQSS/dataverse/blob/v5.2/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java#L95

PR #7447, opened 2020-12-01
7398 saved search performance

scolapasta:
Adds some optimization for saved searches and some code cleanup

None for sure - it may fix #7398if the performance is improved enough

While I was at it, I made some Named queries and did some cleanup.

just make sure saved searches work as expected. You could also compare the timing of how long it takes to update a saved search before and after. (I will be doing some timing test with prod once released)

No

No

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetLinkingServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetLinkingServiceBean.java
@@ -27,27 +27,9 @@ public class DatasetLinkingServiceBean implements java.io.Serializable {
     @PersistenceContext(unitName = "VDCNet-ejbPU")
     private EntityManager em;
     
-    /**
-     * @param linkingDataverseId
-     * @return 
-     * @todo Should this method simply be deleted? It isn't used anywhere and is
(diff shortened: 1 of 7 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetLinkingServiceBean.java
```diff
@@ -60,31 +42,15 @@
 
     public List<Dataverse> findLinkingDataverses(Long datasetId) {
         List<Dataverse> retList = new ArrayList<>();
-        for (DatasetLinkingDataverse dld : findDatasetLinkingDataverses(datasetId)) {
-            retList.add(dld.getLinkingDataverse());
+        Query query = em.createQuery("select object(o) from DatasetLinkingDataverse as o where o.dataset.id =:datasetId order by o.id");
```
sekmiller, line 45: Any reason this wasn't added as a named query?

scolapasta: No, just because I was only changing the ones related specifically to saved search. I did think about it. And sure, I can add the others too, while I'm at it.

sekmiller (approved):
Looks good. passing to QA.

PR #7458, opened 2020-12-04
7406 access requested refresh

### LINKED ISSUE #7406: "Access Requested" only shown after page refresh on File and Dataset page
When access is requested to a file from the file table on the dataset page, the dropdown menu option switching from "Request Access" to the grayed-out "Access Requested" does not take place until after a page refresh. The same issue takes place on the file page. The expectation is that it would change without a refresh since this is the indication to the user that the request has gone through. If it doesn't change they are left wondering if the request was successful. 

To reproduce on the file page:

- Log in to demo.dataverse.org
- Go to a file on demo.dataverse.org that is restricted (pick one of these https://demo.dataverse.org/dataverse/demo?q=&fq1=fileAccess%3A%22Restricted%22&fq0=dvObjectType%3A%28files%29&types=files&sort=dateSort&order=)
- "Request Access" from the top right under the "Access File" dropdown
- In the "Access File" dropdown, note that "Request Access" is still an option and access can be requested again
- Refresh the page
- Note that "Request Access" has been grayed out and changed "Access Requested"

To reproduce on the dataset page, go through the same process but using the button on the file table instead. 

Running into this on 5.2, but may be an issue for 5.0 installations.

### LINKED ISSUE #7426: Restricted file with "Request Access" disabled shows empty "download" dropdown
If request access is enabled, you get a dropdown with a request access option (as designed:

image

But if request access is disabled, you get an empty dropdown:

image

sekmiller:
Fixes display/update issues around requesting access to restricted files

Closes #7406 and closes #7426 

There is some inline style sheet code in the file-download-button-fragment that will hopefully go away when we update prime faces/Bootstrap.
All manner of requesting access with and without terms of access on the dataset page and file page including datasets where requesting access is not allowed. 

no

**Additional documentation**:


pdurbin (approved):
Approved. The only thing that was a head scratcher to me was the extra inline CSS but I see from the pull request description that it may go away when we update primefaces/bootstrap. I didn't actually run the code but it seems reasonable.

PR #7422, opened 2020-11-18
7418 datasourcedefinition

poikilotherm:
**What this PR does / why we need it**:
This will move the database connection configuration from the application server down to the application.
This greatly simplifies how to configure the connection details by using MicroProfile Config API.

**Which issue(s) this PR closes**:

Closes #7418 
Requires #7416
Requires #7419 
Requires #7420 

**Special notes for your reviewer**:
None.

**Suggestions on how to test this**:
Be sure to apply the PRs mentioned above first. Follow release notes instructions. Do usual testing.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Nope.

**Is there a release notes update needed for this change?**:
:battery: included

**Additional documentation**:
None.

PR #7459, opened 2020-12-04
db_password_alias changed to dataverse.db.password #7418 #7422

pdurbin:

Pull request #7422 didn't include a doc change having to do with db_password_alias being renamed to dataverse.db.password.

None but is part of the issue that the original pull request closed: #7418

None.

Just a doc change.

No.

I don't think so. The upgrade instructions from the previous pull request explain about deleting the old alias.

**Additional documentation**:

None.


poikilotherm:
There's also a possibility to stay with the alias as is. Just explored this. See https://docs.payara.fish/community/docs/5.2020.6/documentation/payara-server/server-configuration/var-substitution/usage-of-variables.html#examples-ver-ref, which would allow us to add a `payara-expression-config.properties`:

```
dataverse.db.password={ALIAS=db_password_alias}
```

It would allow for great backward compatibility, also never installations should use the new format. :smile:

poikilotherm:
As requested on IRC by @pdurbin 

The name change in the alias (`db_password_alias` -> `dataverse.db.password`) happened because Payara has a MicroProfile Config Source "password", which reads from the aliases store, providing the password to the `DataSourceDefinition`.

This way, we can provide the database password from any of the possibe mpconfig sources consistently as "dataverse.db.password". It's important to be consistent here and make a single, reusable scope for anything database related.

My above comment was about an option I discovered to provide a backward compatibility, so existing installations wouldn't have to mangle the alias store during update.

landreev (approved):
@poikilotherm thanks for the explanation. 
@pdurbin once again - good eye, it would be confusing to keep mentioning this "db_password_alias" in the docs.

PR #7462, opened 2020-12-07
7449 restrict files fix

### LINKED ISSUE #7449: Rogue popup displayed for restrict files on Add New Dataset pg
Discovered in a development branch, and replicated in develop, and demo (v. 5.2 build 254-4951505).

1. create new dataset and upload file(s)
2. restrict file(s), Save Dataset, see File Restrictions popup displayed
3. (leave form empty or fill it out) click Continue, see Select File(s) popup mistakenly displayed

ALSO... if you "X' out of Select File(s) popup, it closes popup but the overlay remains covering the pg, same if you cancel on the File Restrictions popup.

The rogue Select File(s) popup does not appear on Edit Files pg, or when restricted on dataset pg, only on the Add New Dataset pg.

As you can see in this screenshot, the Select File(s) popup is displayed over a form that clearly shows there are in fact files selected, never mind the fact that it is displayed at a point in the restrict workflow that is completely inappropriate.

Screen Shot 2020-12-02 at 4 20 46 PM

### LINKED ISSUE #7456: Restricting / Unrestricting  Deleting a file from a published version via the kebab throws an error
Create a dataset with a file. 
Publish.
Then click Restrict under the kebab, and click save on the popup.
Save will fail with following UI:
image

The second message makes sense (since the save fail, the draft was not created).

from the server.log:

```Caused by: javax.persistence.OptimisticLockException: Exception [EclipseLink-5010] (Eclipse Persistence Services - 2.7.4.payara-p2): org.eclipse.persistence.exceptions.OptimisticLockException
Exception Description: The object [edu.harvard.iq.dvn.core.study.FileMetadata[id=33]] cannot be merged because it has changed or been deleted since it was last read. 
Class> edu.harvard.iq.dataverse.FileMetadata
```

scolapasta:
Fixes some issues with restrict access that were introduced with kebabs. Also, when editing / uploading files, moves the access popup earlier in the workflow to when you click restrict, instead of when you click save. Also, continued cleanup.

Closes #7449 
Closes #7456

Just to note, that there is still further work in the consolidation of this code. This work is primarily to fix these two issues and introduce this new workflow.

Test the cases in the issues, of course. But then generally, make sure these edit options work from different scenarios - select button, kebab, file page, from a published dataset, from a draft dataset, etc. Most changes related to restrict specifically (the workflow stuff), but one of the bugs had similar affect with delete.
Make sure canceling works as expected.

Yes, the above mentioned workflow change. Same popup (with some altered text), just shows up earlier.

Maybe the workflow change? I don't think it's necessary.

**Additional documentation**:


TaniaSchlatter:
I updated the popup header and text based on discussion in the design meeting, to better reflect that the popup operates on both the file and the dataset.

sekmiller (approved):
Looks good. I really like what you did with code consolidation especially all of those "select files popups"

Issue #7424, opened 2020-11-18
Move Java Mail configuration to application


poikilotherm:
As with JDBC #7418 and JMS #7423, we can move the mail configuration to the application. This further reduces the necessary bash glue to install Dataverse or use in containers. It eases reconfiguration, too.

Using Java EE 7 [`@MailSessionDefinition`](https://javaee.github.io/javamail/docs/api/javax/mail/MailSessionDefinition.html) for this.

This should make things much easier in cases like #3061, #5986, #5723, #4969. #5707

Let's ping some people from the mail service code base: @pdurbin @qqmyers @landreev @sekmiller

poikilotherm:
OK while this is fairly easy to do (add annotation to `MailServiceBean`), I stumbled over something.

The database setting `:SystemEmail` is used to set the "From" address of the message. It is even used to explicitly avoid blocking for mails. (Thx @qqmyers !)

`MailServiceBean.sendSystemMail()`:
https://github.com/IQSS/dataverse/blob/3fcc239491b878041b82f7d19619e5df47deb2c9/src/main/java/edu/harvard/iq/dataverse/MailServiceBean.java#L131-L131

`MailServiceBean.sendMail()`:
https://github.com/IQSS/dataverse/blob/3fcc239491b878041b82f7d19619e5df47deb2c9/src/main/java/edu/harvard/iq/dataverse/MailServiceBean.java#L183-L184

What I find very confusing: why should we do this? The "From" address is already configured for the JavaMail resource:

https://github.com/IQSS/dataverse/blob/3fcc239491b878041b82f7d19619e5df47deb2c9/scripts/installer/as-setup.sh#L180-L182

When a sysadmin configures a mail account from, lets say GMail, this might cause a situation where the authenticated user may not send a mail with a differing mail from `:SystemEmail`.

So while I add the annotation, I'd vote to ease the situation by:
1. Remove the DB setting
2. Make the annotation easily configurable via MicroProfile Config API
3. Rely on the `Session` to set the from mail for us (as configured).

After wrting all of this I found a prior issue on this: #4210 (thanks @donsizemore !)

PR #7446, opened 2020-12-01
7439 - 5.3 release notes

djbrooke:
**What this PR does / why we need it**:

The release notes for the 5.3 pre-holiday break release spectacular. 🎁 

**Which issue(s) this PR closes**:

Closes #7439

**Special notes for your reviewer**:

- Please check my usage of "obsolete" in regards to the DB settings for banner management. Not sure if it should be obsolete or deprecated, or something else. 
- ~~Please check where I put the Payara upgrade in the release steps. I put it after "Undeploy the Previous Version"~~ this was handled in the updates for a later release note from @poikilotherm 

**Suggestions on how to test this**:

Lint check, grammar/spelling check. :) 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

N/A 

**Is there a release notes update needed for this change?**:

Meta

**Additional documentation**:

N/A

Issue #6953, opened 2020-06-01
As a developer, I want changes I make to the front end to be visible when I deploy from Netbeans


pdurbin:
[Formerly known as _As a developer, I want changes I make to the front end to be visible when I deploy from Netbeans_]

Ever since 5f99037 developers have had to remember to comment out part of `src/main/webapp/WEB-INF/web.xml` in order for changes they are making to the front end to be visible when deploying from Netbeans. At the very least we should document this in the dev guide. This is the part that has to be commented out (but not committed, of course):

    <context-param>
        <param-name>javax.faces.FACELETS_REFRESH_PERIOD</param-name>
        <param-value>-1</param-value>
    </context-param>

Even better would be to somehow fix it so developers don't have to comment this out but I'd be satisfied with some docs for now.

<hr>

[GPD - 2020-12-16 - edited the original description -

Original said to also comment out 

    <context-param>
        <param-name>javax.faces.FACELETS_BUFFER_SIZE</param-name>
        <param-value>102400</param-value>
    </context-param>

but doing so causes flash messages to become "sticky"  - see issue #7305]

pdurbin:
Here's a workaround to avoid accidentally checking in a change to web.xml.

- Edit src/main/webapp/WEB-INF/web.xml and comment out the lines above.
- Mark the file as not changed with `git update-index --assume-unchanged src/main/webapp/WEB-INF/web.xml`
- Confirm that git doesn't think the file has been edited with `git status` and `git ls-files -v | grep "^[a-z]"` (which lists files that have been marked as "assume unchanged")

If you actually need to make an edit to web.xml, you must put it back to normal with this:

- `git update-index --no-assume-unchanged src/main/webapp/WEB-INF/web.xml`

References:

- https://stackoverflow.com/questions/13442130/git-temporarily-ignore-trivial-changes-to-files/26549583#26549583
- https://stackoverflow.com/questions/2363197/can-i-get-a-list-of-files-marked-assume-unchanged/37083903#37083903

poikilotherm:
As @pdurbin already mentioned in #7457, using MicroProfile Config API would be an easy way to completely mitigate this problem without recompiling and just changing a config setting.

When MicroProfile Config API 2.0 gets released, it's even easier by switching profiles and providing different setting inside those accordingly. See https://github.com/eclipse/microprofile-config/issues/418

Problem: this is not yet supported with Payara, `<context-param>` does not support variable substitution from `web.xml`.
[...]

PR #7463, opened 2020-12-08
7457 introduce mpconfig

### LINKED ISSUE #7457: Introduce using MicroProfile Config API
This is a story for epic #7000
This could be done as part of #7424, but as we prefer more narrow scopes, don't.

----

As you might have noticed, #7422 is about to introduce the first pieces to bring us the joy of MicroProfile Config API without calling ourselves a microservice.

This issue is about paving the ground to enable #7424 profit from it and offer full configuration flexibility.

We will introduce:
- Maven dependencies for using the API
- A test dependency for running unit and integration tests with an API implementation but without deploying to a full fledged appserver
- Allow moving and renaming of config, but remain backward-compatible:
   - A config source to enable aliases, as we are likely to rename a few "JVM properties"
   - A config source to allow moving some database settings to MicroProfile Config
   - Scenario: make the old settings deprecated and log their use, giving hints about migration

poikilotherm:
This PR paves the ground to finally enable using MicroProfile Config API within the application.
By providing two special config sources, we can rename settings but still stay backward compatible.
We can also start to migrate database settings to MPCONFIG where we see fit and still be backward compatible.
It will log using a deprecated setting, so we can stay with a setting for a while before removing it.

Closes #7457

None.

Actually this is not introducing a change, as we don't have aliases yet nor are we migrating a database setting here. This will be done for #7424 (PR forthcoming).

For now, you can take a look at the added unit tests and running a testcontainers test via `mvn -P tc verify`.

Nope.

:battery: included.

**Additional documentation**:
None so far.


doc/sphinx-guides/source/developers/configuration.rst
```diff
@@ -0,0 +1,12 @@
+Consuming Configuration
+=======================
+
+By using `MicroProfile Config <https://github.com/eclipse/microprofile-config>`_, we start to move away from
+direct ``System.getProperty()`` calls, replacing it with either injected or programmatic config retrieval.
+
+The API of MPCONFIG allows for great flexibility, type-safety, defaults etc. For system administrators
+and developers it get's much easier to provide their configuration.
```
pdurbin, line 8:
```suggestion
and developers it gets much easier to provide their configuration.
```

pom.xml
```diff
@@ -844,6 +861,7 @@
             <id>tc</id>
             <properties>
                 <skipUnitTests>true</skipUnitTests>
+                <postgresql.server.version>9.6</postgresql.server.version>
```
pdurbin, line 864: Should this postgres server version be moved up to be near the client version (around line 30)?

pdurbin (commented):
I saw this was in review so I took a quick look. I didn't run the suggested `verify` command. It sounds like more docs are coming, which is good.

I think I'm still a little confused about the "why". There's nice work here but what does it mean for people who run or develop Dataverse? What problems does it solve? I'm sure in time, all will be revealed. 😄

pom.xml
poikilotherm: Preferably, yes. But I didn't want to bother others, so I left it in the profile. Would you prefer it in the upper place, too?

pdurbin (changes requested):
Can we get this to pass on Jenkins?

I'm seeing this error:

`[ERROR] Errors: `
`[ERROR]   AliasConfigSourceTest.getValue:25 » IllegalState javax.naming.NoInitialContext...`
`[ERROR]   AliasConfigSourceTest.readImportTestAliasesFromFile:39 » IllegalState javax.na...`

over at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-7463/4/console

Also, while I'm in favor of small chunks generally, I'm eager to see how a fix for #7424 makes use of this groundwork, to make it more real. If it's possible to get a preview of that future pull request, it would be nice.

pom.xml
pdurbin: Hmm. Well, I don't want it in two places, if that's what you mean. But sure, if it can be in a single place, I'd prefer the single place to be at the top of the file with the other properties. (I just tried this locally and it seemed to work.). With the settings side by side, it's a little confusing what's what:

        <postgresql.server.version>9.6</postgresql.server.version>
        <postgresql.version>42.2.18</postgresql.version>

Maybe at some point we should rename the "42" one to something more specific.

poikilotherm: OK if it confuses people, maybe we should leave it in the profile then. It's only used with testcontainers, so it would make sense to leave it there.

pdurbin: That's fine.

pdurbin (approved):
This pull request doesn't seem to do any harm and the API test suite passes so I'm moving it to QA.

That said, other developers are quite welcome to take a look.

My overall take on this is that @poikilotherm is trying to move us toward modern technologies like the MicroProfile Config API which will ultimately make sysadmins happier with their options for configuring the application.

I did note the following from @poikilotherm in IRC but again, since the API test suite passed, deployment must have succeeded so I think we're all set.

> Doh! I found the troubles with 7457 not deploying. It's just the same thing that broke tests, too. OK will need to transform the source into a bean and inject the datasource...

poikilotherm:
Sorry @pdurbin dragging this back to dev. I need to work around the dependency injection obstacles first. Will push soon.

poikilotherm:
@pdurbin this is ready. I refactored it to use dependency injection plus adding monitoring, so we can catch potential race conditions. The docs have been extended once more, release note added. Would be perfect to get this into 5.3 (#7439 / #7446 )

poikilotherm:
@donsizemore @pdurbin Jenkins fails with non-reachable SSH. I'm sure this isn't related to my changes. What now?

pdurbin:
@poikilotherm I just clicked "build now" on Jenkins.

Meanwhile, I'll check out the further tweaks to the docs. Do you have any idea why they won't build? (And can you fix it?) Check this out:

`Warning, treated as error:`
`/Users/pdurbin/github/iqss/dataverse/doc/sphinx-guides/source/developers/configuration.rst:96: WARNING: Definition list ends without a blank line; unexpected unindent.`

poikilotherm:
> Meanwhile, I'll check out the further tweaks to the docs. Do you have any idea why they won't build?

Sorry @pdurbin this slipped through. Was a bit early this morning. Whitespace removed, Sphinx happy.

pdurbin (approved):
I'm going to move this to QA but again, I'd be happy for another developer to take a look.

Question for @poikilotherm ... will all this fancy config stuff helps us solve #6953? 😄 I hope so!

PR #7465, opened 2020-12-08
7130-handle missing contact name

### LINKED ISSUE #7130: Bagit PostPublication Workflow: Some datasets generate invalid bags
While experimenting with the internal archiver workflow and using the [dataverse-sample-data ](https://github.com/IQSS/dataverse-sample-data) script. I noticed only 10 out of 21 datasets are archived and out of those 10 bags 4 are invalid and won't unzip. 

Screen Shot 2020-07-24 at 12 19 15

These are the titles of the datasets from the sample data that generate invalid bags:
1. Cause of Death Data
2. Dataset with Shapefiles, Images, and Tabular Files
3. “Changes in test-taking patterns over time” concerning the Flynn Effect in Estonia
4. Data from the BAFACALO project: The Brazilian Intelligence Battery based on two state-of-the-art models: Carroll's model and the CHC model.

My internal archiver workflow config is as follows:
Screen Shot 2020-07-24 at 16 36 41

This appears to be a bug in the bag generator.

qqmyers:
**What this PR does / why we need it**: This PR handles the case where a contact has an email but not a name which previously resulted in a null pointer exception.

Closes #7130 

**Special notes for your reviewer**: discussion in issue. Note this branch was merged once to close 7130 - reusing the issue and branch to handle another point that has come up.

**Suggestions on how to test this**: create a dataset with a contact that only has an email. Publish. Trigger bag creation (probably easiest to configure the file archiver and then use the API to request bag creation - per guides). Bag should be created. (@mderuijter may also be able to test?)

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
@@ -756,8 +756,10 @@ public class BagGenerator {
                         info.append(CRLF);
 
                     } else {
-                        info.append(((JsonObject) person).get(contactNameTerm.getLabel()).getAsString());
-                        info.append(CRLF);
+                        if(contactNameTerm != null) {
+                          info.append(((JsonObject) person).get(contactNameTerm.getLabel()).getAsString());
+                          info.append(CRLF);
+                        }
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I didn't test this fix but it makes sense. A null check was added to two places.

PR #7468, opened 2020-12-09
send emails from onSuccess

### LINKED ISSUE #7467: Dataverse publication success email sends obsolete info (from prior version) such as title
Currently (v5.2, demo,dataverse,org), the email notification created when publication succeeds includes the title of prior published version and is thus incorrect/out of date if the title has been changed. The happens because the email notification code called from FinalizeDatasetPublicationCommand only gets the id of the DatasetVersion being published and then looks that id up in the database. Because the publication transaction hasn't completed and because the email code looks for version.getDataset().getDisplayName(), which gets the name for the latest released version, it is picking up the prior version's tittle.

Currently, the emails only include the dataset name, owning Dataverse and persistentID and I don't think the latter two should change (at least for stores reserving PIDs at create time, the PID can no longer change at publication). However, QDR, for example, has modified the emails to also send the recommended citation and the same effect can be seen there (title as well as the version being "DRAFT VERSION" in the citation. Changes to the creator list would also show in the citation and be affected.)

A straight forward fix appears to be to just move the success notification call to the onSuccess method, i.e. to after the publication transaction has completed. I'll submit a PR for that.

I have not looked to see if there are other examples of this. Just thinking quickly, I assume any publication failure messages have the same underlying issue, but, since the transaction will abort and leave the latest version as draft, the email should still be 'correct'. Similarly, if there are notifications in other commands, they may have the same issue (haven't checked), but if they aren't reporting info that changes in the transaction, there may not be a noticeable result. (It might still be worth thinking about moving success messaging into onSuccess generally to avoid any possibilities should the message contents change in the future.)

qqmyers:
**What this PR does / why we need it**: Makes sure the email notification sent when publication of a dataset succeeds include changes from the newly published version including any change to its title.

Closes #7467 

**Special notes for your reviewer**: This is related to #7467, but I'm not sure if it should close the issue since there could be similar problems for other messages (Dataverse publication?)

**Suggestions on how to test this**: Publish a dataset, edit the title, and publish again. Check the email and verify it has your updated title.

**Additional documentation**:


pdurbin (approved):
This fix looks straightforward enough. I didn't test it but the sending of the notification was moved to onSuccess. This should only affect the publishing of datasets.

I'm approving this and moving it to QA but I'm going to tag @djbrooke for any housekeeping of #7467 which is pull request references but does not close. Perhaps we should have an issue with narrower scope that this pull request does close? It seems to be fixing a real bug.

pdurbin:
Whoops, this branch is still at 5.2. I'll attempt to merge develop into it before putting it in QA.

djbrooke:
Thanks @pdurbin for the ping. I'd be OK closing #7467 for this and I'll update the description above to reflect. 

I think the last part of #7467 that Jim added is important to live on as a new, separate issue:

"It might still be worth thinking about moving success messaging into onSuccess generally to avoid any possibilities should the message contents change in the future."

@scolapasta any thoughts?

scolapasta:
@djbrooke agreed, this is the kind of thing onSuccess was designed for. One thing we'll need to consider, though, is API calls. I think (but am not certain) that many of these success messages are UI only, because we were worried about creating hundred or thousands of notifications when running through the API (and it was an open question on whether we should or should not). Moving to the onSuccess would have them always generated - if we don't want this, we may also need a mechanism to determine in the Command if it was called via API.

djbrooke:
@scolapasta thanks -- sounds good. Does the concern about API calls generating a bunch of notifications come into play on this PR?

pdurbin:
The results are in from https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-7468/8/testReport/

All the tests passed and the API tests took the appropriate amount on time.

I'm moving this to QA.

@djbrooke thanks for the issue cleanup. In this case the notification is not specific to the UI. It's in a command that both the UI and API use. Also, publishing is a relatively rare event. I'm not sure when people get flooded with notifications but I bet Sonia could tell us. There might even be an open issue about this.

scolapasta:
@pdurbin @djbrooke I could imagine if you create 1000 datasets and publish (all via API) you may not want 1000 notifications. (which may just be something you would do when testing or migrating, but that recent law dataverse comes to mind). Regardless in this case, @pdurbin is correct that is was already in the Command and just moved methods.

We may want to revisit the logic / come up with generalized rules, but there's no reason not to move this one currently forward.

PR #7478, opened 2020-12-14
handle getDisplayNameLang() case with no Bundle entry for a tool

### LINKED ISSUE #7477: Handle externaltools with toolName when there are no i18n entries
Currently, if an external tool has a toolName defined, the Dataverse code looks for i18n entries in the Bundle.properties to find a display name in the desired language. If no such entries exist, the name returned for display is 'null'. Nominally one can avoid this by not adding a toolName for tools until Bundle entries are added, it will make it simpler if we can update the curl example scripts in the GDCC dataverse-previewers repo to include a toolName by default without requiring i18n entries.

I'll submit a small PR that just checks for an i18n entry, and if it doesn't exist, defaults back to the tools default displayName.

qqmyers:
**What this PR does / why we need it**: Avoids a 'null' displayName if external previewers/explorers are defined with toolNames without someone also having added i18n entries for that tool in the Bundle.properties files

Closes #7477

**Special notes for your reviewer**: The mechanism to provide an i18n display name based on the toolName already exists - this PR is just avoiding a null for a case where the toolName is defined (e.g. so installs can use one set of example curl commands for registering previewers regardless of whether they set up 18n or not)

**Suggestions on how to test this**:  Register a previewer using a curl command from the README.md in the develop branch of the github.com/GlobaldDataverseCommunityConsortium/dataverse-previewers repo and verify that it's displayName in, for example, a list of available previewers is correct/not 'null'. I think you'll have to register more than one tool for a given mimetype for that. Alternately, I think if you configure a tool to be both preview, and explore types, you'll see the displayName on the Explore button.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
+++ b/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalTool.java
@@ -1,6 +1,7 @@
 package edu.harvard.iq.dataverse.externaltools;
 
 import edu.harvard.iq.dataverse.util.BundleUtil;
+import edu.harvard.iq.dataverse.util.StringUtil;
 
 import java.io.Serializable;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


pdurbin (commented):
This change looks fine to me but I'm happy give @lubitchv time to take a look before we move it to QA. I suppose the only danger is that a tool name could be in the wrong language but that's probably better than "null".

lubitchv:
Thanks Jim, it does make sense, and I missed that part. The change looks good to me, the only thing, the same change should be probably also added for getDescriptionLang(). It will not mater for explorer types of tools, such as previewers, but it will mater for configure tools such as DCT.

qqmyers:
Thanks. I updated the getDescriptionLang() method as well and merged from the develop branch to bring everything up to date.

mheppler (approved):
👍 Logic looks sound.

lubitchv (approved):

PR #7479, opened 2020-12-15
7291 spam check

### LINKED ISSUE #7291: Check during Dataset Publish and Dataverse Create for Spam Content
In #7238 I mentioned that we were seeing spam in Harvard Dataverse (dataverses and datasets) and I wanted to investigate putting in some captchas. It appears that the spammers aren't using automated means to create spam content but are instead doing it manually. This leads me to believe that a captcha or math problem solution would not be as effective as I'd like. :/ So, some new thoughts:

- We'd like to catch the spam before it's published instead of letting it be published and cleaning it up (@mheppler pointed out that letting users save something that couldn't be published doesn't really make sense, so if we can add a check on save that would make good sense)
- We have a workflow for locking datasets during publish for file integrity checks and PID registration, so if we wanted to run some publish-time check for spam we could incorporate it there and include it in the locking messaging.
- Our friends at Zenodo were nice enough to share a link to the repo that they use to identify spam at https://github.com/zenodo/zenodo-classifier. They are willing to accept contributions, and it may be good to collaborate on this because someone spamming Dataverse today is probably spamming Zenodo tomorrow, and vice versa. 
- How would we handle positives? Delete/Destroy? Keep them locked until a human can intervene? Move them to a holding Dataverse and delete after 7 days? Other? 🤔 

cc: @mheppler @landreev

landreev:

An anti-spam mod that prevents sus datasets and dataverses from being published. See the issue for more info. The plan is to throw this on our prod. servers shortly in order to gauge the real life usefulness of the solution. We are likely going to revise it before adding this to an official release. 

(but this brings up a question, do we even want to merge this into develop at all?)

Closes #7291 

Please keep in mind that in its current form this is intended for local use only. Hence no documentation in the official guide etc. 

Let's talk once it goes to QA. 

**Additional documentation**:


scripts/spam/spamassassin/README.md
```diff
@@ -0,0 +1,49 @@
...
+Configure custom rejection messages and/or the admin override, as needed. For example:
+
+`curl -X PUT -d 'Uhm, this looks like spam. But nice try...' http://localhost:8080/api/admin/settings/:DataverseMetadataValidationFailureMsg`
+
+(note that the "If you think this is in error, contact support at ..." will be automatically added to the custom message by the dataverse or dataset page).
+
+
+`curl -X PUT -d true http://localhost:8080/api/admin/settings/:ExternalValidationAdminOverride`
+
+When the override is enabled, the external check will be skipped for admin users.
```
scolapasta, line 38: should say superusers instead of admins.

scripts/spam/spamassassin/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The actual spam checking magic is defined in the rules file (dataverse_spam_prefs.cf). It can be tinkered with/fine-tuned by specifying what tests to perform and/or the penalities assigned for every detected violation. See the SpamAssassin documentation for how it all works. 
+
+
+
+
+DataverseMetadataValidationFailureMsg,
```
scolapasta, line 45: not sure if this text is in the right place.

src/test/java/edu/harvard/iq/dataverse/export/OpenAireExporterTest.java
```diff
@@ -16,6 +16,7 @@
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import static junit.framework.Assert.assertEquals;
+import org.junit.Ignore;
```
scolapasta, line 19: was this meant to be checked in?

scripts/spam/spamassassin/README.md
scolapasta: Actually, better question - do we need this setting? i.e. what's the use case for not just always allowing superusers to publish?

scolapasta (commented):
Looks good! I added a few minor comments (docs, and a test ignore).

The biggest question is the one about superusers? It seems to me that that aetting may be overkill and not needed.

I'm only "commenting" and not approving, because we don't want this to get merged just yet.

pdurbin:
Closing. This has been superseded by pull request #8155.

PR #7488, opened 2020-12-20
Add Wget to Vagrant Instructions

### LINKED ISSUE #7483: Vagrant based quick setup fails because of missing "wget"
The Vagrant based quick setup (as described in https://guides.dataverse.org/en/latest/developers/tools.html) fails because of a missing "wget" in box 'bento/centos-8.2':
```
default: Installing jq for the setup scripts
default: /tmp/vagrant-shell: line 14: wget: command not found
default: chmod: cannot access 'jq': No such file or directory
default: mv: cannot stat 'jq': No such file or directory
```
Solution: Add `dnf install -y wget` at the top of `./dataverse/scripts/vagrant/setup.sh`.

hkramski:
Install missing wget
Closes #7483
Fix and extend final setup message (port number on localhost was wrong)

### DIFF
--- a/scripts/vagrant/install-dataverse.sh
+++ b/scripts/vagrant/install-dataverse.sh
@@ -26,4 +26,6 @@ if [ -e tmp-${pid}-default.config ]; then # if we moved it out, move it back
 	mv -f tmp-${pid}-default.config default.config
 fi
 
-echo "If "vagrant up" was successful (check output above) Dataverse is running on port 8080 of the Linux machine running within Vagrant, but this port has been forwarded to port 8888 of the computer you ran "vagrant up" on. For this reason you should go to http://localhost:8888 to see the Dataverse app running."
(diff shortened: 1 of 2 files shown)


djbrooke:
Thanks @hkramski for the PR! We'll take a look and let you know if we have any questions.

pdurbin (approved):
@hkramski thanks for the pull request!

I think it adds value as-is so I'm approving it for QA to take a look.

A couple things:

- We mention Dataverse being on port 8888 in Vagrant over at https://guides.dataverse.org/en/5.3/developers/tools.html#vagrant and as far as I know it still works. I understand that technically, 8080 is being forwarded to 8088, which is why you made the change. If you're saying it doesn't work on 8888 at all, we should probably change that "Tools" page too.
- We do have some init scripts for starting Payara in the Installation Guide. Perhaps someday we should use them within Vagrant but for now I'm glad to see that you're pointing out that Payara won't start on its own.

PR #7491, opened 2020-12-21
Support proxies with S3 direct up/download #7490

### LINKED ISSUE #7490: Support proxies with S3 direct up/download
The use of proxies with S3 (e.g. as a firewall) will work when direct-upload and download are not configured in Dataverse. However, because these require pre-signed URLs (which includes signing the protocol/host/port in the URL), an additional setting (and support in the software) is required to allow direct up/download to work. In essence, Dataverse needs to sign the URL using the S3 URL but send the URL path/parameters to the proxy instead (so that when the proxy forwards these to the S3 server, the signature validates).

The capability to handle this was developed to support testing direct-uploads at TDL, but, until now, hadn't been extended to handle direct downloads or documented as an option in the guides. As it looks like TDL will now be using a proxy configuration for some store(s) in production, those have now been added and the capability is being offered as a PR.

qqmyers:
**What this PR does / why we need it**: Adds the ability to use a proxy with S3 stores and support direct upload/download.

Closes #7490

**Special notes for your reviewer**: The proxy option was already in the code but only used for upload URLs. The actual change in this PR is pretty minor - just adding code to replace the endpoint url with the proxy before sending (if it is set) in the generateTemporaryS3Url call (for download URLs) and moving the code to read the endpoint and proxy into the constructor. The rest is just formatting to get rid of the mixed space/tabs, so it is probably clearer to view ignoring the formatting. (i.e. https://github.com/IQSS/dataverse/pull/7491/files?w=1).

**Suggestions on how to test this**: A manual test could be done without configuring a real proxy: Set the custom endpoint to point to an existing S3 store - e.g. the Amazon AWS S3 URL for a test instance - and a proxy URL. Then, attempt a download and look at the generated URL that the browser is redirected to (i.e. in the browser dev console/network tab). Copying that URL and manually replacing the proxy setting with the real URL for S3 should allow the download to work (i.e. the key issue is that the signature on the URL validates.) 

**Is there a release notes update needed for this change?**: Could advertise support for proxies, but probably not necessary.

**Additional documentation**:


landreev (approved):

kcondon:
This passes basic functional smoke test but was not able to test using the scenario above since our aws config doesn't use custom-endpoint-url, that is for non aws s3. Turns out that param is required. So, this is working in prod at texas but should this support an dataverse aws installation without custom endpoint specified, as would normally be configured and use endpoint for other implementations? Discussing with Jim and Leonid but putting back into review until sorted out.

landreev (approved):

PR #7503, opened 2021-01-07
support download of UTF-8 filenames #7188

### LINKED ISSUE #7188: UTF-8 characters in filenames not handled correctly in normal downloads
In looking at #7184, I realized that that problem had been introduced when the S3 direct download code had been updated to correctly encode UTF-8 chars (which caused an error in direct downloads). That code was never applied to the normal download, or related downloads of metadata files, etc. where the filename is used and I confirmed that filenames such as "1976–2016.txt" - with a long dash (UTF xE2 x80 x93 ) are downloaded as "1976 2016.txt" with normal download. "1976–2016.txt" was the example from #4524. 

I didn't test any other UTF-8 characters, or other endpoints but I don't see any UTF-8 encoding outside S3AccessIO. I would expect that the same changes to the content disposition header that were made in S3AccessIO would work elsewhere, e.g. 
instances of 
`"attachment; filename=\"" + fileName + "\""` 
--> 
`"attachment; filename*=UTF-8''" + URLEncoder.encode(this.getDataFile().getDisplayName(), "UTF-8").replaceAll("\\+", "%20")`

pdurbin:

If you upload a file to Dataverse with a filename that has UTF-8 characters, they get munged on the download. For example, an en-dash ("-") is replaced by a space. "READ–ME.md" become "READ ME.md".

Closes #7188

None.

- Upload a file with UTF-8 characters such as READ–ME.md and Save Changes.
- Download the file and make sure the name is the same.
- Try files with spaces in them too.

Emojis seem to work too! 🎉 🎉 🎉 

Screen Shot 2021-01-07 at 4 06 00 PM

No.

I didn't add one but I'd be happy to.

**Additional documentation**:

None. My guess is that people assume this already works. I don't think we declare support for UTF-8 filenames anywhere. To be honest, I'm surprised it works as well as it does but I guess it is 2021. We're living in the future. 😎

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
@@ -302,9 +302,11 @@ public class DownloadInstanceWriter implements MessageBodyWriter<DownloadInstanc
                     
                     // Provide both the "Content-disposition" and "Content-Type" headers,
                     // to satisfy the widest selection of browsers out there. 
-                    
-                    httpHeaders.add("Content-disposition", "attachment; filename=\"" + URLEncoder.encode(fileName, "UTF-8") + "\"");
-                    httpHeaders.add("Content-Type", mimeType + "; name=\"" + URLEncoder.encode(fileName, "UTF-8") + "\"");
+                    // Encode the filename as UTF-8, then deal with spaces. "encode" changes
+                    // a space to + so we change it back to a space (%20).
+                    String finalFileName = URLEncoder.encode(fileName, "UTF-8").replaceAll("\\+", "%20");
+                    httpHeaders.add("Content-disposition", "attachment; filename=\"" + finalFileName + "\"");
+                    httpHeaders.add("Content-Type", mimeType + "; name=\"" + finalFileName + "\"");
                     
                     long contentSize; 
                     boolean useChunkedTransfer = false; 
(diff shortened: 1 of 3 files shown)


mheppler:
👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -302,8 +303,8 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
                     // Provide both the "Content-disposition" and "Content-Type" headers,
                     // to satisfy the widest selection of browsers out there. 
                     
-                    httpHeaders.add("Content-disposition", "attachment; filename=\"" + fileName + "\"");
-                    httpHeaders.add("Content-Type", mimeType + "; name=\"" + fileName + "\"");
+                    httpHeaders.add("Content-disposition", "attachment; filename=\"" + URLEncoder.encode(fileName, "UTF-8") + "\"");
```
qqmyers, line 306: FWIW: comments in https://github.com/IQSS/dataverse/pull/4542/files  re S3 downloads suggests that Firefox doesn't/didn't like this form and needed filename* and an explicit UTF-8. Has that been checked? 

Also - the S3 version includes a fix for spaces being encoded as '+' signs by the addition of the URLEncoder.encode method. That would be a change to this same line (and the next one? I'm not sure that S3 has the equivalent (adding a name to the Content-Type header), so I'm not sure.)

pdurbin: I've been testing with Firefox the whole time so I'm not worried about that browser.

Good catch on how the encoding I added changes spaces to plus signs. I just pushed 702b505 to make sure they stay as spaces on download.

qqmyers (approved):
Looks good.

PR #7504, opened 2021-01-08
IQSS/6497 migrate dataset api

### LINKED ISSUE #6497: Semantic Metadata API calls

qqmyers:
**What this PR does / why we need it**: This PR supports migration of datasets from other systems, keeping their PID, version,  and publicationdate, and allowing file upload via the existing API. It leverages leveraging the json-ld metadata api from #7414.
(Note that #7414 in turn requires resolution of updating our Java version or having a work-around for Java 8 added there - see that PR for details).

Closes #6497 

**Special notes for your reviewer**: This is similar to the :import API but is different (hopefully usefully) in a few ways: 
* that API call requires files to be prepositioned. The api calls here allow you to create a dataset to be migrated, then add files/filemetadata using the existingAPIs (inlcuding direct upload or using the HttpOverlay store in #7325), and then conclude by marking the dataset as published.
* this call uses the json-ld metadata format from the ORE export, Bag archiving, and #7414 which is flatter and supports incremental metadata updates (one doesn't have to download the whole metadata tree, make an edit and reupload everything),
* the last point makes it highly compatible with re-importing an archived Bag which includes the same ORE/json-ld formatted metadata
* since dataset creation and version update/publish capabilities are separated into two calls, it may be simpler to support migration of multiple versions via this API (it may work already - I still need to test that capability).

FYI: The last step is slightly different than the normal publish step since version and publication dates are preserved. The PR does update the PID/DOI (e.g. the landing URL specifically) assuming it has the authority/shoulder configured in Dataverse.

**Suggestions on how to test this**: The DVUploader uses this API to re-import Bags. Nominally running an upload on Bags of the sample datasets would test this fairly well (current versions at https://drive.google.com/drive/folders/1YXAmJLziFs7XhsvRfqfZ6o9AvLXywyBT?usp=sharing , note that work to improve the Bag/ORE contents is ongoing, so these may get out of date w.r.t. code, so check dates/remind me if an update looks like its needed). 

Assuming #7414 is tested separately, the testing here could/should focus on just the added migrate calls, so looking for datasets with a broad range of metadata probably isn't necessary (since testing #7414 covers that.) and the things to look at in a migrated dataset are the DOI, creation/publication dates, whether the DOI is updated at DataCite, etc.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Not directly - there are questions in the community (remote store/registry metadata working group topic, DANS migration, etc.) as to how one should indicate a migrated dataset, what metadata about the source should be captured, etc. This PR doesn't add any metadata or make an display changes, but this PR may increase interest in resolving these other questions.

Probably worthwhile to note it exists. As a v~1 implementation, I'd phrase this as an ~experimental API (I don't think we've decided exactly how to phrase this but there have been other recent API adds in this category).
**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -609,6 +614,97 @@ public Response updateDraftVersion( String jsonBody, @PathParam("id") String id,
             
         }
     }
+  
+    @GET
+    @Path("{id}/versions/{versionId}/metadata")
```
landreev, line 619: Was there supposed to be an extra entry in the path for this API call - something like like `"{id}/versions/{versionId}/metadata/jsonld"`?

landreev (commented):
I think I answered all the implementation questions I had by reading the code. Except the one about the "/api/datasets/{id}/versions/metadata" that appears to be duplicated now. 
Also, should this be mentioned in the api guide as well? Plus a two-line release note; just as you said, to mention that it exists.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
landreev: (or is the idea to have the same `.../metadata`, and to have the client request jsonld by supplying an `Accept:` header for it? 
(that should be legal too, right?...)

qqmyers: I was going for the latter - line 620 means this method picks up the Accept: application/json-ld and the other gets anything else.

qqmyers:
Thanks @landreev - I'll go ahead and add documentation. Note that by looking through all of this, you also reviewed #7414 which is a subset.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
landreev: This may be a cleaner way  of doing it too, not having to declare the format an extra parameter. I guess the only disadvantage is that you can't bookmark a standalone URL for a specific content type (?). But I'm assuming it's ok for this specific format.

djbrooke:
Thx @landreev and @qqmyers. I'll move this back to community dev so you can make the change and get this to QA and into a release. 

@qqmyers, let me know if you'd like me to add the release note and I'll do so.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: Yeah - I think it is ~common. FWIW: I do this in the newmetrics PR (#7178) as well to let you select csv or json and I set the csv as the default so if you bookmark/request via the browser, you get that. The default here would still be the json (not json-ld).

qqmyers:
@djbrooke - added docs.  For the release notes, you could say various things - the basic idea is that this PR adds an experimental  semantic API that allows getting/setting/deleting Dataset metadata (and creating a new Dataset) in the same format as used by the OAI-ORE metadata export, and adds experimental API calls to support migrating a Dataset with its existing PID and publication Date. The work was supported by RDA and is leveraged to enable Dataverse to import Datasets stored as RDA-conformant Bags though tools such as the DVUploader.

djbrooke:
@qqmyers thanks, will add notes ASAP.

djbrooke:
@qqmyers I added release notes, take a look and let me know what you think - we can discuss or you can edit directly.

qqmyers:
@djbrooke  - Looks good.

landreev:
@djbrooke @qqmyers The docs look great.

djbrooke:
Thx @qqmyers @landreev anything remaining to review?

pdurbin:
I saw one typo in the docs. I haven't had a chance to read them.

More importantly, is the idea that if you import a foreign DOI it isn't actively managed by Dataverse? It's like a harvested dataset? Or if it's managed, what happens when you edit it in terms of DataCite (since you can't update the record there)?

Finally, I assume the docs are in the dev guide because the API is experimental.

djbrooke:
@pdurbin any hints on what the typo is, so that we can fix it?

qqmyers:
Yes - dev guide because 'experimental'. 

When you import via this API, the DOI has to match the authority/shoulder configured for the Dataverse instance. The DOI is then managed by Dataverse sw, with the metadata and landing URL updated at DataCite when you do the final publish step. (Since DataCite can transfer management of individual DOIs, future work could potentially test that up front (versus checking for the authority/shoulder match). An alternate/additional future addition to DV sw might be to support managing multiple DOI authority/shoulder combos, making it easier to migrate existing data into an instance that was already using a different auth/shoulder combo. (Initial discussion with DataCite indicated that this might be the better option for lots of DOIs.))

src/test/java/edu/harvard/iq/dataverse/dataaccess/StorageIOTest.java
```diff
@@ -65,7 +65,9 @@ public void testGetDvObject() {
         Dataset d= new Dataset();
```
pdurbin, line 65: It would be great to have more tests in this pull request. It's lots of new code.

doc/release-notes/6497-migrate-api.md
```diff
@@ -0,0 +1,7 @@
+# Release Highlights
+
+### Dataset Migration API (Experimental)
+
+Datasets can now imported/updated following the format of an OAI-ORE export (RDA-conformant Bags), allowing for not only easier migration from one Dataverse installation to another, but also for better support of import from other systems. This experimental endpoint also allows keeping the existing persistent identifier and publication dates.
```
pdurbin, line 5: The DOI authority has to match, right?

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
+Dataset Migration API
+=====================
+
+The Dataverse software includes several ways to add Datasets originally created elsewhere (not to mention Harvesting capabilities). These include the Sword API (see the :doc:`/api/sword` guide) and the /dataverses/{id}/datasets/:import methods (json and ddi) (see the :doc:`/api/native-api` guide).
```
pdurbin, line 4: We should mention that the DOI authority has to match. (That's my understanding anyway.)

pdurbin (commented):
This is quite a lot of new code (great!) with zero tests (boo!). I left some doc suggestions, especially to make sure people know the DOI authority has to match. Also, it looks like a Solr schema update is in store, so we should mention this in the release notes.

doc/release-notes/6497-migrate-api.md
```diff
@@ -0,0 +1,7 @@
+# Release Highlights
+
+### Dataset Migration API (Experimental)
```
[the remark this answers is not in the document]
djbrooke: Good call, will add

pdurbin:
@djbrooke "sfotware" ... noted in the review I just left.

@qqmyers that makes a lot more sense. Thanks. In my review I suggested talking about this is both the release notes and the guides. It sounds like the new migrate will work a lot like the old migrate. The big change is JSON-LD support for creating/updating datasets. Whoops, I didn't mention this in my review but @djbrooke we should probably emphasize this in the release note (creating/updating in addition to migrating).

Is this JSON-LD based on a standard? The BagIt standard, I guess? Can we link to standard from our guides? Is there a JSON Schema?

qqmyers:
re: Is this JSON-LD based on a standard? - json-ld is 'just' the format (and is a ~standard way to represent RDF). What's allowed in the json-ld, in terms of what terms are allowed, is defined by Dataverse internally and the metadatablocks. Metadata blocks can define a base URL or per-term URLs, which is what is sent in the OAI-ORE metadata export and what this API allows coming in. For metadata that isn't in a block, I made choices when creating the OAI-ORE mapping those to things like Dublin Core or Schema.org (when obvious or when the schema.org was already used in the page header metadata) or created a term in the 'dvcore' namespace. Also FYI: The api call to get a metadatablock term now includes the URI to use in the json-ld.

doc/sphinx-guides/source/developers/dataset-migration-api.rst
qqmyers: in line 9

doc/release-notes/6497-migrate-api.md
djbrooke: will add

djbrooke: added this in https://github.com/IQSS/dataverse/pull/7504/commits/45e585334d37be8a7204e7984fa7173412db3689

doc/release-notes/6497-migrate-api.md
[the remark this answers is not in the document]
djbrooke: added in https://github.com/IQSS/dataverse/pull/7504/commits/45e585334d37be8a7204e7984fa7173412db3689

copied from 5.1 notes

doc/sphinx-guides/source/developers/dataset-semantic-metadata-api.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export DATASET_ID='12345'
+  export DATASET_PID='doi:10.5072/FK2A1B2C3'
+  export VERSION='1.0'
+  export SERVER_URL=https://demo.dataverse.org
+ 
+  Example: Delete the TermsOfUseAndAcess 'restrictions' value 'No restrictions' a for the latest version using the DATASET PID
```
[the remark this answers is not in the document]
djbrooke: I went to address this https://github.com/IQSS/dataverse/pull/7504/commits/45e585334d37be8a7204e7984fa7173412db3689 but it was already updated. :)

landreev:
Looking at the Jenkins failure above, it appears to be a false alarm... Unless I'm missing something.

doc/release-notes/6497-migrate-api.md
```diff
@@ -0,0 +1,17 @@
...
+
+Datasets can now imported/updated following the format of an OAI-ORE export (RDA-conformant Bags), allowing for not only easier migration from one Dataverse installation to another, but also for better support of import from other systems. This experimental endpoint also allows keeping the existing persistent identifier (where the authority and shoulder match those for which the software is configured) and publication dates. This endpoint also allows for the update of terms metadata (#5899).
+
+This development was supported by the [Research Data Alliance](https://rd-alliance.org) and follows the recommendations from the [Research Data Repository Interoperability Working Group](http://dx.doi.org/10.15497/RDA00025).
+
+### Additional Upgrade Steps
+
+Update Solr Schema
+
+- copy schema_dv_mdb_fields.xml and schema_dv_mdb_copies.xml to solr server, for example into /usr/local/solr/solr-7.7.2/server/solr/collection1/conf/ directory
```
pdurbin: This is out of date. We're on Solr 8 now.

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
...
+* metadata can be specified using the json-ld format used in the OAI-ORE metadata export
+* existing PIDs can be maintained (currently limited to the case where the PID can be managed by the Dataverse software, e.g. where the authority and shoulder match those the software is configured for)
+* adding files can be done via the standard APIs, including using direct-upload to S3
+* the dataset can be published keeping the original publication date
+
+This API consists of 2 calls: one to create an initial Dataset version, and one to publish the version with a specified publication date. 
+These calls can be used in concert with other API calls to add files, update metadata for additional versions, etc.   
+
+
+Start migrating a Dataset into a Dataverse Collection
```
pdurbin, line 17:
```suggestion
Start Migrating a Dataset into a Dataverse Collection
```

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
...
+
+  curl -H X-Dataverse-key:$API_TOKEN -X POST $SERVER_URL/api/dataverses/$DATAVERSE_ID/datasets/:startmigration --upload-file dataset-migrate.jsonld
+
+An example jsonld file is available at :download:`dataset-migrate.jsonld <../_static/api/dataset-migrate.jsonld>` 
+
+
+Publish a Migrated Dataset
+--------------------------
+
+The call above creates a Dataset. Once it is created, other APIs can be used to add files, add additional metadata, etc. When a version is complete, the following call can be used to publish it with its original publication date
```
pdurbin, line 39:
```suggestion
The call above creates a Dataset. Once it is created, other APIs can be used to add files, add additional metadata, etc. When a version is complete, the following call can be used to publish it with its original publication date.
```

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
...
+* the dataset can be published keeping the original publication date
+
+This API consists of 2 calls: one to create an initial Dataset version, and one to publish the version with a specified publication date. 
+These calls can be used in concert with other API calls to add files, update metadata for additional versions, etc.   
+
+
+Start migrating a Dataset into a Dataverse Collection
+-----------------------------------------------------
+
+.. note:: This action requires a Dataverse installation account with super-user permissions.
```
pdurbin, line 20:
```suggestion
.. note:: This action requires a Dataverse installation account with superuser permissions.
```

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
...
+
+An example jsonld file is available at :download:`dataset-migrate.jsonld <../_static/api/dataset-migrate.jsonld>` 
+
+
+Publish a Migrated Dataset
+--------------------------
+
+The call above creates a Dataset. Once it is created, other APIs can be used to add files, add additional metadata, etc. When a version is complete, the following call can be used to publish it with its original publication date
+
+.. note:: This action requires a Dataverse installation account with super-user permissions.
```
pdurbin, line 41:
```suggestion
.. note:: This action requires a Dataverse installation account with superuser permissions.
```

scripts/api/data/metadatablocks/migration.tsv
```diff
@@ -0,0 +1,5 @@
+#metadataBlock	name	dataverseAlias	displayName	blockURI											
+	migration		Migrated Metadata	https://dataverse.org/schema/migration/											
+#datasetField	name	title	description	watermark	 fieldType	displayOrder	displayFormat	advancedSearchField	allowControlledVocabulary	allowmultiples	facetable	displayoncreate	required	parent	metadatablock_id	termURI
+	metadataOnOrig	Metadata on the original source of migrated datasets.			textbox	1		FALSE	FALSE	FALSE	FALSE	FALSE	FALSE		migration	https://dataverse.org/schema/core#metadataOnOrig
```
pdurbin, line 4: In pull request #7414 @qqmyers and I are talking about removing this block (and code) with "metadataOnOrig".

pdurbin (changes requested):
I found a couple things.

Overall this looks good. I still think it would be nice to have more tests but I guess it's experimental. 😄

qqmyers:
FWIW: This PR should be up to date w.r.t. #7414 again. https://github.com/GlobalDataverseCommunityConsortium/dataverse/compare/IQSS/6497-semantic_api...GlobalDataverseCommunityConsortium:IQSS/6497-migrate_dataset_api?expand=1 should make it clearer what's added for this PR prior to #7414 getting merged.

doc/sphinx-guides/source/admin/dashboard.rst
```diff
@@ -22,7 +22,7 @@ This dashboard tool allows you to define sets of local datasets to make availabl
 Metadata Export
 ---------------
 
-This part of the Dashboard is simply a reminder message that metadata export happens through the Dataverse Software API. See the :doc:`metadataexport` section and the :doc:`/api/native-api` section of the API Guide for more details.
+This part of the Dashboard is simply a reminder message that metadata export happens through the Dataverse Software API. See the :doc:`/admin/metadataexport` section and the :doc:`/api/native-api` section of the API Guide for more details.
```
qqmyers, line 25: FWIW: unrelated - just found it when fixing another link

src/main/java/edu/harvard/iq/dataverse/util/json/JSONLDUtil.java
```diff
@@ -80,19 +80,14 @@ public static Dataset updateDatasetMDFromJsonLD(Dataset ds, String jsonLDBody,
...
-            //Do nothing - a null value and other invalid values will be caught in parsing
-        }
-        Optional<GlobalId> maybePid = GlobalId.parse(id);
-        if (maybePid.isPresent()) {
-            ds.setGlobalId(maybePid.get());
-        } else {
-            if(migrating) {
-              // unparsable PID passed. Terminate.
-              throw new BadRequestException("Cannot parse the @id. Make sure it is in valid form - see Dataverse Native API documentation.");
+        if (migrating) {
```
qqmyers, line 83: FWIW: Submitting an '@id' value for the dataset PID, and a modificationdate (next diff) are only allowed for migrate (not valid in the create dataset case), so these are cleaner fixes to just skip running into the null values in the create case.

qqmyers:
OK - with the semantic api PR merged, this is ready for review - just minor changes beyond the new migrate and releasemigrated methods.

scripts/api/data/metadatablocks/migration.tsv
pdurbin: Removed.

pdurbin (changes requested):
Code-wise, I think this is ready for QA but there are two things I think we should fix:

- The new integrations tests are yielding a 500 error.
- We should update the list of APIs to include this one and others that have been recently added. Otherwise they're hard to find. @qqmyers let me know if you'd like me to make this change.

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -0,0 +1,51 @@
+Dataset Migration API
```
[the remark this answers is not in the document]
pdurbin: > can you please update the "Please note that some APIs are only documented in other guides that are more suited to their audience" section and add "Developer Guide" and under that recent APIs that have been added?

@qqmyers I just opened this pull request for this: https://github.com/GlobalDataverseCommunityConsortium/dataverse/pull/13

qqmyers:
OK - @pdurbin made the doc change he asked for and I fixed the test - moving to QA (feel free to drop it back if there's something else that needs to happen first).

kcondon:
Some questions/issues encountered:
1. Update release notes, remove solr step since it is no longer necessary, per Jim
2. Check and mention super user only in doc for api 
3. In (re)create example in doc, shows PERSISTENT_IDENTIFIER in env variable but does not reference it on command line 
4. When updating sample jsonld file with a real, published PID from demo, fails saying cannot recreate dataset that had yet to be created (sample file missing publication date)

qqmyers:
@kcondon - I addressed 1-4 in the latest commit. For 5), I added #8028 and referenced that in the docs.

PR #7506, opened 2021-01-12
remove worldmap/geoconnect #7409

### LINKED ISSUE #7409: Remove Worldmap/Geoconnect from Application
Following Harvard Worldmap's migration to new infrastructure, we will no longer be able to apply resources to reconfigure and support Geoconnect, which powers the integration. :/ Since we're actively investigating other geospatial exploration and analysis options in the community and with Harvard Library, I'm going to suggest pulling out Worldmap from the application. We'll need to pull it out of the guides and automated tests as well. This integration was built before we had external tools, and any new geospatial exploration support should follow our modular framework.

pdurbin:

As explained in #7409 we are removing WorldMap/Geoconnect support from Dataverse.

Closes #7409

The issue says to grab a backup but I took this to mean a full database backup. I didn't implement anything special.

Poke around in the area where the Map It (Geoconnect) button used to be as well as the area where you could view maps.

Yes, maps are gone. The link to create maps is gone.

Probably a good idea but I didn't add one yet.

**Additional documentation**:

None.


djbrooke:
Thanks @pdurbin - if someone wants to start reviewing the code that would be great. I'll take a look at docs and add release notes later this afternoon.

djbrooke:
Ok - release notes added and docs reviewed. Looks good! Still needs some code review.

qqmyers (approved):
While this is large, it looks like everything is a ~straight-forward deletion of Worldmap code and/or simplifying logic in conditional statements to drop the Worldmap special case. 
I didn't check every line, but spot checking all seems good and due to the nature of the change, I'd think moving this on to QA would make sense. The things that could break would presumably be around the external tools displays.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1794,7 +1765,7 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
                 return false;
             } 
             
-        if (permissionService.requestOn(createDataverseRequest(user), df).has(Permission.DownloadFile)) { 
+        if (permissionService.requestOn(createDataverseRequest(user), df).has(Permission.DownloadFile)) { // FIXME: Indent this line more.
```
qqmyers, line 1768: Can this just be changed? (was this a test to see if reviewers would notice :-) ? )

pdurbin: Fixed in a71841346. I wouldn't say it was a test. I always try to keep my diffs as small as possible. 😄

PR #7516, opened 2021-01-20
3438 email verification 0121

### LINKED ISSUE #3438: Email Verification: 2 issues; refresh un-verifies, verify email button should always be available.

sekmiller:
Fixes a bug where the page refresh of dataverse user resends the email verification. Also keeps the "Verify Email" button available while the user's email is unverified.

Closes #3438 

**Special notes for your reviewer**: The underlying cause of the bug was that the page load/refresh was triggering the sendConfirmEmail method which was in the action of the verify button. To fix it I moved it to a remote command. If there's a better way to handle it, I'm open to suggestions.
Also note that the branch has a -0121 on the end of its name due to some git issue I had last week.

I noticed one thing in my testing that I wasn't able to recreate: at least once there was a time when the user had an active token but pressing the "verify" button did not cause the popup to display. I also had an issue where refreshing the page would resend the confirmation email. I think that is resolved.

UI changes have been reviewed/approved by the design team. https://github.com/IQSS/dataverse/issues/3438#issuecomment-760410406

no
**Additional documentation**:
none


scolapasta:
@sekmiller I sent you a suggestion for how to do this without the remote command via slack.

scolapasta (approved):
Good changes to remove the remote command and clean up some.

scolapasta (approved):

Issue #6672, opened 2020-02-24
Unable to resume a "Paused" workflow


vbernabe:
Bug Report: Unable to resume a "Paused" workflow
Dataverse version: 4.18.1
Glassfish: 4.1

We have a use case to "pause" the publishing of a Dataset for a curator/admin to check and approve before publish. I'm trying out the workflow based on the documentation (http://guides.dataverse.org/en/latest/developers/workflows.html)

Steps:
1. I set the default workflow for prepublish dataset
        curl -X POST -H "Content-Type: application/vnd.api+json" -H "X-Dataverse-key:$API_TOKEN"  -d @internal-pause-workflow.json  $SERVER/api/admin/workflows
    
2. I created a Dataset and click on publish - now it is showing "The dataset is locked until publication."
![image](https://user-images.githubusercontent.com/3328254/75126707-1efc0d80-56f6-11ea-9891-65dd75e8d5e3.png)

 3. I tried to resume the workflow but getting an error
![image](https://user-images.githubusercontent.com/3328254/75126735-3cc97280-56f6-11ea-9323-be91a33d2931.png)

pdurbin:
@vbernabe I don't know a ton about the workflow feature apart from what's written at http://guides.dataverse.org/en/4.19/developers/workflows.html

Have you considered using the "Submit for Review" (then "Return to Author" or "Publish") feature? It's documented at http://guides.dataverse.org/en/4.19/user/dataset-management.html#submit-for-review

Here's a screenshot:

![Screen Shot 2020-02-25 at 4 55 23 PM](https://user-images.githubusercontent.com/21006/75291135-a4171d80-57ef-11ea-8e23-118cce558b74.png)

vbernabe:
@pdurbin our users are allowed to created their own Dataverse and add their Dataset under it. So these users will own the Dataverse hence we are not able to set their role as Dataset Contributor only.

If the workflow pause on publish and resume is not working, is there a way for an Admin role to be notified when a Dataset is created? This way at least we can get the notification for a new Dataset and keep tabs on the newly created Datasets. Right now a notification is only send to the creator of Dataset.

pdurbin:
> our users are allowed to created their own Dataverse and add their Dataset under it. So these users will own the Dataverse hence we are not able to set their role as Dataset Contributor only.

[...]

PR #7519, opened 2021-01-20
IQSS/7517-fix-prepublication-workflow

### LINKED ISSUE #7517: Pre-publication workflows fail
In reviewing the possibility of using workflows for #7479, I noticed that the changes to move calling the Finalize Publication method into the success method of the Publish command appear to make Finalize run in parallel with any pre-publication workflow, which I think also still tries to call Finalize a second time when it finishes. In working with DANS, I've now seen evidence that their pre-publication workflow is triggering an OptimisticLockException that may be the result of this problem. It's possible that #6672 is also related.

I'll make a quick PR that addresses what I think is wrong and work with DANS to test it. 

Assuming that works, it should allow #7479 to be modified to use a pre-publish workflow.

qqmyers:
**What this PR does / why we need it**: In earlier code, the call to FinalizeDatasetPublicationCommand was only called from the PublishDatasetCommand when there was no pre-publication workflow. In the workflow case, FinalizeDatasetPublicationCommand is called when the workflow completes. When the call to Finalize publication was moved to the onsuccess method, it ended up executing even when there is a pre-pub workflow which means a) it is called twice, and b) it is potentially executing in parallel with the workflow (async ones at least).

Closes #7517

DANS will be testing. A relatively simple test would be to configure a workflow with the 'pause' step as a pre-publication workflow and to verify that the dataset is not published until after the command to unpause is sent. (Should also verify that doing this workflow without the PR ends up publishing the dataset even without un-pausing the worklflow.)

Note: from the conversation below, there is currently an issue with the 'pause' step being worked on separately. While it is not necessary to test with it, it does simplify testing a bit, so we suggest waiting until that is merged to test this PR.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: none

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
@@ -213,7 +213,11 @@ public class PublishDatasetCommand extends AbstractPublishDatasetCommand<Publish
 
         if (dataset != null) {
             logger.fine("From onSuccess, calling FinalizeDatasetPublicationCommand for dataset " + dataset.getGlobalId().asString());
-            ctxt.datasets().callFinalizePublishCommandAsynchronously(dataset.getId(), ctxt, request, datasetExternallyReleased);
+            Optional<Workflow> prePubWf = ctxt.workflows().getDefaultWorkflow(TriggerType.PrePublishDataset);
+            //A pre-publication workflow will call this when it completes
+            if (! prePubWf.isPresent() ) {
+                ctxt.datasets().callFinalizePublishCommandAsynchronously(dataset.getId(), ctxt, request, datasetExternallyReleased);
+            }
             return true;
         }
         
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
@@ -212,12 +212,12 @@ public class PublishDatasetCommand extends AbstractPublishDatasetCommand<Publish
         }
 
         if (dataset != null) {
-            logger.fine("From onSuccess, calling FinalizeDatasetPublicationCommand for dataset " + dataset.getGlobalId().asString());
             Optional<Workflow> prePubWf = ctxt.workflows().getDefaultWorkflow(TriggerType.PrePublishDataset);
-            //A pre-publication workflow will call this when it completes
+            //A pre-publication workflow will call FinalizeDatasetPublicationCommand itself when it completes
             if (! prePubWf.isPresent() ) {
+                logger.fine("From onSuccess, calling FinalizeDatasetPublicationCommand for dataset " + dataset.getGlobalId().asString());
                 ctxt.datasets().callFinalizePublishCommandAsynchronously(dataset.getId(), ctxt, request, datasetExternallyReleased);
-            }
+            } 
             return true;
         }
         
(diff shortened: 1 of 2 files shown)


janvanmansum:
@qqmyers : I have tried testing our own workflow, but it contains a bug, which has to be solved first, so I'll get back to you about that.

I am now trying to test a minimal scenario, as you suggested.

```json
{
  "name": "Pause",
  "steps": [
    {
       "provider":":internal",
       "stepType":"log",
       "parameters": {
           "invocation ID": "${invocationId}" 
        }
    },{
      "provider":":internal",
      "stepType":"pause"
    }
  ]
}
```
I have add this step and set it as the default for `PrePublishDataset`:

```bash
curl http://localhost:8080/api/admin/workflows --data-binary @pause.json -H 'Content-Type: application/json'
{"status":"OK","data":{"name":"Pause","id":4,"steps":[{"stepType":"log","provider":":internal","parameters":{"invocation ID":"${invocationId}"}},{"stepType":"pause","provider":":internal"}]

curl -X PUT --data 4 http://localhost:8080/api/admin/workflows/default/PrePublishDataset
{"status":"OK","data":{"message":"Default workflow id for trigger PrePublishDataset set to 4"}}
```

I then created and published a dataset and it gets locked and stays locked. 

However I have trouble resuming it. I tried to log the `invocationId` using the variable that is passed to the workflow, but apparently that doesn't work. Fortunately, it is logged anyway by the log step:

```
[#|2021-01-21T12:56:23.257+0100|INFO|Payara 5.2020|edu.harvard.iq.dataverse.workflow.internalspi.LoggingWorkflowStep|_ThreadID=297;_ThreadName=__ejb-thread-pool11;_TimeMill
is=1611230183257;_LevelValue=800;|
  Logging step:|#]

[#|2021-01-21T12:56:23.257+0100|INFO|Payara 5.2020|edu.harvard.iq.dataverse.workflow.internalspi.LoggingWorkflowStep|_ThreadID=297;_ThreadName=__ejb-thread-pool11;_TimeMill
is=1611230183257;_LevelValue=800;|
  Invocation id e94f6da1-8969-4d46-a848-3dcc5357ce7b|#] <============== INVOCATION ID HERE =======

[#|2021-01-21T12:56:23.257+0100|INFO|Payara 5.2020|edu.harvard.iq.dataverse.workflow.internalspi.LoggingWorkflowStep|_ThreadID=297;_ThreadName=__ejb-thread-pool11;_TimeMill
is=1611230183257;_LevelValue=800;|
  Dataset id:10|#]

[#|2021-01-21T12:56:23.257+0100|INFO|Payara 5.2020|edu.harvard.iq.dataverse.workflow.internalspi.LoggingWorkflowStep|_ThreadID=297;_ThreadName=__ejb-thread-pool11;_TimeMill
is=1611230183257;_LevelValue=800;|
  Trigger Type PrePublishDataset|#]

[#|2021-01-21T12:56:23.258+0100|INFO|Payara 5.2020|edu.harvard.iq.dataverse.workflow.internalspi.LoggingWorkflowStep|_ThreadID=297;_ThreadName=__ejb-thread-pool11;_TimeMill
is=1611230183258;_LevelValue=800;|
  Next version:1.0 isMinor:false|#]
```
I then tried to resume `e94f6da1-8969-4d46-a848-3dcc5357ce7b`:

```
 curl -X POST http://localhost:8080/api/workflows/e94f6da1-8969-4d46-a848-3dcc5357ce7b
{"status":"ERROR","message":"Cannot find workflow invocation with id e94f6da1-8969-4d46-a848-3dcc5357ce7b"}
```

Any ideas?

qqmyers:
I'll check - it looks #6672 is a separate issue. I'll look into it.

qqmyers:
@janvanmansum  Some progress - It looks like the log is not showing the right invocationid. If I look in the pendingworkflowinovcation table and use the id there, things work. So - I'll look into why the wrong id is being logged.

qqmyers:
@janvanmansum  - #7521 should address the issue of not being able to resume a paused workflow. If another issue surfaces, let me know.

scolapasta (approved):

janvanmansum:
@qqmyers : Above scenario is indeed fixed with  #7521

PR #7521, opened 2021-01-21
6672-Fix Workflow Resume

### LINKED ISSUE #6672: Unable to resume a "Paused" workflow

qqmyers:
**What this PR does / why we need it**: This fixes a bug introduced in 2018 :-) that resulted in different workflow Ids being recorded in the db and log (and presumably sent the an external client with the https/r workflow step), making it 'hard' to resume the workflow and, for prepublish workflows to then finalize publication. (Things worked if you could get the correct invocationid from the database).

Closes #6672 

**Suggestions on how to test this**: Set up and running a workflow including the pause step. You should then be able to post a success/fail message and have it work. All per Guides instruction. Note that one can test the combo of this and #7519 by setting this up a pre-publication workflow. This fix will allow you to unpause the workflow and #7519 will assure that publication is not finalized until after you do.

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/developers/workflows.rst
+++ b/doc/sphinx-guides/source/developers/workflows.rst
@@ -16,7 +16,7 @@ Workflow steps are created using *step providers*. Dataverse ships with an inter
 
 Steps can be internal (say, writing some data to the log) or external. External steps involve Dataverse sending a request to an external system, and waiting for the system to reply. The wait period is arbitrary, and so allows the external system unbounded operation time. This is useful, e.g., for steps that require human intervension, such as manual approval of a dataset publication.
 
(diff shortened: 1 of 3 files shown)


scolapasta (approved):
Looks good!

PR #7540, opened 2021-01-26
Update data explorer external tools entry

kaitlinnewson:

Updates the data explorer entry in the docs to point to the new version. Also removes reference to R, which is no longer needed for the tool.


djbrooke:
Hey @kaitlinnewson - this change looks good! During review there were a few other spots that I saw that we can/should update. See https://github.com/kaitlinnewson/dataverse/pull/1 for details and feel free to merge that branch into yours, or use the files changed there to determine what to update. It's possible that I missed an instance or two as well.

djbrooke (approved):

PR #7542, opened 2021-01-26
#7536 update .travis.yml to openjdk11

donsizemore:
**What this PR does / why we need it**: Travis builds are currently failing due to the switch to Java 11.

**Which issue(s) this PR closes**:

Closes #7536

**Special notes for your reviewer**: none

**Suggestions on how to test this**: watch for Travis output

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: no

**Is there a release notes update needed for this change?**: no

**Additional documentation**: none

PR #7552, opened 2021-01-27
use locale value

### LINKED ISSUE #6607: The metadata tab of a dataset does not display the translated values
In the metadata tab of a dataset the fields labels are translated but their values are not. However these values are translated in the edit form as well as in the facets of the search page.

For example I get "Other" instead of "Autre" in the french version:
image

But I get this in the facets:
image

And this in the edit form:
image

I noticed the issue for the fields that have controlled vocabulary but I am not sure if it affects the default fields.

qqmyers:
**What this PR does / why we need it**: This corrects an issue where controlled vocabulary values are not translated in the metadata tab display

Closes #6607

**Suggestions on how to test this**: configure for i18n and add a value - verify that the value (not just the term label which was already translated) is in the appropriate language.

**Additional documentation**:


pdurbin (approved):
Seems like a small, straightforward fix.

kcondon:
@qqmyers this is failing a build test:
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   SchemaDotOrgExporterTest.testExportDataset:282 » NullPointer
[INFO] 
[ERROR] Tests run: 1161, Failures: 0, Errors: 1, Skipped: 7

pdurbin:
In addition, if you merge the latest from develop, you'll pick up pull request #7542 should make Travis happy.

I see that Jenkins is also unhappy.

Sorry for approving this prematurely.

qqmyers:
In trying to figure out the failed test, I found two things:

1. Getting a non-existent key from a Bundle can trigger an NPE as well as a MissingResourceException. Simply catching the NPE and creating a MissingResourceException appears to break other tests, so I just caught that in the ControlledVocabularyValue method (so when there is no translation, you just get the default value). 
2. I found a couple places where the displayValue was getting compared to an explicit string value ("ORCID", "Funder") which will clearly fail if the displayValue is i18n. I think these could/should have been using the rawValue anyway, so I made those changes as well. I think the tests that show authors and funder info verify that this works, but one could also test that before the PR, adding an i18n value for ORCID authro identifier type or the Funder contributor type breaks things.

Issue #7564, opened 2021-01-28
Enable Workflows to send a success/fail response to users


qqmyers:
Currently, when workflows complete, they may succeed/fail. However, because they run asynchronously, there has been no mechanism for them to communicate with users and the only record of success/failure was in log messages and, for pre-publication workflows, the failure to finalize publication (which results in the Dataset going back to being in the draft state). For prepublication workflows in particular and workflows in general, it would be useful for them to be able to send success and failure messages to users. Use cases include:

- prepublication workflows adding metadata (e.g. DANS) 
- prepublication spam detection workflow (proposed) which would indicate why publication was not allowed
- postpublication archiving workflow - indicate success/failure in creating an archival copy

The best option for prepublication workflows, at least visually, seems to be adding a message to the dataset page (as with other success/fail messages). However, because workflows are asynchronous, and potentially long lived, and can be triggered by publishing via API, the user may not be  viewing the dataset page. So, it looks like sending a UserNotification (which appears in their notification tab and triggers an email) seems to be needed for a general mechanism. It may be possible to also display a banner-style message in addition/instead (instead would only be a good option if there's a reliable way to tell if the user is going to see a banner).

Sending a notification should be relatively straight forward, so I'd suggest doing that as an initial step and then, if there's interest, looking into whether workflows can keep track of the view state well enough to send a banner-style message. (Any ideas on how to handle the latter would be appreciated!).

Some things to consider: 

- how to manage messages from multiple steps (do steps interact? an initial choice might be to just have steps messaging be independent as most workflows only have one substantial step anyway (some use the log step before and after the main step).)
- whether success/fail messages are required or optional (and is the choice dynamic?)
- how external workflows using, for example, the http s/r step can convey message text
- the message for an external workflow that can't be contacted (e.g. the http call to the external service fails)

For the last, https/r currently uses a regexp to identify whether the step succeeded or failed. This mechanism is useful assuming legacy apps where the response cannot be specified. It could be extended by using the regexp to identify a subset of the response as the message. Alternately, that step, or a variant could expect a standard ~json format (object with 'status' and 'message').

Thoughts welcome. I'll plan to update the issue with more concrete proposal(s).

janvanmansum:
It would be good to at least have generic message on the dataset details page that tells you that publication failed and maybe refers you to the notifications for the details. I would expect something similar to the banner that currently displays when the dataset is locked.

PR #7568, opened 2021-01-31
Allow workflows to make changes

### LINKED ISSUE #7563: Allow external workflows to make changes
Workflows may need to read/write information in the dataset during their execution. However, for external workflows, i.e. ones that will use the API, this is currently blocked by the lock placed for the workflow itself. Use cases include:
PrePublication:

- Adding metadata from an external service (DANS data migration),
- Extracting geospatial information and writing a csv file to the dataset (UCLA geo integration),
- Future spam or virus check analysis that would need to get file contents

PostPublication:
- The archiver workflows set a field in datasetversion to indicate the archivalcopylocation. If this workflow were to be made external, it would have to make that change via an API.

New triggers (TBD):
- External ingest apps that read files after upload and produce metadata. 

Prepublication is the most critical case - for PostPublish and potential triggers like 'UploadFinished', there's no internal Dataverse processing that has to wait, so workflows could immediately respond with a success message and then process as they wish after the lock is released. For pre-publication, responding would trigger FinalizePublication and the workflow couldn't make changes to go into the published version.

The mechanism for lock enforcement I see in API calls is that locks are only checked within the commands themselves, and different lock types are checked separately. 
I can see two options for fixes: either giving the user who started the workflow, or the holder of a token (such as the workflowID), an exemption to avoid being stopped by a workflow lock. (I think all the use cases would work if other lock types are still enforced and I think many would need the protection of those locks (e.g. Ingest, EditInProgress, etc.).

- Commands already know which user is calling them, so it would be easy to add an exception in the current lock checks to allow you to run if there's a workflow lock and you are the user. I think this would be OK in practice since any action that would cause conflicts will also trigger additional non-workflow locks and, if the UI stays as is (doesn't exempt a workflow lock) the only potential for conflicting calls would be API calls from the same user (in theory something like running the DCT to change variable metadata while also having triggered a pre-publication workflow that is in progress could result in uncoordinated changes, but real conflicts _should_ (?) still be stopped by other locks versus causing an exception/server error.)

- An alternate approach that would be stricter but more work would be to add a way to pass the lockID (short and potentially easy to guess) or workflowID (better - long and unguessable at short timescales) to Commands so that the workflow lock check can verify that the API call originates with the workflow itself. In either case, a way to pass the token to the Command is needed, probably through adding it to the DataverseRequest used to start the Command (DataverseRequest has access to the HttpRequest and should be able to read a param or header). To use the workflowID, another addition to keep the relationship between the Workflow and DatasetLock would be needed.

Unless there are other ideas or a strong opinion, I'm inclined to just go for the latter - it's not too much more work and it is may avoid future problems if long-running apps associated with the user's apikey are also running (external request access tools, QDR's annotation tool (in dev), etc.)

### LINKED ISSUE #7569: Workflow locks are not in the db until an http s/r step initially returns
from DANS testing - while a lock is created and em.persist() ed prior to the step being run (and the external app being called), the lock is not actually visible/in effect until the step returns an initial 200/OK response.

In working on #7563, I discovered this is probably due to the fact that the change is not em.flush() ed until the workflow step completes. For the pruposes of #7563, I had to flush the lock to get it's ID (so that the PendingWorkflowInvocation class can store the lock ID which can then be used to decide if an api call from the workflow can override the lock), which presumably will fix this issue as well. 

If not, this will need to be investigated further.

qqmyers:
**What this PR does / why we need it**: Prior to this PR, external workflows (e.g. ones using the http s/r step) could not make changes to the dataset. This PR adds a mechanism to use a workflow invocationID as an alternative to a user api key and, when doing so, to allow an API call on behalf of the user to ignore the workflow lock corresponding to the specified workflow invocation. In this sense, the invocation ID becomes the equivalent of a short-term user API key (only valid until the workflow completes). When an API call authenticates the user this way, it's permissions are still limited to those of the user and any action taken is still constrained by any other locks on the dataset. 

Closes #7563
Closes #7569

**Special notes for your reviewer**: I added a header to send the workflow invocation ID (and a query param), both analogs of the header/param for the api key. It should be getting checked on any call that checks for the user (via the findUserOrDie() or findAuthenticatedUserOrDie() calls). As discussed in the issue, this is an incremental security improvement - services implementing workflows don't need to have a long-lived api key to work on behalf of users, but it isn't as secure as doing something like presigned-urls which could limit which api calls the workflow could make. Nominally, one could consider this two PRs: enabling use of the invocationID to authenticate to the api, and additionally allowing auth via this method to ignore the lock created by this workflow.

**Suggestions on how to test this**:  Run a pre-publication workflow with a pause step, adding a "parameters" object to the json with "authorized":"true". Get the invocation id from the log. Then, try some API call, such as the one to edit the dataset metadata.  (Alternately, setup an http/authext step and a client that would get the invocationID and other info from Dataverse). Using the users API key, it should fail noting the workflow lock. Then use the invocationID and the 'X-dataverse-invocationID' header or ?invocationID= param key and the same call should succeed. One could also verify that the invocationID works for anything else the user has permissions for (e.g. creating/changing a different dataset). One could also add an additional lock to the dataset (any reason code, but if it is for a workflow, it can't have the same invocationID)) and verify that authenticating via invocationID is stopped by the new lock when trying to make changes to the locked dataset. 

To test w.r.t. #7569, a workflow with an http s/r step has to be created. Verifying that the lock exists and is visible in the api prior to the initial 200/OK response would verify the fix. It may be easiest to just as DANS to check to see if this works as they have an http s/r step configured.

**Additional documentation**:


landreev (commented):
This makes sense overall. But this needs to be documented, in workflows.rst, and maybe in native-api.rst too. I was told a few times that all the API functionality must be documented, and that we shouldn't have any unadvertised API calls or parameters, even if it's  experimental. And this is something we clearly want other installations to use. 
(I'm assuming this is the policy - @scolapasta?)
As for #7569, I haven't tested it myself, but, looking at the code, I'm a little confused about this part: 
> I had to flush the lock to get it's ID ... which presumably will fix this issue as well.

It just looks like it was already being flushed, as of v5.3, or v4.18.1 that they appear to be running at DANS:
```
datasetLock.setDataset(ctxt.getDataset());
em.persist(datasetLock);
em.flush();
```
vs. the corresponding fragment in your branch: 
```
datasetLock.setDataset(ctxt.getDataset());
em.persist(datasetLock);
//flush creates the id
em.flush();
ctxt.setLock(datasetLock);
```
so the only difference is the ctxt.setLock() line - but that's only setting it in the workflow context... which shouldn't have an affect on whether the lock gets saved in the database (?)
I may still have a question about the security implications of this PR, but let me try and figure it out. 
Sorry for the  delay with this; I actually knew next to nothing about workflows. I want to understand how they work, since I may need to use a workflow to clean up the spam branch. So I've been educating myself in real time.

qqmyers:
@landreev - thanks! I tend to write documentation last as review can change the code.
W.r.t. the flush - you're right, which explains why this didn't fix the issue. Given that the locking occurs in a method that RequiresNew transaction, I don't understand why calls to the API to get locks don't see the new lock until the overall workflow thread finishes (e.g. after it starts the workflow, runs a send/receive step and that step has replied with a 200 response, and the pendingworkflowinvocation is stored to the db and the thread ends.). I thought RequiresNew would result in changes when the lock method ends...

landreev:
@qqmyers 
> which explains why this didn't fix the issue

OK, so you have confirmed, that the lock is really not in the database, until that OK is received? (as opposed to not being seen by the page; and/or other class instance that may be looking at a stale ejb copy of the dataset object?)

qqmyers:
~Yes. DANS created an external test (python/flask) that serves as an http s/r step. When that step is called, it calls the /api/datasets/{id}/locks api (no type parameter so should get all lock types) before sending it's OK/200 response back to dataverse. Seems like a good test, but - checking the code, that call does a findOrDie on the Dataset, so retrieving from the database, but it is looking for the list of locks on the Dataset, versus querying the DatasetLock table. Does that give a clue?

landreev:
The security question I have: since knowing the invocation id allows anyone to use it in lieu of the api token of the publishing user, are there any potential use cases of external workflows where we may NOT want the remote service to be able to make changes to the dataset? 
I can think of examples of "http/sr" workflow steps - like getting the dataset registered with an alternative persistent identifier service - where we would not want to give them the ability to mess with the dataset... but then this particular example could be implemented as an *internal*, i.e. synchronous http/sr workflow step - where we don't have to pass the invocation id to the remote service. (Our example in the guide, of an internal http/sr workflow - scripts/api/data/workflows/internal-httpSR-workflow.json - does pass the invocation id along: 
```
{
            "provider":":internal",
            "stepType":"http/sr",
            "parameters": {
                "url":"http://localhost:5050/dump/${invocationId}",
                "method":"POST",
                "contentType":"text/plain",
                "body":"${invocationId}\ndataset.id=${dataset.id} /\ndataset.identifier=${dataset.identifier} 
.../",
                "expectedResponse":"OK.*",
                "rollbackUrl":"http://localhost:5050/dump/${invocationId}",
                "rollbackMethod":"DELETE"
            }
        },
```
... but we don't have to really - ?)
The workflow.rst guide says:

> External steps involve the Dataverse Software sending a request to an external system, and waiting for the system to reply. The wait period is arbitrary, and so allows the external system unbounded operation time. This is useful, e.g., for steps that require human intervention, such as manual approval of a dataset publication.

this example, a "manual approval", does suggest a possible scenario where an external workflow is being used to wait for another human user to review something, without necessarily giving them write access to the dataset - ? 

As always, there's a good chance I'm missing something... like this granting of write access to the dataset already being optional, and not automatic for any http/sr workflow and anyone with the invocation id...

landreev:
> ~Yes. DANS created an external test ...

Speaking of which, I meant to ask if you could add  their workflow json, to this PR, or to #7569. (Could help Kevin with testing too; I'm actually not seeing any examples of external workflows in our documentation...)

> ...  that call does a findOrDie on the Dataset, so retrieving from the database, but it is looking for the list of locks on the Dataset, versus querying the DatasetLock table. Does that give a clue?

Hmm. I'm pretty sure that should be resulting in an equivalent new query on the datasetlock table.

landreev:
Hold on, I was probably confusing the "internal provider" with "internal workflows"... Meaning, ALL http/sr are in fact "external", and always are asynchronous, and always rely on a callback... and therefore always need the invocatioh id (?)
Sorry for the confusion... but that still leaves my main question open, about giving write access to all these external components...

qqmyers:
@landreev - I think your analysis is all on target. Assuming that the use case to allow workflows to make changes is valid, I could see a few ways to alter the design to still allow workflows that don't get access. The simplest would be to add a flag to allow/disable access as part of the workflow registration json. Alternately, a different token could be sent to enable access, and/or the optional ability to send the user api key could be added (so the invocationId would have no power and you'd either need to send Dataverse a separate key (which could live as long as the workflow), or the user api key and invocationId (so you know breaking the lock is allowed). Eventually, finer-grained access using something like presigned URLs as discussed for external tools could work here as well (have to think about the ability to still break the lock with that though).

landreev:
(I edited/streamlined the comments above)

landreev:
I am OK with the idea of (potentially) addressing this in the future, if we need to. Since this PR solves the problem at hand for this installation, and since we are not aware of any real life use cases where granting write access in this situation is undesirable. 
I would however point this out in the guide explicitly, that by creating a remote http/sr workflow, you are granting write access to the outside component. And that if anyone has a use case where this is not ok, this would need to be changed, using one of the approaches you outlined.

djbrooke:
Thanks @landreev and @qqmyers for the discussion - @qqmyers it sounds like some docs are the remaining thing here, so let me know if there's anything I can do to help.

qqmyers:
@landreev, @djbrooke - I've modified this PR to include a new http/authext step which is the only one allowed to use the invocationId as a key, so http/sr - based workflows won't be affected. Also added documentation. (Although I've been testing in concert with other changes recently in #7564, this PR should be ~independent : the one cross-connect with 7564 is in the AuthorizedExternalStep class and documentation for it where I've included the message parameter you can send back. In this PR, it exists but does nothing, and with the PR for #7564, it will get picked up and displayed.)

djbrooke:
Thanks @qqmyers - the docs look good from my perspective. 

@landreev can you take another look when you have a chance? Thx.

landreev (approved):
This is great, I like the idea of a dedicated new class of external workflows for this. 
I'm approving the PR and moving it along. 
A couple of quick questions about the PR description metadata: 
- under "how to test" - probably needs to be changed, to specify the new workflow type, not the existing "http s/r";
- is it still expected to fix #7569?

qqmyers:
Thanks - I adjusted the test info to reference the http/authext step. (FWIW, it looks like testing with pause is still possible as it echos it's input params (in the json you start the wf with) into the pendingworkflowinvocation class so it can be configured to allow auth via invocationId if desired. (Also noted in the docs that it really is for testing since the invocationId is only in the log/db and not available elsewhere (and you need that to restart the wf).

As for #7569 - I think you're right that the original code also does em.flush() so this PR may not change it. That said, I'm not sure why it wasn't working and the other transaction-related changes could have fixed it, so I'd at least leave it in to test (or have DANS test) and then remove the issue from Closes if it doesn't solve that issue.

PR #7577, opened 2021-02-03
7370 zip upload warning

### LINKED ISSUE #7370: Zip uploads - file size limit warning not shown
Short version: 
When an uploaded archive is unzipped, if one or more files exceed the size limit, there supposed to be a warning message offering to either save the zipped version, or discard the upload. Instead (in most cases) the zipped archive is saved quietly. (i.e., without unpacking the individual files, and without any extra feedback to the user). 

Will add more info below, there is an interesting special case there.

sekmiller:
Returns a warning on file upload that was removed as redundant during the file replace changes.

Closes #7370

see notes in the ticket. also on file replace the warning should only be in the popup not the page.

**Additional documentation**:


mheppler (approved):
👍 Looks good. Appreciate efforts to move text to the bundles as well.

kcondon:
@sekmiller Tania reviewed the current message and proposed using this instead: "A file contained in this zip file exceeds the size limit of xyz. This Dataverse installation will save and display the zipped file, rather than unpacking and displaying files."

PR #7581, opened 2021-02-05
#7205 fix dataset page load

### LINKED ISSUE #7205: Dataset: Loading dataset page fails with exception if original file size is null, eg. some legacy Murray
[Leonid]
Updating the issue, since we had some incorrect assumptions back when it was opened (L.A.):
1. Yes, dataset page is bombing if the original size is null in the db. for one of the tab. files. (when it tries to calculate the total size for "download all"). 
2. This has nothing to do with whether it can access S3, or whether the actual saved original exists: there is NO attempt to check the size of the stored physical file at this point. A null in the db results in a null pointer. (The code that looks up the physical file, if the value is null and then saves the size in the db does exist; but only in the Acces API).
3. For most installations seeing this issue, this would be resolved by running `/api/admin/datafiles/integrity/fixmissingoriginalsizes`. We recommended (but not required!) to do this in 4.10  release; but some installations may have missed it. 
4. The code that throws that null pointer in this case should of course be modified, as not to kill the page. (It should simply assume that null = 0). 
5. Some ancient installations (like ours) may have legacy tab. files with no saved originals. In addition to addressing it with the fix in 4., we should probably also modify the `fixmissingoriginalsizes` API, not to leave any nulls in the db, but replace them with -1 in cases where the originals are missing. That's what I did in our prod. manually. (and then maybe we could remind people with pre-4.10 legacy installations to run that API)

All of the above should be very straightforward. 

[Kevin] Original comment
It was discovered that loading the dataset page fails with an exception when original file size is null and cannot be calculated. This happened with a Murray dataset on a test instance. It can happen if 1. a legacy file does not have an original file 2. the file system is temporarily inaccessible and so the get size function fails.

It would seem that this should not be a fatal error in loading the dataset. After consulting with Dev, it seems that this behavior was added when the down all files in a dataset feature was added because it wants to alert you to the size of download.

-----

From @qqmyers in #6343:

When Amazon had a recent S3 outage (DDOS attack), I noted that Dataverse can no longer display file pages when S3 is down. Previously, the only issue with S3 being out related to operations with file upload/download and the Dataverse site itself worked. This issue now is fairly trivial - it looks like the file size to be displayed on the page is being pulled from S3 metadata and hence, if S3 is down, the page breaks/shows a 500 error. I'm not sure its worth fixing, but avoiding 500 errors could be as simple as doing a null check on the file size. (In any case, this issue might help someone debug - 500 errors related to file sizes could be due to S3 being misconfigured/down,)

sekmiller:
The Dataset Page fails to load when a tabular file has an original file size of null. A one line change assigns a value of zero to the size when it is null so that the page will load properly

Closes #7205

This also suppresses the download original at the dataset level if the full size of the original files is zero (none have an original file size.)

Null out the original file size of a tabular file and verify that its dataset page will load.
update datatable set originalfilesize = null where datafile_id = ??;

no
Not really unless we want to remind installations to run the fixmissingoriginalsizes api

**Additional documentation**:

### DIFF
--- a/doc/release-notes/7205-orig-file-size.md
+++ b/doc/release-notes/7205-orig-file-size.md
@@ -1,0 +1,11 @@
+## Notes to Admins
+
+Beginning in Dataverse Software 4.10, the size of the saved original file (for an ingested tabular datafile) was stored in the database. For files added before this change, we provide an API that retrieves and permanently stores the sizes for any already existing saved originals. See [Datafile Integrity API](https://guides.dataverse.org/en/5.4/api/native-api.html#datafile-integrity) for more information.
+
+This was documented as a step in previous release notes, but we are noting it in these release notes to give it more visibility.
+
+## Upgrade Instructions
+
+X./ Retroactively store original file size
+
+Use the [Datafile Integrity API](https://guides.dataverse.org/en/5.4/api/native-api.html#datafile-integrity) to ensure that the sizes of all original files are stored in the database.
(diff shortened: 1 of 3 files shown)


pdurbin (approved):
I didn't test the fix myself but it's small and simple and makes sense.

scolapasta (approved):

PR #7589, opened 2021-02-05
log the stack trace for Throwable/500 errors

### LINKED ISSUE #7586: Add stacktrace for some 500 errors
The edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler promises more info in the log when it catches an Exception and throws a 500 error, but there's often little info as Throwables like NPEs aren't things that are logged elsewhere. So - would be good to print the stack trace to the log to help with debugging. Simple PR to follow.

qqmyers:
**What this PR does / why we need it**: aids in debugging 500 errors

Closes #7586

**Suggestions on how to test this**:  Calling curl -X PUT -H 'Content-Type:application/json'  http://localhost:8080/api/datasets/2/editMetadata --upload-file edit.json 

should trigger this - no credentials, some minimal valid json file as in the guide. (See#7587). The stack trace should show in the log (and point to an IllegalCommandException when the UpdateDatasetVersionCommand runs on behalf of the 'Guest' user.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/ThrowableHandler.java
```diff
@@ -27,6 +27,7 @@
     
     @Override
     public Response toResponse(Throwable ex){
+        ex.printStackTrace();
```
poikilotherm, line 30: While I agree on logging this (scratching my head if we removed that or why it had not been added), why not use a proper Logger instance?

poikilotherm (changes requested):

src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/ThrowableHandler.java
poikilotherm: Ok I see now. While we do log the exception, the stack trace is not.

Here's a thought: let's extend the `.log()` with a logging of the stack trace at level fine plus add a method `logStackTrace(Throwable ex)` to the response builder, which is doing the logging at a more prominent level like SEVERE.

Another thought: we could also extend the present logging function with a bollean parameter allowing to switch logging the stack trace along with the log message.

poikilotherm: Sorry if this looks like nitpicking, it's just that I don't like sending the stack trace to sysout. Strong opinion of sending it the same way as any logs here...

qqmyers: @poikilotherm - this was just something that's bitten me a few times, so I wanted a fix. I think all your suggestions are great, so if you have the time to change this PR or create an alternate, please do.

poikilotherm:
Hi @qqmyers,
I looked into this today and wonder why the stack traces are not logged, because they should when running on Payara.

Looking at https://blog.payara.fish/the-basics-of-logging-in-payara-server and https://docs.payara.fish/community/docs/5.2020.7/documentation/payara-server/logging/logging.html, I wondered what exactly the Payara Log Formatters do.

Digging deeper, [ODLLogFormatter](https://github.com/payara/Payara/blob/2e573a7dd61e3d05c834cd1ec125b883ec62f9bd/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/ODLLogFormatter.java#L411-L423) and [UniformLogFormatter](https://github.com/payara/Payara/blob/2e573a7dd61e3d05c834cd1ec125b883ec62f9bd/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/UniformLogFormatter.java#L450-L459) seem to add the stack trace on their own, when the exception is handed over (as we do).

So now I wonder if this might be related to copying the old Glassfish Domain to Payara, not using a recent [DOMAIN_DIR/config/logging.properties](https://github.com/payara/Payara/blob/f44fa70ead58ecc27281a465b53d637eee79c4f3/nucleus/admin/template/src/main/resources/config/logging.properties) file.

Before investing more hacking time into this, could you take a look at your config? Thx!

djbrooke:
@qqmyers and @poikilotherm thanks for the discussion here. Let me know when there's something here that we should plan to QA and merge. I'm all in favor of moving forward and merging small batches that provide incremental improvement while waiting on larger changes.

qqmyers:
@poikilotherm - I was doing this on a fresh dataverse-ansible install. @donsizemore - any reason ansible wouldn't be setup to use the payara loggers above?

poikilotherm:
Hey @qqmyers,

I tried to create a reproducer with latest Payara 5.2020.6 (tested also with 5.2020.7), but wasn't able to: the stack traces appear just fine. 

https://github.com/poikilotherm/payara-test-jaxrs-logging

Does this work on your machine, too? If so, we can narrow the problem scope to be Dataverse project specific and refocus testing.

qqmyers:
@poikilotherm - I'm using dataverse-ansible and don't see the stack trace when calling real API calls. The logging.properties file does list the two logger classes you mention above.

I thought I might have found a useful difference but I'm no longer so sure: the failing API calls use https://github.com/IQSS/dataverse/blob/fbc81d302ddfb16006ae199371c7ef0c182f59c3/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L99 - which creates an Exception (the WrappedResponse class) without a cause, and without calling the super's constructor. However, I've tried that, added a call to fillStackTrace(), and then even through a NullPointerException from within the constructor, and, with the current dataverse-ansible config, I don't see a stack trace. So - perhaps it is some logging config difference between ansible and your setup?

PR #7592, opened 2021-02-08
Removed extra link from Feature Dataverse carousel [ref #7476]

### LINKED ISSUE #7476: Accessibility - carousel images and text links duplications for screenreader users
Hello Harvard team,

I wanted to flag an accessibility issue, tested in 5.1/5.2, with the "featured dataverse" carousel.

When navigating the carousel via a screenreader, the images and text below them are both links with the same title text. This results in duplicate links and text read for a screenreader user.

For example, a user hits image in carousel via tabbing, and the SR reads "Dataverse name - link". User tabs to next link and the same thing is read again. This also adds extra unnecessary tabbing through duplicate links.

One way this can be resolved is by removing the images in the carousel for screenreader users with something like `tabindex="-1" aria-hidden="true"` in the images.

mheppler:

Improves UX for users who visit Dataverse using a screenreader. Removes extra link inside Featured Dataverse carousel, so there's only one link for screenreaders to read in each carousel item, rather than two duplicate links.

Closes #7476 Accessibility - carousel images and text links duplications for screenreader users 

Quick tutorial on [how to use Mac OS's VoiceOver app as a screenreader in Safari](https://webaim.org/articles/voiceover/).

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dataverse.xhtml
+++ b/src/main/webapp/dataverse.xhtml
@@ -580,17 +580,13 @@
                                     </div>
                                     <div id="featuredDataversesList" class="owl-carousel owl-theme">
                                         <ui:repeat value="#{DataversePage.carouselFeaturedDataverses}" var="dv" varStatus="status">
-                                            <div class="item">
-                                                <div style="vertical-align:middle;">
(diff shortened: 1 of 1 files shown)


sekmiller (approved):
Looks good. needs to be updated from dev

PR #7593, opened 2021-02-09
Added minor UI updates to sample custom hmpg

mheppler:

Recent changes to the Harvard Dataverse custom homepage have been applied to the sample custom homepage provided in the Installation Guide.

>there was also a minor update to the homepage customization to resolve a minor CSS related layout issue where the "Find" button text label wraps under the magnifying glass icon in the search input-group button. The text label was removed from the button and added to a Bootstrap tooltip which displays on hover.

image

Closes # N/A

Related to fixes included in https://github.com/IQSS/dataverse.harvard.edu/issues/94

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/custom-homepage-dynamic.html
+++ b/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/custom-homepage-dynamic.html
@@ -94,9 +94,12 @@
                     <div class="search-widget input-group">
                         <input id="inputDataverseSearch" class="form-control" type="text" placeholder="" onkeydown="if (event.keyCode == 13) document.getElementById('btnDataverseSearch').click();">
                         <span class="input-group-btn">
-                            <button id="btnDataverseSearch" class="btn btn-default" type="button" onclick="window.location = '/dataverse/root?q=' + document.getElementById('inputDataverseSearch').value;return false;"><span class="glyphicon glyphicon-search"></span> Find</button>
+                            <button id="btnDataverseSearch" class="btn btn-default bootstrap-button-tooltip" type="button" data-original-title="Find" onclick="window.location = '/dataverse/root?q=' + document.getElementById('inputDataverseSearch').value;return false;"><span class="glyphicon glyphicon-search"></span></button>
                         </span>
                     </div>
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/custom-homepage-dynamic.html
```diff
@@ -94,10 +94,13 @@
                     <div class="search-widget input-group">
                         <input id="inputDataverseSearch" class="form-control" type="text" placeholder="" onkeydown="if (event.keyCode == 13) document.getElementById('btnDataverseSearch').click();">
                         <span class="input-group-btn">
-                            <button id="btnDataverseSearch" class="btn btn-default" type="button" onclick="window.location = '/dataverse/root?q=' + document.getElementById('inputDataverseSearch').value;return false;"><span class="glyphicon glyphicon-search"></span> Find</button>
+                            <button id="btnDataverseSearch" class="btn btn-default bootstrap-button-tooltip" type="button" data-original-title="Find" onclick="window.location = '/dataverse/root?q=' + document.getElementById('inputDataverseSearch').value;return false;"><span class="glyphicon glyphicon-search"></span></button>
```
sekmiller, line 97: Is this a situation where we're not using something like:
'/dataverse/#{dataverseServiceBean.findRootDataverse().alias}?q='
because we don't have access to the app in the homepage?

scolapasta: I'll let @mheppler confirm, but yeah this is html here and not xhtml. you could imagine if we needed to we could add an API to get the alias if the root. But seems like overkill here.

sekmiller (approved):

doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/custom-homepage-dynamic.html
mheppler: Correct. Static custom html content has the root alias hardcoded all over the place. We do get plenty from the API already. Surprised we dont here or anywhere else already.

PR #7597, opened 2021-02-09
#6972 allow pdf download if "original" is selected.

### LINKED ISSUE #6972: Download: 404 error for pdf downloads in projects with ingested tabular files

sekmiller:
If a user selects "original" file format as the download option for a non-tabular file it fails.

Closes #6972 

Applies a test of whether a file selected to be downloaded as "original" is tabular. Similar to how it is handled for a multiple file download.

If there are tabular files present in a dataset the option for downloading any file as "original" will be present. The user should be able to select a single non-tabular file and successfully download it by selecting "original".

no
no
**Additional documentation**:
none

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -331,6 +331,14 @@ public class Access extends AbstractApiBean {
             // So we need to identify when a service is being called and then let checkIfServiceSupportedAndSetConverter see if the required one exists
             if (key.equals("imageThumb") || key.equals("format") || key.equals("variables") || key.equals("noVarHeader")) {
                 serviceRequested = true;
+                //In the dataset file table context a user is allowed to select original as the format
+                //for download
+                // if the dataset has tabular files - it should not be applied to instances 
+                // where the file selected is not tabular see #6972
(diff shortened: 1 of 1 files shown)


scolapasta (commented):
This looks reasonable enough as a quick fix, but I'll ask @landreev to also give it a quick lookover, as he had commented on the issue.

landreev (approved):
Approving the PR. 
My only suggestion would be to maybe add one sentence under "how to test", and clarify, that the choice of "tabular" vs. "original"  will still be showing in the pulldown. But the download will actually work for a single non-tab. file with the fix in place.

Issue #7601, opened 2021-02-10
API Token - Copy to clipboard REDUX


mheppler:
Phil noticed that when copying the API token in Firefox, you get some extra returns before and after the token when you paste from your clipboard.

<img width="674" alt="Screen Shot 2021-02-10 at 3 14 48 PM" src="https://user-images.githubusercontent.com/687227/107566379-d497c900-6bb2-11eb-9881-164569083444.png">

Also, when you click the copy button, the "OK" success icon class is added to the btn, and it's not intended to be added to that button.

mheppler:
Retro actively closing this issue as it appears I created it after creating PR #7603 and never linked the two. The PR was merged to develop with a82b7c3ff112073cb9b1653f6a400a2520015999, as shown above.

Issue #6039, opened 2019-07-19
Copy buttons for API Token and Private URL


DS-INRA:
When generating an API token and private URL, a "Copy to clipboard" button would ease (and avoid errors) the copy the token or private URL you just generated.
An example of the proposition :
![chrome_p57y42xsLe](https://user-images.githubusercontent.com/46443753/61533585-987ba100-aa2d-11e9-9db3-3a910c23de0d.png)

As @pdurbin pointed out, this could be made as in the QDR installation for SHA-512 hashes :
<img width="376" alt="Screen Shot 2019-07-17 at 8 05 10 AM" src="https://user-images.githubusercontent.com/46443753/61533855-4b4bff00-aa2e-11e9-8497-5fd403b935db.png">

pdurbin:
This week @mheppler presented some great work in pull request #7312 that includes "click to copy". I said I'd love to have this for API tokens (I'd use it right now if I could, which is what reminded me to go look for related issues).

mheppler:
![image](https://user-images.githubusercontent.com/687227/105873039-11679b80-5fc9-11eb-93d7-877bcdd28df8.png)

Issue #7534, opened 2021-01-25
intermittent 500 errors on signup / login


donsizemore:
Depending on the route one takes to either sign up and/or login to Dataverse 5.3, any user might experience an error screen on login:

<img width="1188" alt="Screen Shot 2021-01-25 at 11 29 55" src="https://user-images.githubusercontent.com/12831666/105734479-abafdc80-5f00-11eb-9765-50f6c7f7a66c.png">

server.log says

```
{DataverseUserPage.init}: java.lang.NullPointerException
javax.faces.FacesException: #{DataverseUserPage.init}: java.lang.NullPointerException
        at com.sun.faces.application.ActionListenerImpl.getNavigationOutcome(ActionListenerImpl.java:96)
        at com.sun.faces.application.ActionListenerImpl.processAction(ActionListenerImpl.java:71)
        at javax.faces.component.UIViewAction.broadcast(UIViewAction.java:548)
        at javax.faces.component.UIViewRoot.broadcastEvents(UIViewRoot.java:847)
        at javax.faces.component.UIViewRoot.processApplication(UIViewRoot.java:1396)
        at com.sun.faces.lifecycle.InvokeApplicationPhase.execute(InvokeApplicationPhase.java:58)
StandardWrapperValve[Faces Servlet]: Servlet.service() for servlet Faces Servlet threw exception
java.lang.NullPointerException
        at edu.harvard.iq.dataverse.authorization.providers.builtin.DataverseUserPage.init(DataverseUserPage.java:170)
```

@scolapasta says "The error is happening on: switch (selectTab) { and may be an artifact of moving to Payara."

The error seems to be intermittent, and browser-independent. This morning I could produce it consistently in Chrome while creating dummy accounts through the web interface and signing in directly, though a more reliable way to produce it is to click on the Sign Up tab (but don't sign up), then the Login tab, then log in. Once the user is logged in and moves off of dataverse.xhtml all other pages display correctly.

A cosmetic snafu under the hood, but one an end-user might not understand, and a "Tums" moment for Dataverse administrators who encounter the error while testing configuration changes and forget that it's not uncommon on 5.x.

djbrooke:
Thanks @donsizemore - I've seen this too.

djbrooke:
- we suspect this may be caused by the dataverseuser page is being called without the selectTab parameter. It should have a default value but maybe this is not happening and null is being sent instead
- it could also be due to redirects (@qqmyers may have some fix implemented for QDR related to this)

PR #7603, opened 2021-02-10
Fixed extra returns copied with api token on account pg [ref #7601]

mheppler:

Fixes odd bug in Firefox where extra returns were added before and after the api token when copied to the clipboard. Also fixes the copy function success from trying to add an "OK" success icon to the copy button.

Closes # N/A

Related to #6039 Copy buttons for API Token and Private URL 

To resolve this, rather have the plugin get the token from the `code` element on the pg, I am passing the token directly to the clipboard with a data attribute `data-clipboard-text="#{ApiTokenPage.apiToken}"` on the copy button.

Make sure you're only getting the single string of the api token from your clipboard, with no additional returns before or aftere. Phil was using the terminal commandline to prove this. Something pasting in a WYSIWYG that formats the text will not show you these odd invisible returns.

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dataverseuser.xhtml
+++ b/src/main/webapp/dataverseuser.xhtml
@@ -521,7 +521,7 @@
                                 //]]>
                             </script>
                             <div class="btn-toolbar" role="toolbar">
-                                <button class="btn btn-default btn-copy" data-clipboard-action="copy" data-clipboard-target="#apiToken code" type="button" jsf:rendered="#{ApiTokenPage.checkForApiToken() and !ApiTokenPage.tokenIsExpired()}">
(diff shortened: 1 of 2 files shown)


pdurbin (approved):
I didn't test this but the code change makes sense.

kcondon:
This works but I'm finding page refresh sets tab to account info in FF, stays on api tab on chrome . Doesn't happen on demo, stays on api tab. Note that on develop it refreshes to My Data, there was recently a discussion/fix/change to this page that made my data tab the default, see Stephen or Gustavo.

mheppler:
Pretty sure nothing in this PR would effect the selected tab state on page refresh. Since there was work recently done to this exact issue in #7534 4d430c266f4b652cfec681deb3e0b9b56108bbe1, I would suggest this PR fixing an entirely different bug not get held up by those changes, and any additional issues with the default tab state gets a new issue.

kcondon:
@mheppler Were you able to reproduce the behavior though? I don't want to merge something that alters develop.

PR #7608, opened 2021-02-12
7551 expanded compound datasetfield validation

### LINKED ISSUE #7551: Silent publishing failure when not all fields required by Datacite are present
Short version, this is an attempt to consolidate #6586 and #7518. #7518 has some good logging details and information about the crosswalk.

I'm thinking we'd like to:

- Have an API endpoint that takes into account PID provider requirements as well as Dataverse requirements when returning a response (basically #6586, just not sure if it should be an existing or new API). There is a wrinkle here that some instances don't use DataCite and it may need to be taken into account when returning what's valid or not, but I'm sure we can work this out.
- Have better logging at publish time if there's a failure related to a missing field. Whenever this happens it seems like we have to reinvestigate because I forget about this issue. 🤯 
- Better communication to the user when something can't be published because of this (or require it when the Dataset metadata is being added). Right now it just leaves the data in an unpublished state after a silent failure.

Other thoughts welcome...

### LINKED ISSUE #7606: Custom Metadata: Allow Dataverse Installations to Define Conditionally Required Fields for Compound Fields
Currently, when defining metadata for compound fields, fields can either be optional or required, i.e. if required you must always have (at least one) value for that field. For example Author Name being required means you must have at least one Author with an nonempty Author name.

An example of Optional (and a simplified table for values in the defining tsv file):

image

An example of Required (and a simplified table for values in the defining tsv file*):

image

(Note: the tsv technically currently has FALSE for the parent, and the PR for this issue will have to change them to TRUE; but we show TRUE here to differentiate for how we will define the new "Conditionally Required" fields)

In order to support more robust metadata (and specifically to resolve #7551), we need to allow a third case: **Conditionally Required**, that is, the field is required if and only if any of its "sibling" fields are required. For example, a user does not have to enter a Producer, but if they do, they have to enter a Producer Name. 

Here is an example of Conditionally Required (and a simplified table for values in the defining tsv file):
image

The other needed change will be how to handle Dataverse level input levels (when you can make something required that isn't required by the installation). In order to minimize the design burden, for this issue, we'll just modify the default "Optional" radio button to be "Conditionally Required" for these fields. In other words, a Dataverse Admin can make such a field more strict by making it fully required, but they can not reduce it to just optional. Additionally, they will not be able to make Optional fields Conditionally Required yet.

image

scolapasta:
Adds the ability to set subfields of a compound metadata field as Conditionally Required.

(Still needs any guide changes)

Closes #7606 
Closes #7551

In addition to reviewing the citation.tsv (included in this PR), it would be good to make sure I didn't miss any required fields in the other tsv files.

Check display of Producer Name and validation. Change default requirements for Producer name for your dataverse. Also make sure, currently Required fields still work correctly.
While you don't have to reupload citation.tsv (since the changes are also handled in the flyway script, it would be a goo check, for example, on a fresh installation.
Bonus testing, modify a TSV to add more Conditionally Required fields and see if they work correctly. 

Yes, some screenshots are in #7606

Yes, installations will need to modify custom tsv files

**Additional documentation**:


src/main/resources/db/migration/V5.3.0.3__7551-expanded-compound-datasetfield-validation.sql
```diff
@@ -0,0 +1,16 @@
...
+(SELECT parentdatasetfieldtype_id FROM datasetfieldtype WHERE required=true);
+
+
+-- add input level rows for parents of compound fields  where a subfield has required input level
+-- (first run a delete to clear out any possible existing parent values)
+DELETE FROM dataversefieldtypeinputlevel WHERE datasetfieldtype_id in (
+SELECT parentdatasetfieldtype_id FROM datasetfieldtype dsft);
+  
+INSERT INTO dataversefieldtypeinputlevel (include,required,datasetfieldtype_id,dataverse_id)
+SELECT true, true, parentdatasetfieldtype_id, dataverse_id FROM
```
sekmiller, line 13: Need a "DISTINCT" here (line 13) for cases where there are multiple datasetfieldtypes with the same parent that are set to required.
(I only caught this because I was playing around with input levels while working on another issue.)
Without the "DISTINCT" I couldn't deploy the branch to my local.

src/main/resources/db/migration/V5.3.0.3__7551-expanded-compound-datasetfield-validation.sql
```diff
@@ -0,0 +1,16 @@
...
+dataversefieldtypeinputlevel dsftil, datasetfieldtype dsft
+WHERE dsftil.datasetfieldtype_id = dsft.id
+AND parentdatasetfieldtype_id IS NOT null;
```
sekmiller, line 16: This does not modify the Producer Name to make it conditionally required to match what is happening in citation.tsv 

We need to add something like:

update datasetfieldtype set required = true where name = 'producerName';

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
```diff
@@ -407,20 +407,23 @@ public void setValidationMessage(String validationMessage) {
...
             required = false;
-            if (this.datasetFieldType.isPrimitive() && this.datasetFieldType.isRequired()) {
-                required = true;
-            }
-
-            if (this.datasetFieldType.isHasRequiredChildren()) {
+            
+            if (this.datasetFieldType.isRequired()) {
                 required = true;
             }
```
sekmiller, line 424: Does it make sense to put a return here?

djbrooke:
Thanks @sekmiller for the review.

@jggautier (and @scolapasta) - do you have thoughts about putting this in the Guides? I added a small bit to the Dataverse Management Guide in https://github.com/IQSS/dataverse/pull/7608/commits/9b0cab63b60aff0ebc8a902e9bc34bd161bbb000. I think we should also document this in the "required" section of https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/admin/metadatacustomization.rst#datasetfield-field-properties, but some guidance would be appreciated - thank you!

jggautier:
I can add something to that "required" row in the table at https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/admin/metadatacustomization.rst#datasetfield-field-properties, taking language from the release notes at https://github.com/IQSS/dataverse/pull/7608/files#diff-dbe06af7170aef8078a673760ffe3fb0bbc1a3038853782079ad8308c4567e07.

We mention in the [introduction section of the metadata customization guide](https://guides.dataverse.org/en/5.3/admin/metadatacustomization.html#introduction) how Dataverse collection admins can make dataset metadata fields required or not required. Will the Dataverse collection "General Information" page give collection admins the option to make certain fields conditionally required? I don't know if or how the UI's changed there, just that @TaniaSchlatter [mentioned a basic review of it](https://github.com/IQSS/dataverse/issues/7606#issuecomment-778432099). (I tried spinning up this branch on AWS to see what if any UI changes were made, but something about the process of spinning up a branch has been failing for me the last few times I tried.)

djbrooke:
@jggautier if you feel comfortable making the edits please feel free to do so. Thanks, and let me know if you have questions about formatting (or anything else, but you're more knowledgable about the metadata system than I am :)) and I'll try to help. 

Thanks also @sekmiller for the review! @scolapasta is back this week and he should have something to chew on here.

src/main/resources/db/migration/V5.3.0.3__7551-expanded-compound-datasetfield-validation.sql
scolapasta: Good catch; I think in all my testing I only had one child input level overridden.

src/main/resources/db/migration/V5.3.0.3__7551-expanded-compound-datasetfield-validation.sql
scolapasta: Right. I initially wrote these script to maintain the "status quo". But as producerName is going to conditionally required we either need this here or to change the release notes to  require reloading the tsv file.

sekmiller: I wasn't sure about the intent of making the update correspond to the tsv, so whichever makes the most sense.

jggautier:
Okay I'm done editing the docs. I don't have any practical experience with the new logic for compound fields, so what I added to the "required" row of the "datasetField (field) properties" table should be reviewed.

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
scolapasta: We can't just do that their because if the last if in this method (the one that has the comment "since we don't currently support the use case where the parent is required but no specific children are (the "true/false" case), we override this as false).

That said, it could probably skip some of the code right after by putting it in an else? I don't think it's especially critical, since the other thing that happens is it stores the value in the transient field, so would only get run at most once.

src/main/resources/db/migration/V5.3.0.3__7551-expanded-compound-datasetfield-validation.sql
scolapasta: OK, I decided to have them reload citation tsv to get the Producer Name change. (in this way we make sure they get the latest citation.tsv in case they decide to make any other changes).

sekmiller (approved):

PR #7613, opened 2021-02-16
7612 - fix(ingest): No more JNDI lookup fails during deploy

### LINKED ISSUE #7612: Error "Unsupported deployment descriptors element message-destination value null" during application deployment
When deploying recent Dataverse versions to Payara, I noticed the following error messages after server start, during the initial application deployment phase:

```
Unsupported deployment descriptors element message-destination value null
```

Looking around for similar problems on the internet, I found some old mailing list entries on these log messages regarding JMS.
(First Google results, e. g. [1](https://www.eehelp.com/question/dpl8007-unsupported-message-destination-element-descriptors-deployment/)).

After some investigation, it looks like using the `@Resource(name="")` syntax is not the best way to go. Sth. about referencing JNDI resources not physically located due to the new mechanism. Whatever - didn't dig deeper to understand every detail.

Looks like changing `@Resource(name="")` to `@Resource(lookup="")` might do the trick, but there is manual testing of Ingest needed.

PR forthcoming.

poikilotherm:
As we produce the JMS topic and factory for Ingest in
`IngestQueueProducer` since Dataverse 5.3, error messages
about unsupported missing message-destination appeared.
This commit fixes the underlying cause by using lookup()
instead of name() for the @Resource dependency injection.

Closes #7612

Needs manual testing of Ingest!

Usual setup. Try real ingest, so the JMS part is actually used. Not sure API tests already cover that.

None.

Nope.

**Additional documentation**:
None.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
@@ -136,9 +136,9 @@ public class IngestServiceBean {
     @EJB
     SystemConfig systemConfig;
 
-    @Resource(name = "java:app/jms/queue/ingest")
+    @Resource(lookup = "java:app/jms/queue/ingest")
     Queue queue;
-    @Resource(name = "java:app/jms/factory/ingest")
+    @Resource(lookup = "java:app/jms/factory/ingest")
     QueueConnectionFactory factory;
     
 


pdurbin (commented):
All the tests look fine at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-7613/4/testReport/ and I'm sure some ingest happens as part of them, so I'm pretty sure we can move this to QA. I didn't run the code to see if the offending line is gone from server.log.

I looked at the eehelp link and poked around a bit in javadoc and stackoverflow. I was hoping for more clear guidance on when "name" vs. "lookup" should be used. If we want, we could always add a line about this in our coding style page in the dev guide.

I noticed that `@Resource(name = "mail/notifyMailSession")` ("name" instead of "lookup") is still found in `MailServiceBean`. @poikilotherm do you think we should change this now (in this pull request) or later?

poikilotherm:
> I looked at the eehelp link and poked around a bit in javadoc and stackoverflow. I was hoping for more clear guidance on when "name" vs. "lookup" should be used. If we want, we could always add a line about this in our coding style page in the dev guide.

I wasn't able to find a good explanation in plain english. It's all a bit vague why we cannot use the name if we provide the resource via annotations.

> I noticed that `@Resource(name = "mail/notifyMailSession")` ("name" instead of "lookup") is still found in `MailServiceBean`. @poikilotherm do you think we should change this now (in this pull request) or later?

Certainly. May be done as part of #7424

pdurbin:
@poikilotherm good idea to defer the mail server work until that other issue.

Off to QA.

pdurbin (approved):

Issue #7565, opened 2021-01-29
Improve (simplify) Primefaces (Ajax) / Bootstrap interaction


scolapasta:
With Bootstrap, we do a lot of "binding" of html elements, with javascript, e.g. tooltips or popup previews for image files. This is done by calling a JS function `bind_bsui_components` when we ready the document.

With Primefaces, we do a lot of ajax calls that update parts of pages - when that happens the bootstrap binding is lost. This is because the update replaces the html, but doesn't know to call the rebinding.

We currently solve this by calling the javascript for each update with `oncomplete="bind_bsui_components"`

This is prone to error, because we have to remember to put it everywhere.

scolapasta:
So, I opened this issue because I tried an experiment that worked! I think it needs some more testing / refinement, but something of it's nature should eliminate the need to include `oncomplete="bind_bsui_components"` for all updates.

In `dataverse_template.xhtml`:
add:
```
      <p:fragment>
        <p:autoUpdate/>
         ...
     </p:fragment>
```
around the `<script>` at the end of the page.

That said, @mheppler are a little unsure if *why* it works, as we weren't sure if the document.ready would or should get called when that part of the page us updated. So it would be good to understand better why this is working and make sure there's no harm in calling other aspects.

(in doing some other research, there's another thing i want to try and will report back on shortly.

scolapasta:
OK, tried the other thing I had found:
https://api.jquery.com/ajaxComplete/

But it did not seem to work as I expected. More experimentation to be done, especially with the first solution, which seems to work.

scolapasta:
I keep seeing things like:
"The $(document).ready() event fires once, when the DOM has fully loaded from the initial load. An ajax event may change the DOM, but it doesn't reload the page, so the $(document).ready() won't fire again."

In the experiment above, though, we are effectively redefining the document.ready by updating the part of the page that contains it, and that does trigger it to fire. But I have yet to find any agreement (or disagreement) that this is an ok approach.

That said, I may have found a more "accepted" way.  Omnifaces has an `o:onloadScript` component that will get run after ajax calls:
https://stackoverflow.com/questions/14400624/jsf-primefaces-ajax-updates-breaks-jquery-event-listener-function-bindings
https://showcase.omnifaces.org/components/onloadScript

scolapasta:
So to be clear, I think this issue is ready to be worked on:
• add the onLoadScript
• clean up the oncompletes

mheppler:
[...]

PR #7627, opened 2021-02-17
Removed inline calls to js function, added it to rebind function

### LINKED ISSUE #5653: handle_dropdown_popup_scroll is not defined on `permissions-manage.xhtml`
On 1178-1a12cda

Screen Shot 2019-03-15 at 1 02 29 PM

Does not appear to prevent any functionality (was still able to modify user permissions on that page).  Have not investigated sufficiently to see if it's related to #5589 , but that would be plausible.  `` <script type="text/javascript" defer="defer" src="/javax.faces.resource/js/dv_rebind_bootstrap_ui.js.xhtml?version=4.10.1"></script>`` present in generated html

mheppler:

Resolves ReferenceError in console when `handle_dropdown_popup_scroll()` function is called to fix a scrolling bug with the autoComplete component dropdown menu, when it is inside a popup.

The function call has been moved to the `bind_bsui_components()` function which is properly called when the document is ready. The reason we were seeing the ReferenceError is because the inline function call added under the autoComplete was being called before the javascript with the function is loaded by the page.

Closes #5653 handle_dropdown_popup_scroll is not defined on `permissions-manage.xhtml`

Might need to hold SHIFT and reload the page a few times to clear the browser cache and get a fresh copy of the dv_rebind_bootstrap_ui.js file from the server with the new code fix.

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dashboard-datamove.xhtml
+++ b/src/main/webapp/dashboard-datamove.xhtml
@@ -61,9 +61,6 @@
                                             <p:ajax process="@this" event="itemSelect" />
                                             <p:ajax process="@this" event="itemUnselect" />
                                         </p:autoComplete>
-                                        <script>
-                                            handle_dropdown_popup_scroll();
-                                        </script>
                                         <p:message for="sourceDatasetMenu"/>
                                     </div>
                                 </div>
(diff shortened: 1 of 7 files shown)


scolapasta (approved):
Seems pretty straightforward change. (I am always a little concerned about putting everything in this one file, in case it causes any performance - for pages that don't need it - but if we see that, we can try to rethink our strategy here.

mheppler:
> Seems pretty straightforward change. (I am always a little concerned about putting everything in this one file, in case it causes any performance - for pages that don't need it - but if we see that, we can try to rethink our strategy here.

As I mentioned in my last comment on the original issue #5653...

>This will be a good feature to revisit when working on Improve (simplify) Primefaces (Ajax) / Bootstrap interaction #7565.

PR #7628, opened 2021-02-18
3175 edit dvpage bug

### LINKED ISSUE #1198: Dataverse General Information: Metadata Elements, Can Uncheck the open set of metadata elements and it stays open
Open metadata elements->uncheck->remains open and the close option disappeared

### LINKED ISSUE #3175: Metadata Fields > Edit Dataverse General Information: "Done" btn is broken on new dataverse form only, when viewing metadata fields
Steps to reproduce:
1. Create a dataverse, do not yet save.
2. On any of the metadatablocks, under Metadata Fields, click View Fields.
3. Click Done and notice the block remains expanded, cannot close, and now cannot view any other blocks.

Refreshing the page does restore functionality.

This was reported by a user, see RT 235672

sekmiller:
On create DV if the user went into custom metadata blocks before filling in the required "alias" and "category" fields the "Done" button to close the metadata block view wouldn't work. (also there was no indication why 'hey dummy, put in an alias for your new dv')
Also fixed an older bug where you could edit the input levels of an optional mdb and then deselect it.

Closes #3175 
Closes #1198

@mheppler I used "jsf:disabled" for the 1198 fix. It seems to work but it's not used anywhere else. Kosher?

**Additional documentation**:


src/main/webapp/dataverse.xhtml
```diff
@@ -330,8 +331,9 @@
                                                                             </div>
                                                                         </td>
                                                                         <td>
-                                                                            <p:selectOneRadio id="OptionsRadio#{dsft.id}" value="#{dsft.requiredDV}" widgetVar="OptButton#{dsft.id}" disabled ="#{DataversePage.editInputLevel == false}"
-                                                                                              rendered="#{!dsft.hasChildren and !((!dsft.hasChildren  and dsft.required) or (dsft.hasChildren and dsft.hasRequiredChildren))}">
+                                                                            <p:selectOneRadio id="OptionsRadio#{dsft.id}" value="#{dsft.requiredDV}" widgetVar="OptButton#{dsft.id}" disabled ="#{DataversePage.editInputLevel == false}" 
+                                                                                              onclick="#{DataversePage.toggleInputLevel(mdb.id, dsft.id)}"
```
scolapasta, line 335: what is this onclick doing? onclick is for calling javascript, but what you have here is a backing bean (java) call, so I don't think this would be being called when you click?

scolapasta (changes requested):
I selected "request changes" instead of "comment" because I think the line I commented on will need to change; I could be wrong...

src/main/webapp/dataverse.xhtml
sekmiller: I put it in because in my pre-testing I thought I had seen a case where changes to required/optional weren't being saved and because the method in the backing bean wasn't being used anywhere. I thought they were related so I put in the onclick. After removing both it seems to work so I'll chalk it up to bad pre-testing.

src/main/webapp/dataverse.xhtml
```diff
@@ -287,6 +287,7 @@
                                                         <p:ajax update="@widgetVar(optionBlock)" />
                                                     </input>
                                                     <input type="checkbox" jsf:itemValue="#{mdb}" id="#{mdb.idString}" jsf:value="#{mdb.selected}"
+                                                           jsf:disabled="#{DataversePage.editInputLevel == true}"
```
mheppler, line 290: Does this still work without the `jsf:` passthrough? I would imagine it would. The other sibling basic HTML input's above it do not have it on their disabled attributes.

src/main/java/edu/harvard/iq/dataverse/DataversePage.java
```diff
@@ -511,18 +511,6 @@ public void hideDatasetFieldTypes(Long mdbId) {
         }
         setEditInputLevel(false);
     }
-    
-    public void toggleInputLevel( Long mdbId, long dsftId){
```
scolapasta, line 515: Another question: this is being removed, but I don't see a corresponding removal if it being called? Was it never used (or am I just missing it)?

src/main/webapp/dataverse.xhtml
sekmiller: It doesn't work without the jsf: because if the disabled tag is there even if it contains "false" it disables the component. The others on lines 284 and 286 have disabled="disabled" and the real decision work is done on the jsf:rendered.

src/main/java/edu/harvard/iq/dataverse/DataversePage.java
sekmiller: That was what was being called in the (now removed) onclick on the option buttons and nowhere else.

scolapasta (approved):

PR #7631, opened 2021-02-23
7527 redetect api error

### LINKED ISSUE #7527: Error with Redetect File Type API
Hello,

This issue follows a discussion in the google group :
https://groups.google.com/g/dataverse-community/c/_H8ZdAo85BU

Here is an example to describe the problem :

Dataverse version : 5.0 + S3 storage
file extension saved in S3: .tabular
current MIME type for this file: "application/octet-stream"
The .tabular extension is declared in the MimeTypeDetectionByFileExtension.properties file => tabular = text/tab-separated-values

Here is the problem :
When the redetect API resource is called, because the file is remote, its content is inserted into a temporary file: tempFileTypeCheck.tmp
The file extension is then compared to the list in MimeTypeDetectionByFileExtension.properties but the .tmp is not there.
Server return: "tmp is a file extension Dataverse doesn't know about. Consider adding it to the MimeTypeDetectionByFileExtension.properties file."

Finally, the "application/octet-stream" MIME Type is the result of the redetect API resource for this file :(
The expected result is "text/tab-separated-values"

thanks a lot.
Steven.

landreev:

See the discussion on slack for what's going on in this PR. 

Closes #7527 (maybe?)

See the discussion on slack. 

There are 2 parts of the issue. The first one is trivial - the filename/extension were not being taken into account by the redetect API; this is fairly straightforward to test. 

The second part, the strange EJB rollback, is completely counter-intuitive. I have no idea how to test it, seeing how I haven't been able to reproduce it outside of prod. For example, the API consistently fails in prod. with this EJB error for the following datafile:

```
curl -X POST "http://localhost:8080/api/files/4325124/redetect?key=..."
```

however, it's working perfectly fine for this id on vm-5, which is also running 5.3 on a copy of the prod. db. 

It doesn't happen for *any* file in prod. either. It was in fact a matter of considerable luck/coincidence that when the remote installation reported the error, Danny tried the API on a random file in our archive, and got the same exact error they were getting. (But at least this means that the condition is not limited to just our server). 

I didn't have any other way to fix/investigate this than by deploying a modified  5.3 in prod. on a weekend. Once again, I'm not sure how to QA this. 

(edit: I will see if I can find a datafile that is similarly failing on vm-5, by trial and error)

**Additional documentation**:


scolapasta (changes requested):
The specific comments here are small. The general thing is about whether the API class should be an ejb or not. I wrote more extensively in slack (and we are still discussing) but the TLDR is: API classes should operate similar to backing beans and NOT be ejbs. (I think)

donsizemore:
Just a note that Odum has hit this problem in certain of its datafiles. I can provide more information about our datafiles and filemetadata if it would be helpful, but the results are identical to those described by Mr. Ferey.

kaitlinnewson:
Hi all, what is the status of this PR? We are also having this issue and would love to be able to clean up some of our files marked as "Unknown" formats.

landreev:
I will close this PR in favor of #8835 that fixes the file name issue. 
I am concerned about the second part of #7527 - the mysterious TransactionRolledbackLocalException that's triggered by *some* specific files on some installations. See the description of this PR for more info. 

I just checked and it is still happening in our prod., for that specific file (id 4325124). It would be nice to figure out what is happening there. I will open a new issue for just that part, and then close this PR. My gut feeling is that it has got nothing to do with the actual functionality in the redetect command.

landreev:
Opened #9142 to investigate the remaining mystery part (the selectively-happening weird EJB rollback). Closing this.

Issue #7588, opened 2021-02-05
API calls such as Datasets editMetadata returns blank response


qqmyers:
calling the /api/datasets/\<id\>/editMetadata endpoint with bad credentials results in a 500 error with a blank response. It looks like this is caused by the WrappedResponse from an Authentication failure not getting unwrapped and therefore getting caught as a Throwable in the ThrowableHandler, which treats it as a 500 error.

This is a common pattern in the API - has something changed where api methods that throw WrappedResponses are no longer being handled properly?

qqmyers:
Should have been closed by #7641

PR #7641, opened 2021-02-25
properly report error from editVersionMetadata

qqmyers:
**What this PR does / why we need it**: This PR unwraps a WrappedResponse to make the call properly return a useful error/proper status code if/when the user can't be found (e.g. an invalid apikey)

Closes #

**Special notes for your reviewer**: This issue appears to be one of a few related ones where API calls throw WrappedResponse. That doesn't work. There is a pattern where AbstractAPIBean handles the unwrapping (see [createPrivateUrl()](https://github.com/IQSS/dataverse/blob/3600a0e3498f767473e5e473a67956661c029c70/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1217-L1223) for example, but some calls through the WrappedResponse without using it, and those are broken. The PR fixes one of those. If that's the way to fix others, I can scan for the rest, but if there's an alternate approach, it would be good to define it before doing the rest. Due to this, I haven't added #7588 in the Closes section above yet.

**Suggestions on how to test this**: Send a bad api key to the /editMetadata method.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -805,11 +805,16 @@ public class Datasets extends AbstractApiBean {
     
     @PUT
     @Path("{id}/editMetadata")
-    public Response editVersionMetadata(String jsonBody, @PathParam("id") String id, @QueryParam("replace") Boolean replace) throws WrappedResponse{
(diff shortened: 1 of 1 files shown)


pdurbin (commented):
This looks like a good fix. As you've indicated, the method should not throw WrappedResponse, as it was previously. WrappedResponse should be handled within the code.

scolapasta (approved):
Discussed with @qqmyers - moving this through now, but we'll have a general discussion at tech hours to determine future direction.

PR #7642, opened 2021-02-26
restricted summary stats #7619

### LINKED ISSUE #7619: No Summary Stats and Variable Names/Labels for Restricted Files
As part of the OpenDP project, we're adding an integration with the Dataverse Software. Generally, the idea is that a data depositor or curator would use the OpenDP Tool to create a DP metadata release for a restricted file, and then deposit that DP metadata release back to a Dataverse installation as an auxiliary file using the APIs added in 5.3. The DP release would then be accessible for download or use by external tools as appropriate (#7400). The idea is that a researcher would be able to use this DP metadata release to determine whether or not it's worthwhile to request access to the file. 

The big issue with this workflow is that we already expose Summary Stats in the DDI Export, so if the non-DP version of the Summary Stats remain available, people wouldn't bother creating the DP release, and if they did, privacy would be spoiled anyway. Not great! 

We previously discussed this in #6474 and added some documentation to clarify the current behavior, and we mentioned a few times that we'd have to revisit this availability once we started to support sensitive data. Looks like it's time to revisit. 

I'd welcome feedback on the best approach for this, but generally I'm thinking we:

- Remove the export options that expose summary stats and variable names/labels for restricted files (UI and API) for unauthorized users
- Make these summary stats and variable names/labels unavailable through other APIs for unauthorized users
- Add a release note, with a note to run re-exportall

pdurbin:

In the future we want to offer differentially private summary statistics for restricted data which means that restricted data should not present full summary statistics (and related information). This pull request corrects this.

Closes #7619

Not particularly. See the release note and the code. I'm weirded out by the existing BadRequestException error handling in Access.java and WebApplicationExceptionHandler.java. I just accepted the pattern of putting a string in the former and the real message in the latter.

Please see the release note, especially:

- At the dataset level, DDI exports no longer show "dataDscr" information for restricted files. There is only one version of this export and it is the version that's suitable for public consumption with the "dataDscr" information hidden for restricted files.
- Similarly, at the dataset level, the DDI HTML Codebook no longer shows "dataDscr" information for restricted files. 
- At the file level, "dataDscr" information is no longer publicly available for restricted files. In practice, it was only possible to get this publicly via API (the download/access button was hidden).
- At the file level, "dataDscr" (variable metadata) information can still be downloaded for restricted files if you have access to download the file.

The DDI HTML Codebook will no longer show for restricted files summary statistics, names of variables, or descriptions of variables. Instead, you will only see the count of rows and columns, type of file, and the UNF like this:

Screen Shot 2021-02-26 at 10 56 36 AM

Yes, included.

**Additional documentation**:

Documentation was actually removed because we used to warn people (pull request #6620) about summary statistics being publicly available for restricted data. Now we don't need to.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
@@ -47,7 +47,7 @@ public class WebApplicationExceptionHandler implements ExceptionMapper<WebApplic
                 if ( (ex.getMessage()+"").toLowerCase().startsWith("tabular data required")) {
                     jrb.message(BundleUtil.getStringFromBundle("access.api.exception.metadata.not.available.for.nontabular.file"));
                 } else if ((ex.getMessage() + "").toLowerCase().startsWith("no permission to download file")) {
-                    jrb.message("You do not have permission to download this file.");
+                    jrb.message(BundleUtil.getStringFromBundle("access.api.exception.metadata.restricted.no.permission"));
                 } else {
                     jrb.message("Bad Request. The API request cannot be completed with the parameters supplied. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.");
                     jrb.request(request);
--- a/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
@@ -43,8 +43,11 @@ public class WebApplicationExceptionHandler implements ExceptionMapper<WebApplic
         switch (ex.getResponse().getStatus()) {
             // BadRequest
             case 400:
+                // It's strange to have these "startsWith" conditionals here. They both come from Access.java.
                 if ( (ex.getMessage()+"").toLowerCase().startsWith("tabular data required")) {
                     jrb.message(BundleUtil.getStringFromBundle("access.api.exception.metadata.not.available.for.nontabular.file"));
+                } else if ((ex.getMessage() + "").toLowerCase().startsWith("no permission to download file")) {
+                    jrb.message("You do not have permission to download this file.");
                 } else {
                     jrb.message("Bad Request. The API request cannot be completed with the parameters supplied. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.");
                     jrb.request(request);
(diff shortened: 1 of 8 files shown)


src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
```diff
@@ -43,8 +43,11 @@ public Response toResponse(WebApplicationException ex) {
         switch (ex.getResponse().getStatus()) {
             // BadRequest
             case 400:
+                // It's strange to have these "startsWith" conditionals here. They both come from Access.java.
                 if ( (ex.getMessage()+"").toLowerCase().startsWith("tabular data required")) {
                     jrb.message(BundleUtil.getStringFromBundle("access.api.exception.metadata.not.available.for.nontabular.file"));
+                } else if ((ex.getMessage() + "").toLowerCase().startsWith("no permission to download file")) {
+                    jrb.message("You do not have permission to download this file.");
```
scolapasta, line 50: Should this be in a bundle?

scolapasta (commented):
Overall looks good. Once thing that is worth discussing - I had orginally assumed we were hiding summary stats, but it looks like it's the entire data description section, including variable names/labels. However we still index them and make them searchable. Should we be hiding those in the index similarly? Or should we consider leaving them here and just removing summary stats? (if the former, should we consider in scope or out of scope)?

src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
pdurbin: Good idea. Done in 9813c77.

pdurbin:
> I had orginally assumed we were hiding summary stats, but it looks like it's the entire data description section, including variable names/labels. However we still index them and make them searchable.

Good catch. Here's the data I'm testing with (these are my pets, yes my cats are huge)...

```
name,pounds,species
Marshall,40,dog
Tiger,17,cat
Panther,21,cat
```

... and variable names like "species" are still searchable:

Screen Shot 2021-02-26 at 2 32 44 PM

djbrooke:
Thanks @pdurbin and @scolapasta. It's OK if we keep these in the index for now. We'll possibly need to revisit this as we support more sensitive data, but for the OpenDP MVP this will support the use case of not spoiling privacy (and making the tool useful in the first place :)). I'll move this to QA since it looks like the other feedback in the issue (regarding the move to bundle) is done.

kcondon:
@pdurbin Things I've noticed:
1. When first attempted to download DDI HTML on restricted file via ui, saw 404 error. After a couple more attempts it worked. Not sure whether that is a timing issue in export. DDI worked.
2. Am not able to download variable metadata on restricted file though I pass api key for creator (admin). I am using the view file metadata endpoint.

Otherwise it appears to work as described. I can try testing 1. again.

PR #7651, opened 2021-03-04
7374 postgresql update

### LINKED ISSUE #7374: Upgrade to PostgreSQL 13+
In the present [Installation Guide](https://guides.dataverse.org/en/latest/installation/prerequisites.html#)  there is a strong recommendation of using version PostgreSQL 9.6 which reaches End-of-Life 2021-11-01, while the latest version is already at v.13. Evaluate risk of obsolescense.
Together with PostgresSQL 9.6 goes use of RHEL/CentOS 7, which is now only in maintanence support 2, full support ended 2019-08-06, while latest version RHEL/CentOS 8 was released already 2019-05-07 - more than a year ago.

landreev:

PostgreSQL 9.6 is nearing its end of life (in Nov. 2021). This PR removes the language from the installation guide that specifies 9.6 as as the "strongly recommended" version, saying instead that the application has been tested with up to v. 13. It also tells the existing installations, via a release note, that it is ok to upgrade to a later version up to that. 

Closes #7374

This is not a required upgrade. In this PR we are telling them to knock themselves out and upgrade to whatever they want, if they so desire. But they don't have to really. 
A basic quick test is just to confirm that the application still functions under the latest Postgres (13.2 as of now). 
I've tested upgrading and converting a copy of our prod db to later versions (that was the time-consuming part). I figured that if our database can be upgraded, then it should work for any other installation too. 

**Additional documentation**:


doc/sphinx-guides/source/installation/prerequisites.rst
```diff
@@ -98,25 +98,16 @@ PostgreSQL
 Installing PostgreSQL
 =======================
 
-Version 9.6 is strongly recommended because it is the version developers and QA test with::
+The application has been tested with PostgreSQL versions up to 13. We recommend installing the latest version that is available for your OS distribution. *For example*, to install PostgreSQL 13 under RHEL/CentOS 7::
```
pdurbin, line 101: We should update the dev guide too. Right now it starts with, "PostgreSQL 9.6 is recommended to match the version in the Installation Guide."

I also think we should update Vagrant and docker-aio but especially since the Solr pull request touches both and hasn't been merged, it makes sense to wait.

pdurbin (commented):
We should probably update the dev guide too.

doc/sphinx-guides/source/installation/prerequisites.rst
landreev: I'll definitely update the dev. guide, thanks. 
Still not sure about touching the vagrant and docker-aio in this PR. I'd rather have somebody who uses/works on those extensively handle that. And I don't see a problem with doing that separately/later. I mean, everything there is going to keep working just fine with PG 9.6; touching/changing anything there on the other hand comes with the risk of breaking something in that fragile ecosystem.

scolapasta (approved):
New issue was created for vagrant and docker.

PR #7654, opened 2021-03-04
add back variable indexing #6545

### LINKED ISSUE #6545: Solr search and variable level metadata

scolapasta:
This adds back indexing of variable labels and fields added via the Data Curation Tool

Closes #6545

The change was just to reduce the number of queries from one per variable to 1 (then put that into a Map).

Test and compare performance of indexing, especially index all. Indexing should take a little longer than current (because) it is indexing more, but a reasonable percentage more. In my testing, for example, 1 dataset took just under a second before this change and just over a second after (with the original variable metadata indexing it took 10 seconds).

No

Yes, installation should reindex to catch variable metadata (the plan is to get this into the same release as Solr upgrade).

**Additional documentation**:


sekmiller (approved):
Looks reasonable using an existing query and map for improved performance.

Issue #7387, opened 2020-11-04
Distributor tag automatically added in the OAI-DDI and DDI export


bappun:
A distributor element is automatically added in the OAI-DDI and DDI export.
In our case, "data.sciencespo" is added alongside "Sciences Po, Centre de données socio-politiques (CDSP), CNRS (CDSP)".

We have this in our [dataset](https://data.sciencespo.fr/dataset.xhtml?persistentId=doi:10.21410/7E4/00LYOG):
![image](https://user-images.githubusercontent.com/24225471/98123881-c13dd900-1eb2-11eb-91f2-425de61390fb.png)

In the OAI-DDI and DDI export we have:

- Inside `<docDscr>`:

```
<distrbtr source="archive">data.sciencespo</distrbtr>
```

- Inside `<stdyDscr>`:

```
<distrbtr source="archive">data.sciencespo</distrbtr>
<distrbtr abbr="CDSP">Sciences Po, Centre de données socio-politiques (CDSP), CNRS</distrbtr>
```

pdurbin:
@bappun well if you want to make a pull request, here's the first place where `distrbtr` appears in the code: https://github.com/IQSS/dataverse/blob/v5.1.1/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java#L199

I'm a little confused though. Is it a bug? Is it a preference? Should it be configurable? Is it required by the DDI spec? That's where my thoughts are. 😄

jggautier:
Just describing my understanding of the way these fields are used in Dataverse now, and more importantly asking you how you think those fields **should** be used:

The Citation metadatablock's Distributor field is intended to identify who is distributing the data (as opposed to the metadata), and in the DDI Codebook XML export, that Distributor field is mapped to stdyDscr/distrbtr. It's meant to say, the data/study itself is distributed by whatever's entered in the Distributor field, e.g. "Sciences Po, Centre de données socio-politiques (CDSP), CNRS".

A second distrbtr element is used to say that the data is also distributed by the repository. The source attribute is used to signal that "data.sciencespo" is the "archive" charged with storing the data.

All of the metadata in the XML's docDscr section, like docDscr/distrbtr, describes the metadata document itself. The reasoning is that since the repository is distributing the metadata export, the name of the Dataverse-based repository is hardcoded into that part of the DDI XML export. I think someone in the repository's database settings, "data.sciencespo" is set as the name of the repository, and it used in multiple places to identify the repository.

[...]

PR #7657, opened 2021-03-05
@Singleton util classes

### LINKED ISSUE #7649: Replace static calls with @Inject-able classes
There are a number of classes in Dataverse software that consist of static methods (JsonPrinter, MailUtil, and BrandingUtil in particular) where work-arounds have been made to allow settings to be accessed by the class (e.g. allowing JsonPrinter to use the :ExcludeEmailFromExport) or avoid using other services (e.g. sending the root dataverse collection name into BrandingUtil and MailUtil methods as a parameter, or finding the root name by following the chain of owners in DatasetVersion and JsonPrinter - versus calling dataverseService.findRootDataverse() as in other places.)

In addition to these current cases, the work proposed in #7387 will involve accessing settings in JsonPrinter and there is similar work @ QDR to use an 'institution name' instead of the root collection name in some cases that has avoided the issue by storing the institution name in the Bundle.properties file instead of as a setting.

After discussion with @scolapasta, rather than continuing to work around this issue, I've started working to replace the classes containing static methods with an @Stateless class with non-static methods that can then be @Injected where needed. Since this involves fairly straight forward changes across lots of classes, I'm going ahead to create a separate issue/PR. Nominally this PR should not change outward functionality - it does ~code cleanup and prep work for the work in #7387 (which I hope can then be reused by QDR to reduce the custom code in this area).

From some web searching, the general idea of replacing static methods with @Stateless beans seems to be a reasonable/good practice. If anyone know of a reason we shouldn't do this in the Dataverse software, let me know. Also - if there are other cases where the same refactoring should be done - let me know - I can either look into doing those or at least adding some ToDo notes pointing to the classes above as examples.

qqmyers:
**What this PR does / why we need it**: Makes BrandingUtil and JsonPrinter @Singleton beans that can then inject the services they require instead of having to have callers send the services or the results of the calls (such as the root dataverse name) to the server as method parameters. This is basically clean-up but also preparation for #7387.

Most of the PR is making the simplifications possible made possible by that - i.e. dropping method params that aren't needed. One additional change was to remove local code to find the rootDataverseName in DatasetVersion and JsonPrinter that were only used to get the name so it could be passed to BrandingUtil.

Also added the same design to OREMap to avoid having to send settingsService through the ExportService.

There are also changes to update tests - removing parameters and adding mock services.

Closes #7649

**Special notes for your reviewer**: FWIW: I think the lack of mocks in tests initially made me think that I couldn't keep static methods, but injecting a static service does work and I was able to avoid having to inject the util classes everywhere (as discussed in the issue)

**Suggestions on how to test this**: This should change anything w.r.t. functionality, and it shouldn't add new functionality (a PR for #7387 needs the changes here and will add new functionality). The one potential issue I can see would be that some calls to get the root dataverse name that were done by traversing the chain of owners are now replaced with a call to find the root Dataverse from the db. If that results in a measurable performance difference, adding some simple caching to the call to get the name could be done. (I didn't want to add that to this PR without reason, but it could be added here or as a quick follow-on PR.)

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -195,6 +196,19 @@ public Dataverse findRootDataverse() {
...
     }
     
+    
+    //Similarly - if the above throws that exception, do we need to catch it here?
+    //ToDo - consider caching?
+    public String getRootDataverseName() {
+        Dataverse root = findRootDataverse();
+        if(root==null) {
+            logger.warning("Failed to find root dataverse!");
+            return "";
```
scolapasta, line 206: if no root dataverse, should we throw an exception instead? this seems like something that won't happen, but if it does, should be ignored.

src/test/java/edu/harvard/iq/dataverse/branding/BrandingUtilTest.java
```diff
@@ -1,32 +1,46 @@
...
+    
     @Test
     public void testGetInstallationBrandName() {
-        System.out.println("testGetInstallationBrandName");
-        assertEquals("LibraScholar", BrandingUtil.getInstallationBrandName("LibraScholar"));
-        assertEquals(null, BrandingUtil.getInstallationBrandName(null));// misconfiguration to set to null
-        assertEquals("", BrandingUtil.getInstallationBrandName(""));// misconfiguration to set to empty string
+        
+       System.out.println("testGetInstallationBrandName");
+        assertEquals("LibraScholar", BrandingUtil.getInstallationBrandName());
```
scolapasta, line 26: what is the point of this test?

scolapasta (commented):
Overall, looks as we discussed. Just had one comment and one question.

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
qqmyers: Hmm - Or is it needed at all? - findRootDataverse should throw an exception if it isn't found - can it return null?

scolapasta: Yeah, that's what I meant; let it throw the exception. (it will throw a NoResultException)

src/test/java/edu/harvard/iq/dataverse/branding/BrandingUtilTest.java
qqmyers: It checks that the mock dataverseService is being seen within BrandingUtil and responding with 'LibraScholar'. FWIW: I think it works because this test runs after the next one which has line 38 to set the mock correctly. I can look at adding a redundant call here to make sure it works/be clearer. (Also note that there will be more subtests here when there's a installation.name setting that could change the output of this method.)

scolapasta: yeah, these tests like these feel basically useless to me; i.e. it's effectively testing the mock service, rather than using the mock service to test something real. I'd vote to remove it altogether.

qqmyers: removed

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
qqmyers: "null is not an option" (TM)

qqmyers:
@scolapasta - FYI - I made similar changes to avoid settingsService getting passed through the ExportService since you last looked at this. With that I think this is done.

scolapasta (approved):

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -1,22 +1,37 @@
...
+import javax.ejb.Singleton;
 import javax.mail.internet.InternetAddress;
 
+@Singleton
 public class BrandingUtil {
 
-    public static String getInstallationBrandName(String rootDataverseName) {
-        return rootDataverseName;
+    @EJB
+    static DataverseServiceBean dataverseService;
```
poikilotherm, line 15: I didn't even know this is possible. It doesn't feel very beanish, to be honest.

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -1,22 +1,37 @@
...
+import edu.harvard.iq.dataverse.DataverseServiceBean;
 import edu.harvard.iq.dataverse.util.BundleUtil;
 import java.util.Arrays;
+
+import javax.ejb.EJB;
+import javax.ejb.Singleton;
 import javax.mail.internet.InternetAddress;
 
+@Singleton
 public class BrandingUtil {
```
poikilotherm, line 12: Should we rename to BrandingBean as this is a bean now?

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -1,22 +1,37 @@
...
 
-    public static String getInstallationBrandName(String rootDataverseName) {
-        return rootDataverseName;
+    @EJB
+    static DataverseServiceBean dataverseService;
+    
+    public static String getInstallationBrandName() {
```
poikilotherm, line 17: Is there a particular reason why this is still all static methods? This is an injectable bean now, so does it matter?

If we still want to have static methods, maybe we can do the injection the same way I did with `DBSettingsSource`? (A helper bean, injecting the service on construction into the POJO)

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -1,22 +1,37 @@
...
+        //ToDo #7387 which will make this call return something different than getRootDataverseCollectionName() 
+        return dataverseService.getRootDataverseName();
+    }
+    
+    //Convenience to access root name without injecting dataverseService (e.g. in DatasetVersion)
+    public static String getRootDataverseCollectionName() {
+        return dataverseService.getRootDataverseName();
```
poikilotherm, line 24: Shouldn't we do this more consistent then and go for changing the function in `DataverseServiceBean`, too? (I am aware of the consequences)

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -31,8 +46,11 @@ public static String getSupportTeamEmailAddress(InternetAddress systemAddress) {
         return systemAddress.getAddress();
     }
 
-    public static String getContactHeader(InternetAddress systemAddress, String rootDataverseName) {
-        return BundleUtil.getStringFromBundle("contact.header", Arrays.asList(getSupportTeamName(systemAddress, rootDataverseName)));
+    public static String getContactHeader(InternetAddress systemAddress) {
+        return BundleUtil.getStringFromBundle("contact.header", Arrays.asList(getSupportTeamName(systemAddress)));
     }
 
+    public static void injectDataverseService(DataverseServiceBean dataverseService) {
```
poikilotherm, line 53: This is bad style IMHO. If this is to be a bean, add a constructor to have explicit dependency injection in a non-static field. At the bare minimum, this should not be a `public` function (IMHO package private, as used solely for testing). Or, as the field is package private already, tests could just set the field directly.

poikilotherm (changes requested):

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
qqmyers: I've gone back to using a helper - I don't think it was working correctly (but a little hard to tell given other problems - to be discussed elsewhere). W.r.t. static methods - a) there is no state, and b) since there's no discovery mechanism as with the ConfigSources, it would mean making this a bean (it isn't one with the helper) and injecting many places. There's a closed PR where I started looking into how much would have to change for that - this looked better.)

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
qqmyers: reverted to using a helper.

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
qqmyers: no longer a bean.

scolapasta (approved):
Discussed with @qqmyers. Longer term we want to get rid of the Helpers and make these Stateless EJBs, but for this contains several big steps getting us there AND unblocks SciencesPO work. So we'll move forward and create an issue for the next steps.

kcondon:
Suggested regression testing places were:m where root dv name is fetched and excluding email from export. Have found email exclude is not working. root name is.

PR #7661, opened 2021-03-06
Custom Installation Name, Optional Use as DDI Distributor

### LINKED ISSUE #7387: Distributor tag automatically added in the OAI-DDI and DDI export

qqmyers:
**What this PR does / why we need it**: Adds a new setting that will be used instead of the Root Dataverse Collection Name for the Brand Name (BrandingUtil.getInstallationBrandName() ) if set and a second setting that avoids adding the brand name as a Distributor in the DDI export studyDsc if the Distributor is already set in the metadata (part of the Citation block).

Closes #7387 

**Special notes for your reviewer**: Per discussions/issue comments:

- This PR changes the broadly used getInstallationBrandName method so if the new setting is used, the change will occur everywhere that is called (primarily in metadata exports and email notifications). If there are any places where this isn't desired, there is a BrandingUtil.getRootDataverseName() call that can be used instead.

- One new setting is implemented as an MPconfig variable that also has a settings alias. 
  - The variables and setting name (in microprofile-aliases.properties, better name welcome) are: 
```
dataverse.export.distributor.excludeinstallationifset=dataverse.settings.fromdb.ExportInstallationAsDistributorOnlyWhenNotSet

```
(Note that #7680 prevented doing this for 
```
dataverse.branding.installation.name=dataverse.settings.fromdb.InstallationName
```
which is just a standard setting in the final PR.)

  - In trying to get this to work, (heads up to @poikilotherm ) I discovered [three bugs in the DbSettingsConfigSource class](see [this commit](https://github.com/IQSS/dataverse/commit/0df20d6113fd73da43cdf74056d86e72b9d6010e) and [this one](https://github.com/IQSS/dataverse/pull/7661/commits/3238a5638cc221238a8470021e8e9acf3d6e416f)) which could be considered in a separate PR if needed.
  - As far as I can tell, this is the first use of microprofile vars in the code / the first use of the new DbSetting and Alias ConfigSources. FWIW, I ended up choosing the programmatic approach versus injecting the variable itself - I wasn't able to figure out how to make that work (possibly due to testing using a db setting (see BrandingUtilTest) and therefore relying on the DbSettingConfigSource which requires some initialization.)
  - If we want to use this approach to support incremental change in the code, one additional change might be useful - to switch the warning in AliasConfigSource when an alias is used to use logger.fine() from logger.warning() (at least for now. If at some point we recommend not using the db settings, we could make the warning more prominent.)

-  This PR depends on #7657, so that should be tested/merged first.
- The basic functionality is probably easiest to test by setting the :InstallationName and :ExportInstallationAsDistributorOnlyWhenNotSet properties. The first should override the use of the root dataverse collection name in emails and metadata exports (see #7387 for a list - in a comment near the end). The second should only affect whether the DDI exports (DDI, DDI HTML Codebook, OpenAire) have an extra stdyDsc distr element when the dataset metadata already has a Distributor entry (true -> only the value from the metadata appears) (It's the contributor contributorType="Distributor" element in OpenAire). (Note that the doc dstr in the DDI export will still use the Installation Name regardless, so make sure you're checking the right element.) 
- I assume, but have not tested that you can also set dataverse.export.distributor.excludeinstallationifset via micro-profile directly (perhaps @poikilotherm can provide a suggestion).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: only to the text shown where the brand name previously was always the root dataverse collection name.

**Is there a release notes update needed for this change?**: just noting the new options.

**Additional documentation**: documented the new settings in the guide


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -594,7 +594,7 @@ You may also want to look at samples at https://github.com/shlake/LibraDataHomep
 A simpler option to brand and customize your installation is to utilize the Dataverse collection theme, which each Dataverse collection has, that allows you to change colors, add a logo, tagline or website link to the Dataverse collection header section of the page. Those options are outlined in the :doc:`/user/dataverse-management` section of the User Guide.
 
 Custom Homepage
-++++++++++++++++
++++++++++++++++
```
qqmyers, line 597: FWIW: I think my rst plugin is catching these now. I just added the two entries at the bottom.

djbrooke:
Thanks @qqmyers. Just a quick question - for documentation purposes, if an installation chooses update the setting, will they need to run reexportall to realize the changes?

qqmyers:
@djbrooke - good catch - added notes in the docs.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
```diff
@@ -195,7 +197,15 @@ private static void createStdyDscr(XMLStreamWriter xmlw, DatasetDTO datasetDto)
...
-        if (datasetDto.getPublisher() != null && !datasetDto.getPublisher().equals("")) {
+      //The default is to add Dataverse Repository as a distributor. The excludeinstallationifset setting turns that off if there is a distributor defined in the metadata
+        boolean distributorSet=false;
+        MetadataBlockDTO citationDTO= version.getMetadataBlocks().get("citation");
+        if(citationDTO!=null) {
+            if(citationDTO.getField(DatasetFieldConstant.distributorName)!=null) {
+                distributorSet=true;
+            }
+        }
+        if (!StringUtils.isEmpty(datasetDto.getPublisher()) && !(ConfigProvider.getConfig().getOptionalValue("dataverse.export.distributor.excludeinstallationifset", Boolean.class).orElse(false) && distributorSet)) {
```
poikilotherm, line 208: This is DDI specific, so IMHO we should use `dataverse.export.ddi.distributor.excludeinstallationifset`

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -2224,3 +2224,13 @@ This is the local file system path to be used with the LocalSubmitToArchiveComma
 +++++++++++++++++++
 
 These are the bucket and project names to be used with the GoogleCloudSubmitToArchiveCommand class. Further information is in the :ref:`Google Cloud Configuration` section above.
+
```
poikilotherm, line 2227: The below is missing the MicroProfile Config names. IMHO we should provide those, otherwise they are likely to stay unknown.

FWIW: there is an example for the database connections in the upper config section.

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
```diff
@@ -1,22 +1,39 @@
...
+@Singleton
 public class BrandingUtil {
 
-    public static String getInstallationBrandName(String rootDataverseName) {
-        return rootDataverseName;
+    @EJB
+    static DataverseServiceBean dataverseService;
+
+    public static String getInstallationBrandName() {
+        return ConfigProvider.getConfig().getOptionalValue("dataverse.branding.installation.name", String.class)
```
poikilotherm, line 19: While I'm still baffled that injecting stuff into a `static` member is possible with EJB (I don't think it is with CDI), do you actually want to use the programmatic call or should we go for the injected version?

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
qqmyers: It could/should apply to any export that does this - only the DDI, DDI HTML Cookbook and OpenAire are affected today, but I assume we'd want other formats follow this setting if they include such a field. (ORE probably should do this, doesn't yet).

doc/sphinx-guides/source/installation/config.rst
qqmyers: I had to remove use of the microprofile name for the installationname - could not resolve some recursion problem. (trying to get to adding an issue). That said, my thought was that it probably isn't useful to advertise while there aren't many/all settings working this way (not sure it makes sense to add the complexity of supporting mixed ways of doing settings for just one setting.) I'm open if the group consensus is the other way though. (FWIW - the microprofile-alias file will essentially serve as a list of these as they're added - the microprofile docs you created could point to that to start for those who want to experiment/need to for k8s.)

src/main/java/edu/harvard/iq/dataverse/branding/BrandingUtil.java
qqmyers: I was unable to get the injection of the variable to work, so I picked this way. That said, I've now had to remove this one for other reasons. It's possible that @Injecting could work for the exclude setting used in the DDIExportUtil class - I didn't try there.

src/main/resources/META-INF/microprofile-aliases.properties
```diff
@@ -1,4 +1,6 @@
 # NOTE
 # This file is a placeholder for future aliases of deprecated config settings.
 # Format:
-# dataverse.new.config.option=dataverse.old.deprecated.option
\ No newline at end of file
+# dataverse.new.config.option=dataverse.old.deprecated.option
+dataverse.branding.installation.name=dataverse.settings.fromdb.InstallationName
+dataverse.export.distributor.excludeinstallationifset=dataverse.settings.fromdb.ExportInstallationAsDistributorOnlyWhenNotSet
```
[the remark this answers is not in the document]
qqmyers: It only affects DDI, DDI HTML Codebook, and OpenAire at present but I think it should apply to any export that includes the installation name as a distributor. (The ORE export probably should add it as the intent is for it to be complete  - I followed the json export rather than looking across all the exports when I created the first version.)

src/main/java/edu/harvard/iq/dataverse/settings/source/DbSettingConfigSource.java
```diff
@@ -50,7 +51,7 @@ public static void updateProperties() {
     @Override
     public Map<String, String> getProperties() {
         // if the cache is at least XX number of seconds old, update before serving data.
-        if (lastUpdate == null || Instant.now().minus(Duration.ofSeconds(60)).isBefore(lastUpdate)) {
+        if (lastUpdate == null || Instant.now().minus(Duration.ofSeconds(60)).isAfter(lastUpdate)) {
```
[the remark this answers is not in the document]
qqmyers: FWIW: I found one more w.r.t. mapping the setting names via Alias. I'd say that unless you have concerns about the fixes, it's not worth another PR - these classes aren't used anywhere else yet and the changes are small.

sekmiller (approved):

poikilotherm:
@qqmyers @sekmiller @kcondon

Looks like #7680 might be a blocker for this. Will resolve with a PR for #7680, but to not stop the show going on here, we might remove the MPCONFIG parts first.

qqmyers:
It's not a blocker (as it works as it is now), but I can do a quick edit to remove the DB source class that you plan to remove given your new design.

qqmyers:
An update - with the discussion and new direction with microprofile support, this PR is now independent of them. It now just introduces:
* an :InstallationName setting that, when set, is used instead of the root dataverse name in places where that name was being used for branding (i.e. in metadata exports, but not changing the name displayed for the root dataverse it the UI.)
* an :ExportInstallationAsDistributorOnlyWhenNotSet that avoids adding the repository brand name as a "Distributor" in the DDI metadata export stdyDsc element if a Distributor was already set in a Dataset's metadata. (So there will always be a stdyDsc Distributor element, but if this setting is true, there won't be an additional one listing the repository brand name.)

PR #7682, opened 2021-03-12
Workflow speed improvement

### LINKED ISSUE #7681: Improve workflow speed
From DANS' work to migrate ~100K datasets using a pre-publication workflow, @janvanmansum discovered that the 1 second delay added prior to the workflow start (which was part of changes needed to make post-publication workflows like the archiving workflows succeed) was a) adding ~27 hours to their runtime and b) wasn't necessary for pre-publication workflows (see https://github.com/GlobalDataverseCommunityConsortium/dataverse/pull/4). Building on that, a solution for both just needs a flag that indicates whether it is safe to find() the dataset from the database or if a delay is needed. A PR to follow - with more explanatory comments.

qqmyers:
**What this PR does / why we need it**: Removes a delay in launching prepublication workflows

Closes #7681

**Special notes for your reviewer**: small PR to incorporate a DANS performance improvement

**Suggestions on how to test this**: pre and post pub workflows should still work. Could run a workflow with the log step for prepublication. Setting up the local file archiver for post publication would allow checking that the version of the created bag matches the version of the new publication (versus being one behind, which was an error fixed in this same section of code.)

**Additional documentation**:


landreev (approved):
Makes sense.

PR #7688, opened 2021-03-15
IQSS/7687-Curate command updates

### LINKED ISSUE #7687: CuratePublishedDatasetVersionCommand doesn't update everything
The Curate command has two bugs:

1. 

The listed command correctly updates the fileaccessrequest flag on the latest datasetversion object when it is used, but it has not been updating the duplicate/convenience flag of the same name on the dataset itself. The consequence is that, while the edit TermsofUseAndAccess display shows correctly, the summary display shows the wrong value and the file access remains as it was.

This issue only affects datasets where a superuser has run the Curate command and the update included a change to this flag. 

A script that will find any instances of this is:
`select da.id, dv.id, ta.id, da.fileaccessrequest, ta.fileaccessrequest, dv.releasetime from datasetversion dv, termsofuseandaccess ta, dataset da where dv.dataset_id=da.id and dv.termsofuseandaccess_id=ta.id and ta.fileaccessrequest != da.fileaccessrequest and dv.versionstate='RELEASED' and dv.releasetime in (select max(releasetime) from datasetversion where dataset_id=da.id);`

This could go in flyway but I found one instance on QDR's dev server where out-of-order releasedates on some datasetversions makes the script report a false positive. That issue is separate and predates the creation of the Curate command, but it suggests, along with the fact that this case should be rare, that helping an admin manually inspect any cases rather than automatically making an update might be the better approach.

2. 

With #7337 applied (as at QDR today), file name updates get dropped. Merging the datasetversion at the end of the command resolves this (and shouldn't break things prior to #7337 if the bug doesn't exist there).

I'll submit a PR to fixes both and that changes a Permission annotation per discussion with @scolapasta (the command is restricted to superusers but only requires the EditDataset permission in the annotation. The suggestion is to put publishDataset there so that, if superadmin requirements are lifted in the future, the permission annotation will be more reasonable. This should have no effect on current functionality.)

qqmyers:
**What this PR does / why we need it**: Fixes a bug where the Curate (update an existing published dataset) command, usable by superadmins, didn't correctly update the file access request allowed flag if it was changed in the update. The PR also include a couple other fixes updates - at QDR, file name updates were being dropped, noticed after the equivalent of #7337 was merged. Also changing the permission annotation (see issue discussion).

Closes #7687

**Special notes for your reviewer**:  The file access request issue was just discovered and I've verified the fix at QDR. The change to add a merge of the datasetversion makes sense, but I'm just catching up from last year on that and my notes indicate that I thought some change related to #7337 could be involved, but it's possible that the problem affects develop.
W.r.t. addressing legacy db issues - I hope use is fairly rare and that if there were any name issues, those would have been caught/reported. The file access flag was more subtle, so easier to miss. As I noted in the issue, I'm not sure a flyway script makes sense but I could add a release note (that could either point to the query I listed that would help find instances and/or just point out that you can just do the update again to fix any affected datasets). Let me know what makes the most sense.

**Suggestions on how to test this**: Change the file request access flag in the Terms panel and update a file name, then use the Curate command and verify that Terms panel shows the flag update and/or that a restricted file acts as specified by the flag (requesting access is/isn't allowed), and make sure the file name changed.

**Is there a release notes update needed for this change?**: Some note along the lines of 'Bugs were found in the functionality allowing superadmins to update a published dataset version (without creating a new version) involving the Terms setting on whether to allow file access requests and with file name changes.  If you have used this functionality, issue #7687 provides details and includes an sql query that will help in identifying any affected dataset.'

**Additional documentation**:

### DIFF
--- a/doc/release-notes/7687-curate-command.md
+++ b/doc/release-notes/7687-curate-command.md
@@ -1,0 +1,9 @@
+## Notes for Dataverse Installation Administrators
+
+### DB Cleanup for Superusers Releasing without Version Updates
+
+In datasets where a superuser has run the Curate command and the update included a change to the fileaccessrequest flag, those changes would not be reflected appropriately in the published version. This should be a rare occurrence.
+
+Instead of an automated solution, we recommend inspecting the affected datasets and correcting the fileaccessrequest flag as appropriate. You can identify the affected datasets this via a query, which is available in the folder here:
+
+https://github.com/IQSS/dataverse/raw/develop/scripts/issues/7687/
--- a/scripts/issues/7687/file_access_flag_update_bug.txt
+++ b/scripts/issues/7687/file_access_flag_update_bug.txt
@@ -1,0 +1,11 @@
+-- this query will identify datasets where a superuser has run the Curate command and the update included a change to the fileaccessrequest flag, resulting in the file access request updates not being reflected in the published version
+
+select da.id, dv.id, ta.id, da.fileaccessrequest, ta.fileaccessrequest, dv.releasetime 
+from datasetversion dv, termsofuseandaccess ta, dataset da 
+where dv.dataset_id=da.id 
+and dv.termsofuseandaccess_id=ta.id 
+and ta.fileaccessrequest != da.fileaccessrequest 
+and dv.versionstate='RELEASED' 
+and dv.releasetime in (select max(releasetime) 
+from datasetversion 
+where dataset_id=da.id);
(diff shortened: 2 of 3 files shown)


scolapasta (approved):
Looks good (and good catch!). I'll leave it to @djbrooke to determine what we include in the release notes (my vote is a reference to the issue could be enough).

djbrooke:
(will review and add release note when I have a few)

scolapasta (approved):
Re-approved for script and release notes.

kcondon:
@qqmyers The publish in progress banner and page update is missing. Otherwise works.

PR #7690, opened 2021-03-16
Support for  ISO-639-3 language codes

### LINKED ISSUE #7638: Harvesting client fails : Zenodo "Couperin" community / error with "language" field mapping.
Dear community
I'm taking here the liberty to create an issue based on some previous discussions in the Dataverse-Users group. I'm trying to harvest some Zenodo communities. My client is set up as follows :

- Alias : zenodo_test
- Server URL : http://www.zenodo.org/oai2d
- Local dataverse : zenodo_test_couperin
- OAI set : user_couperin
- Metadata format : oai_dc
- Repository type : Generic OAI archive

Only 5 datasets are harvested out of the 20 present here: https://zenodo.org/communities/couperin/?page=1&size=20

<message>Exception processing getRecord(), oaiUrl=https://zenodo.org/oai2d, identifier=oai:zenodo.org:3773762, edu.harvard.iq.dataverse.api.imports.ImportException, Failed to import harvested dataset: class edu.harvard.iq.dataverse.util.json.ControlledVocabularyException (Value 'eng' does not exist in type 'language')</message>
(also : "Value 'fra' does not exist...")

Harvested for example: https://zenodo.org/record/4266132
Not harvested: https://zenodo.org/record/3948266

An exchange with @qqmyers (thanks!) linked that problem to another one regarding a SWORD Atom file import :
"With a quick look, it appears that that element is getting mapped to the Language field in the citation metadata block, whose values are controlled by the list of languages available. Those entries are all complete words (e.g. “English”) rather than the 2 or 3 letter ISO language codes. I haven’t checked to see if there’s an open issue to add support for language codes – might be something that could be addressed via an external service as being discussed in the CVV MDWG."

So it seems the problem is almost trivial, but it's still blocking the harvesting. Until further development solves this problem, is there a way around this that could be tried? Maybe a trick to ignore the language field altogether?

Best 

Thomas

landreev:

This will allow us to import metadata where the "language" field is populated not by a literal value ("French", "English") but by 3 letter ISO-639-3 codes ("fra", "eng"). 
This problem was encountered by a remote installation when harvesting Dublin Core records from Zenodo. DC documentation suggests that using these codes to specify the language is an acceptable practice. 

In this PR these codes are added as "alternate values" for the corresponding controlled vocabulary entries (more info in the issue). Once a record with a field like this (for example, `<dc:language>fra</dc:language>`) is imported, it becomes a controlled vocabulary entry `French` in our metadata. 

Closes #7638

Note that our existing metadata block update API was not updating these "alternate values" found in the TSV. Those were only populated on the initial import. So I had to change that. 

Using the example from the linked issue: 
Once the branch is built and deployed,
Update the citation block: 
```
wget https://raw.githubusercontent.com/IQSS/dataverse/7638-iso-639-3-language-codes/scripts/api/data/metadatablocks/citation.tsv
curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "Content-type: text/tab-separated-values"
```
Create a harvest from the original issue:
```
harvestUrl: https://www.zenodo.org/oai2d
metadataFormat: oai_dc
set: user-couperin
```
(This OAI server offers hundreds of sets. This PR has an extra improvement, unrelated to languages - the sets will appear *sorted* in the pull down menu, making it easier to use)

The harvest should be able to import all 21 records in the set. Including the 16 of them that have these language codes in the metadata, that were failing previously. 

**Additional documentation**:

### DIFF
--- a/doc/release-notes/7638-citation-metadatablock-update.md
+++ b/doc/release-notes/7638-citation-metadatablock-update.md
@@ -1,0 +1,7 @@
+### Citation metadatablock update
+
+Due to a minor update in the citation metadata block (extra ISO-639-3 language codes added) a block upgrade is required:
+
+`wget https://github.com/IQSS/dataverse/releases/download/v5.4/citation.tsv`
+`curl http://localhost:8080/api/admin/datasetfield/load -X POST --data-binary @citation.tsv -H "Content-type: text/tab-separated-values"`
+
--- a/scripts/api/data/metadatablocks/citation.tsv
+++ b/scripts/api/data/metadatablocks/citation.tsv
@@ -137,188 +137,188 @@
 	authorIdentifierScheme	ResearcherID		6
 	authorIdentifierScheme	ScopusID		7
 	language	Abkhaz		0
-	language	Afar		1											
-	language	Afrikaans		2											
-	language	Akan		3											
-	language	Albanian		4											
-	language	Amharic		5											
-	language	Arabic		6											
-	language	Aragonese		7											
-	language	Armenian		8											
-	language	Assamese		9											
-	language	Avaric		10											
-	language	Avestan		11											
-	language	Aymara		12											
-	language	Azerbaijani		13											
-	language	Bambara		14											
-	language	Bashkir		15											
-	language	Basque		16											
-	language	Belarusian		17											
+	language	Afar		1	aar
+	language	Afrikaans		2	afr
+	language	Akan		3	aka
+	language	Albanian		4	sqi
+	language	Amharic		5	amh
+	language	Arabic		6	ara
+	language	Aragonese		7	arg
+	language	Armenian		8	hye
+	language	Assamese		9	asm
+	language	Avaric		10	ava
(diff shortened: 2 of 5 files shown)


sekmiller (approved):
Looks good

kcondon:
Followed the instructions: deployed, applied citation.tsv, created harvest client and it does harvest 99 datasets (there was no subset available) but it failed with an unknown error and nothing in logs. I then attempted a basic harvest against Harvard and it failed outright, no indication of why, just a get request failure. I had been using oai_dc and generic oai client for initial test but switched to dataverse_4+ against Harvard once generic failed and no luck.

landreev:
It sounds like there's something wrong with this system, that's unrelated to this PR; if it can't harvest anything. You definitely want to harvest this particular set; if it's not seeing any sets, it means something is wrong. Let's take a closer look.

kcondon:
as discussed, I can try testing against develop branch. harvesting worked partially against test case.

PR #7695, opened 2021-03-17
6953 mpconfig context param

### LINKED ISSUE #6953: As a developer, I want changes I make to the front end to be visible when I deploy from Netbeans

poikilotherm:
This allows for a very dynamic changing of JSF parameters without recompilation. It adds docs how to do this, too.

This PR also relies on the fact that people should upgrade from Payara 5.2020.6 or later to at least 5.2021.1 to gain support for MPCONFIG 2.0 and default parametes in `${MPCONFIG=...}` variable substitution, which is described in #7701

Closes #6953 
:zap: Blocked by #7701 :zap: 

None.

Change some of the documented vars e. g. via system properties.

Nope.

Yes, two. :battery: included.

**Additional documentation**:
None.


pdurbin:
This seems like a great solution to #6953 but it depends on pull request #7701 so it probably doesn't below in code review.

poikilotherm:
As this relies on Payara 5.2021.1 (see 48d6501 for an extensive explanation), I will put this back to draft.

poikilotherm:
As #7701 is likely to be merged soonish, pushing this PR back into the pipeline.

pdurbin:
It looks like pull request #7701 is a little more involved that we originally thought and since it's a dependency (and since the Review column is pretty full), let's leave this in Community Dev until that other pull request is merged.

poikilotherm:
Putting this back in Review now...

djbrooke:
Thanks @poikilotherm, we'll review. You mentioned two release notes in the PR description but I only saw one in the changed files - is there one missing or was the second release note handled elsewhere?

poikilotherm:
Looks like a leftover. Just ignore that... Thx!

pdurbin (approved):
@poikilotherm I'm so happy for this pull request. In the years since 5f99037 I've had to maintain a local edit to `src/main/webapp/WEB-INF/web.xml` and be careful not to commit it to git in order to for my edits to the GUI to be visible. Now, it's just a simple change to domain.xml, which I added to the docs and the "rebuild" script (using asadmin).

PR #7708, opened 2021-03-22
7634 order dv contents for api

### LINKED ISSUE #7634: Deterministic order of dataverse contents (API)
*Feature request*: when the API returns objects within a dataverse, can the order be consistent across repeated calls?

I don't think the equation for the order matters, but having some consistent order would make our demo/example code more repeatable.  It would also eliminate the ordering that our test suite requires, [like @kuriwaki recently included](https://github.com/IQSS/dataverse-client-r/commit/09eb25aff9b38e976d2feadb2eb336b93991ee3d).

@pdurbin [isolated the following code](https://github.com/IQSS/dataverse-client-r/issues/85#issuecomment-785186798) in the [dataverse-client-r](https://github.com/IQSS/dataverse-client-r) repo:
```
@NamedQuery(name = "DvObject.findByOwnerId", 
            query = "SELECT o FROM DvObject o WHERE o.owner.id=:ownerId"),
```

I don't know the database, but two options that come to mind are:
1. ordering by the primary key for performance reasons, and
2. ordering by the human-assigned name.  So if the owner updates or deletes & recreates the object, the returned order is unchanged.

My slight preference is the latter, but either would help us and be appreciated.

(*@kuriwaki, anything you'd like to add?*)

sekmiller:
**What this PR does / why we need it**: Orders the results of the get contents of a dataverse ordered by type then id. That is all Dataverse Collections are listed first in id order; then datasets in id order.

Closes #7634 Deteministic order of dataverse contents api

ordered by type then id. noted that the return json does not include title of dataset.

**Additional documentation**:
updated doc to reflect ordering of return list


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListDataverseContentCommand.java
```diff
@@ -32,11 +34,14 @@ public ListDataverseContentCommand(DataverseRequest aRequest, Dataverse anAffect
...
     public List<DvObject> execute(CommandContext ctxt) throws CommandException {
+        List<DvObject> availableObjects = new ArrayList();
         if (getRequest().getUser().isSuperuser()) {
-            return ctxt.dvObjects().findByOwnerId(dvToList.getId());
+            availableObjects = ctxt.dvObjects().findByOwnerId(dvToList.getId());
         } else {
-            return ctxt.permissions().whichChildrenHasPermissionsForOrReleased(getRequest(), dvToList, EnumSet.of(Permission.ViewUnpublishedDataverse, Permission.ViewUnpublishedDataset));
+            availableObjects =  ctxt.permissions().whichChildrenHasPermissionsForOrReleased(getRequest(), dvToList, EnumSet.of(Permission.ViewUnpublishedDataverse, Permission.ViewUnpublishedDataset));
         }
+        availableObjects.sort(Comparator.comparing(DvObject::getDisplayName));
```
pdurbin, line 43: Since we don't display the name of datasets, wouldn't it make more sense to sort by id?

pdurbin (commented):
Sorting by id probably makes more sense.

Also, should we update the API Guide to state that the order is now deterministic?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListDataverseContentCommand.java
scolapasta: And if sorting by id, can we just do that in the db query? (seems better performance that sorting after in the java. We could also sort first by dtype, then by id if we wanted to effectively group the dataverses, then datasets)

sekmiller: Is there agreement that we should just use id? (just use order by in the query and ignore type)

scolapasta: My vote is first by dtype, then id, which can both be done in the query.

sekmiller: I updated the query to order by dType (descending to list dataverses first) then dvObject ID. I also updated the doc to note this ordering.

scolapasta (approved):
Good, simple change.

PR #7713, opened 2021-03-23
Update harvestserver.rst (add an example)

### LINKED ISSUE #7699: Question: OAI set with all datasets of a dataverse including those only included by link.
Quick question:
Is it possible to define an OAI set in such a way that it contains all datasets of a dataverse including those that are only included via link?
parentId=... does not work (as already suspected).

lmaylein:
Add an example for OAI Set search queries (subtreePaths).

Closes #7699 

**Additional documentation**:


djbrooke (approved):
Thanks @lmaylein !

Issue #7677, opened 2021-03-11
Reset Password button does not result in info banner feedback, shows no feedback in develop branch


kcondon:
Go to forgot your password, enter email, click Submit Password Request button and get no feedback where on demo/5.3 you get an info message saying it was sent if email exists.

djbrooke:
Thx @kcondon. I reported this in #7106 but it's been frustratingly intermittent. I'll move to up next so we can dig in.

kcondon:
So, after discussing with Stephen, found that disabling the updated google analytics code allowed it to work again: https://guides.dataverse.org/en/5.4.1/installation/config.html#web-analytics-code

Is this a case of the analytics code breaking the event handling/listener on the page? I don't know how that works so am speculating ;)

sekmiller:
I added web analytics to my local environment, but I am still seeing the proper messaging on reset password.

qqmyers:
FWIW: There was a time when the analytics code examples caused trouble - I think because they didn't have everything in CDATA blocks, so it's possible that out of date versions would break this and the latest wouldn't.

sekmiller:
@qqmyers as a test I took out the CDATA blocks and now the messaging works. So what was the issue that adding it was trying to solve?

qqmyers:
#5967

PR #7725, opened 2021-03-25
added missing method to BuiltinUser required by passwordreset.xhtml

### LINKED ISSUE #7724: Error Rendering View[/passwordreset.xhtml]
While experimenting with upgrading from an older encryption version in Dataverse (SHA to  BCRYPT, @qqmyers ), I bumped into a bug in `/passwordreset.xhtml` when the password reset function is called.

Log output:
`Error Rendering View[/passwordreset.xhtml]`
`javax.el.PropertyNotFoundException: /passwordreset.xhtml @104,99 value="#{PasswordResetPage.user.displayInfo.title}": The class 'edu.harvard.iq.dataverse.authorization.providers.builtin.BuiltinUser' does not have the property 'displayInfo'.`

I'm using the latest develop branch (5.3)
Should be solved quite easily by adding a `getDisplayInfo()` method to BuiltinUser

mderuijter:
Fixes bug in password reset functionality

Closes #7724 

No
You have to trigger an "upgrade" type of password reset; the way to do this is to update the encrypted password of a built in user and its encryption type of 0 directly in the db. You can get the correct encrypted password by running this from a command line (thanks @qqmyers), replacing "password" with your chosen password (within quotes):
```
echo -n "password" | openssl dgst -binary -sha1 | openssl enc -base64
```

No
No
**Additional documentation**:
No


djbrooke:
Thanks @mderuijter - would you expect this would fix #7677 as well?

mderuijter:
> Thanks @mderuijter - would you expect this would fix #7677 as well?

@djbrooke I'm afraid it will not fix #7677 as the function I added is only used here:
image

qqmyers:
I think this PR should work but it essentially duplicates a method that is marked as deprecated. I'm not sure what the best approach would be but I can see two alternatives:
1) Change the passwordreset.xhtml to call the deprecated method (or change it's name from getDisplayInfoForApiCreation() to getDisplayInfo() (and remove the @Deprecated since it is really used?), or 
2) Add a PasswordResetPage.getUserDisplayInfo() method that can call authService.getAuthenticatedUser(user.getUserName()).getDisplayInfo() - using the builtinUser to get the AuthenticatedUser which has the undeprecated getDisplayInfo method. (and change the xhtml to use that method instead.)

@scolapasta - any preference? I'm starting to prefer the former as it appears the reason the method is deprecated was just to indicate that BuiltinUser only has transient values now (e.g. see https://github.com/IQSS/dataverse/blob/ac15ff5e0fd956ca4181d49d76ebf9dc6f6fdc43/src/main/java/edu/harvard/iq/dataverse/api/BuiltinUsers.java#L79-L85).

scolapasta:
So I agree the first approach would be fine, except that it turns out, it doesn't actually get the name of the user.

image

Discussed it with @djbrooke and we're fine with dropping the user's name from the text - so this change could just be to change  the bundle message to not include the 2nd parameter (i.e. {1}) - and then you can also change the xhtml to no longer reference displayInfo at all. That should be the easiest fix to get this page working again.

mderuijter:
> So I agree the first approach would be fine, except that it turns out, it doesn't actually get the name of the user.
> 
> image
> 
> Discussed it with @djbrooke and we're fine with dropping the user's name from the text - so this change could just be to change the bundle message to not include the 2nd parameter (i.e. {1}) - and then you can also change the xhtml to no longer reference displayInfo at all. That should be the easiest fix to get this page working again.

@scolapasta should be fixed now :)

mderuijter:
@djbrooke Hi my PR update has probably slipped your attention, but I think it has all the suggested modifications now.

qqmyers (approved):

djbrooke:
Thanks @mderuijter and @qqmyers !

scolapasta:
Thanks @qqmyers. (I was out since late last week, so just catching up today)

kcondon:
Hello @mderuijter, would you mind refreshing your branch from /develop? I'm about to test this and found it was still using the old release version. Thanks, Kevin

mderuijter:
> Hello @mderuijter, would you mind refreshing your branch from /develop? I'm about to test this and found it was still using the old release version. Thanks, Kevin

Hello @kcondon, should be okay now. Let me know if it isn't

PR #7734, opened 2021-03-29
Adding formatting and updating text for publish message

### LINKED ISSUE #7675: Update Success Messaging to indicate that Publishing/Submit for Review is required
We have a few spots in the application where we've been asked to add an additional nudge to get people to publish so their changes can actually take effect. The curation team is finding that some content is unexpectedly left unpublished or not submitted for review as expected. This is part of a larger effort around these workflows, but some text changes are a good first step. Ex.

dataset.message.createSuccess=This dataset has been created
dataset.message.filesSuccess=The files for this dataset have been updated

Review these messages and update them working with @sbarbosadataverse @jggautier or another member of the curation team.

djbrooke:

Some updates to the text for publishing and submit for review, in order to encourage those actions.

(re) closes #7675 

N/A

Create a dataset draft in a Dataverse collection where you can publish, then create a dataset draft in a Dataverse collection where a submit for review workflow is enabled. 

Yes, will link to an AWS instance here.

No

**Additional documentation**:

None


sekmiller (approved):
Looks good.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -727,6 +728,27 @@ public void exportAllDatasets(boolean forceReExport) {
 
     }
     
+    //get a string to add to save success message
+    //depends on dataset state and user privleges
+    public String getReminderString(Dataset dataset, boolean canPublishDataset) {
+
+        if( dataset.getId() == null || !dataset.isReleased() ){
```
scolapasta, line 735: I wonder why we have to check the null here? I know you just moved this from somewhere else, so it's fine if we leave it. That other place had a comment that said "dataset id is null before first save -- needed for create mode", but even in that circumstance wouldn't isReleased still also be false? So seems extraneous.

scolapasta (commented):
Looks fine. I'm not officially approving here because @djbrooke still wanted to check something.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
sekmiller: We don't. When I wrote the method originally I had it higher in the Save method so that newly created datasets did not have an id, but the permissions wrapper requires an id before it can determine canPublishDataset so I moved the call to just before rendering the success method. I will remove it here.

PR #7748, opened 2021-03-31
Document superuser requirements for locks APIs

kaitlinnewson:

The "DELETE" and "POST" commands in the Native API for dataset locks appear to be limited to superusers. This PR adds this to the documentation.

No

No

**Additional documentation**:

Links to API code which indicate superuser requirement:
- https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L2033
- https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L2096


pdurbin (approved):
Looks good. Thanks for linking to the code, @kaitlinnewson!

PR #7765, opened 2021-04-05
Adding a reference to the docs so they build

djbrooke:

Fixes a build error:

doc/sphinx-guides/source/installation/config.rst:2240: WARNING: undefined label: :installationname (if the link has no caption the label must precede a section header)

Closes N/A

**Additional documentation**:


pdurbin (approved):
Seems like a good fix. Works on my machine.

Issue #5989, opened 2019-07-03
Add new Solr fields via an API call (investigation)


pkiraly:
When it comes to adding custom metadata block, there is a manual step involved: adding Solr fields to schema.xml, however Solr provides an API (Schema API) to make this manual step unnecessary, however it invloves some changes in Dataverse.
Note: this ticket is about investigation instead on implementation. First we have to understand every aspect of this change to make sure the existing technologies are reliable and fully support the request.

Solr has a Schema API, which lets you to modify the Solr schema (the list of fields and their properties). Solr can handle the schema in two different ways, and it can be controlled in the solrconfig.xml file. There is a "classic" way, which is based on schema.xml file, and a newer way, called managed schema (its materialization is the "managed-schema" file, and it is editable via the Solr user interface or via API, but it is not advised to edit this file manually).

In the Dataverse provided solrconfig.xml you have this:
 <schemaFactory class="ClassicIndexSchemaFactory"/>

The schema API doesn't work with the ClassicIndexSchemaFactory. If you try, Solr returns an error message: "schema is not editable". To enable Schema API, we have to change this setting:

Set ManagedIndexSchemaFactory in solrconfig.xml:
```
 <schemaFactory class="ManagedIndexSchemaFactory"/> 
```

After this you have to restart Solr, and the Schema API will work this way:

```
curl -X POST -H 'Content-type:application/json' \
  http://localhost:8983/api/cores/collection1/schema --data-binary '{
  "add-field":{
     "name":"title", "type":"text_en", "multiValued":false,
     "stored":true, "indexed":true
  },
  "add-copy-field":{"source":"title", "dest":"_text_", "maxChars":"3000"}
}'
```

The details of the Schema API can be found here: 
https://lucene.apache.org/solr/guide/7_3/schema-api.html

The details of change from classic schema:

https://lucene.apache.org/solr/guide/7_3/schema-factory-definition-in-solrconfig.html#SchemaFactoryDefinitioninSolrConfig-Switchingfromschema.xmltoManagedSchema

The problems:
The documentation says: "Once Solr is restarted and it detects that a schema.xml file exists, but the managedSchemaResourceName file (i.e., “managed-schema”) does not exist, the existing schema.xml file will be renamed to schema.xml.bak and the contents are re-written to the managed schema file." When I tried it, the schema.xml were not copied, and not renamed. However since the same searches, even fielded searches are working.
[...]

PR #7801, opened 2021-04-14
6070 Update Apache Commons Lang3

### LINKED ISSUE #6070: Dependency update: Apache Common Lang v2 -> v3
This is related to my other dependency housekeeping things. See #5288 and #5360.

----

While programming unit tests for #6000, I wanted to use `org.apache.commons.lang3.RandomStringUtils`. In Maven POM I saw that Dataverse is still on [`org.apache.commons.lang` v2.6](https://mvnrepository.com/artifact/org.apache.commons/commons-lang/2.6), which is dead since 2011, when 3.0 was released.
See also: http://www.h-online.com/open/news/item/Apache-Commons-Lang-3-0-makes-a-break-with-the-past-1283761.html

Dataverse uses a lot of Commons Lang stuff. It should be refactored to use v3 of the lib, which gets included anyway, as it is a dep of `com.lyncode:xoai-common`. (see `mvn dependency:tree`, grepping for `commons-lang`). Nothing else is relying on the v2 lib.

Notes about upgrading from upstream project: https://commons.apache.org/proper/commons-lang/article3_0.html

poikilotherm:
I want to use modern [Apache Commons Lang3 functionality (FailableStream)](https://commons.apache.org/proper/commons-lang/apidocs/index.html?org/apache/commons/lang3/stream/Streams.FailableStream.html) for #5989. ([see usage in code](https://github.com/poikilotherm/dataverse/blob/c1a7e56d08fe4a7d88ed383edd5aea60a88c1f00/src/main/java/edu/harvard/iq/dataverse/search/schema/SolrFieldType.java#L60))
To keep the scope of the PR for that issue smallish, I created this PR, as you folks prefer small steps.

The dependency update has been a long standing thing. See the issue. 

Closes #6070

:bug: THERE IS A HIDDEN BUG.
Some tests for the CSV ingest fail.

Switching http://github.com/IQSS/dataverse/blob/eca8b8fd403002327439f81f54c6ed9edde391c7/src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/csv/CSVFileReader.java#L211-L211 like this makes all tests pass:
```diff
-&& StringUtils.isNumeric(varString.substring(1))));
+&& org.apache.commons.lang.StringUtils.isNumeric(varString.substring(1))));
```

Switching that function from Lang3 back to Lang2 should NOT make the tests pass. Playing around with the CSV files and the expected tests results makes the tests flaky with v2 and pass with v3. It looks like there is a hidden bug in the CSV parser not revealed by the tests. (The only change in `isNumeric()` from v2 -> v3 is that empty strings are also not counted as numeric. The test cases have no example covered by this change.)

As I am not very familiar with the ingest code, I would appreciate some help.

Usual unit tests and integration tests. UI uses some escaping routines, but they have just been moved in upstream and there is no functional changes expected.

Nope.

I don't think so.

**Additional documentation**:
None.


djbrooke:
Thanks @poikilotherm for flagging the failing test. We've been dealing with some production issues on dataverse.harvard.edu this week, but we'll take a look at this as soon as we can.

qqmyers:
FWIW: In general, I think these changes should be fine - just updating to a later version.  (We actually had a few lang3 calls already (it must be a dependency of something already in the pom)). 

For the failing code - it just looks like it is trying to check that the field is either null, empty, or a number potentially starting with a + or -. The code does expect that the string is at least two chars long and should probably be fixed by splitting the cases up a bit further - a one char string has to be numeric and if it is two or more chars, the existing check would work (and not give an empty value to isNumeric()).

Others may know more, but it looks straight-forward enough that I think you could just make the change and let people review the result.

poikilotherm:
> FWIW: In general, I think these changes should be fine - just updating to a later version. (We actually had a few lang3 calls already (it must be a dependency of something already in the pom)).

Aye, the transitive dependency has been used. [See also my extensive review from last year.](https://github.com/IQSS/dataverse/issues/6070#issuecomment-527371642)

> For the failing code - it just looks like it is trying to check that the field is either null, empty, or a number potentially starting with a + or -. The code does expect that the string is at least two chars long and should probably be fixed by splitting the cases up a bit further - a one char string has to be numeric and if it is two or more chars, the existing check would work (and not give an empty value to isNumeric()).

Jim I know that the code is expected to do that. Yet it seems to be more complicated, which is why I consider there is a bug hidding either in the test code or within the app code.

grafik

Test `testRead` is expecting "null" to become "0", yet it becomes "0.0" [for this line of CSV](http://github.com/IQSS/dataverse/blob/3e68a9ccc21765d27bb2b8c6f8b77f34f97d505d/src/test/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/csv/IngestCSV.csv#L7-L7)

The other tests also are failing because of detecting as continuous instead of discrete.

Removing all but L1 (header) plus L7 from `IngestCSV.csv` is producing even weirder test results for `testRead()`:
grafik

Who want's to go for a :bug: hunt?

landreev:
I believe @qqmyers has been right all along - it was simply the single digit numbers (like the "2" and "0", in the lines 3 and 4 of the CSV file, respectively) that were breaking the integer test. The code was checking the first character - which is also the *last* character of the string, in this case - against `[+-0-9]`; and then trying to do StringUtils.isNumerc() on the *remainder* of the string - which as of lang3 would return false on an empty string... So then, in the second/final pass, that character string "null" in line 7 would be formatted as "0.0", on the assumption that it was a floating point zero, not an integer kind... (or "continuous" vs. "discrete", for the purposes of the other tests). 
So yes, what Jim suggested above could be a possible fix - a separate case for a single-char string (`[0-9]`), the current test for all other cases...; but it may be easier to just replace these lines:
```
   || (firstNumCharSet.contains(varString.charAt(0))
       && StringUtils.isNumeric(varString.substring(1))
```
with 
```
   || varString.matches("^[+-]?[0-9]+$")
```
So I'm going to check that in. (this change makes the test pass; I'll give it another thought, whether there are advantages to sticking with `StringUtils.isNumeric()`...)

landreev:
I don't really get why Jenkins failed to build the branch again, after I checked in the fix for the CSV ingest test. (The branch is building fine for me locally). 
Anyone?

donsizemore:
@landreev Travis failed as well:

```
[ERROR] COMPILATION ERROR : 

[INFO] -------------------------------------------------------------

[ERROR] /home/travis/build/IQSS/dataverse/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteRegisterService.java:[178,51] cannot find symbol

  symbol:   method escapeXml(java.lang.String)

  location: class org.apache.commons.text.StringEscapeUtils
```

landreev:
@poikilotherm Could you please sync the branch up with develop? 
The change I have committed fixes the CSV ingest tests that were failing. The bad news is, there appears to be another conflict between your branch, and something that has been added to develop since the PR was made. Hence the most recent Jenkins failure (Jenkins automatically syncs up its checkout with develop before attempting to build). 
So let's update the branch, and then investigate further.

poikilotherm:
Hi. Have been on vacation, thus offline.
I updated the branch as requested and fixed the failing `escapeXML()` call. All local unit tests now running smoothly.

poikilotherm:
One remaining thing pops up in my mind: as formerly there had been v2 and v3 usage in parallel, should be make `checkstyle` rule check for v2 usages, so we don't introduce transitive dependency use?

We could let @reviewdog bark in PR reviews based on these rules. https://github.com/marketplace/actions/run-java-checkstyle

landreev:
Great, thank you. I may check in a better permanent fix for the CSV ingest (trying to decide on that now), and will move the PR after that.

landreev:
Also, yes, it would be a good idea to add a check for transitive dependencies like this, as part of the overall automated review framework.

landreev (approved):

poikilotherm:
I created a test branch and a test PR to show the usage of a barking reviewdog in action...
https://github.com/poikilotherm/dataverse/pull/441

The check action is reporting the failed state. It does not leave comments as it's running in a fork, only showing the check report inline.

grafik

It's fairly simple to use and only barks on files of the PR's changeset. Shall I include this in this PR?

kcondon:
Hi @poikilotherm I'm testing this today but due to recent dataverse version change to 5.5, cannot build/deploy. Would you mind updating this branch from develop? Thanks.

poikilotherm:
Done @kcondon

PR #7806, opened 2021-04-15
7714 configure download all ds

### LINKED ISSUE #7714: Configuration for Availability of "Download All Dataset"
The "Download All Dataset" feature introduced to the Dataverse Software in 5.0 has been working great on the Harvard Dataverse Repository for the last several months. :) However, we should consider that:

- On the Harvard Dataverse Repository we in parallel activated the zipper external service to move the zipping off the application server, which means we haven't run into system stability issues. Not every installation will enable this service.
- Because of improvements in large data uploads, we're now seeing some datasets that are pretty big. See Ryan Enos’ recent dataset at 100+ GB (https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/A40X5L). Providing a good user experience around 100+ GB Zip downloads through the UI may be difficult, and there may be cost considerations for some installations. 

A proposal surfaced in #7557 is to use the :ZipDownloadLimit setting to control whether or not the Download All option is available. I think this could serve the two use cases of:

- As a Dataverse installation administrator I want to provide download options that are more suited for my infrastructure/cost
- As a Dataverse installation downloader I want to only be presented with a "Download All" option when I'm going to get a zip of all of the files

I think we should move forward with this. 

Three things to note:

- The UI impact here would be to either not have the button, or provide the button and some messaging about why the zip cannot be downloaded. To discuss. 
- after implementing this, we'll still have the issue of incomplete zips (with manifest.txt showing the missing files) when someone goes through and selects a number of files that are above the :ZipDownloadLimit, but it won't happen with the prominently featured "Download All" option at the top of the page. I don't want to solve this right now. 
- that the :ZipDownloadLimit is ignored when the zipper service is available, so in those installations running the zipper service this would solely be a setting that controls the availability of Download All. I considered suggesting a new option to be more specific, but I'll let more technical people discuss the best approach.

sekmiller:
**What this PR does / why we need it**: Modifies messaging when a download exceeds the limit for zipping. Previously we would zip as many files as possible and put the names of the missing files in the manifest file. This was confusing and annoying to users who assumed they got all of the files requested. With this change we will inform the user that they have selected files in excess of the limit and let them know what their options are.

Closes #7714 Configuration for Availability of "Download All"

The Access Dataset Button will display in various ways depending on the size of the files relative to the ZipDownloadLimit.
Assuming the user may download files the Access Dataset button will be available, but when the user clicks into it instead of a download link they may see a note saying the files are too large to be downloaded. Also if there are tabular files present and the original format is too large, while the archival version is smaller than the limit they will see one download link and one note saying the format is too large (or vice versa). 

In order to exercise the various scenarios it will be useful to put various values in:
curl -X PUT -d 5000000 http://localhost:8080/api/admin/settings/:ZipDownloadLimit
to test 

https://docs.google.com/presentation/d/1aSTQKu5t2fEM9oGUuvKo0yyj-wbjKfYuYqrsmdhe4RQ/edit#slide=id.g7673497ae2_0_0

**Additional documentation**:


src/main/webapp/dataset.xhtml
```diff
@@ -144,41 +144,65 @@
...
+                                                                <h:outputText value="#{bundle['dataset.accessBtn.too.big']}"  >                                           
+                                                                </h:outputText> 
+                                                            </li>
+
                                                             <!-- DOWNLOAD ORIGINAL BUTTON (TABULAR FILES PRESENT) -->
-                                                            <li jsf:rendered="#{DatasetPage.versionHasTabular and DatasetPage.sizeOfDatasetOrig != '0 B' }">
-                                                                <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadOriginal()}" styleClass="btn-download">
+                                                            <li jsf:rendered="#{DatasetPage.versionHasTabular and DatasetPage.sizeOfDatasetOrig != '0 B'  and !(DatasetPage.sizeOfDatasetOrigNumeric > settingsWrapper.zipDownloadLimit) }">
+                                                                <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadOriginal()}"
```
scolapasta, line 168: it probably make sense, but can you explain the reason behind this change (and the similar one below it?)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -3025,6 +3010,11 @@ public boolean isSelectAllFiles() {
     public void setSelectAllFiles(boolean selectAllFiles) {
         this.selectAllFiles = selectAllFiles;
     }
+    
+    public void handleFileSelectClick(){
```
scolapasta, line 3014: is this used anywhere? It's new, but I don't see any other reference.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -3074,6 +3064,27 @@ public void updateFileCounts(){
...
+    }
+
+    public Long getArchivalSelectedFileSize() {
+        return archivalSelectedFileSize;
+    }
+
+    public void setArchivalSelectedFileSize(Long archivalSelectedFileSize) {
+        this.archivalSelectedFileSize = archivalSelectedFileSize;
+    }
+
```
scolapasta, line 3087: are these used? I couldn't find their references.

scolapasta (changes requested):
Nice work. A lot of suggested small tweaks, and some changes due to calls being made from rendered attributes.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2846,58 +2847,110 @@ public void setSelectedNonDownloadableFiles(List<FileMetadata> selectedNonDownlo
...
+        boolean original = false;
+        return DatasetUtil.getDownloadSizeNumeric(workingVersion, original);
+    }
+    
+    public String getSizeOfSelectedAsString(){
+        return FileSizeChecker.bytesToHumanReadable(getSizeOfSelectedOrigNumeric());
+    }
+    
+    public String getSizeOfSelectedMaxAsString(){
+        return FileSizeChecker.bytesToHumanReadable(Math.max(getSizeOfSelectedOrigNumeric(), getSizeOfDatasetArchivalNumeric()) );
```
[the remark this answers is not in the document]
sekmiller: yes. good catch.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2846,58 +2847,110 @@ public void setSelectedNonDownloadableFiles(List<FileMetadata> selectedNonDownlo
...
         }
-        for (FileMetadata fmd : this.selectedFiles){
-            if(this.fileDownloadHelper.canDownloadFile(fmd)){
+        
+        for (FileMetadata fmd : this.selectedFiles) {
+            if (this.fileDownloadHelper.canDownloadFile(fmd)) {
                 getSelectedDownloadableFiles().add(fmd);
+                DataFile dataFile = fmd.getDataFile();
+                if (downloadOriginal && dataFile.isTabularData()) {
+                    bytes += dataFile.getOriginalFileSize() == null ? 0 : dataFile.getOriginalFileSize();
```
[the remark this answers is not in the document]
sekmiller: OK, this is kind of weird. I copied the code from the Util without really thinking about it, but after thinking about it I thought it might be for cases of an old file where we didn't fill in or backfill the original file size - but then I looked at my local DB and I have a bunch of cases (well 5 out of 23) where the the original file size on the datatable record is null. I'm going to dig a little more.

src/main/webapp/dataset.xhtml
sekmiller: We don't want to render the download all here if the size exceeds the zip limit. Instead we show a message pointing the user to use the files table to select files.

scolapasta: Right, I guess I'm trying to understand why there's a rendered on the li but also the validate method called on the link.

sekmiller: The validate method on the link does a bunch of things like check to see if you have permission to download the files and compare the size of files you are asking for to the zip limit (which is more for the case where you select from the files table, but I put it in one method to prevent a lot of duplicate code from the prior version.) Then depending on whether you have selected files for which you don't have permission or too large of a download shows different popups - which is why your suggestion that we put the logic for displaying popups in the page may be hard to accomplish. i'll see what I can do there.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: Removed. I think I put it in in anticipation of giving immediate feedback when the files selected were too large for zip.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: Removed. abandoned in favor of other/existing methods.

src/main/webapp/dataset.xhtml
scolapasta: right, but then what is the rendering checking? I must be missing something, if it helps explain it, we could do a quick call later.

Re: the popups, we should be trying to minimize the popups as much as possible; i.e. single message popups that have just a close. I would suggest trying to see if you can use the validate message to set what the message is in the popup (making sure the page updates the popup) and then the code to show the popup happens on the page. Let me know if that makes sense.

sekmiller: The rendering here is checking for whether there are tabular files because it shows multiple links if there are (original or archival downloads). Maybe there's a different approach to take here, but I didn't really change it much except to add the various size tests(and I did set booleans in the init so that the renders won't have to do the same calculation multiple times)
Re: popups I thought you were objecting to the method actually opening them. I can work with the idea of having a single popup where the message changes - we already do that because we list the files that you don't have permission to download.
And Tania asked for the "just close" popup for when the size is too big - it also includes a link to the Download API in the guides.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2825,197 +2830,219 @@ public void setSelectedRestrictedFiles(List<FileMetadata> selectedRestrictedFile
...
+    public List<FileMetadata> getSelectedNonDownloadableFiles() {
+        return selectedNonDownloadableFiles;
+    }
+
+    public void setSelectedNonDownloadableFiles(List<FileMetadata> selectedNonDownloadableFiles) {
+        this.selectedNonDownloadableFiles = selectedNonDownloadableFiles;
+    }
+
+    public String getSizeOfDataset() {
+        boolean original = false;
```
scolapasta, line 2855: not enough reason for the extra boolean here

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -4054,53 +4080,31 @@ public String getDownloadType() {
...
         this.downloadType = downloadType;
     }
-    
-    
+
+
     public void openDownloadPopupForMultipleFileDownload() {
         if (this.selectedFiles.isEmpty()) {
             //RequestContext requestContext = RequestContext.getCurrentInstance();
             PrimeFaces.current().executeScript("PF('selectFilesForDownload').show()");
```
scolapasta, line 4088: should we consider removing this one while we're here as well? (or is that out of scope)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -4054,53 +4080,31 @@ public String getDownloadType() {
...
-//        }
 
         // There's a chance that this is not really a batch download - i.e.,
         // there may only be one file on the downloadable list. But the fileDownloadService
         // method below will check for that, and will redirect to the single download, if
         // that's the case. -- L.A.
+
         this.guestbookResponse.setDownloadtype("Download");
         //RequestContext requestContext = RequestContext.getCurrentInstance();
         PrimeFaces.current().executeScript("PF('downloadPopup').show();handleResizeDialog('downloadPopup');");
```
scolapasta, line 4099: another one here.

src/main/webapp/dataset.xhtml
```diff
@@ -144,41 +145,61 @@
...
                                                         <!-- DOWNLOAD -->
-                                                        <ui:fragment rendered="#{DatasetPage.canDownloadFiles()}">
+                                                        <ui:fragment rendered="#{DatasetPage.canDownloadFiles()  and !(DatasetPage.tooLargeToDownload) }">
                                                             <li class="dropdown-header">#{bundle['dataset.accessBtn.header.download']} <span class="glyphicon glyphicon-download-alt"/></li>
                                                             <!-- NORMAL DOWNLOAD BUTTON (NO TABULAR FILES) -->
-                                                            <li jsf:rendered="#{!DatasetPage.versionHasTabular}">
-                                                                <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadArchival()}" styleClass="btn-download">
+                                                            <li jsf:rendered="#{!DatasetPage.versionHasTabular and !(DatasetPage.tooLargeToDownload)}">
```
scolapasta, line 153: isn't the 2nd part of the rendered always true? (since it's in the ui fragment)

src/main/webapp/dataset.xhtml
```diff
@@ -144,41 +145,61 @@
...
                                                                     <h:outputFormat value="#{bundle['dataset.accessBtn.download.size']}">
                                                                         <f:param value="#{DatasetPage.sizeOfDataset}" />
                                                                     </h:outputFormat>
                                                                 </p:commandLink>
                                                             </li>
+
                                                             <!-- DOWNLOAD ORIGINAL BUTTON (TABULAR FILES PRESENT) -->
-                                                            <li jsf:rendered="#{DatasetPage.versionHasTabular and DatasetPage.sizeOfDatasetOrig != '0 B' }">
-                                                                <p:commandLink update="@form" actionListener="#{DatasetPage.validateAllFilesForDownloadOriginal()}" styleClass="btn-download">
+                                                            <li jsf:rendered="#{DatasetPage.versionHasTabular and DatasetPage.sizeOfDatasetOrig != '0 B'  and !(DatasetPage.tooLargeToDownloadOriginal) }">
```
scolapasta, line 164: same. May make sense to review all the renders in these sections.

scolapasta (commented):
Some additional review.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: Yeah, it's only used for the download mixed popup (some restricted; some not) so we know there are valid files selected so we can just put the openguestbook popup in the onComplete.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: I removed the whole method - see above.

src/main/webapp/dataset.xhtml
sekmiller: The difference here is that we render different command links depending on whether there are tabular files in the dataset - line 153 for no tabular files and 164 below for has tabular files.

scolapasta: I get that, but that explains the tabular check on the renders of the li tags. But I'm talking about the ui:fragment surrounding these li tags. Let me know if it's easier to discuss on a call, viewing the code.

sekmiller: Ok I get what you're saying now. I can remove it.

scolapasta (approved):
@sekmiller Thanks for making those latest changes - I think this hierarchical format is so miuch easier to understand (so maintainable).

I'll approve, but can you add to the QA notes to make sure to test all these different scenarios of original vs archival, etc? Thanks.

PR #7832, opened 2021-04-28
update Sphinx to 3.5.4, jQuery to 3.5.1

### LINKED ISSUE #7233: Support newer version of Sphinx so that Guides Search still works
I acknowledge this issue is a bit slim on content, but I'll do my best to summarize and let more technical folks fill in the details. 

In #7228 we ran into an issue with Guides search that we plan to resolve by doing a local build of the guides on version 1.5.x of Sphinx, which we'll then put up on the guides.dataverse.org server. This course of action was taken because my primary goal was to get search working.

In a future version of Dataverse I'd like to use newer versions (3.2.x) of Sphinx, and I'm happy to assign some sprint time to this. 

Thoughts/discussion welcome.

donsizemore:
**What this PR does / why we need it**: We've currently pegged Sphinx at 1.5.6 because upgrading broke search. This PR updates Sphinx and jQuery without breaking search (or at least, "it's working for me locally").

Closes #7233

**Suggestions on how to test this**: `make html`

**Additional documentation**: will need to upgrade Sphinx on the Jenkins nodes when the time comes, though this PR should allow us to start automatically building guide branches by pull request trigger in GitHub Actions.


doc/sphinx-guides/source/conf.py
```diff
@@ -224,9 +224,10 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
-html_js_files = [
-    'js/jquery-3.4.1.min.js',
-]
+# this breaks $
```
scolapasta, line 227: should we consider just removing this, if it's not needed?

scolapasta (commented):
One small comment on something you commented out; if there's reason to leave, just let me know and I can approve.

doc/sphinx-guides/source/conf.py
donsizemore: I commented out that setting and left it as pseudo-documentation in case anyone wondered why it was removed, and in case it needed to be put back in the future for whatever reason. Happy to remove it if you think I should.

scolapasta: I honestly don't know what it would need to be there for - we always have version control if we need, so I'd vote just remove.

scolapasta (approved):

PR #7842, opened 2021-05-03
7677 reset password banner

### LINKED ISSUE #7677: Reset Password button does not result in info banner feedback, shows no feedback in develop branch

sekmiller:
**What this PR does / why we need it**: Returns feedback to user when they reset their password.

Closes #7677 Reset Password button does not update banner messages.

**Special notes for your reviewer**: simply narrowing the update from "@all" to the form and message panel does the trick.

This is only failing when Google Analytics are in place so be sure the testing environment has a valid analytics file.

curl -X PUT -d '/$PATH/analytics-code.html' $SERVER_URL/api/admin/settings/:WebAnalyticsCode


pdurbin (approved):
Seems like a reasonable fix.

scolapasta (changes requested):
There is another @all in this page; while it may not be broken, let's replace it here too. (it may or need the messages in the "update" - I'll leave that for you to determine)

PR #7848, opened 2021-05-04
7826 deactivated default

### LINKED ISSUE #7826: Add Null Constraint: add null constraint and default value of false to deactivated column of authenticateduser table.
While testing pr #7814 found that creating new users may result in null values for the deactivated column and that breaks the pending sort users pr. So, this pr adds a null constraint to deactivated with default value of false.

sekmiller:
**What this PR does / why we need it**: missed entity update and formal release notes

Closes #7826 Add Not Null Constraint to Deactivated on Authenticated User Table 

**Suggestions on how to test this**: Really nothing to test here. If you could look at the release notes and see if they make sense and are adequate.

**Is there a release notes update needed for this change?**: Added.

**Additional documentation**:


pdurbin (approved):
Looks good.

Issue #7721, opened 2021-03-24
Refactor and clean up the Access (download) API


landreev:
This is a "technical debt" and legacy issue. 

Access is one of the oldest APIs in the application, originally implemented in the context of DVN 3. There is a lot in the code that is archaic and cumbersome and hard to work with. 
Prime examples: 
- Unlike most of our other APIs, file downloads need to serve potentially massive amounts of data. So the content cannot be returned as a buffer or a string, and needs to be streamed. To implement streaming, in most places we are still relying on the code written for early versions of Jersey REST library; that requires providing standalone `MessageBodyWriter` classes (`DownloadInstance.java`, `DownloadInstanceWriter.java` etc.) It would be cleaner to switch to using `javax.ws.rs.core.StreamingOutput` (`/api/access/datafiles`, download-multiple-files-zipped, a newer API, uses it already). 
- There's a system of `OptionalAccessService` classes, meant to support extra download options - extra formats/conversions and such. It was meant to be dynamic/possibly plugin-driven. It became hard-coded in the original Dataverse 4.0. And by now it only makes the code more confusing and difficult to work with.

landreev:
(if we already have open issues for this matter, this issue should supersede them)

PR #7851, opened 2021-05-05
7720 s3 redirects for auxiliary files

### LINKED ISSUE #7720: Enable S3 redirects for "stored originals" of tabular files.
(this should be a simple fix/low-hanging fruit type of an improvement)

Even when S3 redirects are enabled, tabular files are still served locally. This is because of the variable names line (aka "variable name header") that needs to be added to the file contents in real time. (As stored, on disk/s3, an ingested file contains data columns only).
However, this should not be preventing us from issuing redirects for the stored originals; since nothing needs to be added there. 

Note that I will also open an issue to review and reconsider the above-mentioned scheme, of storing the ingested files without the var. name header and adding it in real time. This issue however is for an improvement that can be done quickly and without having to make any complicated decisions.

landreev:
This PR extends support for S3 redirects to the extra (auxiliary) files that are stored in addition to the main physical file: saved originals and cached RData frames for tabular files; resized thumbnails for image files; and uploaded auxiliary files (like the OpenDP ones), for which we don't currently support direct S3 downloads.

Closes #7720

The original issue specifically has the "stored originals" in the title. But, as was discussed during the planning meeting, we wanted to implement this redirect support for all the types where it is possible. (i.e. for all the files that are cached and/or stored on S3). 

No reason NOT to support S3 redirect on any of these file types, is there? 

The implementation is a bit of a mess of special cases, added to the part of the code that's already messy. But I have an open issue for refactoring/cleaning it up for real (#7721). 

Testing it may be a little tedious - all the download types listed will need to be tested, and confirmed that they are now being redirected, AND working properly, as before. (for example, for a saved original that is a stata file, the downloaded file should have the correct name with the .dta extension... etc.)

Any software clients that are currently expecting to just get a 200, and the file, from our API (i.e., not following redirects automatically) may experience temporary difficulties when this is deployed. One case we should probably test is Data Explorer. (it downloads the "prep" file from the application; this will become a redirect as of this PR). 

no

yes

**Additional documentation**:

### DIFF
--- a/doc/release-notes/7720-extended-s3-redirects.md
+++ b/doc/release-notes/7720-extended-s3-redirects.md
@@ -2,10 +2,9 @@
 
 If your installation uses S3 for storage, and have "direct downloads" enabled, please note that, as of this release, it will cover the following download types that were not handled by redirects in the earlier versions: saved originals of tabular data files, cached RData frames, resized thumbnails for image files and other auxiliary files. In other words, all the forms of the file download API that take extra arguments, such as "format" or "imageThumb" - for example:
 
-```
-   /api/access/datafile/12345?format=original
-   /api/access/datafile/:persistentId?persistentId=doi:1234/ABCDE/FGHIJ?imageThumb=true
-```
+`/api/access/datafile/12345?format=original`
+
+`/api/access/datafile/:persistentId?persistentId=doi:1234/ABCDE/FGHIJ?imageThumb=true`
 
 etc., that were previously excluded.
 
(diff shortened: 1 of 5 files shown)


qqmyers (approved):
Makes sense to me

scolapasta:
> Makes sense to me

Agreed, and discussed with @landreev earlier. He is going to add a release note and possibly some minor related doc changes.

(I also alerted @kcondon so he'd be aware).

Code wise, this isn't expected to change and could be QAed.

PR #7860, opened 2021-05-08
#7859 Thumbnail/preview shows as broken on restricted files

### LINKED ISSUE #7859: Thumbnail/preview shows as broken on restricted files
The code in file-info-fragment.xhtml doesn't properly check for permissions and tries to show a thumbnail and preview popup for restricted files. The thumbnail itself is not available so the display shows as broken. PR to follow.

qqmyers:
**What this PR does / why we need it**: Fixes a display issue where broken thumbnail/previewer elements are shown for restricted files (those elements are shown but the files themselves get a 403 response). The fix is to not show those elements when the user doesn't have access.

Closes #7859 

**Additional documentation**:


sekmiller (approved):
Looks good. Thanks!

PR #7865, opened 2021-05-10
Quick fix for #7864 (Simple search of most metadata fields broken)

### LINKED ISSUE #7864: Simple search of most metadata fields broken
The change to using a schema_dv_mdb_copies.xml file breaks the ability to search for most metadata fields in the metadata blocks beyond the most common few (see below for details). For example: https://demo.dataverse.org/dataverse/demo/?q=outerspace returns nothing whereas https://demo.dataverse.org/dataverse/demo/?q=originOfSources:outerspace returns my test dataset, even though originOfSources is one of the fields included in the schema_dv_mdb_copies.xml, which is intended to make the contents of those fields searchable as simple text/without specifying the field name.

The problem is that 1) the schema_dv_mdb_copies.xml file encloses the <copyField> elements in a parent <schema> element, 2) the include statement copies the parent element into the schema.xml file, and 3) solr evidently doesn't read elements within a <schema> sub-element.

This is unlike the schema_dv_mdb_fields.xml file, which includes a <fields> parent element, with solr able to parse additional <field> elements from within that <fields> element after the schema_dv_mdb_fields.xml file is included. (So actually including the fields themselves into the schema works as intended).

The problem has been somewhat hidden by the use of the 'edismax' search handler which automatically searches several fields (including title, subject, keyword, authorAffiliation) in addition to the main _text_ field. This has the effect that searches for words in titles work, even though the title field is not being copied to the _text_ field as it is supposed to. Metadatablock fields not [directly in schema.xml now](https://github.com/IQSS/dataverse/blob/fd1eb33fff86aba8c21694f87651668e653cbaba/conf/solr/8.8.1/schema.xml#L233-L258) and not in [the edismax list](https://github.com/IQSS/dataverse/blob/fd1eb33fff86aba8c21694f87651668e653cbaba/conf/solr/8.8.1/solrconfig.xml#L696-L717), are affected - many citation block fields and all fields in other metadata blocks. (Sites not using the relatively recent schema_dv_mdb_copies.xml aren't affected.)

The problem was discovered at QDR where the edismax search handler isn't used and therefore more fields are affected. (FWIW: https://guides.dataverse.org/en/latest/installation/prerequisites.html says that using this search handler with the same fields as Harvard is optional and the guide describes how to disable it, so other sites could also have more fields affected.)

The quickest solution I can see is to simply cut/paste the <copyField> elements from schema_dv_mdb_copies.xml manually into schema.xml. Changing to dynamic fields, which has been suggested elsewhere should also resolve the issue (and would be preferable since that would avoid the manual update to schema.xml). (Note that the issue is independent of whether schema_dv_mdb_copies.xml is being generated by the update script or managed manually. It's the inclusion of that file with it's enclosing <schema> element that is the problem. One also can't just remove the <schema> element from schema_dv_mdb_copies.xml because XML requires a single root element.) 

I can submit a quick PR to resolve this by putting the elements in schema.xml directly. We may also want to let people know that this works on prior versions as well: if you use schema_dv_mdb_copies.xml, copy the elements into schema.xml, restart solr and do an incremental re-index.

qqmyers:
**What this PR does / why we need it**: Restores the ability to search across all metadata block fields in the simple search box

Closes #7864

**Suggestions on how to test this**: Add a value in a field such as OriginOfSources (see issue and the example there), typing your text in the search box should return no result. Using the PR, and restarting solr and reindexing if using an existing machine, the same search should find the dataset.

**Additional documentation**:


Kris-LIBIS:
How will this affect installations that have customized the metadata blocks if the copyField tags are no longer updated based on the metadatablock?

qqmyers:
The problem and fix are independent of whether your are manually updating the schema*.xml files or using the updateSchemaMDB.sh script. In either case, having the <copyField> entries in the separate schema_dv_mdb_copies.xml files won't work and the entries need to be copied into the schema.xml file itself.

If you have custom metadata blocks, using the script and then adding an extra step to copy the results from the schema_dv_mdb_copies.xml into schema.xml would work.

Editing the script to make it write directly into the schema.xml file would be another option to fix this, as would moving to using the solr api to let dataverse add the fields and copyField requests directly.

pdurbin (changes requested):
I haven't tested to confirm the bug (assuming it's valid, it's a bad one, thanks for finding it!) nor played around with this quick fix but I think before we send this to QA, we need to add some documentation.

I'm aware of two places where updateSchemaMDB.sh is used:

- The last step after enabling a new custom metadata block: https://guides.dataverse.org/en/5.4.1/admin/metadatacustomization.html#updating-the-solr-schema
- In a scenario where Solr has been upgraded and the installation uses custom metadata blocks: https://github.com/IQSS/dataverse/blob/v5.4/doc/release-notes/5.4-release-notes.md#additional-release-steps

These are both still valid scenarios so I think we should explain (probably in the "Updating the Solr Schema" section of the "Metadata Customization" page) the extra step that is necessary, which is what you wrote in a comment:

"If you have custom metadata blocks, using the script and then adding an extra step to copy the results from the schema_dv_mdb_copies.xml into schema.xml would work."

That way, we have something to link to in either of the scenarios mentioned above.

And yes, longer term let's fix up updateSchemaMDB.sh do this extra step or switch to using the Solr API (#5989).

As a side note, a test for this bug would be nice ("is copyField working?") but we can wait for a future pull request.

Kris-LIBIS:
We are relying on the updateSchemaMDB.sh script. We are currently in a phase where users are still discussing about the metadata schema and we do updates to the citation block on a regular base and sometimes even reset our pilot installation database and index from scratch. We are also running dataverse and solr in docker containers. The way it is currently setup, the schema files are stored in the container itself and any changes are lost when the container restarts. I therefore added a script that performs the metadata schema update at startup to get it synchronized with the current state of the metadata blocks. When a metadata block is updated, I just restart the solr container(s) and we're done.

I also had the intention to use the script at least during installation of the production repository.

BTW, the updateSchemaMDB.sh script as it is now, does not work - at least for us. I have already fixed it locally and was about to create an issue and PR, but I'm holding off for now. With this change in place, it needs to incorporate the changes into the schema.xml and I'd like to tackle that first.

qqmyers:
@Kris-LIBIS - do you have time to fix the script for this? All that should be needed is to change creating the separate schema_dv_mdb_copies.xml file with scanning through schema.xml and replacing all of the lines between the 
`<!-- Dataverse copyField from http://localhost:8080/api/admin/index/solr/schema -->`
and
`<!-- End: Dataverse-specific -->`
with the same set of <copyField> elements. (No wrapping <schema> element  as in the copies files though.)

Kris-LIBIS:
@qqmyers - yes, that is what I was planning to do. But give me another week or so. I'm in the middle of upgrading our installation to 5.4.1 and the Solr upgrade is hurting me. I have to sort that out first.

pdurbin:
@Kris-LIBIS thanks for being willing to work on this! For now I co-assigned you to this pull request but you are also very welcome to make a fresh one or whatever works. I think review is done for now so I'm going to move it to the "community dev" column.

pdurbin (approved):
This looks good now that a release note has been added. Thanks @qqmyers 

I'm going to go ahead and move this to QA as-is. It's important to get the word out.

@Kris-LIBIS we are still very much interested in your fixed to the script but please make a fresh pull request when you are ready.

Kris-LIBIS:
@pdurbin I have a fix for the script running locally for testing. But I was wondering why don't we do the same for the <field> entries? We could as well add them directly to the schema.xml too.

pdurbin:
@Kris-LIBIS that's great news! Stopping now might be good because @poikilotherm and @pkiraly are having a little Solr-related hackathon on Friday. I'm sure you'd be very welcome to join!

pdurbin:
@Kris-LIBIS when I met with @poikilotherm and @pkiraly on Friday we didn't actually talk about the script much. The focus was on #5989 (Add new Solr fields via an API call).

I think it would be best to go ahead and create a pull request for what you have working so we can take a look. Thanks!

Kris-LIBIS:
Issue #7903 and PR #7904 created.

PR #7881, opened 2021-05-19
7766 linking dv bug

### LINKED ISSUE #7766: Linking a dataverse into a "grandchild" of root not working
If you link a dataverse to a dataverse that is a direct child of root, everything works fine. However, if you link to a child of a child of root, then the linked dataverse (nor its dataverse or datasets) show. Note, that the link is correctly created in the db, so this appears to be an indexing issue.

This is not working wither in current production (5.3) or develop (5.4).

sekmiller:
**What this PR does / why we need it**: Fixes the indexing of datasets and dataverses to dataverses that are more than one generation removed from Root DV

Closes #7766 Linking a dataverse into a "grandchild" of Root not working

**Special notes for your reviewer**: also removes extraneous paths not needed for display. see original ticket for more info.

**Suggestions on how to test this**:  verify that linking to dataverses more than one level down from Root DV works/(shows linked datasets and dataverses in the linking DV) Also that re-index all and re-index individual DV retains links to linked dataverses.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1472,32 +1493,15 @@ private void updatePathForExistingSolrDocs(DvObject object) throws SolrServerExc
...
-
+        /*
+        First get all linking dataverses
+        Then get their respective paths
+        And finally add to the "normal" path
+        */
+        List<Dataverse> linkingDataverses = findAllLinkingDataverses(dvo);
+        List<String> linkingDataversePaths = findLinkingDataversePaths(linkingDataverses);
         for (String dvPath : linkingDataversePaths) {
             dataversePaths.add(dvPath);
```
scolapasta, line 1506: can this be simplfied? since findLinkingDataversePaths reutns a List<Strings> just add that directly to dataversePaths (with addAll) instead of looping through?

scolapasta (approved):

PR #7245, opened 2020-08-31
5292 mvn container

poikilotherm:
**What this PR does / why we need it**:
This pull request will add official images to the upstream repository.

1. There is no way around this, as any modern container building tools reusing the Maven environment have to reside in the projects `pom.xml`. 
2. Adding the necessary bits here allows to reuse `pom.xml` as the source of truth by reusing its dependency versions written down in properties.
3. Providing this here allows for using any version/commit of Dataverse in a container right away. Building is fast, reusing layers and local Maven cache.
4. Images from this will be reusable with Testcontainers and Docker-Compose.

**Which issue(s) this PR closes**:
Closes #5292

**Special notes for your reviewer**:
- Build with `mvn -P container package docker:build`
- Building the Solr image is still TBD.
- Fun fact: is Dataverse 5 in general much faster to deploy or is it the new layer arch providing deployment times of about 45 secs?

**Suggestions on how to test this**:
Nothing yet.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
No.

**Is there a release notes update needed for this change?**:
Later on.

**Additional documentation**:
We should debate if @IQSS is willing to provide testing and QA for this type of stuff. Adding a label for good distinction.

PR #7904, opened 2021-05-26
Fix for #7903 update schema.xml in place

### LINKED ISSUE #7903: Update schema.xml in-place
With PR #7865 in place, the updateSchemaMDB.sh script no longer works as intended. It used to update the Solr schema by updating two XML files that were included in the schema.xml. Issue #7864 indicated that this update mechanism did not work and after the fix, the schema.xml file should be updated manually.

In order to restore the fully 'automatic' update without breaking #7864, the updateSchemaMDB.sh script should modify the schema.xml directly instead of updating the included XML files.

Kris-LIBIS:

With PR #7865 the updateSchemaMDB.sh no longer works. This PR restores that functionality without breaking #7864 that the PR fixed. Some institutions may rely on the script to automate their metadata customization.

Closes #7903

This PR also removed the schema_dv_mdb_fields.xml file and included its contents directly in the schema.xml. The updated script will however still work with older schema.xml files that still have the include lines in them.

Perform some customization on the metadata blocks - removing or adding metadata fields and/or blocks - and run the script. The schema.xml should contain an updated list of <field> and <copyField> entries.

No UI changes.

Probably. Some of the changes for PR #7865 in release 5.5 release notes and documentation will need to be updated.

**Additional documentation**:

I have not yet looked at the documentation changes required, but overall it should be simplified.


qqmyers:
I didn't run this but the changes look straight forward and I think with a Q/A check that it does what's intended can quickly verify that. The one thing I'd suggest checking for - if you don't specify a target on the command line, the script defaults to the /tmp dir and worked (creating the two new files there) - does the updated script handle that case some way (i.e. if schema.xml isn't in the target dir). Just warning that schema.xml doesn't exist in that dir and quitting (or similar is fine) but it would be good to make sure the user gets a useful message.

Kris-LIBIS:
@qqmyers That is a very good point you made. I did not realize this before, but the meaning of the TARGET directory modified ever so slightly from 'the location to write the included XML files' to 'the location of the schema.xml file'.

In most installations this will not make a difference, but that may not be the case for everyone. Notably in @poikilotherm 's kubernetes Dockerfile for solr the included XML files will be generated in a separate directory and are included in the schema.xml via symlinks.

Maybe we should leave the original script alone and create a new script instead? updateSchemaInline.sh for instance?

On the other hand, with the changes in #7864, the old script no longer does what it was originally expected to do anyway. The changes in this PR would make the script work again as expected, if some changes are made to the Dockerfile.

What would do you suggest?

qqmyers:
Since the original script won't work for Docker either, I'd suggest just having one script and getting the Dockerfile updated. That said, I haven't worked with the Docker version, and others who do may have a different/better idea. In any case, we should avoid delaying getting this out - it will simplify setup for those who haven't upgraded yet.

conf/solr/8.8.1/schema.xml
```diff
@@ -228,7 +228,166 @@
     <field name="filePersistentId" type="text_en" multiValued="false" stored="true" indexed="true"/>
     
     <!-- Dynamic Dataverse fields from http://localhost:8080/api/admin/index/solr/schema -->
```
poikilotherm, line 230: Can we please create more visible Search & Replace Guards here? Someone might delete that by accident, then things break.

Same for the `<copyField>` part.

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
+fi
+
 ### Retrieval
 echo "Retrieve schema data from ${DATAVERSE_URL}/api/admin/index/solr/schema"
 TMPFILE=`mktemp`
 curl -f -sS "${DATAVERSE_URL}/api/admin/index/solr/schema${UNBLOCK_KEY}" > $TMPFILE
 
 ### Fail gracefull if Dataverse is not ready yet.
-if [[ "`wc -l ${TMPFILE}`" < "3" ]]; then
+if [[ `wc -l < ${TMPFILE}` -lt 3 ]]; then
```
poikilotherm, line 72: We should start using `$()` here and let [`shellcheck`](https://www.shellcheck.net/) bark at other things it finds.

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
 
-### Processing
-echo "Writing ${TARGET}/schema_dv_mdb_fields.xml"
-echo "<fields>" > ${TARGET}/schema_dv_mdb_fields.xml
-cat ${TMPFILE} | grep ".*<field" >> ${TARGET}/schema_dv_mdb_fields.xml
-echo "</fields>" >> ${TARGET}/schema_dv_mdb_fields.xml
+### Processing field entries
+echo "Replacing field lines"
+START='<!-- Dynamic Dataverse fields from http://localhost:8080/api/admin/index/solr/schema -->'
+INCL='<xi:include href="schema_dv_mdb_fields.xml" xmlns:xi="http://www.w3.org/2001/XInclude" />'
```
poikilotherm, line 80: Don't. Just let it fade away in the new release. Add a release note with a hint about this.

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
+END='<!-- End of Dynamic Dataverse fields -->'
+
+### -- create temp file with new content
+TMPF=`mktemp`
+echo "$START" > ${TMPF}
+cat ${TMPFILE} | grep ".*<field" >> ${TMPF}
```
poikilotherm, line 86: I don't think this is necessary anymore. Let's just save the filtered multiline string in a var to be used in the replacement text of `sed`

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
+echo "$END" >> ${TMPF}
+
+### -- first replace INCL with END if it is still there
+sed -e 's@'"$INCL"'@'"$END"'@' -i ${TARGET}/schema.xml
+
+### -- replace START->END with content of temp file
+sed -e '\@'"$START"'@,\@'"$END"'@!b' -e '\@'"$END"'@!d;r '"$TMPF" -e 'd' -i ${TARGET}/schema.xml
```
poikilotherm, line 93: Might become easier if just sending the variable into the command, right?

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
+
+### -- remove temp file
+rm "${TMPF}"
+
+### -- field processing done
+echo
+
+### Processing copyField entries
+echo "Replacing copyField lines"
+START='<!-- Dataverse copyField from http://localhost:8080/api/admin/index/solr/schema -->'
```
poikilotherm, line 103: Same as outlined above: we should use guards around all of this that are self-explaining and less prone to be deleted by accident.

conf/solr/8.8.1/updateSchemaMDB.sh
```diff
@@ -56,27 +56,67 @@ if [ ! -z "${UNBLOCK_KEY}" ]; then
...
+END='<!-- End: Dataverse-specific -->'
+
+### -- create temp file with new content
+TMPCF=`mktemp`
+echo "$START" > ${TMPCF}
+cat ${TMPFILE} | grep ".*<copyField" >> ${TMPCF}
```
poikilotherm, line 109: Same as above: maybe just send to a var instead of tmpfile.

poikilotherm (changes requested):
LGTM with some minor tweaks.Thank you for working on and fixing this @Kris-LIBIS!

About Docker/K8s: yes, these community projects follow this repos lead, so they will need to adapt.

Will become part of #7245

conf/solr/8.8.1/schema.xml
Kris-LIBIS: Fine with me. I was using whatever already was there to make sure the script works with schema.xml files that are already out there. Since that that is no concern, I'll add better markers in the schema.xml.

PR #7907, opened 2021-05-27
#7906 delete extra space from the end of an API success message.

### LINKED ISSUE #7906: When delete a user via API the success message contain an extra space.
edu.harvard.iq.dataverse.api.Admin contains the following line:

```Java
return ok("AuthenticatedUser " + au.getIdentifier() + " deleted. ");
```

It contains an unnecessary extra space at the end on the message, which we should get rid of

pkiraly:
**What this PR does / why we need it**: When an administrator delete a user, s/he receives a JSON message containing a string: "AuthenticatedUser [identifier] deleted. ". At the end of this message there is an unwanted extra space.

Closes #7906 

**Special notes for your reviewer**: no

**Suggestions on how to test this**: no

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
@@ -379,9 +379,8 @@ public class Admin extends AbstractApiBean {
         authSvc.removeAuthentictedUserItems(au);
         
         authSvc.deleteAuthenticatedUser(au.getId());
-        return ok("AuthenticatedUser " + au.getIdentifier() + " deleted. ");
-
-    }  
+        return ok("AuthenticatedUser " + au.getIdentifier() + " deleted.");
+    }
 
     @POST
     @Path("authenticatedUsers/{identifier}/deactivate")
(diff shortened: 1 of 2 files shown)


scolapasta (approved):

pkiraly:
I extended a test to cover the success message content.

scolapasta:
> I extended a test to cover the success message content.

I don't think this is needed (we already check the status code). The actual message isn't critical, just success of failure. This is actually a great example, had the test already contained this assert, then it would have failed for the removing of a space -  having that fail adds no (or at least little) value.

pkiraly:
> I don't think this is needed (we already check the status code).
@scolapasta Yes, it did not have a real value per se, but the test would be a safeguard in the future, when somebody want to change the content of the message, e.g. change username to email address. A test proves that there is an expectation regarding to the output, and changing the message might involve changes in the client software and documentation.
If you still think testing it is useless, I delete the test changes.

scolapasta:
> > I don't think this is needed (we already check the status code).
> > @scolapasta Yes, it did not have a real value per se, but the test would be a safeguard in the future, when somebody want to change the content of the message, e.g. change username to email address. A test proves that there is an expectation regarding to the output, and changing the message might involve changes in the client software and documentation.
> > If you still think testing it is useless, I delete the test changes.

@pkiraly yes, let's remove - alternatively, if you think it's important output, we should return it as part of a more structured json and not as just part of the message itself. (but that could also be out of scope for now)

pkiraly:
@scolapasta done

scolapasta (approved):

pkiraly:

PR #7908, opened 2021-05-27
Iqss/1724 anonymized access

### LINKED ISSUE #1724: Allow For Anonymous Peer Review of a Dataset for Journals
Palgrave Communications journal and many other journals have a blind / anonymous peer review process for their journal articles and any underlying research data. To support this in Dataverse we would need to allow for datasets to hide the names/affilitation of authors for an anonymous peer review of data to take place. 

This is somewhat related to #1012

qqmyers:
**What this PR does / why we need it**: This PR enables a new option when creating a Private URL - to create one that only allows 'anonymized access'. When a user accesses a dataset using this type of Private URL:

- author names are withheld from the dataset and file citations
- citation downloads are disabled
- the version tables (dataset and file) do not show contributor names
- a configurable set of metadata fields (e.g. author, datasetContact, depositor, contributor, publication, and/or grantNumber) are also 'withheld'/their values not shown
- metadata downloads are disabled

Anonymized access is intended for situtations such as making a dataset available for anonymized journal review.

Closes #1724

**Special notes for your reviewer**: The code includes a mechanism to add a boolean flag to private url role assignments to track whether the privateURL is limited to anonymized access along with UI/API code to request this type of url, and the changes in .xhtml files to not render the relevant page parts. An :AnonymizedFieldTypeNames property (comma separated list) allows configuration of which fields (and potentially ones in custom blocks) that should be anonymized. As of now, there is no default list of fields to hide - we could add a default set and/or only show the anonymized access url option if the property is set.

Note that with the addition of code to limit API access (to avoid leaking metadata), previewers that try to retrieve metadata (all our standard ones) also fail. Since the don't actually display metadata when used in an iframe, they don't need to make this call in that case. I have a modification to the previewer javascript that will handle this and restore the in inframe views. (Changing to use signed URLs could avoid this as well.) In any case - if the solution here is merged, we should make the previewer change as well. Alternately, a mechanism that removes anonymized fields from the json returned by various API calls could be developed instead.

**Suggestions on how to test this**: Create a PrivateURL with/without anonymization and verify that access using the URL works in both cases and that, in the anonymized case, the parts listed above are not included.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: changes to the dialog to create a PrivateURL, changes in the dataset/file views to indicate the withheld info listed above.

**Is there a release notes update needed for this change?**: The existing of support for anonymized access to draft datasets could be announced. No default changes are required.

**Additional documentation**: documentation of the setting and user doc on creating an anonymized access url are included.


adam3smith:
From presentation comments at DV community meeting:

> Hi! When I downloaded the metadata (xml) file for the dataset, it shows your name as the author. Is there a way to also anonymize the metadata file?

qqmyers:
I assume this via API? (Any metadata download buttons are turned off afaik). The API may still not be checking the flag on the private URL.

adam3smith:
@qqmyers Datafile citation metadata is available & not anonymized

djbrooke:
Thx @qqmyers for the PR. A developer should feel free to review the code here, but I believe you and @TaniaSchlatter were also discussing a walkthrough of the functionality, so I'll assign her as well. Happy to join in - this is a neat feature.

pdurbin:
@qqmyers great demo! One thing you may not be aware of is that I wrote developer-oriented docs at https://github.com/IQSS/dataverse/blob/v4.5/src/main/java/edu/harvard/iq/dataverse/privateurl/package-info.java . I don't know if you're interested in updating them for this pull request but you should definitely feel free!

qqmyers:
Per demo/review, I've added logic to disable the anonymized access button if there's a published version and added a popup title that only appears when the button is disabled:
image

Also added the same check to the command so the API will also enforce this.
Moved errors strings in the command to the bundle as well for i18n
Added notes to package-info
Updated tests to not check exact error messages, added test for anonymized and released (should fail)

sekmiller:
We need a sql script to add privateurlanonymizedaccess to RoleAssignment

src/main/java/edu/harvard/iq/dataverse/RoleAssignment.java
```diff
@@ -71,13 +71,21 @@
     @Column(nullable = true)
     private String privateUrlToken;
 	
+    @Column(nullable = true)
+    private Boolean privateUrlAnonymizedAccess;
+
```
sekmiller, line 76: We need a sql update script for this new field:

ALTER TABLE RoleAssignment ADD COLUMN IF NOT EXISTS privateUrlAnonymizedAccess BOOLEAN;

UPDATE RoleAssignment SET privateUrlAnonymizedAccess = false WHERE privateUrlAnonymizedAccess IS NULL;

qqmyers: Good catch. I don't think the second line is needed (at least it works for QDR without it).

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -362,7 +362,16 @@ protected User findUserOrDie() throws WrappedResponse {
             return GuestUser.get();
         }
         PrivateUrlUser privateUrlUser = privateUrlSvc.getPrivateUrlUserFromToken(requestApiKey);
+        // For privateUrlUsers restricted to anonymized access, all api calls are off-limits except for those used in the UI
+        // to download the file or image thumbs
         if (privateUrlUser != null) {
+            if (privateUrlUser.hasAnonymizedAccess()) {
+                String pathInfo = httpRequest.getPathInfo();
+                if (!(pathInfo.startsWith("/access/datafile/") && !pathInfo.substring(17).contains("/"))) {
```
sekmiller, line 370: I'm wary of using "17" here. will this fail if we change the access path in any way? is there another way to express this?

TaniaSchlatter:
Thanks for the demo and making the changes @qqmyers. Looks good to me. I took screenshots and will add them to the issue for reference.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
qqmyers: "/access/datafile/".length() ?

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -1736,6 +1741,7 @@ file.dataFilesTab.versions=Versions
 file.dataFilesTab.versions.headers.dataset=Dataset
 file.dataFilesTab.versions.headers.summary=Summary
 file.dataFilesTab.versions.headers.contributors=Contributors
+file.dataFilesTab.versions.headers.contributors=Contributor name(s) withheld
```
sekmiller, line 1744: missing .withheld here

sekmiller (approved):
Looks good. just one missing tag in the bundle and merge the latest from dev

PR #7916, opened 2021-05-31
Account migration

janvanmansum:
**What this PR does / why we need it**: Currently, when a user logs in to a migrated builtin user account for which the password is encrypted with a legacy hash algorithm (e.g., SHA-1), Dataverse will display a password reset page that also requires the user to accept the Terms of Use again. After a new password is provided it gets encrypted with the latest and greatest algorithm.

There are several limitations and issues with the current reset page:
* The introductory message talks about an upgrade from Dataverse 4.0, which is not necessarily the case.
* There is no way to only re-encrypt the current user password, i.e. the user is forced to provide a new one. However, for example DANS would like to just require the user to accept the Terms of Use again and not change the password.

Example of the current screen:
image

**Which issue(s) this PR closes**: N/A

**Special notes for your reviewer**: ~This PR introduces the following database settings:~
* ~`:SilentPasswordAlgorithmUpdateEnabled`: if set to `true`, the "New Password" and "Retype Password" entry fields are hidden and the existing password is automatically re-encrypted on clicking the continue button.~
* ~`:PVCustomPasswordResetAlertMessage`: the text that should appear in the blue banner at the top of the passwordreset page.~
* ~`:CustomPasswordResetButton`: the caption of the button to continue. (If `:SilentPasswordAlgorithmUpdateEnabled` is enabled you should override the default text, which is "Reset Password", with something else, for example "Continue")~
* ~`:CustomPasswordResetAlertIntro`: the introductory text.~

*Settings have been removed at the request of @scolapasta (see https://github.com/IQSS/dataverse/pull/7916/#issuecomment-908730125 and https://github.com/IQSS/dataverse/pull/7916/#issuecomment-917026398)*

Note that, if the password turns out not to comply with the minimal security requirements, the user is presented with the regular password reset page on clicking Continue, and will be required to provide a new password.

1. ~Configure the Dataverse instance to automatically reencrypt password, e.g.:~
   ```bash
    curl -X PUT -d "true" http://localhost:8080/api/admin/settings/:SilentPasswordAlgorithmUpdateEnabled
    curl -X PUT -d '{0} Our Terms of Use have been updated:{1} Please check the box below and click Continue' \
         http://localhost:8080/api/admin/settings/:PVCustomPasswordResetAlertMessage
    curl -X PUT -d "Continue" http://localhost:8080/api/admin/settings/:CustomPasswordResetButton
    curl -X PUT -d 'Please accept our new Terms of Use.  Also, note that you can connect your account to your institutional account, or ORCID, Google or GitHub account. To do so, log out after completing this step. Log in again with one of “Log in with our institutional account”, “ORCID”, “Google” or “GitHub”. ' http://localhost:8080/api/admin/settings/:CustomPasswordResetAlertIntro
    ```
2. Create an SHA-1 password, e.g.,  `echo -n mypassword001 | openssl dgst -binary -sha1 | openssl base64` 
3. Update the `builtinuser` table: 
    ```sql
    UPDATE builtinuser 
    SET encryptedpassword = '<result of previous command>'
    WHERE username = '<user to edit>';

    UPDATE builtinuser 
    SET passwordencryptionversion = 0
    WHERE username = '<user to edit>';
    ```
4. Log in. You should see a page similar to the one below:
    image

(Note: you can customize the build number with the script `dataverse/scripts/custom-build-number` in the Dataverse code.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: See previous item.

**Additional documentation**:


djbrooke:
Hi @janvanmansum, thanks for sharing this. I appreciate the simplification of the workflow here. 

Just so I understand, when the user encounters this screen in the new Dataverse installation, will their passwords have already been loaded into the Dataverse installation's database? Would this only be able to be used by installations where passwords from the old system were known and transferred?

Regarding the 4.0 text, I've adjusted the text on a currently-unmerged branch so that the old version is not referenced specifically.

djbrooke:
Hi @janvanmansum, good to see you at the community meeting last week. I'm just making sure you saw my note above. Any thoughts?

janvanmansum:
@djbrooke : sorry for the late response. Yes, this assumes that the user password is in the database already (just not encrypted with the preferred algorithm).

pdurbin:
@janvanmansum hi, I'm just checking if this pull request is ready for review or if you're still working on it. I noticed a new commit from a couple weeks ago.

djbrooke:
Thanks @qqmyers for discussing, and thanks @janvanmansum for following up. I think we both missed each other a few times. :) 

I want @scolapasta to weigh in before we decide the merge this because we talked in some detail (prompting my question above in https://github.com/IQSS/dataverse/pull/7916#issuecomment-856253074) - if there's a specific security concern about implementing this, let's discuss.

scolapasta:
Hi, sorry for the delay in getting to this! I think the idea makes sense overall, but I wonder about having these settings? i.e. when we first added this we knew that we were requiring stricter passwords which was one reason to force the new password. But since this incorporates that (i.e. it forces new password if not strict enough), let's just make this the default behavior. In that way we can save a few settings and have all installations operate this way.

In short, rather than have this provide an alternative approach, let's have that alternative be standard.

@janvanmansum would you be able to make these changes?

janvanmansum:
@scolapasta : OK, then we will remove the settings and make the behavior the same as currently with the following values:

* SilentPasswordAlgorithmUpdateEnabled = true
* PVCustomPasswordResetAlertMessage = "Please, accept the new terms of use before continuing"
* CustomPasswordResetButton = "Continue"
* CustomPasswordResetAlertIntro = "Our terms of use have changed, please check the box below and click Continue"

janvanmansum:
@scolapasta : @mderuijter made the requested changes. Please, have another look at it. Thanks!

djbrooke:
Thanks @janvanmansum, we'll take another look!

src/main/java/edu/harvard/iq/dataverse/LoginPage.java
```diff
@@ -166,6 +167,9 @@ public String login() {
         for ( FilledCredential fc : filledCredentialsList ) {       
             authReq.putCredential(fc.getCredential().getKey(), fc.getValue());
         }
+
+        FacesContext.getCurrentInstance().getExternalContext().getFlash().put("silentUpgradePasswd",authReq.getCredential(BuiltinAuthenticationProvider.KEY_PASSWORD));
```
scolapasta, line 171: what exactly are we passing here via flash? Is it the password (encrypted or not, should we have any concerns about security here?)

tagging @qqmyers also to see his thoughts

scolapasta (changes requested):

src/main/java/edu/harvard/iq/dataverse/LoginPage.java
qqmyers: As far as I know, flash is like a session in that the value isn't sent to the browser (unless you use the variable in the display somewhere - that is done here but to a password field which blanks the display.) And the value only stays around for one redirect (so better than a session).

I think returning the password to the reset page could be done, but I'm not sure there's much benefit and we'd lose things like the validation against current password rules that is triggered now (same as if the user types a new password). (Basically if you can see the password in the reset page https transmission, you could have gotten it when the user entered it on the login page.)

One suggestion I would make is to at least not set it unless the redirect to password reset will happen, line 212 below if I'm read code correctly. It wouldn't be used in other cases, but no need to set it either.

mderuijter: @janvanmansum and I tested the flash variable to see if it remained in the FacesContext after a redirect. But as soon as the `newPassword` variable was set in `PasswordResetPage.java` line 106, the value was gone. I've implemented @qqmyers last suggestion so that the password only gets set when that redirect really happens, making the risk of exploitation even lower I think.

scolapasta (approved):
Looks good - thanks for the PR!

djbrooke:
@scolapasta @janvanmansum I'll move this back to review until we can take a look at the test failures. I can't access Jenkins right now, but as soon as I can I'll add the information here.

djbrooke:
(rerunning tests now)

djbrooke:
@janvanmansum Can you update from develop to capture the fixed tests? Once those run successfully we can move this to QA. Thank you!

janvanmansum:
Done by @mderuijter , merged to PR branch by me.

djbrooke:
Great, thanks @janvanmansum and @mderuijter. Tests pass!

pdurbin (changes requested):
It looks like :PVCustomPasswordResetAlertMessage has been removed from the code but not the guides. At the very least, it should be removed from the guides and a release note should be added saying that the setting was removed.

scolapasta:
Actually before removing :PVCustomPasswordResetAlertMessage from the guide - it seems there may have been some confusion?

This setting existed before this PR and was not added here. The request I made during review was for removing the new settings you added. The confusion, I think is that in the initial comment this setting was listed under the "This PR introduces the following database settings", so seemed like it was also new and maybe why you remove dit?

However, @pdurbin looked and it seems as if it was actually added in: https://github.com/IQSS/dataverse/pull/4128

So the question I guess should be, should this one setting be left alone and not removed since it predates this PR and arguebly has a use outside of this PR? Or does this PR also remove the need for it (and if so, it should be removed from the guides?

@janvanmansum @mderuijter Could you review and let us know your thoughts? Thanks!

janvanmansum:
@scolapasta : you are right, `:PVCustomPasswordResetAlertMessage` was never part of this PR, so I have restored it now.

@pdurbin @mderuijter

kcondon:
@janvanmansum Thanks for the update. I think @scolapasta still needs to review but I decided to do a quick test of the password complexity workflow. It appears it is not functioning correctly in your branch at the moment. What happens is if I change :PVMinLength to be 10 rather than 6 and log in with an existing account, without setting :PVCustomPasswordResetAlertMessage, I see different results:

On develop branch, I see this:
Screen Shot 2021-10-01 at 2 29 13 PM

On your account migration branch, I see this:
Screen Shot 2021-10-01 at 2 41 37 PM

scolapasta (approved):

scolapasta:
@janvanmansum @mderuijter could one of you get the latest from develop (now that 5.8 has been released?)

janvanmansum:
@scolapasta : done.

janvanmansum:
@pdurbin @scolapasta : I have merged back the latest `develop` commits to this branch.

PR #7923, opened 2021-06-04
Fix #7922 translation for subfield with controlled vocabullary

### LINKED ISSUE #7922: no translation for subfield with controlled vocabulary
When I change display language, fields with controlled vocabulary are translated, but not the subfield with controlled vocabulary.

English version:
image

French version:
Capture d’écran 2021-06-04 à 19 38 47

tcoupin:

Closes #7922 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetField.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetField.java
@@ -241,7 +241,7 @@ public class DatasetField implements Serializable {
             return datasetFieldValues.get(0).getValue();
         } else if (controlledVocabularyValues != null && !controlledVocabularyValues.isEmpty()) {
             if (controlledVocabularyValues.get(0) != null){
-                return controlledVocabularyValues.get(0).getStrValue();                
+                return controlledVocabularyValues.get(0).getLocaleStrValue();                
             }
         }
         return null;


tcoupin:
Capture d’écran 2021-06-04 à 19 34 20

scolapasta (approved):
Looks good!

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValue.java
```diff
@@ -122,7 +122,7 @@ public String getUnsanitizedDisplayValue() {
             if (StringUtils.isBlank(format)) {
                 format = "#VALUE";
             }           
-            String value = this.getValue();    
+            String value = this.getDisplayValue();
```
qqmyers, line 125: @tcoupin - this doesn't seem related to the issue - is it solving a different problem or is it not needed? (FYI - I'm working on a few i18n issues at the moment and am making changes to getDisplayValue and related calls so I want to make sure I understand what the issue is here.)

qqmyers: @tcoupin - checking again - this looks like it calls the getDisplayValue() method which will sanitize the HTML as part of the getUnsanitizedDisplayValue() method which seems odd. I haven't looked through the call chain but presumably the callers want unsanitized values here and if that's a problem, perhaps the calling methods should use getDisplayValue instead of getUnsanitizedDisplayValue instead of making this change?

tcoupin: Hi @qqmyers ,
sorry for my late response. I made this fix for 4.20 so I don't remember why I made this change bu I think it's related to dropdown list https://github.com/IQSS/dataverse/blob/a875db778a53db3353e6b3095e52bdbe6fff01e9/src/main/java/edu/harvard/iq/dataverse/DatasetField.java#L323

qqmyers: @tcoupin - since a lot has changed since 4.20, can you drop that line from the PR and we can get the other change merged? I'm definitely concerned about calling the method to sanitize the values inside the getUnsanitizedDisplayValue method  - it might have fixed a 4.20 issue but presumably something that needs unsanitized values would break with this change. If it is still needed to fix a bug in 5.x, it may make more sense to, for example, change line 323 in your comment above rather than changing the getUnsanitizedDisplayValue method itself.

tcoupin: I remove the change on line 125 and it's still working.
I've just rebased on develop and pushed my commit

kcondon:
@tcoupin This works, thanks! @qqmyers Any review comments?

qqmyers:
Nope - good to go!

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValue.java
pdurbin: >  I've just rebased on develop and pushed my commit

Just a quick comment that I think we could learn from @tcoupin about rebasing and keeping a clean git history. It's nice that this pull request is just a single commit with only the change we want and no merge commits. Thanks!

tcoupin:
Many projects on github make rebase on master before accepting pull request. They generaly squash commits before accept. 


De: "Philip Durbin" ***@***.***> 
À: "IQSS/dataverse" ***@***.***> 
Cc: "thibault coupin" ***@***.***>, "Mention" ***@***.***> 
Envoyé: Mardi 29 Juin 2021 17:20:32 
Objet: Re: [IQSS/dataverse] Fix #7922 translation for subfield with controlled vocabullary (#7923) 





@pdurbin commented on this pull request. 


In [ https://github.com/IQSS/dataverse/pull/7923#discussion_r660724080 | src/main/java/edu/harvard/iq/dataverse/DatasetFieldValue.java ] : 
> @@ -122,7 +122,7 @@ public String getUnsanitizedDisplayValue() {
             if (StringUtils.isBlank(format)) {
                 format = "#VALUE";
             }           
-            String value = this.getValue();    
+            String value = this.getDisplayValue(); 




I've just rebased on develop and pushed my commit 



Just a quick comment that I think we could learn from [ https://github.com/tcoupin | @tcoupin ] about rebasing and keeping a clean git history. It's nice that this pull request is just a single commit with only the change we want and no merge commits. Thanks!

Issue #6751, opened 2020-03-19
Use the translation of a controlled vocabulary in the oai_ddi


bappun:
In our dataverse we use the identifiers of the controlled vocabularies as their value in the TSVs to be able to translate them easily. However these identifiers are not really human readable and are present in the output of the oai_ddi.

Would it be possible to use the translation of the controlled vocabularies in the oai_ddi output instead of their default value (in my case the identifier)?

For example in the oai_ddi output we have:
`<sampProc>TotalUniverseCompleteEnumeration</sampProc>`
But we would like the value present in our languages.zip:
`<sampProc>Total universe/Complete enumeration</sampProc>`

Maybe it is a bit too specific, but if you have any insight I will be glad to read them.

djbrooke:
Hey @bappun, that's a good question. I'll defer to @jggautier (who may defer to a developer) but maybe we remove spaces and non-alphanumeric characters form this output because of interoperability reasons?

jggautier:
Sorry for the late reply @bappun.

I'd like to make sure I understand, using the example you gave: When someone views the English version of your repository, on the metadata edit she sees the English version of the vocabulary terms of the Sampling Procedure metadata field, one of the terms being "Total universe/Complete enumeration", right? And that string is ultimately coming from your repository's languages.zip?

But currently the code that creates the DDI XML export is using the string that's in the metadatablock TSV file, "TotalUniverseCompleteEnumeration", which is supposed to be the language-agnostic identifier for that term?

I agree that "Total universe/Complete enumeration" should be what's used in the DDI xml export, and not "TotalUniverseCompleteEnumeration". But I don't know much about how these language files interact with the TSV files, so I'd have to defer to a developer who's familiar with the internationalization changes over the past couple of years.

bappun:
Hi @jggautier 

You're right. I used identifiers as the vocabulary values in my TSV and I'd like to be able to output their translations in the oai_ddi instead of their values so we can have the oai_ddi in different languages.

Thanks!

PR #7958, opened 2021-06-21
SPO/7388 lang attributes in ddi export

### LINKED ISSUE #7388: Language missing in the OAI-DDI
Most of our datasets are described as "french" datasets in the metadata. For example: https://data.sciencespo.fr/dataset.xhtml?persistentId=doi:10.21410/7E4/00LYOG (detailed metadata are embedded in the dataset [here](https://data.sciencespo.fr/api/access/datafile/305)).

We imported a DDI file where the French language is set at a file level also at a study level:
```
<?xml version="1.0" encoding="UTF-8"?>
<codeBook xml:lang="fr" xsi:schemaLocation="ddi:codebook:2_5 https://www.ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd" version="2.5" ID="fr.cdsp.ddi.OV70"
	xmlns="ddi:codebook:2_5">
```
```
<stdyDscr>
	<citation>
		<titlStmt>
			<titl xml:lang="fr">L'ouvrier français en 1970</titl>
			<parTitl xml:lang="en">The French Working Class in 1970</parTitl>
			<IDNo agency="CDSP">fr.cdsp.ddi.OV70</IDNo>
			<IDNo agency="DataCite">doi:10.21410/7E4/00LYOG</IDNo>
		</titlStmt>
```

But this information is lost at study level, either when harvesting in oai-ddi or when downloading metadata as DDI, for [example](https://data.sciencespo.fr/oai?verb=ListRecords&metadataPrefix=oai_ddi&set=CDSP):

```
<OAI-PMH
	xmlns="http://www.openarchives.org/OAI/2.0/"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
	<responseDate>2020-11-04T13:29:10Z</responseDate>
	<request verb="ListRecords" metadataPrefix="oai_ddi" set="CDSP">https://data.sciencespo.fr/oai</request>
	<ListRecords>
		<record>
			<header>
				<identifier>doi:10.21410/7E4/00LYOG</identifier>
				<datestamp>2020-08-22T00:00:07Z</datestamp>
				<setSpec>CDSP</setSpec>
				<setSpec>ALL_SCPO</setSpec>
			</header>
			<metadata>
				<codeBook
					xmlns="ddi:codebook:2_5"
					xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="ddi:codebook:2_5 https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd" version="2.5">
					<docDscr>
						<citation>
							<titlStmt>
								<titl>L'ouvrier français en 1970</titl>
								<IDNo agency="DOI">doi:10.21410/7E4/00LYOG</IDNo>
							</titlStmt>
							<distStmt>
								<distrbtr source="archive">data.sciencespo</distrbtr>
								<distDate>2020-05-05</distDate>
							</distStmt>
							<verStmt source="DVN">
								<version date="2020-05-13" type="RELEASED">2</version>
							</verStmt>
							<biblCit>Adam, Gérard; Bon, Frédéric; Capdevielle, Jacques; Mouriaux, René; Lavau, Georges, 2020, "L'ouvrier français en 1970", https://doi.org/10.21410/7E4/00LYOG, data.sciencespo, V2</biblCit>
						</citation>
					</docDscr>
```

However it is kept in [oai-dc harvesting](https://data.sciencespo.fr/oai?verb=ListRecords&metadataPrefix=oai_dc&set=CDSP): `<dcterms:language>French</dcterms:language>`

Maybe this is two different issues: one about the import that does not save the xml:lang attribute, and another about the dataset language that is not added to the OAI-DDI.

qqmyers:
**What this PR does / why we need it**: This PR adds xml:lang attributes to the DDI xml metadata export to improve compatibility with CESSDA's requirements. To support this, per Merce's preference, the PR adds a new metadata languages setting and a mechanism to let collection admins set the metadata language for new datasets (from the allowed list in the setting). The mechanism does not enforce using a specific language but does indicate the requested language when editing and again when metadata is displayed.

Closes #7388

**Special notes for your reviewer**: This is one of several PRs from Sciences PO work that increate i18n support. This PR is required for #6751 and should be reviewed/merged before the PR for that issue.

**Suggestions on how to test this**:  Testing the basic mechanism should be simple - add a setting to list the allowed languages (e.g. {"English":"en", "French":"fr"} ) and verify that one can pick the language when editing a collection (edit/general info) and that a new dataset created in that collection uses that language and retains it even when the collection is changed to have a new choice. Publishing a dataset created this way should then allow checking that xml;:lang tags are added to the DDI XML output. The PR adds tags to the codebook element as well as the title, abstract, and distributor elements.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Yes - see description above.

**Additional documentation**:will document new setting in guide


src/main/java/edu/harvard/iq/dataverse/DvObjectContainer.java
```diff
@@ -56,4 +67,32 @@ public void setStorageDriverId(String storageDriver) {
...
+    
+    public String getMetadataLanguage() {
+        if (metadataLanguage == null) {
+            return UNDEFINED_METADATA_LANGUAGE_CODE;
+        }
+        return metadataLanguage;
+    }
+
+    public void setMetadataLanguage(String ml) {
+        if (ml != null && ml.equals(UNDEFINED_METADATA_LANGUAGE_CODE)) {
```
scolapasta, line 91: Is there a reason to not just allow it to be null for undefined instead of transforming back and forth in the getter and setter?

src/main/java/edu/harvard/iq/dataverse/settings/SettingsServiceBean.java
```diff
@@ -433,8 +433,17 @@ Whether Harvesting (OAI) service is enabled
...
+
+        /**
+         * A Json object containing key/values corresponding the the allowed languages
+         * for entering metadata with the key being the human-readable string and the
+         * value being an ISO-639 code for that language (2 and 3 letter codes from the
+         * 639-2 and 639-3 standards are allowed. These will be used directly in
+         * metadata exports.) These values are selectable at the Dataverse level and
+         * apply to Dataset metadata.
+         */
+        MetadataLanguages;
```
scolapasta, line 446: would it possibly make sense to not have two sets of languages - one for interface and one for metadata and expect installation to use same languages for both? I guess there could be a use case for different, but I wonder if it's small enough that the more simple approach (one shared set) isn't more beneficial?

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -1063,5 +1073,55 @@ public String getDataCiteRestApiUrlString() {
...
+    public Map<String, String> getMetadataLanguages(DvObjectContainer target) {
+        Map<String,String> currentMap = new HashMap<String,String>();
+        currentMap.putAll(getBaseMetadataLanguageMap(true));
+        languageMap.put(DvObjectContainer.UNDEFINED_METADATA_LANGUAGE_CODE, getDefaultMetadataLanguageLabel(target));
+        return languageMap;
+    }
+    
+    private String getDefaultMetadataLanguageLabel(DvObjectContainer target) {
+        String mlLabel = Locale.getDefault().getDisplayLanguage();
+/*        Locale l = Locale.getDefault();
```
scolapasta, line 1102: should we remove this commented out section?

src/main/java/propertyFiles/socialscience.properties
```diff
@@ -1,80 +0,0 @@
-metadatablock.name=socialscience
```
scolapasta, line 1: Can you explain why all this is being removed?

src/test/java/edu/harvard/iq/dataverse/export/ddi/dataset-finch1.json
```diff
@@ -4,6 +4,7 @@
   "persistentUrl": "https://doi.org/10.5072/FK2/PCA2E3",
   "protocol": "doi",
   "authority": "10.5072/FK2",
+  "metadataLanguage": "en",
```
scolapasta, line 7: is this needed? (or is the default "en"?

scolapasta (commented):
Reviewed, I added some questions. Does this neeed a UI review / demo?

src/main/java/edu/harvard/iq/dataverse/settings/SettingsServiceBean.java
qqmyers: Sciences PO has use cases where these will be different. (My understanding is that they may only support a couple :Languages and allow more :MetadataLanguages, but I could image others wanting the reverse (support lots of display languages but restrict entries to a few.))

qqmyers (commented):
@scolapasta - I think I've addressed your comments. If you want a demo, let me know.
(Not sure how/why I 'started a review'...)

djbrooke:
Thanks @qqmyers for the update. @scolapasta can you take another look and see if this is good to go?

scolapasta (approved):

kcondon:
Testing Status:
1. The functionality, as designed, works -setting the :MetadataLanguage setting selects from among configured languages at the individual dataverse level to be the language used both when editing a dataset, esp. CVV list, and persists when viewing saved dataset's metadata CVV. Note that the setting only applies to data entered and currently only CVV -the field lables, eg. title, etc, will reflect whichever language is selected from menu, eg. fr but CVV could be in  en if :MetadataLanguages is set that way.
Issues:
1. There is currently no documentation -the new :MetadataLanguage setting needs documenting, as well as behavior, including how it relates to existing :Languages setting, esp for installations currently using that setting. Also, whether or not there is a default value if not set and the behavior in that case.
2. As discussed, a change in behavior when :MetadataLanguage setting is not explicitly configured might be a good idea, as in the case of a current installation using :Languages but not interested in using :MetadataLanguage. Currently, if :MetadataLanguage is not set, the selection list still appears in the UI (perhaps it shouldn't?) it shows the languages of en and null and null is selected. This corresponds to "default", which is en. The suggestion or discussion to retain consistent behavior with :Languages only setting would be for it to behave as it does now, so the language setting from the menu or "view" setting is the only reference in this case so when editing a dataset, CVV and labels and subsequent viewing of metadata will reflect the currently selected menu language. (after some thought, though this makes sense to me, maybe it should get a quick ok from Design/Danny).

kcondon:
OK, we're almost there! These are my test cases, please correct the technical details or otherwise provide feedback as needed :)

Config scenarios: default, :Languages only, :MetadataLanguages+:Languages, :MetadataLanguages only
1. no lang settings, check edit labels, check cvv list lang, check saved display labels, cvv, check ddi export text, xml lang setting, check harvest still works.
2:03
2. language only settings, check the above, all ui elements should reflect the currently selected lang in menu. export xml lang, not sure, maybe en/default, harvest still works
2:05
3. languages + metadatalanguages, check the above except, UI elements differ: Labels always reflect the currently selected languages from menu, cvv for edit and subsequent saved viewing will be in the dv selected metadata lang, dii export xml lang value should match dv metadatalanguage selection at time of save or pub?, actual cvv should be in selected lang and match xml lang, harvest still works
4. metadatalanguages only (n/a?)

PR #7959, opened 2021-06-21
SPO/6751 cvv i18n

### LINKED ISSUE #6751: Use the translation of a controlled vocabulary in the oai_ddi

qqmyers:
**What this PR does / why we need it**: This PR adds support for reporting (internal) controlled vocabulary values in the DDI XML metadata export, improving CESSDA conformance. When translations are available, CVV values are reported in the dataset's metadata language (as defined in #7958) and in the Dataverse instance's default locale's language, if that is different.

This PR applies the new mechanism to the subject field in the citation block and includes support for several socialscience metadatablock fields that are currently not CVV, but have been changed by Sciences PO using what I think are standard values - (see https://github.com/IQSS/dataverse/commit/28eafaa0ca2a69b937dd14b4865adf352f97691a) which will probably result in another PR.

Closes #6751

**Special notes for your reviewer**: This PR includes #7958 as those changes are required. Once that's merged, I'll update this. FWIW, I think the only changes unique to this PR are in DdiExportUtil.java. 

It should be possible to add to the tests for this, but only if there are translations supplied for another/test language. Since those are currently managed separately, I'm not sure how if there's a good way to add such a test.

**Suggestions on how to test this**: Configure dataverse with a properties file translating the citation block to another language. Pick that language (as in #7958) as the dataset's metadata language. Add a subject value(s) to a dataset using a metadata language and publish. The resulting ddi xml export should have subject entries in both languages, with xml:lang attribute tags specifying the languages used.

**Is there a release notes update needed for this change?**: Maybe? For the whole set of Sciences PO PRs, a note could be made re: the improved CESSDA compliance. I can provide some verbiage.

**Additional documentation**:


pdurbin:
> **Special notes for your reviewer**: This PR includes #7958 as those changes are required. Once that's merged, I'll update this. FWIW, I think the only changes unique to this PR are in DdiExportUtil.java.

@qqmyers can you please provide a "compare" link that shows just the changes in the file above. I tried but I'm getting many more changes with https://github.com/GlobalDataverseCommunityConsortium/dataverse/compare/SPO/7388-lang_attributes_in_DDI_export...GlobalDataverseCommunityConsortium:SPO/6751-cvv-i18n

@djbrooke it probably makes sense to get pull request #7958 merged before we review this one.

qqmyers:
@pdurbin - https://github.com/GlobalDataverseCommunityConsortium/dataverse/compare/SPO/7388-lang_attributes_in_DDI_export...GlobalDataverseCommunityConsortium:SPO/6751-cvv-i18n?expand=1 should be cleaner now.

scolapasta (approved):

PR #7976, opened 2021-06-29
Non-integer custom identifiers using stored procedure

### LINKED ISSUE #7548: Allow non-integer sequential identifiers
Dataverse provides a [mechanism](https://guides.dataverse.org/en/latest/installation/config.html#identifiergenerationstyle) for generating integer sequential identifiers through a database stored procedure.

In our implementation, we would like instead to generate fixed length string hashes that keep chronological order.
We already have a stored procedure for that, written in Perl, that returns a `char` instead of a `int`.

However, Dataverse [casts](https://github.com/IQSS/dataverse/blob/fb772d1dfbe78c7b96c88dbafa6218f8511f3a15/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java#L290) the output of `generateIdentifierAsSequentialNumber` to an `Integer`, and then [converts it](https://github.com/IQSS/dataverse/blob/fb772d1dfbe78c7b96c88dbafa6218f8511f3a15/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java#L296) to a `String`.

It would be nice if the output of `generateIdentifierAsSequentialNumber` could be considered a `String` or if another function could be added (e.g., `generateIdentifierAsSequentialString`).

claudiodsf:
This is my first contribution to Dataverse and first time I code Java, so please be understanding! 😄 

This PR replaces the option `sequentialNumber` for `:IdentifierGenerationStyle` with `storedProcGenerated`, and allows generating non-integer custom identifiers.

Closes #7548.

TODO:

- [x] Update [Documentation](https://guides.dataverse.org/en/latest/installation/config.html#identifiergenerationstyle) : which kind of example would you like to have in docs? The same as now?
- [x] Unit Tests? Not sure how to test this.
- [x] Update release notes

Tested on my local installation with the SQL script [createidentifier.txt](https://github.com/IQSS/dataverse/files/6732331/createidentifier.txt), which generates sequential identifiers from a base36 representation of current timestamp.

`[from L.A.:]` A very good test would be:
- Configure a 5.5 instance to use sequential identifiers (using the instructions from the 5.5 guide: execute the script `createsequence.sql`, and add one setting;)
- Confirm that you're getting sequential numeric identifiers; 
- Upgrade to this build; confirm that the instance is still assigning numeric identifiers. 

(so the goal is to confirm that Flyway successfully upgrades an existing old-style stored procedure)
`[/]`

No.

Probably yes.

**Additional documentation**:

A "flyaway script" was mentioned here https://github.com/IQSS/dataverse/issues/7548#issuecomment-771986937. Not sure where it should go and where it should be mentioned in documentation.


landreev:
Hi @claudiodsf!
I just started looking into this PR today. 
Let me answer some of the questions above, and I may ask some of my own later. 

> A "flyaway script" was mentioned here https://github.com/IQSS/dataverse/issues/7548#issuecomment-771986937. Not sure where it should go and where it should be mentioned in documentation.

Flyway is the library/tool we use to automatically update existing databases when things change in new releases. (Even though @scolapasta it is spelled "flyaway", lol).
The scripts live in `src/main/resources/db/migration/` in the source tree. There is a precise naming convention, to make sure the scripts are applied in the correct order when a new release is deployed. This is explained in our developers guide. But for the purposes of this PR, the script name should start with `V5.5.0.5__` (5.5 being the current base version; with the scripts 1-4 having already been added since its release). So the script you need to add should be named something like 
```
src/main/resources/db/migration/V5.5.0.5__7538-stored-procedure-update.sql
```
and it should contain this one line:
```
ALTER FUNCTION generateIdentifierAsSequentialNumber RENAME TO generateIdentifierFromStoredProcedure;
```
So if any installation is currently using a sequential number procedure, it will be renamed for them automatically when they deploy the next release, 5.6. 

**Before you add this file, please make sure** to sync your branch up with the develop branch! Flyway may end up very confused, if you end up with a script checked in and deployed out of order. (The latest script, `V5.5.0.4`, was just added to the develop branch the other day). 

Does this make sense? - Please feel free to ask if you have any questions!

landreev:
@scolapasta, @pameyer : If I'm reading it right, this PR is also changing a setting (`sequentialNumber` to `storedProcGenerated`). I understand that there may not be enough installations currently using the seq. numbers procedure (are there any, actually - except HMS?), to worry about backwards compatibility too much. But, if we are bothering to provide a flyway script to rename the procedure, we should probably auto-rename the setting as well, if exists - ?

pameyer:
@landreev I also read it as changing the name of the setting; my impression is that it's not worth worrying too much about maintaining backwards comparability (either updating the setting in flyway, or mentioning it in the release notes should cover it).

poikilotherm:
Quick thought: should we add some convenience for users/admins by adding a Java based migration, allowing to add the SQL function, reading it from somewhere?

We could provide a few standard scripts out of the box (seq. number and base36 timestamp as shown for a start) and let people configure which one to choose when initializing or provide their own.

The script seems similar to the other oneshot database change we handle with the "after migrate" hook.

claudiodsf:
Hi and thanks for all the comments!

> and it should contain this one line:
> 
> ```
> ALTER FUNCTION generateIdentifierAsSequentialNumber RENAME TO generateIdentifierFromStoredProcedure;
> ```
> 
> So if any installation is currently using a sequential number procedure, it will be renamed for them automatically when they deploy the next release, 5.6.

I think that the Flyway script cannot be as simple as that: we need  `generateIdentifierFromStoredProcedure` to return a `VARCHAR`. I don't think that `ALTER FUNCTION` can do that (but maybe I'm wrong).

The alternative would be to create a new function named `generateIdentifierFromStoredProcedure` that wraps the old `generateIdentifierAsSequentialNumber` and returns a `VARCHAR`.

claudiodsf:
> If I'm reading it right, this PR is also changing a setting (`sequentialNumber` to `storedProcGenerated`).

Yes!

landreev:
> I think that the Flyway script cannot be as simple as _[just renaming an existing function]_: we need  `generateIdentifierFromStoredProcedure` to return a `VARCHAR`.

@claudiodsf You are right of course. I got that idea from the earlier comments (in the issue), but yes, we either need to change that procedure (can still be done from the flyway script too); or maybe change the Java code slightly - to be able to handle _any_ return value from the stored procedure (type `Object`), and then use its `String` representation for the identifier... 

OK, let's figure this out tomorrow (on Monday). Sorry we are making you spend a bit more effort on this. But I still expect to be able to resolve all this easily; and then it will be included in the next release. (And there is a chance this will be useful to another Dataverse installation eventually).

landreev:
@pameyer 

> ... my impression is that it's not worth worrying too much about maintaining backwards comparability (either updating the setting in flyway, or mentioning it in the release notes should cover it).

Yes, when I said "backwards compatibility", I meant supplying a flyway script to make the changes. As opposed to simply mentioning it in the release notes, telling anyone potentially affected how to fix it manually. 
(Flyway script still sounds like a doable solution).

landreev:
_Can_ we provide a flyway script that would resolve this automatically? - I'm actually not sure anymore. 
I.e., it's easy to insert a `CREATE OR REPLACE FUNCTION ...` into a flyway script, to create an updated version of the current `generateIdentifierAsSequentialNumber()`, that returns a string... But is there an _easy_ way to only do this if the function currently exists? - you cannot use `IF EXISTS` with `CREATE OR REPLACE`. And it feels silly, to create this procedure for ALL installations. 

There may be some simple way I'm not seeing... But I am ok by now with the idea of not bothering with flyway at all. And simply providing a release note with the instructions.

@claudiodsf, our release notes go into doc/release-notes. You can look at the existing notes for examples. But I could quickly write it myself. That's not the hard part.

landreev:
Everybody, 
One other thing this PR would need to include - the guide section that explains how to create the currently supported stored procedure needs to be completely re-written. (The script that creates it, `createsequence.sql`, is supplied as part of the guide too). 

I know I'm late to this discussion, so I'd be ok with being outvoted on this, but I'm just not sure why we are changing so much, and I no longer feel that it's justified. 
If it were up to me, I would leave the current `generateIdentifierAsSequentialNumber()` intact. And simply add the stored procedure that returns a string as the second option. So instead of removing/renaming the current setting in the code, there would be 
```
switch (doiIdentifierType) {
   ...
   case "sequentialNumber":
      ...
   case "storedProcGeneratedString":
      ...
```
etc. And zero changes for any installation that may be using, and happy with, the current setup. 

I get that the original rationale was that an installation would only have a use for one of the two methods, so why not combine them into one. But I just feel that it may be outweighed by having to make all these changes and and having to supply a release note - with a possibility that an affected installation will miss the warning. Changing an existing setting always bothers me. Even if are not aware of that many installations using it.

scolapasta:
@landreev I hear your concerns, but I still think it's cleaner to do the complete solution this time around. I'm not aware of many installations using the sequence (though that doesn't mean they don't exist), but I do think the release notes should cover it and otherwise, it will be pretty obvious failure that occurs.

claudiodsf:
Hi and sorry for the delay.

> maybe change the Java code slightly - to be able to handle _any_ return value from the stored procedure (type `Object`), and then use its `String` representation for the identifier...

I tried with `@StoredProcedureParameter(mode = ParameterMode.OUT, type = Object.class)` at this line: https://github.com/IQSS/dataverse/blob/8c4d651709f53c1a79c8c570df2f05ff117a8bfa/src/main/java/edu/harvard/iq/dataverse/Dataset.java#L85

But had no success: I get exactly the same behaviour than using `type = String.class` 🤔.


So, I decide to move to the SQL side of things and came out with the following script, which wraps `generateIdentifierAsSequentialNumber()` into `generateIdentifierFromStoredProcedure()`.

```sql
DO $BODY$
BEGIN
  IF EXISTS ( SELECT * from generateIdentifierAsSequentialNumber() ) THEN
    CREATE OR REPLACE FUNCTION generateIdentifierFromStoredProcedure()
    RETURNS varchar AS $$
    DECLARE
        identifier varchar;
    BEGIN
        identifier := generateIdentifierAsSequentialNumber()::varchar;
        RETURN identifier;
    END;
    $$ LANGUAGE plpgsql IMMUTABLE;
  END IF;
END $BODY$;
```

This could go into the Flyway script.

I checked it, of course, and it works.

What do you think?

landreev:
@claudiodsf OK, that looks good, thanks! We could put this update into a flyway script; but it may be easier to just supply it as part of the release note. 
OK, let me try the flyway approach here; and then I can updating the documentation for this PR, so that we can close it.

claudiodsf:
Thinking about it, it is better to not `REPLACE` the function `generateIdentifierFromStoredProcedure()` if it already exists, since the user might have customised it after first running the update script.

Here's a new version of the script that fixes that:

```SQL
DO $BODY$
BEGIN
  BEGIN
    PERFORM generateIdentifierAsSequentialNumber();
  EXCEPTION
    -- If the above function does not exist, we can stop executing this script
    WHEN undefined_function THEN
      RETURN;
  END;
  BEGIN
    PERFORM generateIdentifierFromStoredProcedure();
  EXCEPTION
    -- We only create this function if it doesn't already exist,
    -- to avoid overwriting user modifications 
    WHEN undefined_function THEN
      CREATE FUNCTION generateIdentifierFromStoredProcedure()
      RETURNS varchar AS $$
      DECLARE
          identifier varchar;
      BEGIN
          identifier := generateIdentifierAsSequentialNumber()::varchar;
          RETURN identifier;
      END;
      $$ LANGUAGE plpgsql IMMUTABLE;
  END;
END $BODY$;
```

landreev:
I agree, this is even better. 
All that said, I still feel like declaring the function with the return `type = Object.class`, and then using the `.toString()` of that return value should've worked... so I'm a bit surprised that it didn't - wondering why. Let me try something here.

landreev:
(BTW, I really like using that `PERFORM ... EXCEPTION` block, to check if the function already exists. I was thinking of trying instead an `IF EXISTS` on the standalone sequence that the procedure uses... but this is much cleaner!)

claudiodsf:
> (BTW, I really like using that `PERFORM ... EXCEPTION` block, to check if the function already exists. I was thinking of trying instead an `IF EXISTS` on the standalone sequence that the procedure uses... but this is much cleaner!)

Yes, Stack Overflow rules 😛

landreev:
OK, I still find it strange, that casting the return value of the function to Object does not work - but it just doesn't. So let's not worry about it. 

I built your branch yesterday, and tried your script with Flyway, it worked like magic. I added one line to it, to modify the setting as well: 
```
-- If the installation is using a stored procedure for generating
-- sequential numeric identifiers, create a wrapper function that
-- works with the new framework (the stored procedure now needs to
-- return a string).
DO $BODY$
BEGIN
  BEGIN
    PERFORM generateIdentifierAsSequentialNumber();
  EXCEPTION
    -- If the above function does not exist, we can stop executing this script
    WHEN undefined_function THEN
      RETURN;
  END;
  BEGIN
    PERFORM generateIdentifierFromStoredProcedure();
  EXCEPTION
    -- We only create this function if it doesn't already exist,
    -- to avoid overwriting user modifications 
    WHEN undefined_function THEN
      UPDATE setting SET content='storedProcGenerated' WHERE name=':IdentifierGenerationStyle' AND content='sequentialNumber';
      CREATE FUNCTION generateIdentifierFromStoredProcedure()
      RETURNS varchar AS $$
      DECLARE
          identifier varchar;
      BEGIN
          identifier := generateIdentifierAsSequentialNumber()::varchar;
          RETURN identifier;
      END;
      $$ LANGUAGE plpgsql IMMUTABLE;
  END;
END $BODY$;
```
It can be checked in as `src/main/resources/db/migration/V5.5.0.5__7538-stored-procedure-update.sql`. But please update the branch first, to sync it up with the develop branch.
Thank you!

claudiodsf:
Thank you @landreev!

I just added the Flyway script to the commit and force-pushed.
Note that the script name should rather be `V5.5.0.5__7548-stored-procedure-update.sql` 😉 

I moved the setting update line at the beginning of the script, since it will correctly work (i.e., do nothing) even if the setting does not exist or it's already changed.

claudiodsf:
Things still missing:

- release notes (@landreev: you offered to write this part 😉)
- documentation: which example of stored procedure would you like to have in the doc? Still the numeric sequence? Base36 encoding of timestamp? Other?
- tests??? Let me know if tests are required (and how to tackle this 😉)

landreev:
Thanks for correcting the issue number. 
I would keep the sequential numbers procedure as the documented example, just changing its name and return value; plus changing the value of the setting where it's mentioned in the guide. 
But if you want to add your procedure, as a second example - that could potentially be useful to somebody too.

claudiodsf:
Ok, I will go for the two examples, then 😉 

Should we mention in the documentation that this setting has changed respect to the previous version? Or is this something just for the release notes?

Also, I would add a note in the documentation saying that the Dataverse administrator is responsible for creating unique IDs and that Dataverse may behave oddly if the same ID is used multiple times (is that true?).

landreev:
> Should we mention in the documentation that this setting has changed respect to the previous version? Or is this something just for the release notes?

The release note should be enough. (Will do).

> Also, I would add a note in the documentation saying that the Dataverse administrator is responsible for creating unique IDs and that Dataverse may behave oddly if the same ID is used multiple times (is that true?).

The uniqueness of the identifier is enforced on the database level. There is a constraint that will prevent a dataset with an already existing identifier from being saved. Also, Dataverse itself will check the generated identifier against what's already in the database; and simply try to generate another one, until it's unique. So it shouldn't be catastrophic, if somebody writes a stored procedure that occasionally returns the same value... But it may still be worth mentioning that the procedure should be generating _unique_ identifiers; because that's the goal. 
(everyone - correct me if I'm wrong on any of the above please)

landreev:
OK, yes, it appears that the Jenkins tests failed because the Ansible scripts execute (the old) `createsequence.sql` AFTER the application is deployed; so Flyway could not modify the procedure, because it wasn't there yet. 
This is being addressed (in https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible).

claudiodsf:
Hi, I just made a new commit with the documentation update.

I reflowed the text of the two paragraphs which I modified (`:IdentifierGenerationStyle` and `:DataFilePIDFormat`), to make it easier to edit with my editor. I hope that's fine with you.

I built the documentation locally and everything seems fine. I don't know if you have a mechanism to build it online.

Finally, let me know if you want me to squash the two commits of this PR into a single one.

Thanks!

claudiodsf:
> OK, yes, it appears that the Jenkins tests failed because the Ansible scripts execute (the old) `createsequence.sql` AFTER the application is deployed; so Flyway could not modify the procedure, because it wasn't there yet.
> This is being addressed (in https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible).

Thanks, all tests pass now! 🎉

landreev:
No, it's not necessary to combine multiple commits - there can be as many as needed. 
I'll add a release note shortly (will try to commit into your branch directly, if I can; otherwise will post it here). 
Thanks again.

landreev:
May I make just one small request - could you rename the sql script that creates the numeric sequence function *back* to `createsequence.sql` please? 
The system that runs automated api tests on pull requests and branches (https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible) that I mentioned, it downloads `createsequence.sql` from the master (i.e., the most recent release) branch. If we keep the script under the old name, we won't need to remember to make a change there after the next release. 
Sorry to keep making up more work for you! :)

landreev:
The documentation updates look good.

claudiodsf:
> May I make just one small request - could you rename the sql script that creates the numeric sequence function _back_ to `createsequence.sql` please?

Done! (After doing some mess with rebasing. Hopefully all is good now).

landreev (approved):
Everything's looking great! Moving into QA.

landreev:
> Done!

Great! The PR is in QA, and will be tested and merged soon, hopefully.
Thank you again for contributing.

claudiodsf:
It's very nice to see this PR merged! 🎉

Thank you all for your support and guidance!

landreev:
> It's very nice to see this PR merged! 🎉

Thank you, it was a pleasure! 
Hope you weren't discouraged by how many things need to be addressed, in order to get something merged. (So maybe we will see more contributions from you in the future! :smile:)

PR #7983, opened 2021-06-30
7980 enhance data source definition

### LINKED ISSUE #7980: DB connection: defaults and more options
Now that #7701 is merged and we are on 5.2021.4, we can enable real defaults for the database connection.

### Context

I learned that any variable expansion supporting annotations or files, that will have Payara create resources, containers or sth else from, will not retrieve default values from the `microprofile-config.properties` file bundled in the WAR.

Instead, we need to provide defaults with the new `${MPCONFIG=name:default}` syntax.

### Todo

- [x] Add defaults to the connection URL. Using "localhost", "5432" and "dataverse" will allow most installation and devs to go without config.
- [x] Use the newly introduced caps to configure more Payara specific properties regarding connection monitoring etc.
- [x] Edit release notes of 5.3 to be more precise about database configuration not having defaults...
- [x] Maybe others

poikilotherm:
Create real defaults for the database connection, enable configuring advanced settings, add docs ~~and fix 5.3 release notes.~~

Closes #7980 

The release docs of 5.3 do not need to be changed. The statements on how to upgrade etc are working just fine.
Only in a few occasions people might have hit the regression.

1. Remove the properties from domain.xml. Should still work (if you had used the defaults in there)
2. Try setting the JDBC logging to enabled via system property: `asadmin create-system-properties "dataverse.db.log-jdbc-calls=true"`

Nope.

:battery: included.

**Additional documentation**:
None.

### DIFF
--- a/doc/release-notes/7980-enhanced-dsd.md
+++ b/doc/release-notes/7980-enhanced-dsd.md
@@ -1,0 +1,10 @@
+### Default Values for Database Connections fixed
+
+Introduced in Dataverse release 5.3 a regression might have hit you:
+the announced default values for the database connection never actually worked.
+
+With the update to Payara 5.2021.4 it was possible to introduce working
+defaults. The documentation has been changed accordingly.
+
+Together with this change, you can now enable advanced connection pool
+configurations useful for debugging and monitoring. See the docs for details.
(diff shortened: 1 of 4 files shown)


poikilotherm:
Looks like this will stay in "draft" for another while, until payara/Payara#5322 is resolved, landed in a new release AND we update to that or a later release. :see_no_evil:

poikilotherm:
Closed in favor of all new #8915

Issue #6863, opened 2020-04-28
FileMetadataIT.testJsonParserWithDirectoryLabels failing


pdurbin:
FileMetadataIT.testJsonParserWithDirectoryLabels is failing for me locally as well as at https://jenkins.dataverse.org/job/IQSS-dataverse-develop/435/

scolapasta:
This test still is failing (at leas intermittently) so I'm reopening up this issue.

scolapasta:
(Oh, I I didn't realize the April date above was 2020, not 2021 (i.e. a year ago). Still not much noise here, so I think it's ok to track here.

sekmiller:
This is still failing via Jenkins in the Download All branch, though not for me locally in that branch.

qqmyers:
Can using the first 8 chars of a UUID as the testName (line 44) cause issues? I.e. some char that isn't allowed in a dataverse name/alias?

pdurbin:
@qqmyers yes, good call. It's probably the dv alias. If we switch to `UtilIT.getRandomDvAlias()` a "dv" will be prepended, making a valid dv alias (rather than all digits, occasionally).

PR #7984, opened 2021-06-30
SPO/7786 Add holdings element to DDI and use PID URIs

### LINKED ISSUE #7786: No URI in OAI-PMH records (with oai-ddi metadata prefix)
From our understanding, we could have an URI Provided inside a metadata "holdings" tag pointing to the dataset (in the form of an URL https://doi.org/… rather than the DOI itself doi:10.21410/7E4/SRVNPE).
This uri, in ddi-codebook 2.5 xml, should be in codeBook -> docDscr -> citation -> holdings or/and codeBook -> stdyDscr -> citation -> holdings

This should allow Dataverse to provide "self sufficient" records in OAI-PMH records.

Example of desired result : http://services.fsd.tuni.fi/v0/oai?verb=GetRecord&identifier=oai:fsd.uta.fi:FSD1262&metadataPrefix=oai_ddi25

qqmyers:
**What this PR does / why we need it**:  This PR adds a studydsc/citation/holdings element with a URI attribute whose value is the full PID/DOI URI for the dataset. It no longer updates the studydsc/citation/titlestmt/IDNo element to use the URI form of the DOI as this causes a compatibility issue when older servers harvest from newer ones using the ddi export and because using the URI form does not appear to be required/recommended by the CESSDA validator. (which does check the holdings element).

Closes #7786

**Special notes for your reviewer**: This seems like an improvement that addresses #7786 but I think further work to understand best-practices for what should be in these elements might be worthwhile. Examples I've seen use the local URL (i.e. what the DOI resolves to) in the holdings element, and do not use the dataset DOI in the docdscr/citation/titlestmt/IDNo (not changed in this PR) as Dataverse does. Further, per the issue discussion, it seems like holdings could reference the individual files that are part of the dataset. In any case, I think this PR should improve compliance and doesn't interfere with making further changes going forward.

**Suggestions on how to test this**: the DDIExportUtilTests check the changes. Manually I think you can check the DDI metadata export to see the changes as well. One could also test the OAI-PMH outputs directly (I haven't but afaik they use the same ddi generation code.

**Additional documentation**:


pdurbin (commented):
Overall, this looks fine but it does represent a backward-incompatible change since `<IDNo agency="DOI">` becomes a URL instead of a DOI. (I can imagine scripts or systems breaking.) @qqmyers can you please add a release note?

pdurbin:
I was about to move this to QA when I noticed a test failure at https://jenkins.dataverse.org/blue/organizations/jenkins/IQSS-Dataverse-Develop-PR/detail/PR-7984/3/tests

Screen Shot 2021-07-16 at 3 43 17 PM

Digging a little more, it seems like this is one of the tests that fails intermittently:

- FileMetadataIT.testJsonParserWithDirectoryLabels failing #6863 

So it should be safe to move to QA.

pdurbin (approved):
The release notes seems to warn well enough that this is a backward incompatible change. The code looks fine.

kcondon:
Passes smoketest holdings element in export verified, harvest between two dataverses of same pr works but when harvest client is 5.5 it fails due to parsing error of doi (IDno)

qqmyers:
@kcondon - I reverted the change to the IDNo element so the harvesting-related compatibility issue should be gone.

PR #7992, opened 2021-07-07
update rserve documentation

### LINKED ISSUE #7893: guides: prerequisites.rst doesn't link to scripts/r/rserve/rserve-setup.sh
though the script refers to other contents of scripts/r/rserve/ so perhaps the installation instructions might simply link to this folder?

happy to submit a PR once @pdurbin @landreev and others agree on what should/might be done.

donsizemore:
**What this PR does / why we need it**: currently the Rserve documentation mentions individual scripts used to install Rserve and where to find them, but there is no explicit link or mention that one needs all files in the subdirectory. This PR links to the directory and spells out that one needs copies of all files contained therein.

Closes #7893

**Suggestions on how to test this**: none

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: none

### DIFF
--- a/doc/sphinx-guides/source/installation/prerequisites.rst
+++ b/doc/sphinx-guides/source/installation/prerequisites.rst
@@ -365,9 +365,9 @@ network connections on a dedicated port. This requires some extra
 configuration and we provide a script for setting it up.
 
 You'll want to obtain local copies of the Rserve setup files found in
-https://github.com/IQSS/dataverse/tree/develop/scripts/r/rserve
+:fixedwidthplain:`scripts/r/rserve/<../../../../scripts/r/rserve/>`
 either by cloning a local copy of the IQSS repository:
-:fixedwidthplain:`git clone https://github.com/IQSS/dataverse.git`
+:fixedwidthplain:`git clone -b master https://github.com/IQSS/dataverse.git`
 or by downloading the files individually.
 
 Run the script as follows (as root)::
--- a/doc/sphinx-guides/source/installation/prerequisites.rst
+++ b/doc/sphinx-guides/source/installation/prerequisites.rst
@@ -362,7 +362,14 @@ The Dataverse Software uses `Rserve <https://rforge.net/Rserve/>`_ to communicat
 to R. Rserve is installed as a library package, as described in the
 step above. It runs as a daemon process on the server, accepting
 network connections on a dedicated port. This requires some extra
-configuration and we provide a  script (:fixedwidthplain:`scripts/r/rserve/rserve-setup.sh`) for setting it up.
+configuration and we provide a script, rserve-setup.sh, for setting it up.
+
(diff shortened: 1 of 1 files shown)


djbrooke (commented):
Thanks @donsizemore for the PR. I'm not a user of this particular guide, but from a consistency standpoint it looks like in other places we don't point to a specific branch but instead point to the script within the scripts folder. Ex.: 

- https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/installation/advanced.rst
- https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/developers/making-releases.rst

I think the advantage of this is that it forces the installation admin/installer to go to the specific version that they are installing instead of introducing a potential version mismatch in the guides (as the develop branch in this case may be a different script than the version they are installing). 

Am I understanding this correctly and, if so, let me know what you think about the advantages/disadvantages of the different approaches?

landreev:
@djbrooke We did discuss this yesterday, before Don created the final PR. We definitely don't want to point the user to the specific script for this one: They need to download the contents of the entire directory for the whole thing to work. The reason Don created this PR was that a user apparently had just downloaded the script, without downloading the other files, and tried to run it. 

But I understand that your main point was pointing them to the specific branch, and that is a valid concern. Except for this particular thing - the Rserve configuration - the chances of us having to make any changes there are slim. And the chances of any such changes introducing an incompatibility are slimmer still... I feel it would be ok to point to the master; under the assumption that the currently released version of this config should be stable. But simply replacing that link with the directory path, as in other docs, should also be ok, I think.

djbrooke:
Thx @landreev - I'll defer to you and @donsizemore since you're both more familiar here. If we do want to link to the main branch we should just update the PR to point there instead of develop. Once that's done I think it would be OK to approve. Thanks for discussing (yesterday and in this issue)!

pdurbin:
As I said yesterday in the DMs between me, Don, and Leonid, I'm fine with what's being proposed (linking to the "develop" branch) but if in the future it turns out we do need to have tighter versioning around these Rserve files ("these are the files that work with for Dataverse 5.6", for example), we could consider adding them to the installation zip (and rewriting the doc, of course).

landreev:
For the record, I agree - if it ever becomes an issue (or if we ever hear from another user about this config step being a problem) - we'll just add these extra files to the installer bundle. 

> As I said yesterday in the DMs between me, Don, and Leonid, I'm fine with what's being proposed (linking to the "develop" branch) but if in the future it turns out we do need to have tighter versioning around these Rserve files ("these are the files that work with for Dataverse 5.6", for example), we could consider adding them to the installation zip (and rewriting the doc, of course).

djbrooke (approved):
Thanks @donsizemore and @landreev @pdurbin for discussing. LGTM now.

PR #7996, opened 2021-07-09
append deaccessionDialog.reasons with periods for proper display

### LINKED ISSUE #7936: Please add punctuation to standard deaccession reasons.
Because of how Dataverse assembles information from the Deaccession Dataset form (reason for deaccession, additional information about the reason, alternative URL) the resulting Deaccession Reason information that appears on the landing page of deaccessioned datasets does not read correctly, like so:

Please add punctuation to the standard deaccession reasons as an easy fix for this.

donsizemore:
**What this PR does / why we need it**: deaccessionDialog.reasons aren't terminated with punctuation but wind up concatenated with other text in run-on sentences

Closes #7936

**Suggestions on how to test this**: deaccession a dataset, then attempt to view it


pdurbin (approved):
Looks good.

One additional thing to test or at least think about is that you can see the deaccession details in the database with `select versionnote,archivenote from datasetversion` and the "versionnote" will now have a period after the reason and before the (optional) additional info (screenshot below). I think this is fine. We can't facet on (or GROUP BY) "versionnote" anyway since it's a combination of these two fields. I only mentioned "archivenote" above because that's where the URL goes, like this (from "develop", no period after "IRB request"):

```
dvndb=> select versionnote,archivenote from datasetversion;
    versionnote     |      archivenote       
--------------------+------------------------
                    | 
 IRB request foobar | https://www.google.com
```


Screen Shot 2021-07-09 at 2 52 09 PM

PR #7998, opened 2021-07-13
Sec36 - update Google Cloud Storage

poikilotherm:
Updating the Google Cloud BOM to use a more recent version of the cloud libraries. Some contained security issues.

Closes IQSS/dataverse-security#36

We should think about moving the Guava and Gson deps to `<dependencyManagement>` with their version and add a version-less direct dependency. This way we properly align the Google Cloud libs with their version of Guava and Gson with ours.

Also, Google Auto Service is at v1.0 - we are using v1.0-rc2.

Well... Do everything to make Dataverse archive data to Google Cloud Storage. Or cross fingers. 

Nope.

Nope.

**Additional documentation**:
None.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
         <commons.logging.version>1.2</commons.logging.version>
         <commons.lang3.version>3.12.0</commons.lang3.version>
         <httpcomponents.client.version>4.5.13</httpcomponents.client.version>
+        <google.cloud.version>0.157.0</google.cloud.version>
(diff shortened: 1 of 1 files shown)


poikilotherm:
After testing this locally, the issues still persist with updated deps. As written in the issue, looks like false positives. :unamused: 

If you want, we could still update the dep.

qqmyers (approved):
Looks like this works fine at QDR so we should probably go ahead with the update and stay current.

PR #8003, opened 2021-07-15
Permissions management screen: display warning only if needed

### LINKED ISSUE #8002: Display issue on permissions management screen
I created a custom role with only one permission: publish dataset.

When I manage permissions on a dataset, there is a display issue since this role has no effect on files but only on current dataset.

With an existing role:
Capture d’écran 2021-07-15 à 16 59 37

With my role:
Capture d’écran 2021-07-15 à 16 59 44

When I manage permission on dataverse, no problem.

I think that the warning block is useless when the role has effect only on current object type, eg a dataverse role on a dataverse or a dataset role on a dataset

tcoupin:

Closes #8002


sekmiller (approved):
This looks good. Thank you for this fix.

PR #8007, opened 2021-07-20
7858 dup file upload

### LINKED ISSUE #7858: S3 direct-upload: uncaught duplicate file warning
One of our archivists, using an S3 bucket with upload-direct enabled, reported:

image

searching for Demo_Geo.zip yields

```
[2021-04-27T17:45:30.410+0000] [Payara 5.2020.6] [INFO] [] [] [tid: _ThreadID=107 _ThreadName=http-thread-pool::jk-connector(1)] [timeMillis: 1619545530410] [levelValue: 800] [[
  upload error]]

[2021-04-27T17:45:30.411+0000] [Payara 5.2020.6] [INFO] [] [] [tid: _ThreadID=107 _ThreadName=http-thread-pool::jk-connector(1)] [timeMillis: 1619545530411] [levelValue: 800] [[
  Error! You may not replace a file with a file that has duplicate content.]]

[2021-04-27T17:45:30.411+0000] [Payara 5.2020.6] [WARNING] [] [javax.enterprise.resource.webcontainer.jsf.lifecycle] [tid: _ThreadID=107 _ThreadName=http-thread-pool::jk-connector(1)] [timeMillis: 1619545530411] [levelValue: 900] [[
  #{EditDatafilesPage.handleExternalUpload()}: java.lang.NullPointerException
javax.faces.FacesException: #{EditDatafilesPage.handleExternalUpload()}: java.lang.NullPointerException

Caused by: java.lang.NullPointerException
        at edu.harvard.iq.dataverse.EditDatafilesPage.handleExternalUpload(EditDatafilesPage.java:2033)

javax.faces.el.EvaluationException: java.lang.NullPointerException
Caused by: java.lang.NullPointerException
        at edu.harvard.iq.dataverse.EditDatafilesPage.handleExternalUpload(EditDatafilesPage.java:2033)
```

I ask Jim whether it's possible this duplicate content error isn't percolating back up to the browser, who says:

> Probably - line 2033 of v5.3 includes fileReplacePageHelper.getNewFileMetadatasBeforeSave() which will return null if 'phase 1' of adding the file fails, so any problem in phase 1 would break it. The problem still exists in the develop branch, so worth an issue there.

sekmiller:
Catches a failure in the file upload in order to inform the user. Previously there was no feedback in the ui - just a note in the log file. Also adds some error messages when s3 setup is faulty - unable to retrieve rsync script.

Closes #7858 s3 direct-upload uncaught duplicate file warning.

**Special notes for your reviewer**:  We were never able to reproduce this issue in 5.5 or the current dev environment. I added some error handling around the reported lines from the logging of the issue and some additional error messaging around improper setup of s3 support.

**Suggestions on how to test this**: try to reproduce the original error. to exercise the additional messaging try adding an erroneous storage driver or dcmurl

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1904,8 +1905,11 @@ private String init(boolean initFull) {
                         }
                     } catch (RuntimeException ex) {
                         logger.warning("Problem getting rsync script: " + ex.getLocalizedMessage());
+                        FacesContext.getCurrentInstance().addMessage("Dataset Page Runtime", new FacesMessage("message")
```
pdurbin, line 1908: Is "Dataset Page Runtime" really what we want to show the user when there's a RuntimeException? Maybe "Dataset Page RuntimeException" instead? And perhaps ex.getLocalizedMessage too? Or "Failed to get rsync script."?

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
```diff
@@ -2037,7 +2049,11 @@ public void handleExternalUpload() {
     			// -----------------------------------------------------------
     			if (this.isFileReplaceOperation()){
     				this.handleReplaceFileUpload(storageLocation, fileName, contentType, checksumValue, checksumType);
-    				this.setFileMetadataSelectedForTagsPopup(fileReplacePageHelper.getNewFileMetadatasBeforeSave().get(0));
+                                if (!saveEnabled){
+                                    uploadWarningMessage = fileReplacePageHelper.getErrorMessages();
```
qqmyers, line 2053: From line 1889, isn't uploadWarningMessage already set? And you could just check for non-null on that rather than checking saveEnabled? (It wasn't immediately clear to me why you were checking saveEnabled (in an error case it remains false), but I'm mostly asking because if you do need to set the warning here, I'm not reading the code correctly.)

pdurbin (commented):
Overall, these seems ok but I pointed out where one of the user-visible errors looks odd.

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
sekmiller: Point taken. I guess I had two things in mind - not calling the "get(0)" if the array was null and making it obvious that the save was not enabled. I can simplify here.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: Right. Thanks. I moved the Exception Type to the logger for diagnostics. The user doesn't need to see it.

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
qqmyers: Looks good

pdurbin (approved):
I don't full grok the fix but it has to do with error handling, which this pull request addresses.

Heads up to QA that there are a number of fixes for rsync in here, so I guess that should be tested.

PR #8015, opened 2021-07-21
fix for the default zip download limit

### LINKED ISSUE #7932: Zip downloads disabled unless :ZipDownloadLimit is set, mention in Guides that setting ZipDownloadLimit now toggles UI options
- In v5.5, not having a ZipDownloadLimit set results in a limit of 0 being returned and that disables all Zip Downloads. The docs say there is a default limit of 100MB, but the code in SystemConfig sets the default to 0.
- In the docs, we should clarify that the ZipDownloadLimit now controls the workflows implemented in #7714 (this is especially important in installations running the custom zipper service)

landreev:

See the issue; self-explanatory (#7932)

Closes #7932

copy-and-pasting from a comment in the issue: 

*This was happening because the 100MB default was set in the class DataFileZipper.java, outside of the normal framework of keeping the defaults in SystemConfig. Under the original assumption that no other components would need that value, probably? As a result, the page wasn't getting it for the purposes of disabling large downloads.
[this] fixes it by moving the default value to the standard place*

**Additional documentation**:


qqmyers (approved):
Looks good to me.

landreev:
It looks like we've come across a small preexisting issue in QA... Once the "too much selected" popup is shown, it'll be shown again, even after the user unselects enough files; before the download starts. 
It may be super easy to fix that too...

landreev:
Giving up on the preexisting issue with the popup. Not obvious at all how to fix it. 
Everything is handled correctly in the page code - the page isn't trying to show that popup again. It's something on the java faces level; something not getting properly refreshed. 
I can't spend too much of my time on this.

PR #8016, opened 2021-07-22
Improve schema.org md

qqmyers:
**What this PR does / why we need it**: Improves schema.org metadata handling

Closes #

**Suggestions on how to test this**: The only changes should be to the schema.org metadata in the `<head>` of the html on published dataset pages. Any html tags/special chars added in author name/affiliation or other fields included in the schema.org metadata should not appear in the metadata.

`From L.A.:` Sure, no need to add anything too specific here. But there are some examples in the slack discussion to try, before and after. 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
@@ -1707,11 +1707,11 @@ public class DatasetVersion implements Serializable {
         JsonArrayBuilder authors = Json.createArrayBuilder();
         for (DatasetAuthor datasetAuthor : this.getDatasetAuthors()) {
             JsonObjectBuilder author = Json.createObjectBuilder();
-            String name = datasetAuthor.getName().getValue();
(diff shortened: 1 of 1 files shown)


landreev (approved):
I checked what's inside the MarkupChecker.stripAllTags() method, that is being used to sanitize the entire fragment now: 
```
   return Parser.unescapeEntities(Jsoup.clean(unsafe, Whitelist.none()), true);
```
and read the Jsoup man pages on the above, and it all sounds good.
As long as it's been tested, that aside from any markup tags, the cleaned-up json fragment is intact (including the schema.org and doi.org urls in it).

qqmyers:
FWIW: If this had been exploited prior to the fix, the result would have been cached in the Schemadotorg metadata export file which is read/used to create the jsonld metadata in the dataset page.  We saw this at QDR in a test dataset that appeared not to be fixed due to the cached info. Hopefully we caught this before there were any wild cases but thought I'd at least mention this hear in case someone does find an old instance - re-exporting should fix it.

PR #8026, opened 2021-07-26
118 file level perms performance

sekmiller:
**What this PR does / why we need it**: Improves performance of the Manage File Permissions page. It is timing out for a dataset with many files.

Closes #
https://github.com/IQSS/dataverse.harvard.edu/issues/118

Most gains are achieved by adding a method for the onchange event of the Show Deleted Files button. Prior to this the init maps was being called three times on page load. Also stopped checking for "deleted" if the file was not restricted.

Try a dataset with a lot of restricted files and see that the page loads faster without losing any functionality.

**Additional documentation**:None

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
@@ -115,6 +115,17 @@ public class ManageFilePermissionsPage implements java.io.Serializable {
     }
     
     
+    private boolean backingShowDeleted = true;
+
+    public String showDeletedCheckboxChange() {
+
+        if (backingShowDeleted != showDeleted) {
+            initMaps();
+            backingShowDeleted = showDeleted;
+        }
+        return "";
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
```diff
@@ -115,6 +115,17 @@ public void setDataset(Dataset dataset) {
     }
     
     
+    private boolean backingShowDeleted = true;
+
+    public String showDeletedCheckboxChange() {
```
scolapasta, line 120: @sekmiller Only small comment, does this need to return "" or can it just be void?

sekmiller: you're right. thanks. I made the change.

scolapasta (approved):

kcondon:
Tested on reported dataset: old: 3m45s new: 1m17s

PR #8033, opened 2021-07-28
IQSS/8028 enable import with test dois

### LINKED ISSUE #8028: Import APIs don't work with DataCite test DOIs
Discovered during @kcondon's testing of #7504: The [test in ImportDataCommand](https://github.com/IQSS/dataverse/blob/c284ffa0580264f4744bb6b70517c3b4c26ca04f/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ImportDatasetCommand.java#L63-L80), which is intended to see if a DataCite DOI exists, always fails for test DataCite DOIs now (since their change from using the open 10.5072 test authority and requiring test accounts. As of that change, test DOIs are not indexed even when findable and thus URLs of the form https://doi.org/10.33564/FK27U7YBV always return 404.)
This affects the new migrate API in #7504 and should also affect the existing Dataverse APIs /api/dataverses/\<id\>/datasets/:import and /api/dataverses/\<id\>/datasets/:importddi APIs. 

Since the intent of the test is just to verify that the DOI exists at DataCite, calling DOIDataCiteServiceBean.alreadyExists() for the DOI instead should work.

Note that this is a testing-only issue.

qqmyers:
**What this PR does / why we need it**: Allows use of DataCite test DOIs with the various import/migrate API calls

Closes #8028

**Special notes for your reviewer**: Does anyone know if the :import and :importddi API calls are intended to allow import of DOIs that don't have the right authority/shoulder to be managed by the Dataverse instance? If not, this PR could be simplified to just check, using credentials, whether the PID exists.

**Suggestions on how to test this**: Should be able to repeat the migrate tests with a test DOI now


landreev:
> Does anyone know if the :import and :importddi API calls are intended to allow import of DOIs that don't have the right authority/shoulder to be managed by the Dataverse instance?

I'm not sure what the original design rationale was. But having thought about it - this sounds like a legitimate use case/scenario. Maybe when somebody needs to be able to import datasets that are under their _legacy_ authority? Or to absorb datasets from another installation (the "how do I merge two Dataverse installations" question we occasionally get). 

So this was most likely the reason for that GET attempt in the original code, instead of .alreadyExists(). We could ask - but I'm inclined to just approve the PR as is.

landreev (approved):
Approving.

PR #8034, opened 2021-07-28
IQSS/8032 Add optional flag to trigger update of PID at provider during migrate

### LINKED ISSUE #8032: Allow PID updates to be made after migration
The new migrate API /api/datasets/\<id\>/actions/:releasemigrated tries to contact the PID provider to update metadata and the landing URL for the dataset. To support a use case where datasets are being migrated while the source datasets remain the official copy, a mechanism (e.g. a query param for the api) is needed to avoid updating the PID at the provider during this call. (When migration is complete, the /api/datasets/\<id\>/modifyRegistrationMetadata (or modifyRegistrationMetaAll) calls can be made to update the PID provider's records). 
PR to follow. Since accidentally changing a PID record at the provider would be problematic, this will switch the default behavior to not change the PID provider record.

qqmyers:
**What this PR does / why we need it**:  Disables updating the PID at the provider during the /api/datasets/\<id\>/actions/:releasemigrated call unless ?updatepidatprovider=true is added to the URL.

Closes #8032

**Special notes for your reviewer**: This is probably safer than the just merged API call which will try to change the PID at the provider by default.

**Suggestions on how to test this**: Should be able to repeat the migrate tests and verify that the DOI doesn't redirect without the param and does when it is true.

**Additional documentation**: added notes about the new param

### DIFF
--- a/doc/sphinx-guides/source/admin/dataverses-datasets.rst
+++ b/doc/sphinx-guides/source/admin/dataverses-datasets.rst
@@ -41,7 +41,7 @@ Recursively assigns the users and groups having a role(s),that are in the set co
     curl -H "X-Dataverse-key: $API_TOKEN" http://$SERVER/api/admin/dataverse/$dataverse-alias/addRoleAssignmentsToChildren
     
 Configure a Dataverse Collection to store all new files in a specific file store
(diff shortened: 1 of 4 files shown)


landreev (approved):
Agree, that this is safer, than to do the update automatically.

PR #8035, opened 2021-07-29
7790 remove files of deleted draft

### LINKED ISSUE #7790: S3 Storage, Deleting a draft of a published dataset does not remove files added in that draft.
Steps to reproduce:
1. create and publish a dataset with 1 file, eg. file1.txt
2. upload another file, file2.txt, which creates a new draft of the published dataset.
3. delete the draft of the dataset, notice file2.txt remains, happens on S3 and local storage too. (thx @qqmyers )

sekmiller:
**What this PR does / why we need it**: Deleting a dataset version from the dataset page does not remove the physical files that had been added to that version. Because of this there are files in storage that were never part of a published or current draft dataset version.

Closes #7790 Deleting a draft of a published dataset does not remove files added in that draft

**Special notes for your reviewer**: Not much since the code had already been written for the delete draft version via the api/

**Suggestions on how to test this**: Add files to a draft version. See that they are saved to the file system. Delete that draft and verify they have been removed while those files from previous published versions remain. 

No. unless we want to note that existing files created for deleted versions prior to this change will not be removed by this change. There's a separate issue for that clean up.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -2777,13 +2777,21 @@ public class DatasetPage implements java.io.Serializable {
 
     public String deleteDatasetVersion() {
         DeleteDatasetVersionCommand cmd;
+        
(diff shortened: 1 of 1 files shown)


landreev (approved):
Looks straightforward, approving.

PR #8055, opened 2021-08-05
document missing Grizzly JVM-option

### LINKED ISSUE #8054: Missed grizzly library
At QDR, I've noticed a 
`java.lang.NoClassDefFoundError: org/glassfish/grizzly/npn/AlpnServerNegotiator` error in the server log at startup. 
It is not fatal. I've tracked it down to a missing jvm-option in the domain.xml file:

<jvm-options>[1.8.0u251|]-Xbootclasspath/a:${com.sun.aas.installRoot}/lib/grizzly-npn-api.jar</jvm-options>

I think this is an option that should have been included in the 5.4 release notes for people whose domain.xml began under glassfish. (We included all of the [9|] options since we were moving to Java 11 and didn't catch that this option is for 1.8.0u251 and higher and should also have been included.)

As this is not part of our code, I think all that is needed for a fix is add a line to "Add/confirm the existence of the following JVM option to the <config name="server-config"><java-config> section:" as in step 5 of the v5.4 upgrade instructions to the next release notes (and/or add this jvm option to the list in the v5.4 notes).

donsizemore:
**What this PR does / why we need it**: without the JVM option in question, Payara throws a `java.lang.NoClassDefFoundError: org/glassfish/grizzly/npn/AlpnServerNegotiator` error at startup

Closes #8054 

**Suggestions on how to test this**: perhaps launch Payara without, then with this JVM option.

**Is there a release notes update needed for this change?**: yes

### DIFF
--- a/doc/release-notes/5.4-release-notes.md
+++ b/doc/release-notes/5.4-release-notes.md
@@ -274,8 +274,6 @@ Add the following JVM options to the `<config name="server-config"><java-config>
 
 `<jvm-options>[9|]--add-opens=java.logging/java.util.logging=ALL-UNNAMED</jvm-options>`  
 
-Add/Confirm the Existence of the Following JVM Option to/in the `<config name="server-config"><java-config>` section:
-
 `<jvm-options>[1.8.0u251|]-Xbootclasspath/a:${com.sun.aas.installRoot}/lib/grizzly-npn-api.jar</jvm-options>`
 
 6\. Start Payara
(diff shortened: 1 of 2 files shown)


doc/release-notes/5.4-release-notes.md
```diff
@@ -274,6 +274,10 @@ Add the following JVM options to the `<config name="server-config"><java-config>
 
 `<jvm-options>[9|]--add-opens=java.logging/java.util.logging=ALL-UNNAMED</jvm-options>`  
 
+Add/Confirm the Existence of the Following JVM Option to/in the `<config name="server-config"><java-config>` section:
```
qqmyers, line 277: I'd drop this for the 5.4 notes - all of the options above are the same case and the note about the path is at the start of step 5 already.

djbrooke (approved):

PR #8067, opened 2021-08-19
IQSS/8066 bag updates

### LINKED ISSUE #8066: BagIt Doesn't Get All Files
For Datasets that have files with the same filename, only one file (of that name) is added to the "data" directory in the bag, or appears in the manifest file. 

But the bag-info.txt file has "counted" ALL files, but the "Payload-Oxum" doesn't match the number of files in the manifest (or in the folder "data").

qqmyers:
**What this PR does / why we need it**: With the addition of file directory paths and the ability to have files with the same but different paths (added since the original Bag code), an update to the bag structure is needed to avoid collisions. This PR adds the directory path from Dataverse to the file's path within the bag's data dir to avoid collisions. There are also a couple minor updates to replace deprecated classes/simplify logic, and some adds in logging to make it clearer which Bags errors are coming from.

Closes #8066

**Suggestions on how to test this**: Configure to export Bags to the local file system and generate Bags for datasets that include files with the same name (e.g. readme.txt files with different paths). The Bag should contain all files and have the right file count in places such as the Payload-Oxum in the baginfo.txt file (the number of files is the integer after the decimal point). The number of entries in the manifest file and the metadata/pid-mapping.txt file should also match the number of files (pid-mapping has one extra entry for the dataset itself too).

**Is there a release notes update needed for this change?**: I added one if we want to use it. It is possible that Bags created before this change could be incomplete for datasets that had multiple files with the same name.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -208,6 +205,8 @@ public boolean generateBag(OutputStream outputStream) throws Exception {
 
         bagID = aggregation.get("@id").getAsString() + "v."
                 + aggregation.get(JsonLDTerm.schemaOrg("version").getLabel()).getAsString();
+        
+        logger.info("Generating Bag: " + bagID);
```
pdurbin, line 209: Probably better to lower this to logger.fine but not a big deal.

pdurbin (approved):
I didn't test this but it looks fine, seems to properly getting the file part from `directoryLabel`.

PR #7967, opened 2021-06-23
QDR/6886 external status labels

qqmyers:
**What this PR does / why we need it**: Adds a mechanism to display an additional status label on draft datasets to aid in coordinating with an external curation workflow mechanism

**Which issue(s) this PR closes**:

Closes #

**Special notes for your reviewer**: Relates to #6886 but doesn't cover all aspects. Created for QDR - will be discussed/reviewed/potentially modified/tested there before inclusion in QDR's version. This draft PR is to make it easier for community exploration.

**Suggestions on how to test this**: Call the API to set/delete the label (suggested values such as 'Queued', 'In Process', etc. - up to 32 chars, alphanumeric and spaces allowed), verify that the label is displayed and that any existing label is removed when the version is published.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #8070, opened 2021-08-19
Iqss/8069 dataset creation notices

### LINKED ISSUE #8069: Dataset creation notifications for admins
Per community discussion (https://groups.google.com/g/dataverse-community/c/_GlUWtJF1so/m/2KiWCclOAAAJ?utm_medium=email&utm_source=footer), there is interest in the ability to configure Dataverse to send email/notifications to admins when a new dataset is created. QDR created an initial capability with some hardcoded aspects. For the community version of Dataverse, this functionality needs to be optional (e.g. enabled via a setting), to use Bundled strings for i18n, and leverage the overall email/user notification mechanism.

The intent is to send messages like 
`Dataset "Notification Test 3" was just created by Jim Myers in the QDR Main collection.`

PR to follow.

qqmyers:

Closes #8069

**Special notes for your reviewer**: Note - this PR assumes that #7967 will be merged and includes a placeholder enum for it's UserNotification type. Both PRs add new notification types, so if both PR is just appended their own enum to the list, the notification table would get type 21 for both when testing. With the placeholder, this PR gets type 22 and leaves type 21 for the status notifications.

**Suggestions on how to test this**: Set the setting, create a dataset as one user and verify that accounts that can publish that dataset get an email/user notification about the dataset. (Also verify that the user doesn't get a second email - they already get the 'your' dataset has  been created message that already exists.)

**Is there a release notes update needed for this change?**: could announce it - release note included

**Additional documentation**: setting documented in config guide.


sekmiller (approved):
Looks good in general. When I started looking at the code I wondered why the notifications were being created in the new postDBFlush method instead of onSuccess.  Looking further I realized that notifications pre-dated onSuccess by several releases and generally are created outside of the commands. This might be something to address in a future refactoring ticket.

pdurbin (commented):
Overall, I think this pull request is fine but I'm confused by how it's named in both the docs and the code. Notifications are already sent on dataset creation by "CREATEDS". Now we have "DATASETCREATED". Maybe the docs and code should emphasize that the new thing sends notifications to admins?

Issue #6045, opened 2019-07-23
Bootstrap 5 Upgrade


djbrooke:
We should upgrade to Bootstrap 4.

djbrooke:
Approach to discuss:

- Investigate ability to phase in the changes page by page, login page or advanced search may be good candidates. As part of this, investigate whether or not the pages could be released page by page.
- If investigation indicates this is the appropriate path forward, then get a list of the pages in the application and start working through them.

TaniaSchlatter:
Good discussion with @mheppler. The technical effort to support two UI frameworks in the same app in order to slowly transition from Bootstrap 3 and 4 should not be necessary, and would be a throwaway effort. Small chunks can be used to carved up the ~20 pgs of the app in the development process, and released together. Alternate, single release approach:
- develop new front end code page templates in Bootstrap 4 for the new redesigned dataset and file page 
- apply the same page layout and various style changes to all pages and review
- release the new Bootstrap 4 UI for all pages together

TaniaSchlatter:
High-level overview of Bootstrap 4 and accessibility. It is not surprising from what I have seen, but disappointing that Bootstrap 4 default colors lead to _insufficient_ color contrast using WCAG 2.0 standards. https://getbootstrap.com/docs/4.0/getting-started/accessibility/

mheppler:
Recently I have suggested a few deliverables could be included in this app-wide UI effort. Adding those hear for reference and discussion.

- configurable custom Bootstrap themes, making it easier to apply branded color schemes to button, link and other UI components
- right-to-left language support
- selenium automated testing selector support

BPeuch:
Howdy-do, IQSS crew.

We like the potentialities afforded by Bootstrap 4 a lot, so we fully support the upgrade. Good luck!

How are things coming so far?

djbrooke:
Hey @BPeuch, this isn't something that we've started yet. Just wondering, can you let us know what things you're interested about specifically?

youssefOuahalou:
Hello @djbrooke  our web designer strongly advised us to do this update. Do you think this tool can help you? : http://upgrade-bootstrap.bootply.com/

djbrooke:
Thanks @youssefOuahalou, we'll take a look. This is a pretty big change that requires touching most of the pages in the application, so we haven't had a chance to plan and work on it yet. Watch this space! :)

BPeuch:
[...]

PR #8071, opened 2021-08-20
Use jquery 3.5.1

### LINKED ISSUE #8068: Remove local override of jQuery version, get version from Primefaces instead
In https://github.com/IQSS/dataverse/blob/develop/src/main/webapp/resources/primefaces/jquery/jquery.js version 3.4.1 of jQuery is defined. Instead of defining it here, we want to get it from the version of Primefaces defined in pom.xml (currently version 10, jQuery 3.5.1) so that we can stay up to date with changes.

qqmyers:
**What this PR does / why we need it**: Per the issue, Primefaces 10 uses 3.5.1 by default, but having a 3.4.1 version in our source tree overrides that. So this PR removes it.

Closes #8068

**Suggestions on how to test this**: Look for the jquery files used in the webpage, e.g. looking in the browser dev console network tab, and open them in a separate window to verify that the version shown in the source is 3.5.1. The change could potentially break UI functionality (anywhere we or PrimeFaces uses jquery) but I'd expect 3.5.1 would if anything work better than having us override to 3.4.1 (since PrimeFaces ships with it.)

**Is there a release notes update needed for this change?**: no? 

**Additional documentation**:


donsizemore:
Removing jquery.js per @qqmyers' instructions renders the following on the 5.6 release:

Screen Shot 2021-08-20 at 10 52 17

I can create a collection and dataset, but haven't tested much beyond that.

pdurbin (approved):
I haven't done any testing but this seems like a nice, easy fix, and simple to revert if we need to.

@donsizemore already commented that simply removing the 3.4.1 version from a server worked.

As of this writing, if you go to https://demo.dataverse.org/javax.faces.resource/jquery/jquery.js.xhtml?ln=primefaces&v=10.0.0 it will show "3.4.1". Once you remove the file, it shows the version Primefaces provides (3.5.1).

So it's all much more maintainable.

The last thing I'll mention is that in pull request #6079 both Bootstrap and jQuery were updated to an identical version (3.4.1). I know we want to upgrade to Bootstrap 4 in #6045 but meanwhile I don't know if it's crucial to keep Bootstrap and jQuery versions in sync.

PR #7154, opened 2020-07-31
6641 search ui bug

sekmiller:
**What this PR does / why we need it**:
This PR prevents the display of duplicate filter query facets on the Dataverse Page. It also removes the subtree and dvobject type facets from the selected filter queries display.

**Which issue(s) this PR closes**:

Closes #6641 Search UI Bug

**Special notes for your reviewer**:

**Suggestions on how to test this**:
From a sub-dataverse select multiple facets and verify that search works as expected.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Bug fix

**Is there a release notes update needed for this change?**:
No
**Additional documentation**:
None

PR #8079, opened 2021-08-26
7304 facet filter bug

### LINKED ISSUE #7304: Bug: Removing facet filter removes file count from search
Testing on demo.dataverse.org but this also occurs on qdr, both in 4.20 and in 5.0 testing:

Steps to reproduce:

1. Go to dataverse root, e.g. https://demo.dataverse.org/ -- note number of files
image

2. Filter by any facet, e.g. "Researcher" (the bug appears to be unrelated to which facet I choose)
image

3. Click the x next to the facet to remove the filter

Expected behavior: The search returns exactly as previously
Actual behavior: The number of files is displayed as 0
image

sekmiller:
Fixes an issue where if a user selects a facet and then deselects it the number of files available shown in the type link reverts to 0. also fixes a bug where clicking any type facet after selecting another facet can yield incorrect results.

Closes #7304 

The underlying issue is that the types are included in the fq filter query array and when this comes in conflict with the selected types string incorrect results are returned from solr.

Select and deselect various facets and verify that you get the expected results

no
none
**Additional documentation**:
none


pdurbin (approved):
I didn't run the code but I spoke with @sekmiller about it.

It sounds like the bug may have been introduced in pull request #7154 for Dataverse 5.0. I can't reproduce the bug on https://dataverse.unc.edu running Dataverse 4.16.

From what I understand, the fix involves keeping the URL the same with regard to types (Dataverses, Datasets, and Files) but setting the types as necessary before the query goes to Solr.

PR #8083, opened 2021-08-31
8063 Tworavens Example Manifest Update, Disclaimer

### LINKED ISSUE #8063: Cannot add TwoRavens as an external tool (Dataverse v5.5)
The following error : {"status":"ERROR","message":"contentType is required."} occurs when I run the curl command 
curl -X POST -H 'Content-type: application/json' --upload-file twoRavens.json http://localhost:8080/api/admin/externalTools
with the twoRavens.json downloaded from https://guides.dataverse.org/en/latest/installation/r-rapache-tworavens.html#e-enable-tworavens-in-a-dataverse-installation.
A solution is provided by Jim from the dataverse community (https://groups.google.com/g/dataverse-community/c/yA-g3wlFH48)

jim_solution

djbrooke:

In #8063 it was suggested we update the TwoRavens manifest so that it still works with our recent changes to external tools. @pdurbin noted that we should rely on the external developers to maintain these manifests, so along with the update I added a disclaimer and pointed people to the TwoRavens team for support. If this breaks again, we can consider taking more extensive action. 

Closes #8063

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/_static/installation/files/root/external-tools/twoRavens.json
+++ b/doc/sphinx-guides/source/_static/installation/files/root/external-tools/twoRavens.json
@@ -1,9 +1,11 @@
 {
   "displayName": "TwoRavens",
   "description": "A system of interlocking statistical tools for data exploration, analysis, and meta-analysis.",
+  "toolName": "TwoRavens",
   "scope": "file",
-  "type": "explore",
+  "types": "explore",
   "toolUrl": "https://tworavens.dataverse.example.edu/dataexplore/gui.html",
+  "contentType": "text/tab-separated-values",
   "toolParameters": {
     "queryParameters": [
       {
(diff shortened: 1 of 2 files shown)


doc/sphinx-guides/source/_static/installation/files/root/external-tools/twoRavens.json
```diff
@@ -1,9 +1,11 @@
 {
   "displayName": "TwoRavens",
   "description": "A system of interlocking statistical tools for data exploration, analysis, and meta-analysis.",
+  "toolName": "TwoRavens",
   "scope": "file",
-  "type": "explore",
+  "types": "explore",
```
pdurbin, line 6: Does this work? I think `types` takes an array.

`type` still works for backward compatibility if we want to revert this.

pdurbin (commented):
I didn't test it but I don't think the change from `type` to `types` will work as written.

Otherwise, it seems fine but someday I still think we should stop hosting this manifest file for TwoRavens. I know they're grandfathered in.

doc/sphinx-guides/source/_static/installation/files/root/external-tools/twoRavens.json
djbrooke: good catch, thx. Changed to type in https://github.com/IQSS/dataverse/commit/582f85e95702b04cbebed52a9b97fc29121d824c

pdurbin (approved):
I didn't test this, but it should work. I'm glad for the note to contact the TwoRavens team as necessary.

PR #8084, opened 2021-09-01
8073 guestbook response api

### LINKED ISSUE #8073: "Retrieve Guestbook Responses for a Dataverse Collection" fails for some collections
Trying to run the ["Retrieve Guestbook Responses for a Dataverse Collection" endpoint](https://guides.dataverse.org/en/latest/api/native-api.html#retrieve-guestbook-responses-for-a-dataverse-collection) to get the very large guestbooks of two collections (https://dataverse.harvard.edu/dataverse/covid-rt and https://dataverse.harvard.edu/dataverse/) failed after running for exactly 2 minutes and 20 seconds. Both times the following error message was returned:

> curl: (22) The requested URL returned error: 504

The curl command I used in my Mac terminal was 
`curl -O -J -f -H X-Dataverse-key:APIKEY https://dataverse.harvard.edu/api/dataverses/ALIAS/guestbookResponses`, where ALIAS was the alias of the collection (covid-rt or data_ncov) and APIKEY was the APIKEY for my superuser account.

The Harvard Dataverse Repository was at v5.6.

I was able to use the command to get the guestbook of another collection in the repository that had much fewer guestbook responses (and that number matched the results of a database query I did to get that collection's guestbook responses :)

sekmiller:
**What this PR does / why we need it**: The Retrieve Guestbook Responses api was failing for DVs with large numbers of responses. Creating a file for the responses was stepping on the streaming of results.

Closes #8073 Retrieve Guestbook Responses fails for some collections

Edit (L.A.): The guestbook for the covid-rt dataverse (the one with the regularly updated datasets/hundreds of versions) started this issue (it could not be downloaded because of the prod. timeout). So it should be a good test on a prod. db copy (`/api/dataverses/covid-rt/guestbookResponses`). Before this PR, you would start receiving output after about 2 min 40 sec. Should be ~40 sec now. Still not instant, but an improvement. 
Also, note the simplified curl command line in the guide. 

**Additional documentation**: updated doc for end user.

### DIFF
--- a/doc/release-notes/8073-gb-response_api_update.md
+++ b/doc/release-notes/8073-gb-response_api_update.md
@@ -1,0 +1,3 @@
+### Update Retrieve Guestbook Response API to remove default output file
+
+With this release the Retrieve Guestbook Responses for a Dataverse Collection will no longer produce a file by default. You may specify an output file by adding a -o $YOURFILENAME to the curl command.
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -587,7 +587,7 @@ You should expect a 200 ("OK") response and JSON output.
 Retrieve Guestbook Responses for a Dataverse Collection
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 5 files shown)


landreev (commented):
@sekmiller In its current state, did you have a chance to test this code with an actual giant guestbook? (on a copy of the prod. db, for example)

landreev:
@sekmiller The API in its current form still doesn't stream the output it generates. 
I'm going to move this back into "in progress", and switch to the slack thread.

landreev (commented):
It is definitely streaming now - as in, it starts sending the output back to the client as soon as it becomes available. 
**Unfortunately**, "as soon as becomes available" may still not be soon enough for that covid-rt dataverse guestbook (from the original report that started this issue) as to start downloading before the prod. timeout. This is because for that particular dataverse it is the database query itself that appears to be taking more than 2 minutes. I am assuming this is because of that monstrous mess of versions in every dataset there. (This is that dataverse with datasets that are being continuously updated, resulting in versions in the hundreds). I should have realized this sooner, when I tested the previous iteration of the PR.

Before moving this into QA, I'm trying to quickly determine if there is a simple database optimization approach that may make this API much more responsive. (our approach of generating the result list with just one db query may be counterproductive there).

landreev (approved):
Moving into QA.

PR #8086, opened 2021-09-03
8085 decoupling tsv

### LINKED ISSUE #8085: Decoupling TSV metadata schema parsing from API endpoint
Currently, any TSV file loaded into a Dataverse instance flows through `DatasetFieldServiceApi.loadDatasetFields()`.
This method isn't just the endpoint, it also contains all the code to parse and convert the TSV into living model objects.

This needs to be torn apart:
- Enable proper testing of the parser
- Enable reading TSV metadata schemas during unit & integration testing. A bare necessity for #5989
- Enable reading the files not only via API endpoint:
    - Possibility to move the code into a reusable module outside main codebase
    - Load TSVs from disk/WAR during startup and bootstrap (Hello, #5361)
- Enable validation on load
- Refactor some dusted code

poikilotherm:
Testing, reuse for tests and future features with TSV based metadata schema files need abstraction and parsing outside the API code, where currently the conversion happens.

This PR is about to change the underlying infrastructure for this, add lots of testing and implement actual checks on restrictions like no circular deps of fields, max depth of compounds etc etc etc.

This is a blocker :zap: for #5989

Closes #8085

None yet.

Extensive unit and integration testing will be provided.

Nope.

I don't think so, as it's part of the code infrastructure.

**Additional documentation**:
None yet. (Might make the docs on metadata blocks more precise in a few places)

### DIFF
--- a/doc/sphinx-guides/source/admin/metadatacustomization.rst
+++ b/doc/sphinx-guides/source/admin/metadatacustomization.rst
@@ -364,49 +364,51 @@ Each of the three main sections own sets of properties:
 #controlledVocabulary (enumerated) properties
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-+-----------------------+-----------------------+-----------------------+
-| **Property**          | **Purpose**           | **Allowed values and  |
-|                       |                       | restrictions**        |
-+-----------------------+-----------------------+-----------------------+
-| DatasetField          | Specifies the         | Must reference an     |
-|                       | #datasetField to which| existing              |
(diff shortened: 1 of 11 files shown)


poikilotherm:
Hey @bencomp thanks for the :rose: :tulip: :white_flower: !

I really had to look what I did there. And I have to say: I kind of gave up on this, as I also saw (some of) the problems you mentioned. The biggest obstacle, however, was the inability to reuse the TSV parser library with what we have as our custom metadata block format.

I initially created this to iron out the problems with the TSVs and Solr configuration. So my other approach, which I hope to continue at some point, can be found here: #8320 I started to write a validating TSV-format parser in there, as I also found there is no good and complete spec of our format.

PR #8087, opened 2021-09-03
support HTTP "Range" header #6937

### LINKED ISSUE #6937: Add byte offset functionality to StorageIO and Access API

pdurbin:

We'd like users to be able to download parts of files: the beginning (like `head`), the end (like `tail`), and a range in the middle.

Closes #6937

Support is not full. As noted in the docs, only a single range is supported and the "If-Range" header is not supported.

Try the examples in the docs.

No.

Yes.

**Additional documentation**:

Included.


src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -80,7 +81,28 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
             StorageIO<DataFile> storageIO = DataAccess.getStorageIO(dataFile, daReq);
 
             if (storageIO != null) {
+                List<Range> ranges = new ArrayList<>();
+                long fileSize = storageIO.getDataFile().getFilesize();
```
landreev, line 85: So the reason we tried this hack, of getting the size from the underlying DataFile `fileSize` db attribute (`storageIO.getDataFile().getFilesize()`) - that was because `storageIO.getSize()` was returning NULL, correct? 
That was simply because `.open()` hasn't been called on the StorageIO yet.
Note that `storageIO.getSize()` is working further down in this class; the method
`getContentSize()` relies on it. For example, in this line further down (430):
```
if ((contentSize = getContentSize(storageIO)) > 0 && ranges.isEmpty()) {
```
(Note that `getContentSize()` does something extra for tabular files - because of the variable name header that we add in real time)

But it's not just a matter of moving that `StorageIO.open()` line (106) 30 lines up either. I believe it should be the other way around. I.e., we should be checking that Range header and setting the offset much further down in the code (closer to the line 430 quoted above). At which point we have already checked all the extra arguments potentially passed to the Access API, and the method `getContentSize(storageIO)` can be trusted. 

Where we are now, we instantiate the StorageIO and open it for reading, to confirm that this is a valid file that's accessible in storage. But we don't know yet if the stream we have opened is what we are sending to the client - since there are some special cases (in addition to the saved original, there are format conversions and thumbnails, etc.) where we'll need to close it and open another one. 

Sorry for the delay with this. I'll add a bit more. 
This code is a bit of a mess of special cases and it can be confusing. (I have an open issue for refactoring it). But it should still be straightforward-ish to work out the remaining kinks.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -271,6 +293,8 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
 
                             if ("original".equals(di.getConversionParamValue())) {
                                 logger.fine("stored original of an ingested file requested");
+                                // Get orginal file size before storageIO is reassigned to prevent an NPE.
+                                fileSize = storageIO.getDataFile().getOriginalFileSize();
```
landreev, line 297: Note that switching the `fileSize` to the original size is not enough here, if the size of the tabular file has already been used to calculate the offset for a "last N bytes" ("bytes=-N") range type.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -433,12 +462,26 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
                             }
                         }
 
+                        // Default to reading the whole file. We'll count down as we write.
+                        long leftToRead = fileSize;
+                        if (!ranges.isEmpty()) {
+                            // Read a range of bytes instead of the whole file.
+                            // For now we only support a single range.
```
landreev, line 469: In the current implementation, if multiple ranges are supplied, are we simply using the first range, and ignoring the remaining ones? 
I'm wondering if it would be more appropriate to throw an exception ("multiple ranges not supported") instead.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -403,9 +427,14 @@ public void writeTo(DownloadInstance di, Class<?> clazz, Type type, Annotation[]
...
-                        if ((contentSize = getContentSize(storageIO)) > 0) {
+                        if ((contentSize = getContentSize(storageIO)) > 0 && ranges.isEmpty()) {
                             logger.fine("Content size (retrieved from the AccessObject): " + contentSize);
                             httpHeaders.add("Content-Length", contentSize);
+                        } else if (!ranges.isEmpty()) {
+                            // For now we only support a single range.
+                            long rangeContentSize = ranges.get(0).getLength();
+                            logger.fine("Content size (Range header in use): " + rangeContentSize);
+                            httpHeaders.add("Content-Length", rangeContentSize);
                         } else {
```
landreev, line 438: Note this case in the original code ("do nothing, if content size is not known").
This is under the assumption that there are some download cases (subsetting, specifically, and maybe others) where the final stream we are sending to the client is being generated dynamically, on the fly. In that case, `getContentSize(storageIO)` will return -1. So, if we know the size of the stream we are about to serve, we add this explicit `Content-Length` header. Otherwise, we don't - and then Glassfish/Payara automatically switches to the "chunked encoding" mode. (this is why the lines like `//httpHeaders.add("Transfer-encoding", "chunked");` are commented out - we no longer have to do it in our code ourselves). 
In the current PR, the code further below is written under the assumption that we will always know the total size of the download (and thus will know the value of `leftToRead`). It needs to be changed back to where this is not necessarily the case. (So ideally the `leftToRead` value should *only* be used when it's a Range request). And we should probably throw an exception when a Range is requested, but the final size is not known/i.e., the output is a dynamic stream.

landreev:
One other thing I'm debating/wondering is if we should throw a "not supported" exception not just on requests like subsetting, that are generated dynamically - but on ANY byte range requests except for the main physical file, or the saved original. 
I'm not sure I can think of a practical real world scenario where someone would want to read byte ranges on say, thumbnails. Or any other auxiliary files we store, except for the original of an ingested tab. file.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
landreev: OK, I see that you meant to do just that - throw an exception, if more than one range. Except I don't think the code is throwing it now, because of a missing ! in line 651 (?).

landreev: OK, I deleted one previous comment... It looks like the code IS throwing an exception when multiple ranges are supplied...

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -585,4 +628,78 @@ private long getFileSize(DownloadInstance di, String extraHeader) {
...
+        if (range != null) {
+            logger.fine("Range header supplied: " + range);
+
+            // Technically this regex supports multiple ranges.
+            // Below we have a check to enforce a single range.
+            if (!range.matches("^bytes=\\d*-\\d*(,\\d*-\\d*)*$")) {
+                throw new RuntimeException("The format is bytes=<range-start>-<range-end> where start and end are optional.");
+            }
+
+            if (ranges.isEmpty()) {
```
landreev, line 651: I was confused by this line - at first I thought it was missing a "!" - but on a closer look it appears to be completely redundant :)

landreev:
Let me know if you have any questions/need any assistance with this PR. 
It should really just be a matter of rearranging the code and logic you've already added a bit to make it fully ready.

pdurbin:
Thanks @landreev for discussing this today. I went ahead and pushed what I have so that you can look at the backend. I'll keep working on the tests and will think more about the problem we discovered with ranges for tabular files. (In short, the header is always written, even if you request the last bytes of a file.)

To summarize the decisions made:

- We're only using the `leftToRead` logic for range requests. Non-range requests get the same logic as in previous versions of Dataverse.
- We'll support range requests for original and tabular files and also (for now at least) the other types of auxiliary files, such as thumbnails. There may not be a use case, but it seems to work fine.

Note to self that I plan to add a release note, work more on the tests, and see what I can do about tabular files.

pdurbin:
@landreev in adab5546bb and added some logic to try to deal with variable headers. Basically:

- write the headers if no range
- if the range starts with 0, write as much as you can within the range

I'll keep working on tests and the release note.

src/main/java/edu/harvard/iq/dataverse/dataaccess/FileAccessIO.java
```diff
@@ -509,6 +509,7 @@ public FileInputStream openLocalFileAsInputStream () {
 
         try {
             in = new FileInputStream(getFileSystemPath().toFile());
+            in.skip(this.getOffset());
```
landreev, line 512: I believe this is what we want to rearrange: this in.skip() should not be happening here, in the open method, but in the setOffset() method itself. Because we want to be able to change that offset after the initial open. 

The setOffset() method will need to throw an IOException, if it's called while the InputStream is still null; or if the skip() call itself results in an IOException()

pdurbin: Yes, excellent idea. Implemented in 4724e11 and now offset seems to be working again, at least for non-tabular files. Thanks!

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
pdurbin: I'm going to mark this is resolved. Leonid and I have been chatting a fair amount and we're on the same page. I refactored the code to move it down, after the open in 29ea256 and then got offset mostly working again in 4724e11.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
pdurbin: Thankfully!

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
pdurbin: As part of the refactoring in 29ea256 I'm now preserving the old logic for non-range downloads. Only range downloads have this `leftToRead` concept now.

src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
pdurbin: Huh. I think you're right that it is redundant. I'll put this on the list.

pdurbin:
On Friday I merged the latest from develop into this branch to pick up the 5.8 pom.xml change but the API tests failed because certain things hadn't happened yet:

- The 5.8 dvinstall.zip hadn't been uploaded to the release yet. This has since been done.
- dataverse-ansible hadn't been bumped to 5.8 yet. This has since been done at https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible/commit/85e9dc68685a6c0eaa7ea449c0b425bdac3c0a65

This morning I made a small change to the release note to force the API tests to run and they passed.

PR #8088, opened 2021-09-07
6752 empty ds via api

### LINKED ISSUE #6752: API: empty datasets can be created resulting in 502 errors
Hi!

As written in the title, we can create "empty" datasets which in turn returns 502 when we try to access them via the dataverse page.

## How to reproduce
The endpoint used is the one described here http://guides.dataverse.org/en/latest/api/native-api.html#id47

`curl -H X-Dataverse-key:$dataverse-key-value https://demo.dataverse.org/api/dataverses/${dataverse}/datasets --upload-file example.json`

The contents of `example.json` are as follows:
```
{
  "datasetVersion": {
    "metadataBlocks": {
      "citation": {
        "fields": [
        ]
      }
    }
  }
}
```
Which is the bare minimum amount of data allowed to create the dataset.

## Current behaviour
This results in the following dataset being created:

image

And consequently when trying to enter the dataset page

image

## Expected behaviour
Dataset is not created in the first place because the required minimum information is not there

or

The dataset page is able to load with no initial information (maybe using some default values if not set?)

We are not really actively using the endpoint, but just wanted to let you guys know since I found this when testing some of our endpoints using dataverse.

sekmiller:
Prevents users from creating datasets with no dataset field values. The validator was only validating fields that exist in the version, not that there are fields at all.

Closes #6752 API: empty datasets can be created resulting in 502 errors

**Special notes for your reviewer**: Basic fix was to change the order of the Validate and "tidy up fields" method. When "Tidy" was run first it would remove all of the blank fields so they wouldn't be validated for required. There is nothing that addresses the existing prod datasets that were created without fields.

**Suggestions on how to test this**: try to use the api to create a dataset using a file where the fields array is empty - and with a dataset with some fields but not all required fields. Also, the Constraint Validation error messaging has been consolidated so you may want to retry the "update dataset metadata" endpoint for required fields. (the integration tests are all passing.)

**Additional documentation**:


scolapasta:
@sekmiller why were calls not failing for required fields? Rather than add a whole new validator, wouldn't the fix be to guarantee that is happening? (because otherwise you'd still be able to create a non empty dataset version with but with missing required fields?)

sekmiller:
@scolapasta there's a for loop at line 1632 in the validate method of the Dataset Version that loops through the "flat dataset fields" if that is empty then there's no checking for required fields happening. I didn't try the case where the fields array wasn't empty but didn't include all the required fields. in the ui version of create dataset I think all of the dataset fields are present so all of the fields are tested for required. Let me think about how to handle it in the api if fields are there but required fields are missing.

sekmiller:
I was able to create a dataset with just a title via the api and publish it. I only got a validation error if i made a draft in the ui and tried to save it without filling in the required fields.

sekmiller:
The real problem was that the Create Dataset Version Command was stripping out blank fields before validating the version so it was missing empty required fields  (which are caught in the Page because that does a version validation before sending to the command). With this fix I was able to remove the "No fields present" validator as suggested.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -258,8 +257,35 @@ public Response createDataset(String jsonBody, @PathParam("identifier") String p
             ds.setIdentifier(null);
             ds.setProtocol(null);
             ds.setGlobalIdCreateTime(null);
+            Dataset managedDs = null;
+            try {
+                managedDs = execCommand(new CreateNewDatasetCommand(ds, createDataverseRequest(u)));
+            } catch (WrappedResponse ww) {
```
scolapasta, line 263: The one thing I wonder about here is if there's a better place for doing this constraint logic so it's all in one place (instead of having to wrap the create call here with a try catch); as I know we have this logic repeated in a few places... Maybe out of scope, but @sekmiller could you at least comment on what you think?

scolapasta (commented):
@sekmiller I made one comment to get your thoughts, otherwise looks good (though the PR description still refers to the initial attempt to fix this.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
sekmiller: I moved the building of the error string to a util so that was able to reduce redundancy quite a bit.

scolapasta (approved):

PR #8130, opened 2021-10-04
7903 solr config hack

### LINKED ISSUE #7871: updateSchemaMDB.sh script not working
In our installation, the updateSchemaMDB.sh script does not perform as expected. We always get the message "Dataverse responded with empty file. When running on K8s: did you bootstrap yet?".

The problem is with the test ```[[ "`wc -l ${TMPFILE}`" < "3" ]]``` which always returns true, even when the result is much more than 3 lines. If changed to ```[[ `wc -l < ${TMPFILE}` -lt 3 ]]```, it correctly returns true or false depending on whether the file size is less than 3 lines.

I'll create a PR with the fix, but I'm holding off for now because of PR https://github.com/IQSS/dataverse/pull/7865#issuecomment-840436115.

### LINKED ISSUE #7903: Update schema.xml in-place

### LINKED ISSUE #8025: Unexpected termination of "conf/solr/8.8.1/updateSchemaMDB.sh" because of not correct empty-file checking
Currently there is a not correct "if-then" statement in the script to update the metadata schema in solr, see https://github.com/IQSS/dataverse/blob/develop/conf/solr/8.8.1/updateSchemaMDB.sh

```
### Fail gracefull if Dataverse is not ready yet.
if [[ "`wc -l ${TMPFILE}`" < "3" ]]; then
  echo "Dataverse responded with empty file. When running on K8s: did you bootstrap yet?"
  exit 123
fi
```

The output of `wc -l ${TMPFILE}` will be generated as a string with 2 values - `LINE_COUNT` and `${TMPFILE}`, what will be wrong interpreted in the comparison with "3" - as a comparison of strings and not as a comparison of digits. This bug was found during the testing of large amount of metadata in the dataverse instance. E.g. if `${TMPFILE}` has more than 1000 lines, the output of `wc -l ${TMPFILE}` will be `1000 ${TMPFILE}`, where only the first digit "1" (not the complete "1000") will be compared with "3", what as a result activates wrongly the echo message and exit command.

As a solution could be the code:

```
### Fail gracefull if Dataverse is not ready yet.
if [[ "`cat ${TMPFILE} | wc -l`" -lt "3" ]]; then
  echo "Dataverse responded with empty file. When running on K8s: did you bootstrap yet?"
  exit 123
fi
```

or 

```
### Fail gracefull if Dataverse is not ready yet.
if [[ "`<${TMPFILE} wc -l`" -lt "3" ]]; then
  echo "Dataverse responded with empty file. When running on K8s: did you bootstrap yet?"
  exit 123
fi
```

based on these posts about ["wc -l"](https://askubuntu.com/questions/741686/how-to-remove-the-filename-from-wc-l-output) and [comparison operators](https://tldp.org/LDP/abs/html/comparison-ops.html).

poikilotherm:
- Removes broken `updateSchemaMDB.sh`
- Introduces new `update-fields.sh` (single purpose tool, no side tasks like reload)
- Adds docs etc

Closes #7871 
Closes #7903
Closes #7904
Closes #8025

Looks like a huge beast to review... But most of it is test data. Have no fear! :smile: 

Try to update the Solr schema.xml with the script. e. g. by following the docs advise.
`curl "https://demo.dataverse.org/api/admin/index/solr/schema" | conf/solr/8.8.1/update-fields.sh conf/solr/8.8.1/schema.xml`

Easier and quicker: take a look at running the Shellspec tests on your machine!
```
curl -fsSL https://git.io/shellspec | sh -s 0.28.1
cd tests/shell
shellspec
```
(You can easily install shellspec via homebrew on Mac, too...)

Nope.

:battery: included

**Additional documentation**:
This is one half of a solution to the chicken-and-egg problem described in gdcc/dataverse-kubernetes#188. (The other half will be a new subcommand in [gdcc/dvcli](https://github.com/gdcc/dvcli) to convert TSV files to Solr schema without a running Dataverse instance.)

Pinging @Kris-LIBIS here as he opened #7904 which was never finished. He said they're relying on scripting in their installation, so this new script might be helpful. (We can discuss on Slack/Matrix, if you want! :smile: )


poikilotherm:
I started adding `shellspec` tests and will create a Github workflow, too. That way we can test the script on Ubuntu and MacOS and make sure things break when someone accidentally deletes sth. like the include guards.

poikilotherm:
Alright folks - consider this done for now.

I didn't add a check for MacOS. Looks like the old `bash` version is having troubles (or maybe the `sed` command not being GNU-flavored?) and I would need to violate A LOT of [shellcheck rules](https://www.shellcheck.net).

As I do not own a Mac, I cannot really develop and test. Please feel free to add this.

There are CI checks in place, running shellcheck via @reviewdog (good doggo!) and running `shellspec` tests on Ubuntu, CentOS 7 and RockyLinux 8.

kcondon:
@poikilotherm Thanks for the contribution -a nice improvement. One quick question: how would you recommend a multi host installation, where solr and payara are on different hosts, run your script to update schema.xml?

poikilotherm:
Glad you liked it @kcondon - thx for merging! :smile: 

> How would you recommend a multi host installation, where solr and payara are on different hosts, run your script to update schema.xml?

This depends on wheter you need to do this because of a metadata block change tied to a database migration or just for a small change in a metadata block. (And of course if you deployed in a way to use A/B testing/rollouts.)

Lets say this is a multi-host installation without A/B. Then you will need to update any Dataverse installation / load metadata blocks / do database migrations first, then update the schema.

As Solr is not required to deploy the Dataverse application (at least not as of writing this), you can use the API endpoint to retrieve the new blocks via `curl` or `wget` and pipe the output into the script: `curl "https://demo.dataverse.org/api/admin/index/solr/schema" | update-fields.sh schema.xml`. This does not affect a running Solr. If you had Solr running, make it reload via API call (`curl http://solrhost:8983/solr/admin/cores?action=RELOAD&core=collection1`) or restart. Depending on your changes to the schema you might need to reindex - this hasn't changed from before.

kcondon:
@poikilotherm Thanks, that's helpful. The fact the script doesn't affect Solr means it *could* be run local to the web server, then the schema.xml moved to the Solr box, since the admin api is blocked by default, or one could change the api policy temporarily to use unlock-key and run the curl command from the Solr box. Good point about A/B set up, I've used that approach elsewhere.

PR #8142, opened 2021-10-13
File directoryLabels were not copied when creating a new draft version.

janvanmansum:
Adds the directoryLabel when copying file metadatas.

N/a

This showed up when creating a new version with a metadata-update through the API. If there were files with directoryLabels, the directoryLabels were lost.

No

No

**Additional documentation**:


pdurbin:
@janvanmansum thanks! I was definitely using a different API endpoint that you were and from testing it just now (I added a test in 1a4e4f9) I realize that the publish step is important when reproducing the bug.

@kcondon I didn't try the curl example above but the test I added should be equivalent. Definitely a bug. Happy to help you reproduce it if you have trouble. I'll move this to QA.

PR #8202, opened 2021-11-02
Query counting script

landreev:
This PR does not touch any application code; and there is no corresponding issue. 
This is just to have an updated version of my old query-counting script in the develop branch. 

Closes #

**Additional documentation**:


scripts/database/querycount/count.pl
```diff
@@ -25,7 +25,7 @@
```
pdurbin, line 25: For me, "diff" is zero so nothing goes into the "tail" file.

10486132 - 10486132 = 0  (`$newsize - $pglogfilesize`) 

I can work around this by setting diff to the size of the file like this:

```
 my $newsize = (stat($pglogfile))[7];
 my $diff = $newsize - $pglogfilesize;
+$diff = $pglogfilesize;
 
 system "tail -c ".$diff." < ".$pglogfile." > tail";
```

But I'm probably just confused. Not sure what's going on here.

pdurbin (commented):
I'm having some trouble with count.pl.

scripts/database/querycount/count.pl
landreev: If the log file really isn't growing, between the start and the end of the task you are testing, maybe verbose logging isn't enabled (the `log_statement = 'all'` in the PostgresQL config). 

Wrong log file?  (it starts a new log file every day). 

What/how are you testing, anyway?

pdurbin: Whoops! I simply forgot how it works and obviously didn't study the README closely enough. I was thinking that it would operate on any old postgres log file (I have hundreds of these). Now I remember (and see in the README) that I'm supposed to get the script started, then go do something, and then go back to the script and tell it I'm done. Sorry for the noise.

pdurbin (approved):
Works on my machine. PostgreSQL 9.6 on Mac.

scripts/database/querycount/count.pl
landreev: Yes, that was the intended use case (human operator-defined "start" and "end" of the operation). If you want to count everything in the log file, then yes, setting that "diff" to the full size would achieve that. That's what @donsizemore is doing when using the script to count all the queries issued during as RestAssured run. 
Thinking about it, I'll add a flag for this mode - and this way Don would not need to hack the script in order to use it w/ Jenkins.

landreev:
(so yes, let me add a couple of things for Don's Jenkins non-interactive use case)

donsizemore:
I've just run Leonid's latest two scripts through my dataverse-ansible branch, now using ``--non-interactive`

integration test query_count reports:

```
Parsed and counted the queries. Total number:
207329

Queries, counted and sorted: 

  31408 SELECT ID, ADVANCEDSEARCHFIELDTYPE, ALLOWCONTROLLEDVOCABULARY, ALLOWMULTIPLES, description, DISPLAYFORMAT, DISPLAYONCREATE, DISPLAYORDER, FACETABLE, FIELDTYPE, name, REQUIRED, title, uri, VALIDATIONFORMAT, WATERMARK, METADATABLOCK_ID, PARENTDATASETFIELDTYPE_ID FROM DATASETFIELDTYPE
  17993 SELECT ID, DISPLAYORDER, IDENTIFIER, STRVALUE, DATASETFIELDTYPE_ID FROM CONTROLLEDVOCABULARYVALUE
  13399 SELECT t0.ID, t0.DTYPE, t0.AUTHORITY, t0.CREATEDATE, t0.GLOBALIDCREATETIME, t0.IDENTIFIER, t0.IDENTIFIERREGISTERED, t0.INDEXTIME, t0.MODIFICATIONTIME, t0.PERMISSIONINDEXTIME, t0.PERMISSIONMODIFICATIONTIME, t0.PREVIEWIMAGEAVAILABLE, t0.PROTOCOL, t0.PUBLICATIONDATE, t0.STORAGEIDENTIFIER, t0.CREATOR_ID, t0.OWNER_ID, t0.RELEASEUSER_ID, t1.ID, t1.AFFILIATION, t1.ALIAS, t1.DATAVERSETYPE, t1.description, t1.EXTERNALLABELSETNAME, t1.FACETROOT, t1.GUESTBOOKROOT, t1.METADATABLOCKROOT, t1.METADATALANGUAGE, t1.NAME, t1.PERMISSIONROOT, t1.STORAGEDRIVER, t1.TEMPLATEROOT, t1.THEMEROOT, t1.DEFAULTCONTRIBUTORROLE_ID, t1.DEFAULTTEMPLATE_ID FROM DVOBJECT t0, DATAVERSE t1
  13317 SELECT ID, AFFILIATION, CREATEDTIME, DEACTIVATED, DEACTIVATEDTIME, EMAIL, EMAILCONFIRMED, FIRSTNAME, LASTAPIUSETIME, LASTLOGINTIME, LASTNAME, POSITION, SUPERUSER, USERIDENTIFIER FROM AUTHENTICATEDUSER
  13088 SELECT ID, CONTENT, LANG, NAME FROM SETTING
  12409 SELECT ID, AUTHENTICATIONPROVIDERID, PERSISTENTUSERID, AUTHENTICATEDUSER_ID FROM AUTHENTICATEDUSERLOOKUP
```
and so on.

pdurbin (approved):
I'm approving this. I tried both interactive and non-interactive modes. I LOVE the reminder of what to do when starting up interactive mode ("hey, go do something that generates queries").

My only nitpick is that it should be "PostgreSQL" rather than "PostgresQL" but whatever. 😄

PR #8190, opened 2021-10-26
8177 - 5.8 Release Notes

djbrooke:
**What this PR does / why we need it**:

It's the release notes! 

**Which issue(s) this PR closes**:

Closes #8177 

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #8203, opened 2021-11-02
More efficient embargo labels

### LINKED ISSUE #8201: Dataverse page: Null pointer exception on stale index entries (develop/5.8 RC)
If one of the search results on the page is stale - i.e., an index entry for a dataset that no longer exists in the database - dataverse page fails ungracefully with a Null pointer exception and a 500 error. 
This was observed on the develop branch, v5.8 release candidate. The null pointer happens in SearchIncludeFragment.java, in the second line in the fragment below, added as part of the "embargo" PR: 
```
public boolean isActivelyEmbargoed(SolrSearchResult result) {
        if(result.getEntity().isInstanceofDataset()) {
```

qqmyers:
**What this PR does / why we need it**: As previously implemented, calculating whether to show an embargo label on datasets in search results significantly slowed displaying the results (per @landreev). This redesign indexes the date the last embargo on any file in a dataset version ends (as well as indexing the embargo end date on individual files) - as the 'epoch day' (long # of days since 1970). Upon retrieval, the isActivelyEmbargoed calculation now just has to compare that with the current date.

Closes #8201 

**Special notes for your reviewer**: As a byproduct, it removes the null pointer exception in #8201 - not sure if that should be considered to close that issue or not.

**Suggestions on how to test this**: Add embargoes, verify that embargo labels appear correctly on datasets and files in search results. Check with @landreev w.r.t. performance.

**Is there a release notes update needed for this change?**: Includes a change to schema.xml so the usual update solr info has to be included. Since no one could have created an embargo prior to this release, reindexing isn't necessary for this PR.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1253,6 +1268,7 @@ private String addOrUpdateDataset(IndexableDataset indexableDataset, Set<Long> d
                     }
                 }
             }
+            solrInputDocument.addField(SearchFields.EMBARGO_END_DATE, embargoEndDate.toEpochDay());
```
sekmiller, line 1271: I'm getting a null pointer here because none of my files are embargoed(?)

qqmyers: Sorry - try that - I switched to using toEpochDay at the end and forgot to check if embargoEndDate was null.

sekmiller (commented):
Looks reasonable. The null check works. thanks! I'll leave it to Leonid to check on the query load for now.

djbrooke:
@landreev, I'll send this one your way since in Slack you mentioned you'd take a look.

@qqmyers let me know if you'd like me to take a pass at the release note update @pdurbin mentioned

qqmyers:
@djbrooke - the release notes need to be whatever is standard now after the solr hack pr. schema.xml needs to be updated and then whatever the new instructions should be.

landreev:
@qqmyers @djbrooke 
So it sounds like all it needs to say is "This release requires a solr schema update. Please follow the *new* schema update instructions in the guide (link). For installations that are not using non-standard metadata blocks, all that's involved is copying the distributed schema.xml into your solr config directory." - something like that? with or without the double negative in the above?

Jim, what's the worst consequence of an installation skipping this schema update step? - would it only be the embargo status not shown in the search results? or are they going to have indexing failures when  attempting to index embargoed datasets?


> @djbrooke - the release notes need to be whatever is standard now after the solr hack pr. schema.xml needs to be updated and then whatever the new instructions should be.

landreev:
It doesn't appear to be necessary to actually run the build through the query counter - since the methods that were generating the extra instantiations proportionate to the number of files in the datasets on the page have been clearly taken out. I can still do that of course. But otherwise I'm ready to move it into QA.

> Looks reasonable. The null check works. thanks! I'll leave it to Leonid to check on the query load for now.

landreev (approved):
Moving to QA. It's still showing "changes requested" - but that's related to Phil's comment about a release note; and that's being addressed/is on the radar.

pdurbin (approved):
I'm switching to "Approve" because pull request #8190 about release notes already indicates that the Solr schema.xml needs to be updated.

qqmyers:
To answer the question a  few hours back - since the embargo end dates would be part of normal indexing (if you have embargoes), not updating the schema.xml will cause indexing to fail. From prior experience with curation labels, if you don't use the new feature, that field isn't included in what's sent to index and indexing works, so unless embargo is enabled, indexing would still work fine.

PR #8207, opened 2021-11-03
8204 - handle empty AllowedCurationLabels setting

### LINKED ISSUE #8204: Default value for :AllowedCurationLabels needs appropriate value or additional checks to prevent log error on create dv page load.
This is happening in v5.8 code on develop branch. If :AllowedCurationLabels isn't set, the default fails json parsing and throws a server log error when loading the create dataverse page. Setting the default to {} made the error go away and @qqmyers said the following:

The code to read that property was originally written to get a string where a default of "" was OK. When it became a json object to support multiple sets of labels, that default is no longer valid/can't be parsed as json. It could be fixed a couple ways - make the default "{}" / an empty json object, or just check for 0 length before passing to the json reader. The work-around/ a test would be to set the property to that default which should make the error go away.

qqmyers:
**What this PR does / why we need it**: Fixes warning in log when the setting is empty/not set. See issue for details.

Closes #8204

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
@@ -1,6 +1,8 @@
 package edu.harvard.iq.dataverse.util;
 
 import com.ocpsoft.pretty.PrettyContext;
+import com.rometools.utils.Strings;
+
 import edu.harvard.iq.dataverse.DataFile;
 import edu.harvard.iq.dataverse.Dataset;
 import edu.harvard.iq.dataverse.Dataverse;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -1,6 +1,8 @@
 package edu.harvard.iq.dataverse.util;
 
 import com.ocpsoft.pretty.PrettyContext;
+import com.rometools.utils.Strings;
```
pdurbin, line 4: This seems like a new dependency on rometools. Do we need it?

qqmyers: Good catch - it's gone.

pdurbin (approved):
Looks good as of f627ae3.

PR #8211, opened 2021-11-04
2912 revoke role command description add

### LINKED ISSUE #2912: Add more detail to ActionLogRecord
The ActionLogRecord table records (among other things) every command that is called. However, the detail of the info for the commands is very sparse. Recently, we've been trying to debug a support issue and it became obvious that a mechanism for a command to provide more detail would be useful.

In this issue, we track the design and development of that infrastructure. Adding details for individual commands should be tracked separately.

(but we'll probably start with a few, for example, for RevokeRoleCommand - add detail on the user whose role was revoked.

mdmADA:
**What this PR does / why we need it**: Recording specific details of what role has been revoked for a specific user in the actionlogrecord is valuable information. It also complements the AssignRoleCommand actionlogrecord detail.

**Which issue(s) this PR closes**: Goes towards closing 2912, not sure if it closes the issue completely

Closes #2912 

**Special notes for your reviewer**: A quick edit to override the describe() method - similar to AssignRoleCommand - to add more detail to the actionlogrecord

**Suggestions on how to test this**: Assign various roles to a user then Revoke each role and check the actionlogrecord after each to verify the revoke role describe() is writing the actionlogrecord row and with the correct detail.


djbrooke:
Thanks @mdmADA we'll take a look!

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RevokeRoleCommand.java
```diff
@@ -38,5 +39,9 @@ protected void executeImpl(CommandContext ctxt) throws CommandException {
         return Collections.singletonMap("",
                 toBeRevoked.getDefinitionPoint() instanceof Dataverse ? Collections.singleton(Permission.ManageDataversePermissions)
                 : Collections.singleton(Permission.ManageDatasetPermissions));
-    }	
+    }
+    
+    @Override public String describe() { 
+	    return toBeRevoked.getAssigneeIdentifier() + " has had the role: " + toBeRevoked.getRole() + " REVOKED on " + toBeRevoked.getDefinitionPoint().accept(DvObject.NameIdPrinter);
```
pdurbin, line 45: I assume...

```
" REVOKED on " + toBeRevoked.getDefinitionPoint().accept(DvObject.NameIdPrinter); 
```

.. is used instead of something shorter like...

```
" REVOKED on " + toBeRevoked.getDefinitionPoint().getDisplayName(); 
```

... because we want the database ID's as well as the name in the output.

pdurbin (approved):
I didn't test this but it should work. I left a comment about how the DvObject is printed. Seems fine.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RevokeRoleCommand.java
scolapasta: I'm assuming it was done this way, as that is what is in AssignRoleCommand:

https://github.com/IQSS/dataverse/commit/7cfc8bab11e24daca562f2a3ae16cf737e94487e#diff-634bf68267b3070eb7ee7afe2a4e2fde1e282e73d3ad4f553bfbe4c458a00794

(which was originally added by... @pdurbin :) )

pdurbin: homer

mdmADA: I followed the AssignRoleCommand way of writing it as I assumed it was approved in a code review so did the same.

If there's a better/preferred way, I can change it.

pdurbin: Nope. Seems fine to me. Consistent. Thanks.

mdmADA: I will create another issue for the revokeAll() command.

PR #8228, opened 2021-11-09
Fix 7874-Dataset-template-authors-ordering

### LINKED ISSUE #7874: Dataset template: authors ordering
Hi, 

I'm trying to build a dataset template but after I save the template, the order in wich I added the authors doesn't seem to stay that way and if I modify said template, dataverse reoders the list of authors again in another way.

Shouldn't my list of authors stay in the order that I added them?

Thanks in advance!

alejandratenorio:
**What this PR does / why we need it**: When saving a template, the order of authors is lost. This PR fixes that error.

**Which issue(s) this PR closes**: #7874

Closes #7874

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/TemplatePage.java
```diff
@@ -183,6 +184,10 @@ public String save(String redirectPage) {
         Long createdId = new Long(0);
         Template created;
         try {
+            Iterator<DatasetField> dsfItSort = template.getDatasetFields().iterator();
```
qqmyers, line 187: This looks like a fix, but - wouldn't the other checking in tidyUpFields also make sense before saving the template? If so, perhaps there's also an opportunity to refactor - make tidyUpFields() take dsv.getDatasetFields / template.getDatasetFields as the parameter (perhaps moving that method to a util class).

Also - seeing that the current tidyUpFields is called from other Commands - should the tidying here also be done in the template commands (on principle/to help if there's a future template API where order also needs to be defined)?

alejandratenorio: Hi @qqmyers,

Yes, my first idea was to run the tidyUpFields() method, however, tidyUpFields also removes all empty fields and this creates another error when editing a template,

image

If you save an author without affiliation or identifier, when you try to edit it, these fields are no longer displayed. What do you think?

qqmyers: Hmm, but that doesn't happen in dataset metadata. Probably some other method getting called to compensate in datasets. While it would be nice to have templates and datasets work the same, I don't know that its worthwhile to ask you to go investigating when what you've put in the PR is a fix for the issue. Unless somebody else has info or has a reason that not adding the empty field removal and trailing space removal that happens in tidyUpFields() is going to a big deal, I'd say it's OK as is - perhaps just adding a note in the code about this issue/difference with datasets.

scolapasta: So the method that makes sure tou don't have missing fields when editing a dataset is DatasetVersionUI (it also looks like it might handle the orderingsinc it has a sort:
    private List<DatasetField> initDatasetFields(boolean createBlanks) {

it is private because it's called by:
    public DatasetVersionUI  initDatasetVersionUI(DatasetVersion datasetVersion, boolean createBlanks) {

which does more than the template would need. 

But to avoid technical debt, we'd probably want to rely on trying to use these (possibly needing to be centralized) methods and tidyUpFields.

djbrooke:
Hey @alejandratenorio! Thanks for the PR, I hope you are well. @scolapasta and @qqmyers have provided some feedback. Please take a look when you have a chance.

I'll assign this to myself and keep this in the review column for now so that we don't lose track of this.

alejandratenorio:
Hi @djbrooke ,
Sorry, this week I was working in other tasks, I will check their comments and let you know my comments as soon as possible.

djbrooke:
Thanks @alejandratenorio, please let us know if you have any questions!

alejandratenorio:
Hi all,

I was checking the initDatasetVersionUI and initDatasetFields methods, it would be complicated to use them because they are linked to the datasetVersion,  we would have to rebuild a lot of methods. Do you think we should try it?

djbrooke:
Hi @alejandratenorio - I'll move this over to the review column so we can provide some guidance.

qqmyers:
I'm not sure how to connect things to the initDatasetFields method, but here's a ~concrete proposal:

- Change AbstractDatasetCommand.tidyUpFields(DatasetVersion) to be a new DatasetFieldUtil.tidyUpFields(List<DatasetField>, boolean removeBlanks) method.
- Only run the remove blank fields loop (https://github.com/IQSS/dataverse/blob/80a42de869fbff1bc84a6781ecf859256b3b79c7/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java#L124-L129) if removeBlanks==true
- Change the three commands that call tidyUpFields to use the new method tidyUpFields(dsv) -> tidyUpFields(dsv.getDatasetFields(), true);
- Change your code to call tidyUpFields(template.getDatasetFields(), false);

The net affect of that is just having your code also run the trimTrailingSpaces on the template fields (probably good), and a small amount of code consolidation. Whether this is worth doing, or if there's a better option, I'll leave to @scolapasta.

(Separate issue: when I was looking at DatasetVersionUI, it seemed to me that https://github.com/IQSS/dataverse/blob/80a42de869fbff1bc84a6781ecf859256b3b79c7/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java#L374-L395 defines and runs the same sort twice?)

scolapasta:
I think a similar approach as above could be done for initDatasetFields, ie. it could be removed it a DatasetFieldUtil taking in a list of Datasetfields (and some other parameters: a list of metadatablocks and an object (either the version or the template)).

That said, I'm also fine with an incremental improvement as the one @qqmyers suggested since it's a step on this direction. If you're willing to try more and also refactor initDatasetFields (so you wouldn't need the extra boolean for blanks, that's fine too.

@qqmyers I agree about the extra sort, so maybe that's a small bonus cleanup that can be done when this is worked on.

alejandratenorio:
Hi guys,

Thanks for clarifying the way, we are fine with an incremental improvement, two stages. This way we'll be working with tidyUpFields, then we do the PR and fix the problem in our installation. Later, we continue with initDatasetFields.

alejandratenorio:
Hi @qqmyers,

Happy new year! Returning to this issue, tidyUpFields method is a protected methid, CuratePublishedDatasetVersionCommand, UpdateDatasetVersionCommand and CreateDatasetVersionCommand can use  it because this classes extends from AbstractDatasetCommand. So, I have a problem with TemplatePage class, this class can´t see it. 
Do you think it would be worthwhile to do public & static tidyUpFields? Would there be any problem?

qqmyers:
@alejandratenorio - Happy New Year to you too! I was suggesting that rather than leave the in AbstractDatasetCommand, create a new DatasetFieldUtil class (e.g. in edu.harvard.iq.dataverse.util) with just this one new public static method. I think that's a little clearer than leaving it in the abstract command class and having non-commands call it. (It could go in some other util class if you see an appropriate one.)

alejandratenorio:
Hi @qqmyers 
I added the DatasetFieldUtil class and have done the commit, I would like to know your opinion.

src/main/java/edu/harvard/iq/dataverse/util/DatasetFieldUtil.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public static void tidyUpFields(List<DatasetField> datasetFields, boolean removeBlanks) {
+        if(removeBlanks){
+            Iterator<DatasetField> dsfIt = datasetFields.iterator();
+            while (dsfIt.hasNext()) {
+                if (dsfIt.next().removeBlankDatasetFieldValues()) {
+                    dsfIt.remove();
+                }
+            }
+            Iterator<DatasetField> dsfItTrim = datasetFields.iterator();
```
qqmyers, line 31: Why only trim if removeBlanks is true? If I understand, trimming would help by removing any trailing spaces in the values entered in a template, which I think is better than leaving the template alone and only trimming in datasets? This isn't a big deal if there's a reason to not trim in templates, but otherwise I'd suggest doing it all the time.

qqmyers (changes requested):
Looks good overall to me. I had one suggested change unless there's a reason it was done as it is.

src/main/java/edu/harvard/iq/dataverse/util/DatasetFieldUtil.java
alejandratenorio: Hi @qqmyers,
I looked inside the edu.harvard.iq.dataverse.util package and I haven't found a class where to drop the method, so I added a class.
About trimTrailingSpaces, if i don't run removeBlankDatasetFieldValues and run trimTrailingSpaces a nullpointerexception occurs in the templates, at this [line](https://github.com/CIMMYT/dataverse/blob/7874-Dataset-template-authors-ordering/src/main/java/edu/harvard/iq/dataverse/DatasetField.java#L643).

qqmyers: Ah - OK. Trim can't handle null values (which is what blank fields are I assume). I guess you could add a null check there but I'm OK leaving the PR as is too - it's a minor issue since dataset metadata will always get trimmed.

alejandratenorio: @qqmyers  thanks so much, I will keep an eye on this,  in case there are any other comments.

qqmyers (approved):

alejandratenorio:
> 

Hi @qqmyers,

It's done!

PR #8229, opened 2021-11-09
prevent (some) special characters in metadata #8018

### LINKED ISSUE #8018: dataset.xhtml and user input of special characters
dataset.xhtml in Dataverse-5.5 allows any old user to insert special characters into dataset metadata on dataset creation, which render the dataset metadata un-editable thereafter. On our instance nothing is written to server.log; instead the browser console conveys the rendering complaint when the user attempts to edit metadata. Here's a sample broken dataset on demo.dataverse.org:

Screen Shot 2021-07-22 at 13 26 16

Tough to say which UTF characters render the XML invalid, but it would be nice to sanitize them on user input.

pdurbin:

Special characters that are causing problems (such as an inability to edit dataset metadata) have been entered into the database. This pull request prevents specific special characters (form feed and start of text) from being entered and provides a SQL update script to remove them from the database.

Closes #8018

- Should we be handling more special characters? If so, which ones?
- Enforce at database level?
- See the "form feed" screenshot below. I would rather show the error if I can but the page becomes fairly broken. In Chrome you can see the error "Input is not proper UTF-8" under "partial-response" as in the screenshot below:

Screen Shot 2021-11-09 at 2 15 49 PM

- Try entering special characters like (form feed or start of text) into dataset fields such as description or title.
- Try creating or editing a dataset via API with the same special characters.

Yes, if you enter a "start of text" character, you should see an error like this: "An invalid character such as form field (U+000C) or start of text (U+0002) was entered."

Here's a screenshot:

Screen Shot 2021-11-09 at 1 58 01 PM

Note that if you enter a "form feed" character, you don't see the error and you are no longer to edit the page (the "Save Dataset" button is greyed out). See screenshots below.

Screen Shot 2021-11-09 at 1 58 25 PM

Screen Shot 2021-11-09 at 1 58 32 PM

Yes, added.

**Additional documentation**:

None.

### DIFF
--- a/doc/release-notes/8018-special-characters.md
+++ b/doc/release-notes/8018-special-characters.md
@@ -1,0 +1,1 @@
+After upgrading, do an "index all" to remove special characters that have been removed from the database.
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValidator.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValidator.java
@@ -11,6 +11,8 @@ import javax.validation.ConstraintValidatorContext;
 import edu.harvard.iq.dataverse.util.BundleUtil;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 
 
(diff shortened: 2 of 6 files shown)


pdurbin:
We discussed this in tech hours. A few points:

- Given the bad UX (potentially lost work) when a user enters a form feed, we'll try stripping characters out instead.
- I mentioned that I couldn't get the "start of text" character to cause a problem but Don clarified that it was on XML export.
- As predicted, if you go to demo (5.8) and enter a form feed in the description and have some other validation error (missing title, for example), the page breaks in the same, uneditable way (potential lost work). Screenshot below.

Screen Shot 2021-11-09 at 3 34 18 PM

pdurbin:
I'm closing this pull request in favor of pull request #8242 which implements the alternative strategy above of removing invalid characters rather than reporting on them.

>* See the "form feed" screenshot below. I would rather show the error if I can but the page becomes fairly broken. In Chrome you can see the error "Input is not proper UTF-8" under "partial-response" as in the screenshot below:

For posterity, I believe the issue to follow for the error above is https://github.com/eclipse-ee4j/mojarra/issues/4516

PR #8192, opened 2021-10-27
remove variable metadata API endpoints

pdurbin:
**What this PR does / why we need it**:

Both of these have been removed:

- /api/access/variable/{varId}/metadata/ddi:
- /api/meta/variable/{varId}:

**Which issue(s) this PR closes**:

Closes https://github.com/IQSS/dataverse-security/issues/43

**Special notes for your reviewer**:

See issue.

**Suggestions on how to test this**:

See issue.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

**Is there a release notes update needed for this change?**:

Perhaps. Let's discuss.

**Additional documentation**:

No.

Issue #6846, opened 2020-04-21
 edu.harvard.iq.dataverse.api.FilesIT.testForceReplaceAndUpdate Test Failing


djbrooke:
Expected status code <200> doesn't match actual status code <500>.

See build 425 on dataverse.org Jenkins for more info.

pdurbin:
At first I could reproduce the `FilesIT.testForceReplaceAndUpdate` failure locally but the error went away after I doubled the following 

```
-XX:MaxMetaspaceSize=512m
-XX:MetaspaceSize=256m
```

(I did this because I'm getting this `java.lang.OutOfMemoryError: Metaspace` in my logs when running even a single API test.)

Now I can't reproduce the original errors but when I run the whole API test suite (which I couldn't do with Glassfish 4.1 without it falling over) I get a seemingly random assortment of failing tests, just like we're seeing on Jenkins.

When I look at my logs, I see many cases of `org.postgresql.util.PSQLException: ERROR: deadlock detected` which we discussed but never fixed in #2460. I think these deadlocks should be discussed in tech hours.

I'm not sure what to do with this issue.

djbrooke:
Thanks @pdurbin for the details.

@scolapasta can you get automated testing on the docket for a future tech hours? We've been picking off the some failing tests one by one, but I saw on some recent runs that 4 new ones were failing so I'm not sure if we should re-evaluate. Generally I'm happy to spend sprint time in the automated test area. :)

poikilotherm:
About the MetaspaceSize and MaxMetaspaceSize: I stumbeld over https://dzone.com/articles/permgen-and-metaspace and my attention was catched here:

> Whenever there is a need to resize PermGen/Metaspace, JVM will do it as it does with the standard heap. Resizing those spaces requires a full GC, which is always an expensive operation. It can usually be observed during a startup when a lot of classes are being loaded. Especially if the application has dependencies on many external libraries. If there are a lot of full GCs during the startup, it’s usually because of that. If that case, increasing the initial size can boost the startup performance.

Has anyone taken a look at this during startup? During playing a bit with container memory limits, I noticed a few times rising and sudden drops in memory use. I wonder if this related to GC and if the metaspace is related to our very long deploy times, next to our massive amount of beans we are loading...

donsizemore:
[...]

PR #8237, opened 2021-11-10
Auxiliary File API Enhancements

### LINKED ISSUE #8235: Feature Request/Idea: Allow the Deletion of Auxiliary Files via API
Currently the Dataverse API includes endpoints to **add** or **download** Auxiliary Files as described in the documentation: https://guides.dataverse.org/en/latest/developers/aux-file-support.html

However, there is no way to **delete** an auxiliary file that is accidentally or incorrectly added. Please add an API endpoint to remove an existing auxiliary file.

An API User who has permissions to add auxiliary files

Testing with a Dataverse-integrated application prompted the request after duplicate auxiliary files were added as well as an incorrectly formatted auxiliary files. As these are auxiliary files we did not consider them falling under the same category as regular data files--which would need to be formally de-accessioned.

qqmyers:
This PR adds a few enhancements to the Aux file API:
- Auxiliary files can now also be associated with non-tabular files
- Improved error reporting
- The API will block attempts to create a duplicate auxiliary file
- Delete and list-by-origin calls have been added

Closes #8235

- w.r.t. stopping duplicates - this is an existing bug - you can add two aux files with the same format params which then causes downloads to fail (as there isn't a single response to the query to find the file.)
- w.r.t. list-by-origin - this was all that QDR needed and I thought perhaps that requiring the origin to find files might be useful, but one could redesign to just list all aux files and require the caller to filter the results and/or to make the origin param optional (e.g. query param)
- w.r.t. non-tabular files: QDR is using aux files with document file types (Word, PDF) and I'm not sure there's a good reason to limit aux files to certain mime types. Hopefully not controversial to just remove the restriction
- w.r.t. error reporting - again hoping there's no controversy - just trying to report the various ways the add call can fail more clearly.

As usual, just following the doc examples should be good. Specific things to try would be
- try to add an aux file twice (same formatTag/formatVersion - doesn't have to be same exact file) and verify that the second call fails
- add an aux to a non-tabular file
- try delete
- try to list files from the origin you use in uploads

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: api only

**Is there a release notes update needed for this change?**: draft provided - this is a tweak to an experimental API so probably nothing that needs to be prominent

**Additional documentation**:


djbrooke:
I added a reminder to expand on use cases at release time (but also so the build would be re-tried automatically :)). I'll unassign myself so someone can review the code.

raprasad:
thanks @qqmyers, :thumbsup:  for "API will block attempts to create a duplicate auxiliary file"

pdurbin:
@qqmyers heads up that I resolved merged conflicts with develop in dedd9f8. As I mentioned in the commit, they seem to be caused by pull request #8192 and only affected src/main/java/edu/harvard/iq/dataverse/api/Access.java

pdurbin:
That's odd, testForceReplaceAndUpdate failed with a 500 error according to https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-8237/3/testReport/junit/edu.harvard.iq.dataverse.api/FilesIT/testForceReplaceAndUpdate/

Screen Shot 2021-11-16 at 10 26 37 AM

I just ran it locally, individually, and it passed.

It might be a recurrence of this issue:

-  edu.harvard.iq.dataverse.api.FilesIT.testForceReplaceAndUpdate Test Failing #6846 

I'll probably push a small commit, maybe to docs, just to see if I can get the test suite passing.

qqmyers:
FWIW: Looks like the 500 error would be from the UpdateDatasetVersionCommand failing - since the test does a replace file right before the metadata update, it could just be a conflict due to indexing still running on the dataset, etc. - i.e. a random race condition that a delay or checking and waiting for the lastindex timestamp to update would fix.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -119,6 +134,35 @@ public AuxiliaryFile lookupAuxiliaryFile(DataFile dataFile, String formatTag, St
             return null;
         }
     }
+    
+    @SuppressWarnings("unchecked")
```
pdurbin, line 138: We only use this `@SuppressWarnings("unchecked")` a few times in the code base and (without looking deeply into it) it seems like a code smell. Do we need it? What problem does it solve? I commented it out locally and the app still compiled fine.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -119,6 +134,35 @@ public AuxiliaryFile lookupAuxiliaryFile(DataFile dataFile, String formatTag, St
...
+    public List<AuxiliaryFile> listAuxiliaryFiles(DataFile dataFile, String origin) {
+        
+        Query query = em.createQuery("select object(o) from AuxiliaryFile as o where o.dataFile.id = :dataFileId and o.origin = :origin");
```
pdurbin, line 141: My understanding is that we prefer our queries like the one below (typed and with the query added to the entity):

    TypedQuery query = em.createNamedQuery("AuxiliaryFile.findAuxiliaryFiles", AuxiliaryFile.class);

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -119,6 +134,35 @@ public AuxiliaryFile lookupAuxiliaryFile(DataFile dataFile, String formatTag, St
...
+                
+        query.setParameter("dataFileId", dataFile.getId());
+        query.setParameter("origin", origin);
+        try {
+            List<AuxiliaryFile> retVal = (List<AuxiliaryFile>)query.getResultList();
+            return retVal;
+        } catch(Exception ex) {
+            return null;
```
pdurbin, line 149: Should we handle the exception from getResultList somehow?

pdurbin:
Whoops, when I merged develop I accidentally deleted the "list by origin" API endpoint. I added it back in 59d0cbb.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -510,6 +512,48 @@ public String tabularDatafileMetadataDDI(@PathParam("fileId") String fileId,  @Q
...
+        JsonArrayBuilder jab = Json.createArrayBuilder();
+        auxFileList.forEach(auxFile -> {
+            if (isAccessAllowed || auxFile.getIsPublic()) {
+                JsonObjectBuilder job = Json.createObjectBuilder();
+                job.add("formatTag", auxFile.getFormatTag());
+                job.add("formatVersion", auxFile.getFormatVersion());
+                job.add("fileSize", auxFile.getFileSize());
+                job.add("contentType", auxFile.getContentType());
+                job.add("isPublic", auxFile.getIsPublic());
+                job.add("type", auxFile.getType());
```
pdurbin, line 550: A 500 error is thrown if type is null. Type is optional.

pdurbin (changes requested):
Overall, looks good.

Bugs:

- "Listing Auxiliary Files for a Datafile by Origin" gives a 500 error and NullPointerException if no type is given. You could consider using NullSafeJsonBuilder but whatever solution is fine.

Would appreciate these:

- Yes, it would be nice to have a "list all aux files given a file id" endpoint. Users will know the file id. They may not remember the origin they used.
- I added a few tests in b66fb55a72 but only for the new "delete" and "list by origin" APIs. It would be nice to have tests for the new "non-tabular files can have aux files" rule and the "duplicate aux files" are prevented rule. Possibly the new error handling could be tested too. Maybe tests around permissions.

Things to think about:

- Various comments I left in the review.
- When should we move the aux file API docs from dev/experimental to prod?
- 404 gives a somewhat confusing "no API endpoint exists" error. This happens for any API endpoint that returns a 404, I think, so it's not unique to this code. It looks like this: `{"status":"ERROR","code":404,"message":"API endpoint does not exist on this server. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.","requestUrl":"http://localhost:8080/api/v1/access/datafile/1441/auxiliary/myApp","requestMethod":"GET"}`

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
qqmyers: The smell part is the broad use of things like` (List<'SomeEntityClass'>)query.getResultList();` with a query that doesn't specify the class type in the list. Adding the  `@SuppressWarnings("unchecked")` is an improvement - hopefully indicating that a developer has checked the query and knows that only the right class can be returned in the list since the compiler can't (as with ~all of the @Entity queries in Dataverse). In general, I think I add the `@SuppressWarnings("unchecked")` because Eclipse flags the warning in the UI. 

That said, the better approach is to use a TypedQuery<AuxiliaryFile> class to start with. I've changed the code I added and the other examples in this Bean that weren't type checking before.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
qqmyers: Changed.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
qqmyers: Since the exceptions for getResultList(); all indicate real server errors, I've removed the try block so they bubble up to be reported via the API. I also changed the existing getSingleResult() call where one exception (NoResultException) is OK and should return null whereas other exceptions represent a real server error.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
qqmyers: Fixed. FWIW: from looking at the database, it looks like many of these are optional, even though I think the add API call wouldn't allow them to be null. Not sure if it is worth adding non-null constraints in the AuxiliaryFile class.

qqmyers:
re: main review comment:

- I added an additional API call to list all aux files, regardless of origin. (FWIW: I sorta like the security-by-obscurity aspect of requiring the origin to be known - since these files have meaning to some app, making them available to other apps/users who may not know what constraints there are on aux file existence/contents/etc. seems potentially problematic. That said, it is only by-obscurity so some future mechanism would really be needed to stop apps from messing with each other's files, and perhaps at that point the api to list all origins could become an admin/superuser call.)
- I added tests for trying to re-upload the same aux file and to test that the list api returns the right number of files. I did not add any test to assure that aux files can be added to non-tabular types - just seems odd to have a test that is mostly about checking that an old constraint has been removed. (If desired, the csv DataFile used in the test could be changed to a json one.)
- when to move from experimental? - not sure. This PR makes things more robust but I wouldn't be surprised if there are more changes w.r.t. UI and/or organization of aux files, etc. so I think it's fair to stay experimental as a warning that things may still be in flux.
- re 404: I think this was discussed when the updated error handling classes were added and I thought there was some reason it wasn't changed (no agreement on better wording?). Not sure - and perhaps it was just about merging the original changes without waiting for further updates. In any case, I'd agree that this is confusing, and that a better message - perhaps even just some more RESTful generic message - might help - i.e. the resource represented by this URL does not exist. 

Assuming the IT tests pass, I think this is ready to go forward again unless the changes raise further points.

pdurbin:
@qqmyers thanks for all the changes. I haven't looked through them all (and probably will) but I wanted to give you a heads up that AuxiliaryFilesIT is failing with a 500 error.

qqmyers:
Fixed the tests. Also discovered that the original code didn't calculate checksums correctly. Code is now fixed, but all existing entries have bad checksum values. They could be recalculated from the files themselves, but it's not something that can be addressed just by a flyway script. Removing/re-adding files via api would probably work, so a script to find/list all entries and then delete/re-add could be made. Since QDR is only using this in testing, I think Harvard is probably the only place with many aux file entries that need updating.

djbrooke:
Thanks @qqmyers and @pdurbin. I'll move this back over to review.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
pdurbin: Looks good with the TypedQuery. Thanks.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
pdurbin: Yay, TypedQuery. Thanks.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
pdurbin: Thanks for adding NullSafeJsonBuilder. I think we can revisit adding any extra constraints when we make this an official API.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -73,12 +82,20 @@ public AuxiliaryFile processAuxiliaryFile(InputStream fileInputStream, DataFile
...
             // by removing the auxfile from storage.
             storageIO = dataFile.getStorageIO();
-            MessageDigest md = MessageDigest.getInstance(systemConfig.getFileFixityChecksumAlgorithm().toString());
-            DigestInputStream di 
-                = new DigestInputStream(fileInputStream, md); 
-  
-            storageIO.saveInputStreamAsAux(fileInputStream, auxExtension);          
-            auxFile.setChecksum(FileUtil.checksumDigestToString(di.getMessageDigest().digest())    );
+            if (storageIO.isAuxObjectCached(auxExtension)) {
+                throw new ClientErrorException("Auxiliary file already exists", Response.Status.CONFLICT);
```
pdurbin, line 86: 409 (Conflict) seems slightly exotic for our code base but I guess it's fine.

"For example, you may get a 409 response when uploading a file which is older than the one already on the server resulting in a version control conflict." https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

pdurbin (approved):
Looks good. I went ahead and merged the latest from develop.

PR #8242, opened 2021-11-15
remove invalid characters from input #8018

### LINKED ISSUE #8018: dataset.xhtml and user input of special characters

pdurbin:

Currently users can input invalid characters into the database via both the GUI and the API. This pull request strips them out before they are written to the database.

Closes #8018

- Should we be handling more special characters? If so, which ones? ~~If you enter all of https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt as a description we have the same problem of not being able to edit the dataset. The error looks something like this in Chrome: `error on line 13 at column 71: Input is not proper UTF-8, indicate encoding ! Bytes: 0xEF 0xBF 0xBE 0x26`~~ (fixed as of 2d8fe44)
- Any ideas for enforcing this at the database level as well? The pull request uses Bean Validation.

- Try entering special characters like (form feed or start of text) into dataset fields such as description or title. See the SQL script or DatasetFieldValidator.java for the full list of characters that are being removed.
- Try creating or editing a dataset via API with the same special characters.

The special characters should be removed before hitting the database as in the screenshots below.

Screen Shot 2021-11-15 at 3 38 01 PM
Screen Shot 2021-11-15 at 3 38 31 PM

No.

Yes, included.

**Additional documentation**:

I didn't include any but perhaps we could mention under "Adding a New Dataset" that invalid characters will be removed: https://guides.dataverse.org/en/5.8/user/dataset-management.html#adding-a-new-dataset


donsizemore:
I'm testing this branch for @pdurbin and am iterating through the Groups of Bad Characters at https://raw.githubusercontent.com/minimaxir/big-list-of-naughty-strings/master/blns.txt

I broke metadata editing and XML export with `"Byte order marks", U+FEFF and U+FFFE, each on its own line.` server.log sez `SAX error Character reference "&#xfffe" is an invalid XML character.`

pdurbin:
@donsizemore thanks! `U+FFFE` was definitely causing trouble. As of 2d8fe44 (where I started stripping it out) I was able to copy https://raw.githubusercontent.com/minimaxir/big-list-of-naughty-strings/master/blns.txt into the dataset description and things seem fine. That is, I'm able to edit metadata and all the exports seem ok. Off to review.

For fun, here's how the dataset description looks:

Screen Shot 2021-11-17 at 10 20 51 AM

pdurbin:
A pull request with a SQL script got merged ahead of this one so I just merged the latest from develop and renamed the script (bumping the version) to avoid conflicts.

landreev (commented):
> Any ideas for enforcing this at the database level as well? The pull request uses Bean Validation.

I personally feel like using bean validation is the way to handle this; and I feel like it should be sufficient. (If anyone has different ideas, please speak up). 

Ready to approve this otherwise, everything makes sense. But I *may* need to ask you to rename the flyway script one more time, since 5.8.0.2 is used in my branch (after multiple renaming also!), that's currently in QA. 
Let's consult with Kevin in the morning, as to what his preference is. I'm still working on fixing something in my branch; and yours may actually be easier to merge first.

landreev (approved):
Consulted with @kcondon. Since there's a good possibility this PR can be merged ahead of mine, moving it into QA as is.

PR #8246, opened 2021-11-18
allow multiple required subfields #8236

### LINKED ISSUE #8236: Cannot create Dataverse collection after editing required metadata fields

- Add a new Dataverse collection
- Enter in the required content (alias, type, etc.)
- Under “Metadata Fields” deselect “Use metadata fields from Harvard Dataverse” (or whatever Dataverse installation you're using)
- Expand citation metadata
- Make all the fields under “Related Publication” required instead of optional (See screenshot)
- Try to save the page

When creating a new Dataverse collection and trying to edit the required fields.

https://dataverse.harvard.edu/dataverse.xhtml

**What happens?**

The page fails to save and an error is shown:

`Error – Command edu.harvard.iq.dataverse.engine.command.impl.CreateDataverseCommand@11c92da3 failed: Transaction aborted If you believe this is an error, please contact Harvard Dataverse Support for assistance.`

From the logs, it also appears there are some SEVERE messages related to duplicate key errors around this time. 

Anyone with Dataverse collection create permissions.

I expected a new Dataverse collection to be created and I expected those fields under related publication to be required for new datasets.

5.6 (dataverse.harvard.edu), but this occurs in 5.8 (demo.dataverse.org) as well. 

I looked, and will take another look, but none popped out.

**Screenshots:**

Screen Shot 2021-11-10 at 10 08 32 AM

pdurbin:

Without a fix like the one in this pull request, as of Dataverse 5.4 (which contains pull request #7608), you cannot mark more than one subfield as required such as "Other ID Agency" and "Other ID Identifier" in the screenshot below.

Screen Shot 2021-11-18 at 3 50 40 PM

Closes #8236

If using a set to ensure uniqueness is preferred, we can merge 291dbbe0d2 into this pull request. To compare that solution against develop: https://github.com/IQSS/dataverse/compare/8236-required-subfields-list2set

See the issue. My preferred way to test is the following:

- Create dataverse, save.
- Edit dataverse, change "Other ID Agency" to required (subfield of "Other ID"), save.
- Edit dataverse, change "Other ID Identifier" as required (subfield of "Other ID"), save.

On develop or https://demo.dataverse.org you get this:

- BOOM! PSQLException: ERROR: duplicate key value violates unique constraint "unq_dataversefieldtypeinputlevel_0"

No.

I don't think so. It's just a bug fix.

**Additional documentation**:

None.


pdurbin:
> If using a set to ensure uniqueness is preferred, we can merge [291dbbe](https://github.com/IQSS/dataverse/commit/291dbbe0d21f2b7dbb32299d5d8519a1037de046) into this pull request.

As discussed at standup, @sekmiller and @scolapasta prefer the "set" approach so I merged the commit above in.

@sekmiller please take a look and move it to QA or kick it back to me, if necessary. Thanks!

sekmiller (approved):

kcondon:
Good news: error isn't happening.

Issues found:
1. Saving all subfields at once doesn't work, it only saves the one, whether saving all at once or one at a time.
2. For the ones that are marked as required, they are not enforced as required in the ui on dataset save.

pdurbin:
@kcondon good catch. Sorry, I hadn't tested the "set" approach as much. My original solution seems to work fine so I reverted back to that. I also merged the latest from develop.

I moved this back to review.

qqmyers:
FWIW: my guess is that the Set approach fails because the equals method on DataverseFieldTypeInputLevel relies on id comparisons, which aren't set until those objects are added to the db in the commands. With two objects with null ids, the Set calling equals gets true, so the Set only ever gets one entry.

scolapasta:
> FWIW: my guess is that the Set approach fails because the equals method on DataverseFieldTypeInputLevel relies on id comparisons, which aren't set until those objects are added to the db in the commands. With two objects with null ids, the Set calling equals gets true, so the Set only ever gets one entry.

Good point - the default equals methods for entities have always had this challenge. There is a part of me that wants us to clean this up and do it "right", but not sure if it's worth the effort right now, especially since it would be good to get this fix into 5.9.

scolapasta:
@qqmyers @pdurbin or anyone else? Any opinions on whether we try to clean up the equals now and use Set vs take on technical debt for this? (the main. reason for the former is that Set is exactly what we really want and is less code - so is technically the right approach for this). If we do go with this PR as is then lets at least add a TODO about changing to a Set and needing to fix equals for that to work.

qqmyers:
My opinion: With only 4 subfields, I think adding another line to equals to check -  if both ids are null, check that the four subfield values are equal -  wouldn't be hard. So if the goal is Set, I'd do it now rather than a ToDo. This is one of many classes though that only check for id in equals, so we probably want a note in either case (i.e. //FYI: Set works here because the DatasetFieldTypeInputLevel class has an equals method that can compare objects that haven't yet been assigned ids in the database.)

scolapasta:
@qqmyers I was about to agree with you, but then I realized that one of those 4 subfields is a dataverse (i.e. an object), so I'm fairly certain then you'd have to change the equals on that (you can set these inputfieldlevels on a new Dataverse). 

Since we'd like to get this onto 5.9, let's just go with this fix here, then, and a comment. I can add that comment and then move to QA.

scolapasta (approved):

kcondon:
Retested, this is still not working for me only now it saves no compound fields as required.

PR #7502, opened 2021-01-06
add more mime types based on use frequency

atrisovic:
**What this PR does / why we need it**:

Adds new mime types based on frequency of use in HDV:

```
MISSING MIME TYPES                    FREQUENCY IN HDV
application/vnd.isac.fcs                          2794
chemical/x-xyz                                    2315
application/vnd.ms-excel.sheet.macroenabled.12    1261
application/postscript                             833
application/vnd.flographit                         732
application/cnt                                    476
application/x-tgif                                 192
application/vnd.ms-pki.stl                         190
text/x-pascal                                      186
text/x-vcard                                       181
text/x-fortran                                     163
application/vnd.ms-excel.sheet.macroEnabled.12      61
text/vnd.fmi.flexstor                               60
application/vnd.oasis.opendocument.chart            43
application/vnd.wolfram.mathematica.package         43
text/x-java-source                                  42
text/x-sh                                           40
application/download                                40
application/winhlp                                  36
application/msaccess                                31
application/x-research-info-systems                 31
application/vnd.tcpdump.pcap                        30
application/java-vm                                 29
application/x-compressed                            27
application/gml+xml                                 27
application/x-iwork-keynote-sffkey                  27
application/x-r-data                                26
application/photoshop                               25
application/vnd.palm                                24
application/vnd.ms-cab-compressed                   24
application/macbinary                               23
text/x-log                                          21
application/rat-file                                20
application/x-photoshop                             20
audio/x-ape                                         19
application/x-graphpad-prism-pzfx                   19
application/x-download                              18
application/vnd.oasis.opendocument.text             18
application/vnd.google-earth.kml+xml                17
text/x-objcsrc                                      17
```

**Additional documentation**:

References:
Data formats:
https://en.wikipedia.org/wiki/XYZ_file_format
https://en.wikipedia.org/wiki/Flow_Cytometry_Standard
https://www.stata.com/manuals13/g-4conceptgphfiles.pdf
https://en.wikipedia.org/wiki/STL_(file_format)
https://www.digipres.org/formats/mime-types/#text/x-comma-separated-values
Archives:
https://en.wikipedia.org/wiki/Optical_disc_image
http://fileformats.archiveteam.org/wiki/Microsoft_Excel
Code:
https://en.wikipedia.org/wiki/PostScript
The rest:
https://fileinfo.com/extension/cnt
https://en.wikipedia.org/wiki/ReStructuredText
https://en.wikipedia.org/wiki/Shapefile
[...]

PR #8262, opened 2021-11-30
Identify geojson file types

### LINKED ISSUE #8261: GeoJSON files marked as "Unknown"/"application/octet-stream"
Hi Dataverse team,

I'm reporting this issue as I've developed a file previewer for GeoJSON files, but don't believe we can use it without being able to enable it for this specific content type, and in order to do that, GeoJSON files need to be individually identified. Hoping this is possible!

1. Upload a GeoJSON file to Dataverse
2. View the file type, which is "Unknown" in the UI
3. In the API, in the files list the content type for the file is "application/octet-stream". [Here's an example from the Harvard DV](https://dataverse.harvard.edu/api/datasets/:persistentId/?persistentId=doi:10.7910/DVN/YQX0SW).

When uploading GeoJSON files.

**What happens?**

After uploading a GeoJSON (.geojson) file to Dataverse, the file type displayed in the UI is "Unknown". Through the API, I can see the content type being set as "application/octet-stream". I have checked older files and also uploaded a few geojson files to the Harvard demo (v 5.8). [Here's an example of a GeoJSON file in the Harvard DV](https://dataverse.harvard.edu/file.xhtml?persistentId=doi:10.7910/DVN/YQX0SW/GLBKGD&version=1.0).

All users

The MIME type should be "application/geo+json". I'm hoping it's possible to detect this (as opposed to it being marked as "application/json" or "text/plain").

5.1/5.8

kaitlinnewson:

Mark .geojson files with the correct file type, and add the GeoJSON display name.

Closes #8261 

Upload a .geojson file and confirm the file type/content type is set correctly. An example file is linked in the issue.

No

Based on #5934 and #7502, I think a full re-index is needed to update the facets, as well as optionally running the "redetect file type" API on existing GeoJSON files.

### DIFF
--- a/doc/release-notes/8261-geojson.md
+++ b/doc/release-notes/8261-geojson.md
@@ -1,0 +1,3 @@
+.geojson files are now correctly identified as GeoJSON files rather than "unknown".
+
+A full re-index is needed to update the facets, as well as optionally running the "redetect file type" API on existing GeoJSON files.
--- a/src/main/java/propertyFiles/MimeTypeDetectionByFileExtension.properties
+++ b/src/main/java/propertyFiles/MimeTypeDetectionByFileExtension.properties
@@ -4,6 +4,7 @@ dbf=application/dbf
 dcm=application/dicom
 docx=application/vnd.openxmlformats-officedocument.wordprocessingml.document
 emf=application/x-emf
+geojson=application/geo+json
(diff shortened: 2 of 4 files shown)


pdurbin (approved):
Looks good. Heads up to @kaitlinnewson that I added a release note.

PR #8265, opened 2021-12-01
8258 preview restricted file fix

### LINKED ISSUE #8258: Preview tab not showing for restricted files even though user has permission to access
1. Install and enable a previewer.
2. Create dataset with one or more restricted files.
3. Grant permission to a user to one or more restricted file.
4. User logs in and navigates to the dataset
5. User sees Preview 'eye' icon in the Files list for the restricted file they have access to
6. User clicks the 'eye'
7. the file.xhtml page is rendered for the user
8. there is no Preview tab

* When does this issue occur?
Only for restricted files

In every dataset page when a Previewer is enabled.

* What happens?
See above ^^^
It happens because FilePage.java calls FileUtil.isPreviewAllowed() which returns false if the file is restricted (and does not check if the user actually has access). FilePage should call FileDownloadHelper isPreviewAllowed() as that is the method called by all other classes/objects and allows the eye icon to show in the files list in the first place.

Anyone who has access through a role.

The preview tab in the file.xhtml should show.

This will occur for every version of dataverse since the file previewers were created. I tested it on demo.dataverse.org v5.8

**Screenshots:**

See the Dataverse Users Community conversation with Phil: https://groups.google.com/g/dataverse-community/c/kjhzKAuncSs
for screenshots

mdmADA:
Allows the Preview tab to show in file.xhtml for restricted file that logged in user has permission to view.

Closes #8258

Grant access to a restricted file to a user.
As that user, navigate to that file and click the preview button.
In file.xhtml, see that the file now has the preview tab and shows the file's contents.

No

No

**Additional documentation**:
None


scolapasta (changes requested):
Looks good, but requesting some small changes to reduce some code duplication.

djbrooke:
Hi @mdmADA - I'll move this back into community dev for now, but please ping us when the changes are complete or if you have questions. Thank you and keep the PRs coming!  :)

src/main/java/edu/harvard/iq/dataverse/FilePage.java
```diff
@@ -832,7 +832,7 @@ public boolean isPreviewAllowed() {
             // Always allow preview for PrivateUrlUser
```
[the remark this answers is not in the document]
mdmADA: I updated the code as suggested which of course makes sense @scolapasta. 

Perhaps checking to ensure preview of embargoed unrestricted files (only those with permissions to access it can see it?) works as the Dataverse team expects would be a good idea before merging.

Let me know if further changes are required...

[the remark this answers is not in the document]
mdmADA: I have to test it further before merging so if you leave it in community dev, I will let you know when it's good to go.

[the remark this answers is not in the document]
mdmADA: It's now ready to go. The previewTab wasn't rendering in my code because I had an extra } in my code.
I can confirm that I have tested it and that the previewTab now shows for restricted files that a user has permission to download.
The dataverse team should probably test it with an unrestricted but actively embargoed file to make sure the behaviour is what the Dataverse team expects.

djbrooke:
Thanks @mdmADA. @scolapasta, tagging you since you did the initial review.

src/main/webapp/file.xhtml
```diff
@@ -337,7 +337,7 @@
                                 </ui:include>
                             </p:tab>
                             <p:tab id="previewTab" title="#{bundle['file.previewTab.header']}" 
-                                   rendered="#{FilePage.toolsWithPreviews.size() > 0 and FilePage.previewAllowed}">
+                                   rendered="#{FilePage.toolsWithPreviews.size() > 0 and fileDownloadHelper.isPreviewAllowed(FilePage.fileMetadata)}">
```
qqmyers, line 340: FWIW: At QDR, I had changed this to fileDownloadHelper.canDownloadFile() - both here and line 402 below. A couple things:
- Not sure if that's better, but I think it works with PrivateUrls, so I'm not sure the isPreviewAllowed method is needed any more? (It's possible that there is some other difference that makes this work at QDR).
- Independently, I think you'd want the same change for 402?
- Since the canDownloadFile method checks for the embargo state (which is called by isPreviewAllowed too), I think the logic should be OK w.r.t. embargoes (embargoes only stop people without permissions (and don't let you request access, etc. - they don't override if you can see the draft dataset, download the files etc as a curator, privateURL user, etc.)

mdmADA: Thanks for the comments Jim. I'm not sure if I am the one to reply but will anyway.

-  the fileDownloadHelper.isPreviewAllowed() prevents users who don't have permission to view/download a file from previewing it but allowing those who do have permission to preview it. So it's needed imo but maybe I am missing something for you to say it may not be necessary?
- the fileDownloadHelper.isPreviewAllowed() is already called at line 402 (if that's what you meant).
- I tested the changes with embargoed files and it 'works' but I would just like the Dataverse team to verify the behaviour is as expected.
- testing it with privateUrl would also be beneficial.

scolapasta: I think what @qqmyers is suggesting is that we could get rid of fileDownloadHelper.isPreviewAllowed() all together and just use fileDownloadHelper.canDownloadFile() in its place. (i.e. a user can preview if and only if they can download).

Can we confirm that is definitely the case? It sounds so to me? If so it may still be worth modifying fileDownloadHelper.canDownloadFile with the user check from isPreviewAllowed, since that seems less expensive than checking permissions.

qqmyers: I agree. I missed that 402 was already doing this - just saw that it wasn't changed in the PR - so a change there is only needed if we change to canDownloadFile. I see @scolapasta's point that the isPreview method is more efficient for PrivateURLs since it bypasses the checks. It seems like that same logic could/would make sense for canDownloadFile as well - PrivateURL users can always download files so we could do the check for them in canDownloadFile and make all of those checks more efficient for privateURL users too (and if that check works in the isPreview method now, it woudl work in canDownloadFile even if I'm wrong about the permission checks. 

Again, I think this is all minor but having additional methods around that end up doing the same thing are just a pain when making other changes so it would be nice to get rid of it with a small change here.

mdmADA: Thanks for the comments. And I agree that these issues are minor in the grand scheme of things and other code that needs to be implemented but the Preview functionality should work. Reducing the number of calls that essentially do the same thing, or result in the same thing, is also beneficial. 

I can add the privateUrl user check to canDownloadFile() and replace the calls to isPreviewAllowed() and delete the latter altogether. And then test again.

mdmADA: So I added the user is PrivateUrlUser check to canDownloadFile() and then changed all previous calls to isPreviewAllowed() to canDownloadFile() and deleted isPreviewAllowed() from FileDownloadHelper.

I believe the previewTable is now rendered as it should but will leave it to @sekmiller to verify the behaviour is as expected for privateUrls, embargoes, restricted/unrestricted, etc.

scolapasta (approved):
Looks great! Thanks for the extra clean up.

kcondon:
@mdmADA This generally works but noticed that responses in the text area type that span multiple lines is handled differently by different opening apps -dataverse converts cr to spaces, excel keeps up to the first cr and puts the rest in a new row, first column, numbers appears to do something similar to Excel. Anyway that content can be confined to the cell it is supposed to be in, maybe writing out response in quotes?

here are some examples:
Screen Shot 2021-12-13 at 5 07 35 PM
Screen Shot 2021-12-13 at 5 08 30 PM
Screen Shot 2021-12-13 at 5 09 17 PM
Screen Shot 2021-12-13 at 5 20 04 PM
Screen Shot 2021-12-13 at 5 20 54 PM
Screen Shot 2021-12-13 at 5 22 30 PM

kcondon:
@mdmADA Would you refresh this branch from develop? It also needs the version to be updated. Thanks!

PR #8268, opened 2021-12-02
8266 fix handle command click

### LINKED ISSUE #8266: handleCommandLinkClick() should call isRequestAccessPopupRequired() so that the Request Access popup is shown only when there are terms of use/access
Create a dataset with no terms of access, assign a guestbook.
The handleCommandLinkClick() calls the isDownloadPopupRequired() method that returns true if there is a guestbook.
The requestAccessPopup is shown even there are no terms of use or access and it shouldn't be.

* When does this issue occur?
Always

It occurs for the request access command link buttons (the single request access to a single file)

* What happens?
See above^^

All users who want to request access to a single file

If there are no terms of use or access for the dataset, then the popup should not appear when the user clicks 'Request Access' - access should just be requested.

This is for all versions of dataverse from when the command link was introduced for the single file functionality.
I tested it on demo.dataverse.org (v5.8) and confirm the bug.

Not sure.

mdmADA:
Fixes FileDownloadHelper.handleCommandLinkClick() method

Closes #8266 

Create a dataset. No terms of use, no terms of access. Add guestbook.
Click 'Request Access' for a single file.
The Request Access will be submitted with no empty popup beforehand even though there is a guestbook.

No

No

**Additional documentation**:
None


sekmiller (approved):
Looks good. Thank you for this fix.

PR #8274, opened 2021-12-03
IQSS/8263-Fix Google analytics button tracking

### LINKED ISSUE #8263: Analytics: Button tracking broken in some cases
"Can you add an issue - looks like mheppler changed the analytics-code.html example a couple years back and, with the change, the button tracking only works for the file edit mode (haven't checked) but not for the file table in view mode."

qqmyers:
The change in the reverted commit tried to make an element selector in the analytics-code.html file more precise. Unfortunately, the sub-element  chosen only appears in some of the views that have the buttons that should be tracked, so the change stopped tracking of those buttons (the download menus in the file table - only the editFilesFragment tracking worked, tracking in the filesFragment broke with this change).

The only change in this PR is to revert the commit that made this change: Changed selector in analytics exmaple (sic) to be more precise [#6725], commit 1e5951fe35257dc0d0a110590475fa54ae55db36.

Closes #8263

**Suggestions on how to test this**: @kcondon discovered this via a null pointer error reported in the java console. With the updated anayltics-code.html configured, that error should be gone when a download button is clicked in the view mode for a dataset. One could also check Google for the actual button click event (which, if it doesn't appear might indicate other issues with analytics reporting). 

**Is there a release notes update needed for this change?**: - The fact that Google analytics wasn't catching download button events for view mode - what you'd expect for published dataset versions - might be relevant, but I don't know that many people were using that (not actually sure, but there haven't been any (?) questions about it.)

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html
+++ b/doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/analytics-code.html
@@ -117,7 +117,7 @@
       var row = target.parents('tr')[0];
       if(row != null) {
         //finds the file id/DOI in the Dataset page
-        label = $(row).find('div.file-metadata-block > a').attr('href');
+        label = $(row).find('td.col-file-metadata  a').attr('href');
       } else {
         //finds the file id/DOI in the file page
         label = $('#fileForm').attr('action');


scolapasta (approved):

kcondon:
I think one piece that is confusing is that the top download button, not the inline download button was working so perhaps was masking the broken inline download functionality. We do use it in production (it is enabled and I can view the events) but not sure whether we are actively reporting on it. Once tested, I'll reopen the original issue. Thanks!

qqmyers:
FWIW: Yes - the issue is in detecting which row/file you're downloading so it only affected the download kebabs

kcondon:
Opened https://github.com/IQSS/dataverse/issues/8280 to track inline download issue

PR #8277, opened 2021-12-03
IQSS/8276 - skip validation when embargo isn't enabled.

### LINKED ISSUE #8276: File Page menus broken in 5.8 w/o embargo set
As reported by @pdurbin in slack, trying to delete a file using the edit menu on the file page fails. Further investigation shows that some items in the Access (Citation downloads) and the Edit (Restrict, delete) menus/the form is broken. The underlying issue is 1) that this menu is in the same form as the hidden embargo dialog and when a delete, restrict, etc. option is chosen, the embargo form is validated as well, and 2) that validation attempt ends with a null pointer exception (FWIW nulls are handled w.r.t. attempting to read the setting but there's an error that causes the validation to run and there's eventually an error in trying to convert the null max date to a string).

* When does this issue occur? 
Menus on the file page, only when the MaxEmbargoDurationInMonths setting is not set (or is set to 0 to disable embargoes)

Only on file page. (Which is somewhat strange as the edit menu in the files table on the dataset works and it nominally uses the same code.)

* What happens?
Actions from a menu fail silently in the UI, with a null pointer exception in the log.

5.8

**Screenshots:**

*

qqmyers:
**What this PR does / why we need it**:  This fixes items in the file page menus that are now (v5.8) broken when the new embargo functionality is not enabled (which would be the default). The change avoids an NPE when the empty embargo (from the dialog that isn't shown) is found to be invalid and then getMaxDate() returns null and the code tries to format it for display in the error message (which also wouldn't show).

Closes #8276

**Special notes for your reviewer**: any ideas as to why the validation runs and caused a problem on the file page but the same menu on the dataset appears to work fine would be welcome. This PR should fix the problem but if there is a difference between the .xhtml code in the two places that stops the overall embargo validation (and potentially the restrict access validation currently being developed) from running, it might be worth adding that to the file page as well.

**Suggestions on how to test this**: Unset the MaxEmbargoDurationInMonths setting and/or set it to 0 and confirm that all entries in the file page access and edit menus work.

**Is there a release notes update needed for this change?**: Could report that this fixes a bug introduced in 5.8 that breaks some actions on the file page menus (see the issue for the list I know of).

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
+++ b/src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
@@ -360,36 +360,39 @@ public class SettingsWrapper implements java.io.Serializable {
     
     public void validateEmbargoDate(FacesContext context, UIComponent component, Object value)
             throws ValidatorException {
-        UIComponent cb = component.findComponent("embargoCheckbox");
-        UIInput endComponent = (UIInput) cb;
-        boolean removedState = false;
-        if (endComponent != null) {
-            try {
-                removedState = (Boolean) endComponent.getSubmittedValue();
-            } catch (NullPointerException npe) {
-                // Do nothing - checkbox is not being shown (and is therefore not checked)
+        if (isEmbargoAllowed()) {
+            UIComponent cb = component.findComponent("embargoCheckbox");
+            UIInput endComponent = (UIInput) cb;
+            boolean removedState = false;
+            if (endComponent != null) {
+                try {
+                    removedState = (Boolean) endComponent.getSubmittedValue();
+                } catch (NullPointerException npe) {
+                    // Do nothing - checkbox is not being shown (and is therefore not checked)
+                }
             }
-        }
-        if (!removedState && value == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/SettingsWrapper.java
```diff
@@ -360,36 +360,39 @@ public boolean isEmbargoAllowed() {
...
             throws ValidatorException {
-        UIComponent cb = component.findComponent("embargoCheckbox");
-        UIInput endComponent = (UIInput) cb;
-        boolean removedState = false;
-        if (endComponent != null) {
-            try {
-                removedState = (Boolean) endComponent.getSubmittedValue();
-            } catch (NullPointerException npe) {
-                // Do nothing - checkbox is not being shown (and is therefore not checked)
+        if (isEmbargoAllowed()) {
```
qqmyers, line 363: The only change is adding this if (and then indenting the whole method so github shows a big change) which stops validation from running when embargoes are not enabled (and maxEmbargoDate is then null).

scolapasta (approved):

PR #8282, opened 2021-12-06
allow users to override content type for aux files #8241

### LINKED ISSUE #8241: auxiliary file download: JSON file extension/type is lost

pdurbin:

Some auxiliary were being saved with the wrong content type (MIME type) but now the user can supply the content type on upload, overriding the type that would otherwise be assigned.

Closes #8241

Markdown files can now be given a better content type of `text/markdown` but on download, no file extension is added. This is because Tika's getDefaultMimeTypes doesn't include "text/markdown". I left a TODO about how perhaps we could look up more esoteric content types in the MimeTypeDetectionByFileExtension.properties file we maintain (.geojson was added recently, for example).

Run the curl commands. You should be able to set the content type to whatever you want.

No.

Included but feel free to leave it out.

**Additional documentation**:

Included.

### DIFF
--- a/doc/sphinx-guides/source/developers/aux-file-support.rst
+++ b/doc/sphinx-guides/source/developers/aux-file-support.rst
@@ -10,14 +10,15 @@ To add an auxiliary file, specify the primary key of the datafile (FILE_ID), and
 .. code-block:: bash
 
   export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
-  export FILENAME='auxfile.txt'
+  export FILENAME='auxfile.json'
+  export FILETYPE='application/json'
   export FILE_ID='12345'
   export FORMAT_TAG='dpJson'
   export FORMAT_VERSION='v1'
   export TYPE='DP'
   export SERVER_URL=https://demo.dataverse.org
   
-  curl -H X-Dataverse-key:$API_TOKEN -X POST -F "file=@$FILENAME" -F 'origin=myApp' -F 'isPublic=true' -F "type=$TYPE" "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
+  curl -H X-Dataverse-key:$API_TOKEN -X POST -F "file=@$FILENAME;type=$FILETYPE" -F 'origin=myApp' -F 'isPublic=true' -F "type=$TYPE" "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
 
 You should expect a 200 ("OK") response and JSON with information about your newly uploaded auxiliary file.
 
--- a/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
@@ -101,7 +101,7 @@ public class AuxiliaryFileServiceBean implements java.io.Serializable {
 
             // The null check prevents an NPE but we expect mediaType to be non-null
             // and to default to "application/octet-stream".
-            if (mediaType != null) {
+            if (mediaType != null && (!mediaType.toString().equals("application/octet-stream"))) {
                 auxFile.setContentType(mediaType.toString());
             } else {
                 Tika tika = new Tika();
--- a/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
@@ -99,6 +99,8 @@ public class AuxiliaryFileServiceBean implements java.io.Serializable {
             storageIO.saveInputStreamAsAux(di, auxExtension);
             auxFile.setChecksum(FileUtil.checksumDigestToString(di.getMessageDigest().digest()));
 
+            // The null check prevents an NPE but we expect mediaType to be non-null
+            // and to default to "application/octet-stream".
             if (mediaType != null) {
                 auxFile.setContentType(mediaType.toString());
             } else {
--- a/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
@@ -25,6 +25,7 @@ import javax.persistence.TypedQuery;
 import javax.ws.rs.ClientErrorException;
 import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.ServerErrorException;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.apache.tika.Tika;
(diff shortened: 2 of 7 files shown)


src/main/java/edu/harvard/iq/dataverse/api/DownloadInstanceWriter.java
```diff
@@ -577,6 +577,7 @@ private boolean isAuxiliaryObjectCached(StorageIO storageIO, String auxiliaryTag
         }
     }
     
+    // TODO: Return ".md" for "text/markdown" as well as other extensions in MimeTypeDetectionByFileExtension.properties
```
qqmyers, line 580: FWIW: We have FileUtil.generateOriginalExtension for tabular files but it uses a hardcoded list rather than the properties file.

qqmyers (commented):
Looks good!

doc/sphinx-guides/source/developers/aux-file-support.rst
```diff
@@ -10,14 +10,15 @@ To add an auxiliary file, specify the primary key of the datafile (FILE_ID), and
...
+  export FILENAME='auxfile.json'
+  export FILETYPE='application/json'
   export FILE_ID='12345'
   export FORMAT_TAG='dpJson'
   export FORMAT_VERSION='v1'
   export TYPE='DP'
   export SERVER_URL=https://demo.dataverse.org
   
-  curl -H X-Dataverse-key:$API_TOKEN -X POST -F "file=@$FILENAME" -F 'origin=myApp' -F 'isPublic=true' -F "type=$TYPE" "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
+  curl -H X-Dataverse-key:$API_TOKEN -X POST -F "file=@$FILENAME;type=$FILETYPE" -F 'origin=myApp' -F 'isPublic=true' -F "type=$TYPE" "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
```
landreev, line 21: I'm *very mildly* rubbed the wrong way by the fact that there are now 2 instances of `type=` in the command line above... Would have been better if we had named that `type` parameter something more specific (`auxtype`? `dvtype`? - idk, something indicating that this is our, proprietary type...) to differentiate it from the standard mime type... But I'm not going to propose changing it now. 

This API is for software clients only; nobody should ever be using it on the command line except an occasional developer.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -97,8 +99,12 @@ public AuxiliaryFile processAuxiliaryFile(InputStream fileInputStream, DataFile
             storageIO.saveInputStreamAsAux(di, auxExtension);
             auxFile.setChecksum(FileUtil.checksumDigestToString(di.getMessageDigest().digest()));
 
-            Tika tika = new Tika();
-            auxFile.setContentType(tika.detect(storageIO.getAuxFileAsInputStream(auxExtension)));
+            if (mediaType != null) {
```
landreev, line 102: Quick question - is `mediaType` (`formDataBodyPart.getMediaType();`) definitely going to be `null` here, if not supplied by the client? - As opposed to defaulting to something like `application/octet-stream`?

doc/sphinx-guides/source/developers/aux-file-support.rst
pdurbin: I had the same thought: two cases of "type" is weird and unfortunate. Perhaps we'll do a final cleanup of this API before we move its docs from the dev guide to the API guide and right some of these wrongs.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
pdurbin: Hmm. You're right. When I leave off `;type=WHATEVER` (the old way), `mediaType` is non-null and its `toString` method prints `application/octet-stream`.

landreev: So it should just be `if (mediaType != null && (!mediaType.equals("application/octet-stream")))` here. We are doing something similar throughout the "normal" datafile upload process. 

Otherwise everything looks good in the PR.

doc/sphinx-guides/source/developers/aux-file-support.rst
landreev: Whether this specific parameter is worth changing or not, I really like the fact that we have a note in the doc saying that this API is experimental, so we reserve the right to keep changing it without guaranteeing backward compatibility, etc.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
pdurbin: I didn't make any code changes but I tweaked the tests so I could make assertions on what happens if you leave off `;type=WHATEVER`: 787dbed359

I'm happy to make changes, of course. Right now we're saying the user can change the content type to whatever we want. We could change is such that if they supply `application/octet-stream` as a content type we'll run it though Tika and save whatever Tika says it is. That is, we could do something like this:

            if (mediaType != null && !mediaType.toString().equals("application/octet-stream")) {
                auxFile.setContentType(mediaType.toString());
            } else {
                Tika tika = new Tika();
                auxFile.setContentType(tika.detect(storageIO.getAuxFileAsInputStream(auxExtension)));
            }

landreev: Yes, that's what I was saying above, how that "if" line should look like. Nobody would ever want to supply "application/octet-stream" as the type. That type just means "type unknown" or "type not specified".

pdurbin: Yes, sorry, I didn't see that comment until I had written mine. I just pushed the change in 985b552a4b. Please take a look.

landreev: Yeah, I think it looks good now - but I'd like @scolapasta and @qqmyers to chime in, on whether we really want to be doing this *instead*, and not *in addition* to saving the file name, as uploaded.

src/main/java/edu/harvard/iq/dataverse/AuxiliaryFileServiceBean.java
```diff
@@ -97,8 +99,14 @@ public AuxiliaryFile processAuxiliaryFile(InputStream fileInputStream, DataFile
             storageIO.saveInputStreamAsAux(di, auxExtension);
             auxFile.setChecksum(FileUtil.checksumDigestToString(di.getMessageDigest().digest()));
 
-            Tika tika = new Tika();
-            auxFile.setContentType(tika.detect(storageIO.getAuxFileAsInputStream(auxExtension)));
+            // The null check prevents an NPE but we expect mediaType to be non-null
+            // and to default to "application/octet-stream".
```
landreev, line 103: FWIW, I believe the null check is necessary, since the mediaType can still be null, if an input stream is used for the file body.

landreev (approved):
OK, the code looks good, and everybody appears to be in agreement, that this is how we want to address *both* of the underlying issues, the mime type and the file name.
If somehow some picky user requests being able to have non-standard extensions, and/or custom download filenames for these, we'll just addres it separately.

kcondon:
@pdurbin Would you update this pr from develop? I'm getting a flyway mismatch error:
Caused by: org.flywaydb.core.api.FlywayException: Validate failed: Migration checksum mismatch for 
migration version 5.8.0.2
-> Applied to database : 1609268027
-> Resolved locally  : 343737208

djbrooke:
Hi @kcondon - I just updated from develop. 

@pdurbin apologies if I messed something up, but I also trust your git skills to revert in the morning if needed :)

PR #8296, opened 2021-12-08
bump Flyway to support PostgreSQL 14

### LINKED ISSUE #8295: support PostgreSQL 14
The integration test suite results look normal on PostgreSQL 14:
develop_PG14
however, in server.log, Flyway 5.2.4 complains that it hasn't with PostgreSQL 14.

I bumped Flyway to its current version, 8.2.0, and the PostgreSQL JDBC driver to 42.3.1 for all the bugfixes. Integration test results appear similar when the above configuration is run on PostgreSQL 13, which IQSS currently recommends:
develop_PG13

Pull request forthcoming.

donsizemore:

Flyway 5.2.4 doesn't recognize the current version of PostgreSQL, 14.

- Closes #8295

**Suggestions on how to test this**: run the integration test suite using PostgreSQL version 14


scolapasta (approved):

kcondon:
@donsizemore This breaks any installation on postgres 9.6:
Caused by: org.flywaydb.core.internal.license.FlywayEditionUpgradeRequiredException: Flyway Teams Edition or PostgreSQL upgrade required: PostgreSQL 9.6 is no longer supported by Flyway Community Edition, but still supported by Flyway Teams Edition.
        at org.flywaydb.core.internal.database.base.Database.ensureDatabaseNotOlderThanOtherwiseRecommendUpgradeToFlywayEdition(Database.java:127)

scolapasta:
We had a good discussion in slack about this, the conclusions seemed to be:
• this was a proactive switch for going to Postgres 14 (which we love), but no installation is in need of it just yet
• some installations may not want to move to 14 for various reasons, but we do want them to move past 9.6 (which is EOL in January)
• we should first get installation to move up, either via the release notes of the next version; or a dedicated intra release message stating that the next version will require a Postgres of higher than 9.6
• we can then make this change in the following release

@djbrooke I'll assign to you to decide how to handle (whether as release notes or separate message) and to move this to the appropriate column while it waits

djbrooke:
@scolapasta - ok, I think I understand. So we could say in the 5.10 release notes that Postgres 10+ is required, and then add this PR in 5.11?

donsizemore:
Once merged, this will also close #8384.

pdurbin (approved):
I haven't tested this but the changes make sense the and the release notes look good.

@kcondon please not that Jenkins uses PG 13 but @donsizemore is offering to do a run with PG 14 if you'd like. The guides say, "The application has been tested with PostgreSQL versions up to 13 and version 10+ is required." We can certainly update that line to 14, if desired.

The big thing, of course, as we said in the 5.10 release notes, is that after this is merged, PG 9.6 will not be supported. I'm personally on PG 13 but I'll probably drop a note to the dev list after this is merged that they'll need to get off PG 9.6 if they haven't already.

kcondon:
@pdurbin Is there a good time in the sprint to do this? As in at the end after everybody has completed their work before they need to alter their dev env?

pdurbin:
> Is there a good time in the sprint to do this? As in at the end after everybody has completed their work before they need to alter their dev env?

Not a ton of chatter about this in Slack: https://iqss.slack.com/archives/C010LA04BCG/p1652455817558209

In short, we think most developers are probably off postgres 9.6 by now but I did just email the dev list that now it the time to upgrade if they haven't already: https://groups.google.com/g/dataverse-dev/c/NOy16jExwyM/m/3xSiJmBwAwAJ

Again, in the 5.10 release notes we notified installations about the need to upgrade.

kcondon:
I spoke to members of local team and all were good.

PR #8304, opened 2021-12-09
adding API Token to Aux Files docs

### LINKED ISSUE #8303: Aux Files Doc Updates
We need to add 

> $API_TOKEN 

in a few places in https://guides.dataverse.org/en/5.9/developers/aux-file-support.html

PR to follow.

djbrooke:

Adds API tokens to the docs for commands where they are needed and where they are optional.

Closes #8303 

Please check my formatting closely. Feel free to edit directly if there's anything to correct.

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/developers/aux-file-support.rst
+++ b/doc/sphinx-guides/source/developers/aux-file-support.rst
@@ -35,7 +35,7 @@ formatTag and formatVersion (if applicable) associated with the auxiliary file:
   export FORMAT_TAG='dpJson'
   export FORMAT_VERSION='v1'
   
-  curl "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
+  curl -H X-Dataverse-key:$API_TOKEN "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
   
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/developers/aux-file-support.rst
```diff
@@ -35,7 +35,7 @@ formatTag and formatVersion (if applicable) associated with the auxiliary file:
   export FORMAT_TAG='dpJson'
   export FORMAT_VERSION='v1'
   
-  curl "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
+  curl -H X-Dataverse-key:$API_TOKEN "$SERVER_URL/api/access/datafile/$FILE_ID/auxiliary/$FORMAT_TAG/$FORMAT_VERSION"
```
pdurbin, line 38: It's completely fine to include the API token here in the "download" command but it's not always necessary so we should probably note this in the text above. A suggestion:

To download an auxiliary file, use the primary key of the datafile, and the formatTag and formatVersion (if applicable) associated with the auxiliary file. An API token is shown in the example below but it is not necessary if the auxiliary file was uploaded with `isPublic=true` and the dataset has been published..

pdurbin (commented):
Overall, looks good and adds value but please see my comment about clarifying when an API token is not necessary for aux file downloads.

PR #8308, opened 2021-12-13
8191 require toa or req access

### LINKED ISSUE #8191: Require selection of enable access request or adding Terms of Access for restricted files

When a file is restricted, a popup appears that allows the user to add terms for accessing the restricted files or allows enabling an access request. The popup can currently be closed without making any changes. After some discussion in the multiple license PR (https://github.com/IQSS/dataverse/pull/7920#issuecomment-947688589), I'd like to propose that either terms of access or request access must be added. I've added the reasoning below. It makes sense to me but I'd be interested to hear from other installations.

- If request access is enabled for a file, it's OK if no terms are defined because the thinking is that the person will request access and it will allow the requestor and the depositor/curator to start a conversation (since login is required for requesting access, we can be sure we will always have the requestor's contact information). A conversation can take place outside of the Dataverse installation about the access request if needed. Another option is that the depositor/curator can just make a decision about giving access with no conversation. This provides the flexibility for both.
- If request access is not enabled for a file, the terms are required because they provide the context about why request is not enabled. For the Murray Dataverse collection, the use case is that there's a separate process outside of the Dataverse installation. Same for restricted files in IFPRI collection. Generally, we don't want to just provide a wall for access with no information about why the access is not possible - the terms allow us to provide context or provide the option(s) for accessing the data. 

All of this being said, we need to do a better job of surfacing the terms during the data access process. Issue to follow on the presentation of terms to accessors, but I wanted to get this issue in as it's pretty neatly scoped (hopefully ;)).

Curator, Depositor, Superuser

Discussion in the multiple license PR, linked above. This is related but not within the scope of the license work.

See above, but the general idea is that our current behavior provides the opportunity to upload data with no real way to get at it, and we're all about data sharing! :) 

See above, some extra validation!

#7691 is related. It covers this and also dips in the license requirement.

sekmiller:
For datasets that have restricted files we want to require that Request Access is enabled or Terms of Access are provided. For newly created datasets we are setting Request Access to true. For any existing datasets with restricted files we are putting up a banner with the new requirements and a link to the edit terms page. In all cases where there are restricted files we are disabling the applicable save button until the requirement is met. Also for those existing datasets in draft status with restricted files and out of compliance terms we are disabling the Publish/Submit button until the dataset is in compliance.

Closes #8191 Require Selection of enable access request or adding Terms of Access for restricted files.

**Special notes for your reviewer**: There was an existing validator on Terms of Use and Access that really wasn't doing anything that I enabled for this. I also tried to cache the tests for has restricted files and has valid terms so that these aren't recalculated too often. Some changes had to be made to various apis and constructors because the terms of use and access validator relies on knowing its dataset version (for getting "has restricted files") and to implement the new default value (true) of RequestAccess 

**Suggestions on how to test this**: Since any new datasets created in this branch  will be in compliance you may have to search for old datasets with restricted files that don't have Request Access or TOA. Here's a query I used to find prior datasets:

select max(v.id), dvo.id from datasetversion v
join termsofuseandaccess ua on ua.id = v.termsofuseandaccess_id
 join filemetadata fm on v.id = fm.datasetversion_id
 join datafile f on f.id = fm.datafile_id
 join dvobject dvo on v.dataset_id = dvo.id
where ua.fileaccessrequest = false and ua.termsofaccess isnull
and f.restricted = true
group by v.dataset_id, dvo.id

You could also just use an update query on a compliant dataset to set termsofuseandaccess.requestaccess to false and 
termsofuseandaccess.termsofaccess to blank or null.

There were also some resfactoring changes to simplify code related to popups - so it would be good to test delete and other functionality from the edit file page.

We have met with the design team about the UI changes.

Screen Shot 2021-12-13 at 1 53 28 PM
Screen Shot 2021-12-13 at 10 07 19 AM

**Is there a release notes update needed for this change?**: We need to highlight that the new default is for datasets to allow access requests and what will happen to existing noncompliant datasets. I will write a first draft.

**Additional documentation**: For the Harvard Dataverse we'd like to get in touch with the owners of noncompliant datasets with an email that includes a link to editing the Terms. I will work on that. (starting with that query above)


djbrooke:
Thanks @sekmiller! Is this ready for review?

sekmiller:
This query will create a list of email addresses and links to datasets for all users with an edit role on a dataset that has restricted files but does not allow access requests and does not have an entry for Terms of Access
[BulkEmailQuery.txt](https://github.com/IQSS/dataverse/files/7714096/BulkEmailQuery.txt)

djbrooke:
@sekmiller thanks for inviting me to the party. I've updated the release notes and I added the query to the scripts folder - let me know if you need anything else here!

src/main/java/edu/harvard/iq/dataverse/TermsOfUseAndAccessValidator.java
```diff
@@ -12,21 +12,43 @@
...
+         */
+        boolean valid = value.isFileAccessRequest() == true || (value.getTermsOfAccess() != null && !value.getTermsOfAccess().isEmpty());
+        if (!valid) {
+            try {
+                if (context != null) {
+                    context.buildConstraintViolationWithTemplate("If Request Access is false then Terms of Access must be provided.").addConstraintViolation();
+                }
+
+                String message = "Constraint violation found in Terms of Use and Access. "
+                        + " If Request Access to restricted files is set to false then Terms of Access must be provided.";
```
pdurbin, line 45: This English should probably go in Bundle.properties so it can be translated.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -392,6 +392,43 @@ public void setRsyncScript(String rsyncScript) {
...
+    public Boolean isHasValidTermsOfAccess() {
+        //cache in page to limit processing
+        if (hasValidTermsOfAccess != null){
+            return hasValidTermsOfAccess;
+        } else {
+            if (!isHasRestrictedFiles()){
+               hasValidTermsOfAccess = true;
+               return hasValidTermsOfAccess;
+            } else {
+                hasValidTermsOfAccess = TermsOfUseAndAccessValidator.isTOUAValid(workingVersion.getTermsOfUseAndAccess(), null);
```
pdurbin, line 407: It seems like the new validator is only called from the GUI? Are the same rules enforced via API?

pdurbin (commented):
I didn't run the code but took a quick look and left some comments.

scripts/issues/8191/datasets_without_toa_or_request_access
```diff
@@ -0,0 +1,28 @@
+-- this query will identify datasets where at least one file does not have either terms of access or request access enabled, and will include owner information for those datasets
```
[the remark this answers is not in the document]
djbrooke: Thx @pdurbin - I just changed this to .txt to match what we did in 

https://github.com/IQSS/dataverse/tree/develop/scripts/issues/7398
https://github.com/IQSS/dataverse/tree/develop/scripts/issues/7687

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: As I mentioned before we all left for break, the validator is called by the commands via the validateOrDie method so it will be applied to all dataset versions created or updated using the various apis.

src/main/java/edu/harvard/iq/dataverse/TermsOfUseAndAccessValidator.java
sekmiller: Ok. these have been moved to the Bundle. Possibly moot because the save button is disabled until TOA are in compliance.

scolapasta (commented):
overall looks good. I did see some places where there are primefaces calls in the backing bean and generally it would be better to have that kind of logic in the xhrml (driven by variables that are set in the bean). @sekmiller did you see if there was a way to do this? I'm, ok with it if you think this is the best way.

sekmiller:
> overall looks good. I did see some places where there are primefaces calls in the backing bean and generally it would be better to have that kind of logic in the xhrml (driven by variables that are set in the bean). @sekmiller did you see if there was a way to do this? I'm, ok with it if you think this is the best way.

- I did move the logic to the page for the Dataset Page. I wasn't able to do it in the Edit Files Page since it's essentially a link so there isn't a way to capture an onclick event.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -3580,6 +3624,8 @@ public String save() {
 
         if (editMode != null) {
             if (editMode.equals(EditMode.CREATE)) {
+                
+                System.out.print("adding files...");
```
scolapasta, line 3628: no System.outs!

src/main/java/edu/harvard/iq/dataverse/TermsOfUseAndAccess.java
```diff
@@ -287,11 +289,25 @@ private void clearCustomTermsVariables(){
...
+    public void setValidationMessage(String validationMessage) {
+        this.validationMessage = validationMessage;
+    }
+    
     /**
      * @todo What does the GUI use for a default license? What does the "native"
      * API use? See also https://github.com/IQSS/dataverse/issues/1385
      */
-
+    /*
```
scolapasta, line 307: does this belong here?

src/main/webapp/editdatafiles.xhtml
```diff
@@ -65,7 +65,17 @@
                                 <ui:param name="lockedFromEdits" value="#{EditDatafilesPage.lockedFromEdits}"/>                             
                             </ui:include>
                         </div>
-                    </ui:fragment>  
+                    </ui:fragment> 
+                    <p:blockUI block="datasetForm" widgetVar="blockDatasetForm"/>
+                    <ui:include src="file-edit-popup-fragment.xhtml">
```
scolapasta, line 70: TO be clear, are you adding the file-edit-popup-fragment to this page? I ask because we hadn't yet consolidated all the different kebab options for all the pages and my concern is that just adding it here might consolidate some logic but it might also break other things (if there are duplicates). How well did you test this? Not just the functions you were working on, but all of the ones that have a popup extracted to this fragment?

scolapasta (changes requested):
One small fir sure change (remove a System.out) and some questions on other things.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: oops. I put those in for debugs and i usually search for them before checking in.

src/main/java/edu/harvard/iq/dataverse/TermsOfUseAndAccess.java
sekmiller: Remnant of the multiple license merge conflict. removing.

scolapasta (approved):

kcondon:
Issues so far:
Using UI:
1. a restricted file in an unpublished draft dataset with no toa or request access does not prevent publish, rather request access is automatically enabled with no notice: "for those existing datasets in draft status with restricted files and out of compliance terms we are disabling the Publish/Submit button until the dataset is in compliance"

kcondon:
1. I think the User guide link is not longer accurate: says, https://guides.dataverse.org/en/5.9/user/dataset-management.html#restricted-files-terms-of-use but that anchor doesn't exist on that page, perhaps use was renamed to access? (https://guides.dataverse.org/en/5.9/user/dataset-management.html#restricted-files-terms-of-access)

mreekie:
There is a meeting about this tomorrow.
Most of the work is complete.
Next is fine-tuning some of the loose ends and what to do about datasets that are in compliance with new requiremnt re: restricted files.  

There are around 3K restricted data files.

sekmiller:
thanks to Kevin for running the query to see which email addresses/users are associated with what number of non-compliant files. There are something like 470 users involved. The vast majority are in the low single digits when it comes to number of datasets. (more than 1/2 have only one, and nearly 400 are at 1, 2, or 3). At the upper end there are 5 with 100 or more datasets with the big winner being dataverseadmin@iq.harvard.edu with nearly 1000 datasets.
Full results attached:

[toa_query.txt](https://github.com/IQSS/dataverse/files/8484506/toa_query.txt)

pdurbin:
It looks like the latest Jenkins run failed but the one before that (83) is showing a couple SWORD tests failing:

- edu.harvard.iq.dataverse.api.SwordIT.testXmlExampleInGuides
- edu.harvard.iq.dataverse.api.SwordIT.testLicenses

Details at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-8308/83/testReport/

Anything to worry about? Fixed already?

sekmiller:
Fixed the test Phil was alluding to above. Terms in incoming sword datasets weren't being associated with their dataset versions and that was breaking the version validation.

PR #8310, opened 2021-12-14
Postgres 9.x EOL messaging, release notes updates for Postgres 10

### LINKED ISSUE #8309: Postgres 9 EOL doc updates, Release note about Postgres versions
We still have 9.x noted in the docs in a few places, and we should update these:

https://guides.dataverse.org/en/latest/developers/dev-environment.html?highlight=postgres#install-postgresql

Also this issue will include a release notes update to be included in 5.10 to give a heads up about the need for 10+ in the following release, per this note:

https://github.com/IQSS/dataverse/pull/8296#issuecomment-992935527

djbrooke:

This is an attempt at supporting the plan laid out by @scolapasta in https://github.com/IQSS/dataverse/pull/8296#issuecomment-992935527 

Closes #8309 

N/A

N/A

(edit from L.A.: As somebody who does perform upgrades, please see if the release note makes sense/answers questions you might have yourself)

N/A

N/A

**Additional documentation**:

N/A


djbrooke:
@scolapasta I'll tag you with this since you had laid out the plan. Let me know if you want to discuss, or feel free to edit directly. Thanks!

pdurbin (commented):
Overall, this looks good but we might want to update `scripts/installer/README_python.txt` as well since it references 9.6 as `/usr/pgsql-9.6` and `yum install postgresql96-devel`. Please below for some context. I could just change it but @landreev might want to take a look since he wrote that doc.

```
make sure the version that you will be using with Dataverse is the
first on your PATH. For example,

   PATH=/usr/pgsql-9.6/bin:$PATH; export PATH

Certain libraries and source include files, both for PostgresQL and
Python, are also needed to compile the module. On
RedHat/CentOS/etc. you may need to install the -devel packages, *for
the specific versions* of PostgreSQL and Python you will be using. For
example:

	yum install postgresql96-devel
	yum install python37-devel
	etc. 
```

pdurbin:
I upgraded postgres on my Mac from 9.6 to 13.5. It turns out that the directory is now "13" rather than "13.5" so I adjusted the docs to reflect this.

I also noticed that "md5" is no longer in pb_hba.conf. It's now "scram-sha-256". So I updated that too.

I think the only thing left is the readme in `scripts/installer` I mentioned above.

I may have heard that 9.6 is not technically EOL? I guess we could reword that a bit, if necessary.

Finally, I'll just mention that I don't upgrade postgres very often so I had to google how to stop the old instance. Otherwise the new instance would have used a non-standard port.

To stop postgres, this worked fine:

    sudo -u postgres /Library/PostgreSQL/9.6/bin/pg_ctl stop -D /Library/PostgreSQL/9.6/data

I also ran this to (hopefully) have my old postgres not start on boot (haven't tested it yet):

    sudo launchctl unload -w /Library/LaunchDaemons/com.edb.launchd.postgresql-9.6.plist

I got these from https://gist.github.com/sandrods/9c4873a18bbcd9051b56e8dde7ebd946

We may want to add these on the Tips page in the dev guide some day.

scolapasta (commented):
Reviewed and looks fine by me. I know @landreev was planning on adding a few words on our own upgrade experience, so I'll assign to him, but can otherwise me moved through.

landreev:
Edited the release note, added upgrade instructions and command line examples. Going to move the PR back into the review column for somebody else to take a quick look. (Or can move into QA myself if everybody's ok with it)

doc/release-notes/8309-postgres-required.md
```diff
@@ -0,0 +1,37 @@
+## Notes for Dataverse Installation Administrators
+
+### PostgreSQL Version 10+ Required
+
+If you are still using PostgreSQL 9.X, now is the time to upgrade. PostgreSQL is now EOL (no longer supported, as of January 2022), and the PostgreSQL driver in the next release of the Dataverse Software will no longer work with versions prior to 10.
```
donsizemore, line 5: The PostgreSQL driver, or the FlyWay thing-a-ma-bob? (or both?)

landreev: Did I get this wrong? I thought it was on account of Flyway, but the actual requirement was the driver... no? 
Will fix!

donsizemore: I think the Postgres driver is intentionally as backwards-compatible as possible; I understood that Flyway dropped 9.n when they added support for 14. A very minor point but... it might come up?

landreev: OK, corrected - thank you.

pdurbin:
I see this is back in review. I going to go ahead and tweak `README_python.txt` as I suggested in my initial review: https://github.com/IQSS/dataverse/pull/8310#pullrequestreview-832991472

pdurbin (approved):
Looks good. Heads up that in 2ca786e3e9 I updated the README_python.txt file for the installer, from 9.6 to 13.

PR #8313, opened 2021-12-16
5733 s3 cred chain

### LINKED ISSUE #5733: Refactor S3 credential provisioning
In #5478 the configuration of Swift storage is pretty much streamlined to contain endpoint(s), access credentials and others in JVM options. This is very usefull when configuring for usage on K8s and similar.

Currently, a S3 storage endpoint can be configured via JVM options. For credentials, you need to create `~/.aws/credentials`. (see http://guides.dataverse.org/en/4.12/installation/config.html#manually-set-up-credentials-file)

The code in `S3AccessIO.java` should be refactored to use a credentials provider chain to read from AWS profile files or retrieve secrets via MicroProfile Config API.

poikilotherm:

Containerized and test setups benefit from an ability to provide S3 keys via MPCONFIG instead of having to create AWS CLI files.

Closes #5733

This approach added a cred provider chain to retain backward and future compatibility. Docs state very precise how to deal with these secrets. Also reformated the s3 option table as it is becomes harder and harder to read anytime an option is added.

Setup a Dataverse instance plus S3 storage. Do not configure AWS CLI stuff, but add the secrets via the documented settings. All other options regarding S3 need to be done as usual, no changes. You might want to mix'n'match with profile files, though to see that they'll win.

[From Leonid: Kevin, for the purposes of QA on our end, my take is that it may be enough to a) Review the S3 documentation in the guide, from the point of view of an installation admin and b) Confirm that the traditional AWS auth. (via ~/.aws/*) is working uninterrupted]

Nope.

Maybe. @djbrooke please ring a bell if you think this should be advertised.

**Additional documentation**:
:battery: included.

### DIFF
--- a/doc/release-notes/5733-s3-creds-chain.md
+++ b/doc/release-notes/5733-s3-creds-chain.md
@@ -1,0 +1,7 @@
+# Providing S3 Storage Credentials via MicroProfile Config
+
+With this release, you may use two new options to pass an access key identifier and a secret access key for S3-based
+storage definitions without creating the files used by the AWS CLI tools (`~/.aws/config` & `~/.aws/credentials`).
+
+This has been added to ease setups using containers (Docker, Podman, Kubernetes, OpenShift) or testing and developing
+installations. Find added [documentation and a word of warning in the installation guide](https://guides.dataverse.org/en/latest/installation/config.html#s3-mpconfig).
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -550,14 +550,23 @@ List of S3 Storage Options
     dataverse.files.<id>.custom-endpoint-url     <?>                 Use custom S3 endpoint. Needs URL either with or without protocol.          (none)
     dataverse.files.<id>.custom-endpoint-region  <?>                 Only used when using custom endpoint. Optional.                             ``dataverse``
     dataverse.files.<id>.profile                 <?>                 Allows the use of AWS profiles for storage spanning multiple AWS accounts.  (none)
-    dataverse.files.<id>.access-key              <?>                 :ref:`Provide static access key ID. Read before use! <s3-mpconfig>`         ``""``
-    dataverse.files.<id>.secret-key              <?>                 :ref:`Provide static secret access key. Read before use! <s3-mpconfig>`     ``""``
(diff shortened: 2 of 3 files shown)


qqmyers (approved):
This looks OK. Two points though:

- The new key options are in a table labelled JVM Options but the guidance is to not use them as jvm options in production. For people who don't read closely, I wonder if breaking them out would be better (or just don't add them in the table and only show them in the section below it?). 
- It seems slightly odd to add an empty credential to the chain when there is no microprofile version configured. As noted, that empty cred is only a backup and won't be used otherwise, so there's no real effect (do we get a different/confusing debug message when you don't have an .aws dir in place?)

I don't think either is a reason to hold this up, but I guess I'd suggest doing something more to make it less likely people cut/paste the table list into their jvm options and do something insecure.

poikilotherm:
> This looks OK. 
:smile: 

>     * The new key options are in a table labelled JVM Options but the guidance is to not use them as jvm options in production. For people who don't read closely, I wonder if breaking them out would be better (or just don't add them in the table and only show them in the section below it?).

Breaking them out into a table below the other one right now. Push forthcoming.

>     * It seems slightly odd to add an empty credential to the chain when there is no microprofile version configured. As noted, that empty cred is only a backup and won't be used otherwise, so there's no real effect (do we get a different/confusing debug message when you don't have an .aws dir in place?)

AFAIK the SDK will silently ignore non present things. (At least I didn't see any weird messages during testing without an `~/.aws/...` profile)

> I don't think either is a reason to hold this up, but I guess I'd suggest doing something more to make it less likely people cut/paste the table list into their jvm options and do something insecure.
:+1:

djbrooke:
Thanks @poikilotherm and @qqmyers 

re: release notes I think it would be great to mention/advertise this.

poikilotherm:
> Thanks @poikilotherm and @qqmyers
> 
> re: release notes I think it would be great to mention/advertise this.

Done in c3a1e21 - @djbrooke please review wording and feel free to propose or push changes as you see fit. 🙋‍♂️

landreev (approved):
Sorry this got further delayed; in hindsight, this PR should've been approved and gone straight to QA back in December. Since everything brought up in the first review was addressed; and nothing else looks remotely controversial. 
Thank you Oliver and Jim...

PR #8322, opened 2021-12-22
IQSS/8321 Accessibility

### LINKED ISSUE #8321: Accessibility Improvements
QDR has been working to improve Accessibility in Dataverse, primarily using the Deque Axe plugin to identify issues.  The changes made significantly reduce the number of issues reported. For example, the main page on demo.dataverse.org lists 89 issues whereas the main page of a test server with the updates reports only 26 issues and all of these appear to be spurious color contrast issues (Axe flags any elements where there is a gradient background or an overlapping element, even if the displayed contrast is OK). There are similar reductions in almost all pages.

A PR to follow.

There are a wide range of accessibility issues including #7523 which is broad. Earlier accessibility issues could be reviewed to see if they are still relevant after the changes included in the PR for this issue.

In general, testing with Axe involves having the page and/or pop-ups involved displayed, so it is easy to miss issues if you don't open a menu, are in a different state (logged in/out, edit versus create versus view modes, with/without custom header/footers, etc. 

There are some issues that look to me like they are due to PrimeFaces problems that are not easy to solve without editing their classes, including:
- Interactive controls must not be nested - issue with PrimeFaces tabs
- ARIA hidden element must not contain focusable elements - PrimeFaces dismissable alert message 
- All page content should be contained by landmarks - PrimeFaces datepicker adds elements outside main content area and doesn't add a landmark role.
- ARIA hidden element must not contain focusable elements - PrimeFaces selectCheckboxMenu

qqmyers:
**What this PR does / why we need it**: As noted in the issue, this PR includes improvements to accessibility developed at/by QDR.

Closes #8321

**Special notes for your reviewer**: As noted in the issue, the fixes primarily address warnings from Deque Axe. For those, it is often not completely clear what the best fix is, especially given our use of PrimeFaces and Bootstrap 3. Rather than make many PRs, I've tried to make the individual commit messages informative enough that one could find the page involved and the warning(s) fixed in the PR and decide if the fix makes sense.

For the remaining issues, especially those related to PrimeFaces, it does look like some of them have been raised with PrimeFaces already, but it could be worth adding notes/updating existing issues or adding new issues. 

There are a few color changes at the bottom of structure.css that override the Bootstrap defaults. I kept these separate as a bootstrap update (as in #6045) would probably mean these could be dropped.

Also note that QDR's issue list at https://github.com/QualitativeDataRepository/dataverse/issues from ~80 to 104 may include additional info including screenshots (from QDR's fork) and descriptions of the problems/solutions. (Issues primarily before 80 involve QDR/s Drupal/SSO components, but were added to the dataverse repo issue list to keep them together and since there are common css/header/footer links - basically one should ignore issues that don't refer to Dataverse specifically and realize that specific color contrast issues may be only for QDR's altered styling.)

**Suggestions on how to test this**: The simplest option would be to use the Deque Axe plugin and simply verify that warnings that exist on the dev branch go away with the PR. (Note that not all warnings go away, but the PR should not create new ones.) In general, there should not be any significant changes to the UI, i.e. the fixes are intended to help screen readers, mouse-less users, etc. There are a couple - the user page did not have any main title, so I added a title element. Beyond that, I think there are some places where I added a popup title as the way to provide a label for accessibility. Overall, nothing should be obviously different or broken across any area of functionality.

Community members who have other/better tools and/or testers with accessibility training may want to look at this PR as well. (I mentioned this in the latest community meetings - have not yet gotten any feedback though.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Two minor changes noted in the testing section. Otherwise, the changes should be only related to accessibility. That does include providing screen-reader labels, etc. where the wording might be of interest - scanning the Bundle.properties file would be a good way to see new strings.

**Is there a release notes update needed for this change?**: might be nice to mention the overall accessibility improvement, but there is nothing admins have to do/no new config options,etc.

**Additional documentation**:


scolapasta (approved):
Changes here seemed straightforward and are clearly an improvement for accessibility.

kcondon:
Issues/questions:
1. Dataverse homepage title and icon is much larger and more prominent than on develop.
Screen Shot 2022-01-12 at 4 05 51 PM
Screen Shot 2022-01-12 at 4 14 26 PM

qqmyers:
For 1), I think I missed adding the css change required to reduce the size of h1 elements (the main headings were h2 but needed to be h1 for accessibility, and the css just resets the sizing back to what the h2 elements had)
For 2), the color change is intentional  - the contrast between white text and that label is not sufficient. I've reduced the brightness of the label to make the contrast acceptable. One could make other color changes to get enough contrast (shifting the hue, etc.) but in general the label has to be duller/darker than it was. It's an easy change in the structure.css file if there's a preferred color. (FWIW: The original coloring comes from bootstrap and is really their problem in v3 which I think they've fixed in later versions.)

PR #8343, opened 2022-01-12
#8193 strip commas from email response

### LINKED ISSUE #8193: Commas in email addresses break guestbook CSV files
- Create and publish a Dataverse collection and add a guestbook
- In that Dataverse collection, create and publish a dataset with at least one file and add the guestbook to that dataset
- Download one of the dataset's files and in the guestbook form's Email field, enter an email address with a comma anywhere in the string
- Go back to the Dataverse collection and download the guestbook CSV file
- Open the CSV file in any spreadsheet application and see that the comma in the email address is interpreted as a delimiter, so the string is split into one or more columns on the row, which pushes the rest of that response's answers out of the columns that they're supposed to be in

When downloading a guestbook (over UI or the newer API endpoint) that has any responses where a downloader has entered one or more commas in the email field

NA

**What happens?**
The comma in the email address that the downloader has entered into the guestbook form is interpreted as a delimiter, so the string is split into one or more columns on the row, which pushes the rest of that response's answers a column over

All users

The email addresses with commas in them wouldn't be split into two or more columns.

Maybe this could be avoided by adding quotes (“) to each field and data to avoid the content being misplaced. This was suggested in https://github.com/IQSS/dataverse/issues/4671 and it's what pgAdmin does when exporting database query results to CSV.

I'm not sure if commas should even be allowed in the Email address field, so perhaps field validation could also be added for the Email field. But the guestbooks in the Harvard Dataverse Repository already contain around 1,621 responses in 30 guestbooks (including the default guestbook) where there's one or more commas entered in the Email field, so something would have to be done with those. Adding the quotes would help spreadsheet applications interpret and display the guestbook data correctly.

Sometimes it's obvious that people entered commas in their email addresses in place of periods and other times people entered multiple email addresses in the one field. So even if field validation were added to the Email field to ensure that commas couldn't be entered, existing emails would still break the CSV.

v5.6 and v5.7

https://github.com/IQSS/dataverse/issues/4671 and https://github.com/IQSS/dataverse/issues/3449

**Screenshots:**

<kbd>Screen Shot 2021-10-27 at 6 41 18 PM</kbd>

I entered `juliangautier@g.harvard.edu,in` in my guestbook's Email field. **in** is put in the next column, pushing the rest of the answers (Institution and Position) out of the columns they're supposed to be in.

Here's the pictured guestbook CSV file:
[Testing_guestbooks_216_GuestbookReponses.csv](https://github.com/IQSS/dataverse/files/7429765/Testing_guestbooks_216_GuestbookReponses.csv)

sekmiller:
**What this PR does / why we need it**: fixes broken csv output from guestbook responses download

Closes #8193 Commas in email addresses breaks guestbook csv download

**Special notes for your reviewer**: Instead of removing commas from the email response, which had been the approach for all other responses, now all responses are enclosed in double quotes allowing the commas to remain without breaking the csv display

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
@@ -185,14 +185,14 @@ public class GuestbookResponseServiceBean {
             // string fields, or the structure of the file will be broken. -- L.A.
             
             // Guestbook name: 
-            sb.append(((String)result[1]).replace(',', ' '));
+            sb.append("\"" + (String)result[1] + "\"");
             sb.append(SEPARATOR);
 
             
             // Dataset name: 
             Integer datasetId = (Integer) result[2];
             String datasetTitle = datasetTitles.get(datasetId);
-            sb.append(datasetTitle == null ? "" : datasetTitle.replace(',', ' '));
+            sb.append(datasetTitle == null ? "" : "\"" +  datasetTitle + "\"");
             sb.append(SEPARATOR);
             
             // Dataset persistent identifier: 
@@ -211,7 +211,7 @@ public class GuestbookResponseServiceBean {
             sb.append(SEPARATOR);
 
             // file name: 
-            sb.append(((String)result[5]).replace(',', ' '));
+            sb.append("\"" + (String)result[5] + "\"" );
             sb.append(SEPARATOR);
 
             // file id (numeric):
@@ -224,19 +224,19 @@ public class GuestbookResponseServiceBean {
             sb.append(SEPARATOR);
             
             // name supplied in the guestbook response: 
-            sb.append(result[7] == null ? "" : ((String)result[7]).replace(',', ' '));
+            sb.append(result[7] == null ? "" : "\"" + (String)result[7] + "\"");
             sb.append(SEPARATOR);
             
             // email: 
-            sb.append(result[8] == null ? "" : ((String)result[8]).replace(',', ' '));
+            sb.append(result[8] == null ? "" : "\"" + (String)result[8] + "\"");           
             sb.append(SEPARATOR);
             
             // institution:
-            sb.append(result[9] == null ? "" : ((String)result[9]).replace(',', ' '));
+            sb.append(result[9] == null ? "" : "\"" + (String)result[9] + "\"");
             sb.append(SEPARATOR);
             
             // position: 
-            sb.append(result[10] == null ? "" : ((String)result[10]).replace(',', ' '));
+            sb.append(result[10] == null ? "" : "\"" + (String)result[10] + "\"");
             
             // Finally, custom questions and answers, if present:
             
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I'm going to approve this because I think it'll allow guestbooks to be downloaded.

However, in the example by @jggautier in the issue, he was entering `juliangautier@g.harvard.edu,in`

As of this code I expect it to become `juliangautier@g.harvard.eduin`

I guess this is better? In either case, the `,in` doesn't make sense to me. I don't know why it's there.

In a future pull request, perhaps we should add some input validation so that `juliangautier@g.harvard.edu,in` isn't accepted as a valid email address.

jggautier:
Thanks for checking @pdurbin.

I used `juliangautier@g.harvard.edu,in` as an example because that's similar to what someone actually put in the email field in the guestbook popup. But people have also entered multiple email addresses in one field, separated by commas, and in those cases this PR remove those commas too and the email addresses will be one long string.

The discussion about this in a sprint planning meeting made me think that we thought it was okay that people entered multiple email addresses in that email field, which would mean we should expect commas in that field. So when Danny wrote in #8193 that commas would be escaped, I thought that meant that the commas would appear in the email columns in the guestbook CSV files.

Can we agree that we're okay with people entering multiple emails in that email field and that it should expect commas? Otherwise, I agree that it would be better for the field to accept only one valid email or that the text box should accept multiple entries without them needing to be comma-separated.

sekmiller:
I re-did the coding so that all responses are double quotes which allows for commas to be displayed without breaking the csv output.

landreev (approved):
OK, this appears to work with most software packages; all the CSV parsers I tried do understand that commas inside double-quoted strings are not field separators. I even tried our own CSV ingest; not becuase anyone would want/need to ingest guestbooks as tabular data, but because we use a standard Apache CSV parser library that's used in many other opensource packages. 
(I thought when we talked about this issue during a meeting that we wanted encode these commas; as a hex notation maybe - '%2C'? - but this is even better probably... So, approving)

landreev:
(P.S. Please note the message above about a merge conflict)

sekmiller:
@landreev that merge conflict was caused by the introduction of a method for escaping csv strings for the text area guestbook response PR. I applied the method to other response strings because it makes the code cleaner. i'll put it back in review if you want to take a look.

landreev:
> @landreev that merge conflict was caused by the introduction of a method for escaping csv strings for the text area guestbook response PR. I applied the method to other response strings because it makes the code cleaner. i'll put it back in review if you want to take a look.

Sorry about my confusion during the standup. I missed the comment above. 
So this escapeCsv method does the same thing, adds double quotes... but only when necessary. OK, yes, this does seem cleaner.

landreev (approved):

landreev:
Oh yeah, we're still having an outbreak of the github automation plague (the approval didn't move the PR into QA automatically). Moving by hand...

PR #8345, opened 2022-01-12
Fix line break in config.rst

kaitlinnewson:
Fixes a line break in `config.rst` which is causing text to display incorrectly.


pdurbin (approved):
This looks much better (screenshot below) that what's at https://guides.dataverse.org/en/5.9/installation/config.html#sendnotificationondatasetcreation

Thanks, @kaitlinnewson !

Screen Shot 2022-01-12 at 4 43 09 PM

Issue #5277, opened 2018-11-05
As a curator, I want to more easily add metadata about related resources so that it's more discoverable


jggautier:
While working on exporting more dataset metadata in schema.org (https://github.com/IQSS/dataverse/issues/4371), the team talked about the best way to include the URLs of related publications. (In schema.org, it's best practice to use a URL, instead of an ID number, for related publications.) But not all depositors/curators enter related publication metadata the way we expect (the way the fields have been designed), which has and will lead to significant curation work to correct the metadata so that it's exported in different metadata formats (I'm thinking of Schema.org and DataCite).

This is what the Dataverse 4 relatedPublication compound fields look like:
![screen shot 2018-11-05 at 1 00 51 pm](https://user-images.githubusercontent.com/18374574/48016919-ea2c0780-e0fa-11e8-9e7d-7d04cceaf60d.png)

For the schema.org issue (https://github.com/IQSS/dataverse/issues/4371), we decided to use what's entered in the URL field.

But there are plenty of cases where:
- what the depositor enters in the URL field is not the URL form of the included DOI or other PID (which is what the picture above shows)
- nothing is entered in the URL field
- the URL form of a PID is included, but nothing is entered in the ID Type and/or ID Number fields

So not all possible related publication metadata will be exported without curation work to update the metadata.

For DataCite schema, Dataverse needs to know what are the ID Types and ID Numbers of related publications, other datasets and software. For Schema.org, Dataverse needs to know what's the URL of related publications. (As of this issue, there's no recommended way to include in schema.org metadata about other related datasets and software.)

For other datasets and software, Dataverse doesn't have fields that are meant to collect URLs, ID Types or ID Numbers:

Field for related software (this might change with a planned software metadata block):
![screen shot 2018-11-05 at 1 30 27 pm](https://user-images.githubusercontent.com/18374574/48018630-590b5f80-e0ff-11e8-9b07-17248c6bed9b.png)

Field for related datasets:
![screen shot 2018-11-05 at 1 30 46 pm](https://user-images.githubusercontent.com/18374574/48018642-60326d80-e0ff-11e8-846f-53846d10bb76.png)

Can we improve the metadata fields (such as the number of fields, how they're labelled and described) and/or how Dataverse uses what's entered in metadata fields (such as text parsing, getting more information from external sources) in order to reduce the curation and training needed now to make sure Dataverse is more capable of exposing information about related resources?

jggautier:
[...]

Issue #5889, opened 2019-05-28
Align or merge DataCite metadata exports


jggautier:
This issue is meant to record the differences between Dataverse's two newest metadata exports as of v4.14, **"DataCite"/"Datacite"** and **"OpenAIRE"/"oai_datacite"**, and discussion about how to align (or possibly merge) the very similar exports.

As part of v4.10 (released in Dec. 2018), Dataverse makes available through the UI, API and over OAI-PMH dataset metadata in the DataCite schema (https://github.com/IQSS/dataverse/issues/5043). This lets Dataverse export dataset metadata in a widely-used, discipline-agnostic schema that's more standardized than Schema.org and has more metadata than Dublin Core.

As part of v4.14 (released in May 2019), Dataverse makes available through the UI, API and over OAI-PMH DataCite metadata that complies with OpenAIRE requirements (https://github.com/IQSS/dataverse/issues/4257). Repositories need to follow these requirements in order for their dataset metadata to be made discoverable (harvested) by OpenAIRE ([OpenAIRE EXPLORE](https://explore.openaire.eu/search/find)). The OpenAIRE metadata requirements follow the DataCite schema, with [some differences between OpenAIRE and DataCite listed in their documentation](https://guidelines.openaire.eu/en/latest/data/use_of_datacite.html).

What both exports are called depending on the export method:

![openaire-datacite-export-graphic](https://user-images.githubusercontent.com/18374574/58607489-d1896780-826d-11e9-9b44-95bd8b890098.jpg)

Both metadata exports are based on DataCite 4 and are meant to be valid against the DataCite 4 schema (although the xml records available over OAI-PMH in "Datacite" format reference DataCite's 3.1 schema). But Dataverse exports them as separate formats for several reasons:
- The two metadata exports were worked on at different times by different groups
[...]

Issue #2778, opened 2015-11-30
Add "Relation Type" to related publication metadata fields to send DataCite related publication metadata


posixeleni:
comment from @bencomp in #2774 

> the DataCite metadata schema v3.1, if a related identifier is included (e.g. for publications that build on the dataset), a relation type must be given.

augustfly:
commenting for reasons of tracking this thread and opinion, but jumping on board with DOI/DataCite relation types is something Dataverse really should do. the controlled vocab and use cases are there for the reuse. i'm sure there are UI issues downstream from implementing it though.

jggautier:
In the article "Pre-Metadata Counseling: Putting the DataCite relationType Attribute into Action" ([PDF](https://www.ideals.illinois.edu/bitstream/handle/2142/95785/Curating%20Research%20Data_v2_Case%20Study_Pre-Metadata%20Counseling.pdf?sequence=3&isAllowed=y)), the team behind [Illinois Data Bank](https://databank.illinois.edu/) write about using DataCite's relationType terms. They decide in their longer ["overly honest" article](http://journal.code4lib.org/articles/11980) about the repository's development that the terms are too difficult to use (hard to define, too much overlap) to expect depositors to apply them consistently, so their curators apply them after dataset publication and they use only six of the 25.

This makes me think it would be a good idea to lessen the confusion when this is implemented by:
- Starting with a limited number of relationType terms available in the deposit form
- Letting depositors choose multiple relationTypes (considering the overlap in terms)

pdurbin:
> they use only six of the 25

The six seem to be "Article, Code, Dataset, Presentation, Thesis, or Other" unless I misunderstand. This issue may be a candidate for adding a user story (as discussed in retrospective this afternoon) because I'm getting a little lost on who wants what and why.

augustfly:
Those are the object types, Phil. The 6 relations are IsSupplementTo, IsCitedBy, IsNewVersionOf/IsPreviousVersionOf and the IsPartOf/HasPart pair. 

I think narrowing terms/relations is wise; it has always felt like the DataCite list was made to appear to cover edge cases, which leads to unnecessary confusion. I don't see much overlap in those 6 relations though someone who hasn't thought it through for a particular object might.  

[...]

PR #8357, opened 2022-01-20
issue #5277: first implementation step for exporting related publicat…

pkiraly:
…ion to DataCite

It was a feature request from our users, and as I found in the issue queue others also missed that related publications and also other "Related ..." metadata fields are not exported into DataCite metadata. 

It is a "baby step" towards a better solution. Its intention is to move forward the discussion and to highlight theoretical and practical topics.

1. modify an existing dataset or create a new one
2. add some values into Related Publications's ID Type and ID number field.
3. save it
4. publish it
5. click on Metadata > Export metadata > Datacite
6. check if there is a line similare to
```xml
<relatedIdentifier relatedIdentifierType="ARK" relationType="Cites">ark:222</relatedIdentifier>
```
in the `<relatedIdentifiers>` section.

It changes the XML output of the DataCite export - see previous point.

If community accepts this approach, it would worth to mention in the release not since I believe it increases the number of connections between papers and dataset in the PID network, thus somehow it might increase the discoverability.

**Additional documentation**:

---

Related to #5277

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteRegisterService.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DOIDataCiteRegisterService.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -23,6 +24,8 @@ import javax.ejb.Stateless;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
 import javax.persistence.TypedQuery;
+
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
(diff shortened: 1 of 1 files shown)


djbrooke:
Hi @pkiraly - thanks! I know that this is a long-standing issue, and I appreciate the PR to start conversations around a potential solution. I think any solution would need to require user selection of the relationship, and that would require some UI changes. I don't think we'll review and merge this in it's current form, and this is one of my last few days with the project, but:

- @jggautier I know you've thought about this more than I have, so if I'm wrong about anything above let me know :)
- @TaniaSchlatter @scolapasta I think this focus on relationships and interoperability would fit in well with the NIH work (https://datascience.nih.gov/news/nih-office-of-data-science-strategy-announces-new-initiative-to-improve-data-access) and potentially the DASH work of Harvard Data Commons. There may be an opportunity to generate some use cases around this and better inform/prioritize a solution but I'm not sure.

jggautier:
I think we'd want to move forward with this PR because it's a small step toward functionality that's been on our radar since at least 2015, but it would also be best to agree on some details about the next steps. I've always thought that QDR's and Madrono's repositories took this PR's approach because they figured that doing something to reap the benefits of sending this metadata to DataCite (such as contributing to the Make Data Count effort, whose metrics they were also early to implement in their repositories) was better than potentially waiting years for a longer lasting solution, which is what wound up happening. If this work was done as part of the NIH work or the Harvard Data Commons work, would it be another year or longer before Dataverse repositories would be able to start sending related publication metadata to DataCite?

The Make Data Count database considers several relationTypes when tracking and reporting relationships between resources, so in practice when Make Data Count says that a dataset has 3 "citation counts," that might not mean that three other resources formally cited that dataset. When looking at Make Data Count's data, I think we can reliably say only that the dataset has some sort of relationship with three other resources (and we know that we can't rely on the formal citation of datasets for measuring the influence of data). I heard in some DataCite webinar last year that researchers have been analyzing Make Data Count data collected so far, and maybe they're seeing patterns in publishers' uses of relationship types that can help DataCite and Crossref better define and maybe somehow enforce relationship type definitions.

But right now I don't think that the benefits of specifying relationship types have been realized, while I think there are already some benefits to reporting to DataCite that a dataset is somehow related to another resource. And Dataverse repositories have been collecting this information for years. Many Dataverse repositories don't know and I think would be hard pressed to figure out the types of relationships to apply to the Related Publication metadata they already have. To me that means that repositories will need to apply some type of general relationship type by default. And I think that's what this PR does.

Maybe the term IsCitedBy is too specific and shouldn't be used as a general, "somehow related to" term. When we were talking about relationTypes in https://github.com/IQSS/dataverse/issues/2778, in [a Google Doc](https://docs.google.com/document/d/1hc3Pa9fNJdEecKzxi1QkwCZLdciQjxv45eYZCvBKmeE) I proposed using IsReferencedBy, which DataCite defines as "indicates A is used as a source of information by B". But if we do use IsCitedBy to say that a dataset is somehow related, we should make sure that we all know this so that it's easier later to update the metadata with another term for "somehow related" if we want to reserve IsCitedBy for "this dataset appears in a citation in this publication". The more consistently the Dataverse community can use these terms, even if other publishing communities (like Zenodo) aren't using them the same way, the easier it'll be to apply changes if, for example, DataCite and/or CrossRef settle on fewer relationship types or revise their definitions based on how they're actually being used.

pkiraly:
@TaniaSchlatter @scolapasta Did you have a chance to take a look on this issue? For us it would be quite urgent, and seems nothing happened in the last  2 months. Do you think it is this out of scope/interest for the majority of the Dataverse community?

qqmyers:
My sense is that one reason this is stuck is the choice of isCitedBy. From the discussion in [#2778](https://github.com/IQSS/dataverse/issues/2778#issuecomment-446999587). isCitedBy is usually reserved for cases when a dataset is in the reference section of a paper. QDR decided on 'IsSupplementTo' in it's fork as an incremental step while waiting for support for a broader range of relationships to be supported (e.g. by adding a relationship type field to the related publications entry). (That said, it sounds like there is another site using isCitedBy).

While recognizing that support for multiple types is of interest, perhaps a way forward would be to select one of the more vague relationship types (IsSupplementTo, IsReferencedBy) as a default and poll the community (e.g. via email) on a few questions:

- Is adding related publications to the DataCite xml (sent to DataCite and in the DataCite export) with relationship X something we should do now?
- If no, would making the relationship type reported configurable at your site make this something we should do now? And, if yes, what relationship would you use?
- If neither of these is acceptable, it would also be possible to add an option to not report related publications (as is currently the case), but please tell us why you would not want to turn this feature on.

Hopefully that would provide enough feedback to quickly tweak this PR (e.g. by adding a setting(s) to select the type and/or turn off relationship reporting) as needed.

pdurbin:
I'm just noting that we're still discussing what to send to DataCite and how: https://groups.google.com/g/dataverse-community/c/vxAw8vs7_K8/m/DRppjQjnEAAJ

Also, @jggautier made a nice list of related issues with regard to sending more and better data to DataCite: https://docs.google.com/document/d/1z6pPVT4_fc833thD9MyhoAMy2lwKDmCFX5QtKq8I04Y

Finally, some related discussion here:

- On 2022-10-06 there was a meeting about sending more metadata to DataCite: https://docs.google.com/document/d/1tNnvVh8jYY1g53BEwpJmMmm9w6Vgy_Q7RrmFjGnYOyA/edit?usp=sharing
- On 2022-10-20 there was another meeting about sending more metadata to DataCite: https://docs.google.com/document/d/1XMDGBNofnZbYp1vkP-qp497hxUr3GXvq2XDBjZMPK14/edit?usp=sharing

scolapasta:

jggautier:
[I wrote that my proposal for merging the DataCite and OpenAIRE exports](https://github.com/IQSS/dataverse/issues/5889#issuecomment-2025510299) proposal will do what @pkiraly proposed in this PR, and I wrote that I thought this PR could be closed (or I guess deleted?) after the DataCite and OpenAIRE exports are merged.

@pkiraly if you're available, I'm hoping you can check out that proposal if you haven't already. @qqmyers said he'd work on it.

The proposal also references work that we're planning, outside of that proposal, about defining how datasets are related to other research objects.

pkiraly:
@jggautier Is it OK for you if I do it early next week?

jggautier:
Definitely, that's OK with me. Thanks!

pkiraly:
@jggautier I read your plan, and it is absolutely fine for me.

pdurbin:
It sounds like there's agreement to close this PR in favor of the proposal above. Great. Closing.

PR #8361, opened 2022-01-23
IQSS/8348 - don't remove expired token

### LINKED ISSUE #8348: API token disappears
Overview: API token disappears and replaced with message "API Token for user [user] has not been created".  API token is erased from apitoken table.

None (althouth possibly related to token expiraration), although it has occured before the token expiry date sometimes.

* When does this issue occur?
It's not clear when this occurs, as the api token exists until the moment it no longer exists, which is very Zen.

https://[installation.url]/dataverseuser.xhtml?selectTab=apiTokenTab

* What happens?
Any action requiring an API token fails as the token is missing from the apitoken table

API users. To my knowledge, though, all those affected have been administrators (at least) of a collection all the way up to superusers.

The API token should remain in effect until its expiry date

v5.6

https://github.com/IQSS/dataverse/issues/6389

From a message from James Myers 13 January 2021 on the Dataverse Community Group:

Looks like I broke this in 2019 when fixing a bug that allowed multiple tokens to exist in the db: https://github.com/IQSS/dataverse/issues/6389  - I don’t think we caught that the UI would be affected and wouldn’t display the token expired message. I think a github issue to revert that change would be appropriate.

-----
The complete text of the email posted to the Dataverse Community Group:

Hi everyone,

Today I was doing some work using the API . It was working fine then suddenly just stopped working with the message "{'status': 'ERROR', 'message': 'Bad api key '}".

I had looked at my API key a few days ago to see if it was about to expire and it wasn't. Having a look at my token via the GUI, the message said "API token has not been created"

From experience and my admittedly cursory search of github which showed a string in 
dataverse/src/main/java/propertyFiles/Bundle.properties, had my token expired I would have expected something like:

"This token is expired, please generate a new one." to be displayed in the GUI.

On a whim, I checked the apitoken table and my api token that I had been using for months wasn't in the table at all, not even as a token with an expiretime in the past.

This isn't the first time this has happened to me. It's also happened on our test server, plus the demo and production servers hosted at other institutions.  And it's also happened to [at least] one of my co-workers. Usually I just shrug it off and create a new key, but this time I thought I'd  bring it up here.

I hesitate to bring it up as a github issue, because I'm not sure if the message displayed in the GUI is just the wrong one. It does seem weird, though, that my old token isn't in the database at all.

Has this affected anyone else? It's not a huge issue, but it is odd.

qqmyers:
**What this PR does / why we need it**: Code added in #6394 automatically removed expired API Tokens which a) is potentially confusing (you thought you had a token) and b) avoids the existing logic that would show the expired token and a warning to create a new one.

Closes #8348

**Suggestions on how to test this**: Create an apitoken, backdate its expiretime in the db to be before now, and refresh the API Token tab. The expired token should display and show a warning message about needing to refresh it. Current code just shows no token exists if your old one has expired.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
@@ -380,14 +380,7 @@ public class AuthenticationServiceBean {
         }
         if (tokens.size() == 1) {
             // Normal case - one token that may or may not have expired
-            ApiToken token = tokens.get(0);
-            if (token.getExpireTime().before(latest)) {
-                // Don't return an expired token which is unusable, delete it instead
-                em.remove(token);
-                return null;
-            } else {
-                return tokens.get(0);
-            }
+            return tokens.get(0);
         } else {
             // We have more than one due to https://github.com/IQSS/dataverse/issues/6389 or
             // similar, so we should delete all but one token.


qqmyers:
FWIW: replaced by #8363 - (which removes the ' char in the branch name/PR title, which was causing a build issue)

Issue #8094, opened 2021-09-14
Migrate codebase to Java 17 LTS


poikilotherm:
Today, Java 17 has been released. This is the next LTS version of Java following 11, which will still be around for some time and then EOL.

Dataverse should run on the next generation Java, too! 🎊🥳

Things that pop up:
- Payara has had a roadmap item for JDK 17 support for Q3/2021. Let's see how fast this will become a reality.
- We might already add support for compiling and unit testing etc with Java 17 on CI to watch out for any trouble.
- Maybe it's a good idea so rewrite some DTOs to records...?

pdurbin:
Similar errors came up in [chat](https://matrix.to/#/!AmypvmJtUjBesRrnLM:matrix.org/$dsdEwLzi12LtlCCVyRDP6JnCOWfeeEzmYBW4i76hqiw) the other day on Java 16:

![Screen Shot 2021-11-04 at 3 37 44 PM](https://user-images.githubusercontent.com/21006/140408535-628ee6d7-42ae-4106-a9b0-292c3f7508e7.png)

poikilotherm:
Maybe we can add a build matrix to our Github Unit Test workflow and use a shadow job to make it non-failing?

See https://ncorti.com/blog/howto-github-actions-build-matrix

poikilotherm:
### Current status

![image](https://user-images.githubusercontent.com/5468333/151152860-05444ac6-0f06-472d-bf5e-22208fde1422.png)

### Dependency updates

This means we need to at least upgrade GSON and RestAssured deps: 

```
[ERROR] edu.harvard.iq.dataverse.export.DDIExporterTest.testExportDataset  Time elapsed: 0.163 s  <<< ERROR!
java.lang.reflect.InaccessibleObjectException: Unable to make field private final int java.time.LocalDate.year accessible: module java.base does not "opens java.time" to unnamed module @23ceabc1
	at edu.harvard.iq.dataverse.export.DDIExporterTest.testExportDataset(DDIExporterTest.java:68)
        [...]
        at com.google.gson.Gson.fromJson(Gson.java:689)

[ERROR] edu.harvard.iq.dataverse.export.OpenAireExporterTest.testExportDataset  Time elapsed: 0.532 s  <<< ERROR!
java.lang.reflect.InaccessibleObjectException: Unable to make private java.util.regex.Pattern(java.lang.String,int) accessible: module java.base does not "opens java.util.regex" to unnamed module @23ceabc1
[...]

PR #8320, opened 2021-12-21
7662 make solrconfig less static

poikilotherm:
**What this PR does / why we need it**:

Solr updates become easier when changes are trackable. Moving the config files around results in cumbersome searches for change history. Also, it's hard to track why and when certain additions happened. Declaring necessary changes to a vanilla `solrconfig.xml` makes this much easier.

This pull request will also add necessary bits to installer, docs etc to make use this. Tools like container builds and dataverse-ansible @donsizemore might reuse the XSLTs directly.

TODOs:
- [ ] Integrate distributable ZIP with installer Makefile
- [x] Make it easier to configure stuff like Solr request header size, limits, etc
- [x] ~~Add solr-maven-plugin to start a Solr server with a single Maven command (great for ephemeral dev usage)~~
- [x] Add a container profile to create images from Maven
- [ ] Add GitHub Workflows to test the module on changes
- [x] Add docs to installation guide (changes to how this works when installing Solr)
- [ ] Add docs to development guide (changes to how you can use this to simplify your dev setup)
- [ ] Add docs about the container image + tips and tricks
- [ ] Add release notes
- [ ] Speak to @donsizemore how to integrate this with `dataverse-ansible`

**Which issue(s) this PR closes**:

- Closes #7662 
- Closes #8325 
- Relates to IQSS/dataverse-security#47

**Special notes for your reviewer**:
This is a draft as long as outstanding TODOs are present.

Note the change to Sphinx guides retrieving the Solr version from the parent POM to make updates easier and avoid missing changes.

**Suggestions on how to test this**:
`mvn -f modules/solr-configset/pom.xml package` to compile the config, create the ZIP and execute integration test.
Usual integration testing should make sure the index configuration is working properly.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Nope.

**Is there a release notes update needed for this change?**:
Yes that will be necessary.

**Additional documentation**:
None so far. Will include updated docs later.

PR #8362, opened 2022-01-24
8094 Experimental CI Java 17 Maven Unittest build

poikilotherm:

This adds an experimental build on the Github Workflows for IQSS/dataverse using Java 17. It's easier to recognise what fails with Java 17 if this becomes a part of regular CI.

Relates to #8094 

Well take a look at the failing CI which should not fail the entire build (which is good)

See build log of the PR

Nope, dev only.

Nope.

**Additional documentation**:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

### DIFF
--- a/.github/workflows/maven_unit_test.yml
+++ b/.github/workflows/maven_unit_test.yml
@@ -10,15 +10,17 @@ on:
 
 jobs:
     unittest:
-        name: (JDK ${{ matrix.jdk }} / ${{ matrix.os }}) Unit Tests
+        name: (${{ matrix.status}} / JDK ${{ matrix.jdk }}) Unit Tests
         strategy:
             fail-fast: false
             matrix:
                 jdk: [ '11' ]
                 experimental: [false]
+                status:  ["Stable"]
                 include:
                     - jdk: '17'
                       experimental: true
+                      status: "Experimental"
         continue-on-error: ${{ matrix.experimental }}
         runs-on: ubuntu-latest
         steps:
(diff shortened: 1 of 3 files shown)


qqmyers (approved):
Makes sense to start testing. Others are probably better able to comment on the implementation - from my view, if it works and isn't required, it's good.

scolapasta (approved):
I like the idea of starting to test on Java 17 (and using this framework for other future versions too).

kcondon:
@poikilotherm Repeat from yesterday: please refresh from develop? Thanks!

pdurbin:
Once this is merged does it mean we're going to see "Some checks were not successful" on all pull requests until we fix the code to support Java 17? I like the comforting feeling I get from seeing "All checks have passed". With some failures you have to get used to knowing which failures are expected. It would be nice if we could somehow have "All IMPORTANT checks have passed".

poikilotherm:
Please see this very PR for an example how this will look like.

There is no way to tell GitHub to just print a warning, as GitHub only supports two states: ✅ yay / ❌ nay

This was the exact reason why I added the statement of "Experimental" and "Production" to the action titles.

poikilotherm:
@kcondon rebased on current `develop`.

poikilotherm:
Looks like we are not alone with this... https://github.com/actions/toolkit/issues/399

I don't currently work on JDK 17. Modules #8395 first, so I can push #8320 forward with them (already doing this local).

PR #8369, opened 2022-01-26
IQSS/8367- create apitoken when needed for preview, refactor, add embargo logic

### LINKED ISSUE #8367: In-page preview for draft files shows an error when missing user API token

1. Login to a Dataverse install with previewers installed
2. Verify your account doesn't have a generated API token, or revoke it if it exists
3. Create a new dataset with a file that can be previewed, like an image or PDF
4. After the file is uploaded, visit the file page
5. Note the preview panel error that appears (screenshot below). If you generate an API token, it will load correctly.

* When does this issue occur?

When on a file page for a draft file with an account that does not have an API token generated.

The File view page (file.xhtml).

* What happens?

An error is displayed in the preview panel ("Unable to preview metadata....").

Dataset creators, or anyone that can view a draft file.

I expected to see a preview of the file.

5.8, tested in the Scholars Portal demo environment and Harvard demo.

May be resolved in #8265 - please close if that is the case.

**Screenshots:**

Screen Shot 2022-01-25 at 3 19 41 PM

qqmyers:

When a user tries to preview a protected file (in draft dataset, restricted, or embargoed) and does not have a valid apiToken, the preview would fail to display. We already had logic to create a new token for explore tools when needed (token doesn't exist or expired) - this PR refactors to have the preview pane in the datafile page use the same logic. 

Also - with #8363 merged, the logic needs to be updated to explicitly check for an expired token (previously an expired token would be deleted so the no token and expired token cases were the same). Lastly, in looking through the code, it appears that this code was not updated when embargoes were added so the code would fail to allow previews for an actively embargoed, non-restricted file previewed in a published version. (The problem is that a user who should be able to preview the embargoed file couldn't/it did not expose an embargoed file when it shouldn't.)

Closes #8367

**Suggestions on how to test this**: Create files where an API token is needed to access them (only in draft version, restricted, or embargoed) and try to preview with no api token or an expired token - the preview should now work (and wouldn't with the prior code)

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
```diff
@@ -340,6 +328,24 @@ public void explore(GuestbookResponse guestbookResponse, FileMetadata fmd, Exter
         }
     }
 
+    public ApiToken getApiToken(User user, boolean createIfNeeded) {
```
sekmiller, line 331: Is createIfNeeded ever false? I don't see any usages where it's false

sekmiller (commented):
Looks good in general. I just have one question on the getApiToken method and the branch needs to be updated from develop.

src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
qqmyers: Not now. I thought it might be useful somewhere/someday to be able to call this to see if a token exists without triggering creation, but I can remove this if that's over-designing.

qqmyers: I just went ahead and removed the second param - easy enough for someone to add in the future.

sekmiller (approved):
Looks good. Thanks

PR #8377, opened 2022-01-28
Getting rid of log4j1 in the war file

landreev:

Self-explanatory - this PR gets rid of the log4j-1.2.14 jar we've been shipping as part of the war file.

Closes IQSS/dataverse-security#48

The surprising part here is that it really looks like none of our 3rd party dependencies are actually using log4j; with the single exception of XOAI - the unsupported dependency that we built ourselves. It's long overdue for an overhaul; this will be handled as a separate issue. 

The main (and easily verifiable) goal is for the .war file to no longer contain log4j-1.2.14.jar. Jenkins/restassured tests still passing should ensure that nothing has been broken in the process. 

**Additional documentation**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,8 @@
         <payara.version>5.2021.5</payara.version>
         <postgresql.version>42.2.19</postgresql.version>
         <aws.version>1.11.762</aws.version>
-        <slf4j.version>1.7.32</slf4j.version>
+        <slf4j.version>1.7.35</slf4j.version>
+	<reload4j.version>1.2.18.4</reload4j.version>
         <commons.io.version>2.11.0</commons.io.version>
         <commons.logging.version>1.2</commons.logging.version>
         <commons.lang3.version>3.12.0</commons.lang3.version>
@@ -225,14 +226,23 @@
             </dependency>
             <!-- In *theory* the log4j12 bridge should not be made convergent, as WE decide which way logs flow.
                  Libraries should *only* rely on the api package. But sometimes... :-/ -->
+            <!-- Replacing the dependency slf4j-log4j12 with slf4j-reload4j, -->
+            <!-- in order to make sure we are getting rid of log4j for good... -->
+            <!-- it may not be necessary, to include slf4j-reload4j explicitly -->
+            <!-- but shouldn't hurt either (? L.A. - Jan. 2022) --> 
             <dependency>
                 <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
+                <artifactId>slf4j-reload4j</artifactId>
                 <version>${slf4j.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>jcl-over-slf4j</artifactId>
+                <version>${slf4j.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>jul-to-slf4j</artifactId>
                 <version>${slf4j.version}</version>
             </dependency>
             
@@ -252,7 +262,25 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
         </dependency>
-        
+	<dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+	<dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+        </dependency>
+	<dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+        </dependency>
+	<!-- reload4j is a drop-in security patch/replacement for log4j1. -->
+	<!-- (https://reload4j.qos.ch/) -->
+	<dependency>
+            <groupId>ch.qos.reload4j</groupId>
+            <artifactId>reload4j</artifactId>
+            <version>${reload4j.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.passay</groupId>
             <artifactId>passay</artifactId>
@@ -573,6 +601,10 @@
         <!-- is still buggy. As an experiment, I'm using -->
         <!-- a patched version I built locally. -->
         <!-- (pull requests pending - L.A. -->
+	<!-- These locally-built XOAI libraries are still a serious debt -->
+	<!-- that needs to be addressed. Adding explicit "exclusion" entries -->
+	<!-- below for the immediate purpose of getting rid of the log4j library -->
+	<!-- that was being added to the project via these dependencies. (L.A. - Jan. 2022) -->
         <dependency>
             <groupId>com.lyncode</groupId>
             <artifactId>xoai-common</artifactId>
@@ -582,11 +614,23 @@
             <groupId>com.lyncode</groupId>
             <artifactId>xoai-data-provider</artifactId>
             <version>4.1.0-header-patch</version>
+	    <exclusions>
+                <exclusion>
+                     <groupId>log4j</groupId>
+                   <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.lyncode</groupId>
             <artifactId>xoai-service-provider</artifactId>
             <version>4.1.0-header-patch</version>
+	    <exclusions>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <!-- Added for AutoService -->
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
         <postgresql.version>42.2.19</postgresql.version>
         <aws.version>1.11.762</aws.version>
         <slf4j.version>1.7.35</slf4j.version>
-	<reload4j.version>1.2.18.4</reload4j.version>
+        <reload4j.version>1.2.18.4</reload4j.version>
         <commons.io.version>2.11.0</commons.io.version>
         <commons.logging.version>1.2</commons.logging.version>
         <commons.lang3.version>3.12.0</commons.lang3.version>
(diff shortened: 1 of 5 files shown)


landreev:
(part of the reason for making this a draft PR, I wanted to see the automated tests run on it)

pom.xml
```diff
@@ -252,7 +262,25 @@
...
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+	<dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+        </dependency>
+	<dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+        </dependency>
```
poikilotherm, line 276: **question:** why do you include these three SLF4J libs as direct dependencies? 

`org.slf4j.slf4j-jdk14` has been included with a comment to have the lib on the classpath and serve as the logging bridge between SLF4J API and our common JUL logging. The others should IMHO not be necessary, as they are included as transitive dependencies by Tika and Solr4J.

pom.xml
```diff
@@ -225,16 +226,25 @@
...
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>jcl-over-slf4j</artifactId>
                 <version>${slf4j.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>jul-to-slf4j</artifactId>
+                <version>${slf4j.version}</version>
+            </dependency>
```
poikilotherm, line 247: **praise:** good catch!

pom.xml
```diff
@@ -225,16 +226,25 @@
...
                  Libraries should *only* rely on the api package. But sometimes... :-/ -->
+            <!-- Replacing the dependency slf4j-log4j12 with slf4j-reload4j, -->
+            <!-- in order to make sure we are getting rid of log4j for good... -->
+            <!-- it may not be necessary, to include slf4j-reload4j explicitly -->
+            <!-- but shouldn't hurt either (? L.A. - Jan. 2022) --> 
             <dependency>
                 <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
+                <artifactId>slf4j-reload4j</artifactId>
                 <version>${slf4j.version}</version>
```
poikilotherm, line 236: **suggestion:** It looks like we don't need this anymore - none of our deps is including this as a transitive dependency anymore.

pom.xml
```diff
@@ -252,7 +262,25 @@
...
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+        </dependency>
+	<!-- reload4j is a drop-in security patch/replacement for log4j1. -->
+	<!-- (https://reload4j.qos.ch/) -->
+	<dependency>
+            <groupId>ch.qos.reload4j</groupId>
+            <artifactId>reload4j</artifactId>
+            <version>${reload4j.version}</version>
+        </dependency>
```
poikilotherm, line 283: **suggestion:** as the XOAI lib is the only one using this library, maybe we can move this next to them and mention within the comment to remove this direct dependency once the technical debt has been payed back? I'd also include a hint that we are not using this within our code, it's just to have it on the classpath... (like I did for the `org.slf4j.slf4j-jdk14` package.)

pom.xml
```diff
@@ -573,6 +601,10 @@
         <!-- is still buggy. As an experiment, I'm using -->
         <!-- a patched version I built locally. -->
         <!-- (pull requests pending - L.A. -->
+	<!-- These locally-built XOAI libraries are still a serious debt -->
+	<!-- that needs to be addressed. Adding explicit "exclusion" entries -->
+	<!-- below for the immediate purpose of getting rid of the log4j library -->
+	<!-- that was being added to the project via these dependencies. (L.A. - Jan. 2022) -->
```
poikilotherm, line 607: **nitpicking:** indentation gone rogue :see_no_evil:

pom.xml
```diff
@@ -582,11 +614,23 @@
             <groupId>com.lyncode</groupId>
             <artifactId>xoai-data-provider</artifactId>
             <version>4.1.0-header-patch</version>
+	    <exclusions>
+                <exclusion>
+                     <groupId>log4j</groupId>
+                   <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
```
poikilotherm, line 622: **nitpicking:** indentation gone rogue :see_no_evil:

pom.xml
```diff
@@ -582,11 +614,23 @@
...
         <dependency>
             <groupId>com.lyncode</groupId>
             <artifactId>xoai-service-provider</artifactId>
             <version>4.1.0-header-patch</version>
+	    <exclusions>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
```
poikilotherm, line 633: **nitpicking:** indentation gone rogue :see_no_evil:

src/main/java/edu/harvard/iq/dataverse/EjbDataverseEngine.java
```diff
@@ -51,7 +51,7 @@
 import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
 
-import org.apache.log4j.lf5.LogLevel;
+//import org.apache.log4j.lf5.LogLevel;
```
poikilotherm, line 54: **praise:**: good catch! Voting to remove the import completely.

src/main/java/edu/harvard/iq/dataverse/util/bagit/OREMap.java
```diff
@@ -35,7 +35,7 @@
 import javax.json.JsonValue;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
-import org.apache.log4j.lf5.LogLevel;
+//import org.apache.log4j.lf5.LogLevel;
```
poikilotherm, line 38: **praise:**: good catch! Voting to remove the import completely.

poikilotherm (dismissed):
I left a few comments... :smile: 

Another question, @landreev: should we make sure no future (accidential) usages of Log4J slip through via 

```diff
<module name="IllegalImport">
-            <property name="illegalPkgs" value="org.apache.commons.lang"/>
+            <property name="illegalPkgs" value="org.apache.commons.lang, org.apache.log4j"/>
        </module>
```
in `checkstyle.xml`? Our good doggo @reviewdog would bark at those (newly added lines).

pom.xml
landreev: This may be true. I'm planning to take another pass at cleaning it all up before moving the PR into review officially.

pom.xml
landreev: Yeah, without this explicit entry a different version of this library ends up in the .war file, 1.7.30; not the end of the world, but we should rather have all these org.slf4j libraries be from the same release.

pom.xml
landreev: That's what it looked like to me too - I was going to double-check with others, then remove. But yes, it appears to be safe to remove...

pom.xml
landreev: (also, I agree about adding an explicit "IllegalImport" rule for log4j)

landreev:
@poikilotherm One random question for you; since it looks like you've been looking into all this in parallel: Have you by any chance figured out why we are downloading the version 1.5.6 of slf4j-parent, slf4j-api, etc. during the build (*in addition* to 1.7.35, specified as ${slf4j.version})? 
I figured I'd ask, in case you've already found the answer.

pom.xml
landreev: Sure.
There is also an alternative of NOT doing any of this - adding these exclusion rules or including the dependency for reload - in our pom.xml; by modifying the XOAI .pom files in the local_lib... But I wanted to do this in a more obvious way, so that there is a reminder, with comments, in the main project pom file.

pdurbin (approved):
At standup I volunteered to take a look and everything seems fine to me so I'm moving it to QA.

I didn't deploy the war file but I compared `develop-984219b5be` to `ds48-l4j-upgrade-aae7bbb955` and found that the former has `./WEB-INF/lib/log4j-1.2.14.jar` but the latter (this pull request) does not. Both have `./WEB-INF/classes/log4j.properties` but this is just a config file.

@poikilotherm I see you had various comments but they seem to be are flagged as "outdated" now that @landreev has addressed them.

It's not a bad idea to add something to checkstyle but I'm ok waiting so as not to hold up this pull request.

landreev:
@pdurbin - thanks for approving! The part about adding an explicit "Illegal Import" rule for log4j - I agree that it's a good idea; but we can't do it just yet. Maybe after we overhaul that XOAI library dependency, via #IQSS/dataverse/8372.

poikilotherm:
> @poikilotherm One random question for you; since it looks like you've been looking into all this in parallel: Have you by any chance figured out why we are downloading the version 1.5.6 of slf4j-parent, slf4j-api, etc. during the build (_in addition_ to 1.7.35, specified as ${slf4j.version})? I figured I'd ask, in case you've already found the answer.

I tried to reproduce this on `develop` by removing all in `~/.m2/repository/org/slf4j/*`. Reviewed the download logs, but no SLF4J v1.5.6 or similar in there... :shrug: Didn't test on your PR though.

landreev:
> I tried to reproduce this on `develop` ... Reviewed the download logs, but no SLF4J v1.5.6 but v1.7.5 etc... 

Interesting. I'm clearly seeing it on this branch... 
```
% ls -d ~/.m2/repository/org/slf4j/*/1.5.6
/Users/landreev/.m2/repository/org/slf4j/jcl-over-slf4j/1.5.6
/Users/landreev/.m2/repository/org/slf4j/slf4j-api/1.5.6
/Users/landreev/.m2/repository/org/slf4j/slf4j-jdk14/1.5.6
/Users/landreev/.m2/repository/org/slf4j/slf4j-parent/1.5.6
% rm -rf ~/.m2/repository/org/slf4j/*/1.5.6
```
then run maven build and see it downloaded again:
```
Downloaded from central: https://repo.maven.apache.org/maven2/org/slf4j/slf4j-parent/1.5.6/slf4j-parent-1.5.6.pom (7.9 kB at 158 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/slf4j/slf4j-api/1.5.6/slf4j-api-1.5.6.pom
Progress (1): 2.7/3.0 kB
... etc.
```
I was assuming it was maven itself that needed this for some internal purposes. (trying to remember if I saw this on develop too). Not 100% sure what to make out of the fact that you appear to be getting a different different version ... 
Not the most important mystery - just curious. (I'll trace it if I have a free minute)

PR #8378, opened 2022-01-28
Ensure Single Copy of Release Notes

### LINKED ISSUE #7988: Process for updating release notes after a release
Our newish process for creating release notes is working well: https://guides.dataverse.org/en/5.5/developers/making-releases.html#write-release-notes

However, updating release notes after a release can be problematic. There is drift between the live release notes and the version in git.

Here are the failure scenarios:

- A pull request updates the version in git but the change doesn't make it to the live notes.
- The live notes are edited and the change is not propogated back to git. (This is less of a problem for end users and impossible for 4.15 and below where release notes aren't even in git.)

Possible solutions:

- Delete the version from git as part of the release process. No syncing to worry about. Only one source of truth. This is how the release notes are pre 4.16 anyway.
- As part of the merge process, copy the content from the updated file in git to the live release note.
- Write scripts to check if the files in git and the live release notes are in sync.
- Automate pushing content from the files in git to the live release notes.

Questions:

- With millions of projects on GitHub, we aren't the only one with this problem, are we? What do other projects do?
- Having the files in git makes it easier for contributors to edit the release notes. Should we keep this?

djbrooke:

We're in a state right now where we maintain two copies of our release notes - one in Github.com/iqss/dataverse/releases and one in the /docs folder. This isn't great, because if there are changes needed, people could make them in one of two places and they could fall out of sync. This PR makes documentation changes that will result in the single, post-release source of truth being the release notes in the github.com/iqss/dataverse/releases - any edits or additions would go there. 

Closes #7988

**Additional documentation**:


doc/sphinx-guides/source/developers/making-releases.rst
```diff
@@ -58,7 +58,7 @@ At or near release time:
 - The "tag version" and "title" should be the number of the milestone with a "v" in front (i.e. v4.16).
 - Copy in the content from the .md file
 - For the description, follow post-4.16 examples at https://github.com/IQSS/dataverse/releases
-
+- Delete the .md file in the source code. This is to ensure that only one copy of the notes lives on in the future, as a single source of truth.
```
pdurbin, line 61: @scolapasta this is the line I mentioned in standup that we might want to think about. That is, when does it make sense to remove the combined release note file from git?

I think of "merge develop into master" as the final step before a release with regard to commits. And since the release note file is temporary, we should probably remove it before this final step.

pdurbin (commented):
I left a comment.

pdurbin (approved):
Just testing approval automation.

pdurbin:
Closing. There are merge conflicts because we merged PR #8877 which updates the "Making Releases" page. There was enough of a rewrite of that page that we should simply make a new PR to address #7988 some day.

I agree with where this PR was going. That is, let's only have one version of the release notes, the "live" version, rather than saving .md files in git.

PR #8397, opened 2022-02-03
500 Error on template page

qqmyers:
per @pdurbin - Template page was giving 500 error. This was due to DatasetUtil:getLicenseName being used here without DatasetUtil having its functions include. Since we only need the custom terms string here, that can be included directly.

Closes #

**Suggestions on how to test this**: Check the template page - a) it should work, and b) Viewing one with custom terms should show "Custom Dataset Terms — the following Custom Dataset Terms have been defined for this dataset." in the License/Data Use Agreement area

**Is there a release notes update needed for this change?**: bug was only in dev so far

**Additional documentation**:


pdurbin (approved):
I haven't tested this but the fix makes sense.

PR #8401, opened 2022-02-04
Required single value controlledVocabulary field cannot be left empty when saving template

janvanmansum:
**What this PR does / why we need it**: When you create a template that contains a single value controlled vocabulary field that is marked as required it is currently not possible to save the template. You will get a validation error: 

> Required fields were missed or there was a validation error. Please scroll down to see details.

This is resolved by this PR. The solution was found by @qqmyers .

Closes n/a

**Special notes for your reviewer**: n/a

Create a custom metadata block with a required single value controlled vocabulary field and try to create a template with no values filled in.

**Additional documentation**: n/a


kcondon:
@janvanmansum I am unable to reproduce the problem: I've modified the astronomy .tsv file to include only the type cvv field, marked it as required in the tsv and uploaded it. I've then enabled it and added a template without specifying a value and it saved without error.

qqmyers:
FYI: Looking back at the code, I see the issue/fix are only when "#{!dsf.datasetFieldType.allowMultiples} so the test field above (and other ones I tried) which allow multiple values to be selected don't show the problem. @kcondon is going to test again tomorrow and hopefully can see the problem/fix work for that case.

PR #8402, opened 2022-02-04
link to guides when creating, editing, selecting guestbooks #8244

### LINKED ISSUE #8244: Help users creating a guestbook, including how to download results
This issue for text-only changes is to improve help for API users, Depositors, and Curators when creating guestbooks. It will be most helpful for self-deposit users (Harvard Dataverse), and is likely to add context for less frequent and new users.

1. Provide a link to the user guide (https://guides.dataverse.org/en/5.8/user/dataverse-management.html#dataset-guestbooks) when creating [, editing] and selecting a guestbook. The app contains links to guides in other cases.  

2. Add a link to the API for downloading guestbook data (https://guides.dataverse.org/en/5.8/api/native-api.html#) to the section on guestbooks in the user guides, so it is more complete.

~~3. Add text in the UI when creating and selecting a guestbook that links to the guides.~~

~~4. Add API link as specified in https://github.com/IQSS/dataverse/issues/8244#issuecomment-970696251~~

5. Add a download button on the create guestbook page so the person does not need to go to the responses page to see the button (since the responses page can take a long time to load if there are many... responses)

**Current screens**
When creating a new guestbook, editing a dataverse collection:
Screen Shot 2021-11-16 at 2 27 05 PM

Screen Shot 2021-11-16 at 2 27 32 PM

When selecting a guestbook for a dataset:
Screen Shot 2021-11-16 at 11 36 33 AM

pdurbin:

Users could use some help understanding how guestbooks work by reading the guides. We provide a link to the guides in a few more places.

Closes #8244

Yes. The "To learn more about" line was added in several places.

## create guestbook

Screen Shot 2022-02-04 at 2 31 42 PM

## edit guestbook

Screen Shot 2022-02-04 at 2 32 52 PM

## view terms while no guestbook is selected

(Note that the "help-block" CSS class was added to make the text grey.)

Screen Shot 2022-02-04 at 2 35 15 PM

## select a guestbook

Screen Shot 2022-02-04 at 2 36 13 PM

**Additional documentation**:

### DIFF
--- a/src/main/java/propertyFiles/Bundle.properties
+++ b/src/main/java/propertyFiles/Bundle.properties
@@ -1222,7 +1222,8 @@ dataset.manageGuestbooks.noGuestbooks.why.reason2=You can download the data coll
 dataset.manageGuestbooks.noGuestbooks.how.header=How To Use Guestbooks
 dataset.manageGuestbooks.noGuestbooks.how.tip1=A guestbook can be used for multiple datasets but only one guestbook can be used for a dataset.
 dataset.manageGuestbooks.noGuestbooks.how.tip2=Custom questions can have free form text answers or have a user select an answer from several options.
-dataset.manageGuestbooks.noGuestbooks.getStarted=To get started, click on the Create Dataset Guestbook button above. To learn more about Guestbooks, visit the <a href="{0}/{1}/user/dataverse-management.html#dataset-guestbooks" title="Dataset Guestbook - Dataverse User Guide" target="_blank">Dataset Guestbook</a> section of the User Guide.
+dataset.manageGuestbooks.noGuestbooks.getStarted=To get started, click on the Create Dataset Guestbook button above.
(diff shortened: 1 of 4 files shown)


TaniaSchlatter:
Thanks for the screenshots, @pdurbin. Which user types see the "view terms while no guestbook is selected" example? Dataset downloaders? Dataset owners? Both?

TaniaSchlatter:
@pdurbin ideally dataset users do not see the Guestbook module unless there is a Guestbook (view terms while no guestbook is selected example). I recognize that might be out of scope. As is, the language is written for a data owner, and might be confusing to a data user. Let's discuss options.

pdurbin:
@TaniaSchlatter @scolapasta and I had a nice pow wow about next steps.

I went over some changes to the guides I wanted to met and after we hung up I implemented them (plus a couple more) in d27a45bf7b. This includes putting quotes around button labels, as discussed. Feedback is certainly welcome.

I also lowercased to "guestbooks" in 0d34784a49 as discussed.

A principle we discussed is that downloaders don't need to see the link to the guides. Toward that end, as suggested above, we plan to hide the entire Guestbooks panel for non-editors (people just browsing anonymously, for example) if there is no guestbook selected. (Less is more. Don't make me think.) Since only editors will see this we can change the message from "you will not" to "users will not".

I'll ping Tania when I'm done with the panel so we can take another look.

PR #8404, opened 2022-02-07
8094 test reports

poikilotherm:
After adding experimental JDK 17 build support, it would be nice to have a easier to read test report than filtering build log output. This PR aims to add them.

Relates to #8094 

I edited a single Java file to trigger the build in a commit that would still be useful...

Take a look at the failing tests for JDK 17

Nope.

Nope.

**Additional documentation**:
None.

### DIFF
--- a/.github/workflows/maven_unit_test.yml
+++ b/.github/workflows/maven_unit_test.yml
@@ -43,3 +43,7 @@ jobs:
                 CI_NAME: github
                 COVERALLS_SECRET: ${{ secrets.GITHUB_TOKEN }}
             run: mvn -V -B jacoco:report coveralls:report -DrepoToken=${COVERALLS_SECRET} -DpullRequest=${{ github.event.number }}
+          - name: Publish Test Report
+            uses: scacap/action-surefire-report@v1
+            with:
+                check_name: JDK ${{ matrix.jdk }} Test Report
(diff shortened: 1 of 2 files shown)


poikilotherm:
Closing - there is no way a pull request from a forked repository is allowed to write a report.

See
- https://github.com/ScaCap/action-surefire-report/issues/31#issuecomment-973980120
- https://github.com/mikepenz/action-junit-report/issues/97
- https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings.

poikilotherm:
Ok there MIGHT be a way to do this - using `pull_request_target` instead of `pull_request` runs from the trusted target repo and receives a token that may grant write access. Will need to test this with a fork of a fork though.

poikilotherm:
It might be possible to work around this by uploading the test results (XML files) as a build artifact and process them with another workflow, using the new `workflow_run` event.

https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

PR #8409, opened 2022-02-09
Direct upload multiple files

### LINKED ISSUE #8408: Undesired behaviour when uploading multiple files via GUI and direct upload
During tests with the current Dataverse 5.9, we discovered a strange behaviour with direct uploads via GUI.

- Prerequisite: Direct upload must be enabled
- Create a new dataset
- Click "Upload Files" button
- Click "Select Files to add"
- Select multiple files (at least 20)

The upload is starting correctly, but stops before the last few files (2-6) are uploaded. GUI is still responsive, but nothing happens anymore. When clicking done, none of the files are added to the dataset
grafik

This behaviour can be reproduced, but there's always a different amount of items that are not uploaded. It seems to be a concurrency problem.
In the JS console, some NPEs are visible:
Uncaught (in promise) TypeError: upload is null  (fileupload.js.xhtml:411:36)

I will provide a pull request with a quite simple fix for this issue, but I don't really understand the original intention of the code snippet. So a review would be great.

Thanks
Markus

haarli:
This pull request addresses and fixes issue #8408.

As stated in the issue, I'm not really sure about the intention of the original code, in particular the splice-method with the 'i+1' deleteCount parameter. So a review would be great.

Generally, this seems to be a concurrency problem, there might still be concurrency issues with this bugfix, but it works much better

Closes #8408


qqmyers (approved):
I think this makes sense - fixing an embarrassing bug :-). It's been a while, but I think the code is just scanning the list of pending files and trying to pull one out to work on. It should (with the fix) just pull one entry out of the list (not necessarily the first entry) to process. The original bug would remove more than one entry (more the larger i is - param 2 is number of entries to remove) and only work on the first one. Not a race per se, but a bug that would ignore some files. (I'd have to dig further to see how/when the request to this method wouldn't be trying to get the first element (i=0, which works OK with the bug) but I could imagine it being very intermittent.)

Thanks for the fix!

haarli:
Thank you for the prompt reply and merge.
Yes, I thought so, too. Just wasn't sure about the i+1 and if there was an idea behind it. But probably just a confusion between (startIndex, endIndex), like many methods use, and (startIndex, deleteCount), like splice uses. Not embararrassing at all.

Thanks again!

PR #8413, opened 2022-02-09
IQSS/8407 fix - update CC0 1.0 info in db/align json with other licenses

### LINKED ISSUE #8407: Multilicense flyway not consistent with JSON examples
The [CC0 1.0 license json file](https://github.com/IQSS/dataverse/blob/develop/scripts/api/data/licenses/licenseCC0-1.0.json) is consistent with the other examples in using the title recommended by Creative Commons (including the version number) but this is inconsistent with the update provided by the flyway script (which was created early on). The URL (differing by a trailing slash), short description, and icon location also differ. The db upgrade should match the examples and the pattern for the other licenses.

There seem to be two options for implementing

1. Update the existing flyway script (requires developers to remove the record of running the prior version from their db, but avoids a second flyway script in this release)
2. Add a second flyway script to modify the results of the first (possibly simpler to create, easier for developers, ends up with second flyway script)

@scolapasta /others - any preference? (I think I'm leaning towards 2, but will implement whatever).

qqmyers:
**What this PR does / why we need it**: This is a minor DB update to align the upgrade changes create by the prior flyway script to match the current CC0 1.0 json file and to align the URl in that file with the other licenses (i.e. no trailing slash).

Closes #8407

I spent some time today trying to figure out how to handle the conditional logic required to update the prior script and still work on databases that had deployed develop already (i.e. starting from a 5.9 database or from one that already had the multilicense changes applied and only needing the updates to the name/description/icon for CC0 1.0 . I wasn't very successful and started thinking that that approach would be a bear to test (assuming you test new db, 5.9 update and update from develop). Instead I just developed ~option 2 in the issue which results in a 5 line flyway script that can be then be added as a last step in the original script (i.e. relying on the fact that the original is idempotent).

I also saw the suggestion in the issue to not have a flyway script duplicating what is in the json file. As far as I understand, that would involve the upgrade instructions adding two manual steps (deploy CC0 1.0 and then run a manual update sql script) rather than providing a fully automated option (aside from the prior discussion that people may want to do some manual work to change a few corner cases rather than accept the default automated results). Given that, and the fact that we've done db updates to things like external tools before rather than have people re-register them, and wanting to avoid the work to make those changes and update/explain the manual steps, I didn't pursue that option.

Edit your db to remove the prior run of the V5.9.0.1 script from the flyway_schema_history table if you've run the develop branch before. Verify that the deployment results in a license with the name 'CC0 1.0', shortdescription 'Creative Commons CC0 1.0 Universal Public Domain Dedication.' and pulling the icon from 'https://licensebuttons.net/p/zero/1.0/88x31.png' exists instead of the 'CC0' license with other values for the shortdescription and icon URL that a) the old script produces, and b) would have already been in the table if you had run develop before. Nominally all of the license info is visible in the UI/browser console, but one could check the db entry in addition/instead.

After merge, let me know so we can alert developers that they will also have to delete the latest flyway entry if they were on develop already.

**Additional documentation**:

### DIFF
--- a/scripts/api/data/licenses/licenseCC0-1.0.json
+++ b/scripts/api/data/licenses/licenseCC0-1.0.json
@@ -1,6 +1,6 @@
 {
   "name": "CC0 1.0",
-  "uri": "https://creativecommons.org/publicdomain/zero/1.0/",
+  "uri": "https://creativecommons.org/publicdomain/zero/1.0",
   "shortDescription": "Creative Commons CC0 1.0 Universal Public Domain Dedication.",
   "iconUrl": "https://licensebuttons.net/p/zero/1.0/88x31.png",
   "active": true
--- a/src/main/resources/db/migration/V5.9.0.1__7440-configurable-license-list.sql
+++ b/src/main/resources/db/migration/V5.9.0.1__7440-configurable-license-list.sql
@@ -71,6 +71,12 @@ BEGIN
   EXCEPTION
     WHEN undefined_column THEN RAISE NOTICE 'license is not in table - new instance';
   END;
+  
+  BEGIN
+    UPDATE license SET shortdescription='Creative Commons CC0 1.0 Universal Public Domain Dedication.' WHERE name='CC0';
+    UPDATE license SET iconurl='https://licensebuttons.net/p/zero/1.0/88x31.png' WHERE name='CC0';
+    UPDATE license SET name='CC0 1.0' WHERE name='CC0';
+  END;
 
 END $$;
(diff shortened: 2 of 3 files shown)


pdurbin (approved):
I didn't run this but it seems ok.

pdurbin (approved):
Trying another approval review.

pdurbin (approved):
Another "Approve" after deleting "QA" and renaming "QA2" to "QA".

pdurbin:
As discussed at standup, I just sent a suggestion to devs if they have trouble deploying: https://groups.google.com/g/dataverse-dev/c/937iVHJgV40/m/xD7YPNhSBAAJ

PR #8418, opened 2022-02-11
IQSS#8417 Only show role assignments that are related to the object type (Dataverse/Dataset/DataFile)

### LINKED ISSUE #8417: Improvement: Do not display "Dataset Creator" permissions on dataset level
We want to apply the "Dataset Creator" role to certain users on Dataverse level (which is the only level that makes sense with this role). However, when doing this, the permission list of every dataset in this Dataverse displays this role (View Dataset -> Edit -> Permissions -> Dataset), although this role does not have any impact on this particular dataset. 

grafik

This is really confusing for users, in particular for the real owners of the dataset. They might belive that all persons in the list have access to their dataset, even if unpublished. Wouldn't it make sense to remove Dataset Creators from the Dataset permission list? Of course, they should still be visible in the Dataverse permission list.

I will provide a pull request for this

haarli:
This pull request addresses IQSS#8417.

DatasetCreator roles, which are applied on dataverse level, will not be displayed in the  permission list of every dataset in the dataverse

Closes #8417


haarli:
I changed the method and reused the logic from getAvailableRoles, to only display permissions that are related to the dvobject type.

However, not 100% sure about the following:
I removed the condition 
```
if (!(dvObject instanceof DataFile) || roleAssignment.getRole().permissions().contains(Permission.DownloadFile))
```
in favour of
```
if (availableRoles.contains(roleAssignment.getRole()))
```

The `getAvailableRoles()` method returns for DataFiles only the role "File Downloader". So this is a slight logic change to before, where every role that contains the "DownloadFile" permission was returned.
Not that deep into the Dataverse code to judge if this is ok.

pdurbin:
> I changed the method and reused the logic from getAvailableRoles, to only display permissions that are related to the dvobject type.

Are roles other than DatasetCreator affected? If so, can you please modify the title of this pull request? Thanks.

qqmyers:
Yeah - my guess is that the PR is now good and the issue you point out is really a (currently unimportant) limitation of the getAvailableRoles() logic - I think there is currently only one standard role with the DownloadPermission and given that there was, until recently, only one datafile-assignable permission there was also no reason to create another role. 

However, per @scolopasta, when I recently added the ManageFilePermissions, I made is assignableTo DataFile. That was mainly for future-proofing/preparing for the option to allow someone to have ManageFilePermissions (the ability to assign the Downloader role/DownloadPermission to someone) on individual files. If/when someone implements that the getAvailableRoles() logic would need to be updated to show any role with either DownloadPermission or ManageFilePermissions for DataFiles. (Not sure  -maybe this would be the only change needed?)

In any case, I think your change to use getAvailableRoles won't affect anything important now and it should still work if/when we want ManageFilePermissions to be assigned per file instead of only at the Dataverse/Dataset level.

Last - I didn't intend to be the reviewer who accepts this, but as far as I can tell, the PR code now should be good to go.

scolapasta:
As I was reading through this, I had the same though that @qqmyers beat me to (about making it more general), so 👍 . I'll go ahead and grab it for official review, as I would like to take a look.

scolapasta (commented):
OK, I reviewed and it looks good. I went ahead a made a small PR to your branch to add a comment related to what @qqmyers was referring to. If you can merge that PR into your branch, I will then approve and move to QA. Thanks!

scolapasta:
> OK, I reviewed and it looks good. I went ahead a made a small PR to your branch to add a comment related to what @qqmyers was referring to. If you can merge that PR into your branch, I will then approve and move to QA. Thanks!

Actually, let me go ahead and make the change now, rather than the comment. Be back soon with a the PR.

scolapasta:
OK, *now* it should be all set; if you approve and merge that PR, I can send this off to QA. Thanks!

haarli:
I approved and merged your PR, thanks for that

qqmyers (approved):
Looks good to me - didn't run it but this has the changes we've been discussing.

scolapasta:
Thanks for reviewing my changes = moving to QA.

PR #8424, opened 2022-02-15
Signposting on Dataverse 5.9

### LINKED ISSUE #5962: Signposting support in Dataverse
Signposting is an approach to make the scholarly web more friendly to machines. It uses Typed Links as a means to clarify patterns that occur repeatedly in scholarly portals. For resources of any media type, these typed links are provided in HTTP Link headers.

At [DANS](https://dans.knaw.nl/en) we implemented Signposting in [EASY](https://easy.dans.knaw.nl/ui/home) on our dataset landing pages with patterns for 'bibliographic metadata' and 'identifiers'. See https://signposting.org/adopters/#dans for examples.

It would be nice to have Signposting implemented in Dataverse as well. We have to see which patterns are suitable. Let's discuss that in this issue.

vicding-mi:
This PR adds signposting to Dataverse. Signposting enables easier machine access to datasets and hence increases accessibility. 
N/A
Closes #5962
N/A

This PR does not change any human readable part of Dataverse. 

`curl -I http://linkto.dataverse/dataset` to see the signposting L1 header

Link to L2 linkset can be found in L1 header and can be directly viewed in browser

No
in folder `doc/release-notes/8424-signposting.md`
**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+                    da.getAffiliation(),
+                    da.getIdentifierAsUrl()
+            ));
+
+            String authorURL = "";
+            authorURL = getAuthorUrl(da);
+            if (!Objects.equals(authorURL, "")) {
+                visibleAuthorCounter++;
+                // return empty if number of visible author more than max allowed
+                if (visibleAuthorCounter >= maxAuthors) return "";
```
qqmyers, line 99: So the intent is to show no authors if there are more than max and not the first max authors? (And we're only talking about authors with URIs? If there are ten authors, three who have ORCIDs, you'd still show 3 authors for a max of 5?)

If so, reusing getJsonAuthors might involve adding a param to that method to check the max (if I understand, the level1 header uses the max limit but the linkset always shows however many there are?)

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+        logger.info(String.format("identifierSchema: %s", identifierSchema));
+        return identifierSchema;
+    }
+
+    /**
+     * Get key, values of signposting items and return as string
+     *
+     * @return comma delimited string
+     */
+    public String getLinks() {
```
qqmyers, line 118: A similar question to authors - this is assembling the same info as in the linkset except that this a) results in a string and b) enforces the max limits? Seems like basing this off the json output with the max limits as params would avoid duplicating the code to retrieve this info?

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+
+        List<FileMetadata> fms = workingDatasetVersion.getFileMetadatas();
+        String items = getItems(fms);
+        if (items != null && !Objects.equals(items, "")) {
+            valueList.add(items);
+        }
+
+        String describedby = "<" + describedByJsonObj.getString(ds.getProtocol()) + ds.getAuthority() + "/"
+                + ds.getIdentifier() + ">;rel=\"describedby\"" + ";type=\"" + describedByJsonObj.getString("type") + "\"";
+        describedby += ",<" + systemConfig.getDataverseSiteUrl() + "/api/datasets/export?exporter=schema.org&persistentId="
```
qqmyers, line 140: Just curious - does Signposting prefer this export? Or would the more complete JSON or OAI_ORE exports be more relevant? (Or can the dataset be described by all the exports so you can find the format you understand?)

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+                + ds.getProtocol() + ":" + ds.getAuthority() + "/" + ds.getIdentifier() + ">;rel=\"describedby\"" + ";type=\"application/json+ld\"";
+        valueList.add(describedby);
+
+        String type = "<https://schema.org/AboutPage>;rel=\"type\"";;
+        if (useDefaultFileType) {
+            type = "<https://schema.org/AboutPage>;rel=\"type\",<" + defaultFileTypeValue + ">;rel=\"type\"";
+        }
+        valueList.add(type);
+
+        // TODO: support only CC0 now, should add flexible license support when flex-terms or multi-license is ready
```
qqmyers, line 150: TODO is done

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+//        License license = workingDatasetVersion.getTermsOfUseAndAccess().getLicense();
+        License license = workingDatasetVersion.getTermsOfUseAndAccess().getLicense();
+        String licenseString = license.getUri() + ";rel=\"license\"";
```
qqmyers, line 153:
```suggestion
        String licenseString = DatasetUtil.getLicenseURI(workingDatasetVersion) + ";rel=\"license\"";
```

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+
+        String type = "<https://schema.org/AboutPage>;rel=\"type\"";;
+        if (useDefaultFileType) {
+            type = "<https://schema.org/AboutPage>;rel=\"type\",<" + defaultFileTypeValue + ">;rel=\"type\"";
+        }
+        valueList.add(type);
+
+        // TODO: support only CC0 now, should add flexible license support when flex-terms or multi-license is ready
+//        License license = workingDatasetVersion.getTermsOfUseAndAccess().getLicense();
+        License license = workingDatasetVersion.getTermsOfUseAndAccess().getLicense();
```
qqmyers, line 152: License is null when custom terms are used. The DataUtil method will get you a URI for both standard and custom licenses.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+
+        /*
+        else {
+            TODO: when multilicense is merged, should get license link from multilicense
+             Thus the checking thus would not be necessary?
+             Currently, when the License is None, which means bring-your-own-license,
+             Mostly, we will not have a valid URI in the License, so we skip it here despite that signposting requires 1
+             valid URI.
+        }
+        */
```
qqmyers, line 164: resolved?

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+        return String.join(", ", valueList);
+    }
+
+    private String getAuthorUrl(DatasetAuthor da) {
+        String authorURL = "";
+        if (da.getIdValue() != null && !da.getIdValue().trim().isEmpty()) {
+            authorURL = da.getIdValue();
+        } else if (da.getIdentifierAsUrl() != null && !da.getIdentifierAsUrl().trim().isEmpty()) {
+            authorURL = da.getIdentifierAsUrl();
+        }
```
qqmyers, line 181: Looks to me like getIdentifierAsUrl already includes this type of checking (could be new since when you started coding) - is this still needed?

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+
+    private String getPublicDownloadUrl(DataFile dataFile) {
+        StorageIO<DataFile> storageIO = null;
+        try {
+            storageIO = dataFile.getStorageIO();
+        } catch (IOException e) {
+            logger.warning(String.format("Error getting storageID from file; original error message is: %s", e.getLocalizedMessage()));
+        }
+
+        if (storageIO instanceof SwiftAccessIO) {
```
qqmyers, line 314: If I understand, this whole section is trying to expose the direct URL for the file. I'm not sure I like that in general (it means that a Dataverse instance manager has to never move the data.) I have other comments below, but if this is kept, I think it should at least be well documented and probably made optional/off by default.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+            String fileDownloadUrl;
+            SwiftAccessIO<DataFile> swiftIO = (SwiftAccessIO<DataFile>) storageIO;
+            try {
+                swiftIO.open();
```
qqmyers, line 318: In general, calling stoargeIO.open() will open an inputstream allowing you to read the file. (The S3 store does not, but File does and I think I see that swift does too.) If you call open, you need to call StorageIO.closeInputStream() before you finish.

Ah - and I see you copied this from FilePage.getPublicDownloadUrl() which also has this bug!

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+                logger.warning(String.format("Error opening the swiftIO; original error message is: %s", e.getLocalizedMessage()));
+            }
+
+            //if its a public install, lets just give users the permanent URL!
+            if (systemConfig.isPublicInstall()) {
+                fileDownloadUrl = swiftIO.getRemoteUrl();
+            } else {
+                //TODO: if a user has access to this file, they should be given the swift url
+                // perhaps even we could use this as the "private url"
+                fileDownloadUrl = swiftIO.getTemporarySwiftUrl();
```
qqmyers, line 329: This would be temporary, i.e. access would expire. In the file page, where someone is navigating now to get the file, this works but as a permanent link I think this is problematic.

qqmyers (changes requested):
Overall, when I look in the dataset page HTTP header, I see the new level 1 info and if I follow the linkset link given there, I get the json info expected. When this gets to QA, it would probably be worth pinging @vicding-mi and/or Herbert Van de Sompel (or others) who know the signposting 'spec' to make sure the outputs are indeed valid.

W.r.t. code, I've made a bunch of comments/suggestions - a few are real issues and others are more code style/avoiding duplicate code suggestions. I understand @vicding-mi is no longer at DANS where this started and may have limited time for changes. At a minimum I think getting the key issues resolved and knowing whether the other suggestions make sense or not would help move this forward. (I.e. if @vicding-mi doesn't have time to refactor code or add some documentation but agrees with the suggestions, we can either see about leaving some things as ToDos, or looking for others who can help with pushing those forward. And if there's some reason not to do the suggestions, it would be good to know that.

vicding-mi:
@qqmyers many thanks for the quick response. I will work on your feedback ASAP. Good to (almost) close a ticket ;)

doc/release-notes/8424-signposting.md
```diff
@@ -0,0 +1,42 @@
...
+`signposting.configuration.SignpostingConf`. Please see a sample configuration below with explaination for each of the
+config items.
+
+```json
+{
+  "license": {
+    "CC0": "https://creativecommons.org/licenses/cc0/"
+  },
+  "describedby": {
+    "doi": "https://doi.org/",
```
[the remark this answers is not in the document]
vicding-mi: This is to make sure in case people have a dataverse instance with mixed id types, they can get proper url, `https://doi.org` in this case, by using `describedByJson.getString(ds.getProtocol())` 
If there is a better way to get per dataset `identifier` and `identifier type`, please do let me know. Thanks!

doc/release-notes/8424-signposting.md
```diff
@@ -0,0 +1,42 @@
...
+    "type": "application/vnd.citationstyles.csl+json"
+  },
+  "useDefaultFileType": true,
+  "defaultFileTypeValue": "https://schema.org/Dataset",
```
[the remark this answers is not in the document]
vicding-mi: starting with `userDefaultFileType`, these are all Signposting specific configuration. 
Signposting can show different FileType according to the given metadata, however, in dataverse these metadata is not there yet. I left this to be configurable. If `true`, the type will be the set type, else it will be the actual type given in metadata.

doc/release-notes/8424-signposting.md
```diff
@@ -0,0 +1,42 @@
...
+  "maxItems": 5,
+  "maxAuthors": 5
+}
```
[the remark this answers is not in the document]
vicding-mi: The signposting cannot be disabled :D 

The parameters can be set to the preference of the users. We wanted to make it slightly customizable as in the first draft these are not changeable and the level 1 header becomes really messy and difficult to read, even though it is not particularly set for human to read. ;) 

For example, for now Herbert is ok with a default file type set to `https://schema.org/Dataset`. This might change when there is an upstream schema change or he changes his idea. 
`maxItems` determines max items/files to show in Level 1 profile, as it can get really lengthy. Same goes for `maxAuthor`.

doc/release-notes/8424-signposting.md
```diff
@@ -0,0 +1,42 @@
...
+which describes the resources that is the origin of the link.
+ * `useDefaultFileType` and `defaultFileTypeValue` are used in combination to provide extra `Dataset` type to DV 
+   datasets. `AboutPage` is required by `Signposting`, hence always present in the datasets. Whilst a second type 
+   could be configured to better reflect the actual scholarly type of the dataset. 
+ * `maxItems` sets the max number of items/files which will be shown in `level 1` profile. Datasets with 
+   too many files will not show any file link in `level 1` profile. They will be shown in `level 2` linkset only. 
+ * `maxAuthors` Same with `maxItems`, `maxAuthors` sets the max number of authors to be shown in `level 1` profile. 
+If amount of authors exceeds this value, no link of authors will be shown in `level 1` profile. 
+
+Note: Authors without author link will not be counted nor shown in any profile/linkset.
```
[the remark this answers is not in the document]
vicding-mi: Sure, please let me know the ideal location, I can add some docs accordingly.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,337 @@
...
+     * Get identifier schema for each author
+     * <p>
+     * For example:
+     * if author has VIAF
+     * Link: <http://viaf.org/viaf/:id/>; rel="author"
+     *
+     * @param datasetAuthors list of all DatasetAuthor object
+     * @return all the non empty author links in a string
+     */
+    private String getAuthors(List<DatasetAuthor> datasetAuthors) {
```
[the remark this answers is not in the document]
vicding-mi: sorry for the bad naming, renamed it to getAuthorsIdSchema, the purpose is to get the author Id schema from list of authors. Please see if the modification is fine, if not, please let me know which system call I can use instead. Thanks!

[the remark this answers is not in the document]
vicding-mi: I tried this approach, call and get json instead of using DatasetAuthors directly. It does not simplify the code, actually did just the opposite. 

The Json structure has to be converted anyway to string, while the resulting string is not a json string. The empty link check is redundant comparing to the json call, however, the logic is much simpler using datasetauthors rather than push every author into array, and then read them back, and then put into link string format.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
vicding-mi: same as above. in level 2, the full json structure is needed. In level 1, it's string format, not json. 
Only authors with links should be counted and displayed in either level 1 or 2.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
vicding-mi: This is the format described in the level 1 protocol. Might be a good idea check with Herbert.

vicding-mi:
hey guys, not sure how to carry on the process/discussion. Can @qqmyers please give some guidance? I changed according to the review some of the points, while for the other points I tried to answers the review questions and ask for suggestions. Many thanks!

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
qqmyers: I think URLs like http://localhost:8080/api/access/datafile/15 would be best practice. I don't think you get that from the swiftIO.getTemporarySwiftUrl() call. That download URL - /api/access/datafile - comes from the api/Access.java class where I think you could add a header. (Similarly, I think you could respond to a HEAD instead of GET there as well).

qqmyers (commented):
Tried to add some feedback but it looks like there are still some open questions/change requests too. I'll contact Herbert about which metadata file(s) should be referenced. 
Most of the discussion is about what should be configurable and if it's worth refactoring some parts. Neither are things that I'd hold this up for so I think when the other issues are resolved.

doc/release-notes/8424-signposting.md
```diff
@@ -0,0 +1,38 @@
...
+
+This branch adds [Signposting](https://signposting.org/) support to Dataverse
+
+There are 2 Signposting profile levels, level 1 and level 2. In this implementation, 
+ * level 1 links are shown in 
+HTTP header, which can be fetched by `curl -I https://domain/link-to-article`. 
+ * The level 2 linkset can be fetched by visiting the dedicated linkset page for 
+   that artifact. The link can be seen in level 1 links with key name `rel="linkset"`.
+
+The configuration is stored as JSON string in the `Bundle.properties` file, key name is
```
pdurbin, line 11: Why is the config in Bundle.properties? I see elsewhere that @qqmyers has suggested a database setting called :Signposting, which sounds like a better approach to me.

pdurbin (commented):
@vicding-mi thanks for this pull request! I'm excited for Signposting support in Dataverse. I haven't had a chance to do a proper review yet but I'm at least leaving a comment about the config that's been on my mind.

vicding-mi:
Hi @pdurbin and @qqmyers, many thanks for your comment. 
bundle file is the solution I got from the proof of concept, so I sticked to it. Since db is a better place for storing config, I will change it and update you guys.

@qqmyers , thanks for your review, will revise accordingly.

pdurbin:
@vicding-mi great! Thanks! Please feel free to create as many database settings as you need. I think the record is all the "PV" settings: [:PVMinLength](https://guides.dataverse.org/en/5.9/installation/config.html#pvminlength) [:PVMaxLength](https://guides.dataverse.org/en/5.9/installation/config.html#pvmaxlength) [:PVNumberOfConsecutiveDigitsAllowed](https://guides.dataverse.org/en/5.9/installation/config.html#pvnumberofconsecutivedigitsallowed) [:PVCharacterRules](https://guides.dataverse.org/en/5.9/installation/config.html#pvcharacterrules) [:PVNumberOfCharacteristics](https://guides.dataverse.org/en/5.9/installation/config.html#pvnumberofcharacteristics) [:PVDictionaries](https://guides.dataverse.org/en/5.9/installation/config.html#pvdictionaries) [:PVGoodStrength](https://guides.dataverse.org/en/5.9/installation/config.html#pvgoodstrength)[:PVCustomPasswordResetAlertMessage](https://guides.dataverse.org/en/5.9/installation/config.html#pvcustompasswordresetalertmessage) 😄

doc/release-notes/8424-signposting.md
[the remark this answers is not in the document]
pdurbin: There's no FAIR section but I just created https://github.com/vicding-mi/dataverse/pull/1 to suggest creating a "Discoverability" page in the Admin Guide. @vicding-mi if you like this, please feel free to merge it and add more docs. Thanks.

pdurbin (commented):
A few comments and some doc suggestions in the form of a pull request.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,321 @@
...
+                singleAuthorString = "<" + authorURL + ">;rel=\"author\"";
+                if (Objects.equals(identifierSchema, "")) {
+                    identifierSchema = singleAuthorString;
+                } else {
+                    identifierSchema = String.join(",", identifierSchema, singleAuthorString);
+                }
+            }
+        }
+
+        logger.info(String.format("identifierSchema: %s", identifierSchema));
```
pdurbin, line 106: This should be lowered to logger.fine.

src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -0,0 +1,321 @@
...
+                    da.getIdentifierAsUrl()
+            ));
+
+            String authorURL = "";
+            authorURL = getAuthorUrl(da);
+            if (!Objects.equals(authorURL, "")) {
+                visibleAuthorCounter++;
+                // return empty if number of visible author more than max allowed
+                if (visibleAuthorCounter >= maxAuthors) return "";
+                singleAuthorString = "<" + authorURL + ">;rel=\"author\"";
```
pdurbin, line 97: I'm not sure why but I'm getting null for authorURL like this (I'm on 5fe6a8a8ee):

```
Link: <null>;rel="author", <https://doi.org/...
```

This is just a local dataset in dev. Here's the curl command:

```
curl -s -i http://localhost:8080/dataset.xhtml?persistentId=doi:10.5072/FK2/KQRKB8
```

src/main/webapp/dataset.xhtml
```diff
@@ -88,6 +88,11 @@
                 </ui:fragment>
             </ui:define>
             <ui:define name="body">
+                <!-- Add Signposting-->
+                <c:if test="#{not empty DatasetPage.getSignpostingLinkHeader()}">
+                    <f:event type="preRenderView" listener="#{facesContext.externalContext.response.setHeader('Link', DatasetPage.getSignpostingLinkHeader())}" />
+                </c:if>
+                <!-- End add Signposting-->
```
pdurbin, line 95: I'm leaving this comment here at the end because this is where we usually have tests. Can some tests be written for this code?

pdurbin (commented):
First, great pull request. I'm excited to have Signposting support in Dataverse!

One thing I want to make clear from the start is that core developers can probably jump in and hack on the code and docs to get this over the finish line. Please just let us know.

As I mentioned at https://github.com/IQSS/dataverse/pull/8424#discussion_r824213197 I made a pull request (into this one) to suggest where to document Signposting.

I also left some other feedback about a `<null>` I'm seeing, about tests, agreeing with Jim about the ok method, etc.

qqmyers:
Closing in favor of #8981 which builds on the commits here.

PR #8437, opened 2022-02-17
IQSS/8337_add metadatalanguage to export/import/migrate

### LINKED ISSUE #8337: Feature Request/Idea: Support metadatalanguage in imports
The ability to specify the language used in a datasets metadata introduced in #7958 is not yet supported in import. It would be useful if one could optionally specify the metadatalangauge in imports so that this value would be set for imported datasets. If metadatalanguages are not enabled, any value in the import would have to match the installation default language. If the list of metadatalanguages is configured, the import value would have to match one of them.

API User

Sciences PO interest in importing datasets in multiple languages

qqmyers:
**What this PR does / why we need it**: This PR adds support for the new metadatalanguage value to the OAI_ORE export, and the json import/json-ld migrate API calls. This makes the metadata language used for the dataset visible in one more export and allows sending that value in when importing /migrating from prior exports/other systems.

Closes #8337

**Special notes for your reviewer**: FWIW: Since the metadatalanguage setting constrains which languages are allowed, the import/migrate api calls return a Bad Request error if you send one that isn't allowed. (Also assumes the default if you don't send one.)

**Suggestions on how to test this**: Configure with metadata languages and verify that the OAI-ORE export contains an https://schema.org/inLanguage  value. The import and migrate APIs can also be tested by adding this value. (I'll look at the guide examples - this is optional so I don't want to add it to all examples but I can see about showing it somewhere.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: export and api effects only. (FWIW: This is one of several updates to the OAI_ORE that will be happening - DataCommons is making several more so we may have an overall release note on the package of changes.)

**Additional documentation**:


src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -2417,17 +2417,20 @@ public void testReCreateDataset() {
...
+        response.prettyPeek();
         String expectedString = getData(response.getBody().asString());
         
         // Delete the dataset via native API
         Response deleteDatasetResponse = UtilIT.deleteDatasetViaNativeApi(datasetId, apiToken);
         deleteDatasetResponse.prettyPrint();
         assertEquals(200, deleteDatasetResponse.getStatusCode());
 
+        
+        logger.info("SENDING to reCreate Dataset: " + expectedString);
```
landreev, line 2429: I'm going to assume this info log message was left here on purpose.

landreev (approved):
Makes sense.

kcondon:
Retesting:
1. With no languages or metadatalanguages defined, json import and dataset migration endpoints fail with Specified metadatalanguage ( metadataLanguage, http://schema.org/inLanguage) not allowed in this collection. DDI import works.

src/main/java/edu/harvard/iq/dataverse/api/imports/ImportDDIServiceBean.java
```diff
@@ -182,6 +182,11 @@ private void processDDI(ImportType importType, XMLStreamReader xmlr, DatasetDTO
             throw new XMLStreamException("It doesn't start with the XML element <codeBook>");
         }
         
+        //Include metadataLanguage from an xml:lang attribute if present (null==undefined)
+        String metadataLanguage= xmlr.getAttributeValue("http://www.w3.org/XML/1998/namespace", "lang");
+        logger.info("Found metadatalanguage in ddi xml: " + metadataLanguage);
```
qqmyers, line 187:
```suggestion
        logger.fine("Found metadatalanguage in ddi xml: " + metadataLanguage);
```

kcondon:
Thanks, @qqmyers !

PR #8440, opened 2022-02-18
IQSS/8422 default to allow dataset level choice of metadatalanguage

### LINKED ISSUE #8422: Metadata Languages should be able to be disabled per collection, and set at the dataset level

This is a follow-up ticket to a Dataverse Community Slack conversation on Feb. 1st.

The Metadata Languages feature is currently configurable at the installation level and, when enabled, is not able to be disabled per collection. This becomes and issue in a multilingual Dataverse setting, particularly when a collection can contain datasets with more than one language. To improve this feature and make it more broadly usable, there are two suggestions:

- Allow for the setting to be turned off for each collection. In our case, we want to disable it at the root collection, otherwise it is inherited into all sub-collections.
- Allow for this to be set at the dataset level, instead of only the collection level. A collection may have datasets in more than one language, and this setting makes much more sense since it applies to a dataset. It could default to the collection-level setting, but be changed by the dataset creator.

Curators, dataset creators

We enabled this setting when upgrading at Scholars Portal, but turned the feature off because it was showing all sub-collections as using "English" metadata. This was an issue because we have bilingual and French institutions using our service, and were concerned it would be confusing from a UX perspective as well as it resulted in the metadata record saying the dataset was in English as soon as the feature was enabled ("This dataset has been configured to use English as the language for all metadata entries."), when it may not have been.

The Metadata Languages feature (see above).

Metadata languages set at the dataset level.

qqmyers:
**What this PR does / why we need it**: Per the issue, the recent enhancement to add a choice of metadata language restricted that choice to collections and, if no collection-level choice was made, the global default option (the machine locale/language) was chosen. This PR changes the default behavior such that, if the metadatalanguage isn't chosen in an ancestor collection, the user is allowed/forced (because there is no 'none' option) to chose a language at dataset creation. (This is only if the metadata language setting is configured in the first place.) Thus, this allows a use case where a collection/repository manager can let users decide which language to use and have collections that include datasets using different metadata languages.

Closes #8422

**Suggestions on how to test this**: Configure metadatalanguages, verify that the dataverse/edit/general entry for metadata language includes the new value 'Chosen at Dataset Creation' and that choosing that on the root or child dataverse triggers a change in the dataset creation page to allow a user to chose the metadata language (versus showing the predefined choice from the collection level) -see image below.

The change in the dataset page, only if metadatalanguage is enabled and only in collections where a choice hasn't been made is for the metadata language to be a live input:
image

**Additional documentation**: Change in default behavior note in the config docs

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -241,7 +241,7 @@ As for the "Remote only" authentication mode, it means that:
 - The "builtin" authentication provider has been disabled (:ref:`api-toggle-auth-provider`). Note that disabling the "builtin" authentication provider means that the API endpoint for converting an account from a remote auth provider will not work. Converting directly from one remote authentication provider to another (i.e. from GitHub to Google) is not supported. Conversion from remote is always to "builtin". Then the user initiates a conversion from "builtin" to remote. Note that longer term, the plan is to permit multiple login options to the same Dataverse installation account per https://github.com/IQSS/dataverse/issues/3487 (so all this talk of conversion will be moot) but for now users can only use a single login option, as explained in the :doc:`/user/account` section of the User Guide. In short, "remote only" might work for you if you only plan to use a single remote authentication provider such that no conversion between remote authentication providers will be necessary.
 
 File Storage: Using a Local Filesystem and/or Swift and/or object stores
----------------------------------------------------------------------------
+------------------------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


landreev (commented):
All makes sense.

landreev (approved):

landreev:
(I have approved the PR, but please take a look at the failed integration test. It looked like it was something outside the PR that killed ansible, but please double-check).

PR #8447, opened 2022-02-22
API importddiwithjson

### LINKED ISSUE #8210: Native API: Import ddi dataset does not appear to be working, fails on parsing subject.
There is also an issue opened to provide a working sample ddi in the guides, #8209 .

For now, until proven otherwise, it appears this endpoint isn't working properly. The json endpoint does.

One hypothesis is that a pr that enforced required fields revealed a prior unknown issue in ddi import parsing, showing up as a null subject error.

There are server log errors, only a command line error: {"status":"ERROR","message":"Validation Failed: Subject is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ])."}

Please note there are two paths for testing: with a preexisting, published pid and one without so that the system generates one. The subject error happens when you use the ddi export from a published dataset on another system, ie. demo. The failure for a ddi without a doi is less clear to me: {"status":"ERROR","message":"Invalid file content: ParseError at [row,col]:[17,6]\nMessage: Element content can not contain child START_ELEMENT when using Typed Access methods"}

lubitchv:
importddi API does not work at present since during dataset creation it tries to validate required fields but DDI (xml) does not have fields that Dataverse deems to be required, for example "Subject". DDI does not have such field. DDI has keywords but they correspond to dataverse keywords.

There is also sometimes a situation when XML files do not have all the fields that user wants to import to Dataverse, in this case editing of XML is needed before import or updating metadata after import.

This API tries to resolve both situations. API submits xml file and json file. Json file consists of missing fields, for example it can be Subject, Language etc. Json file is similar to a json file for edit dataset metadata API   https://guides.dataverse.org/en/latest/api/native-api.html#edit-dataset-metadata

Resolves #8210

Integration test DataversesIT#testImportDDI is introduced.

The API can be checked with curl:
``curl -H X-Dataverse-key:$API_TOKEN -X POST $SERVER_URL/api/dataverses/$DATAVERSE_ID/datasets/importddiwithjson?replace=true -F 'xml=@src/test/resources/xml/example.xml' -F 'jsonData=@src/test/resources/json/missing-fields.json'``

API has following optional parameters:

replace =true/false  - if true then  replace fields of xml by json fields if the field is in both and creates dataset. If false then add missing files from json to xml and creates dataset. If field exists in both then error is shown.

release=true/false if true then after creation, dataset is published.

pid - import dataset with existing persistentId 

No


pdurbin:
Not sure why the Jenkins job failed. https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-8447/1/consoleFull ended with this:

```
[JaCoCo plugin] Loading packages..
[JaCoCo plugin] Done.
[JaCoCo plugin] Overall coverage: class: 37.92017, method: 21.559065, line: 18.84934, branch: 15.647721, instruction: 19.508333, complexity: 15.2561245
[Pipeline] script
[Pipeline] {
[Pipeline] fileExists
[Pipeline] error
[Pipeline] }
[Pipeline] // script
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: Ansible run terminated abnormally, failing build.

GitHub has been notified of this commit’s build result

Finished: FAILURE
```

Also, I added this to Review so someone can take a look. Thanks for the pull request, @lubitchv !

pdurbin:
@lubitchv @scolapasta @qqmyers @landreev thanks for all the discussion around this pull request today.

Please see https://github.com/IQSS/dataverse/issues/8210#issuecomment-1064406141 for the direction we're planning. In short, we're closing this pull request that adds a DDI+JSON API endpoint and focusing on fixing the broken `importddi` endpoint. Existing JSON endpoints can be used afterward to add Subject or other metadata fields. Thanks @lubitchv for taking this on!

PR #8455, opened 2022-03-01
Validate URLs before saving as metadata

eunices:
Validate URLs before saving as metadata.

NA

None

Add `http://google.com test` to Alternative URL when editing metadata. It should not work.
Only https://, http:// and ftp:// allowed.

No

No

**Additional documentation**:
None

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
@@ -19,6 +19,7 @@ import java.util.regex.Pattern;
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.validator.routines.UrlValidator;
 
 /**
  *
@@ -168,7 +169,20 @@ public class DatasetFieldValueValidator implements ConstraintValidator<ValidateD
 
         if (fieldType.equals(FieldType.URL) && !lengthOnly) {
             try {
-                URL url = new URL(value.getValue());
+                
+                String[] schemes = {"http","https"};
+                UrlValidator urlValidator = new UrlValidator(schemes);
+                
+                String urlString = value.getValue();
+                URL url = new URL(urlString);
+                
+                if (urlValidator.isValid(urlString)) {
+                    return true;
+                } else {
+                    context.buildConstraintViolationWithTemplate(dsfType.getDisplayName() + " " + value.getValue() + "  is not a valid URL.").addConstraintViolation();
+                    return false;
+                }
+                
             } catch (MalformedURLException e) {
                 try {
                     context.buildConstraintViolationWithTemplate(dsfType.getDisplayName() + " " + value.getValue() + "  is not a valid URL.").addConstraintViolation();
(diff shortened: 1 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
```diff
@@ -168,7 +169,20 @@ public boolean isValid(DatasetFieldValue value, ConstraintValidatorContext conte
 
         if (fieldType.equals(FieldType.URL) && !lengthOnly) {
             try {
-                URL url = new URL(value.getValue());
+                
+                String[] schemes = {"http","https"};
```
qqmyers, line 173: Other than being less popular, is there any reason ftp URLs shouldn't be allowed?

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
```diff
@@ -168,7 +169,20 @@ public boolean isValid(DatasetFieldValue value, ConstraintValidatorContext conte
...
+                UrlValidator urlValidator = new UrlValidator(schemes);
+                
+                String urlString = value.getValue();
+                URL url = new URL(urlString);
```
qqmyers, line 177: FWIW: Not sure if this and the catch of a MalformedURLException is still needed. The urlValidator class tries to create a java URI and returns false if that gives a MalformedURLException

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
```diff
@@ -19,6 +19,7 @@
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.validator.routines.UrlValidator;
```
qqmyers, line 22: There's a similarly named edu.harvard.iq.dataverse.URLValidator class that looks like it was used in bean validation (pre v4.13) and in a couple places to validate URLs from settings. I don't see any place where it it used for displayed URLs. Never-the-less, seems like using the same Apache UrlValidator class there makes sense.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
eunices: I see, removed MalformedURLException.

eunices: Indeed, removed `URL url = new URL(urlString)`.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
eunices: Nope, have added ftp in.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java
eunices: Ok, have updated URLValidator to use org.apache.commons.validator.routines.UrlValidator, see 6de1e959555c315f79826e8ab86796a696aa49ee. It passed the corresponding unit test.

eunices:
I've addressed your comments @qqmyers sufficiently, I hope!

sekmiller (approved):
Looks good. And I checked in with Jim regarding his previous comments. He agrees that it's ready to move.

PR #8475, opened 2022-03-09
Iqss/8474 poi/tika updates

qqmyers:
**What this PR does / why we need it**: This is an alternative for #8474 that also updates the other poi components to the same version, and also updates tika which uses poi. Includes using properties in the pom to get the same version for all poi components, and the same version for tike core and parsers.

Closes #

**Special notes for your reviewer**: See #8474 for comments. The tika update is possibly separable but tika 2.3.0 is the first to use poi 5.2.1. The one potential concern is that tika updates to log4j 2.17.1 (the old 1.x tika line used log4j v1.x). If we stay away from tika 2.x we should at least update to v1.28.1. The other minor tweaks in the PR are just replacing bad/old imports of IOUtils from tika instead of apache commons, etc.

**Suggestions on how to test this**: Since the poi library requires code changes in the XSLXFileReader, I'd suggest testing some xslx ingests to make sure it still works. Tika is used (with poi) in full-text indexing so it may also be worth turning full-text indexing on and making sure nothing explodes/ basic searches for full-text entries work.
There is also a minor refactor /api/datasets/locks command where the old class for concatenating lock names with ", " between them is replaced (the old class was only available through the old versions of poi/tika). You could run that api with various numbers of locks to test that the new string concatenation works. (Not sure the best way to force locks - possibly just db edits?) 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: no - back-end libraries.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2691,9 +2689,12 @@ public Response listLocks(@QueryParam("type") String lockType, @QueryParam("user
...
                 lockTypeValue = DatasetLock.Reason.valueOf(lockType);
             } catch (IllegalArgumentException iax) {
-                String validValues = Strings.join(",", DatasetLock.Reason.values());
+                StringJoiner reasonJoiner = new StringJoiner(", ");
+                for (Reason r: Reason.values()) {
+                    reasonJoiner.add(r.name());
+                };
                 String errorMessage = "Invalid lock type value: " + lockType + 
-                        "; valid lock types: " + validValues;
+                        "; valid lock types: " + reasonJoiner.toString();
```
pdurbin, line 2697: This refactoring seems completely unrelated to poi/tika updates. @qqmyers can you please either revert this or explain under "how to test" how to exercise this code? Thanks.

pdurbin (commented):
I didn't run the code but the changes make sense with the exception of an off topic refactoring. I put more details in a comment.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: FWIW: import com.beust.jcommander.Strings; was only available through the older versions of poi/tika

pdurbin: Ah, makes more sense now. Thanks for explaining and for the update to "how to test".

pdurbin (approved):
I didn't run the code but the changes make sense.

PR #8477, opened 2022-03-09
Resolved ClassCastException

janvanmansum:
**What this PR does / why we need it**:  A `ClassCastException` was thrown when casting the root bean of a constraint violation to `DatasetField` in cases where the root bean was actually a `FileMetadata` object. This happened in the logical that sets `N/A` as the field value if there are constraint violations for that field. Obviously that logic doesn't make sense for files, so the solution is to only execute it if the root bean of the contraint is an instance of  `DatasetField`.

The `ClassCastException` obscured the problem. The server returned `Failed to add file to dataset.` but no explanation. 

**Which issue(s) this PR closes**: See previous item

Closes N/A

**Special notes for your reviewer**: N/A

**Suggestions on how to test this**: Try to create a file with invalid characters in its name of directoryLabel. This will trigger a constraint violation that has a `FileMetadata` as root bean, which should now be filtered out of above mentioned logic.


sekmiller (approved):
Looks good. I was working on this method for another issue and added the same logic.

PR #8488, opened 2022-03-15
IQSS/dataverse#7895 prevent clearing fields in dataset form when hitting enter

### LINKED ISSUE #7895: Hitting enter in text fields triggers metadata form save
In at least 5.4.1+ - hitting enter when editing a single-line text metadata field results in the whole from being saved. In create mode/when required fields are not filled out, this appears to result in a failed save with the validation message appearing (as it does if you hit the Save button). (tested on v5.5 demo.dataverse.org and on a 5.4.1 test instance. I did not check earlier versions.)

I'm not sure how/when this started but I'd suspect some change that either added a handler for those fields or some change to the Javascript being run (returning false stops default handlers from running, so if some change stopped a 'return false;' it might have triggered this issue.)

haarli:
This pull request closes #7895. 
It adds a dummy button with no functionality to the beginning of the dataset form. This way, when pressing enter in a single text field of the form, nothing happens. Before, all fields in the form were cleared, which is an undesired behaviour.


sekmiller (commented):
This looks like a good approach and I've found that it works for Chrome and Firefox, but it doesn't work for Safari. A quick look on my part didn't yield a solution for Safari.

haarli:
Thanks for the review, didn't check with Safari.

Safari seems to ignore submit buttons with "display=none" when it comes to hitting the enter button inside the form (same goes for buttons with the HTML5 attribute "hidden"). Firefox and Chrome do not.

So even without this PR, there's currently a different behaviour when hitting enter in any single-line text field in the datasetForm. And there's also a difference between creating a new dataset and editing:

#### Current behaviour
* Create a new dataset:
  * Firefox/Chrome:
All input is cleared, because the hidden button for changing the owner dataverse is internally fired (As it is the first submit button inside the form)
  * Safari:
The "Save"-button is fired, as all other submit buttons are hidden via "display:none"
* Edit dataset metadata:
Here,  a save button is additionally displayed at the top of the form. So this one is always fired, independent of the browser.

#### Behaviour with PR
Clearing all input is worst case. So with this PR, in Firefox and Chrome nothing would happen when pressing enter (for create and edit). In Safari, "save" would be triggered (for create and edit).
Not perfect at all and still inconsistent, but maybe better than the current situation (clearing all fields).

sekmiller:
I found this bit of jquery which seems to work on Firefox, Chrome and Safari:`
                    $(document).on("keydown", "form", function(event) { 
                        return event.key !== "Enter";
                    });`

haarli:
The problem with this JS snippet is that it prevents any usage of the enter button in the form, e.g. newlines in the description field wouldn't be possible anymore.

haarli:
Thanks for the alternative solution. Looks good.

sekmiller:
I'm going to close this in favor of Erik's solution.  Thanks for your contribution.

PR #8495, opened 2022-03-16
Installer fixes, reflecting the new pom file structure #8494

### LINKED ISSUE #8494: Vagrant Deployment not working: 'can't seem to find an appropriate warfile.'

* When does this issue occur?
When trying to deploy dataverse using vagrant

N/A

* What happens?
Running `vagrant up` starts the deployment process, but after configuring Solr the process fails because the python installer is unable to find a .war file:
```
==> default: Running provisioner: shell...
    default: Running: /tmp/vagrant-shell20220316-16786-1agei0k.sh
    default: enabling non-interactive mode
    default: 
    default: This appears to be a RedHat system; good.
    default: Checking for required components...
    default: Sorry, I can't seem to find an appropriate warfile.
    default: Are you running the installer from the right directory?
```

Deployment using `vagrant up` should be successful

5.9

Not that I could find

**Screenshots:**

*

j-n-c:
Fix python installer search for dataverse .war file. Currently deploying dataverse with vagrant is not working.
[from @landreev:]  The PR now includes the fixes for the old school (perl) installer, and the installer Makefile as well. 

Closes #8494

Check if the revision field in modules/dataverse-parent/pom.xml is a suitable location for retrieving dataverse version

[from @landreev:] It should now be possible to make the installer bundle; and run the installer successfully from the source tree. 
In other words, the QA for this PR is one of the standard steps of making a release. This means this issue would be found eventually - but this PR removes the element of surprise from the process. 

**Additional documentation**:

### DIFF
--- a/scripts/installer/install
+++ b/scripts/installer/install
@@ -277,7 +277,7 @@ unless ($postgresonly)
 
     unless ( -f $WARFILE_LOCATION ) {
 	my $DATAVERSE_VERSION = "";
-	my $DATAVERSE_POM_FILE = "../../pom.xml";
+	my $DATAVERSE_POM_FILE = "../../modules/dataverse-parent/pom.xml";
 	if ( -f $DATAVERSE_POM_FILE ) 
 	{
 	    open DPF, $DATAVERSE_POM_FILE; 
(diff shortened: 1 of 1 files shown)


landreev:
I'd be happy to add the perl installer fix to this PR.

landreev:
Of course the makefile that builds the installer bundle is also broken, for the same very reason. I can add a fix for that too.

PR #8500, opened 2022-03-17
7895 hitting enter in dataset form is disabled in the input text fields that do not use vocabulary.

ErykKul:
This pull request closes https://github.com/IQSS/dataverse/issues/7895.

The enter is only disabled in the input text fields that do not use vocabulary.


sekmiller (approved):
Looks good. Thank you for your contribution!

PR #8507, opened 2022-03-18
8506 powered by footer

### LINKED ISSUE #8506: Dataverse footer with "powered by dataverse" logo gets rendered at relative position iso absolute by default
Use a custom footer via "/CustomizationFilesServlet?customFileType=footer" and use de default style as defined in the custom-footer.html example. Open a page that does not fit in to the screen and the "powered by" footer gets rendered at wrong position (button of the screen iso bottom of the document) and stays there when scrolling.

* When does this issue occur?
With overloaded custom footer with default style block and on pages that do not fit on screen.

All users using default footer structure: footer element with "dvfooter" and "custom-footer" inside, and with default style block.

Footer gets rendered at the bottom, for example like this:
image

Develop branch.

Did not find any.

**Screenshots:**
image

ErykKul:
Dataverse footer with "powered by dataverse" logo gets rendered at relative position iso absolute by default. This happens with overloaded custom footer with default style block and on pages that do not fit on screen. See the linked issue for more detail.

closes #8506


src/main/webapp/resources/css/structure.css
```diff
@@ -49,7 +49,7 @@ body .ui-widget {font-size: inherit;}
 #status-alert {margin-top:0; margin-bottom:0;}
 #status-alert div.alert {border:0; box-shadow:none;}
 
-#dvfooter {position:absolute; bottom:0; width:100%; height:60px; padding-bottom:100px; color:#767676;}
+#footer {position:absolute; bottom:0; width:100%; height:60px; padding-bottom:100px; color:#767676;}
```
pdurbin, line 52: Hmm, so we'd have one line in the CSS file say `footer` and the rest say `dvfooter`? Seems weird.

I don't know the history of all this but I see @qqmyers changed `footer` to `dvfooter` in 734bd63 as part of pull request #8322. At the same time should we have changed `footer` to `dvfooter` in `doc/sphinx-guides/source/_static/installation/files/var/www/dataverse/branding/custom-footer.html`? Is changing that `custom-footer.html` file to `dvfooter` another possible solution?

pdurbin (commented):
@ErykKul thanks for the pull request! I left some questions about `footer` vs. `dvfooter`.

src/main/webapp/resources/css/structure.css
ErykKul: What I see in a rendered page is that footer, dvfooter an custom-footer all exist:
image

That structure is defined in `src/main/webapp/dataverse_footer.xhtml`

I do not understand the intended logic behind the footer. I would assume that it is the footer that should have the absolute position, while the dvfooter and the custom-footer should have relative positions inside the footer. I think that the sturcture.css should describe both, the footer and the dvfooter, while the custom-footer.html should contain the custom-footer together with the style block for the custom-footer only.

qqmyers: FWIW: It's possible that I just missed/incorrectly updated this line. The name changes were related to accessibility - using <footer> so there's one Footer page region. That change moved the custom footer inside the top level <footer> where in v5.9 the custom footer was after the id="footer" element. 

One place that is using a custom footer that would want to test this change is DANS - (heads-up @janvanmansum). 

My general sense is that this line should now be for footer (not #footer) now, but as @ErykKul argues - before my change the element with id=footer was the top level element and the style was applied to it, so now that <footer> is at the top, the style should apply there. The three lines below do still see like they are intended only for the default header (now #dvfooter). 

As for the example custom-footer file - I don't think I saw that at all when making changes so I agree in general that it's style should not be affecting <footer> at this point, but I haven't looked at what makes sense there.

pdurbin: Thanks for all the discussion. I ran the branch locally and the fix works. I'm happy with it. If someone wants to dig in, they can refer to this pull request.

pdurbin (approved):
Seems like a good fix to me. It could always be refactored later, if necessary. Approved.

I ran the code locally and will post my before and after below.

## before (note Dataverse logo)

Screen Shot 2022-03-21 at 4 15 31 PM

## after (logo in a more reasonable position)

Screen Shot 2022-03-21 at 4 16 03 PM

ErykKul:
Thank you for approving! I totally agree that there must be a better solution. I do not have the necessary background on css to improve it by myself, I have tried different options and this one works for me.

pdurbin:
@ErykKul just a heads up that we made a further tweak to this:

- PR #8565.

Issue #8515, opened 2022-03-21
File upload javascript was giving errors for not unique identifier


ErykKul:
**What steps does it take to reproduce the issue?**
Go to a dataset and try to upload a file. Open the development console in your browser; you can see the error about not unique identifier

* Which page(s) does it occurs on?
Upload data fragment.

* What happens?
Error in the console.

**Which version of Dataverse are you using?**
Develop

ErykKul:
![image](https://user-images.githubusercontent.com/101262459/159237375-d006e085-38ab-4156-acd4-a864eced1b8c.png)

pdurbin:
Resolved by PR #8516, closing.

PR #8499, opened 2022-03-17
Minor bug fixes, including fix for null pointers in the new license selector that did not work with licenses without image URI.

ErykKul:
**What this PR does / why we need it**:
I am new for this project and I have run against some problems when trying out the development branch locally. More in particular some validation errors in NetBeans, as well as some errors when running the application and/or tests. I have fixed them and thought to share it with you.

**Which issue(s) this PR closes**:
I did not see any specific issues reported for the problems I did run against:
- License selector did not work for licenses without image URI (null pointer exceptions)
- DDIExporterTest.java crashed on unmarshalling of LocalDateTime class
- NetBeans complained on namespace declarations in some of the xhtml files
- File upload javascript was giving errors for not unique identifier

Closes #

**Special notes for your reviewer**:
- I am very new to this project, started one week ago, I might have missed something, etc.

**Suggestions on how to test this**:
- Uploading licenses without image URL and trying to select them in license selector.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
- No.

**Is there a release notes update needed for this change?**:
- No.

PR #8516, opened 2022-03-21
fixed not unique identifier error and syntax error in the catch block…

ErykKul:
… in fileupload.js

https://github.com/IQSS/dataverse/issues/8515


ErykKul:
This commit was built in the https://github.com/IQSS/dataverse/pull/8499
I do not understand why it is not building anymore. When I try to view the details, I get error that jenkins is not reachable.

scolapasta (approved):

PR #8517, opened 2022-03-21
fixed unmarshaling van LocalDateTime error in the DDIExporterTest.java

### LINKED ISSUE #8514: DDIExporterTest.java crashed on unmarshalling of LocalDateTime class
 Run the test

* What happens?

image

Develop

ErykKul:

- closes #8514

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
@@ -1,6 +1,10 @@
 package edu.harvard.iq.dataverse.export;
 
-import com.google.gson.*;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
 import edu.harvard.iq.dataverse.ControlledVocabularyValue;
 import edu.harvard.iq.dataverse.DatasetFieldType;
 import edu.harvard.iq.dataverse.DatasetFieldType.FieldType;
--- a/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
@@ -1,6 +1,6 @@
 package edu.harvard.iq.dataverse.export;
 
-import com.google.gson.Gson;
+import com.google.gson.*;
 import edu.harvard.iq.dataverse.ControlledVocabularyValue;
 import edu.harvard.iq.dataverse.DatasetFieldType;
 import edu.harvard.iq.dataverse.DatasetFieldType.FieldType;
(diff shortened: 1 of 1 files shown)


src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
```diff
@@ -1,6 +1,6 @@
 package edu.harvard.iq.dataverse.export;
 
-import com.google.gson.Gson;
+import com.google.gson.*;
```
pdurbin, line 3: For a long time we've been discouraging wildcard imports but there was some recent discussion on reconsidering this position. Nothing is written at https://guides.dataverse.org/en/5.10/developers/coding-style.html either way. 😄

src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
```diff
@@ -42,7 +47,10 @@
     private static final SettingsServiceBean settingsService = Mockito.mock(SettingsServiceBean.class);
     private static final LicenseServiceBean licenseService = Mockito.mock(LicenseServiceBean.class);
     private static final MockDatasetFieldSvc datasetFieldTypeSvc = new MockDatasetFieldSvc();
-    private static final Gson gson = new Gson();
+    private static final Gson gson = new GsonBuilder().registerTypeAdapter(LocalDate.class, (JsonDeserializer<LocalDateTime>) (JsonElement json, Type type, JsonDeserializationContext jsonDeserializationContext) -> {
+        Instant instant = Instant.ofEpochMilli(json.getAsJsonPrimitive().getAsLong());
+        return LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+    }).create();
```
pdurbin, line 53: I ran this test on my laptop and it works fine. It seems like a lot of boilerplate code.

I noticed that the `InaccessibleObjectException` error was first reported by @poikilotherm in an issue about Java 17:

- https://github.com/IQSS/dataverse/issues/8094#issuecomment-1022101843

@ErykKul are you running something newer than Java 11? In that issue, @poikilotherm suggests at upgrading Gson might fix this and what is perhaps a related issued here:

- https://github.com/google/gson/issues/1875

I can't reproduce the problem on Java 11 and haven't tried on anything newer. We aren't actively working on that Java 17 issue above but this pull request seems like a step in the right direction. I'm wondering, however, can you please try upgrading Gson to see if that's also a suitable fix?

pdurbin (commented):
I can't reproduce the issue on Java 11. If the goals is to support Java 17, I'm wondering if upgrading Gson is cleaner (less boilerplate). @ErykKul can you please give this a try? I left a few more comments in my review. Thanks!

src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
ErykKul: I have tried upgrading gson to 2.9.0, but it did not solve the problem. Downgrading to jdk11 helped, the test passes and in stead of errors I am getting warnings (with both gson versions; 2.9.0 and 2.2.4):
image

PR #8555, opened 2022-03-30
Iqss/8553 fixrequest access to files in datasets with custom terms

### LINKED ISSUE #8553: Request access broken for datasets with custom terms
Discovered at QDR:
https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/GP5K62&version=1.1 - try to request access for the first file. The popup won't appear. 
* What happens?
There is a null pointer in the log. This was fixed in the requestdownload popup, but not for the request access popup.

5.10

PR to follow.

qqmyers:
**What this PR does / why we need it**: The request access popup logic failed with a null pointer if license=null-> custom terms were used. This issue was caught/fixed for the download popup but missed here. The PR includes a refactor so that both methods now use common code.

Closes #8553

**Special notes for your reviewer**: As far as I could tell, prior to the multi-license code, the isDownloadPopupRequired and isRequestAccessPopupRequired methods used the same code except for isDownload doing a final check for guestbooks. With multi-license it looks like they were changed independently at different times (with isRequestAccess being broken). They are now the same again (except the final guestbook check). If there was some other intended change between the two of them, this PR would be removing it.

Is this a basic enough issue to push a 5.10.1 or 5.11 in the short term?

**Suggestions on how to test this**: Use custom terms on some dataset, add a restricted file(s), publish, login as someone else and verify that you get the request access popup. Could also check for regression (does the download popup appear as before), does everything still work with a standard license, ...). FWIW: There are tests in FileUtil for isDownloadPopupRequired that are passing as before.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
@@ -1523,9 +1523,8 @@ public class FileUtil implements java.io.Serializable  {
                 logger.fine("Popup required because of terms of access.");
                 answer = true;
             }
-            return answer;
-
         }
+        return answer;
     }
 
     /**
--- a/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
@@ -1495,17 +1495,22 @@ public class FileUtil implements java.io.Serializable  {
         return false;
     }
     
+    /* Code shared by isDownloadPopupRequired and isRequestAccessPopupRequired.
+     * 
+     * Returns Boolean to allow null = no decision. This allows the isDownloadPopupRequired method to then add another check w.r.t. guestbooks before returning its value.
+     * 
+     */
     private static Boolean popupDueToStateOrTerms(DatasetVersion datasetVersion) {
 
         // Each of these conditions is sufficient reason to have to
         // present the user with the popup:
         if (datasetVersion == null) {
-            logger.fine("Popup required because datasetVersion is null.");
+            logger.fine("Popup not required because datasetVersion is null.");
             return false;
         }
         // 0. if version is draft then Popup "not required"
         if (!datasetVersion.isReleased()) {
-            logger.fine("Popup required because datasetVersion has not been released.");
+            logger.fine("Popup not required because datasetVersion has not been released.");
             return false;
         }
         // 1. License and Terms of Use:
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1471,73 +1471,59 @@ public static String getCiteDataFileFilename(String fileTitle, FileCitationExten
...
+        return false;
+    }
+    
+    private static Boolean popupDueToStateOrTerms(DatasetVersion datasetVersion) {
+
+        // Each of these conditions is sufficient reason to have to
+        // present the user with the popup:
         if (datasetVersion == null) {
-            logger.fine("Download popup required because datasetVersion is null.");
+            logger.fine("Popup required because datasetVersion is null.");
```
sekmiller, line 1503: If we're returning false here then the logger should say 'Popup not required...',

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1471,73 +1471,59 @@ public static String getCiteDataFileFilename(String fileTitle, FileCitationExten
...
             if (!(datasetVersion.getTermsOfUseAndAccess().getTermsOfAccess() == null) && !datasetVersion.getTermsOfUseAndAccess().getTermsOfAccess().equals("")) {
-                logger.fine("Download popup required because of terms of access.");
+                logger.fine("Popup required because of terms of access.");
                 return true;
             }
         }
-
-        logger.fine("Download popup is not required.");
-        return false;
+        return null;
```
sekmiller, line 1526: Found the use of null here a little confusing, but looking at the rest of the code I see that it means "leave it up to Guestbook if applicable"

sekmiller (changes requested):
A couple of minor issues with the logging and maybe a comment on the meaning of null would be helpful. Overall makes sense and should fix the underlying issue.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
qqmyers: Added a method header comment and a comment at this line.

sekmiller (approved):

sekmiller (approved):
Looks good, passing to QA

EIARDataverses:
thank you

PR #8558, opened 2022-03-30
remove NONE as a license option for SWORD, show available licenses #8551

### LINKED ISSUE #8551: As of 5.10, SWORD API no longer accepts NONE for dcterms:license

- Download `atom-entry-study.xml` from https://guides.dataverse.org/en/5.10/api/sword.html#create-a-dataset-with-an-atom-entry
- Try to create a dataset using the curl command from the URL above.
- Observe that dataset cannot be created.

Here is the error:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<sword:error xmlns="http://www.w3.org/2005/Atom" href="Couldn't find an active license with: NONE" xmlns:sword="http://purl.org/net/sword/terms/">
  <title>ERROR</title>
  <updated>2022-03-29T19:00:40Z</updated>
  <generator uri="http://www.swordapp.org/" version="2.0"/>
  <summary>Couldn't find an active license with: NONE</summary>
  <sword:treatment>Processing failed</sword:treatment>
</sword:error>
```

I expect the dataset to be created.

To be honest, I'm not sure what I expect the dataset to have as a license when I specify `NONE`. We should probably check what happens in Dataverse 5.9.

At the very least, the XML example in the API Guide should work.

It's possible that SWORD-based integrations like OJS and OSF are broken.

5.10

- #7920

pdurbin:

The example XML in the SWORD docs doesn't work due to licenses changes in 5.10 and PR #7920.

- Closes #8551

I added some minimal testing of interactions between license and rights. We could add more.

I'm not 100% sure of all the assertions in the SWORD page of the API Guide. We could add more tests and/or edit the guide as needed.

Generally speaking, I'm thinking we should drop NONE as an option, even if it's an incompatible change. If we want to keep NONE, what should it map to? Custom Dataset Terms? But they can't be blank, right?

- Try passing NONE. You should now see a list of possible licenses.
- Try passing a valid license.
- Test all the assertions in the SWORD section of the API Guide about license, rights, etc.

No.

Maybe?

**Additional documentation**:

None.


src/main/java/edu/harvard/iq/dataverse/api/datadeposit/SwordServiceBean.java
```diff
@@ -186,7 +187,13 @@ public void setDatasetLicenseAndTermsOfUse(DatasetVersion datasetVersionToMutate
             setTermsOfUse(datasetVersionToMutate, dcterms, null);
         } else {
             License licenseToSet = licenseServiceBean.getByNameOrUri(licenseProvided);
-            if (licenseToSet == null) throw new SwordError("Couldn't find an active license with: " + licenseProvided);
+            if (licenseToSet == null) {
+                List<String> licenses = new ArrayList<>();
+                for (License license : licenseServiceBean.listAllActive()) {
+                    licenses.add(license.getName());
+                }
+                throw new SwordError("Couldn't find an active license with: " + licenseProvided + ". Valid licenses: " + licenses);
```
qqmyers, line 195: I think you also have to check that licenseToSet.isActive()  - if it is non-null but not isActive, you should get this same error.

doc/sphinx-guides/source/api/sword.rst
```diff
@@ -143,7 +143,7 @@ Dublin Core Terms (DC Terms) Qualified Mapping - Dataverse Project DB Element Cr
...
-|dcterms:license              |                   license                    |              |  Set the license to CC0 (default in a Dataverse installation for new Datasets), otherwise enter "NONE" and fill in the dcterms:rights field.                |                
+|dcterms:license              |                   license                    |              |  Set the license. Alternatively, use the dcterms:rights field instead.                                                                                      |
 +-----------------------------+----------------------------------------------+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------+
 |dcterms:rights               |                 termsofuse                   |              |  If not using CC0, enter any terms of use or restrictions for the Dataset.                                                                                  |
```
qqmyers, line 148: 'If not using CC0' ->'If not specifying a license' or similar?

qqmyers (changes requested):
Looks OK - didn't run. 

If I understand, the only main code change is to add the list of active licenses in the response. 
My comment there suggests adding a test to make sure the request License is active, which would be a functionality change.

The other changes fix the example, add good tests, update docs. I saw one more CC0 that could be changed but it all makes sense.

src/main/java/edu/harvard/iq/dataverse/api/datadeposit/SwordServiceBean.java
pdurbin: Ah, good point. Fixed in 66eabc0. Thanks.

qqmyers (approved):

PR #8454, opened 2022-02-28
8127 citation field improvements

jggautier:
**What this PR does / why we need it**:
This PR makes grammar, style and consistency improvements to the titles, descriptions/tooltips and watermarks of all metadata fields that appear in the Citation metadata block.

This includes edits to fields that appear in the Citation Metadata accordions (controlled by the citation.tsv and citation.properties files) and whose values are system-generated, like the Dataset Persistent ID, Previous Dataset Persistent ID, and Publication Date fields (controlled by the bundles.properties file).

The changes should provide clearer information to curators, depositors, and people looking for data about what the fields are for.

It also adds a new page in the [Style Guides](https://guides.dataverse.org/en/latest/style/index.html) called "Text" that includes a section called "Metadata Text Guidelines", which includes a link to the Google Doc where the guidelines are being maintained for now since we expect them to be revised frequently.

**Which issue(s) this PR closes**:
Closes #8127 

**Special notes for your reviewer**:

**Suggestions on how to test this**:
Reload the Citation metadatablock. See the [Reloading a Metadata Block](https://guides.dataverse.org/en/latest/admin/metadatacustomization.html#reloading-a-metadata-block) section of the Admin guide.

Create and edit a dataset to spot check that the titles, descriptions/tooltips and watermarks of fields shown on the following pages have changed:
- Dataset page when creating, editing and viewing a dataset
- Search page facet categories
- Advanced search page
- Dataverse edit page (in the Metadata Fields and Browse/Search Facets sections)

The spreadsheet at https://docs.google.com/spreadsheets/d/1sHRvSpGSF0XfKTp4nWxT0zX4z8TORcuX7DTDrktOorg shows what the field titles, tooltips and watermarks were and, if they've been changed, how they've been changed. I wouldn't suggest checking all changes, since every description/tooltip has been edited. I hope checking a few on each page would be enough to spot any problems.

Check the Style Guide to make sure its Contents list includes a link to the "Text" page and check that clicking on that link takes you to the Text page, which should include a "Metadata Text Guidelines" section.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Yes, the mockups below show some of the changes to the field titles and watermarks. The tooltips are shy (they appear only one at a time when hovering over or touching them) so it's tougher to show their changes in the same way, but all of the changes can be seen in the spreadsheet at https://docs.google.com/spreadsheets/d/1sHRvSpGSF0XfKTp4nWxT0zX4z8TORcuX7DTDrktOorg.

[...]

Issue #8127, opened 2021-10-01
Review and improve metadata tooltips, field labels and watermarks


TaniaSchlatter:
This issue is to outline and track the effort to improve tooltips, field labels and watermarks related to metadata blocks. Clear and consistent tip text, labels and watermarks should help depositors curate their datasets and enter data accurately, leading to more and more consistent metadata. 

**Goals:**  
- Improve the collection of dataset metadata 
- Define and document rules for adding or editing metadata tip text, field labels and watermarks

**Scope/definition of done:**
Review metadata tooltip text, field labels and watermarks for consistency and clarity: citation block, social sciences block, journals, geospatial, astrophysics 
- [x] define[ "working rules" to guide grammar and style/consistency](https://docs.google.com/document/d/1tY5t3gjrIgAGoRxVMWQSCh46fnbSmnFDLQ7aLkNLhJ8) improvements to tip text, form labels and watermark text
- [x] apply the working rules to existing tooltips, field labels and watermarks. [Drafting proposed edits](https://docs.google.com/spreadsheets/d/1sHRvSpGSF0XfKTp4nWxT0zX4z8TORcuX7DTDrktOorg/edit#gid=0) for review
- [x] document opportunities for additional improvements, such as form layout and behavior. These could be broken off as separate issues
- [x] review proposed text changes with metadata/topic experts
- [x] communicate with the community about changes (affects languages/translation)
- [x] update tooltip text, field labels and watermarks in Dataverse
- [x] evaluate changes
- [ ] complete & share the style guide

djbrooke:
Thanks @TaniaSchlatter @scolapasta @jggautier for discussing earlier. The above checklist remains accurate.

- @scolapasta @jggautier and @djbrooke will meet to get a shared understanding of how tool tips are managed and updated for those installations that have localized and those that have not. In the meeting, we should also look at how to manage tool tips for system generated content (such as the dataset PID). We're planning on meeting Thursday at 1 PM.
- Groups that were mentioned for follow up were DDI Alliance, Scholar's Portal/Portage, and the curation team at the Harvard Dataverse Repository. @jggautier has a list as well. We will do targeted outreach to these groups and also general outreach to the community.

scolapasta:
@jggautier I would expect those changes to just have to be made to the bundle.

jggautier:
That worked, thanks!

jggautier:
[...]

PR #8592, opened 2022-04-09
IQSS/8533 Update Internal Semantic Mappings

### LINKED ISSUE #8533: Dataset Migration API: Cannot migrate a dataset using example json-ld file, complains about Contact E-mail being null though it is specified. (testSemanticMetadataAPIs)
Update:
The cause was a change to the citation.tsv file where the Contact field name has changed to Point of Contact. If I manually edit the example file to use that, it works. So the fix would be to update the example file to use the new name:
Change: 
        "citation:Contact": {
                "datasetContact:Name": "Admin, Dataverse",
                "datasetContact:Affiliation": "GDCC",
                "datasetContact:E-mail": "admin@demo.dataverse.org"
        },

to be:
        "citation:Point of Contact": {
                "datasetContact:Name": "Admin, Dataverse",
                "datasetContact:Affiliation": "GDCC",
                "datasetContact:E-mail": "admin@demo.dataverse.org"
        },

Error on command line, no server log error:

curl  -H X-Dataverse-key:9aef05f2-eca0-412e-be4b-4c56697700cc -X POST "http://localhost:8080/api/dataverses/root/datasets/:startmigration" --upload-file dataset-migrate.jsonld
{"status":"ERROR","message":"Validation Failed: Contact E-mail is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ])."}

Also tried specifying file type, same results:

curl -H 'Content-Type:application/ld+json' -H X-Dataverse-key:XXX-YYY-ZZZ -X POST "http://localhost:8080/api/dataverses/root/datasets/:startmigration" --upload-file dataset-migrate.jsonld
{"status":"ERROR","message":"Validation Failed: Contact E-mail is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ])."}

Took file from doc:
https://guides.dataverse.org/en/5.10/_downloads/e6c9ad6742e37c64eabaea749a13fa27/dataset-migrate.jsonld

### LINKED ISSUE #8590: Semantic Metadata API can break when labels are changed (e.g. Contact -> Point of Contact)
When we merged pull request #8454 that changed a variety of metadata field labels in the citation block (e.g. "Contact" became "Point of Contact"), we learned that it resulted in a backward-incompatible change to the Semantic Metadata API.

Specifically, testSemanticMetadataAPIs in DatasetsIT started failing.

The goal is for label changes to not affect the API. That is, pull requests against the citation block or other blocks where you are only changing the human-readable label ("Contact" vs. "Point of Contact") should not result in a breaking change to the API.

@qqmyers has provided a nice write up of the situation at https://github.com/IQSS/dataverse/issues/8533#issuecomment-1078353898 and discussed all this during a recent tech hours.

I don't want to speak for everyone but I'm ok with a breaking change to the Semantic Metadata API in order to make it more tolerant of future label changes.  One of the options we discussed is switching from the human readable label (Contact) to the machine readable one (datasetContact). That way, if "Contact" becomes "Point of Contact" in the future, the API  continues to work the same way. So when creating a dataset, we'd change part of the JSON, perhaps like below.

## old/current human readable

```
  "https://dataverse.org/schema/citation/Contact": {
    "https://dataverse.org/schema/citation/datasetContact#E-mail": "finch@mailinator.com",
    "https://dataverse.org/schema/citation/datasetContact#Name": "Finch, Fiona"
  }
```

## new/proposed machine readable

```
  "https://dataverse.org/schema/citation/datasetContact": {
    "https://dataverse.org/schema/citation/datasetContactEmail": "finch@mailinator.com",
    "https://dataverse.org/schema/citation/datasetContactName": "Finch, Fiona"
  }
```

These machine readable names come from the tsv:

```
cat citation.tsv | cut -f2 | grep -i contact
datasetContact
datasetContactName
datasetContactAffiliation
datasetContactEmail
```

These are just ideas. Mostly I'm just trying to capture the problem. I'm not trying to specify the exact solution.

One more thing I'd feel remiss without saying... at some point we should decide that it's time for the Semantic Metadata API to graduate from the Developer Guide to the API Guide. No rush on this. Whenever we're ready.

qqmyers:
**What this PR does / why we need it**: As discussed in the issues, the default semantic URI assigned to metadatablock terms (datasetFieldType), when no term specific URI is defined for it in the metadatablock, was accidentally dependent on the displayed title of the block, meaning that the identifier changes if the display name is changed. This update fixes that and also simplifies the generation of default term URIs, now relying on the fact that all child terms, like any term, must have a unique name. The mapping is now just to append the datasetFieldType name to the URI assigned to the block. Tests, guides, examples also updated. 

- Closes #8533 
- Closes #8590

**Special notes for your reviewer**: This should be a permanent fix and allow #8454 to be reapplied without breaking tests, etc.
It does include updates to the DatasetFieldType class to enforce uniqueness of the name and uri in the db. This was assumed, and hardcoded in some places already (e.g. uses of the DatasetFieldType.findByName query call getSingleResult and would fail if there were more than one.) I also added a flyway script to add these constraints to existing dbs as well. These constraints aren't required for this PR, but thought they'd be useful and are related. I also did some refactoring to shift the code managing namespace/uri generation to the MetadataBlock and DatasetFieldType classes. There shouldn't be code elsewhere assigning those. That said, if these methods should be in the related service beans, let me know (they're relatively simple and based on the variables in the entities).

**Suggestions on how to test this**: The automated tests cover the round-trip of export/recreate of datasets via the semantic api. Manually testing the inline examples and example files changed in the PR would make sense. 

**Is there a release notes update needed for this change?**: yes - this is a breaking change for the 'experimental' semantic api, although a fairly minor one (it affects terms that have not been mapped to external vocabularies). It does mean that json-ld retrieved via the semantic api from older versions won't correctly import into new versions without some minor modifications to match the new naming convention.

**Additional documentation**:


landreev (approved):
Sorry for the delay with this. 
It does make sense. I'm fine with the methods mentioned in the PR intro being where they are. 
I agree it's a good idea to add the constraints to DatasetFieldType. (Are we actually aware of any real life way/reason for an installation to end up with multiple/duplicated entries in normal operation? Or is this just in case? - just curious).

mreekie:
Sprint: 
 - This will go through first.
 - Then #8127 goes into the next sprint (Needs a new PR)

kcondon:
Tests:
1. semantic api endpoints, including sample create dataset jsonld file
2. migrate dataset endpoint, including sample migrate dataset jsonld file
3. use the create/migrate sample files in native, SWORD apis also
4. check export
5. apply field name-modified citation block to copy of db, check the above tests again
6. build guides, check altered pages
7. check release notes for required action on deployment, alert/discuss as needed

pdurbin:
> Then #8127 goes into the next sprint (Needs a new PR)

I just put this in Next Sprint.

PR #8610, opened 2022-04-13
TDL/7493 Batch Archiving

### LINKED ISSUE #7493: Support batch archiving
Currently, the only ways to trigger archiving leveraging the OA-ORE and Bag outputs is to configure a post-publication workflow or to use the API call that can archive a single specified datset version. To support batch archiving, e.g. of all dataset versions not yet archived (e.g. one published since the last archiving run), an API to find/archive all new dataset versions would be useful. 

TDL has created such an admin API call that will query to find unarchived dataset versions and, optionally to start a thread to archive them. The thread is asynchronous, creates Bags sequentially (to avoid having multiple versions taking processing power, memory and temporary disk space), and logs successes and failures. The API has three query parameters that can be used in combination:

- listonly - true: retrieves the list of unarchived versions but does not attempt to archive any
- latestonly - true: only lists/processes the most recently published version of a given dataset
- limit - maximum number of versions to attempt to archive

PR to follow.

qqmyers:
**What this PR does / why we need it**:  This PR implements an api call to allow archiving of all unarchived datasets/only the latest version of unarchived datasets that can be used in lieu of configuring a post-publish workflow, or used as a catchup mechanism to archive datasets that were published prior to an archiving workflow being configured. The api also has a view-only param that will list dataset versions that will be archived without doing it.

Closes #7493 

**Special notes for your reviewer**: The code includes a ToDo to update the logging - once the changes to support Failure/Pending/Success status (#8696) is merged, this code should/will be updated to appropriately count Failure status as failed and Pending status as a success (the processing is async so pending and success are both successful launches w.r.t. this code.

**Suggestions on how to test this**: Setup your favorite archiver (e.g. the Local one), publisha few datasets /versions and then run this api to see if all/the latest versions cause bags to be generated and sent to your archive location. (Can also try the list-only option described in the docs to just list which ones will be done and then confirm that the same list is actually done.

**Is there a release notes update needed for this change?**: planning one release note across PRs

**Additional documentation**:


qqmyers:
FWIW: tagging this as 3a since it includes changes to the archive api (POST vs GET, namechange) that we should have before trying to sync with 3A services again.

qqmyers:
@scolapasta  - 3A could/should have the changes to the API naming that is in the PR. I can pull that out elsewhere if we don't want to handle the new batch call now.

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
```diff
@@ -1187,4 +1187,36 @@ private DatasetVersion getPreviousVersionWithUnf(DatasetVersion datasetVersion)
...
     
+    /**
+     * Execute a query to return DatasetVersion
+     * 
+     * @param queryString
+     * @return 
+     */
+    public List<DatasetVersion> getUnarchivedDatasetVersions(){
+        
+        String queryString = "SELECT OBJECT(o) FROM DatasetVersion AS o WHERE o.releaseTime IS NOT NULL and o.archivalCopyLocation IS NULL";
```
scolapasta, line 1198: Could we make this a named query?

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1805,31 +1805,34 @@ public Response validateDataFileHashValue(@PathParam("fileId") String fileId) {
...
             AuthenticatedUser au = findAuthenticatedUserOrDie();
-			// Note - the user is being set in the session so it becomes part of the
-			// DataverseRequest and is sent to the back-end command where it is used to get
-			// the API Token which is then used to retrieve files (e.g. via S3 direct
-			// downloads) to create the Bag
+            // Note - the user is being set in the session so it becomes part of the
+            // DataverseRequest and is sent to the back-end command where it is used to get
+            // the API Token which is then used to retrieve files (e.g. via S3 direct
+            // downloads) to create the Bag
             session.setUser(au); // TODO: Stop using session. Use createDataverseRequest instead.
```
scolapasta, line 1819: should we make this change while we're in this code?

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1850,6 +1854,94 @@ public void run() {
...
+    public Response archiveAllUnarchivedDatasetVersions(@QueryParam("listonly") boolean listonly, @QueryParam("limit") Integer limit, @QueryParam("latestonly") boolean latestonly) {
+
+        try {
+            AuthenticatedUser au = findAuthenticatedUserOrDie();
+
+            // Note - the user is being set in the session so it becomes part of the
+            // DataverseRequest and is sent to the back-end command where it is used to get
+            // the API Token which is then used to retrieve files (e.g. via S3 direct
+            // downloads) to create the Bag
+            session.setUser(au);
```
scolapasta, line 1878: same as above, if we do decide to make the change

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1850,6 +1854,94 @@ public void run() {
...
+                            logger.info("    " + dv.getDataset().getGlobalId().toString() + ", v" + dv.getFriendlyVersionNumber());
+                            current++;
+                        }
+                    }
+                    return ok(jab); 
+                }
+                String className = settingsService.getValueForKey(SettingsServiceBean.Key.ArchiverClassName);
+                AbstractSubmitToArchiveCommand cmd = ArchiverUtil.createSubmitToArchiveCommand(className, dvRequestService.getDataverseRequest(), dsl.get(0));
+                final DataverseRequest request = dvRequestService.getDataverseRequest();
+                if (cmd != null) {
```
scolapasta, line 1900: could you explain what this line is doing? (i.e. why is a command being created before this and checked for null)

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1850,6 +1854,94 @@ public void run() {
...
+                                        failures++;
+                                        logger.log(Level.SEVERE, "Unexpected Exception calling  submit archive command", ex);
+                                    }
+                                }
+                                logger.fine(successes + failures + " of " + total + " archive submissions complete");
+                            }
+                            logger.info("Archiving complete: " + successes + " Successes, " + failures + " Failures. See prior log messages for details.");
+                        }
+                    }).start();
+                    return ok("Archiving all unarchived published dataset versions using " + cmd.getClass().getCanonicalName() + ". Processing can take significant time for large datasets/ large numbers of dataset versions. View log and/or check archive for results.");
```
scolapasta, line 1933: oh is it to havr the command here?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1156,23 +1158,33 @@ For example:
...
 
 In the Chronopolis case, since the transfer from the DuraCloud front-end to archival storage in Chronopolis can take significant time, it is currently up to the admin/curator to submit a 'snap-shot' of the space within DuraCloud and to monitor its successful transfer. Once transfer is complete the space should be deleted, at which point the Dataverse Software API call can be used to submit a Bag for other versions of the same Dataset. (The space is reused, so that archival copies of different Dataset versions correspond to different snapshots of the same DuraCloud space.).
 
+A batch version of this admin api call is also available:
+
+``curl -X POST -H "X-Dataverse-key: <key>" http://localhost:8080/api/admin/archiveAllUnarchivedDatasetVersions?listonly=true&limit=10&latestonly=true``
+
+The archiveAllUnarchivedDatasetVersions call takes 3 optional configuration parameters. 
+* listonly=true will cause the API to list dataset versions that would be archived but will not take any action.
+* limit=<n> will limit the number of dataset versions archived in one api call to <= <n>.
```
scolapasta, line 1184: I get how this works, but what's the reason to limit this way? (counting both successes and failures)

scolapasta (changes requested):

doc/sphinx-guides/source/installation/config.rst
qqmyers: listonly=true gives you a list, so with limit working this way you can make sure that only the things you listed will get processed when you drop listonly=true.
Overall, the concern is about load, particularly if/when something is misconfigured and everything will fail after all the work to create a bag.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
qqmyers: The create method is trying to read the property and instantiate the specified class (a subclass of AbstractSubmitToArchiveCommand) via reflection. If the class doesn't exist, it would fail/return null.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
qqmyers: Not really - I could return the className string as done in the null case in the else clause.

doc/sphinx-guides/source/installation/config.rst
scolapasta: Are these guaranteed to work with the list in the same order each time (i.e. if something added, it would be added at the end, so limit is guaranteed to get the things from the last listAll)?

src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java
```diff
@@ -1195,4 +1195,25 @@ private DatasetVersion getPreviousVersionWithUnf(DatasetVersion datasetVersion)
...
+     * Execute a query to return DatasetVersion
+     * 
+     * @param queryString
+     * @return 
+     */
+    public List<DatasetVersion> getUnarchivedDatasetVersions(){
+        
+        try {
+            @SuppressWarnings("unchecked")
+            List<DatasetVersion> dsl = em.createNamedQuery("DatasetVersion.findUnarchivedReleasedVersion").getResultList();
```
scolapasta, line 1209: I think Named Query should still be able to take a type, si you don't need the SuppressWarnings.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
scolapasta: Ah gotcha ot could actually return null, then. Maybe just q quick comment to make that clear, since it's different than when we normally create a command?

scolapasta (changes requested):

doc/sphinx-guides/source/installation/config.rst
qqmyers: ~yes - it's the new named query that is getting the list so unless something affects the return order from that (which I think will go in id order by default), it wouldn't change.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1805,31 +1802,33 @@ public Response validateDataFileHashValue(@PathParam("fileId") String fileId) {
...
+            @PathParam("version") String versionNumber) {
 
         try {
             AuthenticatedUser au = findAuthenticatedUserOrDie();
-			// Note - the user is being set in the session so it becomes part of the
-			// DataverseRequest and is sent to the back-end command where it is used to get
-			// the API Token which is then used to retrieve files (e.g. via S3 direct
-			// downloads) to create the Bag
-            session.setUser(au); // TODO: Stop using session. Use createDataverseRequest instead.
+            // Note - the user is being set in the session so it becomes part of the
```
scolapasta, line 1812: just noticed I think this comment can go, no?

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1850,6 +1850,98 @@ public void run() {
...
+     * @return
+     */
+    @POST
+    @Path("/archiveAllUnarchivedDatasetVersions")
+    public Response archiveAllUnarchivedDatasetVersions(@QueryParam("listonly") boolean listonly, @QueryParam("limit") Integer limit, @QueryParam("latestonly") boolean latestonly) {
+
+        try {
+            AuthenticatedUser au = findAuthenticatedUserOrDie();
+
+            // Note - the user is being set in the session so it becomes part of the
```
scolapasta, line 1870: and here.

qqmyers: updated

scolapasta (approved):

kcondon:
Not seeing json output, logger needs to be fine rather than info, seeing null ptr at end of log output.

  A system exception occurred during an invocation on EJB Admin, method: public javax.ws.rs.core.Response edu.harvard.iq.dataverse.api.Admin.archiveAllUnarchivedDatasetVersions(boolean,java.lang.Integer,boolean)]]

Caused by: java.lang.NullPointerException
	at edu.harvard.iq.dataverse.DatasetVersion.getFriendlyVersionNumber(DatasetVersion.java:539)

PR #8627, opened 2022-04-20
7565 update oncomplete

### LINKED ISSUE #7565: Improve (simplify) Primefaces (Ajax) / Bootstrap interaction

sekmiller:
**What this PR does / why we need it**: removes the rebind update that was on almost all of the button in the app

Closes #7565 

**Suggestions on how to test this**: there are no functional changes to the app, test is same as the previous pilot PR

**Additional documentation**:


scolapasta (approved):

kcondon:
Issues found:
1. Harvest client failure, null ptr license [Update: this exists in v5.10.1]
2. Creating a template fails with toa validation error
3. Enabling/disabling guestbook throws tika zero byte file exception in log but works.

Issue #8659, opened 2022-04-28
DDI exports fail for some datasets in KU Leuven Dataverse


ErykKul:
**What steps does it take to reproduce the issue?**
This is specific issue for KU Leuven Dataverse since the field for related materials in the metadata is customized for that installation. DDI exports for datasets with related materials filled in fail.

* When does this issue occur?
When we try to do an DDI export on a dataset that has a related material filed filled in, the export fails with an exception.

* Which page(s) does it occurs on?
Dataset view in metadata tab with export metadata button.

* What happens?
Error is given an en exception is logged.

* To whom does it occur (all users, curators, superusers)?
All users of the KU Leuven dataverse.

* What did you expect to happen?
Since this is not a mandatory field, loggin a warning would be sufficient. The export should work with related materials skipped.

**Which version of Dataverse are you using?**
5.10.1

**Any related open or closed issues to this bug report?**
No, this is specifiek to our dataverse installation.

PR #8660, opened 2022-04-28
DDI export made more defensieve for othrStdyMat element

ErykKul:
It does not change any functionality it only makes exporting DDI more defensieve. For KU Leuven dataverse DDI exports fail for othrStdyMat element since this field is customized in that installation. However, it is not a required element, so it can be skipped with a warning but without exception.

Closes [#8659](https://github.com/IQSS/dataverse/issues/8659)

No, it does not change any functionality.

No.


sekmiller (approved):
Looks good. Thank you for your contribution!

PR #8668, opened 2022-04-29
#8667 fix null pointer exception

### LINKED ISSUE #8667: Suggestion for fixing a NullPointerException
When playing with the setting of external vocabularies, I started to set [:CVocConf](https://guides.dataverse.org/en/5.10.1/installation/config.html#id264). However when I set it with the two commands mentioned in the description the site returned 505 errors, due to a NullPointerException.

The server log recorder the following error message:
```
[2022-04-28T16:59:17.336+0200] [Payara 5.2021.8] [WARNING] [] [javax.enterprise.web] [tid: _ThreadID=104 _ThreadName=http-thread-pool::jk-connector(5)] [timeMill>
  StandardWrapperValve[default]: Servlet.service() for servlet default threw exception
java.lang.NullPointerException
        at edu.harvard.iq.dataverse.DatasetFieldServiceBean.getCVocConf(DatasetFieldServiceBean.java:303)
        at jdk.internal.reflect.GeneratedMethodAccessor862.invoke(Unknown Source)
```
It is a warning level, but the site was virtually down.

We checked the environment of the referenced code, and we found that null value is checked, but the code only reports it and does not prevent the flow to run into a branch which expects that the value is not null:

https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java#L293 

```java
            for (JsonObject jo : cvocConfJsonArray.getValuesAs(JsonObject.class)) {
                DatasetFieldType dft = findByNameOpt(jo.getString("field-name"));
                if (dft != null) {
                    cvocMap.put(dft.getId(), jo);
                } else {
                    logger.warning("Ignoring External Vocabulary setting for non-existent field: "
                            + jo.getString("field-name"));
                }
                if (jo.containsKey("term-uri-field")) {
                    String termUriField = jo.getString("term-uri-field");
                    if (!dft.isHasChildren()) {
                        if (termUriField.equals(dft.getName())) {
                            logger.fine("Found primitive field for term uri : " + dft.getName() + ": " + dft.getId());
                            cvocMapByTermUri.put(dft.getId(), jo);
                        }
                    } else {
                        DatasetFieldType childdft = findByNameOpt(jo.getString("term-uri-field"));
                        logger.fine("Found term child field: " + childdft.getName()+ ": " + childdft.getId());
                        cvocMapByTermUri.put(childdft.getId(), jo);
                        if (childdft.getParentDatasetFieldType() != dft) {
                            logger.warning("Term URI field (" + childdft.getDisplayName() + ") not a child of parent: "
                                    + dft.getDisplayName());
                        }
                    }
                    if (dft == null) {
                        logger.warning("Ignoring External Vocabulary setting for non-existent child field: "
                                + jo.getString("term-uri-field"));
                    }
```

Thanks to Marcel Hellkamp (GWDG) who did the error tracking at the first place!

* When does this issue occur?

All.

* What happens?
The Payara Java error reporting page is displayed.

All users.

Catch the null, and let flow runs only on the safe branch.

6.9

I am not aware of.

**Screenshots:**

*

pkiraly:

To prevent a NullPointerException.

Closes #8667

In the method I did some code format change, because it was not clear at some points where the opening brackets end. 

Do 
```
wget https://gdcc.github.io/dataverse-external-vocab-support/examples/config/cvoc-conf.json
curl -X PUT --upload-file cvoc-conf.json http://localhost:8080/api/admin/settings/:CVocConf
``` 
and compare before/after PR w.r.t. the log and site functionality (see issue).

No

No

**Additional documentation**:
No

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
@@ -278,55 +278,54 @@ public class DatasetFieldServiceBean implements java.io.Serializable {
         if (cvocSetting == null || cvocSetting.isEmpty()) {
             oldHash=null;
             return new HashMap<>();
-    } 
+        }
         String newHash = DigestUtils.md5Hex(cvocSetting);
-        if(newHash.equals(oldHash)) {
-            
+        if (newHash.equals(oldHash)) {
             return byTermUriField ? cvocMapByTermUri : cvocMap;
         } 
-            oldHash=newHash;
+        oldHash=newHash;
         cvocMap=new HashMap<>();
         cvocMapByTermUri=new HashMap<>();
         
         try (JsonReader jsonReader = Json.createReader(new StringReader(settingsService.getValueForKey(SettingsServiceBean.Key.CVocConf)))) {
-        JsonArray cvocConfJsonArray = jsonReader.readArray();
+            JsonArray cvocConfJsonArray = jsonReader.readArray();
             for (JsonObject jo : cvocConfJsonArray.getValuesAs(JsonObject.class)) {
                 DatasetFieldType dft = findByNameOpt(jo.getString("field-name"));
-                if (dft != null) {
-                    cvocMap.put(dft.getId(), jo);
-                } else {
+                if (dft == null) {
                     logger.warning("Ignoring External Vocabulary setting for non-existent field: "
-                            + jo.getString("field-name"));
-                }
-                if (jo.containsKey("term-uri-field")) {
-                    String termUriField = jo.getString("term-uri-field");
-                    if (!dft.isHasChildren()) {
-                        if (termUriField.equals(dft.getName())) {
-                            logger.fine("Found primitive field for term uri : " + dft.getName() + ": " + dft.getId());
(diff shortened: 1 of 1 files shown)


qqmyers (approved):
Makes sense - a straight-forward improvement in handling bad CVOC input.

It would be good to add specific instructions for how to cause/test this issue - in the latest release with standard metadata blocks - presumably you need a CVOC config file that has an entry for a non-existent field so dft is null - can you provide a file for this?

pkiraly:
@qqmyers What I did is simply these two commands:
```
wget https://gdcc.github.io/dataverse-external-vocab-support/examples/config/cvoc-conf.json
curl -X PUT --upload-file cvoc-conf.json http://localhost:8080/api/admin/settings/:CVocConf
```
I do not get what file do you suggest exactly. I can put these commands to a file, but what should be its name, and where to put it?

qqmyers:
Those two commands are enough - I wanted to make sure that the cvoc-conf.json you used is accessible (and it is at https://gdcc.github.io/dataverse-external-vocab-support/examples/config/cvoc-conf.json) and that that would demonstrate the issue with the develop branch (it should because "field-name": "compoundDemo" doesn't exist in any of the standard metadata blocks.)

Issue #7746, opened 2021-03-31
Dataset Success Messages: Some issues discovered in first update for later consideration


kcondon:
1. Ingest complete causes please publish message to go away, should remain. 
2. Should Metadata be plural, replace has with have? 
4. "Metadata has changed" message has 3 wordings for publish (2) and submit for review (1). Essentially a brief one "publish it" and a longer one for publish and review with ending text, "so others can see these changes"
5. There are 3 different messages for three different file update conditions: "Files have been updated", "Selected files have been updated", "The file has been updated". The singular version, "the file has been updated", never appears. Only the first two messages appear ("Files have been updated" when uploading files and "selected files have been updated" when selected/edited or deleted).

kcondon:
<img width="1164" alt="Screen Shot 2021-03-31 at 11 00 22 AM" src="https://user-images.githubusercontent.com/4983816/113177680-ca419500-921b-11eb-9192-c57dbb8e4b35.png">
<img width="1159" alt="Screen Shot 2021-03-31 at 11 00 48 AM" src="https://user-images.githubusercontent.com/4983816/113177683-cada2b80-921b-11eb-8a8f-3a32e4b77e62.png">
<img width="1141" alt="Screen Shot 2021-03-31 at 11 01 10 AM" src="https://user-images.githubusercontent.com/4983816/113177684-cada2b80-921b-11eb-8fcc-5570ab6228e8.png">
<img width="1147" alt="Screen Shot 2021-03-31 at 11 01 40 AM" src="https://user-images.githubusercontent.com/4983816/113177687-cada2b80-921b-11eb-8152-466ac7a6c0e2.png">
<img width="1141" alt="Screen Shot 2021-03-31 at 11 03 24 AM" src="https://user-images.githubusercontent.com/4983816/113177688-cb72c200-921b-11eb-8225-13607749f454.png">
<img width="1144" alt="Screen Shot 2021-03-31 at 11 03 51 AM" src="https://user-images.githubusercontent.com/4983816/113177690-cb72c200-921b-11eb-810e-d6b36f7c763f.png">
<img width="1139" alt="Screen Shot 2021-03-31 at 11 04 43 AM" src="https://user-images.githubusercontent.com/4983816/113177691-cb72c200-921b-11eb-8a58-951266702b1b.png">
[...]

PR #8687, opened 2022-05-10
4153 update success please share messages

### LINKED ISSUE #4153: Publish dataset after editing a file
When on a file landing page and you create a new version aka "draft," there is no way to publish. 

see RT ticket for example: https://help.hmdc.harvard.edu/Ticket/Display.html?id=253546 

Note that there is a work-around: user can go to the dataset page, select the file, and select the new version from the version tab on the file page. Then publish appears on the draft file page.

sekmiller:
**What this PR does / why we need it**: Adds a persistent warning message to all draft versions of datasets reminding the current user to publish or submit for review as applicable to their permissions. Also adds a message to the file page of any dataset with the publish/submit message including a link to the dataset page. We need it because too many datasets are left in a draft state and users don't seem to know that they need to publish them so that changes are visible to others. Also some minor wording changes on the general dataset create/update success messages.

Closes #4153 Publish dataset after editing 
            #7746 Dataset Success Messages

**Special notes for your reviewer**: Sorry about the whitespace diffs on Dataset service bean. I guess I was too aggressive with the Source/Format. The only real changes there are with the getReminderString method around lines 805-845 or so.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Implemented as designed and agreed to in the original ticket #4153 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2081,6 +2081,7 @@ private String init(boolean initFull) {
         }
 
         displayLockInfo(dataset);
+        displayPublishMessage();
```
scolapasta, line 2084: As a check, does the message stay in place when you do a change on the page that updates (via ajax) the messages panel?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2136,6 +2137,12 @@ private void displayWorkflowComments() {
             }
         }
     }
+    
+    private void displayPublishMessage(){
+        if (workingVersion.isDraft() && workingVersion.getId() != null && canUpdateDataset()){
```
scolapasta, line 2142: In the case that workingId is = null, when is that? When you're creating a new version?  Why is that scenario different? (may have an obvious answer)

scolapasta (commented):
@sekmiller just a couple of (hopefully) quick questions. Looks good overall and I think we'll get to QA tomorrow.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: It stays up when a tabular file is uploaded and stays on after the ingest lock is removed (ingest successful)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
sekmiller: it's on dataset create.  the check permissions fails when the id is null.

scolapasta (approved):

kcondon:
@sekmiller minor q on the publish messaging pr, when you have a chance. I'd noticed that the messages seem correct at the right times but once you submit for review or publish, it continues to show the "it must be submitted/published" message in addition to the new message of "locked while publishing" and "has been submitted for review". This seems minor -locked while publishing is temporary, probably :wink: but the submitted for review state's time to live is unknown so it kind of says you need to take this action but also says you've already taken this action. a bit confusing I suppose.

sekmiller:
Fixed the issues noted above. With the fix there's a warning thrown into the log in the case where the dataset is locked and the publish message is required. not sure how to get it to go away.

PR #8691, opened 2022-05-11
License info is now also shown in submit for review popup

### LINKED ISSUE #8561: Better license workflow on dataset creation
The selection of a license should be more explicit and visible upon dataset creation.

Dataset creators

The new license feature and an improved user experience.

The license choice should be more accessible and more visible, perhaps even mandatory.

The license selector is hidden under the Terms tab and Dataverse doesn’t ask the user to pick a license when they first create a dataset.
Our institution doesn’t allow us to set a default license because of legal reasons (I know, we’re not happy about it either). However, we also think it would be best practice to explicitly get the approval of the user for the license applied to their dataset by making the license selection a bit more obvious and visible.

We would like the dataset creation workflow to allow the creator to select a license on creation or at least always see which default license will be applied to the dataset. And therefore not only have it hidden under the Terms tab.

A solution (option A) could be to do have a similar workflow to that of the template creation (see screenshot). When you create a template, you are immediately sent to Terms and not to the draft dataset once you’ve filled in all your metadata.
Another option (option B) would be to add a license field in the create a dataset form above the citation metadata block, like where the ‘Dataset Template’ selector is.

image

- https://github.com/IQSS/dataverse/issues/8554

ErykKul:
We show a warning about the license in the submit for review popup, but not the license itself, like we do in the publish popup.

- Closes #8561

image


pdurbin (approved):
Looks good to me. I ran the code and will add some screenshots below.

I updated the docs in the user guide to mention the popup.

I also added a release note snippet.

I tested with custom terms. I would suggest testing licenses as well but from what I can tell the xhtml was unchanged and simply moved into a fragment.

## Publish popup (unchanged in this PR)

Screen Shot 2022-05-12 at 4 45 06 PM

## Submit for Review popup (develop/before)

Screen Shot 2022-05-12 at 4 45 42 PM


## Submit for Review popup (this PR/after)

Screen Shot 2022-05-12 at 4 53 04 PM

PR #8423, opened 2022-02-14
add CC BY, improve license config docs #8347

pdurbin:
**What this PR does / why we need it**:

Pull request #7920 added multiple license support but we could take it a bit farther by providing new installations additional licenses beyond CC0 and by providing more documentation on how to manage licenses and add popular licenses such as the ones from Creative Commons.

To this end:

- CC BY is now available for new installations.
- New "Configuring Licenses" section added to Installation Guide.
  - JSON files for all Creative Commons licenses are downloadable.
- URL fixed in licenseCC-BY-NC-ND-4.0.json.
- Various cleanup of typos, formatting, etc.

**Which issue(s) this PR closes**:

Closes #8347

**Special notes for your reviewer**:

I tried to take a relatively light touch on existing docs. I used the new "Configuring Licenses" section as a way to weave them together. However, I was tempted to change the following:

- Move the docs on `/api/licenses` from "Admin" to their own section ("Licenses"). They really aren't part of the Admin API. Let me know and I'll include this change.

Also, I used "CC0 1.0" instead of "CC0" in anticipation of pull request #8413 being merged.

**Suggestions on how to test this**:

Check to make sure of the following:

- CC BY is now available for new installations.
- URL fixed in licenseCC-BY-NC-ND-4.0.json.
- Docs look and make sense.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

**Is there a release notes update needed for this change?**:

No. I think we have enough release notes docs for multiple licenses. We can tweak them as needed at release time. The CC BY addition is only for new installations, not upgrades.

**Additional documentation**:

None.

PR #8697, opened 2022-05-16
8671 license sorting

### LINKED ISSUE #8554: Multiple license sorting is inconsistent
Using Dataverse >= 5.10, install multiple licenses. The licenses will be sorted in the order they are loaded.
Then make one of the licenses the default or change active to false and back to true using the API.
The order of the licenses will be changed. Probably the last one you modified using the API will appear last in the drop-down list.

* When does this issue occur?
When a license is changed via API or even when editing the database table directly.

The license selection in the Terms page

* What happens?
The order of the licenses in the drop-down is changed.

Any installation using multiple licenses

The licenses sorting order should remain the same or at least should be consistent.

5.10

**Screenshots:**

*

### LINKED ISSUE #8671: Sorting of licences by admins
The current solution to keeping a consistent order of licenses in GUI is ordering them by ID. See also https://github.com/IQSS/dataverse/issues/8561

The more desired situation is where admins can arrange the order of licenses with the API. This could be achieved, for example, by adding a new "sort order" database column, as pointed out in the linked issue.

Superuser

Originally, the licenses could not be sorted, resulting in inconsistent behavior. The current solution of sorting by ID is temporal, as a more desired solution is allowing manual sorting by the admins, as proposed in this request.

Sorting of licenses in the license selector.

No

https://github.com/IQSS/dataverse/issues/8561

ErykKul:
Right now licenses are sorted only by the ID and this order cannot be changed. With this pull request, admins can sort the licenses with the API calls, where the ordering by ID remains default until changed by manually sorting (``order by sortOrder, id``). SortOrder is a new column in license that becomes also visible in JSON representation of the license.

- Closes #8554
- Closes #8671


src/main/java/edu/harvard/iq/dataverse/license/LicenseServiceBean.java
```diff
@@ -93,11 +93,23 @@ public int setActive(Long id, boolean state) throws WrappedResponse {
...
+    }
     
     public License save(License license) throws WrappedResponse {
         if (license.getId() != null) {
             throw new WrappedResponse(new IllegalArgumentException("There shouldn't be an ID in the request body"), null);
         }
+        if (license.getSortOrder() == null) {
+            throw new WrappedResponse(new IllegalArgumentException("There should be a sort order value in the request body"), null);
+        }
         try {
```
sekmiller, line 113: Since sort order is required here, we should update the api docs to reflect that.

sekmiller (changes requested):
Looks good generally, but we should probably update the doc and example files to reflect that the add license api requires a sort order

ErykKul:
I have updated the documentation and the examples. However, I have also corrected the pre-installed licenses list in documentation by removing CC BY as `V5.9.0.1__7440-configurable-license-list.sql` does not insert it and I could not find any other place where this could happen. I am not sure about this.

pdurbin:
@ErykKul hi! The CC BY license is added in scripts/api/setup-all.sh. Please see this PR:

- #8423

Some more thoughts on what the source of truth should be for licenses (JSON used in setup scripts vs. Flyway scripts): https://github.com/IQSS/dataverse/issues/8407#issuecomment-1034170973

ErykKul:
@pdurbin I have reverted the removal of the CC BY references in the documentation.

qqmyers:
looks like the LicensesIT test is failing due to lack of that sort order value:  `{\"status\":\"ERROR\",\"message\":\"There should be a sort order value in the request body\"}\n[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.886 s <<< FAILURE! - in edu.harvard.iq.dataverse.api.LicensesIT\n[ERROR] edu.harvard.iq.dataverse.api.LicensesIT.testLicenses `

pdurbin:
@qqmyers ah, thanks!

@ErykKul can you please take a look?

I'm going to move this back into review.

ErykKul:
I have added the requested changes. The integration test that stopped working does not run by default for me locally. I will try to figure it out how to run it. Meanwhile, I will wait to see if the build succeeds.

ErykKul:
I have figured out the integration test (my problem was with different JDKs on my machine and picking up the wrong one). I think I know what the problem was with the test, and I will investigate it further, if the test is still failing.

src/test/java/edu/harvard/iq/dataverse/api/LicensesIT.java
```diff
@@ -144,6 +144,20 @@ public void testLicenses(){
         status = JsonPath.from(body).getString("status");
         assertEquals("OK", status);
         
+        //Fail trying to set null sort order
```
sekmiller, line 147: I'm getting an error in a prior test - up in row 111 above where if I leave my existing licenses sort orders null, the get license by id which is expected to fail will succeed, because the license with id 1 gets sorted to the end of the array when id 2 is set to the default license.

ErykKul: I have changed the code such that the `licenseId` is now the one with the highest ID found. It should fix this issue.

sekmiller: Thanks!

src/main/java/edu/harvard/iq/dataverse/license/License.java
```diff
@@ -73,14 +75,17 @@ public class License {
 
     @Column(nullable = false)
     private boolean isDefault;
+
+    @Column(nullable = true)
```
sekmiller, line 79: Since we're requiring that all new licenses have sort order, it seems to make sense that we make this column non-nullable. You'd have to also update the migration script to set the sort order = id(or some other value that makes sense) for existing licenses.

ErykKul: I made the column not null. I have adjusted the migration script to use 0 as the default value when creating the new column. I think that this is OK, the queries sort first by `sortId`, then by `id`. Setting it to 0 makes it clear that it was not set by admin and the logic of sorting by 'id' is kept, until `sortId` is set to some different value.

sekmiller: Setting it to 0 makes sense. Thanks for the update.

I get an error when I run LicensesIT - 

Line 95:
        long licenseId = JsonPath.from(body).<Long>getList("data[*].id").stream().max((x, y) -> Long.compare(x, y)).get();

fails first because the [*] fails to parse 

Caused by: org.codehaus.groovy.control.MultipleCompilationErrorsException: 
startup failed:
Script1.groovy: 1: unexpected token: ] @ line 1, column 33.
   restAssuredJsonRootObject.data[*].id
                                   ^

1 error

	at edu.harvard.iq.dataverse.api.LicensesIT.testLicenses(LicensesIT.java:95)

Then there's a class cast exception the id is integer not long.

When I changed the line to 

        long licenseId = (long) JsonPath.from(body).<Integer>getList("data.id").stream().max((x, y) -> Integer.compare(x, y)).get();

all of the tests pass for me.

ErykKul: Sorry that I missed it, thank you for catching and fixing it! I thought if it was not fixed correctly, it would show up in the maven test, but this one is an integration test, I forgot about that... I did the fix you suggested.

ErykKul:
There was a problem with my flyway script. I have fixed it and got the integration test to run and pass.

kcondon:
This works, waiting for integration tests to run before merging.

ErykKul:
OK. I will try to fix the jenkins build (I cannot see the output, I assume it fails because of the not null column without default value) by dropping the column first and then reverting the drop.

kcondon:
Hi @ErykKul The jenkins build fails due to a component in the test pipeline infrastructure and not due to an integration test. Essentially, the image used for the test box has been updated and we are waiting for the corresponding ansible lib to be able to host the tests. The team is looking at it now. Thanks for the efforts and apologies for the delay.

qqmyers:
The build issue is related to https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible/issues/263 and is not about a failing test. Once that issue is fixed (Don is working now), rerunning the build job should be enough to get the build to show green.  So as far as I can tell, nothing for @ErykKul to do now (except wait a bit).

ErykKul:
OK, thanks!

qqmyers:
@ErykKul FYI - tests have run and there's a new failure:
```
java.lang.AssertionError: 
Expected status code <204> doesn't match actual status code <400>.
	at edu.harvard.iq.dataverse.api.SwordIT.testDeleteFiles(SwordIT.java:863)
```

PR #8706, opened 2022-05-17
8600 duplicate template

### LINKED ISSUE #8600: Duplicate template seems to create db entries even when cancel
I noticed this strange db behavior - clone a template, click cancel and then look in db. A new row is still added to template table (but with no dataverse_id). Looking at prod, there are many of these orphaned templates.

We need to fix this logic and clean up the db.

sekmiller:
**What this PR does / why we need it**: Fixes logic in Copy template that was allowing templates with no dataverse ID to be created. Also adds admin APIs for finding and deleting templates and adds a script for deleting the orphan templates.

Closes #8600 Duplicate template seems to create db entries even when cancel.

**Special notes for your reviewer**: the fix in the cloneTemplate method was fairly simple. Handling the orphan templates added most of the changes

**Suggestions on how to test this**: the orphan templates are on the template table with Dataverse_id = null

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/TemplateServiceBean.java
```diff
@@ -46,6 +46,22 @@ public Template findByDeafultTemplateOwnerId(Long ownerId) {
         query.setParameter("ownerId", ownerId);
         return query.getSingleResult();
     }
+    
+    public List<Template> findByOwnerId(Long ownerId) {
+        TypedQuery<Template> query = em.createQuery("select object(o) from Template as o where o.dataverse.id =:ownerId order by o.name", Template.class);
```
scolapasta, line 51: can we make these named queries?

src/main/java/edu/harvard/iq/dataverse/TemplateServiceBean.java
```diff
@@ -46,6 +46,22 @@ public Template findByDeafultTemplateOwnerId(Long ownerId) {
...
+        query.setParameter("ownerId", ownerId);
+        return query.getResultList();
+    }
+    
+    public List<Template> findAll() {
+        TypedQuery<Template> query = em.createQuery("select object(o) from Template as o order by o.name", Template.class);
+        return query.getResultList();
+    }
+    
+    public List<Template> findOrphan() {
```
scolapasta, line 61: Can this be removed, as with the other code?

doc/release-notes/8600-duplicate-templates.md
```diff
@@ -0,0 +1,9 @@
+###Duplicate Templates in Database
+Prior to this release making a copy of a Dataset Template was creating two copies, one of which is visible in the dataverse, the other not being assigned 
+a dataverse was invisible to the user (https://github.com/IQSS/dataverse/issues/8600). This release fixes the issue. 
+
+If you would like to remove these orphan templates you may run the follwing script:
+
+https://github.com/IQSS/dataverse/raw/develop/scripts/issues/8600/delete_orphan_templates_8600.sh
+
+In order to support the script two admin APIs for finding and deleting templates have been added.
```
scolapasta, line 9: just one, no?

scolapasta (changes requested):

doc/release-notes/8600-duplicate-templates.md
sekmiller: badly worded ("two" as in one for finding and one for deleting)

scolapasta (approved):

PR #8723, opened 2022-05-20
custom JavaScript setting for metadata editing

### LINKED ISSUE #8722: Feature Request/Idea: adding custom controlled vocabulary JavaScript script
We can have controlled vocabulary as a list locally (with optionally translated values). But if the list is large and needs to be maintained, it is more advantageous to have, for example, a lookup functionality that allows to search for the values at an external service. We can have external controlled vocabularies with "skosmos" protocol, but this is an overkill for a simple list (enumeration) for one field (e.g., author name using author lookup). A more desirable solution is to allow a custom JavaScript to control values of specific fields. A custom script can have all the needed functionality, we only need a way to load it when it is needed.

Developer

Our own author lookup service that is now configured as an external controlled vocabulary. However, we do not have URIs, and we see many warnings in the log as the code is trying to resolve the values, while it is not supported by a simple list.

None. We want only the possibility to add custom scripts.

ErykKul:
It allows loading a custom JavaScript for editing metadata.

Closes #8722

Yes.

**Additional documentation**:
For example, when you want a metadata field to be controlled by an external set of possible values that do not have translations or URIs, you can implement a simple lookup in a JavaScript. An example of a person name lookup from within an organization:

image


qqmyers:
If all you want is to drop an arbitrary script on the page, perhaps adding it to a [custom header or footer would work](https://guides.dataverse.org/en/latest/installation/config.html#branding-your-installation)? FWIW: I did this in developing the ORCID script prior to the CVoc mechanism being in place.

pdurbin (commented):
I'm close to approving this but I made some doc suggestions here: https://github.com/ErykKul/dataverse/pull/3

Also, can I have a covoc.js file to play with? Otherwise there's not much to test or review. The js file is loaded just fine.

ErykKul:
@pdurbin 

Some extra explanation about the location of the javascript and how it is used at KU Leuven:

We have a reverse proxy that manages relative paths. `/covoc/`  points to ruby server that also implements the business logic calls to our internal (KU Leuven) servers. Since everything is behind a reverse proxy, you can enforce same origin policies (CORS) at the browser, etc. Reverse proxy, dataverse, ruby server and some other servers (e.g., database) are then docker images served by one docker with internal network with only the reverse proxy being exposed directly. It is a nice architecture done by @Kris-LIBIS for us.

The specific script we want to load: `/covoc/js/covoc.js` has relative path to the reverse proxy where dataverse is hosted. For example, in our production it becomes: https://rdr.kuleuven.be/covoc/js/covoc.js (you can just click on the link to view it in the browser). Note that it uses relative paths itself, and it won't work for you without the `covoc` ruby backend deployed.

pdurbin:
@ErykKul thanks! All the extra context helps a lot. I have no plans to deploy the Ruby backend, etc. 😄 I confirmed that when I use the new setting I can insert Javascript into the dataset page from a URL (to a Javascript file, of course).

Based on this, I'm approving this PR and sending it to QA.

Something for all of us to think about... we allow a custom CSS file (:StyleCustomizationFile). Should we allow a custom Javascript file (:JavascriptCustomizationFile?). Or multiple files? In practice, people tend to add Javascript files in a custom header or footer (as @qqmyers mentions above). This PR includes the extra Javascript on just the dataset page rather than all pages. @Kris-LIBIS mentioned above that the script didn't work in the footer so it seems like the need for this PR is real.

pdurbin (approved):
While I didn't test a Javascript file that actual does something useful like populate author names (@ErykKul explained they have a whole Ruby backend for that), I did confirm that I can add Javascript to the dataset page and at least print using console.log.

There's additional discussion in the PR about alternatives such as putting Javascript in a custom header or footer but the assertion is that this didn't work.

Approving. Off to QA.

pdurbin:
At standup I said I was happy enough with this PR to move it to QA but I'd like another set of eyes on it before we merge.

landreev:
OK, merging.

Issue #8726, opened 2022-05-23
Improved HTTP Range request support


haarli:
We are currently working on a tool which allows to preview the structure of an uploaded zip file without the need to download the whole file. This requires HTTP range requests support in the Data Access API (api/access/datafile/{id}) for files. I found out that range requests are already supported in Dataverse. However, they do not return required HTTP headers and status to comply with 3rd party tools. So I suggest the following changes:

- If a certain range is requested, the returned status should be 206 (partial content) instead of 200
- A Content-Range header should be returned containing information about the returned bytes
- A Accept-Ranges header with value "bytes" should be returned
- CORS: The "Range" header should be allowed by adding it to "Access-Control-Allow-Headers"
- CORS: Content-Range and Accept-Ranges header should be exposed by adding it to "Access-Control-Expose-Headers"

I'll provide a PR with these changes.

qqmyers:
These seem like useful additions. FWIW: I've also been looking at zip preview, with download of a file from within a zip leveraging the ranged requests (at QDR, primarily in the context of a specific zip format used for qualitative data analysis). I'm interested to see what you come up with and would be happy to compare notes if that's ever useful.

haarli:
Thanks. It's nearly ready, with the ability to download files within the zip using range requests. Here's a preview of the previewer :)

![grafik](https://user-images.githubusercontent.com/5524896/169830943-5522b16f-ddb6-4779-8c32-4cb48e95f0b3.png)



Current code is on https://github.com/MPDL/dataverse-previewers/tree/zip-preview-download if you're interested. 
It uses the zip.js library from Gildas Lormeau (https://github.com/gildas-lormeau/zip.js) and the fancytree library (https://github.com/mar10/fancytree) for rendering the treeview. It works with local file storage and S3 (needs some CORS rules configured)

I'll provide a PR with the complete previewer to the gdcc GitHub repo as soon as we are finished and have tested a bit more.

pdurbin:
Closed by this PR:

- #8727

PR #8727, opened 2022-05-23
Iqss/8726 better HTTP range request support

haarli:
This PR addresses #8726 with the following changes:
- Access.java datafile()
  - If a certain range is requested, the returned status is now 206 (partial content) instead of 200
- DownloadInstanceWriter writeTo() 
  - I f a range is requested, a Content-Range header is returned containing information about the returned bytes
  - I f a range is requested, a Accept-Ranges header with value "bytes" is returned
- APIBlockingFilter doFilter()
  - CORS: The "Range" header is added to "Access-Control-Allow-Headers"
  - CORS: The Content-Range and Accept-Ranges header are added to "Access-Control-Expose-Headers"

[from L.A.:]
A quick note on how/what to test/QA:

A call to file download api with the Range request should now return 206 instead of 200; and the extra headers as described above. 
May make sense to confirm that regular downloads (without the range header) are not affected and everything is working as before. 
This would be best to test on locally-stored files; or, if testing with s3 storage, make sure to disable the direct download option. (otherwise the return codes and the headers you're seeing are not ours, but Jeff Bezos's). 
(sorry if the above is all stating the obvious)


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -423,7 +423,10 @@ public DownloadInstance datafile(@PathParam("fileId") String fileId, @QueryParam
         /* 
          * Provide some browser-friendly headers: (?)
          */
-        return downloadInstance;
+        if (headers.getRequestHeaders().containsKey("Range")) {
+            return Response.status(Response.Status.PARTIAL_CONTENT).entity(downloadInstance).build();
+        }
+        return Response.ok(downloadInstance).build();
```
landreev, line 429: Could you please clarify what this change does? 
Specifically, changing the return value from DownloadInstance to Response - are you positive that it's still going to stream the output? (As opposed to trying to read everything from the DownloadInstance into memory at once, and *then* build the Response?)

haarli: Let's start with the requirement:
If the request contains a "Range" header and the requested range of bytes is returned, a HTTP status of 206 (Partial Content) instead of 200 must be returned. See also
https://datatracker.ietf.org/doc/html/rfc2616#section-14.35.2


So, these lines change the status code of the HTTP response dependent on the existence of the Range header in the HTTP request. The only way to change it is by using the Response object. (Setting it directly in the HttpServletResponse would be overwritten by Jersey later).

However, there is no real difference between returning the entity (DownloadInstance) directly, or wrapped in a Response object. Except that Response allows more control over settings like the status code. The custom MessageBodyWriter for DownloadInstance is still called, and that is where the streaming happens.

landreev: Thanks. For the record, this is *very* old code. Written for a very early version of Jersey. 
We have an open issue for refactoring that download subsystem. But it's going to be a project, with years worth of special case handling added there.

landreev (approved):
Moving along. 
(added a quick note for QA to the description)

landreev:
Sorry, I rushed to move this into QA. 
Some API tests are of course failing now, on account of the changed status code. 
(I'm assuming it's just that; and should be localized to FilesIT->testRange() - but I would double-check. but should be trivial to fix regardless)
Thanks.

haarli:
Uh, sorry, didn't have the integration tests on my mind. I'll have a look at them. 
Hope the status code change won't break any external tools already working with Range requests in Dataverse. Do you know of any?

landreev:
I'm not aware of any external tools that this would break, no. This native range support is a relatively new addition; and even if anyone is already relying on it, the change would be trivial on their end. (To be precise: I'm aware of some use cases relying on ranges; but they happen to use Dataverses that store files on S3 - and that happened to have been working because the request is redirected to S3, and gets handled properly there. And this isn't subject to change because of your PR). 

But this makes me realize - it would be proper style to announce this change, even if it's not likely to affect many people. Our normal procedure for having something included in the next release notes is to add your PR-specific note as a file under doc/release-notes (please see the notes already there for an example). Just one sentence - the return code is changing, the extra headers are being added, per the http spec...

And, uh, better late than never: thank you for contributing to the project! 😄

haarli:
Thanks. I've added notes for this PR under doc/release-notes.

FWIW: We just went productive with the ZIP Previewer, which relies on these changes, see e.g. here: https://edmond.mpdl.mpg.de/file.xhtml?fileId=182288
I'll create a PR for the previewer on GDCC soon.

landreev:
> We just went productive with the ZIP Previewer

Very nice, I like it a lot! I'll be looking forward to having it available @ GDCC.

kcondon:
Hi @haarli would you mind refreshing this branch from develop? We've since released a new version and the version numbers have incremented. Regards, Kevin

PR #8743, opened 2022-05-25
8715 importddi termofuse

### LINKED ISSUE #8715: Import DDI dataset with Term of Use
It seems that importDDI API does not import terms of Use from XML.
Although there is a treatment for this section in ImportDDIServiceBean and it parses correctly input xml in processDataAccs, 
dvDTO.setTermsOfUse(parseText(xmlr, "notes")) is not used, instead there is just parseText(xmlr, "notes"). Hence despite parsing the xml for terms of use, the terms never set. 

Create import xml with
`<dataAccs>
            <notes type="DVN:TOU" level="dv">Test</notes>
</dataAccs>`
in stdyDscr section.

`curl -H X-Dataverse-key:$API_TOKEN -X POST http://localhost:8080/api/dataverses/:root/datasets/:importddi --upload-file test_term_of_use.xml`

Example of such xml file is attached.

[test_term_of_use.txt](https://github.com/IQSS/dataverse/files/8718380/test_term_of_use.txt)

lubitchv:
importddi api does not set Terms of Use although there is a treatment for it in ImportDDIServiceBean. This PR fixes it.

- Closes #8715 

Create import xml with
``<dataAccs> <notes type="DVN:TOU" level="dv">Test</notes> </dataAccs>``
in stdyDscr section.

``curl -H X-Dataverse-key:$API_TOKEN -X POST http://localhost:8080/api/dataverses/:root/datasets/:importddi --upload-file test_term_of_use.xml``


lubitchv:
@pdurbin I see that doc/sphinx-guides/source/_static/api/ddi_dataset.xml does not have "Terms of Use" section it only has "Terms of Access". Should "Terms of Use" also be added there?

pdurbin:
@lubitchv I'm fine with adding it there as long as it doesn't make any tests fail. 😄 

I guess you could try it and we could see what Jenkins reports back. 😄 

Or you could test it locally: https://guides.dataverse.org/en/5.10.1/developers/testing.html#integration-tests

When I wrote that comment I was thinking, "What DDI file would a reviewer or QA use to test?" Perhaps you could simply upload one to this issue.

lubitchv:
I added terms of use into ``doc/sphinx-guides/source/_static/api/ddi_dataset.xml``. It seems to pass all the test. there is also an example [test_term_of_use.txt](https://github.com/IQSS/dataverse/files/8718380/test_term_of_use.txt)   hat attached to the issue  #8715  Although it has extension txt, it is xml file. github did not let me to attach xml.

lubitchv:
I am not sure. It does not seem to affect anything, even cached xml files since it is import. The only think to add is that now users can import xml with Terms of Use that was skipped before because of the bug. I can add it if you think it is needed in the release notes.

pdurbin:
Sure, please go ahead. I imagine it would be something like this? Please correct any misunderstandings I have!

- Terms of Use is now imported when using DDI format through harvesting or the native API. (Issue #8715 PR #8743)

Maybe harvesting isn't affected? I'm not sure.

lubitchv:
I haven't thought about harvesting, but you are right. Harvester client in HarvesterServiceBean uses ImportDDIServiceBean. So harvesting would be affected. I will add the note that you have suggested.

sekmiller (approved):

landreev:
Looks good, merging.

PR #8747, opened 2022-05-26
Gdcc/8746 single version semantics for archiving

qqmyers:
**What this PR does / why we need it**: This PR adds a mechanism for archivers to specify that they only allow archiving for a single version and implements logic to enforce that in the archiving/archival status APIs

**Which issue(s) this PR closes**:

Closes #8746

**Special notes for your reviewer**: Review after #8696 as this incorporates the status API updates from that PR. Hopefully that PR can be merged and this will be updated before review.

**Suggestions on how to test this**: The DRS Archiver (not yet in a PR) is the only archiver that supports single-version mode at present, so testing of this by itself could probably be limited to assuring that the Local archiver still works/still allows you to archive multiple versions.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**: Addressed in #8611

**Additional documentation**:

Issue #8611, opened 2022-04-13
Archiving/Bag/OAI_ORE/DataCommons Release notes


qqmyers:
Multiple issues/PRs related to Archiving/Bag creation/the OAI-ORE export included in the Bags are in progress/planned. Similarly, there are many Data Commons PRs that build on each other. Rather than create separate release notes, I'm aggregating them under one issue/PR here.

Related Issues/PRs: #7493, #8602, #8603, #8604, #8605, #8606. #8609, #8610, #8745, #8746, #8770, #8749, #8750, #8770

qqmyers:
TBD: Add BagGeneratorThreads support in archiver classes after #8606.
TBD: Update Archiver classes to report status after PR for #8605

PR #8748, opened 2022-05-26
Gdcc/8745 archival status UI

### LINKED ISSUE #8745: HDC 3A: UI for managers to view Archival Status/location
As a follow-on to https://github.com/IQSS/dataverse/issues/8605 which updates/extends the archiving mechanism with a status CRUD API, the accompanying PR adds a column to the dataset version table that allows those who can view the unpublished dataset to see basic status and for superusers to see details including the location (URL) of the archival copy and, in cases where archiving is being done manually instead of via a post-publication workflow to submit a version(s) for archiving.

The layout/text in the table have undergone UI review as part of the Data Commons project.

This issue and PR also build upon #8747 - the UI here will support traditional archivers as well as those that limit archiving to one version of a dataset.

qqmyers:
**What this PR does / why we need it**: Starting from QDR's implementation of an additional column in the dataset version table to show archival copy location, this PR adds a column that can show the new archival status/message entries that have been developed as part of the Harvard Data Commons work.

Closes #8745

**Special notes for your reviewer**: This has been demoed/reviewed and is being tested within the HDC Objective 3A group.  Builds on #8747 so could/should be reviewed after that.

**Suggestions on how to test this**: Run archiving, e.g. with the Local Archiver. Optionally use the status API added in #8696 to add other status messages, e.g. with json values like
`{"status":"pending","message":"now transferring at 3 bytes per hour!"}`
or
`{"status":"failure","message":"got hungry while waiting and quit"}`

and verify that:

- when not logged in, the dataset page version table looks as it did previously
- when logged in as someone who can view the unpublished dataset that an additional column appears that the end and one can see a Success message for archived versions and Not Archived for others
- when logged in as a superuser that the success message is a link, that hovering over a failure in the column shows the detailed message

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: Adds a column to the dataset page version table visible to privileged users (see above). There are Miro board mockups, a live machine showing the ui available on the Harvard Library's QA network, and videos available from https://github.com/harvard-lts/awesome-lts (not all current).

**Is there a release notes update needed for this change?**: addressed in #8611

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5570,6 +5570,65 @@ public void archiveVersion(Long id) {
             }
         }
     }
+    
+    public boolean isArchivable() {
+        String className = settingsWrapper.getValueForKey(SettingsServiceBean.Key.ArchiverClassName, null);
```
sekmiller, line 5575: For these booleans, does it make sense to make them nullable Booleans where the logic is only performed once and if it's not null just return the value? (like we do for isHasValidTermsOfAccess)

src/main/webapp/dataset-versions.xhtml
```diff
@@ -147,6 +147,25 @@
                 <h:outputText id="versionDate" value="#{versionTab.publicationDateAsString}" />
             </ui:fragment>
         </p:column><!-- end: date column -->
+        <!--  ToDo: change HeaderText to Bundled string, stop this fragment from being loaded for datasets during creation -->
```
sekmiller, line 150: Does this todo still apply? it looks like you took care of both of those things in line 151.

sekmiller (approved):

PR #8751, opened 2022-05-26
GDCC/8749 S3 Archiver

### LINKED ISSUE #8749: HDC 3A: Support archiving to an S3 bucket
**Overview of the Feature Request** As part of supporting archiving to the Harvard DRS, Dataverse has to transfer an archival bag and a datacite.xml file to an S3 bucket. Rather than embedding that in a fully custom DRSArchiver class, this issue/PR pulls out just the generic ability to send the archival files to an S3 store from DRS-specific functional to create a class that should be reusable in the community.

Dataverse instance system managers configuring archiving

**What inspired the request?** The Harvard Data Commons project

qqmyers:
**What this PR does / why we need it**: Implements an S3 Archiver consistent with recent updates to add archival status support, microprifile config of s3 connections, etc.

Closes #8749

**Special notes for your reviewer**: Builds on #8748 - primarily an S3 archiver class and docs with some minor refactoring to pull a few lines of common code into the abstract base class.

**Suggestions on how to test this**: Follow the configuration instructions to setup an S3 archiver and test it the same way as a local archiver. The same bag/datacite xml files should be sent to the bucket as would have been sent to the local file system.

**Is there a release notes update needed for this change?**: addressed in #8611

**Additional documentation**:


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1910,7 +1910,13 @@ The body is a JSON object that must contain a "status" which may be "success", "
   export VERSION=1.0
   export JSON='{"status":"failure","message":"Something went wrong"}'
 
+<<<<<<< HEAD
+  curl -H "X-Dataverse-key: $API_TOKEN" -H "Content-Type:application/json" -X PUT "$SERVER_URL/api/datasets/submitDatasetVersionToArchive/$VERSION/status?persistentId=$PERSISTENT_IDENTIFIER" -d "$JSON"
+  
+Note that if the configured archiver only supports archiving a single version, the call may return 409 CONFLICT if/when another version already has a non-null status.
+=======
   curl -H "X-Dataverse-key: $API_TOKEN" -H "Content-Type:application/json" -X PUT "$SERVER_URL/api/datasets/:persistentId/$VERSION/archivalStatus?persistentId=$PERSISTENT_IDENTIFIER" -d "$JSON"
+>>>>>>> refs/remotes/IQSS/develop
```
sekmiller, line 1919: looks like an unresolved merge conflict above

src/main/resources/db/migration/V5.10.1.3__8605-support-archival-status.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE datasetversion SET archivalCopyLocation = CONCAT('{"status":"success", "message":"', archivalCopyLocation,'"}') where archivalCopyLocation is not null and not archivalCopyLocation='Attempted';
+UPDATE datasetversion SET archivalCopyLocation = CONCAT('{"status":"failure", "message":"Attempted"}') where archivalCopyLocation='Attempted';
```
pdurbin, line 2: @qqmyers can you please bump the version number on this Flyway script? I can't deploy:

Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 5.10.1.3
Offenders:
-> /usr/local/payara5/glassfish/domains/domain1/applications/dataverse-5.11/WEB-INF/classes/db/migration/V5.10.1.3__8599-legacy-templ
ates.sql (SQL)
-> /usr/local/payara5/glassfish/domains/domain1/applications/dataverse-5.11/WEB-INF/classes/db/migration/V5.10.1.3__8605-support-arch
ival-status.sql (SQL)

pdurbin (changes requested):
The flyway script version needs to be updated.

src/main/resources/db/migration/V5.10.1.3__8605-support-archival-status.sql
pdurbin: Oh, actually, I think we simply need to delete V5.10.1.3__8605-support-archival-status.sql since it's the same as V5.11.0.1__8605-support-archival-status.sql. Does that sound right?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5570,6 +5570,65 @@ public void archiveVersion(Long id) {
...
+                // on it, we can archive, so return true
+                return true;
+            }
+        }
+        //not in an archivable collection
+        return false;
+    }
+    
+    public boolean isSomeVersionArchived() {
+        for (DatasetVersion dv : dataset.getVersions()) {
```
sekmiller, line 5625: Does it make sense for this method to be on the DatasetServiceBean?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5570,6 +5570,65 @@ public void archiveVersion(Long id) {
             }
         }
     }
+    
+    public boolean isArchivable() {
+        String className = settingsWrapper.getValueForKey(SettingsServiceBean.Key.ArchiverClassName, null);
```
sekmiller, line 5575: Does it make sense for this method (and the next one) to be on the ArchiverUtil?

qqmyers: @sekmiller  - Maybe? These, and the one above are all used in the dataset-versions.xhtml file so moving them completely to ArchiverUtil (or DatasetServiceBean above) means having to make those classes (with ArchiverUtil not being a bean) accessible in the xhtml and not being able to use settingsWrapper to cache the property values(again because , with ArchiverUtil is not a bean/ is called from the API where there is no view/session scope). 

One way I could simplify a bit would be to shift the reflective call logic into ArchiverUtil, e.g. a getArchiverOption(String className, String methodToInvoke) method. That would reduce duplicate code and minimize what's in DatasetPage. 

Does that make sense? Or is it worth moving them out of DatasetPage completely?

sekmiller: Maybe just isSomeVersionArchived which is in the DatasetPage as well as Admin and Datasets.

sekmiller: or maybe it's not worth the trouble

qqmyers: I'll do some quick de-duplication and push it back. Thanks!

src/main/resources/db/migration/V5.10.1.3__8605-support-archival-status.sql
sekmiller: This file has been removed.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
qqmyers: I moved the guts of someVersionArchived to the util but the rest seem hard - there are methods that require a dataset and some that are static, some use settingsWrapper and some settingsServiceBean due to their context (UI or API) so as I tried to refactor I didn't see good opportunities to consolidate.

pdurbin:
I know at standup I said I'd move this over to QA but I just noticed the API tests didn't run so I just pushed a typo fix to enable another run.

pdurbin (approved):
Ok, tests are passing now so I'm hitting "Approve"  (@sekmiller is the one who did the actual review).

Issue #7000, opened 2020-06-18
As a sysadmin I want to use MicroProfile Config API to configure my installation


poikilotherm:
Now that Payara 5 is actually happening with the release of Dataverse 5 (thank you so much folks!!!), we should re-raise the topic (formerly #5187) of starting to use some of the goodies coming along with this.

As a system admin pushing to cloud, I would greatly benefit from having the possibility to get configuration inside my instance not via mangling with JVM options and database settings. Instead, I would like to see this being more flexible and combined with a hierarchy.

Introduce [MicroProfile Config API](https://github.com/eclipse/microprofile-config) and its support in [Payara 5](https://docs.payara.fish/community/docs/5.2020.2/documentation/microprofile/config.html)

cmbz:
@poikilotherm Could you please review and see if this issue can be closed now, given the amount of work that appears to have been completed towards your request?

PR #8753, opened 2022-05-27
8372 gdcc xoai library

### LINKED ISSUE #8372: Update and reorganize the XOAI dependencies under local_lib
Currently, Dataverse codebase uses a custom patched XOAI 4.1.0, provided in `/local_lib`.

It would be a good idea to

1. Find out what patches have been applied (needs anatomization of patched JAR sources with original sources)
2. Fork https://github.com/DSpace/xoai (which is the moved repo from https://github.com/lyncode) to @gdcc
3. Find out if our patches are still needed
4. As development in https://github.com/DSpace/xoai has stalled (see e.g. https://github.com/DSpace/xoai/issues/72#issuecomment-557292929), make it our package
5. Update LOTS of dependencies and especially get rid of log4j-1.2!
6. Setup Github Workflow, Code Coverage etc.
7. Make releases to Maven Central
8. Use & test in Dataverse upstream code

It might be worth a try to 1) not rename the packages but still publish under our @gdcc Maven group id and 2) create fork and a pull request for upstream plus setting up wei/pull to auto-create pull requests when DSpace updates their branch.

landreev:
Big technical debt PR; eliminates the custom jars of the old lyncode XOAI library; replaces them with dependencies for the new io.gdcc.xoai packages; fixes some known issues in the process. 
Still a draft; basic functionality is there; still working on more tests, although I'm considering splitting that out as a separate issue. 

- Closes #8372

The scope of this PR has ballooned A LOT. It was initially started under the assumption that it was simply to accommodate the move of the XOAI libraries from local_lib into a gdcc repo _as they were_, with only some minimal changes, leaving any further improvements and cleanup of the modifications in the original fork for later, as separate issues/PRs. Instead a very thorough refactoring of the entire xoai package was done before a stable 5.0 RC release was made on gdcc/xoai, reimplementing and/or getting rid of the early dataverse hacks and solving many issues in the process. Huge thanks to @poikilotherm for all the work there. 
This is all a "good thing, not a bad thing"; it was all very real technical debt that we needed to address rather sooner than later. This has allowed to very much streamline the dataverse-side OAI implementation and get rid of most of the code that was maintained there. There's more that still needs to be done on the dataverse side for sure. But since this has already been a much larger effort than was originally scheduled, I really want to move the PR along and then proceed to handle the remaining improvements as separate tasks. 

**Additional documentation**:


landreev:
@poikilotherm Hi, Harvard just gave us all an extra mini-vacation, but I'm back at work. I made this draft PR last week that finally gets rid of the old custom jars in local_lib. Once again, I cannot fully express my gratitude for your leading of this effort.  
I was able to easily get rid of the extra xoai classes on the dataverse side; and to implement almost everything using the new mechanisms provided in gdcc.xoai, as we planned. I created a local `Condition` in order to pass the MetadataFormat to the `getItems()` methods in the ItemRepository implementation via a ScopedFilter. Etc. etc. 
There are extra things that need to be cleaned up: the `Date -> Instant` change is currently handled via quick conversions between the two (i.e., the underlying entities are left unchanged, using `Date` for the stamps); it clearly needs more API tests. I'm working on this, but considering splitting the task into a separate issue. 

There is one thing I could not implement 1:1 as it was: the handling of the proprietary "Dataverse JSON harvesting". I wanted to consult with you about this. The way the JSON api line has been encoded in the original implementation: 
```
<metadata directApiCall="..."/>
```
 \- This doesn't seem to be possible, to implement this using the gdcc.xoai framework without adding ugly hacks on the library side again (?). And it was a somewhat unfortunate choice in the first place. I feel like *if* we want to keep this "json harvesting" going forward, it should be something like
```
<metadata>
   <dataverseCustom directApiCall="..."/>
</metadata>
``` 
instead. This I was able to easily implement in the item repository; but it obviously introduces a backward incompatibility.  Do you have any thoughts on this? I'm also going to talk to the core dev. team, about whether we want to continue supporting this functionality; and/or if we want to change it.

pom.xml
```diff
@@ -382,52 +382,21 @@
...
-        <!-- (note the reload4j explicitly added below as a replacement! -->
-        <dependency>
-            <groupId>com.lyncode</groupId>
+        <!-- New and Improved GDCC XOAI library! --> 
+        <dependency>
+            <groupId>io.gdcc</groupId>
             <artifactId>xoai-common</artifactId>
-            <version>4.1.0-header-patch</version>
+            <version>${gdcc.xoai.version}</version>
         </dependency>
```
poikilotherm, line 390: Not sure it's necessary to include the common dep here, as it is already a dep of the data & service provider.

Yeah, I know, one should add all transitive deps here that one is using within its own code, but this is all the same library...

pom.xml
```diff
@@ -382,52 +382,21 @@
...
+            <groupId>io.gdcc</groupId>
             <artifactId>xoai-data-provider</artifactId>
-            <version>4.1.0-header-patch</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>log4j</groupId>
-                    <artifactId>log4j</artifactId>
-                </exclusion>
-            </exclusions>
+            <version>${gdcc.xoai.version}</version>
```
poikilotherm, line 394: Would it make sense to release a BOM for this lib and use it in `<dependencyManagement>` to avoid these repeats here? (Following [JLBP-4](http://java.jonathangiles.net/JLBP-4))

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvesterServiceBean.java
```diff
@@ -380,6 +353,68 @@ private Long processRecord(DataverseRequest dataverseRequest, Logger hdLogger, P
...
+                tempOut.write(buffer, 0, bufsize);
+                tempOut.flush();
+            }
+
+            in.close();
+            tempOut.close();
+            return tempMetadataFile;
+        }
+
+        throw new IOException("Failed to download extended metadata.");
```
poikilotherm, line 392: IMHO this code is really hacky Java with C origin, also using some Exception anti-patterns.

May I suggest starting to use try-with-resources, considering [`InputStream.transferTo()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#transferTo(java.io.OutputStream)) and looking at the modern [Java included HTTP client](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/package-summary.html)? (See also https://www.mocklab.io/blog/which-java-http-client-should-i-use-in-2020/)

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvesterServiceBean.java
```diff
@@ -380,6 +353,68 @@ private Long processRecord(DataverseRequest dataverseRequest, Logger hdLogger, P
...
+            // check the status of that Delete - see if it actually succeeded
+            deletedIdentifiers.add(persistentIdentifier);
+            return;
+        }
+        hdLogger.info("No dataset found for " + persistentIdentifier + ", skipping delete. ");
+    }
+    
+    private static String getProprietaryDataverseMetadataURL(String baseURL, String identifier) {
+
+        baseURL = baseURL.replaceAll("/oai", "");
```
poikilotherm, line 410: If you don't expect the pattern to occur multiple times, you might want to avoid this. It's very slow compared to the other available options.

src/main/java/edu/harvard/iq/dataverse/harvest/server/OAIRecordServiceBean.java
```diff
@@ -332,17 +339,21 @@ public List<OAIRecord> findOaiRecordsBySetName(String setName, Date from, Date u
...
                 logger.fine("plain date. incrementing by one day");
-                until.setTime(until.getTime()+(24 * 3600 * 1000));
+                untilDate.setTime(untilDate.getTime()+(24 * 3600 * 1000));
             } else {
                 logger.fine("date and time. incrementing by one second");
-                until.setTime(until.getTime()+1000);
+                untilDate.setTime(untilDate.getTime()+1000);
             }
-            query.setParameter("until",until,TemporalType.TIMESTAMP); 
+            query.setParameter("until",untilDate,TemporalType.TIMESTAMP);
```
poikilotherm, line 356: **suggestion:** Please not the clock skewing is [done by the XOAI lib already](https://github.com/gdcc/xoai/blob/88b06ef3cb401373b158cb9b71b57286b59db2a3/xoai-data-provider/src/main/java/io/gdcc/xoai/dataprovider/request/RequestBuilder.java#L239).

You can alter the skewing behavior by overriding the `RepositoryConfiguration.skewUntil(()` method if necessary.

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -233,150 +232,30 @@ private void processRequest(HttpServletRequest request, HttpServletResponse resp
...
-                XmlWriter xmlWriter = new XmlWriter(response.getOutputStream());
-                xmlWriter.write(handle);
-                xmlWriter.flush();
-                xmlWriter.close();
-            }
+
+            XmlWriter xmlWriter = new XmlWriter(response.getOutputStream(), repositoryConfiguration);
+            xmlWriter.write(handle);
+            xmlWriter.flush();
+            xmlWriter.close();
```
poikilotherm, line 244: **suggestion:** You might consider creating the writer within a try-with-resources to avoid any spilled beans.

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -233,150 +232,30 @@ private void processRequest(HttpServletRequest request, HttpServletResponse resp
...
         } catch (XMLStreamException xse) {
             logger.warning("XML Stream exception in Get; "+xse.getMessage());
             throw new ServletException ("XML Stream Exception in Get", xse);
-        } catch (XmlWriteException xwe) {
-            logger.warning("XML Write exception in Get; "+xwe.getMessage());
-            throw new ServletException ("XML Write Exception in Get", xwe);  
         } catch (Exception e) {
             logger.warning("Unknown exception in Get; "+e.getMessage());
             throw new ServletException ("Unknown servlet exception in Get.", e);
         }
```
poikilotherm, line 255: **nitpick**: I'm not sure if this might be an Exception anti-pattern here.

1) Is the ServletException exposed to a user, which will include the pesky internal details? (That would be an anti-pattern)
2) Is the ServletException logged by the app server, too? In that case: would it be wise to avoid double logging the same thing? (Which would also be an anti-pattern)

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -233,150 +232,30 @@ private void processRequest(HttpServletRequest request, HttpServletResponse resp
...
-    
-    private boolean isGetRecord(HttpServletRequest request) {
-        return "GetRecord".equals(request.getParameter("verb"));
-        
-    }
-    
-    private boolean isListRecords(HttpServletRequest request) {
-        return "ListRecords".equals(request.getParameter("verb"));
-    }
-
```
poikilotherm, line 379: **praise:** Yeah! Away with that! :partying_face:

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,314 @@
...
+    
+    private List<DataverseXoaiItem> list = new ArrayList<DataverseXoaiItem>();
+
+
+    @Override
+    public Item getItem(String identifier) throws IdDoesNotExistException {
+        // I'm assuming we don't want to use this version of getItem 
+        // that does not specify the requested metadata format, ever
+        // in our implementation - ? (L.A.)
+        throw new IdDoesNotExistException("Metadata Format is Required");
```
poikilotherm, line 60: **issue:** Please don't do this. There is exactly one valid use case for this: when a [`ListMetadataFormats`](http://www.openarchives.org/OAI/openarchivesprotocol.html#ListMetadataFormats) request is sent, it might carry an identifier to retrieve all available metadata formats for this specific item.

Please be aware that you extended the method signature: you provide a return type "Item", whereas the interface only request an "ItemIdentifier" (which is inherited by "Item", so this works).

The available metadata formats for the item are provided via the `Context`. It will default to add all formats the context knows about, but you can add more specific things if necessary.

I'd strongly advise against throwing the exception as is now.

src/main/java/edu/harvard/iq/dataverse/settings/SettingsServiceBean.java
```diff
@@ -522,7 +522,21 @@ Whether Harvesting (OAI) service is enabled
...
+         * these by SystemConfig. 
+         */
+        oaiServerMaxListIdentifiers,
+        oaiServerMaxListRecords,
+        oaiServerMaxListSets,
+        /* 
+         * The OAI "repository name", shown by the OAI Identify verb, needs
+         * to be configurable as well. 
+         */
+        oaiServerRepositoryName
```
poikilotherm, line 539: **nitpick:** Please make the writing consistent and use a capital letter within the `Key` enum.

**though:** Can we pretty please avoid adding even more database settings for things that definitely are static things? #7000 got pretty far already.

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -1204,4 +1204,31 @@ public Map<String, String[]> getCurationLabels() {
...
+                int intFromSetting = Integer.parseInt(settingValue);
+                if (intFromSetting > 0) {
+                    return intFromSetting;
+                } 
+            } catch (NumberFormatException ex) {
+                // will use the default
+            }
+        }
+        return defaultValue;
+    }
```
poikilotherm, line 1233: **though:** can we please use #7000 for this?

src/main/java/edu/harvard/iq/dataverse/harvest/server/OAIRecordServiceBean.java
landreev: Sure, we discussed this of course, I forgot to remove it.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
landreev: I didn't think/remember about this use case (ListMetadataFormats with the identifier parameter). Makes sense.
(Why is the method called getItem() though; and not getItemIdentifier(), for consistency with the other method?)

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -69,6 +70,18 @@ public class OAIServlet extends HttpServlet {
...
+    @ConfigProperty(name = "dataverse.oai.server.maxidentifiers", defaultValue="100")
+    private Integer maxListIdentifiers;
+    
+    @Inject
+    @ConfigProperty(name = "dataverse.oai.server.maxsets", defaultValue="100")
+    private Integer maxListSets;
+    
+    @Inject
+    @ConfigProperty(name = "dataverse.oai.server.maxrecords", defaultValue="10")
+    private Integer maxListRecords;
```
poikilotherm, line 84: **praise:** All hail to mighty MPCONFIG!

(Might be refactored with what is in #8823, but it's a great start!!! :partying_face: )

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
landreev: Sure, will do.

pom.xml
landreev: This one I'm not sure about honestly, if it's worth bothering.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItem.java
```diff
@@ -1,40 +1,37 @@
...
  *
  * @author Leonid Andreev
  * 
- * This is an implemention of an Lyncode XOAI Item; 
+ * This is an implemention of a Lyncode/DSpace/gdcc XOAI Item.
  * You can think of it as an XOAI Item wrapper around the
  * Dataverse OAIRecord entity.
  */
-public class Xitem implements Item {
+public class DataverseXoaiItem implements Item {
```
poikilotherm, line 23: **nitpick:** shouldn't this class be `final`? (Yes, I'm a huge fan of non-modifiable by default)

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItem.java
```diff
@@ -51,19 +48,28 @@ public Dataset getDataset() {
...
         this.dataset = dataset; 
         return this; 
     }
     
+    private List<About> about;
+    
     @Override
     public List<About> getAbout() {
-        return null;
+        return about;
```
poikilotherm, line 60: **nitpick:** Shouldn't this return an unmodifiable list to avoid changing the objects? (e. g. via `Collections.unmodifiableList()`)

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItem.java
```diff
@@ -1,40 +1,37 @@
...
     
-    public Xitem(OAIRecord oaiRecord) {
+    public DataverseXoaiItem(OAIRecord oaiRecord) {
         super();
         this.oaiRecord = oaiRecord;
         oaisets = new ArrayList<>();
         if (!StringUtil.isEmpty(oaiRecord.getSetName())) {
             oaisets.add(new Set(oaiRecord.getSetName()));
         }
+        about = new ArrayList<>();
```
poikilotherm, line 32: **thought:** Maybe move this to the declaration + make the list `final`?

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiSetRepository.java
```diff
@@ -21,12 +15,12 @@
  *
  * @author Leonid Andreev
  */
-public class XsetRepository implements SetRepository {
-    private static Logger logger = Logger.getLogger("edu.harvard.iq.dataverse.harvest.server.xoai.XsetRepository");
+public class DataverseXoaiSetRepository implements SetRepository {
+    private static Logger logger = Logger.getLogger("edu.harvard.iq.dataverse.harvest.server.xoai.DataverseXoaiSetRepository");
```
poikilotherm, line 19: **nitpick:** Should be final.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+ * representing harvestable local datasets and translates them into 
+ * XOAI "items". 
+ */
+
+public class DataverseXoaiItemRepository implements ItemRepository {
+    private static Logger logger = Logger.getLogger("edu.harvard.iq.dataverse.harvest.server.xoai.DataverseXoaiItemRepository");
+    
+    private OAIRecordServiceBean recordService;
+    private DatasetServiceBean datasetService;
+    private String serverUrl;
```
poikilotherm, line 43:
```suggestion
    private final OAIRecordServiceBean recordService;
    private final DatasetServiceBean datasetService;
    private final String serverUrl; 
```

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+    private String serverUrl; 
+
+    public DataverseXoaiItemRepository (OAIRecordServiceBean recordService, DatasetServiceBean datasetService, String serverUrl) {
+        super();
+        this.recordService = recordService;
+        this.datasetService = datasetService;
+        this.serverUrl = serverUrl; 
+    }
+    
+    private List<DataverseXoaiItem> list = new ArrayList<DataverseXoaiItem>();
```
poikilotherm, line 52:
```suggestion
    private final List<DataverseXoaiItem> list = new ArrayList<>();
```

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+        // should be available for this record. 
+        
+        List<OAIRecord> oaiRecords = recordService.findOaiRecordsByGlobalId(identifier);
+        if (oaiRecords != null && !oaiRecords.isEmpty()) {
+            DataverseXoaiItem xoaiItem = null; 
+            for (OAIRecord oaiRecord : oaiRecords) {
+                // We can return the first *active* record we find for this identifier. 
+                if (!oaiRecord.isRemoved()) {
+                    xoaiItem = new DataverseXoaiItem(oaiRecord);
+                    return xoaiItem;
```
poikilotherm, line 70:
```suggestion
                    return new DataverseXoaiItem(oaiRecord);
```

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+    public ItemIdentifier getItem(String identifier) throws IdDoesNotExistException {
+        // This method is called when ListMetadataFormats request specifies 
+        // the identifier, requesting the formats available for this specific record.
+        // In our case, under the current implementation, we need to simply look 
+        // up if the record exists; if it does, all the OAI formats that we serve
+        // should be available for this record. 
+        
+        List<OAIRecord> oaiRecords = recordService.findOaiRecordsByGlobalId(identifier);
+        if (oaiRecords != null && !oaiRecords.isEmpty()) {
+            DataverseXoaiItem xoaiItem = null;
```
poikilotherm, line 65:
```suggestion
```

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+                            // TODO: Consider an alternative - instead of throwing 
+                            // an IdDoesNotExist exception, mark the oaiRecord as 
+                            // "deleted" and serve it to the client (?). For all practical
+                            // purposes, this is what this oaiRecord represents - it's 
+                            // still in the database as part of an OAI set; but the 
+                            // corresponding dataset no longer exists, because it 
+                            // must have been deleted. 
+                            // i.e.
+                            // xoaiItem.getOaiRecord().setRemoved(true);
+                            break;
```
poikilotherm, line 114: **suggestion:** Why not simple throw the exception here to make it more explicit and reduce the danger of missing this when refactoring? (Is there a unit test for the behavior?)

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+                            // records have been pre-generated ("exported") should be 
+                            // served as "OAI Record". But, things happen. If for one
+                            // reason or another that cached metadata file is no longer there, 
+                            // we are not going to serve this oaiRecord. 
+                            // TODO: see the comment above; and consider
+                            // xoaiItem.getOaiRecord().setRemoved(true);
+                            // instead. 
+                            break;
+                        }
+                        xoaiItem.withDataset(dataset).withMetadata(metadata);
```
poikilotherm, line 133: **suggestion:** This very much belongs together - shouldn't all this be inside the `try` block? (Which also could throw directly, as mentioned before/above)

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItem.java
landreev: Sure. Or I could just delete these method and rely on the empty list implementation in Item.java; since we have no use for it yet.

landreev: Making the returned list unmodifiable will in fact make it impossible to add any Abouts to the item; since there are no withAbout() or setAbout() methods yet. Yes, I feel like just removing both the variable and the method from the class, until we have some need/use case for this "about" thing.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
landreev: I have a better solution.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+    public Item getItem(String identifier, MetadataFormat metadataFormat) throws HandlerException {
+        logger.fine("getItem; calling findOaiRecordsByGlobalId, identifier " + identifier);
+        
+        if (metadataFormat == null) {
+            throw new NoMetadataFormatsException("Metadata Format is Required");
+        }
+        
+        List<OAIRecord> oaiRecords = recordService.findOaiRecordsByGlobalId(identifier);
+        if (oaiRecords != null && !oaiRecords.isEmpty()) {
+            DataverseXoaiItem xoaiItem = null;
```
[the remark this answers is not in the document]
landreev: This could be a matter of taste. I feel like the way this is implemented here makes sense. This loop logic is not only about waiting to create the item until we know that there's at least one record available; it is also to add the extra sets to the item that was created and populated with metadata when the first record was encountered, when there are multiple records.
(there are multiple records in the database per identifier, one record for every set in which it occurs)

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
landreev: I will move the code duplicated between this method and getItems(), and this will be addressed in the process.

src/main/java/edu/harvard/iq/dataverse/harvest/server/xoai/DataverseXoaiItemRepository.java
```diff
@@ -0,0 +1,329 @@
...
+            ResultsPage<ItemIdentifier> result = new ResultsPage(resumptionToken, hasMore, xoaiItems, oaiRecords.size());
+            logger.fine("returning result with " + xoaiItems.size() + " items.");
+            return result;
+        }
+
+        return new ResultsPage(resumptionToken, false, xoaiItems, 0);
+    }
+
+    @Override
+    public ResultsPage<Item> getItems(List<ScopedFilter> filters, MetadataFormat metadataFormat, int maxResponseLength, ResumptionToken.Value resumptionToken) throws HandlerException {
```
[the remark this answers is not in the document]
landreev: ok.

poikilotherm (approved):
Looks good to me. Thanks for incorporating all those changes!

@landreev Shall I bake a 5.0.0-RC2 or do you want me to go "final" directly?

kcondon:
Issues so far:
1. Datacite, oai_datacite, and dataverse_json formats do not work. These also do not work in the prior version so may be out of scope. Datacite, oai_datacite log that those formats are not supported for import and there is an open issue to implement. The dataverse_json issue is unclear at the moment but it had worked in the original implementation with some isolated failing records but currently fails in existing implementation, not sure when it stopped working but could look at licensing.

landreev:
As for the first 2 formats under 1., we really don't support them (at least for now), so we should probably hide them from the pulldown and/or not allow people to select any formats except the supported ones. 
As for dataverse_json, just to confirm, are you saying that it's failing with the same license issue on develop/5.11 as well, not just on my branch?

landreev:
As for 
> 2. Harvesting as a server using a v5.11 client, same formats fail but json reports a different error: unable to find valid certification path to requested target

dataverse_json does require an additional http call (directly to the export API, to obtain json), outside of the OAI library. So it does not surprise me that the code that makes this call may be more sensitive to whether the cert is valid or not. But if the client is 5.11 in this context, then this is the existing behavior.

kcondon:
@landreev I wasn't saying that but since you asked I have now tested json using v5.11 as both client (demo) and server (prod) and get the same license error.

poikilotherm:
@kcondon sounds like this reveals some internal problems?! If there is anything I can do, please reach out (here/Matrix/Slack/Issue in XOAI/...)

landreev:
@poikilotherm 
> sounds like this reveals some internal network problems?! If there is anything I can do, please reach out (here/Matrix/Slack/Issue in XOAI/...)

I don't think it's network problems. It looks like it literally is what it says - a problem trying to validate the cert of the remote server. Which in turn could have happened for various reasons... I'm working with Kevin on taking these cert issues out of the equation, so that we can focus on testing the XOAI functionality proper. 
Regardless, if we need anything from you, we absolutely know where you live, yes!

landreev:
> I wasn't saying that but since you asked I have now tested json using v5.11 as both client (demo) and server (prod) and get the same license error.

OK, good to know. I'll figure this out, even if it's not specific to my branch. Again, to be able to take this error out of the equation/be able to test before vs. after. 
We were having various license-related import issues right after that PR that refactored everything license-related; with somewhat similar errors in the exception stack traces. Most of those were trivial - like, there were places in the code where the license server bean was not properly passed to the json parser (the new requirement). Although I thought we had fixed them all. Anyway,  I'll figure it out, will talk to Jim if necessary.

landreev:
So, about those dataverse_json import errors: 
The `Invalid license: CC0` error was there because some of the prod. datasets in the set you were testing did not get re-exported after 5.10, when the license was changed. All the datasets have now been re-exported, so this should no longer be an issue.
However, *every single* prod. dataset is going to be failing to import/harvest in dataverse_json by 5.11 clients now, with a different error: `Specified metadatalanguage not allowed`. This is on account of #8868. 

I'm guessing this means that the only way to test harvesting dataverse_json now is to have the new `metadataLanguages` setting defined on both the server and client; and have every dataset in the OAI set configured with a `metadataLanguage` value recognized by the client... which frankly feels like serious pain. So waiting to have a fix for #8868 merged may be a better idea. 

On the [somewhat] plus side, I still think that the very fact that these dataverse_json harvests are failing with the same import errors in my branch does mean that the harvesting framework is working as it should be. By the nature of the changes there, if it were broken, it would be failing to obtain the json, and never getting to the import stage.

landreev:
(Just put my name on the PR to update the status, as requested in standup)
The last comment above is still an accurate description of the status of the PR: the serious issues with harvesting JSON  that were identified during the first QA pass are unrelated to the PR itself. The issues in question should be addressed in #8868, which appears to be close to being done. Once it's merged, this PR can be re-tested. To the best of my knowledge everything else in it should be working; at least working as well, or better than it has been up until now, which is the goal of this PR. (for example, one small thing identified above - that we allow an admin to select a format from the pulldown that Dataverse does not know how to import - is a pre-existing issue; that I'm assuming we don't want to try to address during this pass).

landreev:
Since the json export PR has been merged, this issue is ready to be re-tested. 
Re-testing the functionality directly involving json export (i.e., harvesting in dataverse_json format) is fairly straightforward (just needs going through all the combinations of "old" and "new" on the client and server ends). But then I'm not sure if whoever ends up working on wrapping up this PR will want to re-test *everything* from scratch (?). 
I will work with this person to provide guidance/answer questions.

landreev:
OK, everything has been re-tested very carefully. 
Additionally, harvesting dataverse_json forrmat, that could not be fully tested during the first QA attempt mainly because of the export bug in 5.11 was thoroughly tested as follows: 

This branch simplified the process of harvesting of our proprietary json records; it skips the step where the client gets a hacked GetRecord entry from the server and extracts the url for obtaining the JSON record. It just calls the export API directly, since the url for it can be derived from knowing the server address. However, any remaining Dataverse clients not yet upgraded to this version will continue using the old model. So for backward compatibility, this new version of xoai-server still knows how to produce the old-style, custom version of GetRecord with the url embedded (which is unfortunate, because it's really hacky. But we will get rid of it in a future version, once it is safe to assume that few if any old-style clients harvesting this format are still out there). So all three possible combinations of harvesting dataverse_json had to be tested: new-to-new; old-to-new; new-to-old. Testing was done using my own dev. build as the client, and either dataverse-internal or the perf. cluster as the server.

PR #8758, opened 2022-06-01
Iqss/8757 file restriction with public install

### LINKED ISSUE #8757: Bug: File restriction still possible with setting PublicInstall
According to 
https://guides.dataverse.org/en/latest/installation/config.html?highlight=restrict#publicinstall
file restriction is deactivated when setting :PublicInstall to true.

We are using the PublicInstall setting not because it is a public installation, just because we want to deactivate file restrictions.

However, even if it is set to true, there's still a way to restrict the files via GUI.
- Select "File Options" (three dots) next to a file in the list.
- Select "Metadata"
- Click the checkbox of the file
- Click the "Edit" button and you can select "Restrict" and "Unrestrict"
grafik

haarli:
This PR deactivates the possibility to restrict files when the setting ":PublicInstall" is set to true, as it should be according to the guide.

Closes #8757


haarli:
Sure. Latest develop is merged now

qqmyers (approved):
Looks good. FWIW: I'm working in the same area for Harvard Data Commons and have tested these changes there.

qqmyers:
FWIW: an update: A more extensive change in this area is coming soon (PR by June 30) from the Harvard Data Commons Objective 1 work - adding code there to stop both restriction and embargo when data is public and redefining public as when the :PublicInstallation setting is set or when a dataset is using a store that is identified as public (a new jvm option for stores) - the latter allowing use of some stores (like public remote stores or globus accessible S3 buckets) where Dataverse cannot enforce access control (or restrictions or embargo) without requiring the whole instance to be public.

kcondon:
Hi @haarli I'm ready to test this but would you please refresh from develop? We've released a new version and have incremented the version numbers. Regards, Kevin

PR #8606, opened 2022-04-12
TDL/7493 Make BagGenerator threads configurable

qqmyers:
**What this PR does / why we need it**: Improves scaling of Archiving and allows it to use more or fewer threads depending on machine size

**Which issue(s) this PR closes**:

Closes #8602

**Special notes for your reviewer**:

**Suggestions on how to test this**: Configure archiving (e.g. with LocalSubmitToArchiveCommand), change the :BagGeneratorThreads setting and compare the load and/or watch the log to see how many files are being zipped in parallel.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: No

**Is there a release notes update needed for this change?**: multiple Bag/Archiving changes coming - perhaps one overall release note?

**Additional documentation**: guides updated

Issue #8602, opened 2022-04-12
Make BagGenerator parallelism configurable


qqmyers:
**Overview of the Feature Request**
The generation of archival Bags currently uses 8 threads to read datafiles when generating the zipped Bag. As part of work for TDL related to #7493, the associated PR lowers the default to 2 threads and makes threading configurable via a :BagGeneratorThreads setting which can be used with the existing archiver classes.

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
Interest in lowering the server load when archiving, particularly in the context of batch archiving as in #7493

**Any related open or closed issues to this feature request?** #7493

PR #8767, opened 2022-06-02
add 5.11 release notes

### LINKED ISSUE #8763: Release notes for 5.11
We'll use the same process as last time:

- #8570

pdurbin:

- Closes #8763

None right now but I might point out a few things later.

Make sure the upgrade instructions work.

No.

No.

**Additional documentation**:

None.


doc/release-notes/5.11-release-notes.md
```diff
@@ -0,0 +1,185 @@
...
+- Admin API endpoints have been added to find and delete dataset templates. (Issue 8600, PR #8706)
+- The BagIt file handler detects and transforms zip files with a BagIt package format into Dataverse data files, validating checksums along the way. See the [BagIt File Handler](https://guides.dataverse.org/en/5.11/installation//config.html#bagit-file-handler) section of the Installation Guide for details. (Issue #8608, PR #8677)
+- PostgreSQL 14 can now be used (though we've tested mostly with 13). PostgreSQL 10+ is required. (Issue #8295, PR #8296)
+- As always, widgets can be embedded in the `<iframe>` HTML tag, but the HTTP header "Content-Security-Policy" is now being sent on non-widget pages to prevent them from being embedded. (PR #8662)
+- URIs in the the experimental Semantic API has changed (details below). (Issue #8533, PR #8592)
+- Installations running Make Data Count can upgrade to Counter Processor-0.1.04. (Issue #8380, PR #8391)
+- PrimeFaces, the UI framework we use, has been upgraded from 10 to 11. (Issue #8456, PR #8652)
+
```
qqmyers, line 38: - Archival Bag generation is now more robust and the number of threads used when zipping datafiles into the archival bag is now configurable.

qqmyers: Given #8773, This should probably be:
The number of threads used when zipping datafiles into an archival bag is now configurable (Issue #8602, PR #8606)

pdurbin: @qqmyers thanks! I just pushed 29e480c3e4.

doc/release-notes/5.11-release-notes.md
```diff
@@ -0,0 +1,187 @@
...
+## Notes for Dataverse Installation Administrators
+
+### Identifying Datasets Requiring Terms of Access or Request Access Changes
+
+In support of the change to require either Terms of Access or Request Access for all Restricted Files, we have provided a query to identify datasets in your installation where at least one file has neither Terms of Access or Request Access enabled:
+
```
sekmiller, line 45: minor quibble: I think we want to say "neither....nor..." instead of neither or.

doc/release-notes/5.11-release-notes.md
```diff
@@ -0,0 +1,187 @@
...
+https://github.com/IQSS/dataverse/blob/v5.11/scripts/issues/8191/datasets_without_toa_or_request_access
+
+This will allow you to reach out to those dataset owners as appropriate.
```
qqmyers, line 48: What happens if people don't do this? Seems like a bigger warning, i.e. users with this problem won't be able to publish until they fix this (just reading the PR) might be useful. (If I understand there's a warning in the UI - is that true for the API as well?)

sekmiller: Right. Nothing happens to published versions, but owners are prevented from creating, editing or publishing draft versions until they update the terms/allow access requests.

doc/release-notes/5.11-release-notes.md
pdurbin: Good catch. Fixed in 24181f388d.

doc/release-notes/5.11-release-notes.md
pdurbin: I had a nice chat with @sekmiller and played with the feature a bit and made the warning banner appear. In c302cd626a I tried to improve the release notes.

@qqmyers @sekmiller et al., please take a look. This is the last thing before we move this PR to QA and cut 5.11! 🎉

doc/release-notes/5.11-release-notes.md
```diff
@@ -0,0 +1,195 @@
+# Dataverse Software 5.11
+
+This release brings new features, enhancements, and bug fixes to the Dataverse Software. Thank you to all of the community members who contributed code, suggestions, bug reports, and other assistance across the project.
+
+## Release Highlights
+
+### Terms of Access or Request Access Required for Restricted Files
+
+Beginning in this release, Restricted Files must have either Terms of Access or Request Access enabled. This change is to ensure that for each file in a Dataverse installation there is a clear path to get to the data, either through requesting access to the data or to provide context about why requesting access is not enabled.
```
sekmiller, line 9: One final suggestion - can we say here "Datasets with Restricted Files"? ( to match what is in the warning banner and since the terms or Request Access are at the dataset level)

pdurbin: Good idea. Thanks! Fixed in 85d400f3ec.

PR #8665, opened 2022-04-29
Fix support protocol http import OAI_DC

Esteban-Mtz:
**What this PR does / why we need it**: Support identifiers like "http://hdl.handle.net/" or "http://doi.org/" on import OAI_DC

**Which issue(s) this PR closes**:

- Closes #7808

**Suggestions on how to test this**: No functional changes.

PR #8770, opened 2022-06-02
HDC 3A: Update OAI-ORE contents to add collections, PID URL, brandname

### LINKED ISSUE #8768: HDC 3A: update OAI-ORE file to include information about parent collections/clean-up
In reviewing the current OAI-ORE file contents in the context of requirements for using it as part of an archival bag, a few issues came up:

- Archivers may want to know basic information about the collection(s) containing the dataset being archived
- URL forms of identifiers improve the ability to use them as semantic identifiers (also a DANS request)
- The OAI-ORE file still uses the root name instead of the InstallationBrandName (added after the OAI-ORE export was first created) as the repository name.

qqmyers:
**What this PR does / why we need it**: Updates the OAI-ORE file contents to include info about the collection(s) the dataset is in, using the URL form for PIDs, and using the InstallationBrandName.

Closes #8768

**Suggestions on how to test this**: Publish a version and confirm that the OAI-ORE export includes an "isPartOf" section with lists the hierarchy of collections the dataset is in, that the DOI for the dataset is in URI form, and that, with the :InstallationName set (a setting) that it is used for the includedInDataCatalog entry.

**Is there a release notes update needed for this change?**: part of #8611

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -204,7 +205,15 @@ public boolean generateBag(OutputStream outputStream) throws Exception {
         // The oremapObject is javax.json.JsonObject and we need com.google.gson.JsonObject for the aggregation object
         aggregation = (JsonObject) new JsonParser().parse(oremapObject.getJsonObject(JsonLDTerm.ore("describes").getLabel()).toString());
 
-        bagID = aggregation.get("@id").getAsString() + "v."
+        String pidUrlString = aggregation.get("@id").getAsString();
+        String pidString=pidUrlString;
+        //ToDo - put this conversion in GlobalId
```
scolapasta, line 210: Is there a reason to not do this todo now (seems small enough)?

qqmyers: Makes sense. FWIW: I knew #8665 was coming (as well as #8674) when I wrote this . With #8665 merged, I've gone ahead and moved the logic to the GlobalID class (updated for #8665) (and added a ToDo there for further cleanup that I hope to do in #8674).

scolapasta (approved):

kcondon:
Tested, works, need to find out what the integration error is about.

PR #8772, opened 2022-06-03
#165 ensure ordering of MDBs with Citation first

### LINKED ISSUE #165: Dataset Page: Two sets of breadcrumbs appearing
---

Author Name: **Kevin Condon** (@kcondon)
Original Redmine Issue: 3577, https://redmine.hmdc.harvard.edu/issues/3577
Original Date: 2014-02-25
Original Assignee: Elda Sotiri

---

From 2/21/14 Pilot Test Meeting Notes

Two sets of breadcrumbs appearing

sekmiller:
**What this PR does / why we need it**: For at least one dataset users are seeing custom metadata blocks appear before the Citation block on dataset view and edit. We always want to display the Citation block first among all mdbs.

Closes https://github.com/IQSS/dataverse.harvard.edu/issues/165 When creating, editing and viewing datasets in the NegotiationDataRepository collection, its custom metadatablock appears above the Citation metadatablock

**Special notes for your reviewer**: This fix relies on the name of the citation block similar to the "isRequired" on the metadata block object. There is no display order field so the sort for custom mdbs is id.

**Suggestions on how to test this**: verify that the citation block is always rendered first.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
@@ -18,6 +18,8 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
+import static java.util.stream.Collectors.toList;
 import javax.ejb.EJB;
 import javax.faces.view.ViewScoped;
 import javax.inject.Named;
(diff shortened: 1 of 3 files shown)


pdurbin (approved):
I didn't run the code but this looks fine to me. Metadata blocks should now be sorted by id and since the citation block is added first, it should appear at the top.

We don't have a "display sort" field but if we really need one in the future, we could add it.

I don't know why but Jenkins decided to do another test run despite there only being one commit on this PR. Maybe it's because I fixed the link to the issue that gets closed? The timing is about right. Anyway, the first test run was fine so I think we're good: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-8772/1/testReport/

sekmiller:
The code doesn't actually rely on the citation block being added first. It relies on the name being citation, similar to the way isRequired is applied to metadata blocks.

pdurbin:
Right, for the purposes of sorting, it looks for a block called "citation" and gives it a an id of negative one. The rest of the metadata blocks depend on the order in which they were added.

PR #8774, opened 2022-06-03
Upgrade from Payara 5 to 6

pdurbin:
**What this PR does / why we need it**:

Payara 5 is going EOL.

**Which issue(s) this PR closes**:

- Closes #8305

**Special notes for your reviewer**:

- Do not move this to QA until Payara 6 Alpha 3 is out and we can test with it.
- Do not move this to QA until Payara 6 Final is out and we can test with it.
- Do not move this to QA until we've made a final decision to go with the Rewrite/PrettyFaces library I built on my Mac and put in `local_lib`. Check https://github.com/ocpsoft/rewrite/issues/304 for progress toward a compatible release from upstream.
- Instead of d86e220 should we preserve backward compatibility of the SWORD API by inserting `attachment; ` for the user if it isn't supplied?
- The Perl installer has not been updated. Honestly, I'm ok with deleting it (and references to it, of course, if any).
- The release note snippet needs work. Upgrades have not been tested.

**Suggestions on how to test this**:

- Make sure API tests pass.
- A test server is coming, similar to the one for Payara 5: https://groups.google.com/g/dataverse-dev/c/5hfiXl4Hpd8/m/awzoSpupAQAJ
- Test everything. Test, test, test, just like the last major version change.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No

**Is there a release notes update needed for this change?**:

Yes, a stub is included.

**Additional documentation**:

None but references in the guides to Payara 5 have been switched to Payara 6.

Issue #8599, opened 2022-04-12
500 server error when adding a dataset in dataverse where default template has no associated terms


scolapasta:
This was reported by a user via RT:
They click Add Dataset on their dataverse and get a 500 error page.

The reason is that a null pointer is being thrown. After some investigation, it turns out that the dataset has a default template with no associated terms (termsofuseandaccess_id is null on the template) and that in turn causes the null pointer on line 163 of DatasetUtil:
```License license = dsv.getTermsOfUseAndAccess().getLicense();```

The workaround we suggested is to go and edit the template and associate terms (even if empty) to the template.

I did try to reproduce on demo with a new dataverse, but I wasn't able to create a template with no terms (I tried briefly) so it's possible that can't be done anymore and only affects some legacy templates (in prod, there were 89 such templates, 48 of which were set as a default template).

The fix could be one of:
• db update to create empty terms for any existing templates (and verify there is no way to create any such new ones)
• change line 163 to make sure dsv.getTermsOfUseAndAccess() is not null before calling getLicense()

scolapasta:
Looks like we may need to opt for some aspect of 1 or more complicated 2) as right now you also cannot edit the terms on these legacy templates (an error is thrown).

scolapasta:
So, to fix the issue for the one that was from the RT issue, I:
• Created a new dummy template (through ui)
• through db set the associated terms of the existing template to the ones created for the dummy template, and set the dummy template terms to null.
• deleted dummy template (through ui)

Note that means the template now has a cc0 license, so this would need to be reviewed to make sure that is what is desired. If not, the dataverse admin can then change it through the UI.

qqmyers:
From some quick testing at QDR, I don't think you can create new templates that don't have a termsofuseandaccess so this might be a job for flyway. FWIW: I played with this [commit](https://github.com/GlobalDataverseCommunityConsortium/dataverse/commit/51b743e0cc54018f5911440e8f4653ec1814a140) to start - basically null check TermsOfUseAndAccess 'everywhere' it is being used to get the license (everywhere doesn't include in the xhtml files). Without a flyway, this would cause users of an old template to get custom terms instead of the default license though.

mreekie:
Priority - the current work around is a manual fix.  We are getting RT tickets on this issue so this is a production impacting issue. It will likely impact other dataverse implementations outside of harvard.edu

mreekie:
sprint
- writeup a db script
- release to community.
- small

[...]

Issue #8735, opened 2022-05-24
New metadata blocks not appearing in previously created templates 


sekmiller:
**What steps does it take to reproduce the issue?** Add a template to a dataverse. Edit the dataverse to add a custom metadata block. Edit the existing template. The newly added metadata block does not appear in the template in edit mode.

* When does this issue occur? When you edit a template after adding a metadata block to the dataverse the template belongs to.

* Which page(s) does it occurs on? Edit Template page

* What happens? The newly added metadata block is not included in the template.

* To whom does it occur (all users, curators, superusers)? Anyone with edit Dataverse permissions

* What did you expect to happen? I would expect to be able to put in default values for all fields in all metadata blocks, even those blocks that may have been added after the template was created.

**Which version of Dataverse are you using?** 5.10 (also exists in current Dev environment.)

**Any related open or closed issues to this bug report?**

**Screenshots:**
Also, on Create Dataset if a template was created before an MDB was added that block shows up empty if the template is applied - that could hide required fields which would effectively prevent the Template from being used:
<img width="807" alt="Screen Shot 2022-06-09 at 3 12 48 PM" src="https://user-images.githubusercontent.com/2364928/172926267-70751585-33d6-408d-9a92-0f16fd01c15d.png">

No matter the issue, screenshots are always welcome.

To add a screenshot, please use one of the following formats and/or methods described here:

* https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests
*

PR #8775, opened 2022-06-03
GDCC/HDC 3B-Bi-directional LDN messaging, Template Instructions

### LINKED ISSUE #7882: Exploratory support for accepting related publication messages
As a proof-of-concept motivated by the interest at Harvard in allowing paper repositories to alert Dataverse instances when there are dataset references in the papers, this issue is a proposal to implement an API endpoint to accept such messages following https://notify.coar-repositories.org/ and https://www.w3.org/TR/ldn/ standards. To be consistent with the possibility of tying to other Linked Data Notification services in the future, the API will implement the POST call to a Dataverse Software 'inbox' that will accept a json-ld message asserting a relationship between a paper and a dataset. What happens when a message is received is TBD, but the POC will probably do something simple like send an admin message. In the future, more automation to add the paper as a 'Related Publication' to the dataset could be added.

qqmyers:
**What this PR does / why we need it**: This PR includes functionality created for the Harvard Data Commons project and includes:
- a workflow step to send Linked Data Notification (LDN) messages
- an LDNInbox API endpoint to receive incoming messages
- a new DATASETMENTIONED notification type
- a new DASH-NRS publicationIdType in the citation metadatablock
- a new capability to add custom instructions per field in Templates that will show when adding metadata to datasets using that template

Closes #7882

**Special notes for your reviewer**: Having just gone through the process to split HDC 3A work into multiple PRs, and messing things up at least once, I thought I would at least start by presenting this one as a ~coherent whole for discussion, and, if it should be broken apart, would like to get some guidance on how many parts and how best to do that.

At present, there are a couple minor additions that aren't technically part of this work:
- Some of the 3A changes to the OAI-ORE file are here as well and presumably will disappear from the difference as the 3A PRs are merged.
- the clean-up I did to avoid null pointer exceptions for #8599 - the commit is mentioned in that issue - is here as well. It's relatively straight forward, and would avoid NPEs, but a flyway update as discussed in the other issue may be preferred (or perhaps both make sense?). In any case, that's a single commit that can be moved, reverted, etc. as needed.

**Suggestions on how to test this**: The HDC 3B group has a test protocol for the basic LDN message exchanges as well as a test server set up to run this in parallel with a test instance of DASH. At merge time, we could make sure the code there is updated from this branch (i.e. after more merges with develop), so that it can be run with DASH. It would also be possible to craft/send a json-ld message to the LDNInbox manually, and for the outgoing workflow, one could configure the LDN step and at least check the log to see the message contents, etc. 

For the template instructions, it would be straightforward to create a template, click to add custom instructions, create a dataset using the template to see the custom instructions, clone a template to make sure the custom instructions are transferred to the template copy, etc.

Also note that the LDN messaging is intended to ultimately be compliant with the COAR Notify standard, but this doesn't yet exist in final form. For that reason, I think I'll document the LDN send/receive as 'experimental/likely to change' for now.

Also also note that the PR includes a custom copy of the latest titanium json-ld library. The latest version has moved to EE9 and changed all the package names. The copy here is a duplicate except for package names and would go away as soon as #8774 goes through. I originally upgraded trying to resolve an issue in parsing the LDN ~COAR message format but ultimately had to make other changes to get it to work, so it is possible that the update isn't strictly necessary. (If there's a desire to remove it, I'd want to go back and check that the older version doesn't have security issues - it's been quite a while since we last updated.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: There are several additions to the UI, most of which were reviewed by the UI design team as development proceeded. In particular, the user notification messages from receiving an LDN message, the DASH-NRS entry in the citation metadatablock, the placement/style of custom instructions in the edit dataset view have been looked at. The one piece of new functionality is the mechanism to edit instructions in a template. The edit option was placed just above the individual data fields in the same place that the resuls appear for metadata editing in the dataset.

**Is there a release notes update needed for this change?**: Yes - probably mentions of the LDN send/receive and the template instructions as two features.

**Additional documentation**: Lots of 3B documentation exists. I still need to document the LDN send/receive and the template instructions in the guides.


src/main/resources/db/migration/V5.11.0.1__hdc-3b.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE usernotification ADD COLUMN IF NOT EXISTS additionalinfo VARCHAR;
```
landreev, line 1: Please note that we already have a flyway script using V5.11.0.1 in the develop branch (V5.11.0.1__8605-support-archival-status.sql)...

src/main/resources/db/migration/V5.11.0.2__hdc-3b2-template-instructions.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE template ADD COLUMN IF NOT EXISTS instructions TEXT;
```
landreev, line 1: ... and, per a dv-tech slack thread, there's a PR from an outside developer already in QA (https://github.com/IQSS/dataverse/pull/8793) that has a flyway script numbered V5.11.0.2.

qqmyers: updated

landreev (approved):

pdurbin:
Just passing through, not the actual reviewer, but...

- No docs?
- The SQL scripts need to be renamed.

/me ducks

pdurbin:
Passing through again, haven't had a chance to dig into the docs (thanks!) but I noticed a couple API tests failing at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-8775/19/testReport/

kcondon:
Issue found with templates: can add instructional text to template fields and that text is copied when copying the template but can no longer edit template metadata, the fields are not displayed to edit.

qqmyers:
Not sure I understand. After adding template instructions, I can save, copy to create a new template, see the custom instructions, save, and come back to edit that copy and still see the custom instructions.

kcondon:
@qqmyers I can test it again but reediting the template after saving is what seems to be the problem. No fields appear, only the save button. Copying works fine, adding/removing the instructions during the initial edit works fine.

src/main/java/edu/harvard/iq/dataverse/Template.java
```diff
@@ -256,26 +268,31 @@ public void setMetadataValueBlocks() {
         metadataBlocksForView.clear();
         metadataBlocksForEdit.clear();
         List<DatasetField> filledInFields = this.getDatasetFields(); 
+
+        Map<String, String> instructionsMap = getInstructionsMap();
         
         List <MetadataBlock> viewMDB = new ArrayList<>();
-        List <MetadataBlock> editMDB=this.getDataverse().getMetadataBlocks(true);
+        List <MetadataBlock> editMDB=this.getDataverse().getMetadataBlocks(false);
```
qqmyers, line 275: Note this line is a fix for a bug found in testing that appears to come from #8837. With true, this call only gets metadatablocks from the current collection and, if it isn't the metadatablockroot/inherits which blocks to use from its parent(s), then no blocks are returned. Using false (or getMetadataBlocks() which redirects to getMetadataBlocks(false)) the call recurses up through parents until it finds the one where the metadatablocks are defined.

Testing should check the cases from #8735 to make sure this doesn't revert anything. As far as I can tell, selecting a new block in a parent, or in the current collection and then editing an existing template does show the newly selected block as well.

src/main/java/edu/harvard/iq/dataverse/Template.java
```diff
@@ -256,26 +268,31 @@ public void setMetadataValueBlocks() {
...
             
+        //The metadatablocks in this template include any from the Dataverse it is associated with 
+        //plus any others where the template has a displayable field (i.e. from before a block was dropped in the dataverse/collection)
         viewMDB.addAll(this.getDataverse().getMetadataBlocks(true));
```
qqmyers, line 279: My guess is that this call is no longer needed. As is it gets no blocks unless the current one is the metadatablockroot and it is the loop below that adds the right blocks. Looking back at #8837 where this was added, there used to be one list of blocks for edit and view, which is why getting the available blocks was needed. Now that this variable is just for view, I don't think getting the possible blocks is needed. (I also don't think it hurts in that for view, only the blocks with defined fields appear anyway, so having extra blocks in viewMDB doesn't break anything.

kcondon:
Note, release notes are composite here: #8894

Issue #8598, opened 2022-04-12
Feature Request/Idea: Provide tips on writing .rst (reStructuredText) files


pdurbin:
**Overview of the Feature Request**

Provide tips on writing .rst files in the context of our Sphinx-based guides. I find reStructuredText (.rst) to be quirky and intuitive, especially compared to Markdown.

**What kind of user is the feature intended for?**

Documentation writers. Docs are required for most pull requests.

**What inspired the request?**

I simply could not remember or quickly find how to add a custom title to a `:doc:` link.

Finally, I found https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-doc-role which explains that to do this:

    :doc:`Custom title </guides/intersphinx>`

We can add tips like this to our existing "Writing Documentation" page at https://guides.dataverse.org/en/latest/developers/documentation.html

Topics to cover:

- making links
  - linking to other pages
  - creating and linking to custom anchors
  - linking to external websites
 - ...
 - ...

**What existing behavior do you want changed?**

None.

**Any brand new behavior do you want to add to Dataverse?**

None.

**Any related open or closed issues to this feature request?**

None.

PR #8780, opened 2022-06-07
Add migration scripts to apps page

### LINKED ISSUE #8275: Feature Request/Idea: As someone trying to migrate datasets from one Dataverse to another, I'd like to know about scripts that could help

As someone trying to migrate datasets from one Dataverse to another, I'd like to know about scripts that could help. It turns out @lubitchv and @kaitlinnewson wrote some at https://github.com/scholarsportal/dataverse-migration-scripts and announced them at https://groups.google.com/g/dataverse-community/c/4yy3U5RtUAs/m/OLogk12NBgAJ

I'd suggest adding the link to a future version of https://guides.dataverse.org/en/5.8/admin/dataverses-datasets.html which looks like this:

Screen Shot 2021-12-03 at 10 47 00 AM

As @kaitlinnewson mentioned in the thread above, they don't plan to update them...

> Linking in the guides sounds great! I don't expect we'll be updating the scripts for future versions of DV unless we have a use case, but of course anyone could take these and update them.

... so we obviously shouldn't overpromise that they'll be maintained. They should be considered a starting point. We should also link to the thread, which has some discussion.

Example users roles: Superuser, Sysadmin

It comes up fairly often that people want to migrate datasets from one installation of Dataverse to another.

kaitlinnewson:

Adds migration scripts developed by Scholars Portal to the docs, on the "Apps" page.

Closes #8275

Rebuild the docs and view the "Apps" page.

No

No

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/api/apps.rst
+++ b/doc/sphinx-guides/source/api/apps.rst
@@ -87,6 +87,13 @@ Docker image from a code repository that follows the [reproducible executable en
 
 https://github.com/jupyter/repo2docker/blob/master/repo2docker/contentproviders/dataverse.py
 
+dataverse-migration-scripts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
Looks good. I did add an extra link. (I had to look at #8598 to remember the syntax.)

Thanks, @kaitlinnewson!

PR #8782, opened 2022-06-08
Proper handling of exceptions when exporting metadata

### LINKED ISSUE #8484: Failure to export dataset not detected, dataset marked as exported
This is a followup issue to #8452. The user's/installation's issue was addressed by modifying the export code to properly handle a specific metadata field becoming multiple (#8473). But there was one underlying bug that I would still like to address: despite the failure to export, the dataset was marked as exported; and then added to the OAI set. (The fact that the partial, incomplete export was left written on disk made it even worse). 

Must be a straightforward enough issue of properly intercepting exceptions.

landreev:

This is a simple PR that fixes exception handling during metadata exports, making sure the "last exported" time stamp does not get updated unless it was successful. 

Closes #8484

**Additional documentation**:


sekmiller (approved):

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -256,30 +256,23 @@ public boolean onSuccess(CommandContext ctxt, Object r) {
...
-     */
-    private void exportMetadata(Dataset dataset) {
-
+        // Metadata export:
+        
         try {
             ExportService instance = ExportService.getInstance();
             instance.exportAllFormats(dataset);
-
+            ctxt.datasets().updateLastExportTimeStamp(dataset.getId());
```
qqmyers, line 264: Why both this and setting the update time in exportAllFormats? 

Could/should it just be removed here? (Or is a ctxt.datasets().merge(dataset); or similar needed to get the update from the ExportService? )

Can you even get rid of updateLastExportTimeStamp? (I see it called three places, all after a call to exportAllFormats (andexportAllFormats is called a few more places where updateLastExportTimeStamp isn't called)).

landreev: I'm a bit stumped by this actually... With the line 264 above commented out the time stamp does not get updated. And I can't immediately tell why. Is there still something finicky about the transaction under which onSuccess() is called, in this context, of FinalizeDatasetPublicationCommand being executed as a "nested" command? 
Not sure about .merge(dataset), why it would be needed - as opposed to having all the changes done to the dataset getting saved automatically at the end of the transaction. ExportService is not an EJB itself, despite the name. 
OK, I would need to experiment some more to figure this out.

landreev: Uhm, I would like to just declare this out of scope - figuring out why that line is necessary. It has been there for a while.

qqmyers: I'd suggest at least a comment since you found it isn't doing anything / doesn't appear to anyway.

landreev: Actually, an explicit .merge() call there makes calling that method unnecessary.
Guessing it may be possible to get rid of it in the other 2 places.

landreev: ok.

PR #8793, opened 2022-06-10
Added new static facet to show metadata blocks with values

### LINKED ISSUE #8536: Feature Request/Idea: Add new static facet to show the metadata blocks types that are populated.
Allow filtering of all datasets that have any value for a particular metadata block.

We have a requirement to facilitate searching and improve visibility of Computation Workflow metadata.
Computational Workflow will be a new metadata block and we would like a way of selecting all datasets of this type.

As well, we would like this new static facet to be the most prominent and to appear at the top.

This new static facet will be configurable. We will allow administrators to decide what metadata blocks should be shown in this new facet.

All users

As part of the Harvard Data Commons project there is a requirement to tag datasets as computations workflows. This can be done with metadata blocks, but we would like to find these datasets directly from the side panel as we do with the main Dataverses/Datasets/Files checkboxes.

None

Add a new static search facet to Solr, based on metadata blocks with value for a dataset.

Relates to #8462 and #8463
This new issue supersedes these previous ones

adaybujeda:
New configurable facet to improve dataset discoverability.
This change creates Metadata blocks facet values for datasets with populated fields for a particular metadata block.

Closes #8536

This new facet needs to be configured for it to appear in the search side panel.
For example, to display Computational Workflow Metadata in the root Dataverse (and its children), use:

curl http://localhost:8080/api/dataverses/root/metadatablockfacets -X POST -d '["computationalworkflow"]' -H "X-Dataverse-key: API_KEY" -H "Content-type: application/json"

The payload is an array with names of the metadata blocks that you want to display.

No interface change. It just add a new facet called by default "Dataset Feature"

Yes

**Additional documentation**:
New documentation added to the guide in this PR


src/main/java/edu/harvard/iq/dataverse/Dataverse.java
```diff
@@ -322,8 +322,31 @@ public boolean isHarvested() {
...
     */
-    
-    
+    private boolean metadataBlockFacetRoot;
+
+    public boolean isMetadataBlockFacetRoot() {
+        return metadataBlockFacetRoot;
+    }
+
+    public void setMetadataBlockFacetRoot(boolean metadataBlockFacetRoot) {
```
sekmiller, line 331: Should this be set to false when you create a new Dataverse? Also for the root dataverse do we need to set it to true - since it doesn't have an owner? And it seems as if the only way to get a child dataverse to not inherit facets from its parent would be to give it its own facets via the api, correct?

adaybujeda: It should be `false` by default. Because it is a boolean primitive, the default value is `false`. I was following the same pattern as with `metadataBlockRoot` and `facetRoot`.

As for the root dataverse, technically, there is no need to set it to `true` because of the way `getMetadataBlockFacetRoot` was coded (checking for the owner). But if `metadataBlockRoot` and others are set to `true` in the root dataverse, then I guess it should be.

sekmiller: @adaybujeda - yes, sorry I was away for a few days. I just want to do another pass through the code, but I will probably send it along to QA shortly.

sekmiller (approved):
Looks good. thanks for the contribution.

sekmiller:
@adaybujeda Please update this branch with the latest from dev. Thanks!

kcondon:
@adaybujeda Hi, I'm getting a flyway error when attempting to deploy this. Can you refresh from develop once more?

adaybujeda:
Thanks @kcondon.

I have rebased from `develop` and renamed the DB update script to `V5.11.0.2__8536-metadata-block-facet.sql`

I hope it deploys now.

kcondon:
@adaybujeda Thanks, it deploys now. Would you mind providing an example metadata_blocks.json file? I see the example values sentence and I recognize that mirrors the sentence from the preexisting facets api but it would be useful to have actual example files (for both) but you are only responsible for your api :)

adaybujeda:
Thanks @kcondon. Looking into the issues.

adaybujeda:
> 1 . Please add an example metadata_block.json file to the doc.

I have update the documentation to add the json files with the examples. I have updated the references to the files to be links to download the files. I saw several examples that looked similar. I have updated the facets API too.

Screenshot 2022-07-26 at 12 42 28

> 2 . Attempting to add metadata facet to root dv with non root admin fails with permissions error but still enables feature and adds facet to root.

I have added an explicit command to update the block facets. The problem before is that I was making an update to the dataverse object in the controller and the JPA library was automatically updating he DB regardless of the command errors.
With the new commands, the updates will only be done after the user verification process.

> There doesn't seem to be a way to disable the feature, once enabled? Using -X DELETE fails with HTTP 405 and setting a blank metadatablock entry fails with invalid block name.

Yes, I missed a delete function to remove this feature from a dataverse. This is now implemented. I have updated the PR description with more info and updated the docs

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteMetadataBlockFacetsCommand.java
```diff
@@ -0,0 +1,39 @@
...
+    private final Dataverse editedDv;
+
+    public DeleteMetadataBlockFacetsCommand(DataverseRequest aRequest, Dataverse editedDv) {
+        super(aRequest, editedDv);
+        this.editedDv = editedDv;
+    }
+
+    @Override
+    public Dataverse execute(CommandContext ctxt) throws CommandException {
+        editedDv.setMetadataBlockFacetRoot(false);
```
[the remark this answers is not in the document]
adaybujeda: Not really.
The next line: `editedDv.setMetadataBlockFacets(Collections.emptyList());` will clear the records in the `DataverseMetadataBlockFacet` table.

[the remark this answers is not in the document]
sekmiller: OK, I just looked at Dataverse.java and I see the cascade remove and orphan removal = true. Thanks!

kcondon:
@adaybujeda Thanks for working on this. I have tested it and the issues I've reported are addressed. Passing back to review as Gustavo wanted to take a look.

pdurbin:
`V5.11.0.2__8536-metadata-block-facet.sql` needs to be renamed because we released 5.11.1.

However, I'd suggest waiting for PR #8890 to be merged, which renames another SQL script.

Then, if develop is merged into this branch and no other SQL scripts are added, the script in this PR can be renamed to this:

`V5.11.1.2__8536-metadata-block-facet.sql` (since the other PR is `.1`). If other scripts are merged first, we can rename to `.2` or `.3`, etc.

adaybujeda:
Thanks @pdurbin. Rebasing and renaming....

scolapasta:
@adaybujeda Sorry for the delay in getting back to you on this - I wanted to think thoroughly through how the API should work. I think what would be best is for:
• the edit API to allow for empty - this would allow you to override your parent and not have and blocks in the facet (as a hack you can do this now by selecting a block for which you do not have data, but that does feel hacky)
• instead of a delete API (since you can use edit to clear out the blocks), an API that allows you to reset the dataverse collection to inherit from it's parent. (perhaps this could just be an API that sets as root or not; though if you set as not, you would usually want to follow up calling the edit). This would reflect how it would work in a UI, once we add it there.

Are you avaialbe to make these updates?

adaybujeda:
Thanks for the comments @scolapasta.

> • the edit API to allow for empty 

This is already supported in the edit API by passing an empty array as the payload, `[]`. There is an example in the issue description. For what I understand in the use case you are describing, this is what passing the empty array would do.
Do you want me to update to allow for an empty payload as well?

> • instead of a delete API 

When @kcondon raised the issue of removing this feature for a Dataverse, I thought that the delete function was consistent with how the API works in general.

I can change the delete to be another update and set/unset the value for `metadataBlockFacetRoot`. My concern is that it could leave the DB in an inconsistent state as there could be blocks defined for the dataverse that this call will not clear.

scolapasta:
@adaybujeda 
• For empty array as the payload, that should work
• for the root changingh API, I think if you set root to be false (i.e. inherit), it should clear the blocks; and if you set to true (i.e. use your own), it should make a copy of what is in the parent. In both these cases, it should only do this if you are changing the value. Additionally, the edit, show only succeed if the root is set to false; otherwise it should send a message saying you first need to set the root boolean to true.

Does that work?

scolapasta:
Also, could you add the example for emptying to the guides? I thnk that would be helpful.

adaybujeda:
@scolapasta I have implemented the new requirements for the API.

This is ready for review.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateMetadataBlockFacetRootCommand.java
```diff
@@ -0,0 +1,64 @@
...
+                    DataverseMetadataBlockFacet metadataBlockFacet = new DataverseMetadataBlockFacet();
+                    metadataBlockFacet.setDataverse(editedDv);
+                    metadataBlockFacet.setMetadataBlock(blockFacet.getMetadataBlock());
+                    return metadataBlockFacet;
+                }).collect(Collectors.toList());
+            }
+            editedDv.setMetadataBlockFacets(newBlockFacets);
+        }
+
+        editedDv.setMetadataBlockFacetRoot(metadataBlockFacetRoot);
```
scolapasta, line 50: @adaybujeda  very small thing (and it could be wrong), should these lines be in the if clause so they only get called when you "switch" - I can't recall if the save or anything here would update any "update times" or anything like that.

scolapasta (commented):
@adaybujeda I left one small comment - if you could reivew and let me know what you think. Otherwise it's ready to go back to QA.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateMetadataBlockFacetRootCommand.java
adaybujeda: I left the update outside the `if` to refresh the `modificationTime` (which is done on save) even if there is no change in the data.

Do you want me to change?

scolapasta: I think that's my point - nothing is really getting modified, so let's change it. And acually one more related request then, in the API where you would call this command, first check to see if ti changed so that it doesn't call the command if it doesn't have to (so that it doesn't get logged)

adaybujeda: ok. Updating

adaybujeda: Update completed.
Updated as well the db script filename to: `V5.11.1.5__8536-metadata-block-facet.sql` as requested.

scolapasta (approved):

kcondon:
Hi @adaybujeda thanks for all the examples in how to test. I am running into a minor issue with the first one: curl http://localhost:8080/api/dataverses/root/metadatablockfacets/isRoot -X POST -d true -H "X-Dataverse-key: <real key removed>"
{"status":"ERROR","code":415,"message":"HTTP 415 Unsupported Media Type"}

adaybujeda:
I have updated the API docs with the missing `Content-Type` header

kcondon:
@adaybujeda So far, so good but when I try to disable a child collection from using metadatablock facets using the example, it complains: curl http://localhost:8080/api/dataverses/kc081822b/metadatablockfacets -X POST -d '[]' -H "X-Dataverse-key: <real key removed>" -H "Content-type: application/json"
{"status":"ERROR","message":"Dataverse: kc081822b must have metadata block facet root set to true"}

adaybujeda:
This is part of the new requirements set by @scolapasta. In order to disable a child collection from using metadatablock facets from its parent, you need to first set the facet root in that collection to true.

adaybujeda:
Yes, sorry. Updated the description for the disable curl example. Added the Content-Type header

kcondon:
@adaybujeda Ok, that's fine, I was able to figure it out but maybe there is a slight issue with the explanation in the doc. I'll elaborate, one moment.

adaybujeda:
@kcondon both statements are true.
Before a dataverse can set its own block facets, it must set its facet root to true. Regardless of the state of the parent.

When setting the facet root:
if it changes from true to false, it will clear the block facets (if any).
if it changes from false to true, it will copy the blocks from the first parent that is root (if any)
if it doesn't change, nothing is saved/copied

I will update the docs to refer to them as collections and clarify how to use the API

kcondon:
@adaybujeda OK, I think I get what you are saying and it mirrors the UI: when you inherit, you don't have a local copy of the list, just query the parent list, right? So setting isroot=false (I'm inheriting) clears the local facet list? When you become root, isroot=true, you stop inheriting but as a convenience in the UI and so here, I imagine, you make a local copy of the parent facet list so you can modify it?

adaybujeda:
Yes @kcondon, that is the implemented behaviour as specified by Gustavo.

I have updated the documentation:
 - Changed the references to dataverse to collection
 - Clarify that the root needs to be set to true before setting/clearing any blocks.

kcondon:
@adaybujeda Ok, that makes sense now. I am having trouble setting the metadatadata block facet for a child isroot=true. The command returns without error or status and doesn't change the facets and there is a server.log error:
[chg_facet_err.txt](https://github.com/IQSS/dataverse/files/9375950/chg_facet_err.txt)

adaybujeda:
Thanks @kcondon, could you paste the actual curl request to see If I can reproduce locally.
Not quite sure the error is  related with reading the request or rendering the response.

kcondon:
curl http://localhost:8080/api/dataverses/kc081822b/metadatablockfacets -X POST -d '[socialscience]' -H "X-Dataverse-key: <real key removed>" -H "Content-type: application/json"

kcondon:
@adaybujeda OK, everything test ok, thanks. I am seeing an integration test failure though: https://jenkins.dataverse.org/blue/organizations/jenkins/IQSS-Dataverse-Develop-PR/detail/PR-8793/15/tests
Screen Shot 2022-08-18 at 2 33 17 PM
Screen Shot 2022-08-18 at 2 33 36 PM

adaybujeda:
I am able to run this test locally and it runs successfully (I have tried 3 times).
Could this be a transient error?
Could you run it again?
@kcondon @scolapasta ^^

kcondon:
@adaybujeda OK, will try it again. The suite takes 42 mins to run.

adaybujeda:
Just for the record:

Screenshot 2022-08-18 at 20 04 38

kcondon:
@adaybujeda Thanks for all the effort.

PR #8810, opened 2022-06-20
7000 mpconfig for JVM settings

### LINKED ISSUE #7000: As a sysadmin I want to use MicroProfile Config API to configure my installation

### LINKED ISSUE #7680: Microprofile recursive update exception when trying to use the Alias/DBSetting sources
In #7661, I tried to use the new Alias and DBSetting ConfigSources to support using a normal db setting aliased to an underlying microprofile variable for two cases. One appears to work fine, but the other was consistently giving me a recursive update exception. The problem occurred in #7661 prior to [this commit](https://github.com/IQSS/dataverse/pull/7661/commits/6697e16c00d6fffdaa82956ef92789a3b2c5e47f), which includes the (hopefully) relevant part of the stack trace I was seeing in the commit notes.

At this point, I don't know if the problem was due to how I was using it, or if the problem is general- the only difference I can see between the setting that worked and the one that didn't is that the method retrieving the :InstallationName setting/associated mp variable is used in more places/possibly in async and regular code, but that could be a red-herring. 

From the stack trace, it looked like the issue was related to the Alias source getting called recursively (it's getValue method calls ConfigProvider.getConfig.getOptionalValue() when you request the new variable (that is aliased to an old variable name served by the DbSettings source)). That could be a design issue (as far as I can tell, #7661 is the first use of this code outside of tests) so I though it was worth an issue. If it just innapropriate use, it might be worth some guidance in the dev docs for microprofile stuff.

**At this point, the issue doesn't affect any code that is/will be merged,** but it will probably appear again if we try to update other settings. I think #7661 prior to the commit above is ~working code (there are some changes after that commit needed to make everything work, but as of that commit Dataverse runs and nothing is broken except this issue, which can be triggered by publishing or reExporting via api), and I'm happy to answer questions if anything is unclear.

poikilotherm:
This will introduce MicroProfile Config usage instead of any `System.getProperty()` calls. This enables configuring the JVM settings via the different MPCONFIG in a backward compatible manner, but eases container usage a lot.

However, it will *not* interfere with Database Settings in this iteration, to have a more focused, smaller step to take. There is one exception: the Solr host and port will be retrievable from MPCONFIG as well as database. (`SystemConfig.getSolrHostColonPort()`)

Closes #7680
Closes #7000

Not ready yet.

Will provide unit tests. For more manual testing, you could use another config source with Payara.

Nope, this is sysadmin only.

Yes. Batteries will be included.

**Additional documentation**:
- This PR will include in-depth sysadmin documentation when ready
- This PR might include a mechanism to ensure future PRs will not use `System.getProperty()` by accident
- This PR will add developer documentation about retrieving and adding JVM settings in code


poikilotherm:
Note to self: find all `System.getProperty` properties we ask for via
`rg System.getProperty src/main -INU --no-heading | sed -e "s#.*getProperty(\"\?##" | sed -e "s#\"\?).*##" | sed -e "s#\", .*##" | sort | uniq -c | sort`

poikilotherm:
Closing in favor of smaller scoped pull requests to avoid smoking reviewer brains.

PR #8823, opened 2022-06-30
7000 mpconfig basic infra

### LINKED ISSUE #7680: Microprofile recursive update exception when trying to use the Alias/DBSetting sources

poikilotherm:

This will introduce the most basic infrastructure necessary for MicroProfile Config usage to replace any System.getProperty() calls - see also #7000. This enables configuring the JVM settings via the different MPCONFIG in a backward compatible manner (existing installations can keep their configuration but will receive infos in the logs about migrating). 

This will ease container and (to some extent) development usage a lot. However, it will not interfere with Database Settings for now.

*THIS PULL REQUEST IS THE MOST BASIC INFRASTRUCTURE FOR #8823, #8824, #8825, #8826, #8827, #8828*

Closes #7680

This pull request is basically just setting the JVM settings retrieval infrastructure in place. It doesn't touch any existing code yet (this will be done in succeeding small pull requests). It has unit tests added where necessary.

As there are no changes to existing functionality and the whole thing is still decoupled from all of our code in this PR, there isn't much to test. You can run the test included and play around with `@JvmSetting` for JUnit 5 tests.

Nope, this PR is only dev related.

Not from this PR IMHO. PRs following up that change actual code should contain hints if necessary.

**Additional documentation**:
Included.


poikilotherm:
Hey @scolapasta as requested, merged latest develop.

Is it fine to leave the other PRs for actual features untouched for now and rebase once this is merged?

scolapasta:
@poikilotherm sure, since you'll want to rebase then, that makes perfect sense. (especially since there's a decent chance we'll have 5.12 out by then and you'd have to do it an even extra time...)

pdurbin:
I'm basically ready to approve this PR but I thought I'd try the new `JvmSettings.<SETTING NAME>.lookup(...)` syntax it introduces.

I looked at PR #8824 and played around with our `getVersion` code. 

First I tried changing the method to this...

    public String getVersion(boolean withBuildNumber) {
        return JvmSettings.VERSION.lookup();
    }

... and added the following to domain.xml ...

     <jvm-options>-Ddataverse.version=awesome</jvm-options>

... but at http://localhost:8080/api/info/version I got and empty JSON object (`{}`) and `java.lang.IllegalStateException: Recursive update` in server.log.

Hmm. What did I miss? I noticed that PR #8824 also included this line...

    private static final Config config = ConfigProvider.getConfig();

... and while the `config` object is never used, this seemed to fix it. That is, the version from the API was "awesome". But why is that line necessary? Should we document it?

pdurbin:
@poikilotherm and I chatted. The bottom line is that I'm going to update to the version of Payara in PR #8949 (5.2022.3) to see if I can still reproduce the `IllegalStateException: Recursive update` error.

This time I reproduced it slightly differently.

First, I made the following change to exercise "lookup":

```
$ git diff
diff --git a/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java b/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
index bd27405fae..814b5311e7 100644
--- a/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
@@ -8,6 +8,7 @@ import edu.harvard.iq.dataverse.DvObjectContainer;
 import edu.harvard.iq.dataverse.authorization.providers.builtin.BuiltinAuthenticationProvider;
 import edu.harvard.iq.dataverse.authorization.providers.oauth2.AbstractOAuth2AuthenticationProvider;
+import edu.harvard.iq.dataverse.settings.JvmSettings;
 import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
 import edu.harvard.iq.dataverse.validation.PasswordValidatorUtil;
 import java.io.FileInputStream;
@@ -132,6 +133,10 @@ public class SystemConfig {
     // candidate for being moved into some kind of an application-scoped caching
     // service... some CachingService @Singleton - ? (L.A. 5.8)
     public String getVersion(boolean withBuildNumber) {
+
+        if (true) {
+            return JvmSettings.VERSION.lookup();
+        }
         
         if (appVersionString == null) {
```

As before I have "dataverse.version" defined in domain.xml:

        <jvm-options>-Ddataverse.version=awesome</jvm-options>

Then I built and deployed with `mvn clean && mvn package && scripts/dev/dev-rebuild.sh`.

Somewhat surprisingly, it worked! I saw "awesome" as the version:

$ curl http://localhost:8080/api/info/version
{"status":"OK","data":{"version":"awesome","build":null}}

But the I changed the version in domain.xml like this...

        <jvm-options>-Ddataverse.version=spectacular</jvm-options>

... and instead of seeing a new version, I got the same 500 error and `{}` as output and `IllegalStateException: Recursive update`. Here's the stacetrace: [server.log.mpconfig.txt](https://github.com/IQSS/dataverse/files/9507059/server.log.mpconfig.txt)

Again, I'll try again with a newer Payara.

pdurbin:
Upgrading from Payara 5.2021.5 to 5.2022.3 made the `IllegalStateException: Recursive update` error go away. I noted the need to upgrade in the release notes in dd27401.

This is ready for QA.

pdurbin (approved):
See comments in the PR. No real change happens and devs should be ready to use "lookup" once they upgrade Payara.

doc/release-notes/7000-mpconfig-support.md
```diff
@@ -0,0 +1,8 @@
+# Broader MicroProfile Config Support for Developers
+
+As of this release, many [JVM options](https://guides.dataverse.org/en/latest/installation/config.html#jvm-options)
+can be set using any [MicroProfile Config Source](https://docs.payara.fish/community/docs/Technical%20Documentation/MicroProfile/Config/Overview.html#config-sources).
+
+Currently this change is only relevant to developers but as settings are migrated to the new "lookup" pattern documented in the [Consuming Configuration](https://guides.dataverse.org/en/latest/developers/configuration.html) section of the Developer Guide, anyone installing the Dataverse software will have much greater flexibility when configuring those settings, especially within containers. These changes will be announced in future releases.
+
+Please note that an upgrade to Payara 5.2022.3 or higher may be required. Payara 5.2021.5 threw exceptions, as explained in PR #8823.
```
poikilotherm, line 8: I looked at the stack trace and went looking at [PayaraConfigL228](https://github.com/payara/Payara/blob/e3bee021061f7ded022b92419a1f92e0780d7e50/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java#L228) as seen in the log.

Guess what, this was fixed with payara/Payara#5431 and released with [Payara 5.2021.8](https://github.com/payara/Payara/releases/tag/payara-server-5.2021.8).

Should we add that detail?

```suggestion
Please note that an upgrade to Payara 5.2021.8 or higher is required to make use of this. Payara 5.2021.5 threw exceptions, as explained in PR #8823.
```

pdurbin: Sure! Committed! Thanks!

pdurbin:
At standup I said I'm basically ready to merge this (maybe after poking a bit more) but as the reviewer I'm happy to let someone else take another look and merge it. As people free up, someone might grab it.

pdurbin:
At standup I said I'm still ready to merge this so I assigned myself. Sounds like we want to merge another PR first.

PR #8824, opened 2022-06-30
7000 mpconfig version

poikilotherm:
By simplifying the way to set the version and build number, the container setup is already simplified. There is no need to include some special Maven files anymore, as Maven takes care of providing the version number via the META-INF/properties file.

Relates to #7000

None.

It is a good and simple start to see how the pieces laid out in #8823 are coming together in a simple first usage.

Play around with the settings described in the little dev docs section and watch it work. Or take a look at the nice test example, also using the new shiny `@JvmSetting` test extension!

Not really.

Maybe? Dunno. Please leave comment. IMHO this is very much dev-only.

**Additional documentation**:
Included.


src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -50,6 +46,7 @@
 public class SystemConfig {
 
     private static final Logger logger = Logger.getLogger(SystemConfig.class.getCanonicalName());
+    private static final Config config = ConfigProvider.getConfig();
```
qqmyers, line 49: Since JVMSettings calls this internally in the lookup/lookupOptional calls, is this needed here?

src/test/java/edu/harvard/iq/dataverse/util/SystemConfigTest.java
```diff
@@ -1,13 +1,46 @@
...
+    @JvmSetting(key = JvmSettings.BUILD, value = "FOOBAR")
+    void testGetVersionWithBuild() {
+        // when
+        String result = systemConfig.getVersion(true);
+    
+        // then
+        assertTrue(result.startsWith("100.100"), "'" + result + "' not starting with 100.100");
+        assertTrue(result.contains("build"));
+        
+        // Cannot test this here - there might be the bundle file present which is not under test control
```
qqmyers, line 40: A bundle entry would affect the real app the same way, right? So should this test be here to assure the default app is not shipping a bundle file that's going to override this setting? Or should the code prioritize the mpconfig value over any hardcoded bundle entry (that would allow the test to work, but is it also a better choice for the app - should you have to delete a bundle entry to be able to control build number? I'm asking and not recommending - you/others probably have a better sense of whether a bundle entry should have precedence or just be a default.)

qqmyers (approved):
Looks fine to me. I raised a couple minor questions for @poikilotherm where he might want to make changes. I'm approving since I don't think either thing would require re-review if he does make changes.

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
poikilotherm: Good catch! Removed with [4b60983](https://github.com/IQSS/dataverse/pull/8824/commits/4b60983e360b3ee4b5a50535b769852fc9ea67ef)

src/test/java/edu/harvard/iq/dataverse/util/SystemConfigTest.java
pdurbin: @poikilotherm my first thought is that this shouldn't be at the top of the community backlog if there's still stuff to figure out! 😄

poikilotherm: 😲 

Why is that? It's the first time this code is getting a review and raised a question from a core member. I was fine with the code as is, it works as designed and tested. Maybe I'm getting the scope of "review" wrong?

pdurbin: When I run this test I get this as the result:

    100.100 build pdurbin

This is because I have a file here:

```
$ cat src/main/java/BuildNumber.properties
build.number=pdurbin
```

The test passes just fine, which is good! As a developer I should be free to specify a custom build number.

> So should this test be here to assure the default app is not shipping a bundle file that's going to override this setting? 

No. In fact, we do ship a BuildNumber.properties in each war file. I'm not sure why we would assert that we don't do this but I'm probably misunderstanding what's being discussed above.

In short, I think we should get this into a sprint so core team members can focus on a proper review while it's in a sprint. Again, it sounds like @poikilotherm is ok with the PR as-is. Great.

mreekie:
added to sprint Dec 15, 2022

pdurbin (approved):
I didn't test this but the code makes sense. Should work! I did tweak the docs.

PR #8827, opened 2022-06-30
7000 mpconfig files basic

poikilotherm:
Replace all lookups of the `dataverse.files.directory` with MPCONFIG and provide a sane default `/tmp/dataverse` (as seen with some of the former usages). Also add lots of detail docs to the config guide about this setting.

This PR is fully backward compatible with any existing configurations as it does no renaming or behaviour changes (except for the default, but every installation out there is very likely not relying on a default for this setting!).

Relates to #7000

None.

It is the next simple example to see how the pieces laid out in #8823 are coming together.

Play around with the settings described in the little dev docs section and watch it work. 

Not really.

Maybe? Dunno. Please leave comment. IMHO this is very much dev-only.

**Additional documentation**:
Included.


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1415,7 +1417,19 @@ We are absolutely aware that it's confusing to have both ``dataverse.fqdn`` and
...
 
-This is how you configure the path Dataverse uses for temporary files. (File store specific dataverse.files.\<id\>.directory options set the permanent data storage locations.)
+Please provide an absolute path to a directory backed by some mounted file system. This directory is used for a number
+of purposes:
+
+1. ``<dataverse.files.directory>/temp`` after uploading, data is temporarily stored here for ingest and/or before
+   shipping to the final storage destination.
+2. ``<dataverse.files.directory>/sword`` a place to store uploads via the :doc:`../api/sword` before transfer
+   to final storage location and/or ingest.
+3. ``<dataverse.files.directory>/<PID Authority>/<PID Identifier>`` data location for file system imports, see
```
qqmyers, line 1427: Hmm - has this been tested? Just wondering if the store uses a different dir whether this is still the right place, or if they have to go in the dir for the store being used. From the code it looks like that is the case. If so, the effect here is that #3 should be dropped (it would just be another case for line 1431) but we should either fix or add an issue for changing the documentation for import.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1415,7 +1417,19 @@ We are absolutely aware that it's confusing to have both ``dataverse.fqdn`` and
...
+   :ref:`api-import-dataset`.
+4. ``<dataverse.files.directory>/googlecloudkey.json`` used with :ref:`Google Cloud Configuration` for BagIt exports.
```
qqmyers, line 1429: You're changing this in another PR right? Is this line dropped there? Or should it just get dropped here? (FWIW - not sure that anyone but QDR is using the Google archiving so we may not need to support the legacy option of looking in this dir.)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GoogleCloudSubmitToArchiveCommand.java
```diff
@@ -56,10 +56,11 @@ public WorkflowStepResult performArchiveSubmission(DatasetVersion dv, ApiToken t
             statusObject.add(DatasetVersion.ARCHIVAL_STATUS, DatasetVersion.ARCHIVAL_STATUS_FAILURE);
             statusObject.add(DatasetVersion.ARCHIVAL_STATUS_MESSAGE, "Bag not transferred");
             
-            try {
-                FileInputStream fis = new FileInputStream(System.getProperty("dataverse.files.directory") + System.getProperty("file.separator") + "googlecloudkey.json");
+            String cloudKeyFile = JvmSettings.FILES_DIRECTORY.lookup() + File.separator + "googlecloudkey.json";
```
qqmyers, line 59: Could you add to the docs your usual text about using an alias for this as you've done for other passwords?

qqmyers, line 59: This file should probably just be a json input that can be microprofiled/aliased as a password would be. A future update...

qqmyers (changes requested):
More conversions to microprofile. The documentation here raises an issue as to whether we have out-of-date info elsewhere in the guides (re: the dir to add files to when doing an import) that should be fixed somewhere (with the docs being added here corrected to match).

FWIW: @scolapasta, @kcondon, and I had a discussion about how to QA these mp conversions since they are all repeating the same pattern. I think the outcome was to just check that a JVMoption/mpconfig file, env variable works once on some PR and for the rest of the conversions just test that one mode works (i.e. the existing jvm option today) - versus checking that all options work for every setting that is updated in every PR.

mreekie:
added to sprint Dec 15, 2022

pdurbin:
@poikilotherm heads up that this PR has merge conflicts.

poikilotherm:
All set - forgot to push the merge....

pdurbin:
@poikilotherm as expected, more merge conflicts as other PRs are merged. Can you please resolve them?

pdurbin:
@poikilotherm merge conflicts and it looks like there are changes requested by @qqmyers. Can you please take a look?

doc/sphinx-guides/source/installation/config.rst
pdurbin: @poikilotherm and I talked this out and he will rewrite this.

doc/sphinx-guides/source/installation/config.rst
pdurbin: Yeah, let's leave it last in the list.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1495,6 +1497,8 @@ protocol, host, and port number and should not include a trailing slash.
   https://github.com/IQSS/dataverse/issues/6636 is about resolving this confusion.
 
 
+.. _dataverse.files.directory:
```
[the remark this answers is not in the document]
poikilotherm: Sorry - merge artifact. Resolved.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1503,16 +1507,15 @@ dataverse.files.directory
 Please provide an absolute path to a directory backed by some mounted file system. This directory is used for a number
 of purposes:
 
-1. ``<dataverse.files.directory>/temp`` after uploading, data is temporarily stored here for ingest and/or before
+1. ``<dataverse.files.directory>/<PID Authority>/<PID Identifier>`` is the subdirectory layout when using the target
```
qqmyers, line 1532: dataverse.files.directory isn't used for permanent file storage for file stores - there are new store specific options that may point to the same place. DCM is probably legacy w.r.t. multiple stores and still uses this dir.

qqmyers: Maybe something like this, and move it to the end since DCM is not uses/is deprecated?

 1. ``<dataverse.files.directory>/<PID Authority>/<PID Identifier>``  - The DCM feature for :doc:`../developers/big-data-support`
   is able to trigger imports for externally uploaded files from this area under certain conditions. (This directory may also be used by file stores for :ref:`permanent file storage <storage-files-dir>`, but this is controlled by other, store-specific settings.

poikilotherm: I integrated your suggestion into a new version. Please have another look. Thx!

qqmyers: LGTM

qqmyers (approved):

mreekie:
Sizing:
- for next sprint
- 3

poikilotherm:
🥳 another one merged! Thanks for keeping up the pace @kcondon

PR #8825, opened 2022-06-30
7000 mpconfig solr

poikilotherm:
**What this PR does / why we need it**:
For unknown reasons, the Solr endpoint has been made configurable via a Database Setting only. This is very cumbersome for testing and container usage. Providing simple defaults via MPCONFIG takes lots of this load away. It also properly decouples creating the connection to Solr from the Database connection.

This PR is fully backward compatible with the old behaviour - anyone already having the database setting will let that setting take precedence!

This PR also very much enhances the configuration of the Solr connection, as it allows for the first time to choose transport protocol, core name and/or path, which makes it much more compatible with Solr installations that differ from our classic installations. (It does, however, not yet add support for SolrCloud).

Relates to #7000
Closes #3212

**Which issue(s) this PR closes**:
None.

**Special notes for your reviewer**:
- [ ] Please give advice if the installer should be changed to not use the database setting anymore.
- [ ] Please give advice if a release note update should be included.

**Suggestions on how to test this**:
Play around with the settings described in the added docs and watch it work. Or take a look at the nice test examples, also using the new shiny `@JvmSetting` test extension!

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Not really.

**Is there a release notes update needed for this change?**:
Maybe? Dunno. Please leave comment.

**Additional documentation**:
Included.

PR #8828, opened 2022-06-30
7000 mpconfig pid doi hdl

poikilotherm:
This is a larger cleanup of many PID provider settings.

1. It decouples the settings from different provider from each other by adding new names and scopes.
2. It will reuse existing configuration values to maintain backward compatibility!
3. The docs have been consolidated and changed as necessary to line out more of the differences and which things are important for what
4. Sane defaults have been incorporated, pointing to test and non-production resources

Relates to #7000

None.

This should be done but might need discussion if we want changes to the installer. (Password aliases etc)

- [x] Missing Handle refactoring
- [X] Missing a release notes file explaining the changes
- [x] Maybe change installer to use new configuration names
- [x] HandleNet properties need aliases for old names!
- [x] Add tests for aliases to be functional (linking not testable before as no aliases)
- [X] ~~Maybe extend like done for Solr (#8825) to enable protocol, provider, authority choose etc via MPCONFIG in backward compatible manner? (Much easier config for containers and those settings are VERY static - these shall never change for an installation or you'll be in trouble. Database Settings were introduced for dynamically changing settings? :thinking: )~~ Too large for this PR.

Play around with the settings described in the docs section and watch it work. It's not much of a change despite the renaming.

Not really.

Not included. Let me know if we should write one now or latter (more mpconfig stuff coming). Technically, as the settings have aliases, nothing should change. There is no breaking change, just deprecated names.

**Additional documentation**:
Included.


poikilotherm:
Good catch by @landreev - the HandleNet properties are missing their aliases!

poikilotherm:
The ITs cannot run because Maven fails to install... Sending this to review anyway.

poikilotherm:
I suppose this is a size 10 thing. "3" seems not be enough time to review all the text I wrote. And folks might want to test a few things with the PID providers during QA.

mreekie:
sizing:
- was moved ahead during the last priority meeting

poikilotherm:
Just merged in latest `develop` which contains a lot of changes to the PID providers, merge conflicts solved!

mreekie:
new sprint:
 - Oliver just pulled this back for more work.
 - Moving back to off the board.
 - Sounds like this will likely need to be resized once Olivery figures out what's left.
 -

poikilotherm:
I'm putting this back on the table. @mreekie automation dropped this into "Ready for review", could you move it to the sprint column? Thx!

About sizing: this does not need a resizing, it's still a 10.

The overall added stuff was that Permalink provider is now configurable via MPCONFIG as well as the others.
And the EZID provider is created in any case, but no login is attempted if no username and password is given. This was not possible before because we reused the same property name for all places. As this is changed now, it can properly detect if to attempt login or not.

mreekie:
> I'm putting this back on the table. @mreekie automation dropped this into "Ready for review", could you move it to the sprint column? Thx! (I am not allowed to do this for some unclear reason...?)
> 
> About sizing: this does not need a resizing, it's still a 10.
> 
> The overall important added stuff was that Permalink provider is now configurable via MPCONFIG as well as the others. And the EZID provider is created in any case, but no login is attempted if no username and password is given. This was not possible before because we reused the same property name for all places. As this is changed now, it can properly detect if to attempt login or not. The rest are some cosmetic changes.

@poikilotherm This is fine. Let's leave it be.

src/main/java/edu/harvard/iq/dataverse/api/Pids.java
```diff
@@ -47,9 +48,13 @@ public Response getPid(@Context ContainerRequestContext crc, @QueryParam("persis
...
         }
-        String baseUrl = systemConfig.getDataCiteRestApiUrlString();
-        String username = System.getProperty("doi.username");
-        String password = System.getProperty("doi.password");
+
+        // FIXME: Even before changing to MPCONFIG retrieval, this was pinned to be DataCite specific!
+        //        Should this be extended to EZID and other PID systems like Handle?
+        String baseUrl = JvmSettings.DATACITE_REST_API_URL.lookup();
+        String username = JvmSettings.DATACITE_USERNAME.lookup();
+        String password = JvmSettings.DATACITE_PASSWORD.lookup();
```
pdurbin, line 56: This is fair. I added the PIDs API but hard coded it to DataCite. I would like to see it evolve to be useful across multiple PID providers.

pdurbin (approved):
@poikilotherm great work on refactoring and cleaning up the code. Off to QA!

@kcondon the code looks good but please advised that I only did the lightest of testing, just check the defaults for the FAKE and perma providers. Both worked just fine as shown in the screenshot below. I did not touch DataCite, Handle, or EZID.

Screen Shot 2023-03-31 at 5 06 38 PM

kcondon:
Issues found:
1. Existing, known good handle config fails with ui error and stack trace:
[hdl_create_err.txt](https://github.com/IQSS/dataverse/files/11141597/hdl_create_err.txt)
Seems like it really needs new hdl passphrase property.

poikilotherm:
@kcondon I fixed this with b7d64df, only looking up the passphrase if it is required. Hope that helps :smile:

poikilotherm:
Woo-hoo!!! And another one in! Thanks for testing in depth @kcondon and the positive review @pdurbin ! :party:

PR #8830, opened 2022-07-04
7000 mpconfig rserve

poikilotherm:
Now also changing the Rserve configuration to be MPCONFIG enabled. Also making sure updated settings would be used for every new file read during ingest.

None.

None?

There are no tests. How have test for Rserve integration been done so far?

Nope.

IMHO no.

**Additional documentation**:
Doc updates included.


mreekie:
added to sprint Dec 15, 2022

kcondon:
@poikilotherm @pdurbin Not sure how this should fail but previously, it uploaded the original file. Now it throws a server log error but leaves the user on the grayed out "action" screen and if you click on it, shows upload page with no files, so a silent error.

poikilotherm:
Hey @kcondon this is really strange... where is that "x" coming from? It's clearly not added by the code...

src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -8,6 +8,14 @@ dataverse.db.host=localhost
 dataverse.db.port=5432
 dataverse.db.user=dataverse
 dataverse.db.name=dataverse
+
+# RSERVE
+dataverse.rserve.host=localhost
+dataverse.rserve.port=6311
+dataverse.rserve.username=rserve
+dataverse.rserve.password=rserve
+dataverse.rserve.tempdir=/tmp/Rserve
```
landreev, line 17: `/tmp/Rserv` (without the second `e`) would be a better default - this is the default temp directory for Rserve (and that's what our guide says it defaults too; even though it appears that it was not true!)

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1542,27 +1542,60 @@ Can also be set via *MicroProfile Config API* sources, e.g. the environment vari
...
 ++++++++++++++++++++++++
 
-Temporary directory used by Rserve (defaults to /tmp/Rserv). Note that this location is local to the host on which Rserv is running (specified in ``dataverse.rserve.host`` above). When talking to Rserve, Dataverse needs to know this location in order to generate absolute path names of the files on the other end. 
+Temporary directory used by Rserve (defaults to /tmp/Rserv). Note that this
+location is local to the host on which Rserv is running (specified in
+``dataverse.rserve.host`` above). When talking to Rserve, Dataverse needs to
+know this location in order to generate absolute path names of the files on the
+other end.
+
+Defaults to ``/tmp``.
```
landreev, line 1595: The sentence above already says `defaults to /tmp/Rserv`, which is what we want really. I would also add "This setting is only needed if Rserve is configured to use a location different from the default /tmp/Rserv" to the paragraph above.

src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/rdata/RDATAFileReader.java
```diff
@@ -451,7 +434,13 @@ public RDATAFileReader(TabularDataFileReaderSpi originator) {
 
     super(originator);
     
-    
+    // These settings have sane defaults in resources/META-INF/microprofile-config.properties,
+    // ready to be overridden by a sysadmin. Every time a file would be read with this file reader,
+    // a new reader will be created, reading from the cached config source settings with minimal overhead.
+    this.RSERVE_HOST = JvmSettings.RSERVE_HOST.lookup();
+    this.RSERVE_PORT = JvmSettings.RSERVE_PORT.lookup(Integer.class);
```
poikilotherm, line 441: When a non-integer value is given, this code here leads to the problem described by @kcondon: 

> throws a server log error but leaves the user on the grayed out "action" screen and if you click on it, shows upload page with no files, so a silent error.

This problem never occured before, as the setting was never actually used (bug in implementation looking for "dataverse.ingest.rserve.port") and always defaulted to "6311".

How do we proceed? Maybe move the request builder initialization code to `init()` and cache it from there? At least that method is expected to throw an `IOException` (which is not the same), but might at least the calling `read()` function is called from places expecting exceptions.

kcondon:
So to keep track:
1. Fix doc for tempdir per Leonid
2. Port failure with non numeric in value returns user to Upload url with gray overlay rather than to dataset page.
3. Deleted user setting does not use default but fails the same as in 2.

doc/sphinx-guides/source/installation/config.rst
poikilotherm: Done.

src/main/resources/META-INF/microprofile-config.properties
poikilotherm: Done.

poikilotherm:
- [X] 1. Fix doc for tempdir per Leonid
- [ ] 2. Port failure with non numeric in value (x6311) returns user to Upload url with gray overlay rather than to dataset page.
- [X] 3. Deleted user setting does not use default but fails the same as in 2.

Need input for (2) from @landreev @pdurbin et al.

src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/rdata/RDATAFileReader.java
landreev: I'll be honest, I may not be fully grasping the seriousness of this issue. As in, "don't use a non-integer for this setting" does seem like an adequate solution, no? 
I mean, there are situations where it may not be obvious or intuitive, that a specific setting needs to be of a certain type. But a port, of all things? What else would anyone want to set it to - "six thousand three hundred twelve"? 
There's also the fact that it's fairly hard to imagine a practical situation where anyone would need to run Rserve on a different port.

landreev: (otherwise, if we want this to be handled cleanly, I'm guessing the setting rserve.port could be made a String, that will be converted to Integer at the time of use; with the non-integer exception properly intercepted with a rollback to the default 6311 - something like that?)

poikilotherm: Yes and no - the whole thing would need to be moved to something like the `init()` method, as the requests thing is cached in the `static {}` block.

We could of course also catch the exception and only  log it, then make the field read `null` and check for non-`null` in the code using it and throwing an IOException to prevent NPEs.

landreev: That, the above, would work. But then there's also an option of merging it as is. (this came up very briefly during the standup and we all seemed to be in agreement that we trusted you on how to proceed with handling this. so, up to you).

pdurbin:
@poikilotherm can you please resolve the merge conflicts? Thanks!

src/main/java/edu/harvard/iq/dataverse/ingest/tabulardata/impl/plugins/rdata/RDATAFileReader.java
poikilotherm: OK solved by defaulting to `6311` also when invalid integer given. Least impact, better UX.

poikilotherm:
Thanks for going through all the trouble with me @kcondon! Glad this is done. Another step! :heart:

PR #8831, opened 2022-07-05
#8820 render confirm delete button on delete group click

### LINKED ISSUE #8820: CRITICAL: Pressing Enter in Group Members assignment windows deletes the whole group
1. As an admin/curator in a TEST environment, click Edit > Groups > Create Group, and create a new user group for testing this issue.
2. Look up and add a user to the newly created test group, so that the user name displays in the Group Members field:

image

3. With the cursor placed to the right to the added user name, press Enter.

* When does this issue occur?
See above.

On the Group edit page, e.g.
https://demo.dataverse.org/manage-groups.xhtml?dataverseId=XX

* What happens?
The group is deleted, and the following confirmation message appears on screen:

image

To everyone have the rights to add users to an existing user group.

Not sure. I guess many curators would expect the added user to be saved, thus they want to use Enter as a shortcut for scrolling to the bottom of the page and click Save Changes.

The behavior was tested on version v.5.6 (test.dataverse.no) and version v.5.11 (demo.dataverse.org).

None to my knowledge.

**Screenshots:**
See above.

sekmiller:
**What this PR does / why we need it**: Fixes a bug where pressing "enter" on the create group popup deletes the group because it invokes the confirm delete popup continue button

Closes #8820 CRITICAL pressing entet in Group Members assignment deletes the Group

**Suggestions on how to test this**: try adding a group and press enter while the cursor is in the user text box. the group should be created - not deleted


qqmyers:
FWIW: When I first read this issue, I assumed it was similar to #7895 and that a fix like #8500 would be required.  Is the current PR still allowing the default button on the page to be clicked and just stopping that from having an effect (versus avoiding the keypress from triggering the action at all)?

sekmiller:
In this case according to Philipp, the preferred response to the enter key is to submit and save the group. In that other case since it was the whole dataset page, they wanted enter to be disabled.

pdurbin (approved):
I ran the code and it seems to do the job. Crazy that hitting enter was deleting groups!

A couple things I'll note:

- On develop (4e65f2f), in addition to the "Create Group" dialog I was also able to reproduce the "hit enter to delete" but in the "edit group" dialog. That is, got to "Group Members" and hit enter.
- Now hitting enter doesn't delete the group. Great! Perhaps it should trigger "Save Changes"? But this can wait for a future PR, I'd say.

pdurbin:
The Jenkins tests were passing but I went ahead and refreshed from develop.

PR #8834, opened 2022-07-07
8832 toa validation error on ds create

### LINKED ISSUE #8832: Problem with license selection and Terms of Access/'Enable access request' button
In 5.11, create a new dataset with metadata and files. Once created, go to the Terms tab and click 'Edit Terms Requirements' to select a different license. Pick a different license, and try to save. It doesn't give an error, it just doesn't save the changes or leave the 'Edit Terms Requirements' page.

* When does this issue occur?
When you try to save your changed license choice.

On all the newly created datasets.

* What happens?
It doesn't allow us to change the license because neither the 'Enable access request' button is ticked nor are there any 'Terms of Access for Restricted Files', even though there are no restricted files in the dataset and the option to select or fill them in is greyed out.
I can get it to work by first restricting a file and tick the 'Enable access request' button there, then unrestrict said file. If I then go to Terms and edit them to select a different license, it does work because the 'Enable access request' button (though again greyed out) is now ticked.
I don't get this issue in the demo.dataverse.org environment. There it already has the 'Enable access request' button ticked by default, even though it's greyed out. So, I'm wondering if it's a setting that we don't know about.

All users.

That I could change the license selection and save it. Or that the 'Enable access request' button is always ticked for newly created datasets and the licenses even when there are no files restricted.

5.11

no

**Screenshots:**
image

As you can see, the error is given because neither the 'Enable access request' button is selected nor Terms of Access are supplied. But I also can't select either of them, because they're greyed out due to a lack of restricted files in this dataset.

### LINKED ISSUE #8847: Terms of Access for Restricted Files should not be displayed if there are no more restricted files.
For a dataset that previously had restricted files, we wanted to remove the restriction because the preconditions for this no longer exists. To be able to edit the record with v5.11 at all (disabled edit button with warning "Info – Datasets with restricted files are required to have Request Access enabled or Terms of Access to help people access the data. Please edit the dataset to confirm Request Access or provide Terms of Access to be in compliance with the policy."), we first had to enter Terms of Access for Restricted Files. Unfortunately, after the restrictions were removed from all files, we could no longer edit the terms (because this option was disabled now). This leads to a somewhat unattractive display.

terms_files

How do we get rid of this again?

sekmiller:
**What this PR does / why we need it**: Fixes the Terms of Access validator to notice when there are no restricted files in the Dataset version

Closes #8832 Problem with license selection and Terms of Access/Enable access request button
Closes #8847 

**Suggestions on how to test this**: Make sure you can update the terms of use/license in a dataset that doesn't have restricted files. also verify that adding/editing templates Terms still works.
Also there is a minor ui change - see 8847 where terms of access won't be shown on the terms display tab if there are no restricted files.

**Is there a release notes update needed for this change?**: none


src/main/webapp/dataset.xhtml
```diff
@@ -1867,16 +1868,19 @@
                     
                     function testTOADatasetPage() {
                         var termsofAccessHidden, fileAccessRequestHidden;
+                        var hasRestricted;
                         try{
                             termsofAccessHidden = document.getElementById("datasetForm:tabView:termsofAccessHiddenLT").value;
                             fileAccessRequestHidden  = document.getElementById("datasetForm:tabView:fileAccessRequestHiddenLT").value;  
                         }
                         catch (error){
+                            alert("notdatasetform");
```
scolapasta, line 1877: this seems like it should not be here :)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -422,6 +422,14 @@ public Boolean isHasRestrictedFiles(){
         }
     }
     
+    public boolean getHasRestrictedFiles(){
```
scolapasta, line 425: I dont think you need this - being a boolean it should call the isHasRastrictedFiles method. But see my other note for if we even need the ui:hidden field.

scolapasta (changes requested):

src/main/webapp/dataset.xhtml
sekmiller: good catch, thanks

scolapasta (approved):

kcondon:
It's now working with the minor exception of the string showing 0 restricted files once there are terms of access but no restricted files. The terms of access in this case have not been rendered by the file count remains.

From the user who reported it:
[lmaylein](https://github.com/lmaylein) commented [4 days ago](https://github.com/IQSS/dataverse/issues/8847#issuecomment-1192285756)
Would be nice if it could not come to the case that it says:
There are 0 restricted files in this dataset.
Terms of Access for Restricted Files: ...

Issue #8821, opened 2022-06-29
Bean Validation errors calling file redetect api endpoint


matthew-a-dunlap:
During development of CORE2, I've been using pyDataverse to handle our Dataverse interactions.

One aspect of this is uploading files. We ran into #8344 which causes mime type to not be set. Because we want to support older installations, I'm shooting for a solution that doesn't require the fix pushed by @landreev (though I'm glad it exists!).

The solution I've tried is to call the `redetect` endpoint to get the correct file type. This works and there are no errors thrown in the response... BUT there are concerning messages now appearing in our logs. Note this is on our S3-based test Dataverse running 5.3:

[file_redetect_error.log](https://github.com/IQSS/dataverse/files/9014995/file_redetect_error.log)

I'm curious if anyone over at IQSS has insight as to what might be causing this? Maybe this is a pyDataverse issue but it seems like the calls are pretty straightforward. We are concerned specifically that all these warnings indicate that something is corrupting the metadata in our database. 

Thanks much!

p.s. Incase it helps here are the responses from a few of our calls to the pyDataverse `upload_datafile` and `redetect_file_type` functions:

```
[...]

PR #8835, opened 2022-07-08
GDCC/7527-fix-redetect

### LINKED ISSUE #7527: Error with Redetect File Type API

qqmyers:
**What this PR does / why we need it**: For non File Stores, redetect was using the name of the temp file and it's extension (always .tmp) to determine the mimetype, and with .tmp not mapping to a known mimetype could return null as the answer (if other detect steps didn't work). This fix provides the name from the datafile in the db and sends 'application/octet-stream' if no other determination was made.

Closes #7527

**Special notes for your reviewer**: Ran across this when doing other refactoring - would be good if someone who was following the issue could verify that this fixes both aspects.

**Suggestions on how to test this**: per the issue.

**Additional documentation**:


landreev (approved):
I really thought we had fixed this a while ago; but it appears we only talked about fixing it, but never got around to it. Great to see it fixed it finally. 
Looks good. (To be clear, this was broken for all stores, not just the remote kind; it wasn't working with the filesystem storage either - because it was running the check on the internal, storageidentifier-based filename).

pdurbin:
A couple questions:

- Should this PR close #8821?
- If we merge this PR, should we close PR #7631?

landreev:
No, this PR does not close #8821. 
#7631 should be closed regardless, because we decided not to merge it. 
Note that #7527 was 2 issues in one: a trivial issue with the file name; and a strange, not easily-reproducible EJB rollback that happened only on some files in our prod., and on the installation that first reported it. 
This PR fixes the trivial part of #7527 and we should merge it. 
My #7631 fixed that too, but also contained an attempt to add the `@Stateless` annotation to the API bean in question, which for some unknown reason made the strange EJB issue go away. But it was determined that we didn't want to do that. 

I'll check if that weird part of #7527 was still happening in our prod. If it is, it would be great to actually figure out what's going on there (should not block or delay merging this PR though).

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RedetectFileTypeCommand.java
```diff
@@ -62,7 +63,7 @@ public DataFile execute(CommandContext ctxt) throws CommandException {
             }
 
             logger.fine("target file: " + localFile);
-            String newlyDetectedContentType = FileTypeDetection.determineFileType(localFile);
+            String newlyDetectedContentType = FileUtil.determineFileType(localFile, fileToRedetect.getDisplayName());
```
landreev, line 66: Can we add a null check in the next line here? - From looking at #8821, it appears that determineFileType() may return null here, for especially unrecognizable files; and that results in a constraint violation.
Sorry for asking for this while the PR is already in QA; but would be a trivial change. Thanks.

landreev:
(If we add the null check I just asked for, this PR will in fact close #8821 as well)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RedetectFileTypeCommand.java
qqmyers: The change in FileUtil.java in this PR prevents a null return, so I think we're good?

landreev: Oh, of course, doh! Thanks!

landreev:
Yes, this PR closes #8821 as well. Sorry for the extra noise.

PR #8837, opened 2022-07-12
#8735 fix edit fields

### LINKED ISSUE #8735: New metadata blocks not appearing in previously created templates

sekmiller:
**What this PR does / why we need it**: IF you add a custom metadata block to a dataverse that already has a template, when you edit that template the new meatdata block is not available for adding default values. This PR allows you to add default values to the newly added MDB.

Closes #8735 New Metadata blocks not appearing in previously created templates.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Template.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Template.java
@@ -235,10 +235,8 @@ public class Template implements Serializable {
     }
 
     private void initMetadataBlocksForCreate() {
-        metadataBlocksForView.clear();
         metadataBlocksForEdit.clear();
         for (MetadataBlock mdb : this.getDataverse().getMetadataBlocks()) {
-            List<DatasetField> datasetFieldsForView = new ArrayList<>();
             List<DatasetField> datasetFieldsForEdit = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


pdurbin (approved):
I didn't run the code but the fix looks fine to me. Nothing weird in here and some nice cleanup.

PR #8844, opened 2022-07-14
Iqss/8841 restore checksum view

### LINKED ISSUE #8841: Truncated view of checksums missing in v 5.11
(as noticed by @adam3smith) : 
v5.11 on demo.dataverse.org: 
image

v5.10.1 @dataverse.harvard.edu: 
image

FWIW: I've tracked the removal to [the change to not call bind_bsui_components in many places and use o:onloadScript](https://github.com/IQSS/dataverse/commit/bd7634f2a6af679a52d80cf18768f1c2e519e828). The checksumTruncate() method that would have been called from there has https://github.com/IQSS/dataverse/blob/40c6f304de4bbc162dec865b3c7ca5cfb2fe22cd/src/main/webapp/resources/js/dv_rebind_bootstrap_ui.js#L216 - which removes the sr-only class and adds the visisble (sp - which made it easy to find!) class **-if called once**. It looks like it is now running twice and hence resetting the original classes.

I don't think visisble actually exists in any css file so the old function was essentially removing the sr-only class. Thus a simple fix would just be to remove that class in the xhtml file.  Alternately, the checksumTruncate() method could just removeClass("sr-only") instead of toggling it. (Same for "visisble" but again I don't see that used anywhere).

I can do either of those (already tested at QDR) if someone has a preference or something else if anyone can think of a reason to make a different fix. (I'm not sure why bind_bsui was moved or why the checksum truncate toggled these classes...)

qqmyers:
**What this PR does / why we need it**: Restores the truncated checksum display in the file table

Closes #8841

The PR removes the toggle of sr-only in the dv_rebind javascript and removes sr-only in the xhtml. If there was supposed to be some place where dv_rebind did not run and therefore the truncated checksum value remained hidden by design, this PR will break that (and we can pick a different solution). 

**Suggestions on how to test this**: The changes are to the checksum and UNF displays in the file table and search result so 4 tests of tabular/non-tabular files in the file table and in the search display would make sure that some value appears before teh clipboard icon (see images in issue).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: just fixes a regression

**Additional documentation**:


src/main/webapp/resources/js/dv_rebind_bootstrap_ui.js
```diff
@@ -213,7 +213,6 @@ function contentTruncate(truncSelector, truncMoreBtn, truncMoreTip, truncLessBtn
  */
 function checksumTruncate(){
     $('span.checksum-truncate').each(function () {
-        $(this).toggleClass('sr-only').toggleClass('visisble');
```
pdurbin, line 216: I know @qqmyers mentioned this in the issue but it's so weird to see this misspelling: `visisble`. I wonder what is or was going on here? Anyway, I guess it's going away. Looks like it was added in 89023ba.

pdurbin (commented):
As long as this works, I'm fine with it.

It looks like `sr-only` means "screen reader only". It's a Bootstrap thing: https://getbootstrap.com/docs/4.0/utilities/screenreaders/

scolapasta (approved):

PR #8849, opened 2022-07-20
IQSS/8846 - Fix HTML tags popup

### LINKED ISSUE #8846: HTML tags popup is blank, doesn't show allowed tags

- Edit a dataverse collection or a dataset (screenshots are for the former)
- Click on "HTML tags" to see the "Allowed HTML Tags" popup
- Observe that is is empty (screenshot below)

Screen Shot 2022-07-18 at 4 32 38 PM

All users.

I expect to see a list of allowed HTML tags (screenshot below):

Screen Shot 2022-07-18 at 4 34 14 PM

5.11.

The bug does not seem to be present in 5.10.1.

We also see the bug in the develop branch (567e506 as of this writing).

Possibly the bsui changes here:

- #8586

qqmyers:
**What this PR does / why we need it**: Fixes issue. Also fixes a related issue that the popup was not showing on the Edit Template page (i.e. for the dataset description). That may have been broken longer.

Closes #8846

**Special notes for your reviewer**: Looks like the change in how dv_rebind... was called made it's no argument constructor for popoverHTML be the one that initialized the popup. The popup, with the correct title and list of tags is called from the dataset and dataverse pages as well so it looks like just removing the no arg call fixes things.

For the template page, it looks like there hasn't been any call to the popoverHTML function with parameters for a while (the thing that made the dataverse and dataset pages work after removing the line noted above). So - guessing this has been broken earlier.

See issue - 3 places: check the dataset/metadata/description popup and the one in the issue for the dataverse description, and the dataset description in the template edit page.  and verify that you see the tags.

Key thing to check is whether the popup is still there after some ajax refresh of the page. (Spot checking by changing the template used for a new dataset (suggested by @scolapasta) suggests it does still work, but perhaps there are other situations where it won't).

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -1861,7 +1861,7 @@ The API call requires a Json body that includes the embargo's end date (dateAvai
 Remove an Embargo on Files in a Dataset
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-/api/datasets/$dataset-id/files/actions/:unset-embargo can be used to remove an embargo on one or more files in a dataset. Embargoes can be removed from files that are only in a draft dataset version (and are not in any previously published version) by anyone who can edit the dataset. The same API call can be used by a superuser to remove embargos from files that have already been released as part of a previously published dataset version.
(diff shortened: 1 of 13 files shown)


qqmyers:
I'm not sure why this now shows changes to 13 files instead of 2 as in https://github.com/IQSS/dataverse/compare/develop...GlobalDataverseCommunityConsortium:dataverse:IQSS/8846-fix_popup?expand=1 .
Since this was a small PR, I tried a rebase on develop in Eclipse. If we need a new PR, let me know. The difference shown above is all that is supposed to be new to develop.

pdurbin:
@qqmyers bah. Git is a pain. New PR, please, if you don't mind.

PR #8861, opened 2022-07-26
show alert to enable popups for external tools #8766

### LINKED ISSUE #8766: Feature Request/Idea:  Dataverse External Tools Fail if Pop-ups are blocked which is easy to miss
The external tools mechanism in Dataverse attempts to pop tools up in a separate tab (This is not completely true for  previewers, a sub-type of external tool - we first take you to the datafile page with an embedded view (works without popups enabled), but the view in external window option on that page then tries to popup a new tab.) 

For this to succeed, the user must allow Dataverse to pop-up new Windows. While this isn't hard, it isn't obviously required and the only indicator that you haven't done this on some browsers is a tiny icon in the URL bar, e.g. 
image

This can cause new users of Dataverse to think external tools/previewers aren't working. 

It would be good to avoid this situation. There may be multiple possible solutions so a quick Spike to see which make sense (technically and UI-design-wise) could be helpful. Possible options include:

- replacing the Dataverse window rather than popping up a new one (does this work? will it work with the new mechanism to launch tools via POST that will be used by OpenDP?). 
- detecting that popups are blocked (is this possible?) and alerting the user (how? how detailed?), either at login or when attempting to use a tool, that they must enable that for the tool to work.
- ?

pdurbin:

When users click a button in Dataverse to open an external tool in a new window tell their browsers to allow pop-up windows. This message from the browser can be somewhat subtle. Here's an example from Firefox:

Screen Shot 2022-07-26 at 4 23 46 PM

This pull request puts up an alert to let the user know that they have to pay attention to what their browser is telling them, that they should allow pop-ups.

- Closes #8766

None.

I tested using a private (Firefox) and incognito (Chrome) window.

Preview-only tools do not open in a new window so I changed the "Read Text" previewer to be both a preview and an explore tool like this:

```
curl -X POST -H 'Content-type: application/json' http://localhost:8080/api/admin/externalTools -d \
'{
  "displayName":"Read Text",
  "description":"Read the text file.",
  "toolName":"textPreviewer",
  "scope":"file",
  "types":["preview", "explore"],
  "toolUrl":"https://gdcc.github.io/dataverse-previewers/previewers/v1.3/TextPreview.html",
  "toolParameters": {
      "queryParameters":[
        {"fileid":"{fileId}"},
        {"siteUrl":"{siteUrl}"},
        {"key":"{apiToken}"},
        {"datasetid":"{datasetId}"},
        {"datasetversion":"{datasetVersion}"},
        {"locale":"{localeCode}"}
      ]
    },
  "contentType":"text/plain"
}'
```

Given a preview/explore tool like this, there are three places to test (for a .txt file):

- On the dataset page, click "Access File" then "Read Text"
- On the file page, click "Access File" then "Read Text"
- On the file page, click "Explore on Read Text"

In my testing, the alert only appears once unless you refresh the page.

Yes, two examples below.

## Dataset page

Screen Shot 2022-07-26 at 4 20 02 PM

Screen Shot 2022-07-26 at 4 20 08 PM

## File page (dropdown shown but "Explore on Read Text" is similar)

Screen Shot 2022-07-26 at 4 20 27 PM

Screen Shot 2022-07-26 at 4 20 33 PM

No.

**Additional documentation**:

None.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
@@ -7,6 +7,7 @@ import edu.harvard.iq.dataverse.FileMetadata;
 import edu.harvard.iq.dataverse.GlobalId;
 import edu.harvard.iq.dataverse.authorization.users.ApiToken;
 import edu.harvard.iq.dataverse.externaltools.ExternalTool.ReservedWord;
+import edu.harvard.iq.dataverse.util.BundleUtil;
 import edu.harvard.iq.dataverse.util.SystemConfig;
 import java.io.StringReader;
 import java.util.ArrayList;
@@ -200,4 +201,16 @@ public class ExternalToolHandler {
         this.apiToken = apiToken;
     }
 
+    /**
+     * @return Returns Javascript that opens the explore tool in a new browser
+     * tab if the browser allows it.If not, it shows an alert that popups must
+     * be enabled in the browser.
+     */
+    public String getExploreScript() {
+        String toolUrl = this.getToolUrlWithQueryParams();
+        logger.info("Exploring with " + toolUrl);
+        String msg = BundleUtil.getStringFromBundle("externaltools.enable.browser.popups");
+        String script = "const newWin = window.open('" + toolUrl + "', target='_blank'); if (!newWin || newWin.closed || typeof newWin.closed == \"undefined\") {alert(\"" + msg + "\");}";
+        return script;
+    }
 }
--- a/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
@@ -208,7 +208,7 @@ public class ExternalToolHandler {
      */
     public String getExploreScript() {
         String toolUrl = this.getToolUrlWithQueryParams();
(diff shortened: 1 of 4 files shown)


sekmiller (approved):

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
```diff
@@ -200,4 +201,16 @@ public void setApiToken(ApiToken apiToken) {
...
     }
 
+    /**
+     * @return Returns Javascript that opens the explore tool in a new browser
+     * tab if the browser allows it.If not, it shows an alert that popups must
+     * be enabled in the browser.
+     */
+    public String getExploreScript() {
+        String toolUrl = this.getToolUrlWithQueryParams();
+        logger.info("Exploring with " + toolUrl);
```
qqmyers, line 211: fine?

qqmyers (approved):
Looks good to me. FWIW: This could be one to fire up for community QA - the only real concern with this would be if there is some browser/os where this pops up when popups are already accepted and it might be easier to have the community try their favorites than to have Kevin try to cover everything. (That said, these days not supporting older browsers and their security issues could be a good thing.)

src/main/java/edu/harvard/iq/dataverse/externaltools/ExternalToolHandler.java
pdurbin: Whoops! Good catch! Pushed a fix.

Issue #8867, opened 2022-07-29
Internal Server Error on dataset page after deleting files, Vanishing files from already published versions.


lmaylein:
Calling https://heidata.uni-heidelberg.de/dataset.xhtml?persistentId=doi:10.11588/data/DVU14P results in an Internal Server Error.
I have deleted files from this dataset (as I did many times before) to replace them in the resulting draft version and republish the dataset with the replaced files. Already when deleting the files I got an error message. Now the dataset is no longer accessible at all (in any version).

How can I make the dataset accessible again?

<details>
<summary>server.log:</summary>

```


[2022-07-29T11:47:34.701+0200] [Payara 5.2021.6] [SEVERE] [] [javax.enterprise.resource.webcontainer.jsf.application] [tid: _ThreadID=109 _ThreadName=http-thread-pool::jk-connector(4)] [timeMillis: 1659088054701] [levelValue: 1000] [[
  Error Rendering View[/dataset.xhtml]
javax.el.ELException: /dataset-citation.xhtml @16,76 rendered="#{!empty DatasetPage.thumbnailString and !DatasetPage.workingVersion.deaccessioned}": java.lang.NullPointerException
	at com.sun.faces.facelets.el.TagValueExpression.getValue(TagValueExpression.java:77)
	at javax.faces.component.ComponentStateHelper.eval(ComponentStateHelper.java:170)
	at javax.faces.component.UIComponentBase.isRendered(UIComponentBase.java:313)
	at com.sun.faces.renderkit.html_basic.HtmlBasicRenderer.encodeRecursive(HtmlBasicRenderer.java:271)
	at com.sun.faces.renderkit.html_basic.HtmlBasicRenderer.encodeRecursive(HtmlBasicRenderer.java:283)
	at com.sun.faces.renderkit.html_basic.PassthroughRenderer.encodeChildren(PassthroughRenderer.java:80)
	at javax.faces.component.UIComponentBase.encodeChildren(UIComponentBase.java:566)
	at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1647)
	at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1650)
	at javax.faces.render.Renderer.encodeChildren(Renderer.java:152)
	at javax.faces.component.UIComponentBase.encodeChildren(UIComponentBase.java:566)
	at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1647)
[...]

Issue #8742, opened 2022-05-25
broken kebab on file listing page when dataset out of compliance with new (5.11) terms policy


qqmyers:
**What steps does it take to reproduce the issue?** Publish a dataset with files, try to use the kebob menu to restrict the file. The display looks like 
![image](https://user-images.githubusercontent.com/6731983/170359834-7082dff2-ac67-4b58-b427-4eb5294b1290.png)

and clicking the vertical ... kebab (note there's no down arrow after it) gives:
![image](https://user-images.githubusercontent.com/6731983/170359894-7a64fc20-c83d-4e8d-bc49-939a72a02a0e.png)

If I edit some metadata so that a draft version exists, the menus change to have a carat after them:
![image](https://user-images.githubusercontent.com/6731983/170360086-11266890-a752-47af-88da-5f88c80dd5e2.png)
 and it works as normal.

**Which version of Dataverse are you using?**
develop branch - doesn't seem to occur on 5.10.1

mreekie:
Jim came accross this while using the software.
This was working previously.
Small

pdurbin:
@qqmyers I'm on ba0183d1ed (develop) and having trouble reproducing this. My kebabs don't seem to be missing the arrows and I get the normal popup when I go to restrict a file (screenshots below). Any more tips on how to reproduce this?

![Screenshot 2022-07-25 at 16-41-00 test](https://user-images.githubusercontent.com/21006/180870440-a4e165aa-0029-4c6b-ba80-bae5f0222204.png)

![Screenshot 2022-07-25 at 16-42-37 test](https://user-images.githubusercontent.com/21006/180870535-e25fbc35-3519-4ccd-87de-058a06a92760.png)

qqmyers:
https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/CRN2AW shows this issue for me. Not sure if there is something else about it.

pdurbin:
Hmm, that "quick test" dataset you also mentioned in #8845 and it doesn't show any versions?!?

![Screenshot 2022-07-25 at 16-53-31 Quick Test](https://user-images.githubusercontent.com/21006/180872287-fa2968a3-68a2-4ace-ab3d-29428f1dcc40.png)

Very strange. Maybe you and I can do a quick call to try to reproduce this. I can't reproduce it on demo either.

qqmyers:
Hmm - I see 
[...]

PR #8870, opened 2022-07-29
IQSS/8845 - Avoid creation of extra version in TOA check

### LINKED ISSUE #8845: Version table always shows DRAFT?
**What steps does it take to reproduce the issue?** I'm not sure if it's new, but if I look at a dataset's version table, I see a DRAFT row even when there is no draft. Clicking the link brings me to the last version with a warning that draft doesn't exist, this is version x.y

DRAFT only appears if there is one.

5.11 / demo.dataverse.org

?

**Screenshots:**
DRAFT in the table - warning that draft doesn't exist at the top...

image

qqmyers:
**What this PR does / why we need it**: As of 5.11 a new dataset version was appearing when viewing the DatasetPage for datasets with no draft version. This has various effects from turning on the Delete Draft Dataset menu item to showing the 'ghost' draft in the version table. It may also relate to #8867, #8742. 

Closes #8845

**Special notes for your reviewer**: This removes the ghost version creation. I was able to create datasets (some by editing the db) that display this issue while not showing #8867, even if they had fileaccessrequest false and no TofA. So - while this is pretty clearly a bug, I don't know that other issues should be closed without further investigation. (I think it should stop the route we know if that might cause delete from a different version though.) 

**Suggestions on how to test this**: Create a dataset, publish one or more versions, verify that as a logged in user you don't see a ghost draft version in the versions table, don't see a Delete Draft Dataset item in the Edit menu.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -410,7 +410,7 @@ public class DatasetPage implements java.io.Serializable {
                hasValidTermsOfAccess = true;
                return hasValidTermsOfAccess;
             } else {
-                hasValidTermsOfAccess = TermsOfUseAndAccessValidator.isTOUAValid(dataset.getEditVersion().getTermsOfUseAndAccess(), null);
+                hasValidTermsOfAccess = TermsOfUseAndAccessValidator.isTOUAValid(dataset.getLatestVersion().getTermsOfUseAndAccess(), null);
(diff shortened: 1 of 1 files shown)


scolapasta:
Now that we have the other PR, I presume we can just close this one? @qqmyers tagging to make sure I'm not missing anything

qqmyers:
Yes - this was against develop but assuming the other PR is applied to develop as well this should close. That PR has this fix and another one.

PR #8876, opened 2022-08-01
remove cruft from top level: mdc-logs/raw-mdc-2019-01-07.log

pdurbin:
The `mdc-logs` directory and the file inside isn't needed. Delete it.


pdurbin:
I was just testing CI. I can make a new PR for this later if I feel strongly about it. Closing.

Issue #8874, opened 2022-07-29
Documentation issues


kuhlaid:
Below are a list of URLs that need to be removed or updated in the documents as they no longer point to useful resources (I can clean these out but wasn't sure if there were alternate resources to replace them):

1. Under [https://guides.dataverse.org/en/latest/admin/metadatacustomization.html#metadatablock-properties](https://guides.dataverse.org/en/latest/admin/metadatacustomization.html#metadatablock-properties) it references the dead URL ``https://dataverse.org/schema/citation/ `` which no longer points to anything.

2. Under [https://guides.dataverse.org/en/latest/installation/config.html#persistent-identifiers-and-publishing-datasets](https://guides.dataverse.org/en/latest/installation/config.html#persistent-identifiers-and-publishing-datasets) it references the dead URL ``https://dataverse.org/global-dataverse-community-consortium`` which no longer points to anything.

3. Under [https://guides.dataverse.org/en/latest/api/external-tools.html](https://guides.dataverse.org/en/latest/api/external-tools.html) there are multiple references to ``fabulousfiletool.com`` which not longer seems to be a functioning site and a search did not turn up alternatives.

4. Under [https://guides.dataverse.org/en/latest/installation/config.html](https://guides.dataverse.org/en/latest/installation/config.html) there is broken link to ``https://documentation.ands.org.au/display/DOC/ANDS+DataCite+Client+API``.

qqmyers:
Thanks for the heads up on these. As always you're welcome to submit a pull request with these changes as well.

One thing I'll note - the first one is not an error per se. If you look in the OAI_ORE file and use the json-ld semantic API, you will indeed see/use https://dataverse.org/schema/citation/. It is however, just a global semantic identifier (a namespace/root for many identifiers) rather than a live URL currently. So it might be helpful in the short term to avoid showing it as a live link. (In the longer term, it would make sense to provide a list of identifiers/definitions online at those URLs.)

AkiraY1:
Hey! This is the first time I'm contributing to this project. Let me know if there's anything I need to change or any guidelines/rules I didn't follow properly.

pdurbin:
@AkiraY1 hi! I left some comments on your pull requests. Thanks for making each change separate!

AkiraY1:
Thanks for the quick response! I responded to all the changes under each PR

AkiraY1:
[...]

PR #8886, opened 2022-08-03
Removed Dead Link

AkiraY1:
Removed the dead link https://dataverse.org/global-dataverse-community-consortium from https://guides.dataverse.org/en/latest/installation/config.html#persistent-identifiers-and-publishing-datasets. Partial fix for #8874


pdurbin (changes requested):
@AkiraY1 thanks for pointing out that this link is broken! Instead of removing it, can you please change it to https://dataversecommunity.global ?

AkiraY1:
Sure! I added another commit with the link replaced.

pdurbin (approved):
@AkiraY1 thanks for fixing the link. The preview looks good so I'm approving this:

https://dataverse-guide--8886.org.readthedocs.build/en/8886/installation/config.html#persistent-identifiers-and-publishing-datasets

PR #8887, opened 2022-08-03
Removed fabulousfiletool.com

### LINKED ISSUE #8898: fabulousfiletool.com no longer exists
Should remove references to fabulousfiletool.com on page https://guides.dataverse.org/en/latest/api/external-tools.html because fabulousfiletool.com no longer exists.

AkiraY1:
Removed all references to fabulousfiletool.com on https://guides.dataverse.org/en/latest/api/external-tools.html because fabulousfiletool.com no longer exists. Partial fix for #8874. Closes #8898.

### DIFF
--- a/doc/sphinx-guides/source/api/external-tools.rst
+++ b/doc/sphinx-guides/source/api/external-tools.rst
@@ -11,7 +11,7 @@ Introduction
 
 External tools are additional applications the user can access or open from your Dataverse installation to preview, explore, and manipulate data files and datasets. The term "external" is used to indicate that the tool is not part of the main Dataverse Software.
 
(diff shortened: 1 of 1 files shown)


pdurbin (changes requested):
@AkiraY1 thanks for the pull request. I assume you want to remove this because fabulousfiletool.com is not a valid website? Really this is just an example so we wouldn't want to remove it. We could replace it with a better example, maybe? With a real website? Not sure. It's probably easiest to leave it alone and close this pull request. I'm happy to hear your thoughts. Thanks again.

AkiraY1:
Sure, it makes sense to keep it as an example. Perhaps we could add a short note in parentheses that fabulousfiletool.com no longer exists? I was thinking that someone might try to use fabulousfiletool and assume they're doing something wrong before realizing the whole website is down.

pdurbin:
Sure, maybe. I don't think fabulousfiletool.com ever existed though. It was just a goofy URL I made up for the example. 😄

AkiraY1:
OH hahaha ok then I'll close this PR.

Issue #3832, opened 2017-05-16
Update documentation to indicate other DOI providers can be configured for doi.baseurlstring


mdmADA:
To reflect that "https://ezid.cdlib.org" and "https://mds.datacite.org" are no longer the only valid values for doi-baseurlstring, update the documentation at:

http://guides.dataverse.org/en/4.6.1/installation/config.html#doi-baseurlstring

The Australian Data Archive (ADA) successfully uses the Australian National Data Service (ANDS) API (a proxy for DataCite) to mint their DOI's through Dataverse.

ANDS supplied the doi-baseurlstring, doi-username and doi-password and Authority parameter values for ADA.

As ADA did for ANDS DOI minting, any DOI provider (and their corresponding DOI configuration parameters) other than DataCite and EZId must be tested with dataverse to establish if the other provider details will work or not.

dlmurphy:
@mdmADA, thanks for bringing this to our attention. I've written a new draft of this documentation that takes this new information into account. I hope it's alright that I borrowed a little from your helpful explanation! Here's what I've got, let me know if this solves the issue:

> As of this writing, "https://ezid.cdlib.org" (EZID) and "https://mds.datacite.org" (DataCite) are the main valid values. 
> 
> While the above two options are recommended, it is also possible on an experimental basis to use a DataCite Client API as a proxy to DataCite. In this case, requests made to the Client API are captured and passed on to DataCite for processing. The application will interact with the Client API exactly as if it were interacting directly with the DataCite API, with the only difference being the change to the base endpoint URL. 
> 
> For example: The Australian Data Archive (ADA) successfully uses the Australian National Data Service (ANDS) API (a proxy for DataCite) to mint their DOIs through Dataverse. ANDS supplied the doi-baseurlstring ("https://researchdata.ands.org.au/api/doi/datacite"), doi-username and doi-password, and Authority parameter values for ADA. As ADA did for ANDS DOI minting, any DOI provider (and their corresponding DOI configuration parameters) other than DataCite and EZID must be tested with Dataverse to establish whether or not it will function properly.

mdmADA:
Hi - I think that looks good but I am not sure about the 'experimental' part. ADA will be using the ANDS API in production so it won't be an experimental at that point...

Thanks!

pdurbin:
[...]

PR #8888, opened 2022-08-03
Remove broken link and description of DataCite Proxy no longer used by ADA

AkiraY1:
Removed broken link to https://documentation.ands.org.au/display/DOC/ANDS+DataCite+Client+API on page https://guides.dataverse.org/en/latest/installation/config.html. Partial fix for #8874.


pdurbin (changes requested):
Hmm, instead of a removing this link we should probably ask @mdmADA if there is a better link to use. This was added as part of #3832.

mdmADA:
Hi Phil.

ANDS (now ARDC) used to provide an API proxy for interfacing with DataCite in order to mint DOIs. That API was decommissioned ages ago so now ADA's dataverse instances call the DataCite Fabrica API directly to mint DOIs. So there is no better link to provide for this issue.

Thanks,
M.

pdurbin:
@mdmADA ok, thanks!

@AkiraY1 I think what we should do is simply remove all of the following (both paragraphs):

While the above two options are recommended because they have been tested by the Dataverse Project Team, it is also possible to use a DataCite Client API as a proxy to DataCite. In this case, requests made to the Client API are captured and passed on to DataCite for processing. The application will interact with the DataCite Client API exactly as if it were interacting directly with the DataCite API, with the only difference being the change to the base endpoint URL.

For example, the Australian Data Archive (ADA) successfully uses the Australian National Data Service (ANDS) API (a proxy for DataCite) to mint their DOIs through their Dataverse installation using a ``doi.baseurlstring`` value of "https://researchdata.ands.org.au/api/doi/datacite" as documented at https://documentation.ands.org.au/display/DOC/ANDS+DataCite+Client+API . As ADA did for ANDS DOI minting, any DOI provider (and their corresponding DOI configuration parameters) other than DataCite must be tested with the Dataverse Software to establish whether or not it will function properly.

AkiraY1:
Done! I removed the two paragraphs.

pdurbin (approved):
Looks good. Approved. Please see discussion in this pull request about how ADA doesn't use proxy any more. I'll also retitle the PR and probably the issue.

poikilotherm:
Please note I already have a refactoring of this entire section within the MPCONFIG PR for PIDs #8828

PR #8908, opened 2022-08-09
Improve archiving api error handling

### LINKED ISSUE #8907: Submit to archive api can return incorrect errors
The /api/admin/submitDatasetVersionToArchive endpoint returns any failure indicated by a WrappedResponse as 'api key required' and this isn't the only possible error (not finding the dataset for example) - reported by @pdurbin. Same issue in the batch version. The single version call can also fail with a null pointer if the specified version doesn't exist. 

develop

PR to follow.

qqmyers:
**What this PR does / why we need it**: Improves/fixes error responses.

Closes #8907 

**Special notes for your reviewer**:As far as I can tell there are no other ways than having a bad api key to trigger a WrappedResponse in the batch call but it makes sense to send the wrapped error there as well in case other WrappedResponse-generating calls are added to the method at some point (or in case I missed one).

**Suggestions on how to test this**: Use the submitdatasetversiontoarchive api call using a bad api key, a bad dataset id, a bad version number, etc. and verify that the errors are appropriate/not all reporting a bad api key. For the batch submit call - verify that sending a bad api key gives a reasonable message. 

**Is there a release notes update needed for this change?**: no?

**Additional documentation**:


pdurbin (approved):
I didn't test this but the fix looks great. We are now returning the response from the WrappedResponse, which is our usual pattern. API tests are passing too. 🎉

PR #8915, opened 2022-08-11
7980 enhanced dsd

### LINKED ISSUE #7980: DB connection: defaults and more options

poikilotherm:
Create real defaults for the database connection, enable configuring advanced settings, add docs.

Closes #7980 

- The release docs of older releases do not need to be changed. The statements on how to upgrade etc are working just fine.
Only in a few occasions people might have hit the regression.
- This PR replaces #7983.
- This PR requires people to update to Payara 5.2022.2 at least. It requires #8949  to be merged.
- It might be debatable to [add a few notes to this place about monitoring](https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/admin/monitoring.rst#database-connection-pool-used-by-app-server)

1. Remove the properties from domain.xml. Should still work (if you had used the defaults in there)
2. Try setting the JDBC logging to enabled via system property: `asadmin create-system-properties "dataverse.db.log-jdbc-calls=true"`

Nope.

:battery: included.

**Additional documentation**:
None.


mreekie:
sizing: Waiting on Oliver input. He is out today.

mreekie:
Phil is going to talk with Oliver to see if this can be lowered in priority in the community column

poikilotherm:
I've just merged latest `develop`. This is a more or less trivial change, as there isn't much to test. The functionality is provided by Payara, so IMHO there is no point in testing that. One might test setting a setting, but at this point IMHO we should be confident enough that Payara handles these variable replacements well.

pdurbin:
> That said, yes, it may move down the column. I have no urgent business requiring this ATM.

You know what? Based on comments by @landreev and @donsizemore in the issue regarding the usefulness of `sslmode=require` I'm fine with keeping this where it is priority-wise. I'll throw a 3 on it for size. I  hope that's big enough.

By the way, there's a failing test but I suspect it's unrelated:

```
java.lang.AssertionError: 
Expected status code <200> doesn't match actual status code <403>.

	at edu.harvard.iq.dataverse.api.FilesIT.test_008_ReplaceFileAlreadyDeleted(FilesIT.java:846)
```

mreekie:
Pdurbin reviewed and sized. Moving to sprint ready column.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -263,6 +263,153 @@ As for the "Remote only" authentication mode, it means that:
...
+       **Please note the safety advisory above.**
+     - *No default*
+   * - dataverse.db.name
+     - The PostgreSQL database name to use for the Dataverse installation.
+     - | ``dataverse``
+       | (installer sets to ``dvndb``)
+   * - dataverse.db.parameters
+     - Connection parameters, see `Postgres JDBC docs <https://jdbc.postgresql.org/documentation/head/connect.html>`_
+       Note: you don't need to provide the initial "?".
+     - *Empty string*
```
pdurbin, line 325: I tried putting the following in domain.xml...

```
<jvm-options>-Ddataverse.db.parameters=sslmode=require</jvm-options>
```

... and it seems to have worked. Dataverse failed to deploy and I get errors like this in server.log:

Connection could not be allocated because: The server does not support SSL.

Error in allocating a connection. Cause: Connection could not be allocated because: The server does not support SSL.

  Exception while invoking class org.glassfish.ejb.startup.EjbApplication start method
javax.ejb.EJBException: javax.ejb.CreateException: Initialization failed for Singleton StartupFlywayMigrator
...
Caused by: org.flywaydb.core.internal.exception.FlywaySqlException: Unable to obtain connection from database: Error in allocating a connection. Cause: Connection could not be allocated because: The server does not support SSL.


SQL State  : null
Error Code : 0
Message    : Error in allocating a connection. Cause: Connection could not be allocated because: The server does not support SSL.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -263,6 +263,153 @@ As for the "Remote only" authentication mode, it means that:
...
+       | (installer sets to ``dvnapp``)
+   * - dataverse.db.password
+     - The PostgreSQL users password to connect with.
+
+       **Please note the safety advisory above.**
+     - *No default*
+   * - dataverse.db.name
+     - The PostgreSQL database name to use for the Dataverse installation.
+     - | ``dataverse``
+       | (installer sets to ``dvndb``)
```
pdurbin, line 321: I'm not sure how to see that the default database user and database are "dataverse".

To deploy this branch I used `scripts/dev/dev-rebuild.sh` which has the names the installer uses hard coded:

DB_NAME=dvndb
DB_USER=dvnapp

I guess I could try changing that script to this...

DB_NAME=dataverse
DB_USER=dataverse

... and then deleting these lines from domain.xml:

      <system-property name="dataverse.db.user" value="dvnapp"></system-property>
      <system-property name="dataverse.db.name" value="dvndb"></system-property>

That is, without these system properties configured, it sounds like the default database user and database name will be "dataverse".

pdurbin (approved):
Looks good.

The most immediate real world use case for this PR seems to be `sslmode=require` (see the issue this PR closes) so I mentioned this in the docs.

I did some cursory testing but didn't test the new default database name and user (both "dataverse") though I left a comment about how we could probably test this.

I love the new rst table format, by the way. Seems much easier than keeping the correct number of spaces here or there!

doc/sphinx-guides/source/installation/config.rst
poikilotherm: Yes. Sorry, but for container environments it really is common practice to use the application name for the database name. As the installer sets this explicitly for any installation out there, folks should be fine.

kcondon:
@poikilotherm Can you refresh from develop? Getting flyway errors on deployment. Thanks!

pdurbin:
@kcondon done! Refreshed! Ahhh.

PR #8923, opened 2022-08-19
GDCC/Update storageIO processing to match RemoteOverlay

qqmyers:
**What this PR does / why we need it**: PR #7325 added a mechanism to validate inputs for direct-upload-related datafile submissions that occurs earlier in the add pipeline than other checks we have. This PR adds those checks for other store subclasses.

Closes #

**Suggestions on how to test this**:  Should verify that flyway script works if the constraint already exists (i.e. it should just warn and not fail.)

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/dataaccess/StorageIO.java
```diff
@@ -594,10 +594,14 @@ public static String getDriverPrefix(String driverId) {
...
     
+    public static boolean isDirectUploadEnabled(String driverId) {
+        return Boolean.getBoolean(System.getProperty("dataverse.files." + driverId + ".download-redirect", "false"));
+    }
+    
     //Check that storageIdentifier is consistent with store's config
     //False will prevent direct uploads
     protected static boolean isValidIdentifier(String driverId, String storageId) {
-        return true;
+        return false;
```
sekmiller, line 604: Is this going to cause a problem for SwiftAccessIO because it doesn't implement the method so now the DataAccess line 367 will return false instead of true?

qqmyers: Swift doesn't do direct uploads so this nominally doesn't affect the way Swift has been used. This will prevent someone turning direct-upload on/using the /add file endpoints as a workaround for Swift until there's a check added for the Swift store. It will also prevent any new store types from allowing direct upload unless they implement a type-specific isValidIdentifier method. (Note that true was only introduced in the Remote Store PR and the intent was to flip this to false once file and S3 also had stricter validation tests in place.)

src/main/java/edu/harvard/iq/dataverse/dataaccess/FileAccessIO.java
```diff
@@ -656,4 +659,28 @@ private String stripDriverId(String storageIdentifier) {
...
+        String pathString = storageId.substring(storageId.lastIndexOf("//") + 2);
+        String basePath = "/tmp/";
+        try {
+            String rawPathString = basePath + pathString;
+            Path normalized = Paths.get(rawPathString).normalize();
+            if(!rawPathString.equals(normalized.toString())) {
+                logger.warning("Non-normalized path in submitted identifier " + storageId);
+                return false;
+            }
+            System.out.println(normalized.getFileName().toString());
```
landreev, line 674: Is this println line here on purpose, or were you planning to remove it before this is merged? (just checking; there's a couple more in the PR).

qqmyers: Good catch - changed the 2 in the main code to logger.fine (left the one in the test).

landreev:
(As is the custom with your PRs, it is merged, but it's left up to you to delete the branch, or not)

PR #8962, opened 2022-09-07
IQSS/8730 - fix direct upload progress

### LINKED ISSUE #8730: Direct upload progress bar jumps around with many files
**What steps does it take to reproduce the issue?** Uploading many files that take a while to upload and for hashes to be calculated, one can see the progress bar jump around.

**Which version of Dataverse are you using?**5.10.1

The root cause is that the code calculating the md5 hash an uploaded file is updating all progress bars instead of just it's own. This is only a problem when the bar is > 50% (the first half of the bar is the upload time, the second half is when the file is being read to calculate the local hash). It is only the hashing progress that is updating all bars instead of its own. Other bars <50% get reset whenever their thread updates the upload progress. If there is more than one file in the hash calculation stage, all bars will bounce to the progress for whichever file reports, so bars can go up/down repeatedly.

The problem is in https://github.com/IQSS/dataverse/blob/fdef1f6aa4664da872fdbeb65232b51684098c85/src/main/webapp/resources/js/fileupload.js#L321 which should be finding the right progress bar and not just updating all bars.  I haven't yet tried it but presumably [the code used elsewhere to find the current file's row/bar ](https://github.com/IQSS/dataverse/blob/fdef1f6aa4664da872fdbeb65232b51684098c85/src/main/webapp/resources/js/fileupload.js#L160-L163) would work here as well.

qqmyers:
**What this PR does / why we need it**: Direct upload via the UI sends up to 4 files at a time and their progress bars are all changing at the same time. The old code would update all progress bars during the hash calculation phase rather than only the bar for the current file.

Closes #8730

**Suggestions on how to test this**: upload a mix of small and large files (< MB  up to 100 MB+) and assure that when the larger file(s) are showing progress above 50%, the smaller progress bars aren't jumping up/down. It may be easier to see if you set the network to be throttled in the browser dev console to limit how fast bytes are transferred (e.g. fast 3G or even slow 3G).

**Additional documentation**:


mreekie:
added to sprint Dec 15, 2022

scolapasta (approved):

PR #8969, opened 2022-09-09
nullpointer fix when getting notification with newly created user

ErykKul:
This fixes the nullpointer on DataverseUserPage when creating a new user.


ErykKul:
After some testing: when a user is created with shibboleth in UI, that user is directly set in the session. It gives nullpointer until you log out and log in again (user is loaded through database). This pull request should fix that.

ErykKul:
This workaround is valid. Technically, it is a little bit different: with log-out and log-in again nothing is written in DB, but the object is initialized differently (muted e-mails are empty i.s.o. being null). This pull request fixes the initialization of the object when it is constructed by creating a new account (it was the case I had missed when making the initial pull request). Until you mute something, nothing is written in DB.

mreekie:
Phil points out that testing of this will not be straightforward based on our current environment.
The test server will need to be setup.
This will impact the size of this work to QA this.

To do this the right way - will require a full Shiboleth setup.
The shorter way is to do it in the artificial dev mode.

This calls out that we do not have a current shiboleth environment.

For this we'll do this with the artificial dev mode for this PR.

src/main/java/edu/harvard/iq/dataverse/authorization/users/AuthenticatedUser.java
```diff
@@ -147,7 +147,7 @@ void prePersist() {
     }
     
     @PostLoad
-    void postLoad() {
+    public void postLoad() {
```
sekmiller, line 150: @ErykKul - thanks for this. sorry it has been sitting around for a while. I ran it by Gustavo and he suggested naming the method something like "initialize" to better reflect what it does. Other than that it's ready for QA.

ErykKul: @sekmiller I have just renamed this function.

PR #8933, opened 2022-08-24
8932 container base image

poikilotherm:
**What this PR does / why we need it**:

This PR adds a Maven submodule to build a base container image containing a tuned Payara app server and all things necessary to deploy Dataverse WAR in it.

**Which issue(s) this PR closes**:

Closes #8932

**Special notes for your reviewer**:
- It might be debateable if this image should also do the heavy lifting for JHove config.
- This pull request will be the first of more to come. It is necessary to talk about the review/QA process here, as discussed in-person. Looks like #8910 is related.
- It does not yet contain a Github Actions to build and push the image to a registry.

TODOs:
- [x] Create flattened POM and use `install` target instead of `package` to make the build dependable from app image build
- [x] Make uid and gid configurable build args
- [x] Add Github Action to build and push the image to Docker Hub (needs credentials secrets; limit runs to IQSS repo)
- [x] Include with pushing the image also [pushing a description](https://github.com/marketplace/actions/docker-hub-description). (Description should again state the community-only support)
- [x] Add performance tweaks from [LIBIS image](https://github.com/libis/rdm-build/blob/599f9a47188d0cf7021ecf0040830bc304d8849d/images/dataverse/bin/init_0_deploy_dataverse.sh#L149-L160)
- [x] Think about [removing expired certs](https://github.com/libis/rdm-build/blob/main/images/dataverse/bin/init.d/removeExpiredCerts.sh), but at build time
- [ ] Look at the [OpenAPI configuration as done by LIBIS](https://github.com/libis/rdm-build/blob/599f9a47188d0cf7021ecf0040830bc304d8849d/images/dataverse/bin/init_0_deploy_dataverse.sh#L42-L47)
- [x] Think about a `/data` folder for future workaround of storing temporary upload files ([see here](https://github.com/IQSS/dataverse/issues/6656#issuecomment-1057779724)) - maybe also to be used with `<dataverse.files.directory>/temp` location in case of containers?
- [x] Add a place/volume for the Make Data Count files (to be processed by another container, but logs written by us)
- [x] Probably make some places like `/docroot`, `/data`, `/dumps` a `VOLUME` by default? (Not likely to create a pitfall for Docker volumes vs. Kubernetes/Bindmounts chaos here)
- [x] Add some notes about the multiarch build nature on Linux (see [1](https://github.com/docker/buildx/issues/159#issuecomment-539477326))

**Suggestions on how to test this**:

Simply install Docker/... and follow the provided instructions within the new container guide.

[...]

PR #8983, opened 2022-09-20
feat(upload): make upload file storage path configurable #6656

### LINKED ISSUE #6656: Refactor file upload from web UI and temporary storage
## Intro
This issues evokes from https://groups.google.com/d/topic/dataverse-community/IB1wYpoU_H8

A rough summary: when you upload data via web UI, the data is stored at a few different places before commited to real storage, as it needs to be processed, analyzed and maybe worked on.

Nonetheless, this could be tweaked, especially when starting to deal with larger data. This is kind of related to #6489. As not every big data installation might want to expose its S3 storage to the internet or even does not use this new functionality at all, this should still be addressed. From a UI and UX perspective, which is touched here, issues #6604 and #6605 seem related.

Mentions: @pdurbin @djbrooke @TaniaSchlatter @scolapasta @qqmyers @landreev 

## Discovery
I tested this on my local Kubernetes cluster and created a little video of what happens when. After that I digged a bit through the code to see if there are potential problems ahead for our installation.

Please enjoy the following demo (click on the image):
[upload-test-1g-2g-5g](https://imgur.com/a/QAgg85l)

Things to note:
1. Data is uploaded and stored in `glassfish/domains/domain1/generated/jsp/dataverse` folder, as already stated in the Google Group posting.
2. After upload is done, but before marked completed, the data is copied to the `temp` folder, living at the configured place of JVM option `dataverse.files`. Unless this copy is done, the user experience is a "hanging progress bar at 100%".
3. The uploaded data is in most cases cloned, eating more space. When the upload has been marked finished to the user, often the temporary data in the `generated` folder is not deleted.
4. Upload to real file storage starts when the user clicks on "Save". The user experiences a moving circle, which might appear as hanging for large data.

## Analysis
Now let's see what happens in the background from an implementation view. The following diagram is a bit shortened and leaves out edge cases to make it more understandable.

Click on the image to access the source.
[plantuml](http://www.plantuml.com/plantuml/uml/NP91gnen4CVl-HI3XwBmdjuBBKyKhGzAfUhDIzoTPrEooPB9Mkodxqoo3L5F8Vdz_ydyt3UAEiIXiqetqGSu4GRLyvPfJAzTXCdxOGFJVJ0TRdI3D9k09ZZj7g6DiNYAngR3pUdnSEUkIF60uMOQJCJke5eTzKKJvYmS-79znPolJsLrmq0u_s2E5X6xFb77x_iQyvqsRd55Xq57RCzkyPlwHKcdoFPzFOuvsUfmuVlERkMjHTdBn4efKG2lNtb6-0pVZiSzd7hhTQiaQsxSBkCt0MKbbGmopYlcPGw48GSABvISlwPQDOPPLCxoWgDvSwsZ5YCqdPMyW8oKIrQ-luFecDAiz38pzegQpCkA-RscY2LGMWKJ8ASV4y4tIFG2W_jh-XUOp-Srocmnz1FZ45ng1siegXOf1dyNRXnFZ8gRN_xFg0OQt_KM8p9TMrLV-0VBZyW1RMrg9MYzGxZSOP0U8Ue_YNlrZI4HHnWNIPwgI_CZTWT-0Z6wTEsoRHDsz6EgaisWOhQeIBBBMug5GcsaTv8LHm-KF-Ip6yYuw__lUlTZeXZau7Fz6aL3aSQRwVZ9BaKT1sAgVgNCDThDSckiShfTLP-EzOJyge2wWq_FjpEZtj2rGsV_0G00)

Things to note here:
1. [PrimeFaces uses Apache Commons FileUpload](https://stackoverflow.com/a/12037549), so the temporary files created from it are deleted once the `InputStream` closes. This is however not the case in current codebase, which seems to be at least one cause for dangling files as noticed in the discovery above.
2. The GC closes the dangling Streams when the editing is done, but there is a great danger of leftovers. This should be avoided.
3. As noticed in the Google Group and during discovery, the place where PrimeFaces uploads its data is not the same as `dataverse.files` has been configured for. The docs lack of that information and it has a great danger of disks filling up, breaking things.
4. The same data is copied around a few times, and from what I saw in the code it is even copied when moved to real storage when storing on local filesystem. That sound very much like introducing lag for users. 

## Ideas and TODOs
I feel like this is a medium sized issue, as it needs quite a bit of refactoring.
A collection of ideas what to do next:
- Think about optimizing upload, avoiding multiple copies. Things like [this](https://primefaces.github.io/primefaces/7_0/#/components/fileupload?id=filter-configuration) or [this](https://stackoverflow.com/questions/18548977) and [this](https://books.google.de/books?id=NiqzCQAAQBAJ&ots=Xow9N5w8ZB&dq=&pg=PA228#v=onepage&q&f=false) seem to be good starting points.
- Think about the UX - is this acceptable for large data? The upload itself is quite fast, as seen, even for bigger files (this was from my workstation to a remote on premises location with 1Gbit LAN).
- To what issues is this connected, too?

poikilotherm:

As outlined in IQSS#6656, files will be stored in `domaindir/generated/jsp/dataverse` during upload before being moved to our temporary ingest file space at `$dataverse.files.directory/temp`.

With this commit, we enable to configure a different place for these kind of generated temporary files by using MPCONFIG variable substitution inside of glassfish-web.xml. This is kinda hacky, but at least it allows to store the data in distinguished places.

Also sorts the content of glassfish-web.xml into order as specified by the XSD. Documentation of the setting is provided.

Closes #6656

Please see screenshots below. You can also try with `docker-aio`.
I added a few notes about temporary upload storage to the configuration guide with the gist of knowledge from #6656 so admins can learn about potential problems with large file uploads.

The container base image in #8933 certainly will include something using this if this PR would be merged before. Otherwise I'd extend this PR to include something for the container scenario.

Configure your deployment with a system property pointing `dataverse.files.uploads` to a different place and watch data being stored there.

Nope.

Do we need one? This might be a scenario for admins as outlined by a bunch of folks. Please leave a comment.

**Additional documentation**:
:battery: included.


poikilotherm:
I tested this today in `docker-aio`. Here's how it went.

### UI Upload - 1 GiB random data :: Dataset `10.5072/FK2/QACLPF` :: Default location (./uploads)

1. Upload in progress:
    image
2. Upload finished, data moved to `<dataverse.file.directory>/temp`:
    image
3. After hitting "Done", data is moved to final destination:
    image

landreev:
This looks great, we should be able to move it along and merge it in the next sprint. 
Thank you! (I do feel BAD about not having gotten around to doing this ourselves ☹️)

pdurbin:
Small code change. Good docs. Gave it a size of .25.

pdurbin (approved):
Looks great! Thanks, @poikilotherm!! Approved (I made small doc changes). @landreev also made a positive comment: https://github.com/IQSS/dataverse/pull/8983#issuecomment-1330922696

The only testing I did was to confirm that the new unconfigured JSF upload default directory is `/usr/local/payara5/glassfish/domains/domain1/uploads`. I assuming configuration works fine.

Moving to QA.

PR #9010, opened 2022-10-03
Overload DatasetField method to remove type check

### LINKED ISSUE #1240: Cleanup: DatsetField factory methods
To support templates, the factory methods (createNew and copy) were changed to take in an object. Instead we should overload and have individual methods for DatasetVersion and Template.

bencomp:

It replaces a method that did a string comparison on the class name of an `Object` parameter to determine what to do by two methods that have the expected parameter types.
This improves type safety by allowing the compiler to check types.

Closes #1240, removes a TODO from the codebase and relates to #9021.

I added a few basic unit tests so that the coverage doesn't decrease with the new methods.

Please check that the logic of the new tests makes sense and that they pass like the existing unit tests.

No.

No.

**Additional documentation**:


bencomp:
@sekmiller could you have a look at this? I think this was your code.

bencomp:
I just noticed that there actually was an issue that this PR closes: #1240 (I edited the PR description accordingly).

Perhaps @scolapasta could have a look at this PR, since he opened the issue many years ago?

sekmiller (approved):

landreev:
Looks good. Built and tested locally. Automated tests are passing. 
Merging.

PR #9018, opened 2022-10-03
GDCC/9005 replace files api call

### LINKED ISSUE #5924: Is there an option to batch file replace?
Hello, related to file replace, we are wondering if there is a way to replace ALL files in a dataset at once? 

"My question is how do I replace a dataset that has over 47 files? I have deposited my dataset in geodatabase format which has over 46 separate files in the structure. Is there a way to do a bulk replace?"

For now we see individual files can be replaced, using file replace feature, but is there a better way for many files to be replaced? Deleting the replaced files and uploading new files is I guess another option. 

Any other thoughts?
Thanks, 
Amber

### LINKED ISSUE #9005: Bulk replace files API
#7900 added an efficient way to register multiple files with a dataset after a direct upload or out of band transfer.  This was analogous to the original API to add a single file. However, there is no bulk file replace API that is analogous to the single file replace API call. Such a call is needed for bulk migration operations and would enable efficient replace operations from Globus or DVUploader, etc.

**What inspired the request?** DANS has this issue in bulk migration of datasets from other systems where the volume of data makes using the single file replace call prohibitive.

PR to follow soon.

qqmyers:
**What this PR does / why we need it**: This PR adds a /replaceFiles api call to allow bulk direct upload/out-of-band upload replace operations. It also include 

- some minor cleanup, 
- a bug fix for /addFiles where an error with one file would stop processing of further files and falsely report the same error for all the skipped files
- minor improvement to report the bulk edit operations in the edit-drafts logs (as the single file versions do).

Closes #9005
Closes #5924

**Special notes for your reviewer**: No new tests - while this could be tested within an overall direct upload test (only on S3 with upload-redirect=true), it is tricky to test as a single call since one has to be able to place files in storage directly prior to running the /replaceFiles call. If that can be done in an IT test, then this and /addFiles could be tested this way.

**Suggestions on how to test this**: @janvanmansum /DANS should be able to test this as part of their migration activity so their review would be useful. Otherwise, manual testing can be done - easiest on an S3 store with direct-upload enabled. (One can temporarily add upload-redirect=true on a file store but this breaks upload in the UI). The examples in the docs should give a good idea of how to test. W.r.t. to regression, this does make minor changes to the overall add/replace mechanism so retesting the single file add/replace apis would be useful. 

**Is there a release notes update needed for this change?**: added a small one

**Additional documentation**: included.


mreekie:
The crux is that we can now replace multiple files at once.
It is functionality that we have, just in a different API call.

Discussion:
DANS wanted this and may be that they have tested this already.
If we rely on their work then we would only need to do regression tests.
It would make us feel more confident if we can still do some QA
There are 9 files, but some are documentation which makes review easier
Jim will resolve the merge conflicts as part of the review.
Bulk add is already existing. DVUploader supports this using the DirectUpload API.
This is not a bulk add. It is a direct upload. 
For bulk add, you can do it with a direct upload.
This does a bulk replace. It only works with a direct upload.
Will need S3 setup to test this.  The test can be built but it will need to be built out to do 3 things.

It looks like this may close #5924. It may not as they might not be using direct upload. Does Amber using S3?

This gets a 33. 

Next step:
- @qqmyers will check in #5924 to see if this addresses their issue.

mreekie:
added to sprint Dec 15, 2022

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2366,48 +2373,6 @@ The fully expanded example above (without environment variables) looks like this
 Note: The ``id`` returned in the json response is the id of the file metadata version.
 
 
-
-Adding File Metadata
```
sekmiller, line 2370: Not sure why you're removing this doc since the endpoint still exists and it seems like we would want to have the ability to add files' metadata, not just replace

src/main/java/edu/harvard/iq/dataverse/datasetutility/AddReplaceFileHelper.java
```diff
@@ -2166,6 +2164,175 @@ public Response addFiles(String jsonData, Dataset dataset, User authUser) {
...
+        // -----------------------------------------------------------
+        // Read jsonData and Parse files information from jsondata  :
+        // -----------------------------------------------------------
+        try {
+            filesJson = JsonUtil.getJsonArray(jsonData);
+
+
+            if (filesJson != null) {
+                totalNumberofFiles = filesJson.getValuesAs(JsonObject.class).size();
+                workingVersion = dataset.getEditVersion();
```
sekmiller, line 2197: getEditVersion method has been removed.

src/main/java/edu/harvard/iq/dataverse/datasetutility/AddReplaceFileHelper.java
```diff
@@ -2033,15 +2036,14 @@ public Response addFiles(String jsonData, Dataset dataset, User authUser) {
...
-            filesJson = dbJsonReader.readArray();
-            dbJsonReader.close();
+        try {
+            filesJson = JsonUtil.getJsonArray(jsonData);
 
 
             if (filesJson != null) {
                 totalNumberofFiles = filesJson.getValuesAs(JsonObject.class).size();
-
+                workingVersion = dataset.getCreateOrEditVersion();
```
sekmiller, line 2045: method should be getOrCreateEditVersion

doc/sphinx-guides/source/api/native-api.rst
qqmyers: This endpoint doesn't work as ~implied by this section and this placement in the guides. You can't change the metadata of existing files with it. It is a way to add new files that have been uploaded by direct S3, out-of-band means, or when using the remoteOverlayStore. This call is documented already at https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html?highlight=addfiles#to-add-multiple-uploaded-files-to-the-dataset.

sekmiller (changes requested):
a couple of old methods in the add replace helper need to be updated. also a question about addFiles.

doc/sphinx-guides/source/api/native-api.rst
sekmiller: OK. thanks for the explanation

sekmiller (approved):
Looks good, thanks!

kcondon:
Issues found:
1. extra buttons on upload page appear during replace file workflow in UI if file types differ and you select delete.
2. example json for multiple file upload needs different filetoreplace ids.

qqmyers:
@kcondon - hopefully ready for QA again. 1) was not related to the PR, just to replacing when using direct upload, but I included a fix. Doc changes for 2),3) have been made, and I fixed 4) - looks like a change along the way broke things and made only the last file in the list be changed (with earlier ones being deleted with no replacement).

PR #9061, opened 2022-10-14
IQSS/9020- Fix for Update current version issue when a file is the dataset thumb

### LINKED ISSUE #9020: Update current version fails when a file in the dataset is used as it's thumbnail.
**What steps does it take to reproduce the issue?**  Publish a dataset with an image file in it, confirm that it is used as the dataset thumbnail. Edit some dataset metadata and try to run the 'Update Current Version' option when publishing (as a superuser). An error will be shown:
image

Update Current Version should work

Dataverse 5.11.1 and ~5.12

There was a similar issue with files having category tags. The root cause is ~ that the filemetadata is removed from the draft version as a valid part of how the command works but the dataset/thumbnailfile_id column keeps that file and it's filemetadata in the context, hence the conflict shown. The fix should be to remove that thumbnail link as well.

A possible workaround involving a db edit has been found: set the dataset thumbnail to an image file in another dataset `update dataset set thumbnailfile_id=<some good file> where id = <your dataset>;`, do the Update Current Version, and then reset back to the original thumb. Note that setting the thumb to null or a non-image file fails because Dataverse will aggressively reset the thumb back to one of the dataset's image files during the page refreshes to call the Upload Current Version command. This appears to have worked in one test and not another.

qqmyers:
**What this PR does / why we need it**:  See issue -for details.

Closes #9020

**Special notes for your reviewer**: This appears to be another case where there is an indirect reference causing a conflict. In this case, since we deleting the fileMetadatas from the draft version, if a file is the dataset thumbnail, and since files keep a list of file metadatas, we end up with a reference to the deleted filemetadata. We are removing the draft file metadata from the file in line 137 but this is not the same java object as the datafile from dataset.getThumbnail(). Resetting that to be the same Datafile object as in the PR avoids the error.

**Suggestions on how to test this**: Create a dataset with an image file, publish, verify that the dataset thumbnail is from the image file in it. Create a draft with a metadata change. As a superuser, use Publish/Update Current Version and assure that it works. Before the PR, this should fail and update current version should only work if there is no datafile being used as the dataset thumb.

**Additional documentation**: - just a bugfix


scolapasta (approved):

scolapasta (approved):

kcondon:
I'm unable to reproduce this bug on test or demo.
1. create dataset
2. using thumbnail page, upload and select thumbnail file for dataset
3. publish
4. as superuser, edit dataset metadata, ie. subtitle
5. publish as same version,  works.

Issue #9054, opened 2022-10-12
Document how to sign-up for the Google group without a Google account


twindiscale:
**Overview of the Feature Request**

Short explanation how to sign up without a Google account to the group to receive and send emails. Suggested place: https://guides.dataverse.org/en/5.12/installation/intro.html#getting-help

**What kind of user is the feature intended for?**

All

**What inspired the request?**

My attempt to sign up

**What existing behavior do you want changed?**

There is no documentation.

**Any brand new behavior do you want to add to Dataverse?**

no

**Any related open or closed issues to this feature request?**

unknown

PR #9068, opened 2022-10-17
Update HTTP URL to HTTPS URL

bencomp:
This updates (almost) all HTTP URLs found in the documentation and code to HTTPS. It increases privacy of readers and may even save some time if the HTTP address needs to redirect to the HTTPS address.

I found the HTTP URLs after looking at #9054, but this is unrelated.

I manually checked that the HTTPS URLs resolve. Some do redirect to updated URLs, but I only updated the URL if the content matched the expected content. There may be exceptions to this rule.

I couldn't access the Phoenix server, so I kept those URLs the same.

For now, you could scan the changes and perhaps sample a few URLs to see that resolve to what you would expect.
In the future, you could use an automatic link checker.

It does change the template of the Guides, so changes may be recognised on many users.

N/A
**Additional documentation**:


bencomp:
Merge conflicts caused by #9616 efforts have been resolved.

bencomp:
Updated URLs in the "classic dev environment". Hoping for a positive review 🤞

bencomp:
Fixed merge conflict caused by merging #9796.

PR #9075, opened 2022-10-18
clarified moving dataverse collection description

### LINKED ISSUE #8527: documentation: /api/dataverses description is ambiguous
On https://guides.dataverse.org/en/latest/admin/dataverses-datasets.html the "Move a Dataverse Collection" description reads

> Moves a Dataverse collection whose id is passed to a new Dataverse collection whose id is passed.

A Dataverse collection need not be "new" to accommodate a child collection, and in UNIX, the `mv` binary would overwrite its second argument with the first. Propose that it should read

> Moves a Dataverse collection whose ID is passed into an extant Dataverse collection whose ID is passed.

Secondly, I appreciate that this documentation belongs in the Admin Guide, but is there a reason these commands aren't linked from the API Guide?

Pull request forthcoming.

shlake:
Fix ambiguous documentation

**Which issue(s) this PR closes**:#8527

Closes #8527

### DIFF
--- a/doc/sphinx-guides/source/admin/dataverses-datasets.rst
+++ b/doc/sphinx-guides/source/admin/dataverses-datasets.rst
@@ -15,7 +15,7 @@ Dataverse collections have to be empty to delete them. Navigate to the Dataverse
 Move a Dataverse Collection
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
Yep, this helps clarify things. Thank you to @shlake for the pull request and to @donsizemore for pointing out how this could be worded better.

PR #9076, opened 2022-10-18
#9074 - Added support for building sphinx docs using python 3.10+

### LINKED ISSUE #9074: Currently defined Sphinx version (3.5.4) doesn't support python 3.10

- Have python 3.10 installed on host (tested with 3.10.6)
- Install Sphinx as stated in the [documentation](https://guides.dataverse.org/en/latest/developers/documentation.html)
- Build Sphynx guides

* When does this issue occur?
When building Sphynx guides using Python 3.10

* What happens?
An error is returned:

```
from types import Union as types_Union
ImportError: cannot import name 'Union' from 'types' (/usr/lib64/python3.10/types.py)
```

The Sphynx docs to be built successfully

v5.11.1

This issue is seconded by issues from in Sphynx repository:

- https://github.com/sphinx-doc/sphinx/issues/9816
- https://github.com/sphinx-doc/sphinx/issues/9562

j-n-c:

Add support for building Sphynx guides using Python 3.10+

- Closes #9074

Tested Sphynx version 5.3.0 with Python 3.9.7 and the docs were built successfully. However, chose to add rules for Python >= 3.10 and < 3.10 to ensure backward compatibility

- Install Sphynx requirements while using Python 3.9.x and build guides: Sphynx version 3.5.4 should be installed and build should be successful
- Install Sphynx requirements while using Python 3.10.x and build guides: Sphynx version 5.3.0 should be installed and build should be successful


pdurbin:
@donsizemore and I discussed this. The main wrinkle is that https://guides.dataverse.org is built by Jenkins worker notes running Rocky 8 which offers Python 3.9 (not 3.10 or newer). So we'll stick with Python 3.9 for the guides for now.

This means that devs should also stick with Python 3.9 for now. And they should stick with the same version of Sphinx (3.5.4) used in production.

That said, this PR allows developers to use Python 3.10 if they must. It also means they will be using a newer version of Sphinx (5.3.0). However, the older version of Sphinx that matches what is used in production should be preferred.

For now I think we can explain all this in requirements.txt as a comment. That is, a small update to this PR. I'd say it would be nice to mention how Python 3.9 is preferred here as well: https://guides.dataverse.org/en/5.12.1/developers/documentation.html#building-the-guides-with-sphinx

@j-n-c if you feel like doing this, please go ahead. If not someone can jump on your branch (assuming you allow us to push) and make these small doc changes.

mreekie:
added to sprint Dec 15, 2022

pdurbin (approved):
@j-n-c I left your fix in place (thanks!) but rewrote the comment to ask developers to use the same version of Sphinx that we use to build the production guides (which necessitates Python 3.9 or lower).

@kcondon for even a bit more context you can check out the post from @j-n-c to https://groups.google.com/g/dataverse-dev/c/fJ5U8mPyJ8U/m/lADDZVkBAgAJ

If you can still build the guides and it's still using Sphinx 3.5.4, we're good.

PR #9089, opened 2022-10-19
IQSS/7349-4 creator updates in schema.org

qqmyers:
**What this PR does / why we need it**: Per discussion in #7349 and #5029, Google ignores creator entries from Dataverse because they don't include an @type Person or Organization. Since we don't collect that directly, we have to use some mechanism to infer the @type. The OpenAire export format does this by leveraging an algorithm developed by DataCite. This PR abstracts that algorithm from the code producing the OpenAire xml format into a new PersonOrOrgUtil class that is then used to identify an @type for creators in Dataverse. The new functionality is then used to provide the @type in the schema.org export format and in-dataset-page metadata, along with the givenName and firstName for a person if/when the algorithm determines them.

The PR also updates how the 'affiliation' is handled. According to schema.org, only Persons have 'affilation' and that must be to an 'Organization'. The PR makes this change - to send the affiliation as an object of type Organization with the 'name' specified. Since Organizations don't have an 'affilation', but they do have a 'parentOrganization' key, for Organizations the code now encodes the affiliation for Orgs as a 'parentOrganzation' object of @type 'Organization' and 'name' as specified.

**Which issue(s) this PR closes**:

- Closes #7349 
- Closes #5029

**Special notes for your reviewer**:
FWIW: I think the new PersonOrOrgUtil can replace the code in the OpenAire export but I didn't do that in this PR. The algorithm is used in that code with two minor variations that I think can be handled by an appropriate entry for the organizationIfTied param.

**Suggestions on how to test this**: Unit tests cover most of this. As with others, visually inspecting the schema.org export and/or embedded datasetpage json-ld can be done to verify that Person or Organization is added and that affiliation is handled as described here. Note that miscategorizations (a person with @type Organization or vice versa) is not considered a bug and is considered better than not sending a type at all. If there are consistent issues, the algorithm can potentially be improved.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**: included.

**Additional documentation**:

Issue #7349, opened 2020-10-22
Improve/update Schema.org JSON-LD export


jggautier:
In a meeting with folks from the FAIRsFAIR group (namely @kitchenprinzessin3880) who are building and testing tools to access the "FAIRNESS" of datasets in Dataverse repositories (https://www.fairsfair.eu/fairsfair-data-object-assessment-metrics-request-comments), some changes were recommended for the metadata that Dataverse includes in the Schema.org JSON-LD metadata it exports for datasets. I said I'd open a Github issue so we could record and explain these changes.

**For license property, use the `@type` "CreativeWorks" and use "name" instead of "text":**
As of Dataverse 5.1.1, the `@type` for the Schema.org property "license" is "Dataset". Here's an example of what that looks like:
```
license: {
   @type: "Dataset",
   text: "CC0",
   url: "https://creativecommons.org/publicdomain/zero/1.0/"
```
or if CC0 is waived:
```
license: {
   @type: "Dataset",
   text: "Text the depositor entered in the Terms of Use field"
```

   Google's guide for [describing datasets with Schema.org](https://developers.google.com/search/docs/data-types/dataset#dataset) says to use the "CreativeWorks" `@type` for license and use "name".

Here's an example of what the license metadata in the Schema.org export might look like when this issue is merged (after the "multiple license" work described at https://github.com/IQSS/dataverse/issues/7440 and https://github.com/IQSS/dataverse/issues/7742 is also merged):

If the dataset depositor chooses a license from the list of licenses:
```
license: {
   @type: "CreativeWorks",
   name: "CC0",
   url: "https://creativecommons.org/publicdomain/zero/1.0/"
```
```
license: {
   @type: "CreativeWorks",
   name: "CC BY",
   url: "https://creativecommons.org/licenses/by/4.0/"
```
Or if no license is chosen and a custom license is entered:
```
license: {
   @type: "CreativeWorks",
   name: "Text entered in the "Dataset Terms" fields"
```

 
**For files (in the "distribution" property):**
As of Dataverse 5.1.1, here's an example of what the file metadata in the Schema.org export looks like:
```
distribution: [
    {
        @type: "DataDownload",
        name: "cases_by_infection.tab",
        fileFormat: "text/tab-separated-values",
        contentSize: 56377,
        description: "",
        @id: "https://doi.org/10.70122/FK2/LQKU61/SCM19X",
        identifier: "https://doi.org/10.70122/FK2/LQKU61/SCM19X",
[...]

PR #9102, opened 2022-10-28
IQSS/9100 OpenAire update for orgs

### LINKED ISSUE #9100: OpenAire export mis-characterizes some Organizations
**What steps does it take to reproduce the issue?** The algorithm discussed in #9089 fails for some Organization Names. PR #9089 adds two parameters to change the outcome and to suppress spurious given/family names for Orgs. That PR uses the new code for the schema.org export and in-page metadata, but does not back-port it to the OpenAire oai_datacite export so mischaracterizations can still occur there. Examples include 'The <anything> Project' from QDR and 'Portable Antiquities of the Netherlands' from DANS.

* When does this issue occur? In the OpenAire metadata export for Creator/Contact entries

* What happens? An org gets the nameType 'Personal' and has odd give/family names, e.g. givenName 'The' and familyName 'able Antiquities of the Netherlands' for the example above.

5.12 and earlier

qqmyers:
**What this PR does / why we need it**: Makes the OpenAire exporter use the new PersonOrOrgUtil class which allows ways to handle one-off cases where a name should be considered an Organization. Note this PR depends on/includes #9089 which adds changes to the PersonOrOrgUtil class to handle one-off Org names. The unique part of the PR just applies that to the OpenAire export and then refactors to move related classes out of the openaire package (since they aren't used there directly anymore).

Closes #9100

**Special notes for your reviewer**:  Depends on #9089. Also did some refactoring to move the classes detecting first names and organizations from the exporter/openaire package to util (since they are used by util.PersonOrOrgUtil which is used in Schema.org  and OpenAire currently and may get used elsewhere (e.g. if we want to update existing author fields with Person/Org type info if/when we add that to citation.tsv

**Suggestions on how to test this**: For regression, the auto tests assure that the output of the OpenAire code is the same as before. And the existing PersonOrOrg tests from #9089 check that the specific org name in the issue and the mechanism to handle one-off org names works. So - just verify the unit tests pass? Try some manual examples (add metadata, publish dataset, check the OpenAire export).

**Additional documentation**:


qqmyers:
The whole set of #7349-related PRs are fairly straight-forward change to the metadata we embed in the page for Google/others. They should be easy to review/QA. As this one involves some refactoring that involves/improves the OpenAire metadata export as well, I gave it a 10 - probably could have been a 3 as well.

jggautier:
Hey @qqmyers, when this is merged into a release and an installation applies that release, will the OpenAIRE exports of already published dataset versions be updated as well? Or will folks doing the updating need to do additional things to update the OpenAIRE exports of dataset versions that have already been published in that installation?

I'm asking because I saw recently that in 2 of the 3 installations that I could find that are using v5.13, Harvard and Dartmouth, for dataset versions published after v5.13 was applied to those installations, their Schema.org exports have been updated with the creator `@type` updates made in https://github.com/IQSS/dataverse/pull/9089. But the Schema.org exports of dataset versions published before v5.13 was applied have not been updated. And in [the v5.13 release notes](https://github.com/IQSS/dataverse/releases/tag/v5.13) I couldn't find anything about how to update the Schema.org exports of datasets published before the v5.13 update is applied.

So in the Dataverse release that this OpenAIRE update is included in, should there or will there be instructions about how to update the OpenAIRE exports of dataset versions published before the installation updates to that Dataverse release?

Or is this something that most of the folks doing these installation updates should know, and it was just missed when Harvard and Dartmouth were updated to v5.13 with its Schema.org update? All of QDR's datasets seem to have the Schema.org export updates, even its oldest datasets.

I read somewhere in Slack that @donsizemore is planning on helping update another Dataverse installation to v5.13. So maybe we could ask him if during the update he would know to take extra steps to update the Schema.org metadata of datasets published in that installation before v5.13 is applied.

qqmyers:
@jggautier - I don't think this PR changes the OpenAire output, it is just refactoring, but the earlier #9089 one could/should have included a release note about doing a reExportAll() (the main goal for the schema.org work was to change what's embedded in the page (which is not cached, since it is version dependent), but you're right that the schema.org export format is cached (always for the latest version only) and will only get updated if you do a reExport. I'll add a release note here to add a reExportAll to the v5.14 release (It may be that some of the other changes already say to require that in their release notes, but I'm not sure.)

landreev:
Is there a chance we could sneak in a fix for #9546 into this PR? 
(would totally understand if the answer is no)

landreev (approved):
I was considering sneaking in a fix for #9546, an unrelated openaire thing, into this PR but decided against it. (I made a note in that issue)
This is great stuff, like the cleanup of the util and the tests.

PR #9105, opened 2022-10-28
Add quotes around CVoc term URIs on advanced search page

### LINKED ISSUE #9104: Advanced Search of CVoc fields returns too many hits
**What steps does it take to reproduce the issue?** WIth CVoc fields, the search is against term URIs. The current code does not put these in quotes which results in solr matching against any word in the URI. The term URIs should be quoted.

AdvancedSearch

5.12

qqmyers:
**What this PR does / why we need it**: It adds quotes around termURIs for CVoc fields used in the advanced search page. Without them, the URIs are tokenized, resulting in too many hits (i.e. since the host is part of a term URI, any other terms with the same host/in the same vocabulary would be returned instead of only those with this specific term URI.

Closes #9104

**Special notes for your reviewer**: Discovered by DANS

**Suggestions on how to test this**: Should be able to test on a machine with CVoc, i.e. against a skosmos service as in our examples. If demo.dataverse.org or others aren't yet configured, DANS may be able to help with testing as they have CVoc configurations set up.

**Additional documentation**:


janvanmansum:
Tested this. LGTM.

Issue #8860, opened 2022-07-26
Using editMetadata API endpoint fails when editing non-multiple compound or vocabulary fields


doigl:
**What steps does it take to reproduce the issue?**
If one tries to change the value of a single (not multiple) vocabulary field via the edit metadata API endpoint, the response is an empty object "{}", nothing happens and there is a NullPointerException in the server log.

```
curl -H 'X-Dataverse-Key:$API_KEY' -X PUT "https://demo.dataverse.org/api/datasets/:persistentId/editMetadata?persistentId=doi%3A10.70122%2FFK2%2FN7XGAF&replace=true" --upload-file edit_journal.json -H 'Content-Type: application/json'
```

edit_journal.json
```
{
  "fields": [
    {
      "typeName": "journalArticleType",
      "value": "abstract"
    }
 ]
}
```
The same holds for single (not multiple compound fields) like Series

```
curl -H 'X-Dataverse-Key:$API_KEY' -X PUT "https://demo.dataverse.org/api/datasets/:persistentId/editMetadata?persistentId=doi%3A10.70122%2FFK2%2FN7XGAF&replace=true" --upload-file edit_series.json -H 'Content-Type: application/json'
```

edit_series.json

```
{
  "fields": [
    {
      "typeName": "series",
      "value": {
          {
           "typeName": "seriesName",
           "value" : "bla"
          }

      }
    }
 ]
}
```

* When does this issue occur?

* Which page(s) does it occurs on?
bug does not occur if one edits a dataset via GUI

* What happens?
API returns 
``` {} ```

Error message in server.log (for vocabulary field)
```
[2022-07-26T09:57:07.636+0200] [Payara 5.2021.6] [SEVERE] [] [edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler] [tid: _ThreadID=101 _ThreadName=http-api-pool::jk-api-connector(1)] [timeMillis: 1658822227636] [levelValue: 1000] [[
  _status="ERROR";_code=500;_message="Internal server error. More details available at the server logs.";_incidentId="561c1fc5-d9f0-47b4-a095-0149dde59e25";_interalError="NullPointerException";_requestUrl="https://demodarus.izus.uni-stuttgart.de/api/v1/datasets/:persistentId/editMetadata/?persistentId=doi:10.15770/darus-1460&replace=true";_requestMethod="PUT"|]]

[...]

Issue #7605, opened 2021-02-11
Edit metadata API: doesn't add or replace if the value "contains" the current value


scolapasta:
Using the edit metadata API, I tried both adding and replacing an author (note, in practice it is more likely useful as a "replace"):

Current value: Messi, Lionel
New Value: Messi, Lionel A.

Nothing happened - this is because the check uses contains logic (instead of equality) to check the metadata. This is because it is comparing to the display value which concatenates multiple fields, so we'll need to figure out a clever way to replace / fix this logic.

PR #9107, opened 2022-10-28
fix for cvv and editMetadata replace=true, and test

### LINKED ISSUE #9106: EditMetadata replace doesn't work correctly
**What steps does it take to reproduce the issue?** Reported by DANS - When using replace=true with /api/datasets/\<id\>/editMetadata, values that are in-common with the current value are not added. I.e. when the language field has `English, Japanese, Korean` and you replace with `English, Dutch, Korean`, the result is only `Dutch` . DANS also correctly identified the underlying issue: at line https://github.com/IQSS/dataverse/blob/42e2b59811576bd6034eb5c1dd25bce7a57cea31/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L993 cvvDisplay contains the original values even in the replace=true case.

5.12

qqmyers:
**What this PR does / why we need it**: Fixes the case in the issue - editMetadata api call with replace=true and a cvv field.

Closes #9106

**Special notes for your reviewer**: kudos to DANS for making this easy

**Suggestions on how to test this**:  Use the editMetadata api call with replace=true on a cvv field like subject or language and verify that when the replacement list contains one of the values in the current field value that it remains, i.e. that when you replace with n values where one or more was in the original value, you get all n values. FWIW: The change to the test should check this case - Mathematical Sciences should be a current value and the replacement is with Mathematical Sciences and Social Sciences and the test checks that both values exist at the end.

**Additional documentation**:


pdurbin:
@qqmyers any chance this also fixes either of the following issues?

- #7605
- #8860

qqmyers:
maybe - #8860 looks like a good possibility while #7605 seems to suggest 'nothing' happens vs. the term disappears, but perhaps that is without replace=true. This PR only changes the replace=true logic so if #7605 involves not using replace-true, that part of it would still remain. In any case, certainly worth testing those cases.

mreekie:
This is a PR. Dev is don.
This is a one line fix for something that Don found and diagnosed.
This comes with the fix and automated tests
It is very small.

mreekie:
Added to 12/1/2022 sprint on 12/5/2022

sekmiller (approved):

Issue #7195, opened 2020-08-13
DownloadFilesIT/downloadAllFilesRestricted Failing on build 555


djbrooke:
We are down to one failing test 😄 

On build 555 of Jenkins (not linking here so the bots don't find it) there's more info.

landreev:
Copying what I said on slack; this appears to be this message hard-coded in DownloadFilesIT.java: 
```
downloadFiles3.then().assertThat()
                .statusCode(FORBIDDEN.getStatusCode())
                .body("status", equalTo("ERROR"))
                .body("message", equalTo("'/api/v1/access/dataset/%3ApersistentId' you are not authorized to access this object via this api endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org."));
```
and the actual message has changed, after the "error handler" PR got merged. So it simply needs to be adjusted accordingly.

scolapasta:
Do we even need to check the exact message? I think confirming that we got the write status code and error would be enough? (i.e. we shouldn't have to fix this test every time we decide to change the messaging)

landreev:
Agree. I don't think it's necessary at all, to check the exact message.

landreev:
(it's in QA now - but I don't think it needs any QA other than having it pass the tests; which has been confirmed: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-7198/)

PR #9112, opened 2022-10-31
Adding a mechanism for disabling new remote auth signups. 

### LINKED ISSUE #9111: Add a mechanism for turning off OAuth account signups without disabling existing OAuth accounts
Should be a very straightforward fix.

landreev:

We all know why we need it. 
This implementation allows to disable new signups from specific remote/oauth type, without blocking all the existing accounts authenticated by this method. 
This is controlled by a new *compound* setting `:AllowRemoteAuthSignUp`. By default, if the setting is not present, all remote signups are open. If the setting is present, but the value for this specific method is not specified, it is assumed that the signups are allowed. 
Examples: 
```
curl -X PUT -d '{"default":"false"}' http://localhost:8080/api/admin/settings/:AllowRemoteAuthSignUp
```
disables all remote signups. 
```
curl -X PUT -d '{"default":"true","google":"false"}' http://localhost:8080/api/admin/settings/:AllowRemoteAuthSignUp
```
keeps signups open for all the methods except google. (but note that the `"default":"true"` part in this example is redundant, since it would default to true anyway for all the methods other than google). 

I am sure there are prettier ways to implement this. But, seeing how this is a bit of an emergency, I'm not super concerned about prettiness at this point. (I will move the error messages into the bundle though). 

I am concerned about whether there are still ways to go around this somehow. So it would definitely help to know how those users were able to create their google accounts after I had quote-unquote "disabled" it. 

Closes #9111

**Additional documentation**:


scolapasta (commented):
Overall looks good. A few comments:
• I think you noted that you would move the texts into bundles. 
• This setting should be added to the docs
• Are there any tests we can add? There really isn't a concept of logging in through the API, so not sure what exactly we could add.

landreev:
Did the cleanup/added the doc. section. 
I couldn't think of how to add tests for this. It may be possible; just like I think it should be possible to implement this in some more elegant way. But I would handle any such improvements as a separate issue in a normal release cycle. This patch I'm just itching to throw in prod. asap.

scolapasta (approved):

qqmyers:
I think this one has been seen before and may be timing related - another case where a sleepForLocks/sleepForReindex might help?

pdurbin:
I tested this on dataverse-internal and I'm satisfied that it's working properly. Will merge shortly.

I tweaked the docs in cf00e61 to emphasize that even though "remote" is in the name of the new database setting, for now this only affects OAuth accounts. I also did a little reorg, cross-linking, clean up, etc.

Here's how I tested.

## Regression test

I checked that I could still create a fresh Google account with nothing configured.

Then I deleted it.

## Am I prevented from creating a Google account?

Yes, by following the example in the guides, I get this error when I try:

Screen Shot 2022-11-03 at 2 29 48 PM

Screen Shot 2022-11-03 at 2 30 00 PM

Screen Shot 2022-11-03 at 2 30 21 PM

## With signup disabled, can I still log in with Google?

Yes, works fine.

Nice feature!

pdurbin:
> may be timing related

I just checked the tests on develop after I merged this pull request and they all passed: https://jenkins.dataverse.org/job/IQSS-dataverse-develop/1192/testReport/

So, yes, probably timing related. It looks like we fixed DownloadFilesIT.downloadAllFilesRestricted here but it wasn't timing related then:

- #7195

Anyway, we can create a new issue if it comes up again.

PR #9125, opened 2022-11-03
8346 License internationalizataion

### LINKED ISSUE #8346: Support translation of license names and descriptions after #7920
#7920 adds support for allowing users to choose from a set of allowed licenses configured by installation admins. It does not provide a mechanism to allow license names/descriptions to be translated into other languages for the UI (Names, which are often acronym based may not always be things you'd translate, but some might). Scholars Portal previously added the capability for admins to provide translations of external tool names that looks like it could be adapted here (basically a bundle string key is defined based on the tool name and the usual mechanism is then used to look for the value of that key in different language files).

per review feedback

#6748 - the external tool work - note that further changes/fixes were made after the initial PR, but looking at the issue and PR will get you to the right places in the code.

JayanthyChengan:

Closes #8346


qqmyers (approved):
Looks good. The one question I have is whether it would make sense to refactor to have a getLocalizesLicenseDetails(License, String) or (License, Enum) interface. For the former that would mean doing license.getName() once within the method instead of all the callers.

JayanthyChengan:
@qqmyers  - Please take a look at the changes when you have time.  Thanks

pdurbin:
Gave it a 10. Should be just testing with French or whatever.

mreekie:
Added to 12/1/2022 sprint on 12/5/2022

pdurbin (approved):
I didn't test this but the code and docs look fine to me. All the API tests passed. Approving.

JayanthyChengan:
@pdurbin :
At ScholarsPortal, we noticed  "500 Internal Server Error" when we go to the root dataverse and click on edit > dataset templates.
PR is submitted https://github.com/IQSS/dataverse/pull/9226

Issue #1955, opened 2015-04-10
Google analytics for Dataverse.harvard.edu


eaquigley:
Need to take out the GA code currently in and replace it with the GA code for Dataverse.harvard.edu

eaquigley:
Code for dataverse.harvard.edu has been emailed to @raprasad 

Needs to be placed in a way that it shows up on all pages. We've done this for dvn-build and dataverse-demo already.

raprasad:
Updated.
- Placed in a separate .xhtml file: google-analytics-snippet.xhtml
- This snippet is included by the dataverse_template.xhtml

commit: https://github.com/IQSS/dataverse/commit/3a84ca0d39eac275657344f917b62d850dd9f8b7

raprasad:
reference: http://stackoverflow.com/questions/4245132/google-analytics-api-filter-by-uri

PR #9129, opened 2022-11-03
IQSS/9126- Fix workflow token access

### LINKED ISSUE #9126: Some file access methods ignore workflow tokens
As reported by DANS, some file access methods e.g., the individual file download ignore the X-Dataverse-invocationID authentication token used by workflows. Looking at the code, it appears this is because these methods do their own check for an apiKey/header, usually before calling the AbstractAPIBean findUserOrDie method which does its own lookup of the key or header (and checks for workflow tokens). Removing the unnecssary code in the Access class should fix this issue and simplify processing. (The lookups in the Access class also have some historical oddities such as checking that the apiKey is not 64 characters long (in which case it would be rejected)).
5.12

qqmyers:
**What this PR does / why we need it**: This PR removes some of the custom permissions checking in the Access class which had been stopping workflow token access to these endpoints.

Closes #9126 

**Special notes for your reviewer**: I've tried to comment some in the PR - the logic here was pretty convoluted and I think had some dead code (which I noted back when embargoes went in). Hopefully cleaner now but would be good to have someone check my assumptions about the old code as noted in the comments.

**Suggestions on how to test this**: New functionality - easiest to have DANS test it - you need to run an async workflow and use the workflow token to access the download file endpoints to test it. Regression - this could have but shouldn't break any access by Guest/PrivateUrl/Authenticated Users with appropriate permissions (DownloadFile for published restricted or embargoed files, ViewUnpublishedDataset for unpublished files) to the main file, aux files, etc.
[from Leonid:] For regression, I just want to second the above, about retesting extra carefully the less trivial access cases. In addition to the ones listed above, maybe attempting a zip download, via the api,  of a number of files only some of which the user is authorized to access.

**Is there a release notes update needed for this change?**: as a workflow related bug fix, probably not?

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -187,33 +187,26 @@ public class Access extends AbstractApiBean {
...
-    
     //@EJB
     
     // TODO: 
     // versions? -- L.A. 4.0 beta 10
     @Path("datafile/bundle/{fileId}")
     @GET
     @Produces({"application/zip"})
-    public BundleDownloadInstance datafileBundle(@PathParam("fileId") String fileId, @QueryParam("fileMetadataId") Long fileMetadataId,@QueryParam("gbrecs") boolean gbrecs, @QueryParam("key") String apiToken, @Context UriInfo uriInfo, @Context HttpHeaders headers, @Context HttpServletResponse response) /*throws NotFoundException, ServiceUnavailableException, PermissionDeniedException, AuthorizationRequiredException*/ {
+    public BundleDownloadInstance datafileBundle(@PathParam("fileId") String fileId, @QueryParam("fileMetadataId") Long fileMetadataId,@QueryParam("gbrecs") boolean gbrecs, @Context UriInfo uriInfo, @Context HttpHeaders headers, @Context HttpServletResponse response) /*throws NotFoundException, ServiceUnavailableException, PermissionDeniedException, AuthorizationRequiredException*/ {
```
qqmyers, line 197: Most api calls don't add the key as a QueryParam so I've removed them in this class.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -187,33 +187,26 @@ public class Access extends AbstractApiBean {
...
-        }
-        
         // This will throw a ForbiddenException if access isn't authorized: 
-        checkAuthorization(df, apiToken);
+        checkAuthorization(df);
         
         if (gbrecs != true && df.isReleased()){
             // Write Guestbook record if not done previously and file is released
-            User apiTokenUser = findAPITokenUser(apiToken);
+            User apiTokenUser = findAPITokenUser();
```
qqmyers, line 209: This calls findUserOrDie which will retrieve the key param or api token header, or the workflow token header.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -187,33 +187,26 @@ public class Access extends AbstractApiBean {
...
         
         DataFile df = findDataFileOrDieWrapper(fileId);
         
-        if (apiToken == null || apiToken.equals("")) {
-            apiToken = headers.getHeaderString(API_KEY_HEADER);
-        }
-        
         // This will throw a ForbiddenException if access isn't authorized: 
-        checkAuthorization(df, apiToken);
+        checkAuthorization(df);
```
qqmyers, line 205: comments in this method.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -859,15 +833,15 @@ public void write(OutputStream os) throws IOException,
                         logger.fine("token: " + fileIdParams[i]);
                         Long fileId = null;
                         try {
-                            fileId = new Long(fileIdParams[i]);
+                            fileId = Long.parseLong(fileIdParams[i]);
```
qqmyers, line 836: unrelated - removed deprecated methods.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1436,8 +1410,8 @@ public Response requestFileAccess(@PathParam("id") String fileToRequestAccessId,
             List<String> args = Arrays.asList(wr.getLocalizedMessage());
             return error(BAD_REQUEST, BundleUtil.getStringFromBundle("access.api.fileAccess.failure.noUser", args));
         }
-
-        if (isAccessAuthorized(dataFile, getRequestApiKey())) {
+        //Already have access
+        if (isAccessAuthorized(dataFile)) {
```
qqmyers, line 1414: added a comment - my understanding is that in a request access call, if you already have access, this is a BAD_REQUEST

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1787,37 +1761,41 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-         * note that the fragment below - that retrieves the session object
-         * and tries to find the user associated with the session - is really
-         * for logging/debugging purposes only; for practical purposes, it 
-         * would be enough to just call "permissionService.on(df).has(Permission.DownloadFile)"
-         * and the method does just that, tries to authorize for the user in 
-         * the current session (or guest user, if no session user is available):
          */
         
-        if (session != null) {
+        User apiTokenUser = null;
```
qqmyers, line 1781: The logic used to handle 3 cases separately, mostly so that fine logging could be done to indicate which branch was being used: session - logged in user, guest user, and apitoken user. One tricky thing is that both session and apitoken calls could return a GuestUser and this wasn't broken out.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1787,37 +1761,41 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
+        //If we get a non-GuestUser from findUserOrDie, use it. Otherwise, check the session
```
qqmyers, line 1782: The logic now looks for an apitoken authenticated user and uses it if it exists. If not, and a session user exists, we use that. If the apitoken method indicates a GuestUser, we will use that if there's no session.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1787,37 +1761,41 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
+        try {
+            logger.fine("calling apiTokenUser = findUserOrDie()...");
+            apiTokenUser = findUserOrDie();
```
qqmyers, line 1785: With the rearchitecting, I assume this call will end up handling both session and token logins and this part of the code will be able to be replaced. Until then, this is the only part of the api that was already designed to be usable with sessions or tokens.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-                logger.log(Level.FINE, "Message from findUserOrDie(): {0}", wr.getMessage());
-            }
-            
-            if (apiTokenUser == null) {
-                logger.warning("API token-based auth: Unable to find a user with the API token provided.");
-            }
+        //If we don't have a user, nothing more to do. (Note session could have returned GuestUser)
+        if (user == null && apiTokenUser == null) {
+            logger.warning("Unable to find a user via session or with a token.");
+            return false;
```
qqmyers, line 1813: The code had/has a check for both being null at the end, but there's no reason to proceed further if we have no user.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-            } catch (WrappedResponse wr) {
-                logger.log(Level.FINE, "Message from findUserOrDie(): {0}", wr.getMessage());
-            }
-            
-            if (apiTokenUser == null) {
-                logger.warning("API token-based auth: Unable to find a user with the API token provided.");
-            }
+        //If we don't have a user, nothing more to do. (Note session could have returned GuestUser)
+        if (user == null && apiTokenUser == null) {
+            logger.warning("Unable to find a user via session or with a token.");
```
qqmyers, line 1812: Looking in the old code - checking for the key length to not be 64 chars is very old and seems to be backwards even if there was a time the keys were 64 chars. Regardless, since the Access class doesn't need to check the token any more (all handled in the AbstractApiBean.findUserOrDie method), this isn't needed anymore.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
         // (if (published) was already addressed above)
-        
-        if (!restricted && !embargoed) {
+
+        DataverseRequest dvr = null;
+        if (apiTokenUser != null) {
+            dvr = createDataverseRequest(apiTokenUser);
+        } else {
+            // used in JSF context, user may be Guest
+            dvr = dvRequestService.getDataverseRequest();
```
qqmyers, line 1825: Simplifying the code below by getting the appropriate request ready now. This means we can't use the permissionService.on() shortcut call, but it also means that session and token users are handled in the same single lines below.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
+        }
+        if (!published) { // and restricted or embargoed (implied by earlier processing)
```
qqmyers, line 1827: Since published and not restricted/embargoed is handled above, the main split now is whether it is published or not. If it's published, the only case left is with restricted/embargoed. With unpublished, both the restricted/embargoed and not restricted/embargoed both get handled the same way.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-                if (permissionService.requestOn( createDataverseRequest(apiTokenUser), df.getOwner()).has(Permission.ViewUnpublishedDataset)) {
-                    logger.log(Level.FINE, "Token-based auth: user {0} has access rights on the non-restricted, unpublished datafile.", apiTokenUser.getIdentifier());
-                    return true;
-                }
-            }
 
-            // last option - guest user in either contexts
-            // Guset user is impled by the code above.
-            if ( permissionService.requestOn(dvRequestService.getDataverseRequest(), df.getOwner()).has(Permission.ViewUnpublishedDataset) ) {
+            if (permissionService.requestOn(dvr, df.getOwner()).has(Permission.ViewUnpublishedDataset)) {
```
qqmyers, line 1831: Given the changes above, this line now handles all three authenticated session user, token user, and guest cases.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-                        logger.log(Level.FINE, "API token-based auth: User {0} has rights to access the (unpublished) datafile.", user.getIdentifier());
-                        //Same case as line 1843?
-                        return true;
-                    } else {
-                        logger.log(Level.FINE, "API token-based auth: User {0} is not authorized to access the (unpublished) datafile.", user.getIdentifier());
-                    }
-                }
-            } else {
-                logger.log(Level.FINE, "API token-based auth: User {0} is not authorized to access the datafile.", user.getIdentifier());
+            if (permissionService.requestOn(dvr, df).has(Permission.DownloadFile)) {
```
qqmyers, line 1841: same here - all three cases. Also note in the old code I think the lines after #1955 were unreachable, repeating a case handled earlier.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1829,192 +1807,72 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
-                logger.fine("calling apiTokenUser = findUserOrDie()...");
-                apiTokenUser = findUserOrDie();
-                return apiTokenUser;
-            } catch (WrappedResponse wr) {
-                logger.log(Level.FINE, "Message from findUserOrDie(): {0}", wr.getMessage());
-                return null;
+        try {
+            logger.fine("calling apiTokenUser = findUserOrDie()...");
+            apiTokenUser = findUserOrDie();
+            if(apiTokenUser instanceof GuestUser) {
```
qqmyers, line 1862: As noted, the idea here is to not let a guest user returned from findUserOrDie (which happens when there is no key/token, and which we want if there's no session) from overriding an authenticated session user. This could probably be cleaned up further, but I'm again assuming the rearchitecture work will end up doing that in the AbstractApiBean soon.

mreekie:
added to sprint Dec 15, 2022

src/main/java/edu/harvard/iq/dataverse/api/Access.java
landreev: Those 64 characters-long tokens were used by GeoConnect/Worldmap; that system needed token-based authentication, back when our standard REST API did not support tokens. So that subsystem had its own tokens implemented and we had to support them for a while, for backward compatibility. I understand that we got rid of everything worldmap-related in the code base... (but, just in case, somebody correct me if I'm wrong!)

src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1787,37 +1761,41 @@ private boolean isAccessAuthorized(DataFile df, String apiToken) {
...
+            // If they are not published, they can still be downloaded, if the user
             // has the permission to view unpublished versions! (this case will 
             // be handled below)
-            if (published) { 
-                return true;
-            }
+            return true;
         }
         
+        //For permissions check decide if we havce a session user, or an API user
```
landreev, line 1774: typo in "have"

src/main/java/edu/harvard/iq/dataverse/api/Access.java
landreev: Note to self: we need to retire this "bundle access" method.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
landreev: I like how the code has been reorganized and simplified. And the logic seems correct. But could you please clarify the "this wasn't broken out" part above. Aside from the old code being convoluted and unwieldy, was there actually a problem with the old logic, was there a situation where the guest user returned by the apitoken call would overwrite an authenticated user from the session? - I've stared at the old code for a while now and I'm not seeing that, but figured I would ask to confirm.

(We may have assumed in the past that nobody would be calling the api with _both_ a session and a token in real life).

landreev (commented):
I'm glad to see this logic streamlined and simplified. 
I asked a couple of simple questions, and there's a trivial typo, but I'm generally ready to move it into qa. 
I appreciate the comments explaining what's going on. My only question is, is there a reason not to incorporate at least some of them into the code?

src/main/java/edu/harvard/iq/dataverse/api/Access.java
qqmyers: I don't think it was broken, so no real issue. I think my comment related to the old line 1877 and the code after it - there was no distinction between guest coming back from a session or from the apitoken logic. In refactoring, I had to make sure that both were still handled, e.g. in line 1790 below.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
qqmyers: fixed

src/main/java/edu/harvard/iq/dataverse/api/Access.java
landreev: I was looking at that code after line 1877 - it was just meaningless, wasn't it? I.e., it doesn't seem to be doing anything that the line 1861 hasn't already attempted. 
It just looks like the old code may have been written with the assumption that `session == null` when it's a direct API call - and it should never be null of course.

landreev (approved):

mreekie:
QA is left:
Size for this sprint: 3

PR #9132, opened 2022-11-04
Dataset files cleanup

### LINKED ISSUE #9130: Cleanup of the files in dataset storage that are not linked to that dataset
As described by @qqmyers in #9002 comments:
W.r.t. temp files - we have this issue in general (i.e. if a user does a UI direct upload and leaves the page without hitting cancel or save, Dataverse doesn't know and doesn't clean up the files. Similarly in the direct upload API, if the final /addFiles call isn't done, the files are abandoned. In S3 systems that support it, we add a tag so these temp files can be easily found and deleted (i.e. via a cron job), but it would be nicer if there were an API call to remove files unrelated to a dataset on S3 (or file system, etc.) Such an API would have to deal with files in different versions and all of the auxiliary files, thumbnails, metadata exports that should remain but delete any files with storageidentifiers that aren't in the dataset.

#9002

ErykKul:
In order to conserve storage space, you might want to clean up files in datasets' storage location. Most importantly, when uploading files directly, or by user action, there could be files that failed to upload fully and remain unlinked to the dataset. With the new API call, you can remove all the files not linked to the dataset.

Closes #9130

This implementation removes all the files that are not in the file list of a specific dataset. Maybe the thumbnails and export cache files should remain in the storage? It could be useful to be able to remove these files, e.g., in datasets that are not accessed very often. You could also have thumbnails of the files from previous versions of the dataset, possibly no longer present the newest version, etc.

For now, I have kept it simple, and the code removes all not linked files. If required, I could add a functionality that files with names that start with a name of a linked file are left untouched.

Most importantly, I have this code only tested with local file storage. This code could be dangerous if used in production without prior testing...


qqmyers:
@ErykKul - haven't yet looked at code, but thanks for this! I was just considering starting in on it. I would definitely like to see this expanded to preserve all the files that could be there, at least as an option (maybe the default) - thumbs, cached exports, aux files, external dataset thumbs, provenance file, original files (for ingested tabular files). For S3, something that would also remove partial multipart uploads would be useful but could/should probably be a separate PR. A list-only version that just identify extra files without deleting would be useful as well.

I'm happy to contribute, test on S3, etc. to help you move this forward - just let me know when/where you want help. As a first step I'll try to look through your current code soon.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2502,6 +2503,65 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
...
+            for (DatasetVersion dv : dataset.getVersions()) {
+                for (FileMetadata f : dv.getFileMetadatas()) {
+                    String storageIdentifier = f.getDataFile().getStorageIdentifier();
+                    String location = storageIdentifier.substring(storageIdentifier.indexOf("://") + 3);
+                    String[] locationParts = location.split(":", 3);//separate bucket, swift container, etc. from fileName
+                    files.add(locationParts[locationParts.length-1]);
+                }
+            }
+            StorageIO<DvObject> datasetIO = DataAccess.getStorageIO(dataset);
+            List<String> allDatasetFiles = datasetIO.listAllFiles();
```
qqmyers, line 2549: One quick thought would be to just have a StoageIO.cleanDataset() method and keep the list and delete methods out of the public interface. There might be some circumstances where listing or deleting in the main code body might be useful, but I think it would be easier to maintain integrity if stores manage that directly. (FWIW: I think all stores currently use the same authority/identifier/filename format but they don't have to so putting the logic to identify valid files within the store would be important in that case.)

ErykKul: I have done some refactoring. The thing that I am struggling right now is which files to keep. So I have a list of all files from all versions of the dataset (here called `files`) and right now I keep the files that start with any name in that list:
```
Predicate<String> filter = f -> {
  return files.stream().noneMatch(x -> f.startsWith(x));
};
```
I could use some help extending this filter to keep all files we want to keep. I can add an option to delete more files later on, if needed.

donsizemore:
Yes, thank you for this. Is it possible to have the endpoint default to a `listonly` or `dryrun` mode, then actually fire when passed an `execute` or `force` flag? Just so admins can see what would be deleted first...

ErykKul:
I have added the ``dryrun`` query parameter. I have also updated the filter to include the export files:
````
Predicate<String> filter = f -> {
    return !f.startsWith("export_") || files.stream().noneMatch(x -> f.startsWith(x));
};
````

**Comment by mreekie** [comment]
This one needs some care in looking at.
It's about cleaning up files.
Size 33

**Comment by mreekie** [comment]
This may be related to work Oliver did on #8983

File: src/main/java/edu/harvard/iq/dataverse/api/Datasets.java [hunk 2]
```diff
@@ -2502,6 +2504,67 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
...
+        
+        // check permissions
+        if (!permissionSvc.permissionsFor(createDataverseRequest(authUser), dataset).contains(Permission.EditDataset)) {
+            return error(Response.Status.INTERNAL_SERVER_ERROR, "Access denied!");
+        }
+
+        List<String> deleted = new ArrayList<>();
+        Set<String> files = new HashSet<String>();
+        try {
+            for (DatasetVersion dv : dataset.getVersions()) {
```
landreev, line 2541: Is there a reason for iterating through the versions->FileMetadatas here? 
As opposed to just 
```
for (DataFile dataFile : dataset.getFiles()) {
```
? - then you wouldn't have to iterate through the same files that occur in more than one versions. Not a big deal, obviously.

landreev:
> ... I have also updated the filter to include the export files:
> 
> ```
> ...
>     return !f.startsWith("export_") || ...;
> ```

This is the biggest potential problem I'm seeing so far. I don't think it is safe at all, to assume that the only useful _dataset-level_ auxiliary files are the cached metadata export files. The export files are at least mostly safe to erase (they will be regenerated automatically when needed). But the `dataset_logo*` files (custom dataset thumbnails) that it would end up deleting with the filter above are not easily recoverable. The only other real-life examples of non-export dataset-level aux. files I can think of right away were some GEO/WorldMap-specific files that are no longer in use. But we should definitely assume that it is possible, for such files to exist. I.e., if we hard-code the valid filename prefixes that are in use now, future developers will need to know/remember to add to the list if they ever need to add aux. files that start with something else (and that would be asking for trouble). 

Not sure what the best/cleanest way to deal with this is. It's possible that the best we can do is to only delete the files with the filenames that match `^[0-9a-f]{11}-[0-9a-f]{12}\..*`. That should catch all the unlinked datafile-level main files and their derivative that match our current standard file naming scheme. This will NOT catch some ancient files for a (very few) legacy installations - such as Harvard and Odum and a couple of others - that have grandfathered files that don't look like this. But I don't think this is a big issue - if any of us old-timer Dataverses still have some junk files from a decade+ back wasting storage space - that's probably on us. 

Anyone? - Does the regex above really match all possible (reasonably modern) datafile-level physical files? (Does this work with RemoteOverlay?) 

OK, we'll figure this out. (Sorry to be difficult :)

qqmyers:
I agree that just deleting datafiles not in the database ( `^[0-9a-f]{11}-[0-9a-f]{12}\..*` ) would be a safer option. I didn't check that regex just now but the new code that assures you can only send in valid storageidentifiers should be using a regex so we can compare with that. 

There are definitely more files that just export_* (which as  you say, will be regenerated) at the dataset level that need to remain - the json provenance file is one of them - that is user input and can't be recreated. Right now, creating a whitelist would be more trouble than a blacklist approach. If/when the dataset-level aux files are organized like the file level ones (all with the same naming pattern), a whitelist might be easier.

Not sure what that means for this PR. Having the ability to clean failed uploads is probably 90% of the problem and arguably safer, so it would be nice if that were an/the option. If there's interest in trying to assemble a whitelist of all additional files that can exist, I'd be happy to help trying to track down the sources, but I think it would be OK to make that future work as well.

One other thing to be aware of - it is probably an infrequent case but files in a dataset may not all be in the same store. One can change the store used by the dataset to save new files after some files have been uploaded. I *think* the auxiliary files for the dataset are always in the store for the dataset but the datafiles and their auxfiles could be spread out.

landreev:
(sorry, that regex I first posted was specifically for aux. files - i.e., for filenames with extensions; it needed one extra character to match the main files too; I corrected it in my comments above)

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
ErykKul: I did not know I could do that, I have implemented your suggestion. Thank you.

ErykKul:
@landreev, @qqmyers 
Thank you for the comments and suggested improvements. I have refactored the code to make it more readable. I have also fixed some of the problems that were in the code. The code is made safer with the regex:
- first, I ensure that all the files in the dataset follow the regex pattern (no legacy files present), otherwise an error is thrown
- then, I only delete files that match the regex and their names do not start with a name of a file linked to the dataset

The list of files linked to the dataset is now retrieved as suggested: by looping over the files in the dataset (not the file metadata of every version separately).

Finally, I have added a unit test that verifies the working of the "to delete files" filter. It also illustrates which files are deletable.

landreev:
Thank you @ErykKul. I'm looking at the latest changes now.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2502,6 +2506,76 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
...
+        if (!permissionSvc.permissionsFor(createDataverseRequest(authUser), dataset).contains(Permission.EditDataset)) {
+            return error(Response.Status.INTERNAL_SERVER_ERROR, "Access denied!");
+        }
+
+        boolean doDryRun = dryrun != null && dryrun.booleanValue();
+
+        // check if no legacy files are present
+        Set<String> datasetFilenames = getDatasetFilenames(dataset);
+        if (datasetFilenames.stream().anyMatch(x -> !dataFilePattern.matcher(x).matches())) {
+            return error(Response.Status.INTERNAL_SERVER_ERROR, "Dataset contains files not matching the nameing pattern!");
```
landreev, line 2545: > first, I ensure that all the files in the dataset follow the regex pattern (no legacy files present), otherwise an error is thrown ... 

But, do we really need to do this? I mean, just because it's a potentially ancient dataset (and/or a dataset that may have been modified in some custom way) doesn't mean that we are not going to find any recent junk files that our regex is going to match. So, it is possible that we may still be able to recover some wasted storage space - ? 

Or do you feel that it's just safer, to communicate to the admin that "this is a weird/non-standard dataset, we don't want to touch it, you're on your own maintaining it?"

landreev:
Sorry for the delay with reviewing the PR; I keep getting distracted by other things, but will try to wrap it up asap. 
I just tested it with filesystem-stored files; happy to see that my regex appears to be working (w/ the extra backslash to escape the literal . - yes). I mean, I only tried it with egrep, wasn't 100% sure if it was  going to work w/ Java matcher. 
I do want to test it on s3 some more, so I'll go do that. 
(I may suggest that we tighten that regex a bit further, let me think about that).

landreev:
The first part of our generated physical file name - `[0-9a-f]{11}` - that's the timestamp of when the file was created, the number of milliseconds since 01-01-1970, hex-encoded. 
If my quick math is correct, the regex will be valid for 500+ more years, before the number becomes 12 digit... so, I think we are good?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2502,6 +2506,76 @@ public Response addFileToDataset(@PathParam("id") String idSupplied,
...
+        return ok("Found: " + datasetFilenames.stream().collect(Collectors.joining(", ")) + "\n" + "Deleted: " + deleted.stream().collect(Collectors.joining(", ")));
+        
+    }
+
+    private static Set<String> getDatasetFilenames(Dataset dataset) {
+        Set<String> files = new HashSet<>();
+        for (DataFile dataFile: dataset.getFiles()) {
+            String storageIdentifier = dataFile.getStorageIdentifier();
+            String location = storageIdentifier.substring(storageIdentifier.indexOf("://") + 3);
+            String[] locationParts = location.split(":", 3);//separate bucket, swift container, etc. from fileName
```
landreev, line 2567: This limit = 3 in split() - that's because of swift... right? - must be; but I can barely remember how swift works. :(

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1513,6 +1513,34 @@ The fully expanded example above (without environment variables) looks like this
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your Dataverse installation has been configured to support direct uploads, or in some other situations,
+you could end up with some files in the storage of a dataset that are not linked to that dataset directly. Most commonly, this could
+happen when an upload fails in the middle of a transfer, i.e. if a user does a UI direct upload and leaves the page without hitting cancel or save,
+Dataverse doesn't know and doesn't clean up the files. Similarly in the direct upload API, if the final /addFiles call isn't done, the files are abandoned. 
+
+You might also want to remove cached export files or some temporary files, thumbnails, etc.
+
+All the files stored in the Dataset storage location that are not in the file list of that Dataset can be removed, as shown in the example below.
```
landreev, line 1528: Is it necessary to clarify here, that only the files that are looking like leftover *datafiles* will be deleted here, that Dataverse will skip any files that it can't identify as such for sure? So, if they happen to have anything else useful stored in the same directory - not that we ever recommend doing that - it will be spared? 
I'm not sure about this, whether this is in fact necessary, to explain this. So I'm leaving this up to you.

landreev: But it would be great to add a reminder here that this is an _experimental_ feature, like in the release note; and a suggestion to make sure their backups are up-to-date before attempting this on prod. servers. And also, could you please document the "dryrun" parameter, and tell them sternly, to start with that. Thanks!

landreev:
I tested it under S3 a bit, it's all looking good - thank you!
I don't think it's necessary, to mess with that regex any further. 
So, the LAST thing I'm worried about: swift. Do you happen to have a way to test it? Because I don't think we do at this point.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
ErykKul: How I see it, is that it is no problem to let the cleanup do its thing on the legacy datasets. It should never remove something that may not be removed: the regex does not allow it. The only reason to not do that is that it could give a false sense of data being clean. A warning i.s.o. error sounds good. I will change the code that it gives a warning about the legacy files → these will not be cleaned, only the files following the naming conventions will be cleaned.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
ErykKul: I have removed the limit. The reasoning behind it was that there could be ":" somewhere in the name of the container, etc. The current solution with your regex is much better: if whatever that comes after the last ":" does not match the pattern, it will not be cleaned. This way, even if there was a ":" in the filename, we will not delete any files we are not supposed to.

doc/sphinx-guides/source/api/native-api.rst
ErykKul: Thanks. I will rework the documentation to reflect the current solution and add the "experimental feature" warning there. The dry run parameter is a very good idea for testing before using it, and should be explained (I did not do that yet after adding the feature...). I will give extra warning to the swift users, as it is not yet tested on swift...

ErykKul:
Swift is a problem. I can not test it neither. The current solution with your regex is much safer, we do not attempt removing files that do not match the pattern. And, if the remove fails, the file remains in the system. I think it is OK to release it with additional warnings for swift that it is not tested. The better solution would be to find someone who uses swift and can test it...

pdurbin:
Borealis (née Scholars Portal) is using Swift, I believe. They wrote "Ontario Library Research Cloud (SWIFT-S3 setup), locally managed with 5 nodes across Canada" under "indicate where Dataverse deployed" in the [2022 GDCC survey](https://groups.google.com/g/dataverse-community/c/gI7vJRNqrPw/m/Iyp47y3SAwAJ). They also ticked the box for using Swift:

Screen Shot 2023-01-25 at 8 38 42 AM

I'm not sure who would know for sure or who might be able to test this PR so I'll just spam some folks from Borealis (sorry! 😅 ): @amberleahey @lubitchv @JayanthyChengan @meghangoodchild

doc/sphinx-guides/source/api/native-api.rst
ErykKul: @landreev
Thank you for your review! I have done the remaining changes and updated the documentation. Can you have a look at the latest commits? Thanks.

landreev (approved):
I am going to go ahead and approve the PR. Thank you for all the work on this!
As for swift - thank you @pdurbin for spamming some potential testers of this functionality. 

I personally don't think the swift part of it should hold the PR, or be considered a blocker. So, if testing under swift becomes even remotely problematic, I'm just going to suggest that the cleanup implementation in the swift driver be commented out and replaced with a "not yet supported" exception, and a message to any potential swift user suggesting that they review the code and complete the task of adding the functionality.

PR #9137, opened 2022-11-04
Fix for direct download disabled

qqmyers:
**What this PR does / why we need it**: Restores direct download functionality. The refactoring to include globus included a misplaced paren (my bad) that made the check for direct download only occur in the Globus case.

Closes #

**Additional documentation**:


qqmyers:
bad PR

PR #9138, opened 2022-11-04
5.12 Fix - misplaced paren disabling direct download

qqmyers:
**What this PR does / why we need it**: As stated, a misplaced paren I introduced in the Globus PR caused direct download to only be done in the Globus case and disabled it for normal S3 stores.

Closes #

**Special notes for your reviewer**: This is only a paren relocation - the rest of the block is just moved one indent level out.

**Additional documentation**:


landreev (approved):
Nice!

PR #9152, opened 2022-11-08
detect NetCDF and HDF5 files based on content #9117

### LINKED ISSUE #9117: Improve file type detection of NetCDF and HDF5
According to Wikipedia [NetCDF](https://en.wikipedia.org/wiki/NetCDF) and [HDF5](https://en.wikipedia.org/wiki/Hierarchical_Data_Format) have magic numbers that should let us detect these file types more easily and reliably than guessing based on file extensions.

## NetCDF magic number

```
CDF\001
\211HDF\r\n\032\n
```

## HDF5 magic number

```
\211HDF\r\n\032\n
```

I brought this up at standup today and here are some notes from the discussion:

- We should see if JHOVE can detect them.
- Normally, detecting file types by seeking into files is part of detecting tabular files, specifically.
- Given that NetCDF can be big, this might be a case where switching to a ranged request to find the signature might be important.

We should add some NetCDF and HDF5 files to https://github.com/IQSS/dataverse-sample-data to test with, at some point.

Related:

- https://github.com/IQSS/dataverse.harvard.edu/issues/23

pdurbin:

Detection of NetCDF and HDF5 files is currently based on file extensions rather than content.

- Closes #9117

The Java library I'm adding ( https://github.com/Unidata/netcdf-java ) comes from the makers of NetCDF.

My thought was to introduce it here to simply detect NetCDF and HDF5 files and in subsequent pull requests we can make more use of it, such as this issue:

- #9153

Rename files to have no or non-standard file extensions to see if they are detected properly.

**Additional documentation**:


sekmiller (approved):

kcondon:
@pdurbin I've tested using example files from the net and works with extension but with diff extension, fails. Is there some variation in the formats where magic number test isn't valid? Any ideas? I tried examples from https://www.unidata.ucar.edu/software/netcdf/examples/files.html and https://people.sc.fsu.edu/~jburkardt/data/hdf/hdf.html . I can try more from different sources.

pdurbin:
@kcondon thanks. In my testing, I only discovered the following three "file ID types" (NETCDF, NetCDF-4, and HDF5) that were detected based on NetCDF and HDF5 files I found:

            switch (type) {
                case "NETCDF":
                    return "application/netcdf";
                case "NetCDF-4":
                    return "application/netcdf";
                case "HDF5":
                    return "application/x-hdf5";

I'd be happy to test with additional files. As usual, you've been more thorough than I have! Let's please compare notes (and files!) when we get a chance. I've checked a few small files into the branch/pull request for testing but most of them are too big for this.

pdurbin:
@kcondon as we discussed offline, I had a typo in the code such that "classic" NetCDF files (pre version 4) were not being detected correctly. I pushed a fix in 03188e72c9

I also tried to better document in the release note what people should expect. Here's what it says now:

"NetCDF and HDF5 files are now detected based on their content rather than just their file extension.

Both "classic" NetCDF 3 files and more modern NetCDF 4 files are detected based on content.

Detection for HDF4 files is only done through the file extension ".hdf", as before."

Issue #9164, opened 2022-11-11
Support digital signature/timestamps on dataset versions


qqmyers:
**Overview of the Feature Request**
Odum is interested in developing the ability to digitally sign a dataset version (i.e. during a pre-publication workflow) so that it is possible to prove that the dataset version has not changed since the signing. The current plan is to leverage a separate application for this, i.e. one that would use an institutional signing/timestamp authority but doing this would still require some modifications to Dataverse. The proposed changes at this point would be:

- [ ] Allow OAI-ORE/any export to be requested for any dataset version, including drafts (access to this last one would have to be authorized and would not be cached). This would allow the OAI-ORE to be used in the content being signed in a pre-pub workflow and for regeneration of the OAI-ORE file for earlier versions for use in validation efforts.
- [ ] Create a place, perhaps an column in the dataverse table for a signature to be stored (probably json), perhaps as a separate (aux?) file. When a signature exists, it would be added to the OAI-ORE export (and archival Bags) in an element clearly separate from the dataset contents (i.e. avoiding the catch 22 of signing something (a dataset description) that includes the signature itself)
- [ ] Explore use of the new (in review) signedURL capabilities to be able to provide a signer/validator app with only read-only access to the dataset contents.

While the overall development is in its early stages and details may change, the expectation is that these changes in Dataverse could/should be generally applicable for signing applications and perhaps valuable in other use cases as well.

**What kind of user is the feature intended for?**
Regulated environments where being able to detect malicious changes (in addition to accidental changes) is important.

If anyone else interested in this and/or working on something similar, please get in touch.

jggautier:
Just jumping in to say that the first part, asking exports for all versions of a deposit, is related to the request at https://github.com/IQSS/dataverse/issues/4372

qqmyers:
FWIW (heads-up to @matthew-a-dunlap ) - It looks like the ability to export the OAI-ORE metadata for a given version already exists via the `/api/datasets/<id>/versions/<version>/metadata` endpoint (as the 'data' element in the returned JSON response). An API key is required to get the response for the :draft version. 

As that is sufficient for this issue, I've checked the first box as complete. (The existing API does not allow export of any export format for any version which could be generally useful, but that isn't required for signing. If that's desired for other reasons, e.g. #4372, it could/should still be pursued as a separate issue.)

[...]

PR #9171, opened 2022-11-17
Moves JSON-LD Exporter Code from Domainmodel to corresponding Exporter

johannes-darms:

Moves code related to an Export Format from the domain model to the corresponing Exporter. In this case the JSON-LD generation code  is moved to the schemaDotOrgExporter. 

Closes #

**Special notes for your reviewer**: Does not add nor alter functionallity only moved code.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: No.

**Additional documentation**: Not needed.


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5525,6 +5525,13 @@ public boolean isThisLatestReleasedVersion() {
 
     }
 
+    /**
+     * JSON-LD is provided via the normal ExporterService. In favour for the more generic implementation this method is
+     * deprecated.
+     * @see  #getExporters()
+     * @return The Dateset encoded as JSON-LD
+     */
+    @Deprecated
```
qqmyers, line 5534: This method is used to add the schema.org metadata to the header of the dataset page. I don't see any place that you've replaced that functionality so this would need to not be deprecated then?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5534,7 +5541,7 @@ public String getJsonLd() {
                 return jsonLd;
             } else {
                 logger.fine("No cached schema.org JSON-LD available. Going to the database.");
-                String jsonLdProduced = workingVersion.getJsonLd();
+                String jsonLdProduced = instance.getExportAsString(workingVersion.getDataset(), SchemaDotOrgExporter.NAME);
```
qqmyers, line 5544: This is going to get the schema.org for the latest version rather than the one being viewed. I think what's needed is a method in ExportService that allows you to specify the version.
Also - the datasetVersion.getJsonLd() method used to cache the output in memory for reuse. The Exporter currently only writes the schema.org for the latest version to disk (which we may still want) and if we don't cache there, we may still want to store the jsonLd in memory. Perhaps keeping the DatasetVersion.getJsonLd() method (which would cache) but calling a new version specific method in the exporter would handle that?

src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
```diff
@@ -1772,270 +1763,17 @@ public String getPublicationDateAsString() {
...
-    // one metadata export in a given format per dataset (it uses the current 
-    // released (published) version. This JSON fragment is generated for a 
-    // specific released version - and we can have multiple released versions. 
-    // So something will need to be modified to accommodate this. -- L.A.  
     /**
      * We call the export format "Schema.org JSON-LD" and extensive Javadoc can
-     * be found in {@link SchemaDotOrgExporter}.
+     * be found in {@link edu.harvard.iq.dataverse.export.SchemaDotOrgExporter}.
      */
+    @Deprecated
```
qqmyers, line 1770: Based on my earlier comments  - if this gets used in producing the schema.org header, we still need it.

src/main/java/edu/harvard/iq/dataverse/export/SchemaDotOrgExporter.java
```diff
@@ -92,6 +102,266 @@ public void exportDataset(DatasetVersion version, JsonObject json, OutputStream
         }
     }
 
+
+    // The export subsystem assumes there is only
```
qqmyers, line 106: This is the problem I was talking about in earlier comments - ExportService doesn't handle per-version exporting and we need that for schema.org for the page header (and will want it for other reasons soon - see #9164). In any case - once that's fixed, this comment won't be true anymore.

src/main/java/edu/harvard/iq/dataverse/export/SchemaDotOrgExporter.java
```diff
@@ -92,6 +102,266 @@ public void exportDataset(DatasetVersion version, JsonObject json, OutputStream
...
+     */
+    public String getJsonLd(DatasetVersion version) {
+        String jsonLd;
+
+        JsonObjectBuilder job = Json.createObjectBuilder();
+        job.add("@context", "http://schema.org");
+        job.add("@type", "Dataset");
+        // Note that whenever you use "@id" you should also use "identifier" and vice versa.
+        job.add("@id", version.getDataset().getPersistentURL());
+        job.add("identifier", version.getDataset().getPersistentURL());
```
qqmyers, line 123: I'm guessing that the only changes in this block are changing this to version. If not, could you highlight any other changes you made?

src/test/java/edu/harvard/iq/dataverse/export/SchemaDotOrgExporterTest.java
```diff
@@ -116,17 +119,7 @@ public void testExportDataset() throws Exception {
...
-        if(version == null) logger.fine("ver null");
-        if(byteArrayOutputStream == null) logger.fine("bytarr null");
-        if(schemaDotOrgExporter == null) logger.fine("sdoe" + " null");
-        try {
-        schemaDotOrgExporter.exportDataset(version, json1, byteArrayOutputStream);
-        } catch (Exception e) {
-            e.printStackTrace();
-        }
-        String jsonLd = byteArrayOutputStream.toString();
+        String jsonLd = generateeExportAsString(version);
```
qqmyers, line 122: Typos here and below with an extra small e - should be generateExportAsString ?

qqmyers (changes requested):
Mostly a simple refactor but I think the fact that the jsonLd is used in the dataset header (see https://github.com/IQSS/dataverse/blob/26afd103152c91238d2a730da778cf11e2d44a66/src/main/webapp/dataset.xhtml#L72) was missed. That is supposed to be version specific, which the original code is, but the refactoring runs everything through ExportService which always picks the latest version. I'd suggest adding a version specific call to ExportService (probably still only writing to disk for the latest version) either as a separate PR that goes in before this or as part of this PR. (IQSS usually prefers smaller PRs.)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
qqmyers: FWIW: https://github.com/IQSS/dataverse/blob/803a6e94176a8114a96aa06e8c7369aab658731f/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L5529 means the schema.org is only in the page header for the latest version so nominally we don't currently need to keep this method version specific for that reason. That said, other work will need it so removing it now only to have to bring it back would just add work.

pdurbin:
@johannes-darms are you still interested in this? This refactoring hasn't been a priority for us. We have 107 open pull requests and I'm looking for some we can close. Maybe a "soft close" since we could always open it again if it becomes a priority. Please let us know. Thanks.

scolapasta:

scolapasta:
FYI, once we've worked through the existing PRs and we get that # to be more manageable, do let us know if you want to revisit this. Thanks!

PR #9056, opened 2022-10-12
8903 email subject

JayanthyChengan:
**What this PR does / why we need it**:
Some of the notification emails received from the Dataverse system include the dataset name in the subject (e.g., dataset created), whereas others do not (e.g., dataset published).

So, including the dataset name for those missing in the subject line.

**Which issue(s) this PR closes**:
Closes #8903

PR #9173, opened 2022-11-17
QDR/9172-fix messages

### LINKED ISSUE #9172: Some messages have empty "" (bug in #9056)
#9056 changed several messages to include the dataset diplayname in quotes in the message. However the PR code assumes that the object being sent to the message is the dataset and the displayname is left blank when the object is the datasetversion. (FWIW: The bug was introduced during review in trying to avoid a null pointer for messages that don't use either of these object types.)

* What happens?
When messages are sent - they say things like `Qualitative Data Repository: Dataset "" Status Change`
5.12.1

PR to follow

qqmyers:
**What this PR does / why we need it**: #9056 added the dataset displayname to some messages but a bug/partial fix introduced in review led to some messages display an empty `""`. This PR fixes that.

- Closes #9172 

**Suggestions on how to test this**:  Do what's needed to generate messages of any/all types below and verify that the message includes the dataset displayname. If one works, they all should.

```
CREATEDS:
SUBMITTEDDS:
PUBLISHEDDS:
PUBLISHFAILED_PIDREG:
RETURNEDDS:
WORKFLOW_SUCCESS:
WORKFLOW_FAILURE:
STATUSUPDATED:
```

Regression testing: The message to users when a dataset is created should also show the dataset display name. No other messages (such as REQUESTFILEACCESS) should be broken.

**Is there a release notes update needed for this change?**: no? relatively minor message display bug

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/MailUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/MailUtil.java
@@ -34,8 +34,12 @@ public class MailUtil {
         List<String> rootDvNameAsList = Arrays.asList(BrandingUtil.getInstallationBrandName());
         String datasetDisplayName = "";
 
-        if (objectOfNotification != null && (objectOfNotification instanceof Dataset)  ) {
-            datasetDisplayName = ((Dataset)objectOfNotification).getDisplayName();
+        if (objectOfNotification != null) {
+            if (objectOfNotification instanceof Dataset) {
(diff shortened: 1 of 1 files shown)


mreekie:
Raised by Gary.
Definitely required for the next release.
When we send email messages, there is a text string identifying the dataset id that is now blank.
Simple PR 
Sized at a 3

mreekie:
added to sprint Dec 15, 2022

scolapasta (approved):

Issue #7050, opened 2020-07-02
Beyond Dataverse 5.0: make it pluggable (on a code level)


poikilotherm:
**tl;dr**: we should discuss how to make Dataverse more pluggable and open to third-party extensions.

While digging through the causes of why our WAR file is so big (will be ~135MB for Dataverse 5.0, was ~205MB for 4.20), I stumbled over a lots of things that could use a hand. This is mostly about refactoring old code, libraries and dependencies. Great - I like that :pick: . (See an older list to be updated at #5360)

But what is likely not to be shrinkable: Apache Tika. It's used for the great full text indexing component, provided by @qqmyers (thank you! :+1: ). *But*: that's a 45MB increase in WAR size for a feature which is completely optional. I have no figures how many installations actually enabled and use it. For "us" (Jülich DATA) it makes images bigger and adds to deployment time :neutral_face: .

In many cases, new and great features are accepted by IQSS and merged (yeah! :tada: ), also they might not be using it for Harvard. But is this a good approach? Maintenance effort is put onto people that focus on other features. Testing is necessary and I'm not the only one around that has some grey hair from the status quo :face_with_head_bandage:  (looking at you @4tikhonov, @skasberger, @kcondon, @donsizemore, @pdurbin and others).

Back in the day of #4106 people started to fork just to add support for some functionality that was rejected (for good) by IQSS. But it's a pity you're forced into forking, which is always a big tradeoff. :balance_scale: 

Now we have great new working groups! :heart: Metadata! :heart: More of that! Yes please! :+1:  And it's likely that there's more ingest stuff coming down the road. Lot's of new features, which don't fit under "External Tool" or "Integration" like the shiny previewers, but code that needs access on a "Java-level".

Those "extensions" or "plugins" should be easy to install for admins (no compiling and fiddling with Maven like DSpace) and developed independently from IQSS, offloading maintenance, testing and development. Ideally people would start to share their plugins. Or even start selling them. Lots of new options.

There a few ways to do this :brain: . One is using small frameworks like https://pf4j.org. There are more and others.
I am eager for more input. What do y'all think? Especially: what does our beloved architect @scolapasta think about this? :bow: 

This should start small and where we all see fit. It can grow as we go, trying to find the sweet spot of best community support vs. refactoring burden. Is there a chance to find some funding to enable this next generation repository technology? (Yeah, I know this is nothing new from a technical perspective, but it seems to be from a community perspective.)

stevenmce:
[...]

Issue #8512, opened 2022-03-19
Feature Request/Idea: Standardize standard license configuration


philippconzett:
**Overview of the Feature Request**
With version 5.10, the long-awaited multiple-license support was released (see [release notes](https://github.com/IQSS/dataverse/releases/tag/v5.10)). Thanks to all contributors!
To better support interoperability between Dataverse installations and beyond Dataverse installations, I'd like to suggest standardizing the way standard license configuration is managed using multiple-license support as follows:

1. Standardized licenses are provided as authoritative JSON files stored in the IQSS/dataverse GitHub repositories or the GDCC GitHub repository.
2. The section [Adding Licenses](https://guides.dataverse.org/en/latest/installation/config.html#adding-licenses) in the Dataverse Installation Guide links to GitHub folder containing the JSON files.
3. We agree on the source and content of the elements in the JSON files. Here are some suggestions and possible issues to discuss:
- **Source:** Whenever possible, we use the information provided in the [SPDX License List](https://spdx.org/licenses/) and the license webpage provided by the license issuer.
JSON elements:
- **name:** Field _Identifier_ in the SPDX License List, but without hyphens, e.g., Artistic-2.0 > Artistic 2.0
- **uri:** 1) License URI provided by the license issuer; 2) if (1) is not available, SPDX URI for the license
- **shortDescription:** Field _Full name_ in the SPDX License List. **Question 1:** Do we need to add "License" or "Dedication", as is currently done in the JSON files provided in the Dataverse Installation Guide? **Question 2:** Do we need to add a full stop at the end of the shortDescription element, as is currently done in the JSON files provided in the Dataverse Installation Guide?
- **iconUrl:** As provided by license issuer
4. We agree on which standard licenses to provide as JSON files in the GitHub repository. To start with, I suggest we concentrate on the following ones:
- Creative Commons Zero 1.0
- All Creative Commons Attribution (BY) licenses 4.0 and later
- Open Data Commons Open Database License v1.0
- Open Data Commons Attribution License v1.0
- All licenses in the SPDX License List that are FSF Free/Libre and OSI Approved, starting with licenses included in Open Science Framework (OSF):

Content:
- CC0 1.0 Universal
- CC-BY Attribution 4.0 International

Code - Permissive:
- MIT License
- Apache License 2.0
- BSD 2-Clause "Simplified" License
- BSD 3-Clause "New"/"Revises" License

Code - Copyleft:
- GNU General Public License (GPL) 3.0
[...]

Issue #7077, opened 2020-07-14
Support flexible DataCite resourceType metadata in a Dataset


poikilotherm:
**tl;dr:** Dataverse should offer a UI component to select the [dataset type with a CV based on DataCite/Dublin Core](https://support.datacite.org/docs/schema-mandatory-properties-v42#101-resourcetypegeneral). The selected type should be used for metadata registration at DataCite.

**Related issues**: 
- #2917 (this new issue is only about a part of what is mentioned there) 
- #5086 (which is about doing the same but for file level DOIs)

Jülich DATA has an open request by @sciapp to publish software in our repo and get a DOI for releases (so inline with [FORCE11 recommendations](https://doi.org/10.7717/peerj-cs.86) and [DataCite recommendations](https://support.datacite.org/docs/schema-software-citation-principles-mapping-v41)).

Recently, the RDA WG has [published a paper open for community comments](https://docs.google.com/document/d/1MpWGgxet1A0qFhPFJoIs0363wXOUKgzwQIinKc8QqWI) about the different PID options for software publications. The interesting part for Dataverse is the [diagram for registered software datasets at Datacite](https://docs.google.com/document/d/1MpWGgxet1A0qFhPFJoIs0363wXOUKgzwQIinKc8QqWI/edit#heading=h.w1jjaclppedf).

Currently, software that gets published via Dataverse, will have a `resourceTypeGeneral="DataSet"` attached to it, as [the metadata template](https://github.com/IQSS/dataverse/blob/0aaac2546abba56acb20ed3d755063acd7ec6a6a/src/main/resources/edu/harvard/iq/dataverse/datacite_metadata_template.xml#L12) does not allow for customization. (This is also true for #5086, where we might think about using type "Collection" for the dataset automatically and specific types for the files.)

Having software counted as "DataSet" makes things less discoverable and does not push research software engineering forward. [See DataCite Schema Docs for a complete list of types.](https://support.datacite.org/docs/schema-mandatory-properties-v42#101-resourcetypegeneral)

A [full example of software metadata](https://support.datacite.org/docs/schema-metadata-examples-v42#example-for-software) can be found at DataCite.

Things to do for an implementation: 

1. Obviously, this will need a new metadata field in `citation.tsv`, using a controlled vocabulary based on `resourceTypeGeneral`. Should be mandatory, but happy to discuss. This will definitely have to be present `onCreate`, but could use "DataSet" as a default!
[...]

PR #9175, opened 2022-11-18
DANS - Exporters in external jars

qqmyers:
**What this PR does / why we need it**: DANS is interested in being able to add exporters and to replace the existing ones with customized/improved versions. This PR extends the service provider interface code in the ExporterService to dynamically load classes from Jar files places outside the exploded war file.  It is therefore a proof-of-concept for this approach.

Related to #7050

The overall design was discussed in Tech Hour. Aside from the ability to dynamically use external exporter jars, the main change is to the Exporter interface and the introduction of a ExportDataProviderInterface that remove dependencies Exporters had on internal Dataverse classes (DatasetVersion, FileMetadata, DataFile, etc.) This was done by providing Exporters with the standard JSON/ORE/Schema.org/DataCite XML export contents along with a new JSON representation of file variable level metadata. All of the existing Exporters were rewritten to use these JSON/JSON-LD inputs. (Nominally the JSON and ORE exports are 'complete' and are usable for new exporters. The DataCite and Schema exporters were and still are just wrappers of content created deeper in the Dataverse code (and reused there for other purposes.))

Another ~minor change was to add the Locale to the Exporter.getDisplayName() method. (Exporters originally retrieved the Locale using the BundleUtil class which calls PrimeFaces FacesContext.) This change avoids these dependencies (but does not allow external Exporters to use other BundleUtil methods and thereby share Dataverse's existing i18n properties files - that could be done later if desired.)

With those changes in place, I went ahead to create maven modules for an exporters SPI jar file that includes the two interfaces above and the ExportException class. That jar is now used in the main code and in an example  'MyJSON' exporter (that does nothing but wrapper the original JSON export with 
```
{
    "name": "dataverse_json",
    "inputJson": <original export>
}
```

Having schema in place for the JSON/ORE/Variable-level metadata or creating new types of transfer objects would be useful going forward. Presumably their design/devolpment will be driven by the SPA work and changes it requires. In the meantime, this PR should make it significantly easier to develop/deploy exporters.

The one aspect of this work I'm least sure of is the best way to handle the maven modules. Right now, the jar has a 0.0.1 version, the interface jar and the example MyJSON exporter are both under the ./modules dir, the code installs the new jar as a local resource rather than publishing it, etc. All changes welcome if there are ideas of how to version, where code should go in the repo, if the interface jar and/or example should be separate repos, etc. If someone knows how to do that and we have the appropriate credentials to do so, publishing the jar would be a better long-term option. Until then, doing mvn clean package for the ./modules/exportSPI module and then building the main war as usual works. (The auto-build currently fails due to missing this step/the interface jar only being local.)

There are three things to test:

- Regression testing - that existing exporters work as before - much of which is covered by tests already
- That dropping a new exporter in place works with two subcases - adding a new one and replacing an existing one - this requires defining the dataverse.exporters.directory mpconfig/jvm option, putting the new jar there, restarting the server, clearing the cached exports (for the replace case), and verifying that the new exporter appears in the list of options, the export file can be downloaded, etc.
- Building a new exporter works - this is basically demonstrated by the 'MyJSON' example above - just running the maven build for it proves this.

Only by replacing/adding new exporters

Yes - for this and other documentation, I'd like to get through an initial review that might change naming, locations, build instructions, etc. before nailing things down. 
**Additional documentation**:

Older notes:

The PR does raise issues though. The ServiceLoader mechanism looks for files "META-INF/services/edu.harvard.iq.dataverse.export.spi.Exporter" that list the available instances of Exporter that should be loaded. Thus, to successfully load a class, it has to be in one of the META-INF list files and it has to be on the classpath, not necessarily in the same Jar file as the META-INF file in which it is listed. In considering where to put the existing exporter class files - if they stay in the war but the META-INF is moved to a separate Jar, the could be overridden (can't use the same class name but another class could provide the same export (defined by the displayname and provider name). Conversely if the META-INF file is in the war, ServiceLoader doesn't allow skipping it (as far as I can tell), so it would not be possible to override the internal classes. That said, even the current exporters could move to an external Jar (we might need to add some error handling if people removed the exporters that do double duty (DataCite, Schema.org, OAI_ORE, ...). 

I've also explored not using the ServiceLoader - the main difference with not using it would be that the META-INF file wouldn't be needed - we could check all files in the jar to see if they are Exporters or provide a different lookup mechanism, i.e. some setting/microprofile config list.

Since these are techno-policy issues, I think it would be useful to discuss the best approaches in a Tech Hour (and through discussion in this PR, slack, etc.) before I remove 'Draft' and send something like this forward.

Also note - ~this should be usable for ingest, archivers, full-text search, etc., so it would be nice if we had a common pattern across all of those.

Also also note - the PR doesn't currently remove any Exporter class files from the war, etc. I've been testing by manually creating Jars and removing the META-INF file and/or the classes from the exported war to show that it works. Who/how/where Exporters get built/reviewed/tested/packaged is also a good topic.


poikilotherm:
Thank you @qqmyers for your extensive work on this! I'm glad there is a chance it will get picked up so someday we have #7050 resolved.

As I wrote on Slack, it seems to be totally fine not to use a convenient wrapper as PF4J, but rely on standard SPI mechanics like we see them in action for MicroProfile Config API or Jakarta EE. I also completely agree we need to discuss these topics at Tech Hour.

Here's a thought: IMHO we (whoever that may be) shall create a proper "Dataverse Java API" package, preferably as a module under `/modules`. The content should be inspired from the mentioned frameworks:

- Interfaces to provide implementations from a package: `ExporterProvider`, `MinterProvider`, and so on (similar to [ConfigSourceProvider](https://download.eclipse.org/microprofile/microprofile-config-3.0/apidocs/org/eclipse/microprofile/config/spi/ConfigSourceProvider.html))
- Interfaces expressing the contract to be used in the main code: `Exporter`, `Minter`, and so on (similar to [ConfigSource](https://download.eclipse.org/microprofile/microprofile-config-3.0/apidocs/org/eclipse/microprofile/config/spi/ConfigSource.html))
- Data Transfer Objects (DTOs) to abstract our data model `Entity` classes and not expose these internal implementations. Something like [modelmapper](https://github.com/modelmapper/modelmapper) might be handy to do the conversion. Maybe moving our codebase to Java 17 is a win here, enabling use of records for this.

It might be fine to do develop this API package iteratively, adding more as we go.

IMHO this technical discussion is very much related two other topics:
1. The UI reworks in process. Providing DTOs would help in abstracting away our internal data model, which simplifies REST API stabilization. Also, using the class loading approach, it should be possible to add more JAX-RS `Resources` and `Providers` to extend the REST API from an external JAR. Shipping extensible UI functionality, bundling a React addon plus the backing API classes, would be in reach.
2. DataCite metadata issues like #7077, #8512 and many others, as we need a completely renewed exporter/registrar for this. As this is crucial to support things like software publications, I can add funded developer time from me to do enabling groundwork. (Will add a HERMES label to this PR to express interest and collab options).

johannes-darms:
1.) Is is possible to delegate the classloading to the application server instead of writing our own implementation?  If this is not possible I would suggest to move the generic part to a dedicated class as is is likely used by more modules that load custom implementation at runtime (e.g. PID providers). 

2.) Do we plan to add libraries to ease and automate the generation of mainifest files? Something like [google auto](https://github.com/google/auto/tree/main/service)?

qqmyers:
Just updated to prioritize external exporters that register with the same 'providerName', allowing replacement of the internally provider export format(s). This was ~per Tech Hour Discussion and uses the standard ServiceLoader to load classes combined with a Map that is only populated with the 'winning' Exporters.

@johannes-darms - we are using @AutoService already.

FWIW: I'm planning to meet with @poikilotherm next week to think through how to make a separate SPI API jar/artifact that would be all an external Exporter project would have to depend on. Hoping to do that generically so other SPI interfaces can be added as we open those up.

poikilotherm:
Hi there!

I created https://github.com/GlobalDataverseCommunityConsortium/dataverse/pull/35 to fix some very crucial things w.r.t Maven modules. Will leave a review later.

With regards to a release cycle, we need to test a few things - I simply don't know if we can release a package to Maven Central **Snapshots** that doesn't carry "-SNAPSHOT" in the version string.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -276,82 +316,79 @@ public void exportFormat(Dataset dataset, String formatName) throws ExportExcept
...
-    	} catch (IOException ioex) {
-    		//This catches any problem creating a local temp file in the catch clause above
-    		throw new ExportException("IO Exception thrown before exporting as " + "export_" + format + ".cached");
-    	} finally {
-    		IOUtils.closeQuietly(outputStream);
-    	}
+    private void cacheExport(Dataset dataset, ExportDataProviderInterface dataProvider, String format, Exporter exporter)
+            throws ExportException {
+        boolean tempFileUsed = false;
+        File tempFile = null;
```
poikilotherm, line 335: This is not C. Please move to correct (local) places.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -276,82 +316,79 @@ public void exportFormat(Dataset dataset, String formatName) throws ExportExcept
...
+        OutputStream outputStream = null;
+
+        StorageIO<Dataset> storageIO = null;
```
poikilotherm, line 338: Should be inside the first try as well...

src/main/java/edu/harvard/iq/dataverse/export/SchemaDotOrgExporter.java
```diff
@@ -75,20 +73,17 @@ public class SchemaDotOrgExporter implements Exporter {
...
+                try {
+            outputStream.write(dataProvider.getDatasetSchemaDotOrg().toString().getBytes("UTF8"));
+        } catch (IOException ex) {
+            logger.info("IOException calling outputStream.write: " + ex);
+        }
+        try {
+            outputStream.flush();
+        } catch (IOException ex) {
+            logger.info("IOException calling outputStream.flush: " + ex);
         }
```
poikilotherm, line 87: Indentation issues.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -105,6 +105,10 @@ public enum JvmSettings {
     SCOPE_PID_HANDLENET_KEY(SCOPE_PID_HANDLENET, "key"),
     HANDLENET_KEY_PATH(SCOPE_PID_HANDLENET_KEY, "path", "dataverse.handlenet.admcredfile"),
     HANDLENET_KEY_PASSPHRASE(SCOPE_PID_HANDLENET_KEY, "passphrase", "dataverse.handlenet.admprivphrase"),
+
+    // EXPORT SETTINGS
+    SCOPE_EXPORTERS(PREFIX, "exporters"),
```
poikilotherm, line 110: May I suggest we add the scope "exporters" to a another scope "plugins"? The name now seems a bit arbitrary. Is it a target dir where exports are written? Adding "plugins" (or sth similar) make the intention clearer and will allow us to add more switches (like enable/disable, lists, options, ...)

src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -54,3 +54,6 @@ dataverse.pid.datacite.rest-api-url=https://api.test.datacite.org
 
 # Handle.Net
 dataverse.pid.handlenet.index=300
+
+# Directory to load external exporters from
+dataverse.exporters.directory=/tmp/exporters
```
poikilotherm, line 59: Not sure I like this. How about `${com.sun.aas.instanceRoot}/plugins/exporters`?

Also, see my remark on `ExportService` above about how this might log strange warning when absolutely not necessary.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +57,50 @@ public class ExportService {
...
+
+    private static final Logger logger = Logger.getLogger(ExportService.class.getCanonicalName());
 
     private ExportService() {
-        loader = ServiceLoader.load(Exporter.class);
+        /*
+         * Step 1 - find the EXPORTERS dir and add all jar files there to a class loader
+         */
+        List<URL> jarUrls = new ArrayList<URL>();
+        Path exporterDir = Paths.get(JvmSettings.EXPORTERS_DIRECTORY.lookup());
```
poikilotherm, line 69: Just a thought: maybe make this lookup optional and skip if not present. Someone not using the plugin function might not want to see warnings about no directory for exporters.

If we still want a sane default location in `microprofile-config.properties`, maybe add an enabling config property?

poikilotherm (changes requested):
Overall, the concept and approach seem sound to me. This is a fine piece and a great first step towards plugability in Dataverse! Much appreciated!

I do have a lot of detail remarks.

As these files are not present, I'll add this here: any new module carrying Java code should follow good practices and add the `package-info.java`.

src/main/java/edu/harvard/iq/dataverse/export/DCTermsExporter.java
```diff
@@ -25,14 +28,15 @@ public String getProviderName() {
     }
 
     @Override
-    public String getDisplayName() {
-        return  BundleUtil.getStringFromBundle("dataset.exportBtn.itemLabel.dublinCore") != null ? BundleUtil.getStringFromBundle("dataset.exportBtn.itemLabel.dublinCore") : "Dublin Core (DCTERMS)";
+    public String getDisplayName(Locale locale) {
+        String displayName = BundleUtil.getStringFromBundle("dataset.exportBtn.itemLabel.dublinCore", locale);
+        return  displayName != null ? displayName : "Dublin Core (DCTERMS)";
```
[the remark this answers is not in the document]
qqmyers: I'd like to leave general changes to BundleUtil out if this PR

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +57,50 @@ public class ExportService {
 
     private static ExportService service;
     private ServiceLoader<Exporter> loader;
+    private Map<String, Exporter> exporterMap = new HashMap<String, Exporter>();
+
+    private static final Logger logger = Logger.getLogger(ExportService.class.getCanonicalName());
 
     private ExportService() {
```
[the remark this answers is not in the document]
qqmyers: I don't know that hot swap is a requirement at the moment, so I'd prefer to leave this out of this PR.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +57,50 @@ public class ExportService {
...
+         * Step 3 - Fill exporterMap with providerName as the key, allow external
+         * exporters to replace internal ones for the same providerName. FWIW: From the
+         * logging it appears that ServiceLoader returns classes in ~ alphabetical order
+         * rather than by class loader, so internal classes handling a given
+         * providerName may be processed before or after external ones.
+         */
+        loader.forEach(exp -> {
+            String formatName = exp.getProviderName();
+            // If no entry for this providerName yet or if it is an external exporter
+            if (!exporterMap.containsKey(formatName) || exp.getClass().getClassLoader().equals(cl)) {
```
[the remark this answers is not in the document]
qqmyers: Not sure where you would have an isExternal method, but this is a one time operation so hopefully not something that is worth changing now?

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -276,82 +316,79 @@ public void exportFormat(Dataset dataset, String formatName) throws ExportExcept
...
-    		//This catches any problem creating a local temp file in the catch clause above
-    		throw new ExportException("IO Exception thrown before exporting as " + "export_" + format + ".cached");
-    	} finally {
-    		IOUtils.closeQuietly(outputStream);
-    	}
+    private void cacheExport(Dataset dataset, ExportDataProviderInterface dataProvider, String format, Exporter exporter)
+            throws ExportException {
+        boolean tempFileUsed = false;
+        File tempFile = null;
+        OutputStream outputStream = null;
```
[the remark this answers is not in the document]
qqmyers: Not sure I want to tackle this now (it is existing code - just reformatted). The way the code is now, we try creating the outputStream and, if that fails, we create it a different way, after which it is used regardless of how it was created. I don't immediately see how to restructure to use try with resources in that case.

modules/dataverse-exporter-spi/src/main/java/io/gdcc/export/spi/Exporter.java
```diff
@@ -43,8 +46,6 @@ but NOT close() it!
     public String getXMLSchemaLocation() throws ExportException; 
     public String getXMLSchemaVersion() throws ExportException; 
     
-    public void setParam(String name, Object value);
-
 	public default String getMediaType() {
 	    return MediaType.APPLICATION_XML;
```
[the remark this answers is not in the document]
qqmyers: FWIW: I added an XMLExporter extends Exporter interface that gets rid of the three XML specific methods in the Exporter interface (which means that non-XML exporters don't have to implement and throw exceptions for them) and allows only XMLExporter to have a default value. It doesn't get rid of the MediaType dependency, but it might be good to have that so new exporters can use it.

pom.xml
```diff
@@ -739,6 +743,29 @@
...
+                <executions>
+                    <execution>
+                        <id>install-jar-lib</id>
+                        <goals>
+                            <goal>install-file</goal>
+                        </goals>
+                        <phase>validate</phase>
+                    </execution>
+                </executions>
+            </plugin>
```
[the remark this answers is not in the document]
qqmyers: Will remove when done building locally

src/main/java/edu/harvard/iq/dataverse/export/DCTermsExporter.java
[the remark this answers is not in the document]
poikilotherm: Misunderstanding, sorry 'bout that. I was talking about this little trick below. Again, this is a style question and what one prefers to read as code.

```suggestion
        return Optional.ofNullable(BundleUtil.getStringFromBundle("dataset.exportBtn.itemLabel.dublinCore", locale))
            .orElse("Dublin Core (DCTERMS)");
```

src/main/java/edu/harvard/iq/dataverse/export/DDIExporter.java
```diff
@@ -34,17 +38,18 @@ public String getProviderName() {
...
+        return  displayName != null ? displayName : "DDI";
     }
 
     @Override
-    public void exportDataset(DatasetVersion version, JsonObject json, OutputStream outputStream) throws ExportException {
+    public void exportDataset(ExportDataProviderInterface dataProvider, OutputStream outputStream) throws ExportException {
         try {
         XMLStreamWriter xmlw = XMLOutputFactory.newInstance().createXMLStreamWriter(outputStream);
         xmlw.writeStartDocument();
         xmlw.flush();
```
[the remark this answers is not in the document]
qqmyers: I don't know. I'd rather not change in this PR if there is some reason.

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/XMLExporter.java
poikilotherm, line 1: This is a very nice idea! I love it!

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +57,50 @@ public class ExportService {
...
+        List<URL> jarUrls = new ArrayList<URL>();
+        Path exporterDir = Paths.get(JvmSettings.EXPORTERS_DIRECTORY.lookup());
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(exporterDir)) {
+            for (Path path : stream) {
+                if (!Files.isDirectory(path)) {
+                    logger.fine("Adding  " + path.toUri().toURL());
+                    //This is the syntax required to indicate a jar file from which classes should be loaded (versus a class file). 
+                    jarUrls.add(new URL("jar:" + path.toUri().toURL() + "!/"));
+                }
+            }
```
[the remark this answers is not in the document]
qqmyers: FWIW: I just cut/pasted this as I'd already added the Optional<> change above so that was easier than having to merge the conflict.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +58,54 @@ public class ExportService {
...
+
+    private static final Logger logger = Logger.getLogger(ExportService.class.getCanonicalName());
 
     private ExportService() {
-        loader = ServiceLoader.load(Exporter.class);
+        /*
+         * Step 1 - find the EXPORTERS dir and add all jar files there to a class loader
+         */
+        List<URL> jarUrls = new ArrayList<>();
+        Optional<String> exportPathSetting = JvmSettings.EXPORTERS_DIRECTORY.lookupOptional(String.class);
```
poikilotherm, line 70: Just a thought: you might be able to take advantage of auto-conversion via `JvmSettings.EXPORTERS_DIRECTORY.lookupOptional(Path.class)`, as it has a method `Path.of(String)`. (But I would need to try that myself.) You definitely can skip the `String.class`.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -51,9 +57,50 @@ public class ExportService {
...
+        loader = ServiceLoader.load(Exporter.class, cl);
+        /*
+         * Step 3 - Fill exporterMap with providerName as the key, allow external
+         * exporters to replace internal ones for the same providerName. FWIW: From the
+         * logging it appears that ServiceLoader returns classes in ~ alphabetical order
+         * rather than by class loader, so internal classes handling a given
+         * providerName may be processed before or after external ones.
+         */
+        loader.forEach(exp -> {
+            String formatName = exp.getProviderName();
```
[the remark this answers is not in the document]
qqmyers: They look like ~synonyms. I went ahead and used format name everywhere. So far I didn't add the existing ones as constants in the interface. I could, but they are not all consistent in capitalization and there are two with the same format (one is harvest only and not in the export list and the other is vice versa). I'm not sure that cleaning that up is in scope and not sure I know all the dependencies, i.e. what has to change in the harvesting code if a format name changes.

qqmyers:
@poikilotherm - Thanks for the feedback. I think I've addressed most things aside from docs/JavaDocs at this point. Some of what you brought up was just from existing code that I didn't change (or just reformatted/moved) but I tried to address anything straight forward there too. I left a few comments about things I think should go into other PRs if/when those are priorities. I haven't gone back to retest yet but will. (Also plan to move the example out to a new repo but I did the refactoring in here.)

qqmyers:
@poikilotherm - javadocs/docs/release note added, example moved to https://github.com/gdcc/dataverse-exporters, and the local maven install of the jar has been removed. I think all that's needed at this point is whatever change is needed to push the maven artifact (is that something you can set up/tell me how to set up?) and a final review of the documentation. Feel free to edit directly or request more changes. Thanks again.

qqmyers:
@janvanmansum, @PaulBoon - any feedback from DANS?

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/Exporter.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Exporters that return an XML-based format should implement the XMLExporter
+     * interface.
+     * 
+     * @apiNote XML-based exporters must implement additional methods. Returning
+     *          true by overriding this method without implementing the XMLExporter
+     *          interface will result in failure/exceptions being thrown when your
+     *          exporter is called.
+     */
+    default Boolean isXMLFormat() {
```
johannes-darms, line 97: can we push this method down to XMLExporter interface?

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/Exporter.java
```diff
@@ -0,0 +1,122 @@
...
+     * to flush() the outputStream, but NOT close() it! This way an exporter can be
+     * used to insert the produced metadata into the body of an HTTP response, etc.
+     * (for example, to insert it into the body of an OAI response, where more XML
+     * needs to be written, for the outer OAI-PMH record). -- L.A. 4.5
+     * 
+     * @param dataProvider - the @see ExportDataProvider interface includes several methods that can be used to retrieve the dataset metadata in different formats. An Exporter should use one or more of these to obtain the values needed to generate metadata in the format it supports. 
+     * @param outputStream - the OutputStream to write the metadata to
+     * @throws ExportException - if there is an error writing the metadata
+     */
+    void exportDataset(ExportDataProvider dataProvider, OutputStream outputStream) throws ExportException;
```
johannes-darms, line 29: I would not pass an `exportDataProvider` as a parameter to `exportDataset`. First, since the name of the interface and method is `exportDataset`, I assume that I can pass a DataSet to this method to get an external representation. Secondly, to create an export, an implementer may need access to the real typed object and all associated properties. These cannot be contained in a single ExportDataProvider response. Third, ExportDataProvider responses are either JSON or XML objects, and an implementer must parse these even if a typed object tree is available. This leads to unnecessary complexity. Fourth, the responses of `ExportDataProvider` are already results of exports that can potentially be overridden. Therefore, I would change the definition to `void exportDataset(DataSet dataset, OutputStream outputStream) throws ExportException;`

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/XMLExporter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @implNote for the ddi exporter, this method returns
+     *           "https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd"
+     * @return - the location of the XML schema
+     */
+    String getXMLSchemaLocation();
+
+    /**
+     * @implNote for the ddi exporter, this method returns "2.5"
```
johannes-darms, line 32: Isn't that an URI according to the XML Spec?

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -201,157 +252,168 @@ public void exportAllFormats(Dataset dataset) throws ExportException {
...
-            while (exporters.hasNext()) {
-                Exporter e = exporters.next();
-                if (e.getProviderName().equals(formatName)) {
-                    DatasetVersion releasedVersion = dataset.getReleasedVersion();
-                    if (releasedVersion == null) {
-                        throw new ExportException("No published version found during export. " + dataset.getGlobalId().toString());
+
+            Exporter e = exporterMap.get(formatName);
+            if (e != null) {
+                DatasetVersion releasedVersion = dataset.getReleasedVersion();
```
johannes-darms, line 309: This pins the export to `getReleasedVersion` would it make sense to just export the passed Dateset and move the responsible to select the latest/released version to the caller?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
```
johannes-darms, line 2: Default header

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -0,0 +1,98 @@
...
+import io.gdcc.spi.export.ExportDataProvider;
+import edu.harvard.iq.dataverse.util.bagit.OREMap;
+import edu.harvard.iq.dataverse.util.json.JsonPrinter;
+import edu.harvard.iq.dataverse.util.json.JsonUtil;
+
+/**
+ * Provides all data necessary to create an export
+ * 
+ */
+public class InternalExportDataProvider implements ExportDataProvider {
```
johannes-darms, line 31: Is this interface and implementation really needed? Why are there `internal` export Formats and why are those handled different than the others? IMHO: The same result can be achieved without this interface/implementation. Sure this approach protects the internal `DataSet` object from access by an Exporter implementation, but is this needed?  IMHO, this just makes development of new exports more complex to implement (or forces me to use reflection to get the objects ;))

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/Exporter.java
johannes-darms: This will expose the internal structure to an Exporter. IMHO an Exporter is a piece of code that transfers an internal representation to an external one, subsequently it needs access to the internal structure.

poikilotherm: Thank you @johannes-darms for the feedback. Let me add some more context you can't possibly know here, that hopefully makes the design choices easier to follow.

1. This work has been funded by DANS, and funding is limited. The changes you are asking for involve the creation of data transfer objects and code translating between the internal database models (`DvObject, Dataset, DataFile, Dataverse`) and those. This task can't be covered on current funding.
2. @qqmyers is aware this is not the technically best solution and kind of introduces technical debt. But it is a starting point to something very important, that we have been longing for for a long time: real modularity with loadable JARs and externalized Java code. I do agree this solution here is limited and far from perfect. "Aber besser den Spatz in der Hand, als die Taube auf dem Dach."

Please keep on nagging to go for the best. Funding and future PRs improving this are very welcome (but we'd need to coordinate such activities).

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
poikilotherm: Also it might look new code, this is the behaviour as it has been for a long time. Not meaning to say this is good, but changing this is beyond scope for this PR. (Needs new use case, issue, PR, funding, ...)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
poikilotherm: Yes this is needed. Let me explain: the internal model is using lots of database annotations. If we directly move these classes into the `dataverse-spi` package, that means everyone using it will inherit the dependencies. This is not how you design an API like this: you want to decouple internal and external code.

Usually one would use data transfer objects for this that are shipped in the API package and where the main code has converters in place to move the data in and out. However, this is a larger task then one might anticipate and could not be done on current funding.

So the technical compromise is to use JSON / JSON-LD / XML as data transfer, although it needs re-processing on the plugin side. This needs to be shipped somehow, and the external plugin needs a clear interface definition how to access the data and what kind of data is expected.

On the other hand, the main code must provide an actual implementation of the interface, otherwise the external plugin will never be able to query for the data. So this is the implementation of this interface, which is internal to the main code - it will never be shown to the external plugin, which is always relying on the interface.

It's not great, it's not perfect, but very reliable. Even if you would go for DTOs, this provider thing might still be there, as it could keep hold of the conversion process from the DB model to the external model.

Hope this makes all more sense now.

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/Exporter.java
qqmyers: Yes - thanks for the comments! There are definitely aspects of trying to do an incremental update without rewriting the whole export system. 

One thing I'd like to clarify is that the entries in the ExportDataProvider interface are not technically the content of the built-in exporters - they are internal representations of the dataset used in multiple places (to send to DataCite, used in archival Bags, embedded in the header of dataset pages) where a metadata exporter with the same/very similar name has been created to make those formats available for download/use in harvesting. Even before this PR, most of the internal Exporters either parsed the JSON format or just used the DatasetVersion object just to get the DataCiteXML or schema.org output directly. (The two exceptions are the HTML exporter which used the DDI export directly, and the one DDI exporter that provides the detailed variable-level file metadata - it did scan through all the related classes rather than starting from an XML/JSON serialization.) 

Some consequences of all of this are that replacing an internal exporter does not change what goes to DataCite, etc. (True before and after this PR). That would be useful future work - to reverse the logic and have the exporter output used internally rather than vice versa, at which point some of the get* methods could be dropped. Right now the JSON or ORE representations, along with the FileDetails are the ones intended to be complete, so they would be the only ones that would stay or be replaced by something cleaner. 

I would also say that if there is dataset metadata not available to Exporters from the existing interface methods, it's a bug/limitation that should be fixed. With the ORE representation in particular, the intent is to be complete enough to allow recreation of a dataset from an archival Bag, so we want to add anything missing to it for that purpose as well.

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/XMLExporter.java
qqmyers: Yes. In the code (not changed in this PR), it is basically cut/pasted as a String. I've added a comment in the interface Javadoc to note that it has to be a valid URI. (The code could be changed to enforce a URI, but I'd ask that that be done as a separate PR if its a priority.)

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
qqmyers: FWIW: I think this is one of the reasons that the internal representations are used to create the exports and not vice versa (see [other comment](https://github.com/IQSS/dataverse/pull/9175#discussion_r1192435908)) - the underlying code can create representations for different versions. A new PR that provides version-specific inputs to the exporter would be a good step in that direction. (I can't find it now but someone did try to refactor the getJsonLD (schema.org output) method and got stuck here (the job turned out to be bigger because exports aren't version-specific).)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
qqmyers: removed

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/XMLExporter.java
johannes-darms: How often and how frequent can we change the API and expect external develops to adapt their code? I would prefer to integrate this into this MR so there is a well-defined API that won't change in the near future...

kcondon:
Jenkins build is failing: 
[bld_fail_err.txt](https://github.com/IQSS/dataverse/files/11576493/bld_fail_err.txt)

@qqmyers @poikilotherm Any ideas?

poikilotherm:
/push-image

Test the new build also including submodules.

poikilotherm:
/push-image

Try push image again thx

poikilotherm:
Stupid me - obviously won't work... Needs to be pushed to `develop` first as command picks up the version from there, not the PR.

PR #9184, opened 2022-11-23
Iqss/9183 terms of access validation fix

### LINKED ISSUE #9183: Bug: Dataset saved twice when updating terms
We are using a slightly customized dataset terms page, on which we removed some input fields. Since v5.11, a strange effect occurs: When changing terms of a dataset and clicking "Save Changes", the save() method is called twice. This leads to the effect, that a subsequent action like "Publish" throws an OptimisticLockingError.

The cause is the introduction of the Javascript testTOADatasetPage() method in the dataset.xhtml file:
```javascript
function testTOADatasetPage() {
  var termsofAccessHidden, fileAccessRequestHidden;
  try{
      termsofAccessHidden = document.getElementById("datasetForm:tabView:termsofAccessHiddenLT").value;
      fileAccessRequestHidden  = document.getElementById("datasetForm:tabView:fileAccessRequestHiddenLT").value;  
  }
  catch (error){
      //terms not present so save...
      PF('blockDatasetForm').show();
      datasetSaveCommand();                        
  }                         
  if ('#{DatasetPage.isHasRestrictedFiles()}' === 'true' && fileAccessRequestHidden === 'false' && termsofAccessHidden === '') {
     //Not compliant show error keep page open...
  } else {
      PF('blockDatasetForm').show();
      datasetSaveCommand();
  }
}
```

In our case, the two hidden fields are not available, as we removed some blocks. This results in an error, catch-clause is executed and dataset is saved. Afterwards, as we don't have restricted files, the else-block is executed, and the dataset is saved again. 
Possible solution: Move the if-else statement into the try-block?

haarli:
Prevents that an dataset is saved twice (which might result in subsequent OptimisticLockingError problems), when the two hidden input fields for Terms of Access validation are not available (e.g. due to a customized terms page)

Closes #9183


qqmyers (approved):
Interesting issue - fixing a bug that only appears in a fork. W.r.t. the code in the PR, it makes sense to move the code into the try block - it should have no effect for the community version but is reasonable code clean-up - if anything is changed here in the future such that another way to have an error is introduced, the current code could end up doing a double save, as in the fork, and the PR prevents that possibility.

pdurbin:
@haarli can you please explain how to reproduce the problem? In the issue you said you removed some input fields? Which ones? Is there some other way to reproduce it?

haarli:
As you already mentioned, this is a very special behaviour which only takes effect when you mess around with the dataset-license-terms page. So this PR is just a code cleanup.
Here are the steps to reproduce:

**Prerequesites:**
The hidden input fields with ids "termsofAccessHiddenLT" and "fileAccessRequestHiddenLT" must be removed in dataset-license-terms.xhtml
https://github.com/IQSS/dataverse/blob/ac1454b50ee892bdb410d4dd96a56cba6883394d/src/main/webapp/dataset-license-terms.xhtml#L341-L342
(in our case, we just wanted to display the licenses list and the terms of use field, and removed everything else)

**Steps:**
1. Create a new dataset
2. Go to Terms Tab, edit the terms (e.g. changing the license) and click on save => The dataset is saved twice due to the javascript code. 
At first in the catch-clause, which is called because the hidden fields do not exist: https://github.com/IQSS/dataverse/blob/ac1454b50ee892bdb410d4dd96a56cba6883394d/src/main/webapp/dataset.xhtml#L1891
and then in the further code: https://github.com/IQSS/dataverse/blob/ac1454b50ee892bdb410d4dd96a56cba6883394d/src/main/webapp/dataset.xhtml#L1897
The DatasetPage Scope keeps the version after the first save, the database contains the newer version of the second save.
4. Directly click on Publish afterwards -> OptimisticLockingError because the versions of the page scope and the database differ.

scolapasta (approved):

kcondon:
Hi @haarli , would you please refresh this branch from develop since we've since update the version and this prevents us from building and testing. Thanks!

PR #9187, opened 2022-11-23
7940 stop harvest in progress

### LINKED ISSUE #7940: [feature request] stop an harvest job in progress
Hello,

I ran into the issue described here: https://groups.google.com/g/dataverse-community/c/O-NdDtgFrI0/m/os_KjdLxAQAJ

The process worked, but maybe it would be a good idea to have a button that would do the same thing on the harvest page? or maybe an API to reset stuck harvest?

Here is what i've done: 
#manually get the harvest job's id:
select * from harvestingclient;

#fix the issue - where {ID} is the database id of the harvesting client.
UPDATE clientharvestrun SET harvestresult=0 WHERE harvestingclient_id={ID} AND harvestresult = 2;
UPDATE harvestingclient SET harvestingnow = FALSE WHERE id={ID};

#restart payara "just in case"
systemctl restart payara

Take care,

Virgile

landreev:

This provides a mechanism for a dataverse sysadmin to stop a long-running harvesting job in progress. 

Closes #7940

will update w/ instructions

**Additional documentation**:


mreekie:
This just has the normal review and QA left.
It's hard to know how to size what's left as it's waiting for review and hasn't been reviewed in detail.
It looks like the review may be straightforward.

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvestingClientServiceBean.java
```diff
@@ -167,28 +167,20 @@ public void deleteClient(Long clientId) {
...
-            // TODO: what if there's no current run in progress? should we just
-            // give up quietly, or should we make a noise of some kind? -- L.A. 4.4      
-            
-            currentRun.setSuccess();
-            currentRun.setFinishTime(currentTime);
-            currentRun.setHarvestedDatasetCount(new Long(harvestedCount));
-            currentRun.setFailedDatasetCount(new Long(failedCount));
-            currentRun.setDeletedDatasetCount(new Long(deletedCount));
-        }
+        recordHarvestJobStatus(hcId, currentTime, harvestedCount, failedCount, deletedCount, ClientHarvestRun.RunResultType.INTERRUPTED);
```
qqmyers, line 170: Not sure I understand - is this a typo? the setHarvestSuccess method is using the RunResultType.INTERRUPTED?

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvestingClientServiceBean.java
```diff
@@ -198,28 +190,40 @@ public void setHarvestFailure(Long hcId, Date currentTime) {
...
+                clientIds = clientIds.concat(","+client.getId().toString());
             }
         }
         
         try {
-            return (Long) em.createNativeQuery("SELECT count(d.id) FROM dataset d, "
-                    + " dvobject o WHERE d.id = o.id AND o.owner_id in (" 
-                    + dvs + ")").getSingleResult();
+            return (Long) em.createNativeQuery("SELECT count(d.id) FROM dataset d "
+                    + " WHERE d.harvestingclient_id in ("
```
qqmyers, line 225: I can see where the join of dataset and dvobject isn't needed, but why the switch to harvestingclient_id? Was the old code just wrong or is this a typo?

landreev: (Sorry for the delay)
The new query is simpler, and correct, all the harvested datasets are uniquely tied to the specific harvesting configurations (clients). 

The old query must have followed the logic from pre-Dataverse 4 times; back then harvested datasets ("studies") were similarly tied to their "harvesting dataverses". (A harvesting dataverse could only contain harvested studies, all harvested from the same place). That old query is still producing  the expected results in our prod., because we still have dedicated collections for each harvested source. Strictly speaking, the dvobject-dataset joint in it was useful,  although `o.dtype='Dataset'` would achieve the same result.  But I just remembered that this was no longer required. In Dataverse 4+ it is possible to harvest into a collection that has other content - local datasets and/or datasets harvested from other places. So, shorter answer,  yes, that query was wrong.

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvestingClientServiceBean.java
landreev: That was of course a typo, thank you. 
(I'm working on automated tests right now that would catch this)

doc/sphinx-guides/source/admin/harvestclients.rst
```diff
@@ -21,6 +21,21 @@ Clients are managed on the "Harvesting Clients" page accessible via the :doc:`da
...
+
+For example:
+
+.. code-block:: bash
+
+  sudo touch /usr/local/payara5/glassfish/domains/domain1/logs/stopharvest_bigarchive.70916
+  sudo chown dataverse /usr/local/payara5/glassfish/domains/domain1/logs/stopharvest_bigarchive.70916
+
+We recommend that stop stop any running harvesting jobs using this mechanism if you need to restart the application server, otherwise the ongoing harvest will be killed, but may be left marked as if it's still in progress in the database.
```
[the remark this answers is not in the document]
landreev: Thank you, I adopted your version almost verbatim.

scolapasta (approved):

kcondon:
Test results:
1. This worked when halting a no set harvest from demo but did not seem to halt a no set harvest from prod. I will retest.
2. I am also seeing stack trace errors in the log related to this, perhaps we are pulling the rug out from some process and it gets a null ptr?

landreev:
Not sure if all of these problems are related to my PR. 
With the "no set" harvest from demo, I suspect it may just mean that the harvest finished too fast, before the stop file was created (would make sense if the default set on demo is not large enough).  
The import error is just that - the client failing to import a record it got from the server; whatever the reason for that was, it would not be related to anything in this PR (it didn't touch any of the actual harvesting/importing/etc. functionality). Some records failing to harvest should not by itself be considered a problem, as long as the harvesting job finishes, and the failure is properly recorded. 

The part about the long prod. harvest not stopping with the stop file in place does sound like a real problem. So let's run that again on Monday and I'll try to diagnose what's going on.

kcondon:
Retesting this am, found it was a user error around the stop file name syntax. Should be stopharvest.<name> but I confusingly named my  harvest client stop_harvest_demo so the correct file name given the syntax would be stopharvest.stop_harvest_demo. This worked.

Testing rerun behavior, ie pause use case. It works but needed to rerun it twice, will check again.

PR #9195, opened 2022-11-28
IQSS/9194-fix curate command validation

### LINKED ISSUE #9194: CurateCommand doesn't correctly run the new TermsOfAccess validation tests
The validation test in the CurateCommand (Update Current Version capability) was validating the currently published version after the new version's metadata was copied in. When the terms validation was added in that call, it didn't work right in the Curate command as the Terms in the new version had not yet been copied into the currently published version.  

This was detected at QDR where a dataset with an old published version with no terms and requestfileaccess = f could not be updated even though the draft version was OK (had requestfileaccess = t). Not sure if there are other use cases (can one use the API to edit/create a dataset that doesn't have valid terms? If so, using the Curate command on that could let the invalid terms get published).
5.12.1/develop

PR to follow.

qqmyers:
**What this PR does / why we need it**: This moves the call to validate the datasetversion to be after the termsofuseandaccess are updated. The makes the validation run on the terms that are to-be-published in the update-current-version/CurateCommand call which is the right thing to do.

Closes #9194 

**Special notes for your reviewer**: a minor thing but something we found/fixed at QDR

**Suggestions on how to test this**: Possibly just regression test that normal Update-Current-Version scenarios work? 
To test that the existing code is broken, the easiest thing I know of would be to edit the db to set file access request false for a published version with no terms of use and then have a draft with file access request true and try to Publish/Update Current Version for that. In the current code, it should fail and the new code should succeed.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: no change

**Is there a release notes update needed for this change?**: pretty minor.

**Additional documentation**:


mreekie:
Grooming:

This issue was in the New or No Status columns on the Dataverse Global Backlog Board.
Those columns have been removed.
This issue has been removed from the board

This is NOT a reflection of the priority of these issues.
However, as we have worked on the grooming process, it became clear that these issues which are in columns that are not stewarded will likely not get prioritized or sized.

To get this item onto the Dataverse Global Backlog Board, please reach out to one of the Stewards on the board.
You can also reach out to @mreekie and I can connect you with the steward of the appropriate queue

qqmyers:
This is a small PR that fixes a bug found at QDR. Sized at 10 due to the need to configure a test case (could be a 3 too).

sekmiller (approved):
Looks good. @qqmyers can you just make are that the branch is up to date with dev while it's in "ready for QA"? Thanks!

PR #5485, opened 2019-01-23
QDR-982 - sort by category

qqmyers:
Conflicts:
	src/main/java/edu/harvard/iq/dataverse/settings/SettingsServiceBean.java

## New Contributors

Welcome! New contributors should at least glance at [CONTRIBUTING.md](/CONTRIBUTING.md), especially the section on pull requests where we encourage you to reach out to other developers before you start coding. Also, please note that we measure code coverage and prefer you write unit tests. Pull requests can still be reviewed without tests or completion of the checklist outlined below. Thanks!

## Related Issues

- connects to #4959 Allow installations to determine order of files on Dataset page
## Pull Request Checklist

- [ ] Unit [tests][] completed
- [ ] Integration [tests][]: None
- [ ] Deployment requirements, [SQL updates][], [Solr updates][], etc.: None
- [ ] [Documentation][docs] completed
- [ ] Merged latest from "develop" [branch][] and resolved conflicts

[tests]: http://guides.dataverse.org/en/latest/developers/testing.html
[SQL updates]: http://guides.dataverse.org/en/latest/developers/sql-upgrade-scripts.html
[Solr updates]: https://github.com/IQSS/dataverse/blob/develop/conf/solr/7.3.0/schema.xml
[docs]: http://guides.dataverse.org/en/latest/developers/documentation.html
[branch]: http://guides.dataverse.org/en/latest/developers/branching-strategy.html

PR #9204, opened 2022-11-30
IQSS/4959-Support-By-Category-and-Folder-Grouping

### LINKED ISSUE #4959: Allow installations to determine order of files on Dataset page
Some installations will find it useful to define the order in which files in a dataset are presented to users. An example of this is QDR, where "0_" is prepended to the names of documentation files as a not-so-great way to get those documentation files to the top, since the ordering is alphabetical:

https://data.qdr.syr.edu/dataset.xhtml?persistentId=doi:10.5064/F6BUAX58

Consider an installation setting of ordering by a specific tag, date, alphanumeric, etc. or some other solution.

### LINKED ISSUE #6933: Add "sort by file path" to sort options for file listing on dataset page
In the Dataset overview page, there is a Sort button, which presently has the following options: name(a-z), name(z-a), newest, oldest, size, type.
In a round of testing with Dataverse at our institute, the suggestion was made to add an extra option: 'file path'. 

Regards,

Nico Poppelier
University Medical Centre Utrecht, The Netherlands

qqmyers:
**What this PR does / why we need it**:  This PR is an update to the original work in #5485 from QDR in 2019 that added a 'Group by Folder' option for the main dataset file table and an optional 'Goup By Category' option. These work in consort with the main sort options and do a presort/grouping of the files by Category tag (e.g. Documentation, Data, Code, custom tags) and/or Folder before the alphabetical, by age, etc. sorting is done. 

Closes #4959 
Closes #6933

The original work has been in production in QDR since 2019 but was dropped from consideration by IQSS in anticipation of further analysis of other possible options. Since then JHU and ASU have requested this functionality and JHU has supported updating the original branch via GDCC.

- Set the :CategoryOrder setting to something like "Documentation, Data, Code"
- In a dataset with multiple files, sprinkle some category tags and directory path entries
- Verify that the display order puts Documentation first (all files with that tag sorted according to current sort direction) followed by Data (again sorted in that group), etc. and that all files with a folder path appear after the group with no path.
- Verify that setting :OrderByFolder false turns off the by folder sorting
- Set :AllowUserManagementOfOrder true and 
  - Verify that turning the group by toggles on/off turn the groupings on/off
  - Verify that going to page 2 etc with more files or to/from the tree view doesn't mess with the group by checkbox displays or the grouping that is enabled.

image

The image shows the two new Group-By-Folder and Group-By-Tag checkbox selectors and shows some of the effect of the tag sort - the second file with the Data tag would normally be third in alphabetical order but has been pulled up to be before untagged files. (Not shown - a file in the /test sub-folder is now last in the list instead of in the top 3 files due to the group-by-folder setting.)

**Is there a release notes update needed for this change?**: in the PR

**Additional documentation**:


rtreacy (approved):
Checked as per instructions. Both checkboxes checked, unchecked Group By Folder, Grouped By Tag in all combinations. Checked Tree view always maintains folder order

kcondon:
Issues found:
1.:CategoryOrder doesn't appear to have an effect.
2. If user level sort is enabled, tag sorting is always in Data, Code, Doc order, ignores CategoryOrder.
3. Type sort doesn't seem to work correctly but this is preexisting.

pdurbin:
> :CategoryOrder

Wait, is a new database setting slipping in? Shouldn't this be MPCONFIG?

PR #9206, opened 2022-12-01
IQSS/9205-Add role-based provider for S3

### LINKED ISSUE #9205: S3 logging without credentials
Hello,
I'm using dataverse 5.10 and I can't upload files to S3 without credentials file and  by just defining an AWS-IAM role.

It is possible to upload only creating an AWS-IAM user with credentials (API key + secret) and adding credentials file to  the .aws directory in dataverse user home.

The problem would be solved by adding to the file: S3AccessIO.java
 
the AWS class: **InstanceProfileCredentialsProvider**

In the poviderChain at this line:

`AWSCredentialsProviderChain providerChain = new AWSCredentialsProviderChain(profileCredentials, staticCredentials);`

(line 1196 of the develop branch)

Is it enough? 

Is it possible to add the feature?

Thanks,
Alfredo

qqmyers:
**What this PR does / why we need it**: Prior to [the change to enable a static credential provider](https://github.com/IQSS/dataverse/commit/d038f85c94583a9b8fcf3c00008e9316f2b3919c) , Dataverse supported using role-based AWS S3 authentication. This PR restores this as discussed in the issue.

Closes #9205

**Suggestions on how to test this**: - regression test to make sure normal S3 access isn't affected. Could test role-based access - might be easiest to coord with the person reporting the issue since they know the setup for that (Don may also be a good resource). (Note that if you do test role-based access, you have to remove the creds in ~/.aws/credentials and any static creds (probably never set up for the test instance) or they will still get used as defaults and it will be hard to confirm role-based access is what's in use.

**Additional documentation**:


mreekie:
Grooming:

This issue was in the New or No Status columns on the Dataverse Global Backlog Board.
Those columns have been removed.
This issue has been removed from the board

This is NOT a reflection of the priority of these issues.
However, as we have worked on the grooming process, it became clear that these issues which are in columns that are not stewarded will likely not get prioritized or sized.

To get this item onto the Dataverse Global Backlog Board, please reach out to one of the Stewards on the board.
You can also reach out to @mreekie and I can connect you with the steward of the appropriate queue

qqmyers:
This is a fix for a regression reported by the community. I've suggested coordinating with the issue reporter since they would have the best insight into a test scenario. Otherwise, just regression testing for an S3 store is needed. Sized at 3. Could be a 10 if the idea is to test the restored S3 access mode in-house.

scolapasta (approved):

kcondon:
With Don's help we were able to confirm both programmatic (keys) based and roles based S3 access work independently but do not appear to work concurrently..

PR #9212, opened 2022-12-07
#9211 fix render logic display with TOA OR restricted files

### LINKED ISSUE #9211: Restricted Files Count not displayed on Terms Tab
Create a dataset
add a file
restrict the file - leave allow access request true and leave Terms of Access blank
go to the terms tab on dataset page
you'll see the license info and possibly guestbook, but not "Restricted Files + Terms of Access"
If you edit terms and fill in "Terms of Access" You'll see the count of restricted files and info on Request Access/terms of access

* When does this issue occur?
See above

Dataset page

* What happens?
Unless there are Terms of Access you don't see the count of restricted files on the terms tab 

all users who have permission to see the terms tab of a dataset

I would expect to see the count of restricted files and info on Request Access regardless of whether Terms of Access have been entered.

Current Dev 5.12.1

**Screenshots:**
With no Terms of access

Screen Shot 2022-12-07 at 11 31 54 AM

With Terms of Access 
Screen Shot 2022-12-07 at 11 31 09 AM

*

sekmiller:

Fixes render logic on the Terms tab when there are restricted files and no TOA entered

Closes #9211 Restricted files count not displayed on Terms Tab

Verify that if there are restricted files and no TOA you see the count and info on Request Access.

**Additional documentation**:


mreekie:
sized as a 3

src/main/webapp/dataset-license-terms.xhtml
```diff
@@ -237,7 +237,8 @@
                 </p:fragment>
             </div>
         </div>
-       <div class="panel panel-default" jsf:rendered="#{!empty editMode or empty editMode and ((!empty termsOfUseAndAccess.termsOfAccess and DatasetPage.restrictedFileCount > 0) 
+       <div class="panel panel-default" jsf:rendered="#{!empty editMode or empty editMode and (DatasetPage.restrictedFileCount > 0
```
scolapasta, line 240: @sekmiller  Does this even need the first two checks for edit mode since it seems they'd cancel each other out? If that goes away so can a parenthesis layer, I think, making it a little cleaner)

scolapasta (changes requested):

src/main/webapp/dataset-license-terms.xhtml
sekmiller: Ok. that makes sense. In my internal parser I was saying, "you want to render this if you're in edit mode or you're not in edit mode but at least one of the fields is filled in". I'll change it.

scolapasta: Ah, you're right - I missed that it was an AND and not an OR. Either way, removing it is simpler (since `empty editMode` has to be true if `!empty editMode` is false)

scolapasta (approved):

PR #9230, opened 2022-12-14
9229 - enable OIDC bearer token API access

### LINKED ISSUE #9223: Clarify in guides that APIs do not support OIDC
## When does this issue occur?

If you look at https://dataverse.org/software-features you see the following about OIDC:

"Login via OpenID Connect (OIDC)

Log in using your institution's identity provider or a third party. [More information](https://guides.dataverse.org/en/latest/installation/oidc.html)."

This is accurate. In the guides you see an explanation of OIDC support in Dataverse but we don't explicitly explain that OIDC cannot be used as a way to authenticate to Dataverse APIs.

To be clear, some day it would be very nice to support OIDC with PKCE as a way to authenticate to Dataverse APIs, but for now let's simply set expectations better in the guides (and perhaps on the features page). Otherwise, people building apps or even their own frontend to Dataverse might be disappointed.

## To whom does it occur (all users, curators, superusers)?

API users.

## Which version of Dataverse are you using?

5.12.1

### LINKED ISSUE #9229: As a developer, I'd like to use Bearer Token to send authenticated API requests
Placeholder for now. I'll extend it tomorrow. 

poikilotherm:

We want to enable OIDC bearer token based access to the API for existing OIDC users.

TODOs:
- [x] Retrieve provider metadata from existing provider
- [x] Create feature flag mechanism (should be reusable for other stuff, too) and make it configurable via JvmSettings
- [x] Add docs
- [ ] Probably some more...

Closes #9229
- Closes #9223

None yet.

None yet.

Nope.

Yes.

**Additional documentation**:
None yet.


vera:
I pushed a few commits fixing issues I ran into while testing here: https://github.com/vera/dataverse/commits/9229-api-oidc-access

I ran into one more issue which I am not sure how to fix: 

https://github.com/IQSS/dataverse/blob/3694d8119b6ba977c7aa6459c3d5a78b72fe3171/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L453

This line does not succeed in finding the correct user. If I understand correctly, it tries to look up an `authenticateduser` with a `userIdentifier` equal to the subject of the OIDC user info (for me, this is some hex string like `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`). But the `userIdentifier` of the user which should be found isn't that. It's the username I entered when I registered the user via the Dataverse UI. Thus, the lookup fails.

Is the OIDC user ID stored anywhere during user registration?

mreekie:
sizing:
- This is not ready for sizing yet.

mreekie:
sizing:
- Skipping for today.
- @GPortas @poikilotherm update in the comments when you believe this is ready to be sized and I'll catch it in the next pass

mreekie:
Sizing:
- Added it to the queue by adding the size queue label.
- Added bklog: needs discussion label to indicate that we will need to revisit this.
- All that changes nothing in terms of what's next. We'll size this when we get notified that it's ready

poikilotherm:
Replaced by #9532. Thanks @johannes-darms

PR #8901, opened 2022-08-06
GDCC/8449 use original files in archival bag

qqmyers:
**What this PR does / why we need it**: This PR updates the OAI-ORE metadata export to include a consistent set of name/mimetype, size, download URL, and checksum values that are for 

- non-tabular/not ingested: the file as is in the current version 
- tabular/ingested: the original file.

As noted in the issue, the current OAI_ORE, and hence archival bags as well include the checksum of the original file but, for tabular/ingested files, use the name/size/mimetype, and download URL (which is used to generate the bag) of the ingested version which makes the OAI_ORE inconsistent and makes it impossible to use the checksum to validate the bag contents.

**Which issue(s) this PR closes**:

Closes #8449

**Special notes for your reviewer**: This is a relatively straight-forward change to the OAI-ORE generator class. It doesn't address the broader issue(s) related to moving the original file upon ingest as recently discussed in tech hour. The part that's relevant here is that we don't change the checksum/don't calculate the checksum of the ingested version, so in addition to it being reasonable to include the original file in an archival bag (since ingest would occur if/when you read the bag back in), we don't have a checksum to use with the ingested version anyway. (In making the fix here, I also realized that there's also a minor issue in that the per-version filenames are only for the ingested version and the original file download only has the original name, i.e. the original download button for 'newfilename1.tab' in v2 of a dataset is 'oldname.csv' (versus newfilename1.csv). - not sure if that's worth an issue or is something I can address when refactoring how ingested files are stored for the remote overalay post 5.12 work).

**Suggestions on how to test this**: Publish datasets with a  mix of normal and ingested files. Verify that the OAI_ORE output and an archival bag (e.g. using the File Archiver) include the original file/metadata including name/mimetype/size/checksum for ingested files and include the name from the latest version, mimetype/size/checksum for normal files. Also verify that the schema:sameAs link in the OAI-ORE includes '&format=original' for the ingested files (so if you use it, you'll get the original file downloaded).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: no

**Is there a release notes update needed for this change?**: will include in main release note file.
[all the release notes for HDC 1 and 3A/3B, including this PR are combined in #8894 - L.A.]

**Additional documentation**:

PR #9245, opened 2022-12-22
Client-side multifile zip download

### LINKED ISSUE #5864: datasetversion_id not being written to guestbookresponse on initial dataset page open/refresh when 'multi file' Download button clicked
tested on v 4.13 but applies to earlier versions:

Scenario:
1. Open a dataset page (choose one with open files just to make process quicker)
2. Choose any number of files
3. Click the 'multi-file' Download button above the file panel
4. Files download
5. The datasetversion_id value is NOT written to guestbookresponse

Repeat as many times as desired - same result

6. Click one of the individual file 'Download' buttons.
7. File downloads
8. The datasetversion_id IS written to guestbookresponse

9. Choose several files and use the 'multi-file' Download button 
10. Files download
11. The datasetversion_id IS written to guestbookresponse

Now refresh the dataset page. The scenario starts at 1 again. 

So it seems that the datasetversion_id is only set, and therefore written to guestbookrespone, in the DataversePage.java on the Individual download click.

Until one of the individual download buttons is clicked, then, the dasetversion_id is not set and so will not be written to the guestbookresponse table.

qqmyers:
**What this PR does / why we need it**: A possible addition to/replacement for zipping on the server. In this PR, the multi-file download button invokes JavaScript that will download files individually (using direct download if enabled) and create a zip locally, using file names/directoryPaths from the specific datasetVersion being downloaded.

Current issues/limitations:

- It isn't clear that this will work on all browsers
- There's no error handling - should be possible, for example, to default to using the server side zip if things go wrong or if the browser type/version doesn't support what's needed.
- It should be more efficient, but I've done minimal testing on scalability so far. Nominally, one could allow users to download all files and not have a size limit. The underlying zip code is using Blobs and Promises and is supposed to scale, but I'm not sure I'm doing everything to configure it to scale, etc.
- The logic not allowing download of a zip when your over the size limit has not been changed, so this method is also subject to any limit so far.
- The download file is always named dataverse_files.zip as before - could potentially use the dataset PID/version to create a unique name (with full or partial to indicate some/all files)
- There is not currently any manifest file in the zip - should be possible to add one if desired (or to someday make a Bag)

Closes #5864

To enable this, I needed to know the datasetVersion in the download code which required trying to fix #5864 - the multifile button doesn't set the datasetVersion in the guestbook by default. If the rest gets delayed, it may be worth pulling out this one line fix (it's a separate commit).

**Additional documentation**:


donsizemore:
this is wonderful! does it default to original file format, or does it send surrogate copies?

qqmyers:
Right now it is adding `?format=original` when it retrieves each file.

donsizemore:
@qqmyers @jdmar3 corrects me: `From archival theory there is a case to be made either way, but I do argue that prioritizing original file formats over plaintext tabular data incurs tech debt and may make data unusable.` Would perhaps an "original format" check-box or some such require much additional work?

jdmar3:
@qqmyers Would it be possible to present an option to add `?format=archival` in addition to/instead of `?format=original`. Also, we're working on some automated user testing for different browsers, so I would be happy to help with testing on different browsers if needed.

qqmyers:
The Access Dataset menu at the top of the page allows getting either original or archival format. Currently I have not changed those buttons to use the client-size zipping but that's a useful addition once we know that it works well for most browsers and sizable datasets. 

Both forms are also available at the individual file level, so it is mostly a limitation of the bulk 'Download' button for selected files, regardless of whether the existing server-side zipping or this client-side method is used. I don't want to change that as part of this PR for client-side zipping, but I think both client-size and the existing server-side algorithm could handle both cases if the user interface work is done to allow it. FWIW: I think the [API call to download all files](https://github.com/IQSS/dataverse/pull/7086) allows you to specify either form as well.

W.r.t. archiving, I would argue that the Bag exports are better than the zip available from the front end (the Bag has fixity info, all the metadata for the dataset, etc.) and since it is privileged, it doesn't run the risk of files being excluded if you don't have permissions (if I recall the zip options in the UI include a manifest that lists files that weren't included due to permissions or size limits). There has also been discussion of the archival Bag exports w.r.t. whether including the ingested formats would be better than the original, but there are issues that have slowed that work, e.g. [the fact that Dataverse isn't storing the fixity info for the ingested versions](https://github.com/IQSS/dataverse/pull/8901). It would definitely be useful to have some discussion/review of the Bags to decide requirements and priorities.

W.r.t. to testing - thanks! The draft PR should work as is so if we can get a test server(s) set up somewhere, it could be tested with different browsers, larger data, etc. I think DataverseNO was going to try to fire one up, Don could probably do that at Odum as well. Assuming that looks promising, I can look into updating the download all buttons - that shouldn't involve any new risks - if it works for the one format, it will work for the other.

pdurbin:
We're excited about it. Let's let Jim size it.

PR #9222, opened 2022-12-12
Update dataverse.xhtml to add "read more/read less" buttons for dataverse descriptions

ar062800:
Added contentTruncateForDataverse() to allow dataverse descriptions to have "read more/read less" buttons.

**What this PR does / why we need it**: Our organization has a long description for our dataverse, which takes up too much space on the page. We needed a way to collapse the description. I reused the contentTruncate() function from dv_rebind_bootstrap_ui.js and added it to dataverse.xhtml scripts as contentTruncateForDataverse() to add this functionality.

**Which issue(s) this PR closes**: N/A

- Closes #1249
- Closes #9265

**Special notes for your reviewer**: N/A

**Suggestions on how to test this**: N/A

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Read More:
image

Read Less:
image

**Is there a release notes update needed for this change?**:

Yes, included.

**Additional documentation**: N/A

PR #9265, opened 2023-01-05
Truncate Dataverse Description Text

### LINKED ISSUE #1249: Dataverse - Description text truncated
We are allowing unlimited characters for textbox fields (like description), so we should truncate these on the view dataverse page (with a view more link)

qqmyers:

This was developed as a possible alternative to #9222 . It makes changes to the contentTruncate function to allow it to be used in both the dataset and dataverse pages.

It also adds some possible fixes/improvements:

- The `Collapse Description [+]` button shows a + sign like the `Read full Description [+]`  - I've changed that to use a minus sign `Collapse Description [-]` - I assume that was just a typo.
- The current code only truncates the first dataset description when there is more than one. Both would be displayed above the fold so I changed the selector to truncate any description that needs it.
- The original code calls `DatasetPage.datasetVersionUI.description.datasetFieldType.localeTitle` multiple times - I'm guessing one cset and reuse of that variable is more efficient.

The one thing that may be different is exactly where the page scrolls to when one re-collapses the description on a dataset page. The old code would have scrolled to the top of all descriptions whereas it will now scroll to the top of the description being collapsed. Since the old code would only collapse the first one, this should be a matter of a few pixels at most.  (FWIW - the technical reason is that the truncSelector is reused in the contentTruncate function and my change means truncSelector points to the specific dataset description rather than the top of the element containing all descriptions. This could probably be fixed if desired - not sure it is worth it or that scrolling to the top of all descriptions makes sense now that all descriptions get truncated.)

Closes #1249

**Additional documentation**:


ar062800:
Hello @qqmyers!

I tried testing this out, but it seems like it did not work. While trying to dig around to see what the problem was, I noticed that that while I was on a Dataverse page, in Sources Tab of Chrome Developer Tools -> javax.faces.resource -> js -> dv_rebind_bootstrap_ui.js.xhtml?version=5.12.1 was not the same as dv_rebind_bootstrap_ui.js in the source code.



Screen Shot 2023-01-17 at 10 02 43 AM


I am not sure why this is the case, but in dataverse_template.xhtml, 
I changed line 76 from: <script src="#{resource['js/dv_rebind_bootstrap_ui.js']}?version=#{settingsWrapper.appVersion}"></script>
to: <script src="#{resource['js/dv_rebind_bootstrap_ui.js']}"></script>
which removed the '?version=5.12.1' from dv_rebind_bootstrap_ui.js and then contentTruncate worked correctly.

To my understanding, it seems like the wrong version of dv_rebind_bootstrap_ui.js is being called in dataverse_template.xhtml?

qqmyers:
I think this is a caching issue in the browser. The idea of putting the ?version=X at the end avoids it for non-developers - when v5.13 comes out, the browser will auto-update it in users browser. Unfortunately, when we make a change to the script in development and the version doesn't change, your browser doesn't know to update. The work-around is to open that script in a separate pane and do the SHIFT-refresh to force the browser to get a new copy. After that the Dataverse pages will get the updated version.

ar062800:
Got it, thanks! That fixed the issue.

pdurbin:
@qqmyers thanks for showing the way to me and @ar062800

I'm going to go ahead and close this PR now that I've approved his:

- #9222

Issue #9307, opened 2023-01-19
[in progress] Update to documentation tables to convert inline tables to CSV generated tables


kuhlaid:
Going through the documentation, there are pages such as `\doc\sphinx-guides\source\admin\metadatacustomization.rst` which have inline tables. This makes me cringe. I cannot think of any reason why we should have tables created this way. The CSV generated tables on page `\doc\sphinx-guides\source\api\external-tools.rst` are much easier to update and does not clutter the .rst file.  It is a nightmare to make updates to inline tables and might be a reason why the generated PDF of the documentation cuts off the large table `\doc\sphinx-guides\source\admin\metadatacustomization.rst` on one page and does not extend to a second page (I plan to look into this issue while converting the tables to CSV generated to see if that corrects the PDF problem). 

If there are some obscure reasons why inline tables should exist in the documentation let me know, but I will begin converting the documentation currently using inline, over to CSV tables.

pdurbin:
@kuhlaid I'm the one who added the CSV/TSV style table for external tools and I like for some things but there's also a third way. Check out this PR by @poikilotherm

- #9299

It has a table that looks like this:

```
Feature Flags
-------------

Certain features might be deactivated because they are experimental and/or opt-in previews. If you want to enable these,
please find all known feature flags below. Any of these flags can be activated using a boolean value
(case-insensitive, one of "true", "1", "YES", "Y", "ON") for the setting.

.. list-table::
    :widths: 35 50 15
    :header-rows: 1
    :align: left

    * - Flag Name
      - Description
      - Default status
    * - Example flag
      - Replace this with something real
      - ``Off``

**Note:** Can be set via any `supported MicroProfile Config API source`_, e.g. the environment variable
``DATAVERSE_FEATURE_XXX``.
```

Just another option to play with and consider! 😄

kuhlaid:
Suggestions made.

PR #9308, opened 2023-01-19
9307 convert inline tables to csv generated tables

kuhlaid:
**What this PR does / why we need it**: Table formatting in the documentation can be a nightmare if tables are entered inline. In the PDF documentation, long tables will be truncated at the end of the page where the table begins and will not wrap to the subsequent pages. Using csv tables with the `:class: longtable` option fixes this issue for PDF builds of the documentation. Many of the large tables were converted to csv tables in this version of the documentation.

**Which issue(s) this PR closes**: #9307


pdurbin:
@kuhlaid hi! I'm closing this pull request in favor of this one:

- #9474

I realize my fix is much less ambitious than yours but in practice we've found that smaller chunks of work move more easily through the system. I hope you continue to engage and help us improve the PDF!

As I mentioned In that PR, we are now building it regularly (every time a PR is merged) at http://preview.guides.gdcc.io/_/downloads/en/develop/pdf/

I'm going to repeat this comment on your other open related PRs. Sorry! 😅

PR #9316, opened 2023-01-23
Fix for the broken "earliest date" (#9309)

### LINKED ISSUE #9309: OAI Server: "earliestDateStamp" in the repository broken in 5.12.1; causing problems with serving incremental harvests
This was an insane thing for me to have missed. 😢 
In the new XOAI implementation (gdcc/xoai-5.0.0-RC1), validation was added for the "from=" parameter, rejecting the request if the date supplied is before what the OAI Repository thinks is the "earliest date stamp" in it. *) But the RepositoryConfiguration is still initialized with the default value of
```
.withEarliestDate(DateProvider.now())
```
and it's never reset in the Dataverse code. And "now" in this context means the time the OAI service was initialized. 
Which in turn means that the server is going to reject any incremental request from a client that hasn't successfully harvested since the last time this Dataverse application was restarted (!).
Straightforward to fix; and we may want to consider applying the fix as a patch to our (Harvard) prod. before 5.13 is released. 

*) Is this validation really necessary? I'm accepting full responsibility for missing this; and I'm all for fixing it - if nothing else, we want to advertise the correct date in the Identify verb. But I'm not sure I understand the point of rejecting requests on account of this; and I'm not seeing this spelled out in the OAI spec. May not be looking in the right place, etc.

landreev:

A trivial fix for the embarrassing OAI server bug in 12.1

Closes #9309

(I have synced with develop and resolved the merge conflict; it is indeed ready for QA - but please see the note below)

One important note: the PR #9310, that has already been merged, provided a parallel fix for the worst manifestation of this bug, the part where incremental harvest requests were being rejected as described in the issue. This means you can no longer test that symptoms on develop branch in its current form (although it's still present in prod. and demo). (to clarify: it "provided a parallel fix" because the problem had also been fixed in the XOAI library, and in that PR we switched to the latest version of it). 

We still need this PR, because it fixes the remaining part of the bug. Namely, the "earliest record in the repository" time stamp is still advertised incorrectly, and this can still be observed in develop. For example, looking at `https://dataverse-internal.iq.harvard.edu/oai?verb=Identify`, the earliest record's date is shown to be the "epoch", i.e. `1970-01-01`:
```
<OAI-PMH xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
<responseDate>2023-01-30T17:15:05Z</responseDate>
<request verb="Identify">https://dataverse-internal.iq.harvard.edu/oai</request>
<Identify>
<repositoryName>Dataverse Internal Dataverse OAI Archive</repositoryName>
<baseURL>https://dataverse-internal.iq.harvard.edu/oai</baseURL>
<protocolVersion>2.0</protocolVersion>
<adminEmail>root@dataverse-internal.iq.harvard.edu</adminEmail>
<earliestDatestamp>1970-01-01T00:00:00Z</earliestDatestamp>
<deletedRecord>transient</deletedRecord>
<granularity>YYYY-MM-DDThh:mm:ssZ</granularity>
<compression>gzip</compression>
<compression>deflate</compression>
</Identify>
</OAI-PMH>
```
 (and in prod. and demo, it is still being reported as the time of the last Payara restart ☹️ )
 WIth the fix in this branch, it will change to the actual timestamp on the earliest record in the OAI sets on the server. 

**Additional documentation**:


mreekie:
Daily: 
- This has been raised by @landreev and has been ok'd for the next sprint.
- @landreev will size this and Stefano has Ok'd this to put at the top of our thesprint we start tomorrow.

landreev:
@mreekie 
>     * @landreev will size this and Stefano has Ok'd this to put at the top of our thesprint we start tomorrow.

I believe this is an honest 10 at most, review and QA combined. It's straightforward to reproduce, just as straightforward to verify that the fix does indeed fix it. 
I would size it at even lower, but I'm calling it a 10 because there's going to be more than 1 person involved (somebody will have to review it and Kevin will have to test it).

sekmiller (approved):

PR #9322, opened 2023-01-24
bump sample heap size to something large enough for install process

### LINKED ISSUE #9321: sample GLASSFISH_HEAP is too small
In testing Payara6, we found that the sample GLASSFISH_HEAP setting at https://github.com/IQSS/dataverse/blob/develop/scripts/installer/default.config#L7 causes Payara to run out of memory during the installation routine. Will submit a PR to bump this to 4096.

donsizemore:

Install process was failing with database errors; succeeded once we bumped this value.

Closes #9321 

None

Run the installer with the previous and proposed values

No

No

**Additional documentation**:

None


scripts/installer/default.config
```diff
@@ -4,7 +4,7 @@ GLASSFISH_USER = dataverse
 GLASSFISH_DIRECTORY = /usr/local/payara5
 GLASSFISH_ADMIN_USER = admin
 GLASSFISH_ADMIN_PASSWORD = secret
-GLASSFISH_HEAP = 2048
+GLASSFISH_HEAP = 4096
```
pdurbin, line 7: This might be a dumb question but is another option to remove this line? I thought the installer calculated the heap based on what's available, but I could be mistaken.

pdurbin (commented):
I'm just adding a question, really.

scripts/installer/default.config
donsizemore: I don't know about removing the line (since it's an available option) but if you'd like I can replace the longstanding sample value with a comment that says "set this to roughly half of the system RAM, in MB" or some such? I don't like hard-coding a value, either...

scolapasta:

pdurbin:
> If we don’t hear from you by May 22, 2024, we’ll go ahead and close this PR (it can always be reopened after that date, if there is still interest).

Closing.

PR #9303, opened 2023-01-19
9293 - New filter-based design for the API authentication mechanisms (1/2)

GPortas:
## What this PR does / why we need it:
This PR includes the refactoring changes for transitioning to the new filter-based design described in #9293 for the current API authentication mechanisms.

Due to the magnitude of the changes and in order not to overextend the revision work of this PR, I have decided to limit the scope of this PR in relation to issue #9293.

In particular, this PR covers the following points:

- The new authentication model is fully implemented, including all current authentication mechanisms (previously handled in the _findUserOrDie_ method).
- Only _Datasets_ API methods have been evolved to use the new filter-based design. The rest of the API resources (_Files_, _Dataverses_, etc) still use the old model (_findUserOrDie_).
- The _findUserOrDie_ method has been marked as `@Deprecated`, as well as any other methods in the _AbstractApiBean_ class related to the old auth design.

My idea for a next iteration is to evolve the rest of the API methods to use the new auth design so we can remove the old design logic (methods marked as `@Deprecated` in this PR).

## Which issue(s) this PR closes:

Partially completes #9293 but does not close it.

## Special notes for your reviewer:

Not yet

## Suggestions on how to test this:

The behavior of the API must be the same as before, so in addition to the automated IT tests, curl calls to the API using different types of credentials can be used to test the different auth mechanisms.

## Does this PR introduce a user interface change? If mockups are available, please link/include them here:

No

## Is there a release notes update needed for this change?:

Not sure

## Additional documentation:

Not yet / Not sure

PR #9360, opened 2023-02-02
9293 - Apply filter-based auth for all API endpoints (2/2)

### LINKED ISSUE #9293: New filter-based design for the API authentication mechanisms
## Overview of the Feature Request

Instead of calling the AbstractApiBean's findUserOrDie method on each secured endpoint to authenticate the request through one of the possible authentication credentials, this responsibility can be delegated to an outer layer. Since Dataverse already uses the JAX-RS standard in its API, the layer would nominally be a JAX-RS ContainerRequestFilter doing authentication/revalidation as a _PreMatch_ operation which would either reject the request (if authentication fails) or pass the request to the underlying API method with the logged in user account info included. To mark API endpoints that require authentication and therefore filtering, a custom annotation can be used. 

This new design would support Dataverse’s existing API authentication methods into such a filter (or another one(s) - filters can be stacked). This may be most straight-forward for the stateless SignedURL mechanism. The other mechanisms, which require Database lookups, could also be moved to a filter but would still have dependencies on internal Dataverse code.

This new design improves the testability and extensibility of the API authentication mechanisms.

(_Specification extracted from Dataverse - SPA Authentication document_)

## What kind of user is the feature intended for?
API User, future SPA user

## What inspired the request?
- Re-Architecture decisions for Dataverse Authentication and related Proof Of Concept (PoC) developed.
- Dataverse - SPA Authentication document

PoC repository link: https://github.com/GPortas/dataverse/tree/poc/api_security_filter

## What existing behavior do you want changed?
AbstractApiBean's findUserOrDie method

## Any brand new behavior do you want to add to Dataverse?
JAX-RS ContainerRequestFilter for authentication/revalidation

## Any related open or closed issues to this feature request?
- #9229 
- #9063 

The API auth mechanisms requested in these issues should be part of the new filter-based mechanism.

GPortas:
## What this PR does / why we need it:

This is a continuation of the refactoring work done for PR #9303, containing the remaining changes to close issue #9293. **It is necessary to review and merge PR #9303 before reviewing this PR**.

In particular, this PR replaces the calls to the deprecated _findUserOrDie_ and _findAuthenticatedUserOrDie_ methods with the filter-based design and removes the deprecated methods, no longer used by any API endpoint.

## Which issue(s) this PR closes:

Closes #9293 

## Special notes for your reviewer:

Not yet

The behavior of the API must be the same as before, so in addition to the automated IT tests, curl calls to the API using different types of credentials can be used to test the different auth mechanisms.

## Does this PR introduce a user interface change? If mockups are available, please link/include them here:

No

## Is there a release notes update needed for this change?:

Not sure

## Additional documentation:

Not yet / Not sure


qqmyers:
FWIW - In another conversation, it was just pointed out to me that the API calls in the mydata package all appear to use https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L148 instead of findUser/findAUthenticatedUserOrDie. They should probably be updated to use the filter as well, either here or as a new issue/PR.
These may be the only api calls outside the api package - I didn't find any @Path annotations anywhere else.

GPortas:
> FWIW - In another conversation, it was just pointed out to me that the API calls in the mydata package all appear to use
> 
> https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L148
> 
> instead of findUser/findAUthenticatedUserOrDie. They should probably be updated to use the filter as well, either here or as a new issue/PR.
> These may be the only api calls outside the api package - I didn't find any @path annotations anywhere else.

@qqmyers

The only call to the _getUserFromIdentifier_ method used to authenticate a user is wrapped in a DEBUG condition (which is always false and it might be interesting to remove it, since it never gets called), please take a look at the call I'm referring to:  https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L287

The rest of the calls to _getUserFromIdentifier_ are made once the user has been authenticated through another mechanism. These mechanisms are checked separately, not via the old _findUserOrDie_ or _findAuthenticatedUserOrDie_ methods, as you mention.

In effect, that endpoint is a candidate for the Auth Filter to be applied on it. By having already identified this, I think it's a good time to also refactor that endpoint in this same PR.

FWIW - I see that session authentication is also enabled for such endpoint (See https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L291), I don't know if there is any special reason. Perhaps the endpoint is called or has been called in the past from JSF, so it requires to authenticate by Session Cookie. In any case, the Auth Filter can be applied while also maintaining session authentication for this API resource, if necessary.

qqmyers:
@GPortas - Thanks! Sorry for the bad code reference. The real issue that was found is that signedURLs don't work with these calls. It's because this class is checking for a session or api key (I think at https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L307) by itself so it therefore isn't seeing the newer methods like signedURLs and workflow tokens.

MyData pre-dates me, but I think you are probably right about the session being accepted - the MyData page apparently uses different technologies and probably calls that API directly. I would guess that use is current and should be kept, but I haven't checked.

It does look like removing that if DEBUG statement that is always false would be good code clean-up.

GPortas:
> @GPortas - Thanks! Sorry for the bad code reference. The real issue that was found is that signedURLs don't work with these calls. It's because this class is checking for a session or api key (I think at
> 
> https://github.com/IQSS/dataverse/blob/ac7aef757efc88f60179b2724ec99ff3c2a35853/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java#L307
> 
> ) by itself so it therefore isn't seeing the newer methods like signedURLs and workflow tokens.
> MyData pre-dates me, but I think you are probably right about the session being accepted - the MyData page apparently uses different technologies and probably calls that API directly. I would guess that use is current and should be kept, but I haven't checked.
> 
> It does look like removing that if DEBUG statement that is always false would be good code clean-up.

@qqmyers 

Great, thank you for the clarification. That issue will be solved by applying the Auth Filter to the endpoint, which includes the newer methods you mention.

I'm already working on these changes, to add them to this PR.

rtreacy (approved):
status codes, messages, more auth, tests

kcondon:
@GPortas This pr has the same signedurl issue as part one -decoding? Would a refresh of this branch from develop solve the problem?

PR #9366, opened 2023-02-03
Add initial support for .eln file format

### LINKED ISSUE #9363: Feature Request/Idea: Support for .eln file format
Hello everyone,

I am Nicolas CARPI, lead dev of [eLabFTW](github.com/elabftw/elabftw/), an Electronic Lab Notebook. I am also one of the founder of [The ELN Consortium](https://github.com/TheELNConsortium), where we managed to group several ELN software together to improve interoperability of these softwares.

We achieved this by designing a file format called .eln: [see ELN file format repo](https://github.com/TheELNConsortium/TheELNFileFormat).

Fear not, we didn't re-invent the wheel, it's basically a RO-Crate ([Research Object crate](https://www.researchobject.org/ro-crate/1.1/introduction.html)) inside a zip.

This format allows for import/export of research data such as experimental results, protocols, samples descriptions, experiments templates, etc...

Basically, there is a .json file at the root that describes the content + some data such as author name/affiliation/orcid, etc... And the nice thing is that all the fields are standardized.

I'm opening this issue to:

* Let you know about this format
* Ask you if you're interested eventually to add support for it, to facilitate depositing raw data along with how the data was produced: meaning the lab notebook entries corresponding to the raw data

You can look [here](https://github.com/TheELNConsortium/TheELNFileFormat/tree/master/examples/elabftw) to get a precise idea of what it looks like.

I believe adding support for this format would not be too difficult because opening a zip and reading json are very basic things in any language.

Users of ELN often ask the question about Data repositories, so there is definitely a strong demand from researchers to be able to upload their procedures/experiments to a Data repo without having to input everything manually again.

Please let me know what you think about this proposal.

Best,
~Nico

Support upload of .eln files.

Depositor.

The wish to increase interoperability between ELNs and data repositories.

I want depositors to be able to upload a .eln file that will be correctly understood by Dataverse, and populate fields based on what it finds in the .eln file.

Ability to parse .eln files, which are json files inside a zip.

Not that I know of.

NicolasCARPi:

This PR adds initial support for .eln files,  which are zip archives.

It closes not #9363 but it gets the ball rolling. Using yoda speak here to prevent github from closing upon merge.

Given that the file is a zip, I didn't edit MimeTypeDisplay or MimeTypeFacets.

Try and upload a .eln file, as can be found here: https://github.com/TheELNConsortium/TheELNFileFormat/tree/master/examples/elabftw

I don't think so.

Yes.


src/main/java/propertyFiles/MimeTypeDetectionByFileExtension.properties
```diff
@@ -3,6 +3,7 @@ ado=application/x-stata-ado
 dbf=application/dbf
 dcm=application/dicom
 docx=application/vnd.openxmlformats-officedocument.wordprocessingml.document
+eln=application/zip
```
pdurbin, line 6: I saw "Given that the file is a zip, I didn't edit MimeTypeDisplay or MimeTypeFacets" in the PR but it would be nice (some day) to have a mime type for .eln. I couldn't find an issue for this, so I created one:

https://github.com/TheELNConsortium/TheELNFileFormat/issues/22

With a mime type in place, in Dataverse we can show whatever we want such as "ELN file" or "Electronic Lab Notebooks" rather than whatever we show for zip files.

I'd say we can go ahead and merge this in the spirit of incremental progress, but I'd feel remiss if I didn't at least mention it.

doc/release-notes/9363-eln.md
```diff
@@ -0,0 +1,7 @@
+Support for .eln files has been added.
+
+See file format specification: https://github.com/TheELNConsortium/TheELNFileFormat
+
+The .eln file format is used by Electronic Laboratory Notebooks as an exchange format for experimental protocols, results, sample descriptions, etc...
+
+A Solr schema update is required.
```
pdurbin, line 7:
```suggestion
In order for existing .eln files to show as zip files (instead of unknown), you must run the file re-detection API on them and reindex them into Solr.
```

pdurbin (commented):
@NicolasCARPi thanks for the pull request. Please take a look at the comments I left.

pdurbin:
This is a tiny one-line change to detect .eln files as zip files rather than unknown. I'm giving this our smallest size, which is 3.

src/main/java/propertyFiles/MimeTypeDetectionByFileExtension.properties
NicolasCARPi: :+1:

NicolasCARPi:
Hello,

I have update this PR with the now accepted IANA mime type: https://www.iana.org/assignments/media-types/application/vnd.eln+zip

src/main/java/propertyFiles/MimeTypeFacets.properties
```diff
@@ -170,6 +170,7 @@ application/x-7z-compressed=Archive
 application/x-xz=Archive
 application/warc=Archive
 application/x-iso9660-image=Archive
+application/vnd.eln+zip=ELN Archive
```
pdurbin, line 173: I have some doubt about this. We try to put file types in a few buckets...

Screen Shot 2023-02-09 at 8 53 44 AM

... so I'm not sure we should create a new type, "ELN Archive", or not. That is, maybe it should be lumped under "Archive".

Another approach would be to put it under "Data".

I guess I'm ok with "ELN Archive" (though not everyone will know what ELN stands for) but I'm curious if anyone feels strongly about this.

pdurbin (commented):
Just a comment about how we put files in broad buckets.

src/main/java/propertyFiles/MimeTypeFacets.properties
NicolasCARPi: I've pushed a commit in that direction, I agree that there is no need to create another bucket.

NicolasCARPi:
> inspecting the first few bytes of the file or something.

The magic number will be the same as zip (because at the end of the day it's a zip). But a 100% effective way to detect it (on top of the file extension), is to look for the presence of `ro-crate-metadata.json` file in the archive. This file is guaranteed to be there in all .eln. And it will also be the entrypoint when importing its data, so if the file is not present it might raise an error such as "Expecting ELN file format but could not find metadata file!".

pdurbin (approved):
This is looking good. Thanks for the pull request and for responding to my feedback, @NicolasCARPi!

This should be a nice stepping stone to other ideas expressed in the linked issue.

PR #9377, opened 2023-02-09
9376 - Fix the order of metadatablocks in templates

### LINKED ISSUE #9376: Fix order of metadatablocks when creating and editing dataset templates in NDR collection
When creating and editing dataset templates in other collections with custom metadatablocks, the custom blocks are always at the bottom.

But when creating and editing dataset templates in [the NDR collection](https://dataverse.harvard.edu/dataverse/NegotiationDataRepository), the customNDR metadatablock appears at the top, above the Citation block:

<kbd>Screen Shot 2022-08-01 at 12 18 51 PM</kbd>

Can we fix the order of the metadatablocks when creating and editing dataset templates so that the custom block is at the bottom?

(When creating and editing dataset metadata, the Citation block is at the top. This was fixed in https://github.com/IQSS/dataverse.harvard.edu/issues/165)

scolapasta:

This fixes the order of the metadata blocks when viewing or editing templates.

Closes #9376 

The changes made here were parallel to the changes made in https://github.com/IQSS/dataverse/pull/8772 .  

So now, the Maps for metadata blocks in Template.java are TreeMaps. (note: metadata blocks were already made Comparable in that PR, so that change did not need to be carried over)

Create or edit or  a template in a dataverse collection for a dataverse that has a custom block. Confirm that the order of the blocks is the same as they appear on the dataset page. (namely that Citation block is first)

Technically, yes, in that it changes the order the blocks are displayed. But I wouldn't consider that a UI/UX change.

No

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Template.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Template.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 
 import javax.json.Json;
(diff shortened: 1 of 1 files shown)


qqmyers (approved):
Makes sense. A clear parallel with the dataset case. The only minor comment I have is that I'm not sure that interfaces need to all be for a TreeMap - as long as the two constructors change from HashMap, the others could still be Map. That said, the dataset code did the same thing and it's a minor point, so not really worth a change.

scolapasta:
@qqmyers I agree 100% about the interfaces, but since it was done that way for dataset, I decided to keep in that way here as well. Alternatively, we could switch back to Map and also change the ones in DatasetVersionUI? Not sure if it's worth the effort.

qqmyers:
Yep - no change requested...

kcondon:
I'm seeing all blocks, even ones not enabled, appearing when create template in a sub collection. Also, not sure how the ordering in create/edit dataset happens, citation is first but the others can't determine what ordering criteria is used.

PR #9383, opened 2023-02-10
Support for deleting files using native API

### LINKED ISSUE #3913: Is a delete file endpoint available?

ErykKul:
Support for deleting files using native API.

Closes #3913

Yes.


src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+    @Path("{id}")
+    public Response deleteFileInDataset(@PathParam("id") String fileIdOrPersistentId){
+
+        if (!systemConfig.isHTTPUpload()) {
+            return error(Response.Status.SERVICE_UNAVAILABLE, BundleUtil.getStringFromBundle("file.api.httpDisabled"));
+        }
+        // (1) Get the user from the API key
+        User authUser;
+        try {
+            authUser = findUserOrDie();
```
pdurbin, line 341: When you merge the latest from develop (please do!) you'll find that findUserOrDis can't be found. This is the error I got:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project dataverse: Compilation failure
[ERROR] /Users/pdurbin/github/iqss/dataverse/src/main/java/edu/harvard/iq/dataverse/api/Files.java:[335,24] cannot find symbol
[ERROR]   symbol:   method findUserOrDie()
[ERROR]   location: class edu.harvard.iq.dataverse.api.Files

I'm not 100% sure myself what the way is but I'm guessing it's something like this (based on other parts of Files.java after you merge):

```
DataverseRequest req;
try {
    req = createDataverseRequest(getRequestUser(crc));
} catch (Exception e) {
    return error(BAD_REQUEST, "Error attempting to request information. Maybe a bad API token?");
}
```

Please take a look and hack around. We can both learn a bit by studying PR where this change was introduced:

- #9303

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+        } catch (AbstractApiBean.WrappedResponse ex) {
+            return error(Response.Status.FORBIDDEN, BundleUtil.getStringFromBundle("file.addreplace.error.auth"));
+        }
+
+        msg("DELETE!");
```
pdurbin, line 346: Please delete this DELETE! 😄

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+
+        if (!systemConfig.isHTTPUpload()) {
+            return error(Response.Status.SERVICE_UNAVAILABLE, BundleUtil.getStringFromBundle("file.api.httpDisabled"));
+        }
+        // (1) Get the user from the API key
+        User authUser;
+        try {
+            authUser = findUserOrDie();
+        } catch (AbstractApiBean.WrappedResponse ex) {
+            return error(Response.Status.FORBIDDEN, BundleUtil.getStringFromBundle("file.addreplace.error.auth"));
```
pdurbin, line 343: This might change when we fix the code above anyway, but I'll mention that usually when we catch a WrappedResponse we `return wr.getResponse()`.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+        msg("DELETE!");
+
+        // (2) Delete
+        boolean deletePhysicalFile = false;
+        try {
+            DataFile dataFile = findDataFileOrDie(fileIdOrPersistentId);
+            List<FileMetadata> fileToDelete = dataFile.getFileMetadatas();
+            Dataset dataset = dataFile.getOwner();
+            DatasetVersion v = dataset.getOrCreateEditVersion();
+            DataverseRequest dvRequest2 = createDataverseRequest(authUser);
```
pdurbin, line 355: If we do use createDataverseRequest above we shouldn't need it again here.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+                    logger.warning("Failed to delete the physical file associated with the deleted datafile id="
+                            + dataFile.getId() + ", storage location: " + fileService.getPhysicalFileToDelete(dataFile));
+                }
+            }
+        } catch (WrappedResponse ex) {
+            return error(BAD_REQUEST, "There was no file found for ID: " + fileIdOrPersistentId);
+        }
+
+        return ok(deletePhysicalFile);
+    }
```
pdurbin, line 382: I'm just putting this on the last line because this is where tests go.

Can we please have some integration tests to ensure this new functionality is working?

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -317,6 +322,64 @@ public Response replaceFileInDataset(
...
+    /**
+     * Delete an Existing File 
+     * 
+     * @param id file ID or peristent ID
+     */
+    @DELETE
+    @Path("{id}")
+    public Response deleteFileInDataset(@PathParam("id") String fileIdOrPersistentId){
+
+        if (!systemConfig.isHTTPUpload()) {
```
pdurbin, line 335: Why is this isHTTPUpload check here? Is it necessary? Here's what's in the bundle, which seems to be about upload:

file.api.httpDisabled=File upload via HTTP is not available for this installation of Dataverse.

I guess the feeling is that if upload is disabled, delete should be disabled too? However, I don't see this check in the SWORD code for "delete file": https://github.com/IQSS/dataverse/blob/v5.13/src/main/java/edu/harvard/iq/dataverse/api/datadeposit/MediaResourceManagerImpl.java#L146

pdurbin (commented):
Hi @ErykKul thanks for the pull request! People have wanted this for a long time!

I was going to test this but hit a snag when merging the latest from develop. Can you please try and check out my comments about this and other things? Thanks!

src/main/java/edu/harvard/iq/dataverse/api/Files.java
pdurbin: findUserOrDie was deprecated in the PR above and removed in this PR:

- #9360

pdurbin: This is the javadoc that was deleted when findUserOrDie was deleted:

In case you are implementing a new endpoint that requires user authentication, here is an example of how to apply the new filter-based authentication:
{@link edu.harvard.iq.dataverse.api.Datasets#getDataset(ContainerRequestContext, String, UriInfo, HttpHeaders, HttpServletResponse)}

That code is here: https://github.com/IQSS/dataverse/blob/f97fc0567b0984f3be3069d7263c16168ee23061/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L249

But I suppose other examples could be followed as well.

rtreacy: methods that used the old API are now anotated @AuthRequired and have @Context  ContainerRequestContext crc
eg
    @POST
    @AuthRequired
    @Path("{id}/replace")
    @Consumes(MediaType.MULTIPART_FORM_DATA)
    public Response replaceFileInDataset(
                    @Context ContainerRequestContext crc,
                    @PathParam("id") String fileIdOrPersistentId,
                    @FormDataParam("jsonData") String jsonData,
                    @FormDataParam("file") InputStream testFileInputStream,
                    @FormDataParam("file") FormDataContentDisposition contentDispositionHeader,
                    @FormDataParam("file") final FormDataBodyPart formDataBodyPart
                    ){
...
       User authUser = getRequestUser(crc);

where AbstractApiMethod is
    protected User getRequestUser(ContainerRequestContext crc) {
        return (User) crc.getProperty(ApiConstants.CONTAINER_REQUEST_CONTEXT_USER);
    }

ErykKul: I have merged the develop branch and added annotation/replaced getting user method.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
ErykKul: Done.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
ErykKul: Try-catch was no longer needed after adding the annotation etc.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
ErykKul: I have moved the DataverseRequest creation to just after the user retrieval.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
ErykKul: I have removed the check. If we ever decide that it is needed, we can add it back.

ErykKul:
@pdurbin thank you for your review. I have implemented the changes. I am not sure if the integration test that I wrote works correctly, as I could not run any integration tests. Maybe it has something to do with my testing on Payara 6, I do not know yet. Probably something to do with the configuration on my machine. I will try to fix it and verify the test.

ErykKul:
Integration test is working now.

pdurbin:
Yes! Your test is passing on our Jenkins (firewalled off for you, sorry!): https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9383/11/testReport/edu.harvard.iq.dataverse.api/FilesIT/testDeleteFile/

pdurbin (approved):
Looks good. Approved. I fixed up the code and docs and added more tests. Thanks, @ErykKul !

kcondon:
Issues found:
1. Deleting a file published in v1 from an already existing draft dataset fails with command line response:
{"status":"ERROR","message":"Delete failed for file ID 323544 (CommandException): Command [edu.harvard.iq](http://edu.harvard.iq/).dataverse.engine.command.impl.UpdateDatasetVersionCommand@4f5d52a8 failed: Cannot merge an Entity that has been removed: [edu.harvard.iq](http://edu.harvard.iq/).dvn.core.study.FileMetadata[id=303880]"}
and server log entry:
[del_pub_file_err.txt](https://github.com/IQSS/dataverse/files/11023144/del_pub_file_err.txt)

ErykKul:
@kcondon 

I have already this type of integration test:
- create a new draft dataset
- add file
- delete this file (still draft)
- add another file
- publish
- delete the second file

Now I extended it such that two files are added before publish, when the first file is deleted, the dataset becomes draft, then the second file gets deleted. I think this should reproduce the scenario? Unfortunately, I was not able to run the integration tests, all tests are failing. I have the impression that something broke all integration tests? Or did just something happen on my machine? Can you still run the tests?

Greetings,
Eryk

pdurbin:
On Jenkins only one test is failing. From https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9383/14/testReport/edu.harvard.iq.dataverse.api/FilesIT/testDeleteFile/

```
java.lang.AssertionError: 
JSON path data.files[0] doesn't match.
Expected: null
  Actual: {restricted=false, directoryLabel=data/subdir1, description=my description3, dataFile={filename=cc0-1.png, storageIdentifier=file://1870f8168fc-25ecb0f848a7, checksum={type=SHA-1, value=a11205337740bf92502cbab3985bc278931410cd}, description=my description3, id=314, filesize=1124, categories=[Data], creationDate=2023-03-23, contentType=image/png, rootDataFileId=-1}, label=cc0-1.png, categories=[Data], version=1, datasetVersionId=129}

	at edu.harvard.iq.dataverse.api.FilesIT.testDeleteFile(FilesIT.java:1981)
```

kcondon:
@ErykKul I am out today but your updated test would test the failing case. Thanks!

ErykKul:
I thought I have found what the problem was, but I still can't run the integration tests. I am not sure if it only at my machine, or if there is a more general problem with the tests (the resulting aoi docker image does not work properly and many tests fail).

pdurbin:
@ErykKul here's the latest from https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9383/18/testReport/edu.harvard.iq.dataverse.api/FilesIT/testDeleteFile/

```
java.lang.AssertionError: 
JSON path data.files[0] doesn't match.
Expected: null
  Actual: {restricted=false, directoryLabel=data/subdir1, description=my description3, dataFile={filename=orcid_16x16.png, storageIdentifier=file://187143e3d0f-e782829cb196, checksum={type=SHA-1, value=ada69beeb122b099475efdaf88133c82bd0890b2}, description=my description3, id=314, filesize=1261, categories=[Data], creationDate=2023-03-24, contentType=image/png, rootDataFileId=-1}, label=orcid_16x16.png, categories=[Data], version=1, datasetVersionId=129}

	at edu.harvard.iq.dataverse.api.FilesIT.testDeleteFile(FilesIT.java:1979)
```

I haven't tried running the IT tests in docker-aio for quite some months. In the [Containerization Working Group](https://groups.google.com/g/dataverse-community/c/3vJWcAK5F7M/m/h63JizJKBgAJ), we've talked about the possibility of using the new Dockerized Dataverse environment (PR #9439) for this, to allow developers like yourself (or me!) run the integration tests within Docker. Would that be of interest?

pdurbin:
@ErykKul do you need a hand with this? How can I help?

ErykKul:
@pdurbin 
I think that running the integration tests with the Dockerized Dataverse is a great idea. Maybe we could use docker for that as well? For example: https://hub.docker.com/_/docker. Right now, the aio-docker makes some assumptions about the environment that are not true for every system (I still could not run the integration tests lately...). I am not sure about this, but it looks to me that it is no longer possible to build aio-docker using latest Rocky Linux image, like specified in the docker file? I get errors about not sufficient entropy (it also takes very long time to start the Dataverse inside the aio-docker container). Also, the "ps" command is missing, preventing the solr starting properly (procps package is not installed by default). Finally, when I get everything running, then I get errors from the sword API... I am not sure how the continuous integration runs the integration tests, I was not successful in doing that (I did get them running when trying out payara 6, but I had upgraded many packages to do that, upgraded the jdk version to 17, changed some scripts, etc.).

I have tried to fix the integration test, but since I was not able to run it locally, I am waiting for the continuous integration to finish.

ErykKul:
@pdurbin, @kcondon 
I have solved the entropy problem on my machine by installing the rng-tools. Now the tests run, so I was able to fix it.

pdurbin:
@ErykKul thanks for the feedback on testing in Docker. I passed it along at https://dataverse.zulipchat.com/#narrow/stream/375812-containers/topic/running.20API.20tests if you'd like to continue talking there.

It's strange, the API test you wrote seems to run just fine for me in Dockerized Dataverse. To get set up I'm following the steps laid out at http://preview.guides.gdcc.io/en/develop/container/dev-usage.html

```
mvn -Pct clean package
mvn -Pct docker:run
scripts/dev/docker-final-setup.sh
```

Importantly, in that "final setup" script we run `setup-all.sh` with the `--insecure` flag which allows for API testing. (See also https://guides.dataverse.org/en/5.13/developers/testing.html#getting-set-up-to-run-rest-assured-tests )

Anyway, below is the failure I'm seeing. While I'm on this branch I'll see if I can fix it quick:

```
[ERROR]   FilesIT.testDeleteFile:1997 JSON path data.files[0].dataFile.filename doesn't match.
Expected: orcid_16x16.png
  Actual: cc0.png
```

Here's the full test run: [testrun.txt](https://github.com/IQSS/dataverse/files/11091303/testrun.txt)

To run just your "delete file" test I used this:

```
mvn test -Dtest=FilesIT#testDeleteFile
```

ErykKul:
@pdurbin 
"filename doesn't match" was fixed in my last commit - I was testing the wrong version of the dataset (1.0 i.s.o. draft).

pdurbin:
@ErykKul great. Looking good! I also added a way to check the filename with some weird REST Assured syntax. I think we're good now. Back to QA! Thanks!

pdurbin (approved):
Ready for more QA.

PR #9388, opened 2023-02-14
IQSS/9387 - Support protected system metadata

### LINKED ISSUE #9387: Feature Request/Idea: Support protected metadata managed by external applications
There are several related use cases in the community for metadata that would be managed by a third-party application, e.g.
- DANS Data Stations where the archiving system wants to store metadata about the archival process/copy in a custom metadata block. The metadata can be shown but edits by humans or other processes would break archiving.
- Workflow (more hypothetical) - where an external workflow system wants to record information about the processing environment and/or how specific files in the dataset are used (input data, output data, etc.), again where changes by humans or other untrusted systems would break things and destroy reproducibility.

The proposed solution is a setting that can associate a metadatablock with a secret key, which has two effects:

- that block is not shown in the edit metadata display
- all API calls that would allow changes to that block must include a parameter with that key if there are changes being made to that block.

Concept and proposed implementation have been discussed in Tech Hours. Draft PR to follow.

qqmyers:
**What this PR does / why we need it**: Provides a mechanism to restrict metadata edits in specified blocks to those who know the secret key for that block. API only in this PR. See issue for use cases.

Closes #9387

Implementation was discussed in tech hours. As part of the work I created some utility methods to determine which blocks have changes when an update is made. These should be more efficient that keeping track of all the individual fields that have changed. I'm not sure if these could be useful elsewhere, e.g. in the version table where indicating just which blocks have changes might be a cheaper option when there are many versions, etc.

Also - 

- this PR removes the unused AbstractDatasetCommand.registrationRequired flag. (I new harvested flag was needed to avoid checking harvested datasets for system metadata, which led me to check on the use of the registrationRequired flag. So removing it is cleanup and avoids having more variant constructors with two optional flags.)
- this PR fixes some getOrCreateEditVersion logic in the semantic api calls. Prior to this PR, getOrCreateEditVersion was called before a check was made to see if a draft existed, which made that check always true. This is completely unrelated to the system metadata and could be a separate PR if needed. (It is local to those specific calls).

DANS has been testing this PR and @PaulBoon can probably provide more details. In essence, adding a key for a given metadata block should make that block not appear in the edit metadata display and cause all API calls for creating/updating a dataset/version where metadata in this block is changed should now require the extra key parameter for changes to succeed. (see the doc changes). All other blocks and functionality should be unaffected.

included

**Additional documentation**:
included


PaulBoon:
About setting up the testing, activating the feature, for example set '1234' on the citation metadata block. 
I managed this in two ways, the first one is via a properties file. 
- Create META-INF/microprofile-config.properties file in the exploded war and to add the required setting, e.g.
`dataverse.metadata. block-system-metadata-keys={“citation”:”1234”}`.
The second and preferred one is via the as admin command. 
- `/usr/local/payara5/glassfish/bin/asadmin set-config-property --propertyName=dataverse.metadata.block-system-metadata-keys --propertyValue='{"citation":"1234"}' --source=domain --sourceName=domain1
`
  This is persisted in the domain.xml file as: 
` <property name="payara.microprofile.dataverse.metadata.block-system-metadata-keys" value="{&quot;citation&quot;:&quot;1234&quot;}"></property>`. 
 @poikilotherm Not sure why it won’t persists if I use application and Dataverse as the source and sourceName, but I had to use the domain. Also we probably have to secure it like we do with the database password. 

Removing is done by editing the properties file or with `delete-config-property`.

poikilotherm:
Hey @PaulBoon sorry to say this, but it looks like you're doing this in a much more complicated way than necessary.

1. You should not edit values in `META-INF/microprofile-config.properties` from the WAR. This is solely meant to carry defaults added at build time.
2. I very much believe you don't need to go for these type of properties. Simply use "system properties" (also called JVM options), e. g. using `asadmin create-system-properties` (when going for JVM options via `asadmin` don't forget the `-D`, which makes it a system property).
3. Or, for testing / in containers: you can opt for setting an environment variable `DATAVERSE_METADATA_BLOCK_SYSTEM_METADATA_KEYS`

src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -40,3 +40,6 @@ dataverse.oai.server.maxsets=100
 # the OAI repository name, as shown by the Identify verb,
 # can be customized via the setting below:
 #dataverse.oai.server.repositoryname=
+
+# METADATA SETTINGS
+#dataverse.metadata.block-system-metadata-keys
```
poikilotherm, line 45: Not sure why you add this here, IMHO we should only add things that we have a default for. (Yes I am aware that there is a similar thing for OAI above - yet it looks like people start to mistake this file as editable for configuration, which is not how it's meant to be used)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -212,4 +221,28 @@ public void setDataset(Dataset dataset) {
...
+
+    protected void checkSystemMetadataKeyIfNeeded(DatasetVersion newVersion, DatasetVersion persistedVersion) throws IllegalCommandException {
+        Set<MetadataBlock> changedMDBs = DatasetVersionDifference.getBlocksWithChanges(newVersion, persistedVersion);
+        changedMDBs.forEach(mdb -> {
+            logger.fine(mdb.getName() + " has been changed");
+        });
+        
+        String smdbString = JvmSettings.METADATA_BLOCK_SYSTEM_METADATA_KEYS.lookupOptional().orElse(null);
+        if (smdbString != null) {
+            JsonObject systemMetadataBlocks = JsonUtil.getJsonObject(smdbString);
```
poikilotherm, line 233: How about adding a converter instead of fiddling with the String itself? 

It's not very hard to do and makes code here much nicer to read and write. :wink: https://download.eclipse.org/microprofile/microprofile-config-3.0/microprofile-config-spec-3.0.html#_adding_custom_converters

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -212,4 +221,28 @@ public void setDataset(Dataset dataset) {
...
+            HttpServletRequest httpServletRequest = getRequest().getHttpServletRequest();
+            if (httpServletRequest != null) {
+                String mdKey = httpServletRequest.getParameter(MetadataBlockServiceBean.SYSTEM_MD_KEY);
```
poikilotherm, line 236: Wouldn't it make sense to move the parameter extraction to `DataverseRequest`, adding a method there and avoid a bit more if/else stuff by returning an `Optional<>`?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -212,4 +221,28 @@ public void setDataset(Dataset dataset) {
...
+                for (MetadataBlock mdb : changedMDBs) {
+                    if (systemMetadataBlocks.containsKey(mdb.getName())) {
```
poikilotherm, line 238: A thought: it has advantages to use a JSON object here, yes. What would also work: create a `JvmSetting` with a placeholder at the end. Less complexity, no costly JSON parsing.

Asking for `dataverse.metadata.block-system-metadata-keys.citation` would return the key `String` if set for a certain block. You can access this variable setting via the same `JvmSettings.lookupOptional()`, but you need to pass the name of the block as the var arg.

PaulBoon:
I tested the feature and could not find errors, but it seems better if someone else also takes a look at it. 

Prepare the JSON to create dataset with citation (need it by default) and another metadata block. 
Note that you can create a dataset via the web form and then export it to JSON that you can edit before using it with the API. 
When creating or editing add the key as parameter; `mdkey=1234`. 
For example a successful create with curl could look like this:
`curl -k -H X-Dataverse-key:$API_TOKEN -X POST "$SERVER_URL/api/dataverses/$PARENT/datasets?mdkey=1234" --upload-file dataset-test3.json -H 'Content-type:application/json'`

Test scenario: 

Attempt to create dataset or edit metadata

- without a protected metadata block and without key
  result: succes
- with protected metadata block and
  - without key
    result: fail
  - with wrong key
    result: fail
  - with correct key
    result: success

Creation and editing via the web form should also be tested; protected metadata should not be editable.

PaulBoon:
For configuring a deployment (install) we would prefer the `create-jvm-options`, because that is how all the others are done. 
For example: 
`./asadmin create-jvm-options '\-Ddataverse.metadata.block-system-metadata-keys=\{\"citation\"\:\"1234\"\}'`

So, no microprofiling needed if I am correct @poikilotherm ?

PaulBoon:
Could be interesting to see if we can do the config with `dataverse.metadata.block-system-metadata-keys.citation`, preventing JSON parsing. This would also be more in line with specifying the keys using parameters like: `mdkey.citation=1234`. Allowing to add more key's for API calls that 'touch' several protected blocks. 
Would it be feasible to try this @qqmyers

PaulBoon:
Juste tested the new implementation and it works fine. 
Using the jvm-options with an alias for the key and passing the key for a specific metadata block works as advocated.

PaulBoon:
Finally tested it with our Republish workflow that fills metadata fields in a protected block. This also works. 
However I discovered that the logs contain the secret key!
```

[#|2023-03-13T17:04:29.269+0100|INFO|Payara 5.2022.3|edu.harvard.iq.dataverse.engine.command.impl.AbstractDatasetCommand|_ThreadID=74;_ThreadName=http-thread-pool::ht
tp-listener-1(3);_TimeMillis=1678723469269;_LevelValue=800;|
  Found key: 1234|#]

[#|2023-03-13T17:04:29.269+0100|INFO|Payara 5.2022.3|edu.harvard.iq.dataverse.engine.command.impl.AbstractDatasetCommand|_ThreadID=74;_ThreadName=http-thread-pool::ht
tp-listener-1(3);_TimeMillis=1678723469269;_LevelValue=800;|
  Found supplied key: 123456|#]
```

Both the real key and the supplied key, in this test there was mismatch. 
@qqmyers This seems nice when debugging, but is also possess thread when a malicious actor can read the logs of a log fragment is somehow available with this in it.

PaulBoon (approved):
LGTM

landreev (approved):
OK, this was already heavily reviewed, by @poikilotherm and @PaulBoon, with all the feedback having been addressed. So mine is a review of the reviews for the most part. Everything appears to make sense, so I'm marking it ready for QA to be included in v5.14.

kcondon:
Issues found:
1. Need to restart payara for system metadata block setting to be enforced (just an observation)
2. Templates feature continues allowing editing system metadata block fields but cannot create a dataset using that template, throws ui error and server log error saying needs key.
3. Can create a dataset using native api that contains system metadata block entries without a key, see all fields example.

qqmyers:
1. Need to restart payara for system metadata block setting to be enforced using example (just an observation, not sure whether all mp sources need restart)
- noted in docs
2. Templates feature continues allowing editing system metadata block fields but cannot create a dataset using that template, throws ui error and server log error saying needs key.
- fixed, protected blocks won't show in template now (FWIW: I don't see an api call to create a template yet, so nothing to fix there)
3. Can create a dataset using native api that contains system metadata block entries without a key, see all fields example.
- fixed
4. Making citation block a system block comes with some constraints: a. cannot use UI to create or edit a dataset due to required citation block fields on save and b. cannot edit any other block in ui or api without key due to automatic citation block update on save field like deposit date.
- noted in docs
5. Unclear how to reverse/undo the system metadatablock with the example provided. Maybe make it a bit more explicit -in the case of this example, asadmin delete-password-alias dataverse.metadata.block-system-metadata-keys.codemeta , then restart payara for it to take effect.
-  added example to docs
6. The semantic api example should state it is changing the title metadata and add ?replace=true to the command, otherwise it fails against an existing dataset already containing a title with an unclear message: status":"ERROR","code":400,"message":"Bad Request.
- changed in docs
7. The semantic api example shows two command line methods of passing the block key. The header example using -H fails but the second example, passing as an argument ?md.key, works.
- I found the opposite - the header worked and query param didn't. Investigating, I found that the query param is case sensitive, so I noted that in the docs and changed the examples to use the same case everywhere (fwiw: also noticed that the codemeta block is actually named codeMeta20 so I changed the example to that)
8. Was not able to get the native api edit metadata endpoint to work using the example in the guide so could not confirm it enforces system block. However, the update metadata endpoint does enforce it.
- can't reproduce this - possible an issue with block name and/or case sensitivity?

PR #9399, opened 2023-02-20
#9398 - fixing issues with metadata language inherited option label

### LINKED ISSUE #9398: Editing collection : Metadata Language inherited option label is broken
Hello Dataverse Team,

Issue has been found using Metadata Languages setting in a Dataverse Collection :
If a Metadata Language has been set in a collection, when editing, the last option `<Title of language> (inherited from enclosing Dataverse)` is in most cases wrong because it uses the current selected language of the collection itself instead of one parent.
It will work if : One of the parent have the same Metadata Language selected
It will not work if : No parents have Metadata Language selected or one of the parent have a different Metadata Language selected

PR will follow.

Regards.

luddaniel:

This PR fixes some issues with the label of the inherited option label while editing a collection : 
- when no parents have Metadata Language selected it will now show default language key `dataverse.metadatalanguage.setatdatasetcreation`
- when one of the parent have a different Metadata Language selected it will now show the language of the inherited language
- it will no longer display the current select language as inherited.

Closes #9398

This PR was blind coded (I'm not in capacity to do more for now sorry)

I would say yes.


qqmyers (approved):
Good catch & fix. The change looks good to me. 

The original code would work if the collection hasn't had a selection, but not once a non-default pick has been made. The PR fixes that by looking at the parent's effectiveMetadataLanguage instead.

I would suggest reverting the comment or extending it, e.g. - 'If it's undefined, no parent has a metadata language defined, and the global default should be used.'

For testing, I think the key thing is to configure for using metadata language, set the language on a subcollection to the non default value, and then verify when coming back to edit again that the default option is correct (and not showing the current value for the collection).

kcondon:
@luddaniel I'm seeing a stack trace when I attempt to view general information for my root collection:
[metadata_lang_view_collect_err.txt](https://github.com/IQSS/dataverse/files/11579743/metadata_lang_view_collect_err.txt)

luddaniel:
Thanks to your comments, I was able to quickly fix the java code. Once again, this is untested code but it should do the trick. 
Best regards

kcondon:
@luddaniel I can edit root general properties now but as Jim points out, if I have selected a metadata language for root there should be no "inherited from" entry since there is no parent collection to root.

luddaniel:
Hello @kcondon and @qqmyers, I spend some time to re-dive on the subject and I tested the code :D Now this look good, I hope it will please you. Sorry, I didn't have the resources to test before.
Best Regards

kcondon:
[like]  Condon, Kevin M reacted to your message:
________________________________

PR #9402, opened 2023-02-21
Iqss/9150 handle fundreg reqs for ext cvv

### LINKED ISSUE #9150: Create a javascript for the frontend that supports Fundref
Definition of Done:
- [ ] front end is created
- [ ] Code, javascript and a test metadatablock in the other repo

(i.e. not production )

### LINKED ISSUE #9498: CVOC : Searching External Vocab Values and compound fields
Hello there,

I wanted to report that CVOC values indexing does not work properly when you are using external vocab support like skosmos and you are using compound fields.
You will not be able to perform Dataverse Search on child fields external value such as the skosmos term in another language.
You will only be able to search DatasetField direct value (not the variety of a term from a skosmos instance).
When you are using external vocab support with skosmos on a single field, indexing of skosmos values will work.

Use SKOSMOS CVOC with compound fields, save the dataset and try to search for other languages. You cannot find the dataset using other languages for the term.

Using SKOSMOS CVOC with compound fields

Search page

v5.12.1

qqmyers:
**What this PR does / why we need it**: This PR makes changes/extensions to the external vocabulary support mechanism to address requirements from the Crossref Funder registry use case. These include:

- [x] - allowing a script to only manage the one term-uri-field child
- [x]   - when defining an expanded value to be included in the json and ore exports
- [x]   - in the metadata display
- [x]   - in the search response highlights
- [x] - supporting retrieval filtering params to specify a path including all array elements rather than a single one

Closes #9150
Closes #9498

Use the new fundreg and/or ror example external vocab configs and related Javascripts at https://github.com/gdcc/dataverse-external-vocab-support/tree/fundregror and verify that you can add CrossRef fundreg entries in the citation block Funder (Grant Agency) metadata field or an ROR as the author affiliation. For regression testing, one could run the older cvoc example metadatablock and the orcid/skosmos scripts there and verify they still work (skosmos requires that "skosmos.dev.finto.fi" be replaced by "demo.skosmos.org" everywhere in the config). 

Fundreg/ROR should work as with other external vocabs in that they should show correctly in the metadata display, search results, advanced search. Fundreg should also provide an expandedValue entry in the json metadata export that includes the name of the funding org (along with the URI as the "value").

**Additional documentation**:


mreekie:
grooming:
- Looks like this will bleed over to March 15 sprint.
- @qqmyers Thanks for the sizing estimate.

mreekie:
Prio review:
- After the technical review of the code
- Needs to be installed on Demo.
- At that point Julian and possibly others will review the User Experience.
- The intent is that feedback that comes out of the UEX review will spin out additional issues as needed.

@kcondon @jggautier collaborate on the details of how/where the UI testing gets done.

luddaniel:
Hello there, as you are working on CVOC / ROR for v5.14, I just submitted an issue #9498 with a PR, maybe you can have a look in case it's a quick win

kcondon:
@qqmyers @jggautier Is there a simple, brief description of what this does from a user's perspective or do I need to do a deep dive into those "closes" issues? This seems like a tip of the iceberg kind of pr.

qqmyers:
@kcondon - I added some details on testing in the description. Basically this PR does nothing on its own, but it allows you to run the new Fundreg and ROR external vocab config/javascripts that are in a branch/PR in the gdcc external vocab repo. Nominally the setup for that is the same as for the older ORCID/Skosmos ones except that the example configs for fundreg/ror apply them to fields in the citation block rather than a new example block.

qqmyers:
1) There are a few possibilities. The issue is because I refactored so that the main scripts now use a common utility script and the main scripts handle loading it, so that it is only loaded when needed. Options include:

- Document where the scripts have to be placed
- Hardcode so that the main scripts use the util script from the gdcc github.io page (if you local install, you're not getting everything local
- Add the util script to always be added by Dataverse
- Use the setting created by EryK to add the util script manually (may load when not needed, but not all the time)
- Adjust the config mechanism to allow multiple scripts to be required (probably best, not a one-line update)
I'll plan on implementing the last one.

2) Working on it - looks like it is on the script side rather than the Dataverse PR

3) I've made a change (in the Dataverse PR) to help with this.

4) As far as I can tell, this is just from delays at those services. We could tweak the minimum characters to search for (currently 3) or the delay to wait for new chars (so we don't send repeating requests to the service for your 3,4,5,6, ... char strings) - currently 500ms to see if that helps. Both are in the scripts themselves rather that the Dataverse PR.

qqmyers:
5) As far as I can tell, this is a known issue for select2 which will be fixed when jquery 3.7 comes out - see the end of https://github.com/select2/select2/issues/5993#issuecomment-800927000. There is a work-around we could try, but given that it's minor, I'd suggest waiting.

6) This is due to the poor performance of the Fundreg service. They do threaten to throttle so perhaps many requests from testing is causing the slow down. FWIW, I've done what they suggest to get on their faster server (basically providing contact info in the requests themselves) and, for the latest ~100 values you use, I cache the results so things should be faster after the first access as we don't go back to Fundreg for them. It may be worth talking with them (as GREI?) and/or looking into whether we can run a local service (I forget, Fundreg or ROR suggested that, but it may be possible for both).

7) This is a current limitation of the external vocab mechanism that probably could be addressed as a new issue. For now, advanced search, and use in facets should be possible. Facets might be the best way to quickly find things funded by one agency. For simple search, there could still be problems with the stemming and other settings we have on the underlying field, i.e. "Institutes" would match "Institute" as well.

8) I wasn't able to reproduce this on my test server - the oai_datacite was produced when I had one ror affiliation for author and two or three fundreg entries, using controlled and free-text entries.

9) This is ~by design in that the ORE and json outputs are intended to be complete and the other exports then have to be updated if/when someone determines whether those formats could/should display something different (the human readable form, the i18n form, both with the identifier as an attribute, etc.). The raw info should be available to all the exporters through the json they receive so further changes can be done in new issues/by someone familiar with a given exporter.

10) I haven't looked into this but I assume not seeing something in the DDI HTML is a bug (since it is supposed to mirror the DDI XML?) but perhaps just a missing enhancement if that was added to the DDI XML recently.

J2) Unlike ORCID, Fundreg doesn't provide much to help with prioritizing (that I've found anyway). I have added support to search not only by name but by the acronym and tags that they provide. These help the query 'NIH' bring the National Institutes of Health (which doesn't contain NIH) into the search results. I also raise the priority of anything you've chosen recently (last ~100 choices), so NIH should be first once you've used it once. There's code that prioritizes by prior use, then acronym, then tags (various parts of NIH add that tag even though they are one institute within NIH. The relative priorities there could be changed, or perhaps weights could be added. A more useful thing that only Crossref/a service could do would be to prioritize results based on popularity (either hits or the number of existing resources that cite that funder, etc.).

J3 - see 7

J4 - this is in the URL? If so, yes, the advanced search and facets use the id when searching rather than the name (or international variants) for the back end, but all the places in the UI should be showing the name (again, from 6 these may show briefly as the id until the lookup can complete (against the server the first time or more quickly from browser memory after that).

J5 -the popup shows the alternate names and acronyms that Crossref provides for the given id. Thus while the facet shows National Institutes of Health, which is the official name, you may have searched for NIH to add it and the popup will show that value. Whether there's a popup and what it shows (other names or just acronym) is controlled by the javascript in the external vocab repo and changes could be made there. There may be some limitations if we want to just exclude some values, i.e. I'm not sure that alternate names are distinguished from i18n values when Fundreg returns them.

J6 - see 9. I'd definitely suggest changes to other formats should be separate issues.

qqmyers:
The external vocab mechanism tells the script whether an item needs to be displayed, or an input/selection UI is needed. What the script does for each of those is up to the script, but, on the display side, there isn't currently any distinction for display in different parts of the UI, i.e. in the metadata pane, a facet, as a search hit, etc. A couple of the example scripts do use a popup - I think ORCID is the other one - it will show the email of the person (if their email is public at ORCID). For internal text/CVV fields, there is no equivalent mechanism to customize the display (just the basic formatting in the metadata block). If changes are desired, if the change applies to all places a term is displayed, the script can be updated. If a general mechanism is needed to make facet display different from display elsewhere, it would be an extension of the mechanism itself and would involve changes to Dataverse and scripts and the config mechanism, schema, etc.

qqmyers:
1) Most of these were real, which is good news. Turns out simple search was enabled and there were parsing errors. I was able to fix those and now simple search works. I also turned a couple remain logger.info()s into fine()s.
2) The code was not ignoring free-text entries and would ask the servers about those - now fixed.
3) I added the @id to the ORE export - it should be complete like the json export. For other formats, I'd suggest separate issues for any updates.
4) The Dataverse code stores whatever the config tells it to pull from the server's response. In the case of fundreg, I believe it puts multiple English variants in the same structure with other language variants. We currently have the capability to store none of that or all of it, but can't filter to just get other language terms (assuming we don't actually want the English name variations to be sent.) I'd suggest a separate issue(s) to figure out what's desired here given what fundreg returns and the variability between their entries and then implement whatever change is desired. (Any given installation can remove what's there now by editing the example configuration in the repo.)
5) This was a cut/paste type in the ror script. When the value was a string instead of an id, the script added a blank between the () chars instead of the string. Now fixed in the ror.js script.
E2, E3) The blank values are probably due to the fundreg server being slow - the script currently displays nothing until the call to fundreg returns. It is probably possible to add in a spinner or temporary value by editing the script. I'd suggest a separate issue here just because that would all be in the javascript in the external vocab repo, versus changes to Dataverse. For the scaling, it is a ~known issue with the select2 widget that it doesn't resize when the page is dynamically resized. Refreshing on a resized page will update the widget. I believe when I looked before this was something that was reported to the select2 devs but I'm not sure. In any case, I think it would be hard for us to fix (could be wrong) unless there's an update. An alternative would be to move to use an alternate widget in any/all scripts. The original community ror script, for example, just put an icon on the page and if you clicked it would popup a separate dialog to let you find the org you wanted. That, or other design could be done by changing the script - again an issue for the gdcc ext-vocab repo. I did also notice a static issue with narrow pages - as reported in other images, the text can appear below the field input. I verified on my machine that with the current css that only happens when the screen is narrow enough that other non-external-cvoc fields are also messed up  (I think it was an internal CVV field where the drop-down control slid under the adjacent child field.) In any case, the static issue could be fixed by providing different css values depending on the screen width as is done in other places. That would be a Dataverse issue rather than one for the gdcc ext-vocab repo.
E7 - as noted above, simple search is now working.
E10 - I assume this is true regardless of whether the field is associated with an external-vocab? If so, it should probably be a separate issue (like the existing issue about removing the funder contributor type which isn't external vocab related).

kcondon:
Latest testing shows all major issues resolved. It was decided to merge and perform any ux later after spa.

PR #9406, opened 2023-02-23
#9404 Added Harvesting key to customize facet label

### LINKED ISSUE #9404: metadata source : Harvesting facet label is not internationalized
_Issue created by the "entrepot.recherche.data.gouv.fr" team_

Have data harvested

* When does this issue occur?

When the Harvested facet label is translated

Pages with facets are displayed, with harvested datasets

* What happens?

The default "Harvested" label is displayed for all languages

All users

Have the "Harvested" key in the Bundle property file

5.12.1

**Screenshots:**

Capture_metadata_source_harvested

stevenferey:

Adds the ability to customize the Harvesting label

- closes #9404

No

**Additional documentation**:


qqmyers:
Thanks for the PR, but I think more will be needed to fix this. 

I think what's needed is for the specific code writing the word Harvested to actually use the Bundle entry. Looking quickly in the code, the one place I see a static "Harvested" defined is https://github.com/IQSS/dataverse/blob/1a797171cdb73741b5da4a683f38697558349b5c/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java#L159 and I'm guessing that the code using that is what would have to change. (I haven't verified that this is truly the source of that word in the facet, but I don't see any other possibilities so far. ) Also - seeing that there are other constant English strings defined right before that line, I wonder if there are other search facets that are hardcoded as well.

I also see that there's already a key for the word Harvested in the Bundle.properties file: https://github.com/IQSS/dataverse/blob/1a797171cdb73741b5da4a683f38697558349b5c/src/main/java/propertyFiles/Bundle.properties#L69
I don't know where that is used but unless there's a reason to translate differently in that case, we should be able to reuse that entry,

stevenferey:
Thank you for your reply,

There are indeed other Strings declared in the file. They are declared in the Bundle.properties, for example, "In Review":

https://github.com/IQSS/dataverse/blob/1a797171cdb73741b5da4a683f38697558349b5c/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java#L157

https://github.com/IQSS/dataverse/blob/1a797171cdb73741b5da4a683f38697558349b5c/src/main/java/propertyFiles/Bundle.properties#L2676

We translated it into French for our Dataverse in our Bundle_fr.properties file:

`In\u0020Review=En r\u00E9vision`

With the following result:

Capture_preprod_01

For the Harvested key, we also translated the key with the following result:

`Harvested=Harvested`

Capture_preprod_02_EN

`Harvested=Moissonn\u00E9`

Capture_preprod_02_FR

I specify that the new key is declared with a capital letter. 

The lowercase "harvested" key is used here:

Capture_preprod_04_EN


Steven.

qqmyers:
OK - cool. I guess the connection between the Harvested variable and the bundle is ~hidden because the solr field is created and the call to the bundle is using getValue(). In any case, it's working, so a good PR. Thanks again, and for the clear explanation.

qqmyers (approved):
From the discussion, this is working. The only questions I can see is whether this should go with the ~related entries at https://github.com/IQSS/dataverse/blob/1a797171cdb73741b5da4a683f38697558349b5c/src/main/java/propertyFiles/Bundle.properties#L2673-L2677 or if it is worth digging to avoid having keys for harvested and Harvested. 
With SPA changes coming, which will change the whole bundle mechanism, perhaps not.

pdurbin:
At first I was wondering how simply adding `Harvested=Harvested` to Bundle.properties could work but then I found all my todos from many, many years ago to get rid of the reflection I added. 😬 😅 

I agree this should probably work. I'll give this a size of "3" because it's a one line change and should be pretty easy to verify.

Thanks for the PR @stevenferey!

kcondon:
@stevenferey This branch needs to be refreshed from develop branch since there was a flyway db script that was renamed. I cannot deploy this as this is, even though the only change is a text file change.

kcondon:
@stevenferey Thanks! It works great!

PR #9409, opened 2023-02-24
9361 storage quotas

### LINKED ISSUE #9361: Add GENERIC storage quota check to the file upload framework.
Several issues have been opened for adding mechanisms for storage quotas. There is some overlap between them, but different kinds of quotas have been requested - per-user, per-dataset, per-collection, per-user-per-day. Some appear to be more urgent than the others. (case in point, the most recent one, #8549, opened by me, is a specific request from our curation team that we wanted to address soon). 
- #8549 
- #7829
- #4339 
- #3939

(there may be a couple more that are complete duplicates of something requested by one or more of the issues above, those are omitted).  

Each one of these issues does require the actual quota check to be present somewhere in the workflow of uploading and creating a file. So this new issue is just for adding that mechanism. Perhaps it can be tested against a single setting that defines the limit for ALL uploads, but once it's there it should be extendable to add more types of quota checks to satisfy the specific cases and scenarios requested in the issues above. It should be programmatically configurable, and it should be enforced everywhere where the application allows a creation of a new file - via the GUI, the API and in the direct upload workflow. 

It can be as simple as slightly modifying the individual file size check that's already there; but it should probably be made more programmatic, etc.

landreev:
(please note the draft nature of the pr; work in progress, the model/approach need to be reviewed before I can produce a finalized version)

As explained in the issue, this PR will not result in a feature that's ready to be used in real life for enforcing meaningful storage quotas. What it does is add the quota check to the upload workflow; for testing of the concept it provides a mechanism where the same storage quota limit can be defined for all the users of the installation (and only the files directly created by the user are counted towards the quota). Again, this is not going to be useful as a practical solution, but will allow us to further expand the quota definition mechanism once the checks are there. This was meant to be a sensible sprint-or-so-sized chunk of work. The PR also moves some file unpacking and processing code that turns uploads into Datatverse Datafile objects from a static Utility method and into its own dedicated Command (so that files can eventually become subject to command engine-side metering too). 

Closes #9361

As stated elsewhere in the PR (see the long-ish comment from Mar 17, in addition to the description above), all that's implemented here is the added mechanisms in the upload pipeline that prevent a user from uploading more content once they exceed  the quota. The mechanism for defining the actual quota is essentially a placeholder, just enough to test these new hooks, but not yet usable pas a real life practical solution. 
So a test scenario would be to set an arbitrary limit via the `:StorageQuotaSizeInBytes` setting, and make sure you  can't upload more than that once the limit is reached. 
[Additionally:] The second setting, `:UseStorageQuotas` activates the enforcement of the quota (defaults to false)
I can answer any additional questions etc. 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -140,6 +142,36 @@ public class DataFileServiceBean implements java.io.Serializable {
      */
     public static final String MIME_TYPE_PACKAGE_FILE = "application/vnd.dataverse.file-package";
     
+    public class UserStorageQuota {
```
scolapasta, line 145: should we move this out to be its own class? (i.e. any reason for it to be an internal class here?)

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1657,4 +1689,29 @@ public Embargo findEmbargo(Long id) {
...
         return d.getEmbargo();
     }
+    
+    public Long getStorageUsageByCreator(AuthenticatedUser user) {
+        Query query = em.createQuery("SELECT SUM(o.filesize) FROM DataFile o WHERE o.creator.id=:creatorId");
+        
+        try {
+            Long totalSize = (Long)query.setParameter("creatorId", user.getId()).getSingleResult();
+            logger.info("total size for user: "+totalSize);
+            return totalSize == null ? 0L : totalSize;
```
scolapasta, line 1699: when would this be null (as opposed to the exception thrown)? is is where we for some reason don't have sizes stored?

src/main/java/edu/harvard/iq/dataverse/EditDatafilesPage.java
```diff
@@ -341,6 +348,18 @@ public boolean isUnlimitedUploadFileSize() {
...
+    
+    public String getHumanMaxTotalUploadSizeInBytes() {
+        return FileSizeChecker.bytesToHumanReadable(maxTotalUploadSizeInBytes);
+    }
+    
+    public boolean isStorageQuotaEnforced() {
+        return maxTotalUploadSizeInBytes != null; 
+    }
 
     public Long getMaxIngestSizeInBytes() {
```
scolapasta, line 364: should we reconsider naming this? (since it does more now?)

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1657,4 +1689,29 @@ public Embargo findEmbargo(Long id) {
         DataFile d = find(id);
         return d.getEmbargo();
     }
+    
+    public Long getStorageUsageByCreator(AuthenticatedUser user) {
+        Query query = em.createQuery("SELECT SUM(o.filesize) FROM DataFile o WHERE o.creator.id=:creatorId");
```
scolapasta, line 1694: this makes me think (as you point out) how we decide to define the limits even? by user, by dataset - for example a workaround around this limit would be to create a new account and give it permission?)

src/main/webapp/editFilesFragment.xhtml
```diff
@@ -155,6 +160,7 @@
                                   fileLimit="#{EditDatafilesPage.getMaxNumberOfFiles()}" 
                                   invalidSizeMessage="#{bundle['file.edit.error.file_exceeds_limit']}" 
                                   sequential="true"
+                                  previewWidth="-1"
```
scolapasta, line 163: what does this do?

scolapasta (commented):
Overall, this looks great - I did leave a few questions as comments, most not too important. The biggest is how we actually want to define quotas which leads to what the next steps are? We should discuss that and then add the summary here. (e.g. one example is to go and test this as is, and then add user specific (if that's the direction we go) as a new issue / PR; if so we should also add some integration tests here).

src/main/webapp/editFilesFragment.xhtml
landreev: `TL;DR` it disables the browser-side thumbnail previews. 

I was delighted to discover that, as of a couple of recent versions of PrimeFaces, it is now possible to disable their own image thumbnail preview in the upload component. It used to drive me nuts - we don't really need it (because we generate our own thumbnails server-side); but, unlike with our thumbnails, there's no way to define a size limit, so attempting to upload a bunch of large enough images paralyzes the user's browser (since it's all javascript). And, adding insult to injury, the user assumes that it's our site that's crapping the bed... I am embarrassed to acknowledge that, despite all that righteous indignation, I somehow missed the point at which they had added this ability to disable the whole thing, by setting the previewWidth to -1.

mreekie:
grooming
 - It appears this will bleed over from March 1 sprint to March 15th

mreekie:
> grooming
> 
> * It appears this will bleed over from March 1 sprint to March 15th

Leonid noted no change in size for the march 15th sprint

landreev:
> Overall, this looks great - I did leave a few questions as comments, most not too important. The biggest is how we actually want to define quotas which leads to what the next steps are? We should discuss that and then add the summary here. (e.g. one example is to go and test this as is, and then add user specific (if that's the direction we go) as a new issue / PR; if so we should also add some integration tests here).

combining this with another question: 

> ... how we decide to define the limits even? 

The linked issue #9361 lists 4 other issues in which specific types of quota definitions were requested. Specifically, per collection, per dataset, per user, and even per-user-per-day. Per group has never been requested, I believe - not sure if it's needed (per dataset/collection  may be enough to cover a practical scenario of a group of users, a research project/department etc., working together depositing data (?). 

I'm assuming the answer is that we want all of the above. And in each case the answer - how much this user can add to the storage during this specific upload session - will be based on checking all possible quotas that may apply. Somewhat similar to how our permissions work - we check for permissions via direct role assignments, combine with group perms ... etc. Although I **very much hope** this does not need to be as complex as our permission structure, with all the complicated recursive db queries there. It may be possible to keep it relatively simple because, unlike with permissions, we probably don't need/want to apply _the most generous_ of the possibilities. I mean, you can access a restricted file with _either_ a group or a user permission, whichever you may have. With the quotas, however, I'm assuming we can get away with having a clear precedent structure. For example, the quota on the dvobject (dataset, collection), if defined, beats everything else (if not defined, check if the parent has the quota, repeat). If none are defined on the dataset/collection, then the user's own quota applies (both total and per day, possibly), if that's not defined either, then check if the installation has the default user quota and/or default dataset/collection size. (Something like that, just an example of a possible scheme). 

This pr was an attempt to create a sprint-sized amount of work that could be wrapped up and merged independently. We don't need to know how the quotas are going to be defined. We just assume that we can count on a service that will give us these numbers - how many bytes this user can add during this upload session, to this dataset (plus the total applicable quota, for display and/or error messages). All it tests is the behavior of the upload page and the api, with only enough service-side code provided as to be able to test against _some_ defined limit. 

As for the next step, I don't think the 4 linked issues can be worked on separately. I.e., it doesn't seem practical to try to add a per-user quota separately from per-collection (not 100% sure though). But this is a natural next step, to actually work on the real, functional quota service. 

Though there is another task that may be standalone enough - we need to work out some mechanism for counting used storage space. I don't think we can practically obtain this number in real time every time. Going through the physical storage every time is obviously impractical. Calculating it from the database (by adding the filesizes + the saved originals) is not truly accurate (skips many other aux. files), although it may be a reasonable approximation... but still inefficient, to calculate it in real time every time. So I'm assuming we'll need some system of keeping the sizes of datasets and collections cached... Figuring this out and implementing it is definitely a project of its own.

mreekie:
new sprint:
- no change to size.
- was waiting. needed additional input from folks who were out.

landreev:
Per conversation with @scolapasta last week, we decided it made sense to send the PR through the normal QA pipeline and merge it in its current form, before proceeding with the next steps of the overall effort. I will remove the "draft" designation.

scolapasta (approved):

landreev:
Sorry, my "how to test" instructions were incomplete - there are 2 settings, one to define the quota size, the other to activate the enforcement. My bad, will update. 
I'm getting the error messages trying to upload over the quota, but noticing the byte numbers being weirdly formatted/messed up. Will fix and move back.

landreev:
There is also a merge conflict in the branch now that's not immediately trivial to resolve, this will have to wait till tomorrow.

kcondon:
Issues found:
1. Can no longer upload files on dataset creation, upload button is grayed out.

kcondon:
Issues found:
1. Can't upload a file with quota enabled:
Screen Shot 2023-05-01 at 1 43 41 PM

kcondon:
@landreev I guess I would put 5. under "unexpected behavior". So, unless we actively want that then it is a bug. Happy if you want to raise it in tech hours or tech channel. It seemed similar to the other issues.

landreev:
I checked in a few fixes. 
I'm not sure about .6 either. It may have been a real permission issue. Or a fluke, or you may have found some behavior that's actually a bug, but that would be outside of this PR, I think. 
To reiterate, this not a ready feature. The goal of the PR is to test the behavior, in the UI etc. of enforcing quotas. But we will need to add a more flexible/useful system of defining different quotas, for users and/or datasets/collections/etc.etc.

landreev:
(I'll address the cosmetic issues, with the tabs/whitespaces in the code that the bot is complaining about as well)

landreev:
And, just to have it in writing here, I don't think we want to merge this PR  before 5.14 is released.

kcondon:
Testing is complete. Works as expected within the context of test framework. Holding off merging to post v5.14 as requested. Note for possible future check that if :UseStorageQuotas is enabled but :StorageQuotaSizeInBytes is not settling , creating a dataset throws exception.

cmbz:
2023/08/02: Discussed during Sprint Kickoff, there is no capacity to address this during this current sprint, but it should be attempted in 6.1.

landreev:
I don't think this PR is actually "ready for review" at this point. Just syncing it with develop appears to be a non-trivial task now that it has sat for so long. So I'll move it to "in progress" to work on that. 
But even after the branch is synced and builds under 6.0, I no longer feel like it was a good idea to try to merge the PR as it was in the first place. The PR was created as an experiment in merging incomplete/partially implemented pieces. But the whole idea was to merge such incremental parts quickly and keep working on the functionality; so it seems safe to conclude that the experiment failed.  I am wondering if we should just change the goal of the issue to delivering working/practically useful quota functionality and put the PR back in dev.? - idk, let's talk about this part.

kcondon:
@landreev I ended up doing a quick smoke test and found file upload to direct s3 fails: 
[file_upload_fail_direct_s3_quota_not_config.txt](https://github.com/IQSS/dataverse/files/12728785/file_upload_fail_direct_s3_quota_not_config.txt)

landreev:
Good catch! I will look into it. 

> @landreev I ended up doing a quick smoke test and found file upload to direct s3 fails: [file_upload_fail_direct_s3_quota_not_config.txt](https://github.com/IQSS/dataverse/files/12728785/file_upload_fail_direct_s3_quota_not_config.txt)
> 
> I did end up also checking storage quota and it seemed to work on local file and indirect s3 upload.

landreev:
Fixing the problem with direct upload now (once again, good catch!).

landreev:
Direct uploads should be working via the UI now.

PR #9456, opened 2023-03-20
Correct the regex that strips ZIP entry to shortname

janvanmansum:
There is a problem when uploading a ZIP-file that uses backslashes a file path separator. The file labels contain the whole path instead separating the directory path and file name.

N/a

The code in FileUtil gets shortname (base name is intended I think) by stripping everything up to the last slash. However, when getting the directoryLabel, it does take in to account backslashes as well, so I assume that backslashes are supposed to be supported but were overlooked in the line that gets shortname. 

No

No

**Additional documentation**:


PaulBoon:
@pdurbin It is a small change and even though it is not a common problem (old Powershell version on windows produced those zip files) it would be great if this is merged soon.

pdurbin:
@PaulBoon @janvanmansum I think what would help here are unit tests and/or filling in "how to test".

janvanmansum:
@pdurbin : it is hard to create a ZIP file to test this with. This was discovered with a ZIP file from a customer, that we cannot shared. However, I think it is obvious from the code itself that this is a simple slip-up. Compare with https://github.com/IQSS/dataverse/pull/9456/files#diff-4ed4534b8e05afee635f00ee208d526ecf3797776752e34a992e1834f1059963R1026 where correct pattern is already implemented for directories.

qqmyers:
@janvanmansum - this line has been moved to https://github.com/IQSS/dataverse/blob/dc8a03f84de148b080d6c4e97cf4f143827ed3a0/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateNewDataFilesCommand.java#L416 - can we get an updated PR?

W.r.t. testing - I think it would be enough if the core team just regression tests to make sure normal zip files aren't affected and you've tested with the non-standard zip you have. (I tried to search a bit and while there's info about the problem with old Powershell creating zips like this, I didn't see any examples to use.)

qqmyers (commented):
LGTM

jp-tosca:
I am not sure this should be merged, this is the result of my test, I created a zip file with the following files: 

image

This is how this file is uncompressed on Develop vs this branch: 

image

jp-tosca:
This is the file I created, let me know if this is helpful:
[Archive.zip](https://github.com/IQSS/dataverse/files/14947976/Archive.zip)

qqmyers:
@janvanmansum - We've run into an issue when @jp-tosca QA tested this. It looks like Macs allow \ in filenames (as in the test zip above - note my Windows machine flags it as having a virus, but that appears to be a false positive). That creates a couple problems: a file like `Filenames can be fun \o:name/filenames are fun \o:name.html\` gets ignored with this PR since the filename (everything after the last \) is empty, so it doesn't get saved. Also, a resource file like `__MACOSX/Filenames can be fun \o:name/._filenames are fun \o:name.html:` which used to get ignored as the filename starts with a . (`._filenames are fun \o:name.html:`) now gets added since it's filename looks like `o:name.html:` with the PR change. 

I'm sure this is a rare case, but hopefully corrupt zips using a \ separator are rare as well. At present, I don't see a good way to handle valid-but-rare Mac zips and the corrupt zip cases at the same time (at least not without a lot of work). Given that, and that it seems worse to potentially ignore some files and accidentally add resource files in valid zips versus just getting the path in the name for the corrupt case, I think we're leaning towards rejecting the PR unless you've got an idea of how to avoid these issues.

(Sorry for catching this late. @jp-tosca and I were noting that with our Windows backgrounds, we've just assumed that \ isn't a valid filename char.)

qqmyers:
A couple more notes - this could be something that could be added with a flag. Right now, the code isn't great for names with \ as, as was noted early on, the code to find the file name and the code to find the path differ, so if a \ exists, the current code doesn't produce a path + filename in Dataverse that equals the original. Maybe both treat \ the same way and you can turn on/off whether it is treated as a separator in zips?

janvanmansum:
@qqmyers , @jp-tosca I see your point. Closing the PR, hoping that the old ZIP file was a unique case.

Issue #8889, opened 2022-08-03
Feature Request: Enable file PIDs at Dataverse Collection Level


sbarbosadataverse:
The Harvard Dataverse Model for supporting file-level DOIs will support collection level instead of production-wide file-level PIDs.
This requires the ability to enable file-level PIDs at the collection level. Eventually, we may discuss enabling at dataset level, if needed.
-->

**Overview of the Feature Request**

**What kind of user is the feature intended for?**
SUPERUSER/INSTALLATION ADMIN

**What inspired the request?**
THE COST OF PROVIDING FILE LEVEL PIDs FOR GENERALIST, OPEN REPOSITORIES

**What existing behavior do you want changed?**
MANAGED FILE LEVEL PIDs distribution

**Any brand new behavior do you want to add to Dataverse?**

**Any related open or closed issues to this feature request?**
NO

@scolapasta

scolapasta:
Some notes:

We'll want to introduce this as a setting per dataverse collection, and model it after other such settings (metadablocks, facets, etc) in that a child dataverse would inherit the setting from its parent unless it chooses to override it.

For now this will be superuser only, so we can set via API (optionally, we could include a rudimentary UI way, though further design would be warranted if we opened it up to dataverse collection admins).

Additionally, we'll want to modify the current installation wide setting from just true / false to also allow for this option (so likely, off, on, and "set all collection level" - though we may not need "on" and just off and "set all collection level" - something to discuss when we start working on this).

We'll also need an API (or move the current admin only API) to register for existing datasets (for when a superuser turns this on in a dataverse collection with existing datasets).

mreekie:
Sizing:

What is needed is to assign XX to a specific collection
- all datasets automatically get a persistent identifier.
- Files - persistent identifier is option.
- Persistent-data identifiers for files is currently disabled in production.
- It's a global switch.

Sonia would like this to be turned on/off on a per-collection basis.
This feelis like it could be one sprint: Size 80

landreev:
BTW, @poikilotherm 's #9462 is already adding a per-collection (and inheritable) setting that's related to PID behavior. For versions PIDs. Can be used as a model for this setting.

scolapasta:
After a discussion with @qqmyers and reviewing the code and documentation, I'm adding this back to  the board. The issue is that the current implementation allows a user with edit dataverse permissions (e.g. Dataverse Admin) to set this despite it being turned off for the installation. 

[...]

PR #9462, opened 2023-03-22
4499 - dataset version pids

### LINKED ISSUE #4499: DOIs for Dataset versions
Following up on the thread on the google group:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/dataverse-community/34E9foKnxQs 

### Userstory 
As a researcher, I want to be able to cite, using a permanent identifier, a specific version of dataset to avoid any ambiguity and to make the citation machine-actionable. 

### How this should probably look

Zenodo would be a good template here.
1. A dataset has one "generic" DOI that always points to the latest version (this would be used if you e.g. just cite the data generically)
2. A dataset has one DOI per version to allow to point to a specific version
#### Example 
https://doi.org/10.5281/zenodo.1041767 is the generic dataset DOI, always pointing to the most recent version, with 10.5281/zenodo.1188752 being the DOI for the 2nd (current) version and 10.5281/zenodo.1041768 the DOI for the first version

#### File DOIs:
File DOIs are great (and by themselves necessary), but they are not a replacement for dataset version DOIs. Datasets are often made up of multiple files. An analysis script (which itself may be part of the data and thus versioned) may point to multiple files in a dataset. It's not feasible (or desirable) for a researcher to include the DOI for every file used in a citation. They want to point to one single DOI to reference the exact data (made up of multiple files) they've been using. This is equally true for quantiative and qualitative data, btw.

File DOIs of files that are not changed between versions should remain stable (otherwise there's a potential to massively inflate the number of DOIs issued for no good reason)

#### UNFs
UNFs (at least for tabular data) help *ensure* that a cited dataset is the one being used, so they avoid using incorrect versions, but they do not function as identifiers, i.e. UNF:6:edx+kB6SY2N3Zt9OsUbp4A== tells me nothing about where to find that dataset.

poikilotherm:

This pull request adds the long awaited option to generate PIDs for dataset version.

Closes #4499

This is WIP.

This is WIP.

Some minor additions. Will be included once avail.

Yes, will be included.

**Additional documentation**:
None yet.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/pidproviders/VersionPidMode.java
+++ b/src/main/java/edu/harvard/iq/dataverse/pidproviders/VersionPidMode.java
@@ -1,0 +1,114 @@
+package edu.harvard.iq.dataverse.pidproviders;
+
+import edu.harvard.iq.dataverse.Dataverse;
+
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Enumlike class to bundle available options for PIDs of Dataset Versions.
+ * Must be a class to ensure case-insensitive configuration values (not possible/reliable with enum)
+ */
+public final class VersionPidMode {
+    
+    /**
+     * Means feature is switched off, no version PIDs will be minted
+     */
+    public static final VersionPidMode OFF = new VersionPidMode("OFF");
+    
+    /**
+     * Means the feature is activated instance wide for all collections and datasets
+     * (No opt-out so far!)
+     */
+    public static final VersionPidMode GLOBAL = new VersionPidMode("GLOBAL");
+    
+    /**
+     * Means the feature must be activated per Dataverse Collection (opt-in)
+     */
+    public static final VersionPidMode COLLECTION = new VersionPidMode("COLLECTION");
+    
+    /**
+     * A collection of conducts for mode {@link #COLLECTION}, used in {@link edu.harvard.iq.dataverse.Dataverse}
+     * and {@link edu.harvard.iq.dataverse.DataverseServiceBean#wantsDatasetVersionPids(Dataverse)}:
+     * <ol>
+     *     <li>Collection may inherit version pid behaviour from the parent collection(s),</li>
+     *     <li>Collection may choose to be actively enabling it</li>
+     *     <li>Collection may choose to opt out and skip the minting</li>
+     * </ol>
+     */
+    public enum CollectionConduct {
+        INHERIT("inherit"),
+        ACTIVE("active"),
+        SKIP("skip");
+        
+        private final String name;
+        
+        CollectionConduct(String name) {
+            this.name = name;
+        }
+    
+        @Override
+        public String toString() {
+            return this.name;
+        }
+        
+        public static CollectionConduct findBy(String name) {
+            return Arrays.stream(CollectionConduct.values())
+                .filter(cs -> cs.name.equalsIgnoreCase(name))
+                .findFirst()
+                .orElse(null);
+        }
+    }
+    
+    
+    // Init as unmodifiable set
+    public static final Set<VersionPidMode> values;
+    static {
+        values = Set.of(OFF, GLOBAL, COLLECTION);
+    }
+    
+    private final String mode;
+    
+    // Hide the no-arg constructor - no one shall get fancy ideas of extension.
+    private VersionPidMode() {
+        this.mode = "";
+    }
+    
+    // Hide the constructor - no one shall get fancy ideas of extension.
+    private VersionPidMode(String mode) {
+        this.mode = mode;
+    }
+    
+    /**
+     * Used to enable auto-conversion for MicroProfile Config.
+     * Comparison is done case-insensitive to enable all variants of writing the config value.
+     *
+     * Note that a non-matching value will return null, which will trigger a {@link java.util.NoSuchElementException}
+     * for the conversion.
+     *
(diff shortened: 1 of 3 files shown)


poikilotherm:
Key takeaways from first feedback during tech hour:

- Put the business logic for no PID / major version only / major  + minor into the decision in `wantVersionPid()` and make it a three way thing. Don't put the decision into the PID provider.
- Think about overrides: current design says collection decides. The storage option for example can be overridden per dataset via API. Do we need this here?
- Tap into moving a dataset to new owner: give warning / abort / react to the new owner not having versions enabled when the dataset has identifiers for versions (which would be very quite trap)
- Give some thought into possibilities of "going back" or at least send warnings if for example the version was publicized but the dataset was not, which creates inconsistencies.

poikilotherm:
After an elaborate chat with @landreev yesterday (thanks again!), let me write down our key insights we got out of it:

1. Make the JVM setting `dataverse.pid.version.mode` three way: `off` (default), `allow-major`, `allow-minor`. That way, a sysadmin can restrict choice in collections, as this is cost related.
2. Enable collection curators to choose for a collection: `inherit` (default), `skip`, `major`, `minor`. This way they inherit from parent collection, have an opt-out (skip) and can choose between major only or also create for minors. The selection will be limited by the global setting.
3. We'll change the order of  creating and publicizing version PIDs to reduce the risks of incoherence at DataCite and database.
    1. We will create the PID after all file PIDs are publicized but before we publicize the dataset PID. This might be a long list and the chance of things going sideways is highest there.
    2. We will publicize the version PID *after* the dataset PID. If that update goes sideways, the version PID is registered, but not yet charged for and can be reused on next publication attempt. We can already include the version PID in the dataset PID metadata because we created it before (see 1).
    3. There is a tricky edge case. Assume version PIDs to be a suffix of the dataset PID. If someone tries to publish and says "major update" and things go sideways before the version PID is done but after the dataset PID is publicized, on the next publication attempt they might choose "minor update". As we did not yet publicize the major one, we can create a minor one and go with it. The major one can be reused later (it's not findable/resolvable anyway). The business logic to create an identifier *for a version using suffixes* needs to be OK with a PID already existing. (This is different for datasets, where we need to create a new identifier!!!)
4. We were in mutual agreement that we should make this PR smaller scoped. This PR should only be about the "inner core", enabling the general concept to be realized. It should not yet touch the UI, unnecessary API changes or PID provider support. This way, reviewing and quality assurance are much easier to handle without endangering a release hitting the public. Any documentation included in this PR should state clearly this is not yet ready and should only be used in development and experimentation.

qqmyers:
A couple minor points:
- you say 'off' in 1, and 'skip' in 2 - those are the same thing?
- ~3.2, 3.3 - PIDs are only non-deletable once they are publicized (I think that's true for all current PID providers) so I don't think there's ever a need to keep an un-published PID around after a failure, i.e. those can just be deleted, and new ones or the same PID recreated later if/when needed. One would have to make those calls to delete as part of a rollback action. Version PIDs are definitely a new case as the major/minor status and whether those categories get PIDs can change.

poikilotherm:
Ad 1) Yes, on purpose. The selection in the collection will be disabled of the feature is turned off globally. No one can turn it on by accident. 

"Skip" on the other hand is an opt-out. If your parent collection has it on, you might want to turn it off for a certain subcollection.

The decision on the collection is by curators! This is a difference to selecting storage per collection.

Ad 3.3) Yes, one could try to make that rollback. Let me add a note here that we don't do this for dataset PIDs or File PIDs, so maybe it's fine not to do it for version as well.

Also, in case things go sideways because the provider is out shopping 500s, it might be hard to ask for a deletion.

Of course, we could add a service that retries deleting failed registrations. Might be neat for dataset PIDs and file PIDs as well. But sounds like beyond scope here to me.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -574,7 +591,76 @@ public Response deleteDataverse(@Context ContainerRequestContext crc, @PathParam
...
+     * Endpoint to change attributes of a Dataverse collection.
+     *
+     * @apiNote Example curl command:
+     *          <code>curl -X PUT -d "test" http://localhost:8080/api/dataverses/$ALIAS/attribute/alias</code>
+     *          to change the alias of the collection named $ALIAS to "test".
+     */
+    @PUT
+    @AuthRequired
+    @Path("{identifier}/attribute/{attribute}")
+    public Response updateAttribute(@Context ContainerRequestContext crc, @PathParam("identifier") String identifier,
```
landreev, line 605: I just started working on #8889, enabling file pids in specific collections. It should be much more straightforward than the functionality in this PR.
Is it ok if I just copy-and-paste most of this API into my branch and use it for my new collection setting there?

landreev: (I added a simple documentation entry for this API in that PR, if you want to pull into your branch and expand on it)

poikilotherm: > hope that was ok!

That's fine. You could have cherry-picked (so it retain authorship and commit count), but never mind. 🤗

pdurbin:
@poikilotherm this PR is on [your wishlist](https://dataverse.zulipchat.com/#narrow/stream/375707-community/topic/Release.206.2E4.20Proposals/near/452276836) for 6.4 but there are a lot of merge conflicts.

Also, do you still feel that a size of 30 is right?

Thanks.

vera:
We are also interested in this feature. I saw above this feature was planned for 6.4. Is it now planned for a later version?

PR #9467, opened 2023-03-22
Fix for handling storageidentifiers in harvested files (dataverse_json)

### LINKED ISSUE #7736: Fix handling of storageidentifiers in dataverse_json harvests
Dataverses can harvest metadata in our custom json format from each other. Our own proprietary export and import are used in the process. 
A side effect of this method is that the storageidentifier, the physical location of the datafile on the remote installation ends up being imported *verbatim* into the harvesting dataverse; instead of the url of the download API on the remote end. 
We now have all these recently harvested files with strange storageidentifiers; 
like this one:
```
1784af576bf-27316028aabe
```
(no driver prefix; must have been harvested from a pre-4.20 Dataverse)
or this:
```
s3://dataverse-prod-s3:178290c4c23-3a496b7152af
```
(a file in somebody's S3 storage bucket...)
These are of course completely useless for a harvesting installation. We want to handle these the same way as when we harvest DDI between Dataverses. I.e., the dvobjects for these harvested files need to be created with the remote download API url  in storageidentifier field (would be `https://dataverse.tdl.org/api/access/datafile/something-something` for the last file above, for example...)

Aside from these entries being useless as imported, this is not urgent in that we don't use these remote locations for any practical purpose, as of now. (So that's why we haven't noticed until now). But it's still messy (I was completely weirded out when I saw the ones like the first one above; that looked entirely like a local storageidentifier, that somehow got created without a driver prefix...)

landreev:

See description in the issue. It's fairly straightforward. When harvesting in dataverse_json from other Dataverses, we've been importing their storageidentifiers (physical storage location keys) verbatim. I.e., I harvest from another server, I see things in my database that look like `s3:their-bucket-name:xxxxxxx-yyyyyyy` - which are of course meaningless/useless on my system. 
We never use these values for anything for harvested files; so, it's a matter of semantics, if this is even a bug - it doesn't break anything. But, these are meaningless, and also potentially confusing to an admin looking at the database. This PR changes it so that it saves the access api url pointing to the remote location of the file on its home server instead - like `https://demo.dataverse.org/api/access/datafile/173401`. We may or may not have some use for these one day, but for now at least it will be obvious that these are remote locations. 

(this was estimated as 10, so I kept it simple)

Closes #7736

Harvest a known set, like my `control_set` from demo, in dataverse_json format. Look at the last datafile dvobjects in the database - they should have access api urls pointing back to demo, like the one in the description, in the `storageidentifier` column. 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/imports/ImportServiceBean.java
@@ -264,7 +264,7 @@ public class ImportServiceBean {
         JsonObject obj = jsonReader.readObject();
         //and call parse Json to read it into a dataset   
         try {
-            JsonParser parser = new JsonParser(datasetfieldService, metadataBlockService, settingsService, licenseService);
+            JsonParser parser = new JsonParser(datasetfieldService, metadataBlockService, settingsService, licenseService, harvestingClient);
             parser.setLenient(true);
             Dataset ds = parser.parseDataset(obj);
 
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
@@ -71,6 +71,7 @@ public class JsonParser {
     MetadataBlockServiceBean blockService;
     SettingsServiceBean settingsService;
     LicenseServiceBean licenseService;
+    HarvestingClient harvestingClient = null; 
     
     /**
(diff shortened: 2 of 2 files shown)


pdurbin (approved):
I didn't test this but the code makes sense. Should work.

pdurbin:
p.s. And the API tests are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9467/1/testReport/edu.harvard.iq.dataverse.api/

PR #9482, opened 2023-03-28
IQSS/9431-support checksum alg in direct uploads

### LINKED ISSUE #9431: Direct upload in the UI doesn't follow :FileFixityChecksumAlgorithm
The Javascript in the UI has been assuming md5 regardless of whether the Dataverse instance has an alternate algorithm set via  [:FileFixityChecksumAlgorithm](https://guides.dataverse.org/en/4.7/installation/config.html#id104). 

Direct upload should use the specified algorithm.

Could consider blocking uploads that use the wrong algorithm (breaking change) and probably should add some notes about how to update the algorithm on existing files (there's an API for that).

5.13

PR to follow (fixed at QDR)

qqmyers:
**What this PR does / why we need it**:This PR addresses in two ways:
* it adds a new API endpoint to retrieve the name of the configured fixity algorithm, and
* it updates the direct upload javascript to use that API and then use the specified algorithm for direct uploads via the normal UI. (not the dvwebloader which is in a separate repo).

Closes #9431 

**Special notes for your reviewer**: fairly straight forward, just wasn't implemented (support for direct upload may predate the multiple algorithm implementation?). Also include minor code cleanup/recommended fixes from my IDE in the direct upload javascript.

**Suggestions on how to test this**: Do direct upload via the UI with various fixity algorithms specified and assure that those uploads use the correct algorithm. Try the new api/files/fixityAlgorithm call and verify it returns the current algorithm.

included

**Additional documentation**:
new api call documented, older admin call to update the alg on existing files also improved.


pdurbin (approved):
I didn't test the actual fix but I did check that http://localhost:8080/api/files/fixityAlgorithm responds and improved the docs a bit. I'm approving this.

PR #9497, opened 2023-04-05
IQSS/9496: restore legacy behavior including an empty persistentID

### LINKED ISSUE #9496: Restore empty file persistentId value in JSON responses
#8674 changed PID handling and inadvertently changed the no-PID behavior for the JSONPrinter from sending a "persistentId":"" key/value pair to not including a "persistentId" key. One effect of this change is that it breaks Previewers when opened in a separate page for files with no PIDs. THe change could affect any other users of the API, JSON metadata export, etc. (https://github.com/gdcc/dataverse-previewers/pull/29 will make previewers accept either case going forward). 

develop

qqmyers:
**What this PR does / why we need it**: Restores pre #8674 behavior in JSON file metadata - see issue

Closes #9496

**Special notes for your reviewer**: Slack discussion at https://iqss.slack.com/archives/C010LA04BCG/p1680701364259049

**Suggestions on how to test this**: Visually inspect JSON metadata export and confirm files with no PID have "persistentId":"" in their JSON objects. Can also verify that v1.3 and older previewers work when opened in a separate page.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -627,7 +627,8 @@ public class JsonPrinter {
         fileName = fileMetadata.getLabel();
         GlobalId filePid = df.getGlobalId();
         String pidURL = (filePid!=null)? filePid.asURL(): null;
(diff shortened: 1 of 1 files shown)


landreev (approved):
yay!

PR #9499, opened 2023-04-05
#9498 - fixing issue while searching for TermUri field that is a compound field

### LINKED ISSUE #9498: CVOC : Searching External Vocab Values and compound fields

luddaniel:

This PR will allow a compound field that is set as `"term-uri-field"` in `cvoc-conf.json` file to be indexed in SOLR with all values of `externalvocabularyvalue.value` table.
Without this PR, only when `"field-name"` equals `"term-uri-field"`,  values cached from a skosmos term will be indexed properly.

Closes #9498

Make sure that you can search Term in different languages using non compound or compound field with skosmos external vocab support (CVOC).
Example with non compound field :
```json
"field-name": "skosterm",
"term-uri-field": "skosterm",
```
Example with compound field :
```json
"field-name": "cvocDemo",
"term-uri-field": "cvocDemoTermURI",
```

Why not

**Additional documentation**:

related to commit 552a34a
Also, I have no idea if the change is needed is another function like json exporter.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
@@ -828,7 +828,7 @@ public class IndexServiceBean {
             }
 
             Set<String> langs = settingsService.getConfiguredLanguages();
-            Map<Long, JsonObject> cvocMap = datasetFieldService.getCVocConf(false);
+            Map<Long, JsonObject> cvocMap = datasetFieldService.getCVocConf(true);
             Set<String> metadataBlocksWithValue = new HashSet<>();
             for (DatasetField dsf : datasetVersion.getFlatDatasetFields()) {
 


qqmyers:
@luddaniel - thanks for the fix. I think this is already in #9402 - there were several places where trying to use the external vocab mechanism on a single child did not work.

luddaniel:
Yes it's already in https://github.com/IQSS/dataverse/pull/9402/files#diff-1a01af3c6f5f9893743e27eb5b74bcf87abadf4aea927690bd9f8d179926349d

PR #9500, opened 2023-04-06
Change log level from info to fine to suppress noise from netcdf

### LINKED ISSUE #9441: File ingest: Some extra Netcdf logging was left in place, making ingest noisy, especially of non netcdf files.
While ingesting a simple test stata file, I'm seeing several log entries related to netcdf. This adds noise to the log and seems unnecessary.

[ingest_50by1000_dta.txt](https://github.com/IQSS/dataverse/files/10960450/ingest_50by1000_dta.txt)

rtreacy:
**What this PR does / why we need it**: suppress noise in log file from netcdf by setting log level from info to fine

Closes #9441 

**Suggestions on how to test this**: Check log files when uploaded a file, netcdf noise should no longer appear

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
@@ -1269,7 +1269,7 @@ public class IngestServiceBean {
                     String ncml = netcdfFile.toNcml("");
                     inputStream = new ByteArrayInputStream(ncml.getBytes(StandardCharsets.UTF_8));
                 } else {
-                    logger.info("NetcdfFiles.open() could not open file id " + dataFile.getId() + " (null returned).");
+                    logger.fine("NetcdfFiles.open() could not open file id " + dataFile.getId() + " (null returned).");
                 }
             } catch (IOException ex) {
                 logger.info("NetcdfFiles.open() could not open file id " + dataFile.getId() + ". Exception caught: " + ex);
@@ -1307,7 +1307,7 @@ public class IngestServiceBean {
                 logger.fine("Aux file extracted from NetCDF/HDF5 file saved to storage (but not to the database yet) from file id  " + dataFile.getId());
                 ncmlFileCreated = true;
             } catch (Exception ex) {
-                logger.info("exception throw calling processAuxiliaryFile: " + ex);
+                logger.fine("exception throw calling processAuxiliaryFile: " + ex);
             }
         } else {
             logger.info("extractMetadataNcml: input stream is null! dataFileLocation was " + dataFileLocation);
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -1307,7 +1307,7 @@ public boolean extractMetadataNcml(DataFile dataFile, Path tempLocationPath) {
                 logger.fine("Aux file extracted from NetCDF/HDF5 file saved to storage (but not to the database yet) from file id  " + dataFile.getId());
                 ncmlFileCreated = true;
             } catch (Exception ex) {
-                logger.info("exception throw calling processAuxiliaryFile: " + ex);
+                logger.fine("exception throw calling processAuxiliaryFile: " + ex);
             }
         } else {
             logger.info("extractMetadataNcml: input stream is null! dataFileLocation was " + dataFileLocation);
```
pdurbin, line 1313:
```suggestion
            logger.fine("extractMetadataNcml: input stream is null! dataFileLocation was " + dataFileLocation);
```

pdurbin (commented):
I think we need to lower another info.

pdurbin (approved):
I didn't test this but the changes make sense.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -1269,7 +1269,7 @@ public boolean extractMetadataNcml(DataFile dataFile, Path tempLocationPath) {
                     String ncml = netcdfFile.toNcml("");
                     inputStream = new ByteArrayInputStream(ncml.getBytes(StandardCharsets.UTF_8));
                 } else {
-                    logger.info("NetcdfFiles.open() could not open file id " + dataFile.getId() + " (null returned).");
+                    logger.fine("NetcdfFiles.open() could not open file id " + dataFile.getId() + " (null returned).");
```
landreev, line 1272: @pdurbin I'm seeing that @rtreacy has changed the log level to .fine here; so this should remove the noisy message. What confuses me though is why are we trying to run this `extractMetadataNcml()` methods on ALL files, of all types here in the first place? I mean, you do already know which files are netcdf and which are not at this point. Because you have a dedicated test for this format in `FileUtil.determineFileType()`, which actually does the same thing - `NetcdfFiles.open()` - and gives the file the correct mime type `application/netcdf` or `.../x-hdf5` if that works. 
So you should definitely have one of the 2 types above in `dataFile.getContentType()` at this point for any file that can be opened as netcdf. So unless I'm missing something, it seems like this netcdf extraction should only be performed on those types. Just like we do with fits extraction - ?

landreev: (I only noticed this because I was deep in `FileUtil.determineFileType()` myself, on account of the runaway file descriptors)

pdurbin: @landreev yes, good point. I just created this PR as an alternative using your idea of only running `extractMetadataNcml()` on NetCDF or HDF5 files:

- #9510

pdurbin:
@kcondon yes, done. Thanks. Please see my comment above.

PR #9503, opened 2023-04-07
6723 unzip optimizations

### LINKED ISSUE #6723: High levels of open file descriptors while uploading a zip with lots of files in it
We are monitoring the number of open file descriptors ever since we had serious problems with it in the past (last year).
When the number of open file descriptors of the glassfish process is 3000 or higher (max is 4096) I get a mail, this happened several times in the past days. Now, the good news is that it goes down to the 'basis' value of around 700.
But if the peak will be too high, the system will fail and I don't want that to happen, besides this I get a lot of warning emails from the ‘monitoring’ script.
 
A typical run looks like the following:
 
2020-02-27T15:26:01+0100,742
2020-02-27T15:27:02+0100,1744
2020-02-27T15:28:01+0100,1744
2020-02-27T15:29:01+0100,2747
2020-02-27T15:30:01+0100,2747
2020-02-27T15:31:01+0100,1450
2020-02-27T15:32:01+0100,1452
2020-02-27T15:33:01+0100,2453
2020-02-27T15:34:01+0100,3456
 
With each line the timestamp followed by the number of open file descriptors, increments in steps of ~1000 (sampling every minute). 

Below a chart showing what happened on that day:
image
.

After another warning mail, we checked the logs

`[2020-03-02T13:51:04.422+0100] [glassfish 4.1] [SEVERE] [] [edu.harvard.hul.ois.jhove] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464422] [levelValue: 1000] [[
  Testing SEVERE level]]
 
[2020-03-02T13:51:04.431+0100] [glassfish 4.1] [SEVERE] [] [edu.harvard.hul.ois.jhove] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464431] [levelValue: 1000] [[
  Testing SEVERE level]]
 
[2020-03-02T13:51:04.437+0100] [glassfish 4.1] [SEVERE] [] [edu.harvard.hul.ois.jhove] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464437] [levelValue: 1000] [[
  Testing SEVERE level]]
 
[2020-03-02T13:51:04.447+0100] [glassfish 4.1] [SEVERE] [] [edu.harvard.hul.ois.jhove] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464447] [levelValue: 1000] [[
  Testing SEVERE level]]
 
[2020-03-02T13:51:04.450+0100] [glassfish 4.1] [WARNING] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464450] [levelValue: 900] [[
  Zip upload - too many files.]]
 
[2020-03-02T13:51:04.450+0100] [glassfish 4.1] [WARNING] [] [edu.harvard.iq.dataverse.util.FileUtil] [tid: _ThreadID=52 _ThreadName=jk-connector(3)] [timeMillis: 1583153464450] [levelValue: 900] [[
  Unzipping failed; rolling back to saving the file as is.]]
 `

This is from the code in: edu.harvard.iq.dataverse.util.FileUtil
There we see that there is a limit in: 
edu/harvard/iq/dataverse/util/SystemConfig.java

`/**
     * The default number of datafiles that we allow to be created through 
     * zip file upload.
     */
    private static final int defaultZipUploadFilesLimit = 1000; 
    private static final int defaultMultipleUploadFilesLimit = 1000;
    private static final int defaultLoginSessionTimeout = 480; // = 8 hours
 `
It is not set by us, so it should indeed be the default 1000. 

Next I tested uploading a zip with a lot of files >1000 and several things happened. 
On the GUI I got the message: "The number of files in the zip archive is over the limit (1000); please upload a zip archive with fewer files, if you want them to be ingested as individual DataFiles."
The files in the zip where indeed not extracted into the dataset, but the complete zip was added, as is. 

But the following problems occurred: 
- The number of open file descriptors peeked. 
- All the temporary files where left in the Dataverse temp dir. 

It would be much better if those two things did not happen.

Furthermore, I tested it with a special constructed zip with exactly 1001 files in it. 
Then all the files where extracted, although 1001 > 1000. 

If I change the limit to 999, with: 
'curl -X PUT -d '999' http://localhost:8080/api/admin/settings/:ZipUploadFilesLimit'
The zip is added 'as is'. This looks like a off-by-one error, which would be nice to have fixed in order to minimise confusion. 
[1001_files.zip](https://github.com/IQSS/dataverse/files/4287255/1001_files.zip)

landreev:

I was/am confused about what was expected to be done here, as a "size 3 spike" for this issue. If it was just a quick test to confirm that the issue was still present, then it appears that @kcondon has already done that (see his last comment in the issue). 

I put in a day+ of work into this (so, more than "3") and added a couple of optimizations. 

Our file type detection tests (those are run individually on every uncompressed file) were indeed leaving N extra un-closed file descriptors behind - i.e., 1 for each unpacked file. That has been fixed. But please note that not all the temp files that lsof is showing as open right after the upload are wasted resources necessarily. Generally jvm will hold onto some files that have been properly closed in the application, just in case; but it will close them as the resources are garbage-collected, and/or as more fds are needed. 
I don't think the number of file descriptors was necessarily the worst underlying issue here. The other part, that we were uncompressing and saving individual temp files, even if a zip archive contained more than the unpacking limit num. of files until that number was reached, was more serious. More of a waste, IMO. It is indeed very easy to run a quick check through the zip file directory and decide whether to unpack or save as is, beforehand. (Note that it's not just on account of the number of files, we also save zip files un-unpacked if one or more uncompressed files are above the individual size limit. It's just as easy to quickly check for that ahead of time too. 

Closes #6723

lsof should be showing fewer open files in .../files/temp. There should never be _more_ of these open fds than the number of files unpacked; and they should not stay around for long un-garbage collected. 

But, most importantly, you should be getting the warning "too many files inside, saving as is" much sooner, right after the upload is complete pretty much. 
 

**Additional documentation**:


rtreacy (approved):
Code looks good. Tested with a couple of large Consilience zip files I had at hand

kcondon:
@landreev Ready to merge except for merge conflicts

PR #9520, opened 2023-04-12
intermittent harvest tests failures #9240

### LINKED ISSUE #9240: Intermittent error from the new Harvesting Clients test
Has been observed a couple of times on Jenkins since the PR #9227 was merged. Meaning, it's intermittent, the test passes on most runs. 
The error happens in the new test that's somewhat extensive and runs a real harvest of an external archive. However, the 500 below happens before anything heavy is attempted, merely when creating the new harvesting client: 
```
java.lang.AssertionError: expected:201 but was:500
at edu.harvard.iq.dataverse.api.HarvestingClientsIT.testHarvestingClientRun(HarvestingClients0T.java:184)
```
which is really a cheapest operation you can think of. This makes me think that this is another weird timing condition caused by an overall load of a full restassured run on the system. I expect it to be straightforward to fix/work around.

landreev:

When the tests in HarvestingClientsIT were creating clients, they would use a random string 6 characters long for the identifier. Our "random strings" are random numbers in hex. But the identifier in this particular class has the constraint such that the string should not contain decimal digits only. So with the generation method above, the expected probability of a constraint violation should therefore be 0.625^6 = roughly 5.96% of the time. Which seems about right, as to the failure rate we've been seeing. 

I'm pretty sure we've run into this with dataverse aliases in the past. 

Closes #9240

I acknowledge that it's not awesome, the behavior of the api - the 500 it throws because of the constraint. I just wanted to keep this pr purely about fixing the test. The harvesting clients api in question is for admins only; and nobody will ever use it particularly frequently. So maybe we can fix this error handling one day... when everything else in the application is fixed. 

Seeing how the natural probability of failure was really low (i.e., it would keep passing most of the times if left as is), it would be difficult to confirm for sure that it's not going to happen again, ever. Just confirming that the test passes in the next jenkins run should be QA enough to merge it. If it starts failing again for whatever reason, we'll know soon enough. 

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
@@ -79,7 +79,7 @@ public class HarvestingClientsIT {
         // This method focuses on testing the native Dataverse harvesting client
         // API. 
         
-        String nickName = UtilIT.getRandomString(6);
+        String nickName = "h" + UtilIT.getRandomString(6);
         
 
         String clientApiPath = String.format(HARVEST_CLIENTS_API+"%s", nickName);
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I love this fix. Approved!

Issue #9536, opened 2023-04-19
Metrics API downloads problem


sirineREKIK:
Hello,

We detected a bug at Metrics API:
The result is the same for these two api calls: (count:35523)
1/ curl -s https://demo.dataverse.org/api/info/metrics/downloads | jq -r .data.count 
2/ curl -s https://demo.dataverse.org/api/info/metrics/downloads?parentAlias=csvconfv723 | jq -r .data.count
--> It is like the ‘parentAlias’ is not working good.
For the scenario we used this dataset for test: https://demo.dataverse.org/dataverse/csvconfv723

When investigating the code, we found that the SQL Query is missing parenthesis in the condition, and we think that is the origin of the problem: 
Class: MetricsServiceBean
Method: downloadsToMonth(..)
Query: select count(id) from guestbookresponse where date_trunc('month', responsetime) <=  to_date('" + yyyymm + "','YYYY-MM') or responsetime is NULL .......

**--> We propose that you add () to the query that become:** 
select count(id) from guestbookresponse where ( date_trunc('month', responsetime) <=  to_date('" + yyyymm + "','YYYY-MM') or responsetime is NULL ) .......

qqmyers:
Good catch! I can confirm at QDR that the query is wrong as you indicate - 'and' takes precedence over the 'or' in the where clause. I checked the other metrics queries and don't see any others affected by this.

Are you interested in making a pull request with the fix? If so, please do. Otherwise, let me know and I can submit the fix.

Thanks!

sirineREKIK:
It's done for the PR : (https://github.com/sirineREKIK/dataverse/tree/9536_metrics_api_download)
Is it ok for you ?

PR #9538, opened 2023-04-19
Bugfix for /api/metrics/downloads/toMonth/{yyyymm} endpoint: add parenthesis to condition SQL Query

sirineREKIK:
this PR is related to the issue : [https://github.com/IQSS/dataverse/issues/9536](url)

Closes: #9536

Testing (added by @qqmyers ):

Try the api/metrics/downloads/toMonth/{yyyymm} endpoint with/without a '?parentAlias=' parameter and verify that the downloads with the parameter are lower/correspond to downloads from just datasets in that subcollection.
As this is a one-line change in the query for this API call, there shouldn't be any effects to other metrics calls, etc.


qqmyers:
@sirineREKIK - one issue with the PR - can you merge your branch with the latest IQSS develop branch, or create a new branch/new PR from IQSS/develop with your change? Right now, since your branch is at v5.12.1, the automated build and test system is failing to run all of the tests (because it expects to be able to run a v5.13 war file). I don't think your commit can break any of those tests, but we like to see those tests all run before merging changes.

sirineREKIK:
Hello @qqmyers, sorry for the wrong manip, I think all is good now.
Do not hesitate if you still need other modifications.
Thanks!

PR #9523, opened 2023-04-13
extract bounding box from netcdf/hdf5 files and insert into the geospatial metadata block

pdurbin:
**What this PR does / why we need it**:

Prevent users from having to manually enter bounding box values. As with FITS, we get the values from the files themselves.

**Which issue(s) this PR closes**:

- Closes #9331

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

Issue #9416, opened 2023-03-01
 NcML previewer doesn't work with drafts


pdurbin:
Currently we get this error:

"Unable to retrieve NcML file. Please try again. The most common issue is that your login has timed out. If the problem persists, please contact the support team of this data repository. Please include any status code included at the end of this message: Forbidden"

![221693399-e18ececd-197e-431f-b130-7eafb405f900](https://user-images.githubusercontent.com/21006/221956396-9f6bbccf-e2c3-41cb-9088-11c32c0199b1.png)

We should update ncml.js to handle API tokens (like other previewers)

Please note: this issue is only for tracking. The work will be done in a gdcc repo:

- https://github.com/gdcc/dataverse-previewers/issues/25

pdurbin:
Still no luck. I'm sure it's a real bug. I can reproduce it on demo and prod.

However, I can't reproduce this on my laptop or my dev1 server. It just works in both places.

Here's dev1, for example, happily returning the NcML file:

![Screen Shot 2023-05-16 at 5 22 00 PM](https://github.com/IQSS/dataverse/assets/21006/357fc193-fee5-47dc-afef-ed42d4b5b92c)

I'm not sure why it's working in my browser like that for dev1 (and my laptop). As expected when I try to curl the file from the command line it fails:

`$ curl https://dev1.dataverse.org/api/access/datafile/1073/auxiliary/NcML/0.1`
`{"status":"ERROR","code":403,"message":"Not authorized to access this object via this API endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.","requestUrl":"https://dev1.dataverse.org/api/v1/access/datafile/1073/auxiliary/NcML/0.1","requestMethod":"GET"}`

qqmyers:
Did it get fixed? - See https://github.com/gdcc/dataverse-previewers/pull/30#issuecomment-1540707492 - sounds like it was fixed in source but not yet built at that point.

pdurbin:
@qqmyers thanks but no I believe that's Jan pre-emptively fixing the H5Web previewer.

qqmyers:
Sorry - my bad. You don't have the cookie feature flag on or something?

pdurbin:
I'm still not sure why I don't see this problem when I serve up previewers from the same server as Dataverse but I was able to replicate it my pointing Dataverse on my laptop to previewers hosted on my dev1 server.

I'm putting this into "ready for review". The pull request is here:

[...]

Issue #9331, opened 2023-01-26
Extracting lat/long and insert into geospatial bounding box fields


pdurbin:
@atrisovic and I exchanged emails with @plesubc and he has inspired us to use GDAL, ogrinfo or similar to try extracting latitude and longitude from a NetCDF file.

As this is just a spike, some discovery, we're sizing this as 10 or 1 day.

Here's part of the email from Paul (this file happens to span the entire globe):

"Metadata extraction is a relatively simple process assuming you’re using GDAL. The GDAL suite exports file metadata to stdout, so all you really need to do is capture and process the text. Of course, differing formats have differing outputs, because life is never that simple.

So, for example, imagine you downloaded a netcdf from here:

https://data.ceda.ac.uk/badc/ukmo-hadobs/data/derived/MOHC/HadOBS/HadEX3/v3-0-2 (HadEX3-0-2_cwd_ann_1901-2018.nc). This isn’t some special data set, it’s the result of a google for spatial netcdf files.

Basically, filtering this file through ogrinfo (one of the utilities in GDAL), you get something like this as output:

```
ogrinfo *nc
INFO: Open of `HadEX3-0-2_cwd_ann_1901-2018.nc'
      using driver `netCDF' successful.
Metadata:
  NC_GLOBAL#acknowledgement=RJHD was supported by Met Office Hadley Centre Climate Programme funded by BEIS and Defra
  NC_GLOBAL#CDI=Climate Data Interface version 1.9.9rc1 (https://mpimet.mpg.de/cdi)
  NC_GLOBAL#cdm_data_type=grid
  NC_GLOBAL#CDO=Climate Data Operators version 1.9.9rc1 (https://mpimet.mpg.de/cdo)
  [NC_GLOBAL#creator_email=robert.dunn@metoffice.gov.uk](mailto:NC_GLOBAL#creator_email=robert.dunn@metoffice.gov.uk)
  NC_GLOBAL#creator_name=Robert Dunn
  NC_GLOBAL#creator_url=[www.metoffice.gov.uk](http://www.metoffice.gov.uk/)
  NC_GLOBAL#dataset_version=3.0.2
  NC_GLOBAL#date_created=Mon Oct 26, 12:10 2020
  NC_GLOBAL#DOI=https://doi.org/10.1029/2019JD032263
  NC_GLOBAL#geospatial_lat_max=90
  NC_GLOBAL#geospatial_lat_min=-90
[...]

PR #9541, opened 2023-04-19
extract bounding box from NetCDF/HDF5 files and insert into the geospatial metadata block 

pdurbin:
insert the bounding box into the geospatial metadata block

Prevent users from having to manually enter bounding box values. As with FITS, we get the values from the files themselves.

- Closes #9523

I tried pretty hard to get the compound field code working but ended up switching to something that works, `jsonParser.parseField` which is also used by the native API.

Upload various NetCDF files, especially ones that have a bounding box. Here's a nice small one with a bounding box that the code tests: https://www.ncei.noaa.gov/data/international-comprehensive-ocean-atmosphere/v3/archive/nrt/ICOADS_R3.0.0_1662-10.nc

Not really. The bounding box is populated. Here's what it looks like when you are just looking at (not editing) a dataset:

It seems a little weird that we don't show any labels (north, south, east, west). I assume there's an open issue for this but it's out of scope.

Screen Shot 2023-04-19 at 4 55 39 PM

The values look more reasonable when you are editing:

Screen Shot 2023-04-19 at 4 55 56 PM

Yes, included.

**Additional documentation**:

Included. I updated the User Guide.


pdurbin:
I've been saying at standup that I need to tweak something for S3 but I was wrong. S3 works fine so in f5be4a8 I removed the FIXME and replaced it with some details.

I'll note that @JR-1991 @atrisovic and I are still discussing a bit the getStandardLongitude method in this PR. I'm happy to discus this more with a reviewer. Here's an awesome diagram Jan made and [posted](https://dataverse.zulipchat.com/#narrow/stream/376593-geospatial/topic/extract.20lat.2Flong.20.239331/near/354199150):

diagram

I'm putting this in "ready for review".

src/main/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractor.java
```diff
@@ -0,0 +1,168 @@
...
+        float eastAsFloat;
+        float westAsFloat;
+        try {
+            westAsFloat = Float.valueOf(westAndEastLongitude.getWestLongitude());
+            eastAsFloat = Float.valueOf(westAndEastLongitude.getEastLongitude());
+        } catch (NumberFormatException ex) {
+            return null;
+        }
+        if (westAsFloat > 180 || eastAsFloat > 180) {
+            Float westStandard = westAsFloat - 360;
```
sekmiller, line 154: I think that if we're converting 0-360 to -180-180 then we need to subtract just 180 from the given value, not 360.

JR-1991: @sekmiller I was also puzzled with this at the beginning, but subtracting 360° is the correct procedure. The [prime meridian](https://oceanservice.noaa.gov/facts/longitude.html#:~:text=The%20prime%20meridian%2C%20which%20runs,the%20eastern%20and%20western%20hemispheres.) (0° longitude) needs to stay at Greenwich and thus the interval's "cycle point" where both ends meet changes by 180°. Thus, this has to be attributed for, which leads to the 360°.

That's how I understood it so far, but this illustration may explain it better:

<p align="center">
image
</p>

pdurbin: @JR-1991 thanks! Even with that explanation, it's hard to keep straight in my head.

I was relying on a tool to check that maps are the same. That is these two URLs (minus 360) show the same map:

- https://linestrings.com/bbox/#343.68,41.8,353.78,49.62
- https://linestrings.com/bbox/#-16.320007,41.8,-6.220001,49.62

Here's a screenshot of the maps (and a bit more commentary) I originally posted at https://dataverse.zulipchat.com/#narrow/stream/376593-geospatial/topic/extract.20lat.2Flong.20.239331/near/348542658

Screen Shot 2023-05-03 at 9 20 22 AM

sekmiller: @JR-1991  OK. Looking at the globe and the example location makes sense when the value is between 180 and 360, but what the code does here is if the value for either is over 180 it subtracts 360 from both so there is a potential for a situation where one val is > 180 and the other is less and in that case, say 170, and subtracting 360 results in -190 which is off the scale below -180.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -1531,72 +1654,71 @@ private void processDatasetMetadata(FileMetadataIngest fileMetadataIngest, Datas
                         // create a new compound field value and its child 
                         // 
                         DatasetFieldCompoundValue compoundDsfv = new DatasetFieldCompoundValue();
-                        int nonEmptyFields = 0;
```
pdurbin, line 1534: This whitespace change and below is from me removing this code and then (just) re-adding it because it seems to be working afterall! My test still passes at least. 😅

src/test/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractorTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    public void testStandardLongitude() {
+        NetcdfFileMetadataExtractor extractor = new NetcdfFileMetadataExtractor();
+        // Both are over 180. Change.
+        assertEquals(new WestAndEastLongitude("-16.320007", "-6.220001"), extractor.getStandardLongitude(new WestAndEastLongitude("343.68", "353.78")));
+        // One is over 180. Change
+        assertEquals(new WestAndEastLongitude("-260.0", "-179.0"), extractor.getStandardLongitude(new WestAndEastLongitude("100", "181")));
+        // Both are under 180. No change.
```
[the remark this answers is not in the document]
pdurbin: Yes, I agree. I'll ping Jan for some input on this.

src/main/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractor.java
JR-1991: @sekmiller thanks for the review! Indeed subtracting from both would lead to a failure. I tested it visually and in the case of $\theta_1 = [280°, 170°]$ the result should translate to $\theta_2 = [-80°, 170°]$ - Hence if a degree is under 180 it should not be subtracted. I have updated the illustration:

<p align="center">
image
</p>

I understand it in this way, that within this domain transformation only one halb of the circle is changing. Note, in 0° - 360° domain the first half is the same as in the domain of (-180°) - 180° and thus there is a 1:1 relation if degrees are within 0-180. Yet, the other half does change to another range in a more drastic way. Tipping over 180° by a degree suddenly leads you to -179° which in the other domain corresponds to 181° - Hence the 360° provides a valid transformation for this case.

It's quite complicated though 😅

src/test/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractorTest.java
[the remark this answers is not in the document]
JR-1991: @pdurbin just for clarification, the right hand-side extractor transforms by subtracting 360° from both?

If so, then I have to agree, this test should fail and truly evaluate to $[100°, -179°]$ due to the above [comment](https://github.com/IQSS/dataverse/pull/9541#discussion_r1185691536) and that there can't be -260° since the domain has it's lower bound at -180°. 

Given the proposed change in the calculation, the test should look like this:

```java
assertEquals(
    new WestAndEastLongitude("100.0", "-179.0"),
    extractor.getStandardLongitude(new WestAndEastLongitude("100", "181"))
);
```

What do you think?

src/main/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractor.java
JR-1991: Maybe I should add to the figure, that $f(\theta)$ is defined for $\theta \in [0°, 360°]$ and if $\theta < 0°$ its already in the desired domain.

src/test/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractorTest.java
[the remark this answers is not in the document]
pdurbin: I think it's late on a Friday but I powered through and pushed some fixes around longitude at 2edb351.

Almost as importantly, I added docs! Rules in the User Guide about how users (and QA) should expect the code to behave. I'd love to hear what you (@JR-1991) and @atrisovic and @sekmiller (whom I've been bugging all day) think! Here they are:

- West Longitude and East Longitude are expected to be in the range of -180 and 180. (When using :ref:`geospatial-search`, you should use this range for longitude.)
- If West Longitude and East Longitude are both over 180 (outside the expected -180:180 range), 360 will be subtracted to shift the values from the 0:360 range to the expected -180:180 range.
- If either West Longitude or East Longitude are less than zero but the other latitude is greater than 180 (which would imply an indeterminate domain, a lack of clarity of if the domain is -180:180 or 0:360), metadata will be not be extracted.

I also added more tests. I'm happy to add more if you all think of more edge cases. I don't have real NetCDF files to test with that have wacky values. It might be a challenge for QA to find NetCDF files to exercise all the rules.

src/main/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractor.java
pdurbin: I tried to address all this in 690b8a4a just now and left this comment about it: https://github.com/IQSS/dataverse/pull/9541#discussion_r1186507839

Maybe I should have resolve other conversation and gotten it down to just this one. Oh well. Please see the other comment.

pdurbin:
I'm ready for more review so I'm removing myself from this PR. Here are the changes since the last round:

1fb2798c44 add API test for FITS file metadata extraction #9331
189c7749a7 restore original "extract from compound" code #9331
2edb351cab document and fix up the rules for longitude, add tests #9331

Yes, I was convince the code to extract from compound fields wasn't working. I put it back and now it works! Well, it was always working, I suppose. Now I'm feeding it the data it expects!

src/main/java/edu/harvard/iq/dataverse/ingest/metadataextraction/impl/plugins/netcdf/NetcdfFileMetadataExtractor.java
atrisovic: I read through the thread and agree with the comments! Instead of the lines 155-173, we can check each coordinate by itself and it would work well, ie:

```
Float westStandard = westAsFloat;
Float eastStandard = eastAsFloat;

 if (westAsFloat > 180) {
             westStandard = westAsFloat - 360;
}

if (eastAsFloat > 180){
             eastStandard = eastAsFloat - 360;
}

WestAndEastLongitude updatedWeLong = new WestAndEastLongitude(westStandard.toString(), eastStandard.toString());
return updatedWeLong;
```

atrisovic (changes requested):

kcondon:
Issues found:
1. Preview for netcdf file shows permissions issue:
Screen Shot 2023-05-09 at 11 39 47 AM
Screen Shot 2023-05-09 at 11 40 02 AM

pdurbin:
> Preview for netcdf file shows permissions issue:

Sadly, this is a known issue. It's in "sprint ready", at least:

- #9416

The fix will need to happen over in the previewers repo so there's an issue to track it there:

- https://github.com/gdcc/dataverse-previewers/issues/25

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -367,7 +369,20 @@ public List<DataFile> saveAndAddFilesToDataset(DatasetVersion version,
 						} else {
 							logger.fine("Failed to extract indexable metadata from file " + fileName);
 						}
-					} else if (FileUtil.MIME_TYPE_INGESTED_FILE.equals(dataFile.getContentType())) {
+                                        } else if (fileMetadataExtractableFromNetcdf(dataFile, tempLocationPath)) {
```
ErykKul, line 372: This breaks direct uploads for me. When using a direct upload, the file is `unattached` and the owner is set to 'null' on line 336. However, 'fileMetadataExtractableFromNetcdf' needs the owner dataset to determine the folder inside `storageIO.open();`. This results in a `NullPointerException`, preventing adding the files to the dataset. I have only tested with the local filesystem, but from looking at the code, this probably also breaks uploading files directly to S3. Can you investigate and fix it? Thanks.

pdurbin: @ErykKul yikes! This has already been merged. 😬  Could you please create an issue about this?

kcondon: I've reproduced this in develop: direct file upload via UI fails with a grayed out screen as if waiting for a completion signal and a server log error: 
[diredct_upload_ui_err.txt](https://github.com/IQSS/dataverse/files/11517681/diredct_upload_ui_err.txt)

Note that direct file upload continues to work via API. Also, deleting the file continues to work via ui.

pdurbin: I can reproduce it too. I created an issue:

- #9601

Thanks, both!

PR #9548, opened 2023-04-22
allow war files to be deployed by the base image

pdurbin:

https://guides.dataverse.org/en/5.13/container/base-image.html#locations says, "Any EAR or WAR file, exploded WAR directory etc are autodeployed on start" but only exploded WAR directories are currently supported. This PR allows WAR fles to be deployed.

Screenshot 2023-04-22 at 6 34 56 PM

Closes NONE

I tested this with a WAR file I built from https://github.com/IQSS/dataverse-people

I build the base image locally with a custom tag and used the following line in a Dockerfile to copy the war into place.

```
COPY ./target/dataverse-people.war ${DEPLOY_DIR}/
```

No.

No.

**Additional documentation**:

None.


poikilotherm (approved):
Looks good to me!

I believe this is not even a size 3 and can go through the pipeline without much QA. The missing parenthesis have been thoroughly tested for being the correct fix.

PR #9556, opened 2023-04-24
IQSS/9555 update list curation states api call

### LINKED ISSUE #9555: /api/datasets/listCurationStates not documented and only works for first drafts
As per the title, the listCurationStates api call that produces a CSV file listing datasets with the assigned curation label (along with creation/last modified date, and assigned curators (those who have directly assigned permissions to publish the dataset) which was added along with the curation labels functionality, was not documented. Further, the query to get datasets incorrectly only retrieved unpublished datasets rather than all those with a draft version (curation labels can be applied to any dataset with a draft version).
5.13/develop

#9247 

PR to follow

qqmyers:
**What this PR does / why we need it**: Documents the /listCurationStates api call and makes it work with all draft versions

Closes #9555

**Suggestions on how to test this**: Assign some curation labels to first or second draft versions and verify that the datasets appear in the CSV file returned when calling /api/datasets/listCurationStates as a superuser

No

**Additional documentation**:


scolapasta (approved):

doc/sphinx-guides/source/api/curation-labels.rst
```diff
@@ -93,3 +93,22 @@ To get the list of allowed curation labels allowed for a given Dataset
     curl -H X-Dataverse-key:$API_TOKEN "$SERVER_URL/api/datasets/:persistentId/allowedCurationLabels?persistentId=$DATASET_PID"
 
 You should expect a 200 ("OK") response with a comma-separated list of allowed labels contained in a JSON 'data' object.
+
+
+Get a Report on the Curation Status of all Datasets
```
pdurbin, line 98:
```suggestion
Get a Report on the Curation Status of All Datasets
```

pdurbin (commented):
Just fixing a tiny doc thing.

kcondon:
@qqmyers This works except the output goes to standard out rather than to a csv file as documented: "You should expect a 200 ("OK") response with a 'datasets.status.csv' file download."

qqmyers:
FWIW: Curl doesn't respect the file name being sent by the server so the example doesn't download a file (doing the api call in the browser does). I've changed the docs to just note that the response is CSV-formatted. Also merged from dev.

PR #9426, opened 2023-03-06
Payara 6 testing

ErykKul:
**What this PR does / why we need it**:
I was playing with Payara 6 and I had some tests failing. I have merged the develop branch on the 8305-payara6-ee10 branch, fixed some tests, done some library updates, etc.

In order to avoid double work, I thought that I will just share it. Note that two integration test are still failing due to the known issue with slf4j logging:
https://github.com/payara/Payara/issues/6211
https://github.com/payara/Payara/issues/5898

image

**Special notes for your reviewer**:
If it is more work to review it than the amount of work it saves, just reject it.
My commits:
image

**Suggestions on how to test this**:
Run the unit and integration tests; the build uses Payara 6 with jdk 17.

PR #9003, opened 2022-09-29
9002 allow direct upload setting

ErykKul:
**What this PR does / why we need it**:
In some situations, direct upload might not work from the UI, e.g., when s3 storage is not accessible from the internet. This pull request adds an option to allow direct uploads via API only. This way, a third party application can use direct upload from within the internal network, while there is no direct download available to the users via UI.

**Which issue(s) this PR closes**:

Closes #9002

**Special notes for your reviewer**:
I have replaced `Boolean.getBoolean("...")` with `Boolean.parseBoolean(System.getProperty("..."))` in some places, as I think these were bugs. Let me know if I need to revert it.

**Is there a release notes update needed for this change?**:
yes

PR #8940, opened 2022-08-26
8822 incomplete datasets via api

ErykKul:
**What this PR does / why we need it**:
When integrating GitLab or iRods with Dataverse, we want researchers to be able to send files and just enough information to get the dataset created via API. Publishing is prevented until required fields are entered (using the GUI, most often).

**Which issue(s) this PR closes**:

Closes #8822

**Suggestions on how to test this**:
First, you will need to update the solar configuration. For this feature to work, a new solr field is required: "datasetValid". The needed configuration is added in "schema.xml". Notice that when you do not use this feature, updating solr configuration and/or reindexing is not required. Nevertheless, it is reasonable to keep the solr configuration up-to-date and do the configuration update and reindex.

Next, you need to enable these settings:
- :AllowInvalidMetadataThroughAPI: set this to "true". If you do that without updating the solr configuration first, you will not be able to ingest dataset anymore due to the missing field errors.
- :CanReviewInvalid: can be set optionally to "true", you can then submit datasets with invalid metadata for review. Otherwise, you will need to correct the metadata before sending to review. Datasets with invalid metadata can never be published, regardless of any setting.
- :ShowValidityFilter: when set to "true" it enables a filter in "My Data" page (important: this filter will only work correctly when all metadata is reindex on the solr server, otherwise valid datasets will not be shown):

image

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
When the new feature is not used, there are no changes to the user interface. When you use the new feature, you will have the possibility to enable filtering of valid and/or invalid metadata on the "My Data" page, as discussed earlier.

Invalid metadata is marked with a new tag "Invalid metadata":

image

This tag is visible in the "My Data" page and the main page, when logged in. When you do not log in, you can only see published metadata, and since invalid metadata cannot be published, you will not see that tag. The tag, together with a warning, is also shown when viewing a dataset with invalid metadata:

image

A warning is also shown when submitting for review and trying to publish a dataset with invalid metadata:

image

The exact content of the tag and the text of the warning can be modified with the "notValid" and "dataset.message.invalid.warning" properties, respectively.

**Is there a release notes update needed for this change?**:
Yes.

**Additional documentation**:
You can also watch the demo as given during the community meeting. You can find the video at the top of DataverseTV: https://dataverse.org/dataversetv

[...]

PR #9558, opened 2023-04-25
async indexing after update command

### LINKED ISSUE #9557: Performance issues when uploading files to a dataset containing thousands of files
Upload a few thousand files (e.g., 2.5 MB each) and try adding a file to it with the API (e.g., Readme.md). Also, try adding multiple files one by one with the API to the same dataset. Each upload takes minutes i.s.o. seconds (on a single machine deployment, e.g., the docker deployment with Solr, DB and Dataverse on the same machine).

* When does this issue occur?
It happens for all update operations for datasets containing thousands of files.

* Which page(s) does it occur on?
I have tested it with API, UI experience may be different, but it also may be the same because of the reused update command (I did not test that in the UI).

* What happens?
The operation waits for the dataset to be indexed before returning an HTTP response. Indexing takes longer each time a file is added (it is a cumulative problem). When there are already many files in the dataset, the problem gets very bad.

All users.

I would expect a linear behavior, uploading a file to a dataset (or any other update operation) should take the same time, independently of the number of files already present in the dataset.

5.13

ErykKul:
It improves the performance of updating datasets containing huge number of files (thousands of files). With this PR, the updates take seconds i.s.o. minutes (roughly the same as updates of dataset containing small number of files).

Closes #9557 

Making indexing async was not sufficient to fix this problem, as the default flush mode is auto (on query): indexing in the background causes table locks when attempting another update operation while index is busy. Setting the flush behavior to "commit" solves this.

Also, async indexing starts immediately after an update, but when multiple updates happen to the same dataset while indexing is busy, only the last updated dataset object will get reindexed (we skip unnecessary indexing).

Normally, I would expect that the "commit" flush mode should not impact the functionality. For example, all commands executed on the internal engine are done on transactions, with commit at the end. All indexing, ingesting, post-processing, etc. happen after the transaction is committed, e.g., while executing the "onSuccess" method.

Also, not waiting for indexing to be ready before the operation returns should not impact the functionality neither since the index must be processed by Solr first: immediately querying Solr after the Java index operation is ready my result in a response not reflecting the last changes that were indexed. The difference is that it can take longer time now to see the results on Solr, and thus it can become more noticeable.

Upload thousands of files to a dataset and try several update operations. For example, uploading some files directly after each other.

No.

Possibly. Short mention of the performance improvement should be sufficient.


qqmyers (approved):
Very nice! I walked through the logic in getNextToIndex and don't see any problems. That said, comments in the code on how it works would be useful.

qqmyers:
The CI failed after the unit tests and before the integration tests ran. @donsizemore  - any ideas:

```
at TASK [dataverse : allow authenticated users to add to root] ********************
fatal: [localhost]: FAILED! => {"access_control_allow_headers": "Accept, Content-Type, X-Dataverse-Key, Range", "access_control_allow_methods": "PUT, GET, POST, DELETE, OPTIONS", "access_control_allow_origin": "*", "access_control_expose_headers": "Accept-Ranges, Content-Range, Content-Encoding", "changed": false, "connection": "close", "content_length": "74", "content_type": "application/json;charset=UTF-8", "elapsed": 0, "json": {"message": "Can't find dataverse with identifier='root'", "status": "ERROR"}, "msg": "Status code was 404 and not [200]: HTTP Error 404: Not Found", "redirected": false, "server": "Payara Server  5.2022.3 #badassfish", "status": 404, "url": "http://localhost:8080/api/dataverses/root/assignments?key=86a3f945-f565-4e54-a031-bc2551b98062", "x_frame_options": "SAMEORIGIN", "x_powered_by": "Servlet/4.0 JSP/2.3 (Payara Server  5.2022.3 #badassfish Java/Red Hat, Inc./11)"}

NO MORE HOSTS LEFT *************************************************************
```

ErykKul:
I have added some comments to the code. I hope it made it more readable. Also, I hope that this change did not break any integration tests. I do not see the results of the tests, so let me know if something breaks. Thanks!

qqmyers:
Thanks. FWIW, it looks like build issues rather than the PR contents are what is causing the failures right now. Once that's fixed we can rerun and let you know if there are any IT failures.

ErykKul:
@qqmyers , @donsizemore 
I am not sure if this is the case here, but I had a similar problem with the integration tests when my Solr did not start up. I happened because the base image is "latest" rocky linux (for docker-aio image) and apparently procps is not there by default anymore. I added it in the yum line at the end like this:
```
RUN yum install -y java-11-openjdk-devel postgresql13-server sudo epel-release unzip curl httpd python2 diffutils procps
```
This fixed the Solr starting up. I hope it helps.

Last time I had integration tests working was when I played with payara 6, I had to change quite some things to make it work (https://github.com/IQSS/dataverse/pull/9426).

qqmyers:
@ErykKul - Great that you can investigate. FWIW: I think the failure to find root has been seen before - possibly a timing issue that you change to flush-mode is making more likely? Also, FWIW: I cherry-picked your commit to QDR and the call to create a role assignment works fine there (on a pre-existing database).

qqmyers:
FYI: It looks like all of the unit tests are failing with a TransactionRequired exception. The failure on the unit tests job is related to the Run docker/login-action@v2 - Error: Username and password required, but all the unit tests appear to pass in the CI build.

pdurbin:
Yes, please don't worry about the Docker stuff. @poikilotherm and I will fix eventually. 😅

poikilotherm:
Hear hear! 😬🙈

ErykKul:
The transaction required thing was because of the fush in Ejb Dataverse Engine that I added, I moved it to the specific Create Dataverse Command and the test look way better now:
```
[ERROR] Failures: 
[ERROR]   ConfirmEmailIT.testConfirm:112 Expected status code <200> doesn't match actual status code <500>.

[ERROR]   DatasetsIT.testCuratePublishedDatasetVersionCommand:2803 Expected status code <200> doesn't match actual status code <500>.

[ERROR]   FilesIT.testAccessFacet:1112 JSON path data.total_count doesn't match.
Expected: <1>
  Actual: 0

[ERROR]   HarvestingServerIT.testMultiRecordOaiSet:606 Wrong number of items on the first ListIdentifiers page expected:<2> but was:<5>
[ERROR]   LinkIT.testDeepLinks:170 JSON path data.total_count doesn't match.
Expected: <1>
  Actual: 0

[ERROR]   UsersIT.convertNonBcryptUserFromBuiltinToShib:314 Expected status code <200> doesn't match actual status code <500>.

[ERROR]   UsersIT.testMergeAccounts:250 expected:<200> but was:<401>
[ERROR] Errors: 
[ERROR]   DatasetsIT.testPrivateUrl:964 » Connect Connection refused (Connection refused...
[ERROR]   SignpostingIT.testSignposting:91 » Connect Connection refused (Connection refu...
[INFO] 
[ERROR] Tests run: 199, Failures: 7, Errors: 2, Skipped: 8
```

The two errors are exactly the same as I was getting on the develop branch, so I will ignore them for now. I will investigate the 7 failing tests in more detail. All units test passed on my machine, so the failing on git looks incorrect. I think the test running on github has no internet access to pull the xsd schema: https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/codebook.xsd

ErykKul:
Good news: I have found what was causing the locking with flash-mode auto: the indexing updates the dataset while indexing it by setting last indexed time, that is why updating the dataset while the indexing was busy was not possible (deadlock situation). I have fixed the issue by making the update the last thing after indexing and making it asynchronous in a new transaction. It does not lock anymore, so I have reverted the flush-mode setting (back to default "auto"). This should fix some timing issues with integration tests.

Also, I will retest if asynchronous indexing is still necessary. If it still is, I will investigate how to improve the performance of the indexing itself. This may take some time, I think it will be a different pull request, if I find anything.

ErykKul:
With my last commits, the CPU went nicely down in my tests. In my test scenario, indexing of large dataset takes now around 10 seconds instead of more than a minute. Further improvement of the indexing speed would require smart regeneration of the solar docs for files; only regenerating files that have changes would improve the performance.

It looks like this version can handle large datasets (containing thousands of files) very nicely. I will rerun the integration tests to see if there are still problems.

ErykKul:
Most tests pass now. Some that are failing look not related to the changes I made. Some got fixed by adding the "wait for search", since the indexing happens in the background now. It looks safe to me; I will patch our test environment and do some more testing tomorrow. If everything goes well, we will consider patching production (this one is urgent for us since we have a researcher that is unable to upload a large dataset).

qqmyers:
I'll look more later but FWIW: the api.UtilIT.sleepForReindex method might help with tests messed up by timing. That call waits until the lastindexedtime is updated.

ErykKul:
@qqmyers thanks for the hint! I did implement it, and now I get the same test results as on the develop branch (I think they should pass on GitHub since I am missing some settings probably).
```
[ERROR] Failures: 
[ERROR]   HarvestingServerIT.testMultiRecordOaiSet:606 Wrong number of items on the first ListIdentifiers page expected:<2> but was:<5>
[ERROR] Errors: 
[ERROR]   DatasetsIT.testPrivateUrl:964 » Connect Connection refused (Connection refused...
[ERROR]   SignpostingIT.testSignposting:91 » Connect Connection refused (Connection refu...
[INFO] 
[ERROR] Tests run: 199, Failures: 1, Errors: 2, Skipped: 8
```

qqmyers (approved):
Again - very nice (PR of the year candidate)! I left a few comments, mostly to make sure I understand rather than about anything that should change. The one open question I have is whether we should retain/expand the separate logging of indexing failures that was done through LoggingUtil. As I mentioned, I think expanding could be useful.

Other than that, putting a release-note in docs/release-notes that summarizes the impact of this PR (faster indexing, especially for datasets with large numbers of files, and, if true, removing the need to wait for indexing to complete to avoid the possibility of Optimistic Lock Exceptions when making further changes, etc.) would be good.

ErykKul:
@qqmyers 
I am still learning the EJB specifications and I gave my best shot at implementing the fix. However, it is far from perfect. For example, it is a bad practice to synchronize in static method, since it does not work on clustered Payara servers (like some of us use). The better way is to use the JMS, like the ingesting does. I think it is a good idea to implement the same approach here. I will give it a try. Furthermore, I will also try to answer your questions the best I can. It is always a good idea to have more reviewers, especially since my knowledge of EJB specifications is limited. All remarks, suggestions and questions are always welcome.

I have also found the there is still a problem when uploading images, since ingesting makes thumbnails, which can take some time for larger files. Ingesting takes a lock that prevents updating the dataset version. This generates dataset locked errors on our test server. I will also investigate this deeper to see if I find something that can be done.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -348,48 +348,77 @@ public Future<String> indexDataverse(Dataverse dataverse, boolean processPaths)
...
+    }
+
+    private Future<String> indexDataset(Dataset dataset, boolean doNormalSolrDocCleanUp) throws  SolrServerException, IOException {
+        Future<String> result = doIndexDataset(dataset, doNormalSolrDocCleanUp);
+        updateLastIndexedTime(dataset.getId());
+        return result;
     }
     
-    public Future<String> indexDataset(Dataset dataset, boolean doNormalSolrDocCleanUp) throws  SolrServerException, IOException {
+    private Future<String> doIndexDataset(Dataset dataset, boolean doNormalSolrDocCleanUp) throws  SolrServerException, IOException {
```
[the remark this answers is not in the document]
ErykKul: I did not want to refactor too much, such that the core of the change remains better visible. The Future was not used in the code anyway, I think it is not a problem to remove it and make the method void.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
```diff
@@ -270,21 +265,12 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
     
     @Override
     public boolean onSuccess(CommandContext ctxt, Object r) {
-
-        boolean retVal = true;
-        Dataset dataset = (Dataset) r;
-
-        try {
-            Future<String> indexString = ctxt.index().indexDataset(dataset, true);
```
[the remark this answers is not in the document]
ErykKul: This is what I have observed and how I understand it.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1405,17 +1435,29 @@ private String addOrUpdateDataset(IndexableDataset indexableDataset, Set<Long> d
...
         /// Dataset updatedDataset =
         /// (Dataset)dvObjectService.updateContentIndexTime(dataset);
         /// updatedDataset = null;
         // instead of making a call to dvObjectService, let's try and
         // modify the index time stamp using the local EntityManager:
-        DvObject dvObjectToModify = em.find(DvObject.class, docs.getDatasetId());
+        DvObject dvObjectToModify = em.find(DvObject.class, id);
         dvObjectToModify.setIndexTime(new Timestamp(new Date().getTime()));
         dvObjectToModify = em.merge(dvObjectToModify);
+        em.flush();
```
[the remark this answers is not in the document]
ErykKul: I just removed the two lines, they are not needed.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -2485,11 +2499,15 @@ static boolean sleepForReindex(String idOrPersistentId, String apiToken, int dur
...
-            String hasStaleIndex = timestampResponse.body().jsonPath().getString("data.hasStaleIndex");
-            System.out.println(hasStaleIndex);
-            stale = Boolean.parseBoolean(hasStaleIndex);
+            try {
+                timestampResponse = UtilIT.getDatasetTimestamps(idOrPersistentId, apiToken);
+                System.out.println(timestampResponse.body().asString());
+                String hasStaleIndex = timestampResponse.body().jsonPath().getString("data.hasStaleIndex");
+                System.out.println(hasStaleIndex);
+                stale = Boolean.parseBoolean(hasStaleIndex);
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
ErykKul: It was catching illegal argument exception since apiToken can be null. But later I have fixed the UtilIT.getDatasetTimestamps such that it can take null as apiToken, so it is no longer needed. I have removed the try/catch wrapping.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1405,17 +1435,29 @@ private String addOrUpdateDataset(IndexableDataset indexableDataset, Set<Long> d
                 throw new IOException(ex);
             }
         }
+        return docs.getMessage();
+    }
+
+    @Asynchronous
```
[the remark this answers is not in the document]
ErykKul: At least this is the goal of isolating the update operation in the smallest possible transaction. The only locking problems I still have are when ingesting needs to make thumbnails, it takes an ingest lock on the dataset that prevents editting.

Also, if the locking exception would happen on the transaction here, it would be OK since it means that the dataset is being updated and it will be reindexed anyway.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2823,14 +2809,7 @@ public Response lockDataset(@Context ContainerRequestContext crc, @PathParam("id
                 // refresh the dataset:
                 dataset = findDatasetOrDie(id);
                 // ... and kick of dataset reindexing:
-                try {
-                    indexService.indexDataset(dataset, true);
-                } catch (IOException | SolrServerException e) {
-                    String failureLogText = "Post add lock indexing failed. You can kickoff a re-index of this dataset with: \r\n curl http://localhost:8080/api/admin/index/datasets/" + dataset.getId().toString();
-                    failureLogText += "\r\n" + e.getLocalizedMessage();
-                    LoggingUtil.writeOnSuccessFailureLog(null, failureLogText, dataset);
```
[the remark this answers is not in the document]
ErykKul: I am not sure if I did that correctly, but I changed the asyncIndexDataset method to use `LoggingUtil.writeOnSuccessFailureLog`.

ErykKul:
I think that ingesting does not need to block the update dataset version command. Maybe I am missing something, but I did change the code to allow updating the dataset version when ingesting lock is taken. It did not break any tests. I hope it did not break something that I am not aware of.

ErykKul:
Now I see it: when ingesting a file that has been deleted or replaced, the ingesting would fail and lock is not removed. This is why ingesting locks updating. I will work something out...

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -348,48 +349,78 @@ public Future<String> indexDataverse(Dataverse dataverse, boolean processPaths)
...
-        return ret;
+    // nextToIndex contains datasets mapped by dataset id that were added for future indexing while the indexing was already ongoing for a given dataset
+    // (if there already was a dataset scheduled for indexing, it is overwritten and only the most recently requested version is kept in the map)
+    private static Map<Long, Dataset> nextToIndex = new HashMap<>();
+    // indexingNow is a set of dataset ids of datasets being indexed asynchronously right now
+    private static Set<Long> indexingNow = new HashSet<>();
+
+    // When you pass null as Dataset parameter to this method, it indicates that the indexing of the dataset with "id" has finished
+    // Pass non-null Dataset to schedule it for indexing
+    synchronized private static Dataset getNextToIndex(Long id, Dataset d) {
```
poikilotherm, line 369: What I do not like about this approach: it implements lots of custom logic to solve the producer-consumer problem at hand instead of using Java standard mechanisms from `java.util.concurrent`.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -348,48 +349,78 @@ public Future<String> indexDataverse(Dataverse dataverse, boolean processPaths)
...
+        }
+        // otherwise, start a new job
+        indexingNow.add(id);
+        return d;
     }
 
     @Asynchronous
-    public Future<String> asyncIndexDataset(Dataset dataset, boolean doNormalSolrDocCleanUp) throws  SolrServerException, IOException {
-        return indexDataset(dataset, doNormalSolrDocCleanUp);
+    public void asyncIndexDataset(Dataset dataset, boolean doNormalSolrDocCleanUpe) {
```
poikilotherm, line 389: I completely agree with you that, as we ignore the returned `Future` anyway everywhere in the code, we can make this `void`. Yet, I'd argue that we could take this even a step further.

Instead of using `@Asynchronous`, an EJB specific mechanism to execute this method decoupled from the rest of the calling code, maybe it would be a good time to start using an [`Executor`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executor.html). We could deploy tasks/datasets to a queue (both have implementations in `java.util.concurrent`), making an executor pick them up. This would even enable scaling the indexing onto multiple threads, as executors can be pooled.

Using a `Queue` also enables adding metrics for the status and throughput via MicroProfile Metrics, so people can adapt the queue size and executor pools (and all that comes with that like resources etc).

poikilotherm (changes requested):
I have some remarks, as I'd rather see using standard components for queueing instead of custom made.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -348,48 +349,78 @@ public Future<String> indexDataverse(Dataverse dataverse, boolean processPaths)
...
-        boolean doNormalSolrDocCleanUp = false;
-        // return indexDataset(dataset, doNormalSolrDocCleanUp);
-        Future<String> ret = indexDataset(dataset, doNormalSolrDocCleanUp);
-        dataset = null;
-        return ret;
+    // nextToIndex contains datasets mapped by dataset id that were added for future indexing while the indexing was already ongoing for a given dataset
+    // (if there already was a dataset scheduled for indexing, it is overwritten and only the most recently requested version is kept in the map)
+    private static Map<Long, Dataset> nextToIndex = new HashMap<>();
+    // indexingNow is a set of dataset ids of datasets being indexed asynchronously right now
+    private static Set<Long> indexingNow = new HashSet<>();
```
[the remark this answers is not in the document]
ErykKul: Thanks, corrected.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
ErykKul: From what I understand, calls to @Asynchronous methods are executed with the EJB thread pool, which has by default 16 core threads. These are then managed by admins: https://docs.payara.fish/enterprise/docs/Technical%20Documentation/Payara%20Server%20Documentation/Server%20Configuration%20And%20Management/Thread%20Pools/EJB%20Thread%20Pool.html

I am not sure if it is a good idea of creating extra thread-pools (or does it use the same thread EJB thread pool)? Maybe if we want to make a more limited thread-pool with only few threads? But on the other hand, we want the indexing to finish as soon as possible, since some code might be waiting for the results. Nevertheless, it is something worth to consider as it might have some benefits.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
ErykKul: I went with custom logic because of the specific scenario: I am doing many updates (e.g., uploading 1000s of files) one after another, and I want to index only once at the end. Since I can not know when the end would come, and the same solution must work for just one update where we want the indexing result to be available as soon as possible, I can only skip as many index operations as possible, but not more than that.

I explicitly did not want to use a queue and a producer-consumer logic, as it would make it very complex to achieve the same functionality as it is implemented now. I would still have to know if an index is being done for a specific dataset at a given time (indexingNow set) and I would need to implement removing items from the queue, and even worse: reordering of the queue to give a priority to the datasets not being indexed now. Synchronizing the producer-consumer logic the way the current solution does would also be very difficult.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
poikilotherm: Obviously one should use the Jakarta EE Concurrency `Executors`, because we's want the application server to manage the pool. Here's the spec for Java EE 8 https://jakarta.ee/specifications/concurrency/1.1 (note that mostly the Javadoc is useful there, for the docs better look into https://jakarta.ee/specifications/concurrency/2.0/concurrency-spec-2.0.html which is Jakarta EE 9, the renamed EE 8)

IMHO your implementation with the set is kind of like a queue already, but with one important difference: there are no mechanisms in place that avoid saturating the consumer.

A) If you just put in enough requests (for different datasets), the queue grows infinite. This might be OK for us, but the Solr service usually has resource limitations as well, so this might be something a sysadmin wants to tune for.

b) If your indexing is fast enough, it will not avoid double work. If reindexing takes 1 second and the same dataset comes in again after 3 seconds, it will still do two indexing ops. Some might say this is what we want, but on the other hand in large installations you might want to have tunable delay. It's probably OK to have an updated search index after some X seconds when edits come in hot (machine edits etc), but searches are less frequent.

C) There is no feedback to a user when indexing performance degrades. As the command engine is simply dropping of the workload in an infinite queue, we cannot provide a feedback response to the command engine that now is not a good time so please come back later. Such feedback should not result in aborting publications (as the indexing happens after all that done), but at least it should create log messages about performance degration or even notifications to the sysadmin. Then an admin might investigate and tune configuration and resources or schedule a reindex for off-hours to spread the load.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
poikilotherm: The existing queue mechanism provide all this...?

> I would still have to know if an index is being done for a specific dataset at a given time (indexingNow set)

Why would this be necessary? The executors would take from the queue (so when a dataset is being indexed right now it not be in the queue at the time you want to edit the same dataset again). The problem to solve is to make another executor not take up the same dataset as long as it is being indexed by another. There are 2 options for that: a) set a `Lock` or a `Semaphore` and/or b) add a delay for the taking (readily avail as `DelayedQueue`) where delay time > index time

>  and I would need to implement removing items from the queue, 

Nope, you will only need to check if a dataset is in the queue already and not add it again.

> and even worse: reordering of the queue to give a priority to the datasets not being indexed now.

I don't think that is necessary.

> Synchronizing the producer-consumer logic the way the current solution does would also be very difficult.

Synchronizing what exactly? Are you referring to the `synchronized` modifier? The blocking queues in `java.util.concurrent` are thread safe already.

ErykKul: DelayQueue would not work, as index of empty dataset takes a second, and index of a dataset with 7000 files takes 5 minutes (we don't know yet how long it takes for 50000 files as it is not possible with the current code to upload that many in less than 270 days).

The Lock you mention is exactly what I said about the synchronizing, regardless of the fact that the queue is already synchronized. To be efficient and allow indexing of multiple datasets at once, you would need a lock for each dataset. To have a constant time of finding locks, you could store them in a set (exactly what I am doing now with the set indexingNow).

Checking if a dataset is already in a queue is done now with contains, which is a constant time operation for a map. The new solution would look almost the same as the current one, except for the expensive (i.e., not constant time) queue operations of contains and reordering, which would be very ugly and hard to implement.

And reordering makes sense even if there is only one dataset in the queue (or even stronger: especially when there is only one dataset in the queue). If it is the one that takes 5 minutes, some other user would need to wait at least 5 minutes before his dataset would be indexed.

This problem gets worse when we have multiple workers: either they pick a dataset from a queue and wait for the lock to get free (which means that adding a file to that dataset would put another indexing for that dataset in the queue, etc., until all workers wait on the same lock to index the same dataset), or they wait until locked dataset goes away from the queue before picking the next one (leaving the workers with nothing to do for 5 minutes).

A way better solution would be then that workers pick up only datasets that are not locked now. At that point, a queue does not make much sense and can be better replaced with a map having constant operation times. Exactly what happens now.

I have really considered a queue before I implemented this and have given a lot of thought to that problem. It might not look like much, but I believe that my code is a nice and elegant solution to a very hard problem. Please reconsider it as it is now. I am used to using queues in my code (you can even check my integration tool, I think it handles concurrency very nicely with a queue), I just do not see how it fits here. I might be missing something, and will look at it again, but I think I am already biased...

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
ErykKul: I am not an expert in EJB, there might be a solution that is beyond my reach. I think I will not be able to implement it by myself.

I will give it my best shot and finish what I am doing now. We can decide what to do with that later on.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
poikilotherm: Thank you @ErykKul for this explanation! Glad you already gave this that much thought, much appreciated!

I will stop arguing here, as you make some very valid points with the non-linearity of index times.

May I request that you do add explanations about this to the code? All of this seems to be non-trivial and we have no good other place to document architectural decisions aside from the code itself. Some in-depth Javadoc would certainly help to clarify what has been tried and why this is the solution we go for. In case we revisit this in the future, this would provide very valuable insights that might be forgotten about by then.

Thank you for working on this! Much appreciated!

ErykKul: Great, thanks! I also appreciate all your comments and input, I have still a lot to learn about the EJBs, the existing code, etc. I will extend the documentation to clarify the intent (I can see why this code might rise some eyebrows at the first sight). Maybe we will figure out something better in the long term.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
ErykKul: I have moved the implementation to a clustered singleton bean with blocking thread pool executor, where the maximum number of threads is configured via microprofile. I have kept the skipping indexing logic and added more documentation about it. I think we are getting there.

I am not sure about the way that the executor is initialized, should it come from some managed executor factory or something? One that I checked was from the glassfish api and used exactly the same code that I used for its implementation.

It would be also very useful if I could see why the integration test are failing, running them by myself is very time-consuming. Maybe there could be something added that e-mails the errors automatically when the tests fail? I will rerun them later locally.

ErykKul: I had to remove the blocking thread pool executor from the commit
https://github.com/IQSS/dataverse/pull/9558/commits/84116dd30351a4fc4f9bed2e882dafcfc016ac25

I noticed that when 4 users would be uploading large datasets simultaneously, the dataverse could be blocked for days (not allowing any updates to datasets), I went back to the previous solution with @Asynchronous annotation. It could degrade the performance when more users upload large datasets (containing thousands of files), but it should not block, as long as the payara has a decent task scheduling algorithm. I have also noticed that 26 other methods also use the @Asynchronous annotation, the problem could be bigger than just the indexing... I will do some more testing and simulate more users uploading large datasets.

ErykKul: It looks like the @Asynchronous annotation has the same default behavior as the blocking thread pool executor, it blocks when the thread pool is exhausted. Luckily I was wrong, and it does not block for days, as new indexing is not accepted for any dataset until a thread becomes available. I will need some more testing to confirm that for sure, but I notice the wait when uploading multiple datasets simultaneously.

ErykKul: I could not make it work, I gave up and reverted to the previous solution.

ErykKul:
Now I have the same results locally from the integration tests as in the develop branch, but they still fail here. Can someone check what is going? Stack trace would be welcome.

ErykKul:
Hi. I have added a release note. I have also tried reworking the async operation to use blocking queues and thread pool executors (with microprofile settings), and preferably taking into account the clustering of payara servers. However, this did not work and I believe that I am at the limit of what I could do in a short term. For the longer term, I think we should reevaluate the indexing itself. I think it can be improved by detecting which files did exactly change, and generate Solr docks for only those files. There could also be some other improvements to be found.

I think that the thread pool executors with blocking queue that are configurable with microprofile and that work well on clustered payara servers is a very useful feature. There are plenty of other places in the code that would benefit from that. I have a feeling that we should develop something more general, usable in other contexts. We could also consider managed (scheduled) executors, and some other paths. I believe this is something for longer term, and possibly out of scope for this PR.

qqmyers:
java.lang.AssertionError: 
JSON path data.total_count doesn't match.
Expected: <1>
  Actual: 0
	at edu.harvard.iq.dataverse.api.SearchIT.testIdentifier(SearchIT.java:754)

qqmyers:
FWIW: The unit test and readthedocs failures don't look related to the PR. The former is a known issue. The latter is a  `Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168)` error I haven't seen before.

ErykKul:
I have patched our production server v5.12.1 and started an upload yesterday. The CPU remains around 20% (80% idle, similar to my local and pilot tests with the patched version) and production remains responsive. The current upload speed for a dataset with 10,000 files in it is at least 25 times faster than the unpatched version (with a spectacular drop in CPU usage), bringing the upload of 60,000 files from around a year to 11 days. Note that the upload speed is near constant in the patched version and is much less influenced by the amount of files already in the dataset, meaning that a year of uploading is a very optimistic assessment for the unpatched version.

Indexing and ingesting look also fine (with thumbnails and indexing being queued and having small delay, but almost in real time).
I have also patched the v5.13, with similar results. I tested it locally and on our test server.

Overall, this PR looks very stable and safe to use.

qqmyers:
Nice! - Can you remind me - you're using a file store and therefore aren't using the direct upload and multi-file upload APIs? (My guess is that the asynch changes will help in all cases whereas the queuing part only helps when not using the multifile direct uploads. It also could be that the remaining time/load for file stores is mostly due to the uploads going through the Dataverse server. Overall, I'll be curious to see performance on S3.)

ErykKul:
I did not test with S3, but I can test with direct uploads using this pr: #9003
The results should be very comparable to S3. Up until now I have only tested with direct uploads one by one file, it was the same as without the direct upload. I think multifile upload should be the fastest option and probably not much influenced by these improvements. The difference could be more noticeable once that the files are uploaded and when making updates to the metadata, adding one file, or sending the dataset for review or publishing it. I would expect that these operations should be faster now, regardless of the used data store.

ErykKul:
I have tried the multifile upload with direct upload to the filestore, first with the unpatched version (without this PR, only using the #9003). First I uploaded 10,000 files followed by upload of 1 file. The upload itself is only limited by IO and CPU for hashing. It took 4 minutes to upload 10,000 files of 2.5 MB each. The API call to add these file to the dataset took then around 20 minutes. 24 minutes in total. Adding one small file (6.3 kB) after that took 16 minutes. The server became very unresponsive. I have tried deleting the files with UI, deleting the dataset, etc., in order to prepare my local server for the next test with a patched version. Everything fails without a clear error. Once that, I will reset the server, I will try using the patched version to see if there is any difference. The log from my test with unpatched version:
```
2023/05/08 10:04:01 job added for doi:10.80442/Q8KS6G
2023/05/08 10:04:02 doi:10.80442/Q8KS6G: job started
2023/05/08 10:04:02 doi:10.80442/Q8KS6G: processed 10/10000
2023/05/08 10:04:02 doi:10.80442/Q8KS6G: processed 20/10000
...
2023/05/08 10:08:57 doi:10.80442/Q8KS6G: processed 9980/10000
2023/05/08 10:08:58 doi:10.80442/Q8KS6G: processed 9990/10000
2023/05/08 10:28:49 doi:10.80442/Q8KS6G: job ended
```
Uploading of 1 file (6.3 kB):
```
2023/05/08 10:35:06 job added for doi:10.80442/Q8KS6G
2023/05/08 10:35:06 doi:10.80442/Q8KS6G: job started
2023/05/08 10:51:33 doi:10.80442/Q8KS6G: job ended
```

ErykKul:
My test dataset remains locked for edit for some reason, without me doing any edit operation. Quite strange. I would not try an upload like this in production. Server remains slow.

ErykKul:
I can't delete the draft, everything gets time-outs. I think I killed my local test server...

ErykKul:
After trying out the patched version, the results are very similar. Server got slow, trying to delete the files got dataset locked, etc. The upload itself was faster:
```
2023/05/08 12:30:35 job added for doi:10.80442/Q8KS6G
2023/05/08 12:30:35 doi:10.80442/Q8KS6G: job started
2023/05/08 12:30:36 doi:10.80442/Q8KS6G: processed 10/10000
2023/05/08 12:30:36 doi:10.80442/Q8KS6G: processed 20/10000
...
2023/05/08 12:35:11 doi:10.80442/Q8KS6G: processed 9980/10000
2023/05/08 12:35:12 doi:10.80442/Q8KS6G: processed 9990/10000
2023/05/08 12:43:35 doi:10.80442/Q8KS6G: job ended
```
Uploading a single small file after that (6.3 kB):
```
2023/05/08 12:47:29 job added for doi:10.80442/Q8KS6G
2023/05/08 12:47:29 doi:10.80442/Q8KS6G: job started
2023/05/08 12:48:48 doi:10.80442/Q8KS6G: job ended
```
So, roughly speaking, uploading 10,000 files got twice as fast with the patch and multifile upload scenario. Using multifile upload with direct upload to the dataset (#9003) with 10,000 files got well significant speed boost: 12.5 times faster. However, my local test server does not support such big uploads at once very well. It probably needs some time to finish ingesting, etc. I also suspect that because of the slow-down of the server, the small file upload at the end would be faster when the files would have been uploaded one by one, but such upload takes very long time and is difficult to test.

ErykKul:
I did some more testing, and it looks like only one dataset that was slow (the one with many files). Other datasets responded well. Also, the lock was the delete operation that was busy in the background. Once it finished, the dataset got unlocked and does not contain any files. Everything runs smooth now.

poikilotherm:
/push-image

Sorry @ErykKul you're my victim to try out the new push images workflow...

BTW can you update your branch to latest `develop`? That will make the shabby falling check go away. Thx!

ErykKul:
@poikilotherm I am perfectly fine with testing new stuff out :-)
I have just updated the branch.

poikilotherm:
/push-image

Thx for the update! Will push again now with latest develop merged.

ErykKul:
Hi. We have managed to add a large dataset (more than 60,000 files) to our production server. However, it crashes the server when more than one person is viewing it. Uploading, etc., looks fine. I have already optimized retrieving the dataset, and I am still looking at this issue. I was planing to make more commits on this branch, when it is fixed. For me, it is the same issue; it is not useful getting large datasets on the server if they can crash it. Also, the #9003 would make it complete, allowing us to go to the production with the file upload tool that I was working on it.

@pdurbin, @poikilotherm, @qqmyers,
Do you think that it would be possible to get this PR and the #9003 into the new release (v5.14)? This would make it easier for us. I am hoping to finish the performance tweaks tomorrow.

ErykKul:
My last two commits fixed the server crashing for me. It made the datasets load much faster.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -750,6 +750,21 @@ else if (dataset.isLockedFor(DatasetLock.Reason.InReview)) {
...
+                if (lock.getReason() != DatasetLock.Reason.Ingest) {
+                    locked = true;
+                    break;
+                }
+            }
+        }
+        if (locked) {
+            checkEditDatasetLock(dataset, dataverseRequest, command);
+        }
+    }
```
qqmyers, line 767: Not sure I understand this  - the code only calls checkEditDatasetLock if there is an ingest lock, but that call will immediately throw an IllegalCommandException due to the Ingest lock.

ErykKul: The goal here is to call checkEditDatasetLock only if there is at least one lock that is not an ingest lock. If there is no lock, or there are only ingest locks, it is fine to do an update. Otherwise, default checkEditDatasetLock is called.

ErykKul: I have rewritten the code for more readability:
```
boolean hasAtLeastOneLockThatIsNotAnIngestLock = dataset.isLocked() && dataset.getLocks().stream()
    .anyMatch(lock -> !DatasetLock.Reason.Ingest.equals(lock.getReason()));
if (hasAtLeastOneLockThatIsNotAnIngestLock) {
    checkEditDatasetLock(dataset, dataverseRequest, command);
}
```

ErykKul:
I see that the integration tests are failing again. They did pass with https://github.com/IQSS/dataverse/pull/9558/commits/e3751bd32bded7aa1ea6b387979259d1fc25f342. I assume it is something on the develop branch or on the Jenkins that makes them fail. They also fail on my other branches after merging the develop branch to them. If it is something due to my code change, please let me know.

pdurbin:
@ErykKul it's not your code. Something completely unrelated. Thanks for the heads up. I started a thread at https://iqss.slack.com/archives/C03R1E7T4KA/p1683919620384289 about it and I'll try to remember to circle back to you once we have a fix in place. Sorry for the noise!

pdurbin:
> @pdurbin, @poikilotherm, @qqmyers,
> Do you think that it would be possible to get this PR and the #9003 into the new release (v5.14)? This would make it easier for us. I am hoping to finish the performance tweaks tomorrow.

Wow, #9003 wasn't really even on my radar. I don't know if you've heard us talk about the new-ish backlog board/project at https://github.com/orgs/IQSS/projects/34/views/1

I don't have any direct control over which PRs or issues are prioritized but I do manage a column on that board called "Community Backlog" and I just moved that pull request up a few places:

Screen Shot 2023-05-12 at 3 31 49 PM

(I've been prioritizing small PRs. Lots of internationalization fixes, some from first-time contributors).

You can read more about how I'm trying to manage the community backlog here:

- https://github.com/IQSS/dataverse.org/issues/68

Feedback is certainly welcome! 😅 All these PRs from fantastic contributors like you is a good problem to have! ❤️ 

Oh, and by the way, we have already prioritized #8940. We talk about it as standup every day. I see that just a few hours ago you left a comment for Steve on it. I'll make sure he sees it. Thanks!

ErykKul:
We have now intensively tested these changes. It runs in production and allows for large datasets (we needed to extend the heap space to make it run smoothly). I have just fixed one problem we ran against: when an unexpected exception is thrown, the lock was not removed from the dataset, preventing it from reindexing. The exception in our case was the geolocation field that could not be processed (mentioned in the community call), which helped with testing these changes :smile:

ErykKul:
Indexing after publish (and re-indexing of datasets in general) was horribly slow due to the large number of queries for each file. I have now reduced the number of needed queries to the minimum (just one query, no queries by file).

qqmyers:
I've removed the Size 3 label from this as it has continued to grow. @ErykKul - can we freeze this (except for fixing bugs/responding to comments) so we can get this reviewed?

qqmyers:
FWIW: One related thing I've noticed at QDR is that, after a metadata update/other dataset change, the save() request doesn't return until after things in the onSuccess() method of the commands returns. Because of this, one can open a new tab for the dataset and see the change before the original page says the update is done. This PR should help with that by making indexing faster, but someday looking at making the whole execution of onSuccess methods asynchronous might be useful as well.

src/main/java/edu/harvard/iq/dataverse/Dataset.java
```diff
@@ -43,6 +43,8 @@
  * @author skraffmiller
  */
 @NamedQueries({
+    @NamedQuery(name = "Dataset.findById", 
+                query = "SELECT o FROM Dataset o LEFT JOIN FETCH o.files WHERE o.id=:id"),
```
qqmyers, line 47: Can you add a comment - this should only be used if you're going to use the files (otherwise use DatasetService.find() )? Would you ever want to use this outside the findDeep() method (at least people should consider that method rather than using this directly)?

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -105,6 +106,29 @@ public Dataset find(Object pk) {
         return em.find(Dataset.class, pk);
     }
 
+    public Dataset findDeep(Object pk) {
```
qqmyers, line 109: Can you add a comment about when to use this and what it does? If I understand - all of these left-join-fetch hints (and the left-join-fetch in the query) are optimizations to get everything on one call rather than relying on lazy fetching later?

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -105,6 +106,29 @@ public Dataset find(Object pk) {
...
+            .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas.fileCategories")
+            .setHint("eclipselink.left-join-fetch", "o.files.guestbookResponses")
+            .setHint("eclipselink.left-join-fetch", "o.files.embargo")
+            .setHint("eclipselink.left-join-fetch", "o.files.fileAccessRequests")
+            .setHint("eclipselink.left-join-fetch", "o.files.owner")
+            .setHint("eclipselink.left-join-fetch", "o.files.releaseUser")
+            .setHint("eclipselink.left-join-fetch", "o.files.creator")
+            .setHint("eclipselink.left-join-fetch", "o.files.alternativePersistentIndentifiers")
+            .setHint("eclipselink.left-join-fetch", "o.files.roleAssignments")
+            .setLockMode(LockModeType.NONE)
```
qqmyers, line 128: When is this OK? If I understand, in the indexing case this makes sense because a) you're not expecting any writes and b) if there are some, you're going to reindex anyway. Assuming this is saving significant time, can you add a comment about when this shouldn't be used. (Common theme through a few of my comments - trying to making sure that developers who don't have deep JPA/SQL knowledge or who just see the method name somewhere don't misuse them.)

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1981,8 +1980,8 @@ private String init(boolean initFull) {
                 } else {
                     // an attempt to retreive both the filemetadatas and datafiles early on, so that
                     // we don't have to do so later (possibly, many more times than necessary):
-                    AuthenticatedUser au = session.getUser() instanceof AuthenticatedUser ? (AuthenticatedUser) session.getUser() : null;
-                    datafileService.findFileMetadataOptimizedExperimental(dataset, workingVersion, au);
+                    //AuthenticatedUser au = session.getUser() instanceof AuthenticatedUser ? (AuthenticatedUser) session.getUser() : null;
+                    //datafileService.findFileMetadataOptimizedExperimental(dataset, workingVersion, au);
```
qqmyers, line 1984: You're removing the only two uses of findFileMetadataOptimizedExperimental - can you then remove it's definition in DataFileServiceBean? (I assume you found that findDeep() is as fast or faster than when using this method? findDeep() definitely looks more maintainable.) I'd also suggest just deleting these lines instead of commenting them out.

src/main/java/edu/harvard/iq/dataverse/DvObjectServiceBean.java
```diff
@@ -116,6 +116,16 @@ public DvObject findByAltGlobalId(GlobalId globalId, DvObject.DType dtype) {
         return runFindByGlobalId(query, globalId, dtype);
     }
 
+    public Long findIdByGlobalId(GlobalId globalId, DvObject.DType dtype) {
```
qqmyers, line 119: Same - this is a duplicate of the two methods right above them?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
```diff
@@ -142,13 +142,7 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         // TODO: this needs to be moved in to an onSuccess method; not adding to this PR as its out of scope
         // TODO: switch to asynchronous version when JPA sync works
```
qqmyers, line 143: The TODO is done?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDatasetVersionCommand.java
```diff
@@ -66,7 +66,7 @@ public DatasetVersion execute(CommandContext ctxt) throws CommandException {
         prepareDatasetAndVersion();
         
         // TODO make async
-        // ctxt.index().indexDataset(dataset);
+        // ctxt.index().asyncIndexDataset(dataset);
```
qqmyers, line 69: I guess the TODO would be done if this weren't a comment? Maybe just delete both lines?

qqmyers (approved):
Looks good. Lots of much needed optimization! I've asked for a few code comments and some cleanup, but otherwise I think it looks good to move forward. FWIW: I hope to test this over at QDR in the near future - some informal QA.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -105,6 +106,29 @@ public Dataset find(Object pk) {
...
+            .setHint("eclipselink.left-join-fetch", "o.files.thumbnailForDataset")
+            .setHint("eclipselink.left-join-fetch", "o.files.dataTables")
+            .setHint("eclipselink.left-join-fetch", "o.files.auxiliaryFiles")
+            .setHint("eclipselink.left-join-fetch", "o.files.ingestReports")
+            .setHint("eclipselink.left-join-fetch", "o.files.dataFileTags")
+            .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas")
+            .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas.fileCategories")
+            .setHint("eclipselink.left-join-fetch", "o.files.guestbookResponses")
+            .setHint("eclipselink.left-join-fetch", "o.files.embargo")
+            .setHint("eclipselink.left-join-fetch", "o.files.fileAccessRequests")
```
qqmyers, line 122: @ErykKul  - FWIW - at QDR, I've had to remove this line in two different versions. QDR is a fork, but I think the key issue is probably that in once case, the code is ~5.13 which is before the FileAccessRequest class was created to add the request time (a recent PR) and in the other my db has an additional column on the fileaccessrequests table due to some experimenting on another PR (#9599). I don't think that affects this indexing PR directly but - does that make sense? other than performance, is it true that dropping this hint (or any hint?) shouldn't affect anything? Or is this unique because we're not using fileaccessrequests when indexing? Any thoughts welcome.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
ErykKul: I have added comments to that method. As for the lock mode, this is just explicit default ostrich locking (default for em.find() and named query executions). I made it explicit to think about it later. I think this is always OK when reading and writing in a single transaction. Otherwise, you can have behind-your-back modification problems, which should not be common in the Dataverse context, since we do not expect multiple users working on the same dataset simultaneously.
I have removed the setLockMode line just in case when there could be a side-effect I am not aware of. Also, this should prevent any confusion about it.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
ErykKul: I did remove these comments now. I did not compare the two methods directly, but judging from the description of the experimental method, it was optimized for making as few db queries as possible. This is exactly what findDeep() does. I also did not notice any problems with findDeep() and all datasets load fast now. The main advantage of findDeep() is that it can also be used in write scenarios (the experimental method was only used in the read-only cases). It can also be used outside the dataset page, like it is now also used for indexing. I am now more certain of the findDeep() method (after using in production for several days now), so I dared to remove the experimental method from the code.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
ErykKul: Comments removed.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDatasetVersionCommand.java
ErykKul: It was not clear to me why it is in comment in the first place. It might have been an "optimization", or a placeholder for when async indexing would be implemented. I did uncomment it, but I have also placed it after the new version is stored, such that we can be user that the new version will not be missed by the indexing.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessage.java
```diff
@@ -32,49 +31,21 @@
...
-    public static final int INGEST_MESAGE_LEVEL_INFO = 2;
-
     /** Creates a new instance of IngestMessage */
-    public IngestMessage()  {
-        this(INGEST_MESAGE_LEVEL_INFO);
-    }
 
-    public IngestMessage(int messageLevel)  {
-        this.messageLevel = messageLevel;
+    public IngestMessage()  {
```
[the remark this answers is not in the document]
ErykKul: Yes. I did clean it up because it was very confusing of how the message works. I think it changed a lot over the years, but nobody cleaned it. When I was looking at it, I have lost some time to understand it. The next person should have less hard time now, since the unused code is gone.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessageBean.java
```diff
@@ -69,25 +68,30 @@ public IngestMessageBean() {
...
         IngestMessage ingestMessage = null;
 
-        Long datafile_id = null;
         AuthenticatedUser authenticatedUser = null;
         
         try {
             ObjectMessage om = (ObjectMessage) message;
             ingestMessage = (IngestMessage) om.getObject();
 
+            // if the lock was removed while an ingest was queued, ratake the lock
```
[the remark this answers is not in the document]
ErykKul: The "if" is the first thing that addDatasetLock method does. It has some complexity and would result in the code duplication if repeated here.
If that check would be removed from the addDatasetLock method in the future without updating the code using this method, ingest code would still not break because we remove "all" ingest locks at the end (right now, there can be at most one ingest lock).
I have extended the comment to clarify this.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
ErykKul: The hints do result in improved performance because eclipse link picks them up and respects them. They do not influence functionality. Indexing does not use the FileAccessRequest, so its performance would not be affected. In the dataset page it will be, since it does use FileAccessRequest (this is why I have added the hint, I have noticed thousands of queries retrieving the FileAccessRequest objects before showing the page with a large dataset to the user).

ErykKul:
@qqmyers 
I went through your comments and have done the changes. I consider the current version frozen, unless some bugs will show up or someone will request some more changes. Thanks for reviewing!

pdurbin:
> [ERROR]   DatasetsIT.testCuratePublishedDatasetVersionCommand:2803 Expected status code <200> doesn't match actual status code <500>.

Jenkins is showing this on develop as of 5394a9c: https://jenkins.dataverse.org/job/IQSS-dataverse-develop/1358/consoleFull

That line is blank, though. Strange. https://github.com/IQSS/dataverse/blob/5394a9c9ddf4231109faeee341ad0ab8d19b132a/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java#L2803

ErykKul:
It looks like the code was not waiting for index to finish, I have added the sleepForReindex call, it should fix this integration test.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessageBean.java
```diff
@@ -23,7 +23,6 @@
 import edu.harvard.iq.dataverse.*;
 import edu.harvard.iq.dataverse.authorization.AuthenticationServiceBean;
 import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
-import edu.harvard.iq.dataverse.util.BundleUtil;
```
qqmyers, line 26: @ErykKul  - This is needed for lines 140, 142 - causing compile errors.

qqmyers:
@ErykKul - there are 6 test failures in SearchIT where it appears that the sleepForReindex call is just timing out after 5 seconds with hasStaleIndex still true. I'm not sure if that means Jenkins is just that slow or if there's some way indexing isn't getting done. The other failure in DatasetsIT is one that appears in the develop branch as well so is probably not related to this PR:
image

qqmyers:
Not sure I can match an error to each test but I see errors like the following in the server.log from around the time of the failing SearchIT tests: [#|2023-05-31T14:47:15.381+0000|INFO|Payara 5.2022.3|edu.harvard.iq.dataverse.engine.command.impl.FinalizeDatasetPublicationCommand|_ThreadID=227;_ThreadName=__ejb-thread-pool13;_TimeMillis=1685544435381;_LevelValue=800;|
  Finalizing publication of the dataset doi:10.5072/FK2/9G1MU4|#]

[#|2023-05-31T14:47:15.479+0000|INFO|Payara 5.2022.3|edu.harvard.iq.dataverse.engine.command.impl.FinalizeDatasetPublicationCommand|_ThreadID=227;_ThreadName=__ejb-thread-pool13;_TimeMillis=1685544435479;_LevelValue=800;|
  Successfully published the dataset doi:10.5072/FK2/9G1MU4|#]

[#|2023-05-31T14:47:23.947+0000|WARNING|Payara 5.2022.3|edu.harvard.iq.dataverse.DataverseServiceBean|_ThreadID=71;_ThreadName=http-thread-pool::http-listener-1(1);_TimeMillis=1685544443947;_LevelValue=900;|
  Unable to find a single dataverse using alias "dvcdd2f3c2": javax.persistence.NoResultException: getSingleResult() did not retrieve any entities.|#]

[#|2023-05-31T14:47:24.180+0000|WARNING|Payara 5.2022.3|javax.enterprise.ejb.container|_ThreadID=229;_ThreadName=__ejb-thread-pool15;_TimeMillis=1685544444180;_LevelValue=900;_MessageID=AS-EJB-00056;|
  A system exception occurred during an invocation on EJB DatasetServiceBean, method: public edu.harvard.iq.dataverse.Dataset edu.harvard.iq.dataverse.DatasetServiceBean.findDeep(java.lang.Object)|#]

[#|2023-05-31T14:47:24.181+0000|WARNING|Payara 5.2022.3|javax.enterprise.ejb.container|_ThreadID=229;_ThreadName=__ejb-thread-pool15;_TimeMillis=1685544444181;_LevelValue=900;|
  javax.ejb.TransactionRolledbackLocalException: Exception thrown from bean: javax.persistence.NoResultException: getSingleResult() did not retrieve any entities.
	at com.sun.ejb.containers.EJBContainerTransactionManager.checkExceptionClientTx(EJBContainerTransactionManager.java:615)
	at com.sun.ejb.containers.EJBContainerTransactionManager.postInvokeTx(EJBContainerTransactionManager.java:485)
	at com.sun.ejb.containers.BaseContainer.postInvokeTx(BaseContainer.java:4601)
	at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2134)
	at com.sun.ejb.containers.BaseContainer.postInvoke(BaseContainer.java:2104)
	at com.sun.ejb.containers.EJBLocalObjectInvocationHandler.invoke(EJBLocalObjectInvocationHandler.java:220)
	at com.sun.ejb.containers.EJBLocalObjectInvocationHandlerDelegate.invoke(EJBLocalObjectInvocationHandlerDelegate.java:90)
	at com.sun.proxy.$Proxy302.findDeep(Unknown Source)
	at edu.harvard.iq.dataverse.__EJB31_Generated__DatasetServiceBean__Intf____Bean__.findDeep(Unknown Source)
	at edu.harvard.iq.dataverse.search.IndexServiceBean.indexDataset(IndexServiceBean.java:440)
	at edu.harvard.iq.dataverse.search.IndexServiceBean.asyncIndexDataset(IndexServiceBean.java:415)
	at jdk.internal.reflect.GeneratedMethodAccessor364.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.glassfish.ejb.security.application.EJBSecurityManager.runMethod(EJBSecurityManager.java:588)
	at org.glassfish.ejb.security.application.EJBSecurityManager.invoke(EJBSecurityManager.java:408)
	at com.sun.ejb.containers.BaseContainer.invokeBeanMethod(BaseContainer.java:4835)
	at com.sun.ejb.EjbInvocation.invokeBeanMethod(EjbInvocation.java:665)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:834)
	at com.sun.ejb.EjbInvocation.proceed(EjbInvocation.java:615)
	at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.doCall(SystemInterceptorProxy.java:163)
	at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.aroundInvoke(SystemInterceptorProxy.java:140)
	at jdk.internal.reflect.GeneratedMethodAccessor273.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.ejb.containers.interceptors.AroundInvokeInterceptor.intercept(InterceptorManager.java:888)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:833)
	at com.sun.ejb.EjbInvocation.proceed(EjbInvocation.java:615)
	at org.jboss.weld.module.ejb.AbstractEJBRequestScopeActivationInterceptor.aroundInvoke(AbstractEJBRequestScopeActivationInterceptor.java:81)
	at org.jboss.weld.module.ejb.SessionBeanInterceptor.aroundInvoke(SessionBeanInterceptor.java:52)
	at jdk.internal.reflect.GeneratedMethodAccessor140.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.ejb.containers.interceptors.AroundInvokeInterceptor.intercept(InterceptorManager.java:888)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:833)
	at com.sun.ejb.containers.interceptors.InterceptorManager.intercept(InterceptorManager.java:375)
	at com.sun.ejb.containers.BaseContainer.__intercept(BaseContainer.java:4807)
	at com.sun.ejb.containers.BaseContainer.intercept(BaseContainer.java:4795)
	at com.sun.ejb.containers.EjbAsyncTask.call(EjbAsyncTask.java:101)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: javax.persistence.NoResultException: getSingleResult() did not retrieve any entities.
	at org.eclipse.persistence.internal.jpa.QueryImpl.throwNoResultException(QueryImpl.java:985)
	at org.eclipse.persistence.internal.jpa.QueryImpl.getSingleResult(QueryImpl.java:544)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.getSingleResult(EJBQueryImpl.java:404)
	at edu.harvard.iq.dataverse.DatasetServiceBean.findDeep(DatasetServiceBean.java:138)
	at jdk.internal.reflect.GeneratedMethodAccessor365.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.glassfish.ejb.security.application.EJBSecurityManager.runMethod(EJBSecurityManager.java:588)
	at org.glassfish.ejb.security.application.EJBSecurityManager.invoke(EJBSecurityManager.java:408)
	at com.sun.ejb.containers.BaseContainer.invokeBeanMethod(BaseContainer.java:4835)
	at com.sun.ejb.EjbInvocation.invokeBeanMethod(EjbInvocation.java:665)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:834)
	at com.sun.ejb.EjbInvocation.proceed(EjbInvocation.java:615)
	at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.doCall(SystemInterceptorProxy.java:163)
	at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.aroundInvoke(SystemInterceptorProxy.java:140)
	at jdk.internal.reflect.GeneratedMethodAccessor273.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.ejb.containers.interceptors.AroundInvokeInterceptor.intercept(InterceptorManager.java:888)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:833)
	at com.sun.ejb.EjbInvocation.proceed(EjbInvocation.java:615)
	at org.jboss.weld.module.ejb.AbstractEJBRequestScopeActivationInterceptor.aroundInvoke(AbstractEJBRequestScopeActivationInterceptor.java:72)
	at org.jboss.weld.module.ejb.SessionBeanInterceptor.aroundInvoke(SessionBeanInterceptor.java:52)
	at jdk.internal.reflect.GeneratedMethodAccessor140.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.ejb.containers.interceptors.AroundInvokeInterceptor.intercept(InterceptorManager.java:888)
	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:833)
	at com.sun.ejb.containers.interceptors.InterceptorManager.intercept(InterceptorManager.java:375)
	at com.sun.ejb.containers.BaseContainer.__intercept(BaseContainer.java:4807)
	at com.sun.ejb.containers.BaseContainer.intercept(BaseContainer.java:4795)
	at com.sun.ejb.containers.EJBLocalObjectInvocationHandler.invoke(EJBLocalObjectInvocationHandler.java:212)
	... 37 more
|#]

[#|2023-05-31T14:47:30.485+0000|WARNING|Payara 5.2022.3|edu.harvard.iq.dataverse.DataverseServiceBean|_ThreadID=71;_ThreadName=http-thread-pool::http-listener-1(1);_TimeMillis=1685544450485;_LevelValue=900;|
  Unable to find a single dataverse using alias "dv00d2af04": javax.persistence.NoResultException: getSingleResult() did not retrieve any entities.|#]

qqmyers:
FWIW: #9632 fixes the problem in DatasetsIT. So that failure should go away once that PR is merged and merged here.

qqmyers:
Random thought - could making indexing async mean that indexing for many prior tests are backing up and when you need to see indexing results, the wait could need to be much longer than 5 seconds (to let the queue drain?).

poikilotherm:
> Random thought - could making indexing async mean that indexing for many prior tests are backing up and when you need to see indexing results, the wait could need to be much longer than 5 seconds (to let the queue drain?). 

Sorry for asking potentially dumb questions: would it help if we can monitor the queue better? 

There's MicroProfile Metrics builtin with Payara. Adding an application metric isn't that hard. We could check via REST if the queue might be a problem even before running the test / commit the workload.

Here's a link to the annotation docs: https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0/microprofile-metrics-spec-5.0.0.html#api-annotations

poikilotherm:
I played a little with this in an experimentation branch. Here's how you add such metrics: https://github.com/poikilotherm/dataverse/commit/0389b2f01bfac14fbeaa932a7e43051c3637e42f (Really easy once you learn to work around the edges.)

Running `SearchIT` on my machine does not build up a long queue, even with refreshing every 0.1sec:
```
watch -n0.1 http -j localhost:8080/metrics/application
```
(Uses httpie, curl is similar)

poikilotherm:
> The maximum size of the queue is 1, monitoring it would not provide a lot of information. I was thinking that maybe setting the lock mode to none for the find deep query might have more impact than what I thought. I have restored it to see if it changes anything.

You mean in this case the queue is limited to one because of how our tests look like, right? Because in a realworld scenario with multiple users etc that should be different.

FWIW with regards to creating insights about impact of optimizations: there are the `@Timed` and `@Metered` annotations in that metrics package. They provides nice measurements how long things take including min/max/avg. Just add it to a method. Here's an example how this can look like: https://www.adam-bien.com/roller/abien/entry/microprofile_metered_vs_timed

poikilotherm:
@ErykKul here's the result of the last run:

grafik

Still saying "INFO: Still indexing after 5 seconds"

ErykKul:
I think this is the worst case scenario: https://guides.dataverse.org/en/latest/admin/solr-search-index.html#full-reindex
```
curl http://localhost:8080/api/admin/index/clear
curl http://localhost:8080/api/admin/index
```
All datasets are indexed async, and all at once. We do not have many datasets (84), so it works fine for us. If there is a problem with multiple users scenarios, reindexing a large collection should show it. Can someone with bigger collections try it out?

qqmyers:
FWIW: In talking about the queue, I wasn't referring to how many indexing jobs for one dataset there can be, but how many datasets can get lined up for @Async processing. The tests create lots of different datasets that have to be indexed and now with that being async, the next test runs immediately. For the search tests, I was thinking that the pool of other datasets waiting to be indexed could be big enough to make it longer than 5 seconds before indexing for the search test gets done. Hopefully, it would still get done/it's not a flaw in the code, but the 5 second wait that was sufficient before isn't long enough now. (So reindexAll would still finish, probably faster than before, but the time delay from when indexing is triggered for the last dataset and when it's indexing actually runs and completes could be longer. Still just guessing - I haven't tested anything.

poikilotherm:
> For the search tests, I was thinking that the pool of other datasets waiting to be indexed could be big enough to make it longer than 5 seconds before indexing for the search test gets done.

I understood it that way, but didn't check on it on my machine (which would be pointless anyway, as the EC2 machine is the bottleneck anyway). I just added BOTH metrics, so you can check with a simple REST call if the queue is empty or not before executing the test (e.g. in a `@BeforeAll`).

qqmyers:
Looks good! - the only failing test is the DatasetsIT one that is fixed in the other PR.

ErykKul:
Great. I will leave it like this. The findDeep calls are not essential there.

landreev:
BTW, I grabbed this PR for review earlier today, without realizing that there had been an ongoing back-and-forth since April. @qqmyers, @poikilotherm if either of you are ready to approve it, please go ahead and don't wait for me. Otherwise I'm reading it with great interest. 
Thank you, @ErykKul!

landreev:
P.S. I mean, I see that it was already approved back in April, but a lot has been done since then; and I'm assuming that's why it was moved back into "In Review".

qqmyers:
I was just waiting on the commits from June 1 that resolved some minor requests and I see there are more commits from June 9th. Assuming those are all OK, I was going to approve and move it back.

ErykKul:
@landreev @qqmyers 
The changes as requested are all there. The later commits (from June 9th) fixed the integration test issues. I think that all changes are OK now and can be merged. I will be still watching this closely, just in case something pops up (e.g., I am a little bit worried about re-indexing of large Dataverse collections, but normally it should be fine and even work better after this PR).

qqmyers:
OK - looks good! No more changes unless requested in QA!

qqmyers:
FWIW: It looks like the CI failure is the problem with the availability of payara-5.2022.3.zip that was discussed in #dv-tech today and apparently fixed. @kcondon - you may want to rerun to verify that the build now works before you dig in.

kcondon:
Quick test so far shows adding 1 or more files to a dataset containing 1000 files is closer to adding a single file in performance: 6.71s for adding an additional file to 1000 ds before this pr, 2.94s to add a file with this pr. Adding 3 files to the 1000ds with this pr took 3.44s. Note these are rough times. Can repeat with a larger dataset.

PR #9568, opened 2023-04-28
9148 Ensure that license/terms are set via API (create/update/publish)

### LINKED ISSUE #9148: Ensure that datasets published using the Dataverse APIs include a license or have Terms of Use

API User, Curator, Superuser, Sysadmin

Noticing that after the multiple license update (v.5.10) was applied to the Harvard Dataverse Repository, some datasets have been published using the Dataverse APIs without a license (CC0 is the only one available in the dropdown for now) and with nothing entered in the Terms of Use field.

The GUI enforces the requirement to choose a license from the License/Data Use Agreement dropdown or enter something in the Terms of Use field, in order to ensure that the metadata includes some information that makes explicit how the data can be used, but the API doesn't enforce this requirement.

Depositors publishing datasets using the APIs must include a license or add something in the Terms of Use field, to match the rules enforced by the GUI.

https://github.com/IQSS/dataverse/issues/7440

### LINKED ISSUE #9155: Bug report: Json API doesn't allow round-trip of license information
Title updated. In discussion this issue we discovered that the root cause is[ a bug/mismatch between the JSON export and import format for licenses](https://github.com/IQSS/dataverse/issues/9155#issuecomment-1320363684). A resolution to this issue should address that. (Note there is a workaround.)

Dear Dataverse developers,

Unfortunately, I could not find a possibility to attach Terms to a dataset using API. 

API User

There is a section in den docs explaining how to manage the tems dataverse-wide[1], however not how to attach them to datasets.
[1] https://guides.dataverse.org/en/latest/api/native-api.html#id202

When I try to do it the same way that dataverse respond, the data is getting accepted, however there is no result and the dataset is created with "custom terms"
````
 {
        "metadataLanguage": "en",
        "datasetVersion": {
            "license": {
                "name": "CC0 1.0",
                "uri": "http://creativecommons.org/publicdomain/zero/1.0"
            },
            "metadataBlocks": {.......}
        }
    }
````
I would expect that this json would change the license to CC0

All issues about Terms are somehow related, I guess..
Thank you very much
Sergej

### LINKED ISSUE #9364: Creating Datasets via API Should Use Default License - if None Specified
Use the example **dataset-finch1.json** from the documentation: https://guides.dataverse.org/en/latest/api/native-api.html#create-a-dataset-in-a-dataverse-collection
https://guides.dataverse.org/en/latest/_downloads/fc56af1c414df69fd4721ce3629f0c03/dataset-finch1.json

Use the API to create a dataset.
`curl -H "X-Dataverse-key:xxxxxxxx-xEJEJEJEJEE" -X POST "https://demo.dataverse.org/api/dataverses/root/datasets" --upload-file "dataset-finch1.json" -H 'Content-type:application/json'`

* What happens?
The dataset is created, but without the default license - assumes "Custom License" see screenshot below. 

All users.

Since the json file did not have any license specified, I expected the CC-0 license (default) to be automatically added to the dataset.

5.12.1 (on demo.dataverse.org)

- https://github.com/IQSS/dataverse/issues/9155
- https://github.com/IQSS/dataverse/issues/3357

**Screenshots:**

Here is what the dataset shows after create via API:
Screen Shot 2023-02-03 at 9 11 17 AM

### PR DESCRIPTION
**Pull request by landreev**

See the issue for details. 
Basic idea is that we want the API behavior to match the UI. I.e., we want to make it impossible to publish a dataset that doesn't have either a configured license or custom terms of use. 

This has been achieved by making the following changes: 

- The publish API is now going to fail if you try to publish a dataset without either of the above. 
- But, it is now possible, or not-entirely-counter-intuitive to actually create a dataset via the API with a standard license, as well as to modify it via the API without losing said license. That's because the following bugs that used to make this problematic have been fixed: (the following is from the release note)

We fixed the incompatibility between the format in which license information was *exported* in json, and the format the Create and Update APIs were expecting it for *import* (https://github.com/IQSS/dataverse/issues/9155). This means that the following json format can now be imported:
```
"license": {
   "name": "CC0 1.0",
   "uri": "http://creativecommons.org/publicdomain/zero/1.0"
}
```
However, for the sake of backward compatibility the old format
```
"license" : "CC0 1.0"
```
will be accepted as well.

We have added the default license (CC0) to the model json file that we provide and recommend to use as the model in the Native API Guide (https://github.com/IQSS/dataverse/issues/9364).

And we have corrected the misleading language in the same guide where we used to recommend to users that they select, edit and re-import only the `.metadataBlocks` fragment of the json metadata representing the latest version. There are in fact other useful pieces of information that need to be preserved in the update (such as the `"license"` section above). So the recommended way of creating base json for updates via the API is to select *everything but* the `"files"` section, with (for example) the following `jq` command:

```
jq '.data | del(.files)'
```

Closes #9148 
Closes #9364 
Closes #9155

Please see if there is anything I miss, if the logic above, and the change in the guide and the release note make sense. Whoever is going to be reviewing this, please consult with @jggautier if you have any questions or doubts about what we are doing in this PR. 
As submitted, we are definitely introducing a backward incompatibility - some users' API scripts that have been publishing datasets successfully will stop doing so once this change is deployed, until they make some (simple) modifications. Input and/or strong opinions are welcome. But please keep in mind that Julian has already done some research and outreach to the users who have been publishing such datasets in our production, and most or all appear to be interested in being able to publish them with standard licenses, just like our UI users. (as you would expect, nobody appears to have been doing it knowingly, it's just a function of our idiosyncratic APIs). 

This will require re-testing the Create, Update and Publish dataset APIs, per the information in the release note and the updated section of the guide. 

**Additional documentation**:

### CODE REVIEW

**Review by sekmiller — APPROVED** [review]
Fixed a minor typo and updated from dev in Leonid's absence.

**Comment by jggautier** [comment]
Maybe just for my own clarification about maintaining backwards compatibility, does this mean that using a license's URI, instead of the license name, will still work, too? For example, would this work?:
```
"license" : "http://creativecommons.org/publicdomain/zero/1.0"
```

pdurbin:
This merged. It just shows closed.

PR #9582, opened 2023-05-09
9469 accented letters and other utf8 characters in Stata ingest

### LINKED ISSUE #9469: Ingest of Stata file with diacritics characters produces not correct encoded variable metadata
After the upload of Stata files (Version SE17, .dta-format 118 from Stata 14, 15, 16, or 17) the ingest runs correctly. The original binary and tab file can be downloaded correctly. But the metadata of variables isn’t correct encoded. Diacritics are shown as unicode replacement character ��. Variable & value level are affected.
e.g. https://data.aussda.at/api/datasets/export?exporter=ddi&persistentId=doi%3A10.11587/QQ7HTL  
\<var ID="v31494" name="G1" intrvl="discrete">
\<location fileid="f472"/>
\<labl level="variable">laufende Nummer des Gespr��chsturns</labl>
		
\<var ID="v31501" name="G3" intrvl="discrete">
\<location fileid="f472"/>
\<catgry>
\<catValu>201</catValu>
\<labl level="category">Alois Mock (��VP)</labl>
Note: In this metadata export example there are also correct diacritics from other data files, csv and SPSS ingest is correct.

This error can only be found in not restricted data files, where variable metadata can be exported.
But the search can find other corrupted characters, the variable metadata is also ingested from restricted files
https://data.aussda.at/dataverse/AUSSDA/?q=%EF%BF%BD%EF%BF%BDsterreich
 
I think it has something to do with https://en.wikipedia.org/wiki/Variable-width_encoding 
I can also exclude that the Stata files are corrupted, attached a small test file which generates the same error.

Dataverse 5.12.1
the first issue concern the same Dataverse instance

- https://github.com/IQSS/dataverse/issues/7507 
- https://github.com/IQSS/dataverse/issues/6675

[stata14_ingest_diacritics_numericVariable.zip](https://github.com/IQSS/dataverse/files/11051102/stata14_ingest_diacritics_numericVariable.zip)

landreev:

This a 5 line fix for the Stata ingest plugin that I made some weeks ago when looking into a report from a user (#9469). Making a (draft) pr from the branch so that it's not forgotten. 

The problem is straightforward, accented characters are garbled in the variable metadata labels (both the variable-level and the category value labels). (Only in the metadata! The values in the tab-delimited files are saved properly). 

Closes #9469

A Stata file from the remote dataset from the original user report can be used for testing: https://data.aussda.at/file.xhtml?fileId=472&version=3.0 The file has both types of labels with accented characters in them. 

The test is to ingest the file and look at the variable labels as exported in the DDI. 
before: 
```
<labl level="variable">laufende Nummer des Gespr��chsturns</labl>
```
after: 
```
<labl level="variable">laufende Nummer des Gesprächsturns</labl>
```
etc.

**Additional documentation**:


qqmyers (approved):
LGTM

PR #9586, opened 2023-05-10
#9565 Fix metrics API "file/byType/monthly" by using specific key cache

### LINKED ISSUE #9565: Metrics API "file/byType/monthly" may not work
Hi,

If I call metrics API `metrics/file/byType`, and after `metrics/file/byType/monthly` on the same collection, the result on `metrics/file/byType/monthy` doesn't respond correctly. On Recherche Data Gouv repository response have a HTTP 500 error code with a `NullPointerException` on the log file. On Dataverse Demo repository, the response code is 200 but the date is `null`.

I think the problem comes from using the same key in the database cache as the endpoint `metrics/file/byType` : if I call first `metrics/file/byType/monthly` the problem doesn't occur because the cache `filesByType` store the date information.

First I call `metrics/files/byType`.

```
curl https://demo.dataverse.org/api/info/metrics/files/byType?parentAlias=dataverseDeExemplo
```
And after I call `metrics/files/byType/monthly`.

```
curl https://demo.dataverse.org/api/info/metrics/files/byType/monthly?parentAlias=dataverseDeExemplo
```

* When does this issue occur?

When I call first `metrics/files/byType` and after `metrics/files/byType/monthly`.

* What happens?

The information date is `null` or the endpoint respond with a HTTP 500 error code with a `NullPointerException` on [FileUtil.java](https://github.com/IQSS/dataverse/blob/v5.13/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java#L2144).
If want to see the JSON response, I will see the information date is absent but the `NullPointerException` is absent : 

```
curl -s -H "Accept: application/json" https://demo.dataverse.org/api/info/metrics/files/byType/monthly?parentAlias=dataverseDeExemplo
```

This difference comes in the source code in [Metrics.java](https://github.com/IQSS/dataverse/blob/v5.13/src/main/java/edu/harvard/iq/dataverse/api/Metrics.java#L426) because the return type is `application/json`.

All users

If I call first `metrics/file/byType`, and after `metrics/file/byType/monthly`, I can see the date datas on last call.

5.12.1 and 5.13

jeromeroucou:

Fix metrics API "file/byType/monthly" by using specific key cache

Closes #9565 

Call first `metrics/files/byType` :

```
curl https://$SERVER_URL/api/info/metrics/files/byType?parentAlias=$ALIAS
```

And after call `metrics/files/byType/monthly` :

```
curl https://$SERVER_URL/api/info/metrics/files/byType/monthly?parentAlias=$ALIAS
```

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Metrics.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Metrics.java
@@ -412,7 +412,7 @@ public class Metrics extends AbstractApiBean {
         } catch (IllegalArgumentException ia) {
             return error(BAD_REQUEST, ia.getLocalizedMessage());
         }
-        String metricName = "filesByType";
+        String metricName = "filesByTypeMonthly";
 
         JsonArray jsonArray = MetricsUtil.stringToJsonArray(metricsSvc.returnUnexpiredCacheAllTime(metricName, null, d));
 


qqmyers (approved):
Looks good - thanks!

Issue #9264, opened 2023-01-05
give contributors guidance on how to write a release note snippet


pdurbin:
Currently, the information is buried here: https://guides.dataverse.org/en/5.12.1/developers/making-releases.html#write-release-notes

We should add something to the "how to make a pull request" docs: https://guides.dataverse.org/en/5.12.1/developers/version-control.html#how-to-make-a-pull-request (this is a bit buried too, by the way).

We might also want to add something here: https://guides.dataverse.org/en/5.12.1/developers/documentation.html#quick-fix

pdurbin:
Here's the file to edit: https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/developers/version-control.rst

Suggestions on how to edit the file: http://preview.guides.gdcc.io/en/develop/developers/documentation.html

bencomp:
And if you want to stimulate (or enforce) the addition of release notes, you could use a GitHub workflow to suggest or check for them. I am thinking something like "if PR notes say release notes are needed, then check if a corresponding file exists and respond with 'thank you!'. Else, gently remind the contributor that no release notes were found."

pdurbin:
Sure, maybe a reach goal for this issue. 😄 

But just some docs is also fine.

sbondka:
Could you please provide me with more information regarding this issue? I have started making changes in the "Commit Your Change to Your New Branch" and "Find or Create a GitHub Issue" sections of the file "https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/developers/version-control.rst". Should I also include in the "Make a Pull Request" section the automated process related to adding release notes?

pdurbin:
@sbondka yes, that would be a perfect place for it.

However, the process isn't automated. Contributors write snippets and later someone manually combines them.

Please feel free to hit me up on https://chat.dataverse.org with any questions! Thanks for your interest in this!

pdurbin:
Whoops, now I see that you already created this PR:

- #10203

I'll make some suggestions or edits. Thanks!

PR #9596, opened 2023-05-16
add MyData API to dataverse documentaion

### LINKED ISSUE #5042: Document - Make use of MyData API "official"
The MyData API works !! but is undocumented (not in any guides) and not integrated into the Dataverse tests.

The MyData API is a good option for getting a list of just the datasets & dataverses an authenticated user can edit. Per Phil, here's the scope of what needs to be done:

- Document in the API Guide how to use the My Data API.
- Write integration tests for the My Data API so that it is exercised regularly.

sirineREKIK:
add MyData API to dataverse documentaion

- closes #5042

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -4538,9 +4538,6 @@ MyData
 
 The MyData API is used to get a list of just the datasets, dataverses or datafiles an authenticated user can edit.
 
-MyData API
-~~~~~~~~~~
-
 A curl example listing objects
 
 .. code-block:: bash
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -4507,12 +4507,12 @@ security.
 
 
 MyData
------
+------
 
 The MyData API is used to get a list of just the datasets, dataverses or datafiles an authenticated user can edit.
 
 MyData API
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~
 
 List objects::
 
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -4556,8 +4556,9 @@ A curl example listing objects
 
 Parameters:
 
-``key`` is the user token, for this API is must not be passed in the header.
-``role_id`` User roles, several possible values among:
+``key`` Is the user token, for this API is must not be passed in the header.
+
+``role_id`` Roles are customizable. Standard roles include:
 
 - ``1`` = Admin
 - ``2`` = File Downloader
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -4506,6 +4506,33 @@ Please see :ref:`dataverse.api.signature-secret` for the configuration option to
 security.
 
 
+.. _send-feedback:
+
+Send Feedback To Contact(s)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This API call allows sending an email to the contacts for a collection, dataset, or datafile or to the support email address when no object is specified.
+The call is protected by the normal /admin API protections (limited to localhost or requiring a separate key), but does not otherwise limit the sending of emails.
+Administrators should be sure only trusted applications have access to avoid the potential for spam.
+
+The call is a POST with a JSON object as input with four keys:
+- "targetId" - the id of the collection, dataset, or datafile. Persistent ids and collection aliases are not supported. (Optional)
+- "subject" - the email subject line
+- "body" - the email body to send
+- "fromEmail" - the email to list in the reply-to field. (Dataverse always sends mail from the system email, but does it "on behalf of" and with a reply-to for the specified user.)
+
+A curl example using an ``ID``
+
+.. code-block:: bash
+
+  export SERVER_URL=http://localhost
+  export JSON='{"targetId":24, "subject":"Data Question", "body":"Please help me understand your data. Thank you!", "fromEmail":"dataverseSupport@mailinator.com"}'
+
+  curl -X POST -H 'Content-Type:application/json' -d "$JSON" $SERVER_URL/api/admin/feedback
+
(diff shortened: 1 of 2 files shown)


pdurbin:
@sirineREKIK thanks for the pull request! Can you please take a look at these errors? https://github.com/IQSS/dataverse/actions/runs/4989177823/jobs/8932750403?pr=9596

Thanks!

(I'm giving this a size 3 because it shouldn't take much effort to review and QA.)

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4530,3 +4531,46 @@ A curl example using an ``ID``
...
+  export DVOBJECT_TYPES=Dataset
+  export PUBLISHED_STATES=Unpublished
+  export PER_PAGE=10
+
+  curl -H GET http://$SERVER_URL/api/mydata/retrieve?key=$API_TOKEN&role_ids=$ROLE_IDS&dvobject_types=$DVOBJECT_TYPES&published_states=$PUBLISHED_STATES&per_page=$PER_PAGE
+
+Parameters:
+
+``key`` is the user token, for this API is must not be passed in the header.
+``role_id`` User roles, several possible values among:
```
qqmyers, line 4560: Roles are customizable. Standard roles include:

qqmyers (approved):
Looks good. I think my only minor comment is that roles are customizable (i.e. by https://guides.dataverse.org/en/4.15.1/api/native-api.html#id24) so the values shown as parameters are only examples, not all possible values. It would be nice to have a change before this is merged, but I don't think it should be slowed down for that.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4530,3 +4531,47 @@ A curl example using an ``ID``
...
+
+
+MyData
+------
+
+The MyData API is used to get a list of just the datasets, dataverses or datafiles an authenticated user can edit.
+
+MyData API
+~~~~~~~~~~
+
```
pdurbin, line 4543:
```suggestion
```

We can probably remove this second level. Here's how it looks at https://dataverse-guide--9596.org.readthedocs.build/en/9596/api/native-api.html right now:

Screenshot 2023-06-01 at 5 41 22 PM

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4530,3 +4531,47 @@ A curl example using an ``ID``
...
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+  export ROLE_IDS=6
+  export DVOBJECT_TYPES=Dataset
+  export PUBLISHED_STATES=Unpublished
+  export PER_PAGE=10
+
+  curl -H GET http://$SERVER_URL/api/mydata/retrieve?key=$API_TOKEN&role_ids=$ROLE_IDS&dvobject_types=$DVOBJECT_TYPES&published_states=$PUBLISHED_STATES&per_page=$PER_PAGE
```
pdurbin, line 4555: - -H with no header?
- sending -H with the API token is preferred to sending "key" (I just noticed the note about key below.) 🤔 
- I'd suggest removing GET since it's the default
- https is preferred over http
- I'd replace `http://$SERVER_URL` with `$SERVER_URL`
- I'd put quotes around the URL because of all the `&` characters. Bash will get confused.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4530,3 +4531,47 @@ A curl example using an ``ID``
...
+
+Parameters:
+
+``key`` Is the user token, for this API is must not be passed in the header.
```
pdurbin, line 4559: Is this true?

doc/sphinx-guides/source/api/native-api.rst
pdurbin, line 1: Can we have a release note snippet as well, please? For now (sorry), please see this issue:

- #9264

pdurbin (commented):
I haven't tested this and I'm wondering about the "key" thing but I left a little feedback. Thanks for working on this, @sirineREKIK! ❤️

sirineREKIK:
Can you check if everything is ok or if we need to do something else please?

pdurbin (approved):
https://dataverse-guide--9596.org.readthedocs.build/en/9596/api/native-api.html#mydata looks reasonable but I haven't tested it.

@kcondon if you could verify that indeed the API token can only be passes as a query parameter ("key") for this API rather than as header (curl -H ...), I would appreciate it!

sirineREKIK:
Hello,
@kcondon 
From our side, when doing tests to access the endpoint when passing the api_token ("key") in the header, we got an authentication error, you can see the example below:
login

sirineREKIK:
Hello,
@kcondon
From our side, when doing tests to access the endpoint when passing the api_token ("key") in the header, we got an authentication error, you can see the example below:

curl -H "X-Dataverse-key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" "https://demo.dataverse.org/api/v1/mydata/retrieve?dvobject_types=Dataverse&dvobject_types=Dataset&published_states=Published&role_ids=8&per_page=5" | jq .
{
  "success": false,
  "error_message": "Requires authentication.  Please login."
}

Also it is always the same return for all our installation environements.
I don't know if we are missing something, but as mentioned in the example, the test with the datavarese demo site is in error.

kcondon:
@sirineREKIK Thank you for that information. I was able to reproduce the problem. I still believe passing the key in the header is the correct syntax since all of my testing previous testing with endpoints has confirmed this is the correct method. However, I do not yet know what's wrong in this case. Also, please note that this syntax is working on my test server.

sirineREKIK:
@kcondon Thank you for that information. Changes are taken into account.
Regards, Sirine.

kcondon:
@sirineREKIK Thanks, it looks great!

pdurbin:
I'm glad the header works! Probably due to recent auth refactoring? 🤔 

- #9303

Not sure. 😄

kcondon:
@pdurbin Yes. Guillermo fixed it. I believe this unpublished api implementation predated our move to using the header as a more secure way of passing info. So, really was just never implemented, I think?

PR #9257, opened 2023-01-04
8092 timestamp of data access request

ericdevries:
**What this PR does / why we need it**:

This adds a date column to the restricted file access request overview, indicating when the earliest request by that user was made.

It also fixes an issue where the request list was not updated when a request was approved or rejected.

To reproduce this bug (outside this branch) perform the following steps:

* create a dataset with one or more restricted files
* sign in with a different user that does not have permissions already to download the files
* request access on the files
* sign back in as the owner of the dataset
* go to Edit Dataset > Permissions > Files
* approve or reject the request 
* notice how the data table with the approve and reject options does not disappear or update when clicked. 

**Which issue(s) this PR closes**:

- Closes #8092

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

Yes, it adds a date column to the restricted file access request overview. See screenshot below:

Screenshot 2023-01-04 at 18-27-26 Restricted File Permissions - gasdf

**Is there a release notes update needed for this change?**:

Yes, added.

**Additional documentation**:

None.

PR #9602, opened 2023-05-19
Fix Flyway script name

qqmyers:
**What this PR does / why we need it**:  #9257 added a flyway script that is misnamed - starting with V5.13.0.4__ while the next open slot as it ended up being merged is V5.13.0.1__. This PR makes the name change.

Closes #

**Special notes for your reviewer**: I verified that the script can be rerun.

**Suggestions on how to test this**: On an existing database, one has to drop the V5.13.0.4__* entry in the flyway_schema_history table to be able to redeploy Dataverse. One could just check that this works. Additionally one could check that a fresh deploy works (I can't see how this would fail) and/or verify that the contents of the fileaccessrequests table is the same before and after the redeploy (both content and the constraints, etc.).

**Additional documentation**:  When this is merged, developers should probably get the standard email about having to resolve the flyway conflict on existing databases.


pdurbin (approved):
Looks good to me.

The failing Jenkins job is something unrelated: TASK [dataverse : install NodeJS packages from package.json] ******************* https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9602/1/consoleFull

PR #9571, opened 2023-05-03
Internationalization of ingest messages

stevenferey:
**What this PR does / why we need it**:
Internationalization of ingest messages

**Which issue(s) this PR closes**:

- Closes #9570

PR #9618, opened 2023-05-25
Make DDIExportTest XML validation non-flaky

### LINKED ISSUE #9617: Fix flaky `DDIExporterTest.testExportDataset()`
By default, the assertion will load the schema to validate the DDI XML from the remote source announced in the generated XML.

This seems to cause flaky unit tests in CI, as the remote end seems to go away every now and then.

Fix by adding the schema files to the sources and use from there.

poikilotherm:

A lot of unit testing in Github Actions failed because the DDI XML schemas could not be downloaded and validation failed.

This PR adds the necessary files as test resources to the tree, so we don't need to download them. By loading them in the test we avoid the failures.

- Closes #9617 

@landreev I assigned you and requested your review as you flipped the switch to test for valid XML...

Here are some failed actions, where you can see the error: 
- https://github.com/IQSS/dataverse/actions/runs/5073375089/jobs/9112324489
- https://github.com/IQSS/dataverse/actions/runs/5070201120/jobs/9104893386
- https://github.com/IQSS/dataverse/actions/runs/5072343226/jobs/9109918346

 There are plenty more at https://github.com/IQSS/dataverse/actions/workflows/maven_unit_test.yml

Uhm yeah well... The tests should not fail anymore now. Just let 'em run. This fixes a test that failed for the wrong reasons...

No

No

**Additional documentation**:
None

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/export/DDIExporterTest.java
@@ -32,18 +32,25 @@ import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.logging.Logger;
 import javax.json.Json;
 import javax.json.JsonObject;
 import javax.json.JsonReader;
+import javax.xml.transform.Source;
+import javax.xml.validation.Schema;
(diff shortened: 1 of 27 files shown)


pdurbin (approved):
All tests are passing (including API tests). Approved. Thanks, @poikilotherm !

We were aware of this and saw it here, for example: https://github.com/IQSS/dataverse/pull/9571#issuecomment-1536138794

Much appreciated!

PR #9629, opened 2023-05-30
9338 Fixes Roles behavior in the UI and API

### LINKED ISSUE #9338: Inconsistent behavior relating to Roles

I found that the API and the web UI is not working similarly. I tested it in Dataverse v5.11.1. Here is a summary:

If I create new role ("reviewer") for the root dataverse, it is displayed only in the root's roles list. It is not displayed other dataverse' role list, and I can not create another role with the same machine name elsewhere.

Displaying a newly created role in the list of roles:

web:
- root dataverse: it is shown
- elsewhere: it is not shown

API:
- root dataverse: it is shown
- elsewhere: it is not shown

Assign role to user/group:

web:
- root dataverse: possible
- elsewhere: I can not assign the same rule in another dataverse, because it is not displayed among the selectable roles
API:
- root dataverse: possible
- elsewhere: possible

Checking the list of Users/Groups (assignments).

web:
- root dataverse: the assignment is shown in the list
- elsewhere: it doesn't.
API:
- root dataverse: the assignment is shown in the list
- elsewhere: the assignment is shown in the list

I haven't found an explanation in the Dataverse guide.

- Do you know if this inconvenience is by design? or it is rather a bug?
- What would be your expectation regarding to newly created roles (those which is not packed in a freshly installed Dataverse)? Mine would be that if I create a new role in the root, it is visible and usable elsewhere in both the web UI and API, but maybe you have different opinion.

* When does this issue occur?
see above

the Permission page

* What happens?
see above

superusers, dataverse admins

see above

v5.11.1

I am not aware of.

**Screenshots:**

*

landreev:

See the opening comment in the linked issue, it has a detailed description of the 3 major issues: Inherited roles (roles defined in the ancestral collections) not listed by either the API or UI; and then 2 inconsistencies between the API (`/api/dataverses/*/roles` and `/api/dataverses/*/assignments`) and the permissions page in the UI. 

Closes #9338

It really looks like it took very minimal changes after all, 2 lines basically: 
- Changed the `availableRoles()` method in `DataverseRoleServiceBean` to ignore the "permissionroot" boolean when looking for the extra Roles defined on the ancestral collections. 
- Changed the `/api/dataverses/*/roles` to use the service method above, instead of simply listing `definitionPoint.getRoles()`, for consistency between the two. 

(Did I miss anything?)

Please note the PR addresses one more UI/API inconsistency. Copy-and-pasting my comment from the issue below: (please let me know if there's any reason we don't want this!)

> One other difference in how the roles are listed in the UI vs the API: the page show all the available roles, the builtin + the defined ones. The `/api/dataverses/*/roles` api only shows the defined kind. There's probably no reason for it not to show all the *available* roles, just like the page, is there? 
> As of now, we have separate apis for the builtin and defined roles. But the former is under `/api/admin`, so it's not available to regular curator users. 

I.e., as implemented in this PR, the `/api/dataverses/*/roles` API will list ALL the roles available for the collection, just like the UI page. 

(will add)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
@@ -303,7 +303,6 @@ public class DataverseRoleServiceBean implements java.io.Serializable {
         Set<DataverseRole> roles = dv.getRoles();
         roles.addAll(findBuiltinRoles());
 
-        //while (!dv.isEffectivelyPermissionRoot()) {
         while (dv.getOwner() != null) {
             dv = dv.getOwner();
             roles.addAll(dv.getRoles());
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
@@ -303,7 +303,8 @@ public class DataverseRoleServiceBean implements java.io.Serializable {
         Set<DataverseRole> roles = dv.getRoles();
         roles.addAll(findBuiltinRoles());
 
-        while (!dv.isEffectivelyPermissionRoot()) {
+        //while (!dv.isEffectivelyPermissionRoot()) {
+        while (dv.getOwner() != null) {
             dv = dv.getOwner();
             roles.addAll(dv.getRoles());
         }
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListRolesCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListRolesCommand.java
@@ -28,7 +28,6 @@ public class ListRolesCommand extends AbstractCommand<Set<DataverseRole>> {
     @Override
     public Set<DataverseRole> execute(CommandContext ctxt) throws CommandException {
         return ctxt.roles().availableRoles(definitionPoint.getId());
-        //return definitionPoint.getRoles();
     }
 
     
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListRolesCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListRolesCommand.java
@@ -27,7 +27,8 @@ public class ListRolesCommand extends AbstractCommand<Set<DataverseRole>> {
 
     @Override
     public Set<DataverseRole> execute(CommandContext ctxt) throws CommandException {
-        return definitionPoint.getRoles();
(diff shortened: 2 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
```diff
@@ -303,7 +303,7 @@ public Set<DataverseRole> availableRoles(Long dvId) {
         Set<DataverseRole> roles = dv.getRoles();
         roles.addAll(findBuiltinRoles());
 
-        while (!dv.isEffectivelyPermissionRoot()) {
+        while (dv.getOwner() != null) {
             dv = dv.getOwner();
```
sekmiller, line 307: So, we're saying here that Dataverses will always inherit roles from their parents. It looks like we never implemented  permission root that would allow the DV owner to refuse to inherit. There are other usages of isEffectivelyPermissionRoot. Should they be removed in favor of getOwner as well?

landreev: I did run this by @scolapasta to make sure I got it right, and it really appears that the line above was simply a bug. I.e., was a wrong place to apply this "permission root" logic. In other words, yes, the roles themselves (the roles available for assignment) should always be inherited. 
We DO want to continue using the permission root concept elsewhere (even though there doesn't appear to be a way currently to set the boolean to false, other than by changing it in the database). It's just that it is meant for restricting the inheritance of role *assignments*, from the parent dataverse to its descendant dataverses. In other words, it should still be possible to set PermissionRoot to false and have a dataverse/collection automatically inherit the role assignments from the parent. 

> It looks like we never implemented permission root that would allow the DV admin to refuse to inherit.

So, it's the other way around, all our dataverses are "permission roots" as of now and therefore refuse to inherit the assignments. 
It also looks like we DID have it implemented at some point, i.e. there used to be a way to set it to false. There's the `UpdatePermissionRootCommand` and this fragment: `RolePermissionFragment.java` appears to be a remnant of a page where it could be changed (but it's not being used as of now). I didn't remember this, but according to Gustavo we had stopped allowing it because "it was causing too much problems" in some cases. But from what I understand, we want to to keep the code that checks for it around in places like `PermissionServiceBean`. 

Does this make sense? I hope I'm getting it right. This is rather old design history. And also, so much time has passed since I talked to Gustavo about it and made this PR...

sekmiller: OK. that makes sense. I recall that we did once allow DVs to set it to false, I just don't remember when that was changed.

sekmiller (approved):

PR #9632, opened 2023-05-31
External Exporters bug fix

qqmyers:
**What this PR does / why we need it**: When there is no weight variable (which is tested in an IT test but not a unit test), the code give an NPE. This PR checks for a null response in getWeghtvariable() to avoid that.

This is a quick follow-up to #9175 which was merged before the IT test failure was noticed.

Closes #

**Suggestions on how to test this**: Let the IT tests run/succeed.

**Additional documentation**:


pdurbin (approved):
I'm approving this but a couple thoughts:

- Are we worried about more NPEs from this part of the code?
- Should we switch to NullSafeJsonBuilder?

PR #9637, opened 2023-06-14
3818 Delete temporary thumbnail files

### LINKED ISSUE #3818: File Upload: Clean up files left in /tmp after successful ingest
Ingest leaves files in /tmp. This is not normally an issue but ideally we clean up after ourselves so that we don't consume space, especially in the case of large uploads of many files. Though /tmp cleaner jobs take care of this in most cases, it would make the task lighter if we clean up when we can.

For normal file ingest, in v4.6.1 we leave files as a result of csv, excel ingest of the form firstpass*.tab 

In v4.6.2 we have added support for Swift storage and in addition to the behavior in v4.6.1 for local files, Swift files leave the following:
-csv and excel ingest leaves files of form firstpass*.tab
-large images leave files of the form tempFileToRescale*.tmp
-other ingest files (.sav, .spss, .por) leave files of form tempIngestSourceFile*.tmp

Also see related ticket on documenting general temp file use and recommendations for sys admins, #2848

haarli:
With S3 storage activated, Dataverse creates temporary thumbnail files (tempFileForRescale*.tmp) for image and PDF ingests. After the thumbnails are uploaded to S3, the temporary files are not deleted. Our tmp directory currently grew up to over 180GB .

This PR deletes the tempFileForRescale*.tmp files silently after they are uploaded to S3.

- partial fix for #3818


pdurbin (approved):
I didn't test this but the code makes sense.

I'm hitting "approve" but this still need to be prioritized.

@haarli I added it near the top of the community column of the backlog. I hope it helps! Thanks for the pull request!

sekmiller:
@haarli, we'd like to merge this PR. Can you please update this branch with the latest from develop and see if the Jenkins build succeeds? Thanks!

haarli:
@sekmiller Sure, the branch is now updated

PR #9641, opened 2023-06-14
IQSS/9640-citationDate bug

### LINKED ISSUE #9640: Citation Date field appears empty when API used to change year in citation of a dataset with no embargoed files

* When does this issue occur?
When using the Dataverse API to change the year in the dataset citation the Dataverse suggests ([see API endpoint documentation](https://guides.dataverse.org/en/5.13/api/native-api.html#set-citation-date-field-type-for-a-dataset)).

The dataset page

* What happens?
We see the Citation Date field, which is defined as "The citation date of a dataset, determined by the longest embargo on any file in version 1.0.". The no date is shown because no files were embargoed.

All users

If we keep the current definition of the field, I think it should appear only when any of the dataset's files are embargoed. When no files are embargoed, the field shouldn't appear.

   Is there's some code that's checking if Dataverse was told to change the date in its suggested citation, and if so, this Citation Date field appears? And to determine which date to add to the Citation Date field, the code only knows to look for the longest embargo on any file, so when there are no embargoed files, it can't add a date. Is that right?

   Should the definition of Citation Date be broadened to be the date used to change the year in Dataverse's suggested citation? 

   How likely is it that a depositor would want to both embargo any files in their dataset and use the API endpoint to change the year in the dataset citation? And if this is likely and something we need to consider, which date should be used, the longest embargo date or the date specified when the API was used? I'm thinking it should be the date specified when the API was used, since the depositor used the endpoint for that reason.  

v5.13

None that I could find.

**Screenshots:**

_When the API is used to change the year of Dataverse's suggested citation for a dataset with no embargoed files, the Citation Date field appears with no date_
Screenshot 2023-05-31 at 5 49 44 PM

qqmyers:
**What this PR does / why we need it**: The field in the issue was only intended to show when the citation uses the default date and there are embargoed files ( and thus the date is when the latest embargo ends and is different than when the publish button was pressed). When the date used in the citation is set specifically via API, this extra field doesn't need to show.

Closes #9640

**Suggestions on how to test this**: see issue

**Additional documentation**:


pdurbin (approved):
I didn't test it but it seems like a reasonable fix.

scolapasta (approved):

PR #9645, opened 2023-06-15
QDR - fix for full-text indexing when one file fails

qqmyers:
**What this PR does / why we need it**: Previously, a full-text indexing failure in one file would cause basic indexing of that file's metadata to be skipped. This PR fixes that and allows indexing of the name/tags, etc. to occur.

Closes #

**Suggestions on how to test this**: Run full-text indexing on a dataset with a file where indexing is known to fail (due to out-of-memory or other errors - any work), verify that the file is searchable by name, tags. etc. afterwards.

**Additional documentation**:


pdurbin (approved):
Looks like a reasonable fix to me.

scolapasta (approved):

kcondon:
@qqmyers This does not work for a zero length file. That was the easiest test case I could come up with. The indexing fails with a tika zero length error in logs but the filename then is not indexed per testing instructions above.

qqmyers:
@kcondon - I think it may be working as designed. The zerolength.txt file is having a logged failure for full text indexing, but it appears that searching for the file name, e.g. https://dataverse-internal.iq.harvard.edu/dataverse/root/?q=zerolength works. That would not have happened prior to the PR - searching on name, tags, etc. would have failed. I did go ahead and merge with dev to see if that fixes the test.

kcondon:
@qqmyers OK, I'll retest. When I first did this I could not search and find the zero length file from the basic search box.

PR #9649, opened 2023-06-16
9647 bugfix facets on files page

### LINKED ISSUE #9647: Filters for files in dataset page are gone
Open a dataset with more than one file, the filters are gone (see "Filter by"):
image

* When does this issue occur?
When opening a dataset.

Dataset page (files page fragment)

* What happens?
Filters for the files are gone.

Everybody.

Filters are there.

Develop.

#9558 

**Screenshots:**
You can cheat the page into showing the filters by providing a search term:
image

The facets will remain even after removing the search-term:
image

I have found the code causing this and the fix for it. The PR will follow shortly.

ErykKul:
It fixes the missing filters for files in dataset page bug.

Closes #9647 

Open a dataset page for a dataset with more than 1 file, it should contain the facet filters (Filter by)

No, this is only a bugfix.

No.

**Additional documentation**:
image

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -684,12 +684,14 @@ public class DatasetPage implements java.io.Serializable {
 
     private List<FileMetadata> selectFileMetadatasForDisplay() {
         final Set<Long> searchResultsIdSet;
-        if (StringUtil.isEmpty(fileLabelSearchTerm) && StringUtil.isEmpty(fileTypeFacet) && StringUtil.isEmpty(fileAccessFacet) && StringUtil.isEmpty(fileTagsFacet)) {
-            // But, if no search terms were specified, we return the full
-            // list of the files in the version:
+        if (isIndexedVersion() && StringUtil.isEmpty(fileLabelSearchTerm) && StringUtil.isEmpty(fileTypeFacet) && StringUtil.isEmpty(fileAccessFacet) && StringUtil.isEmpty(fileTagsFacet)) {
+            // Indexed version: we need facets, they are set as a side effect of getFileIdsInVersionFromSolr method.
+            // But, no search terms were specified, we will return the full
+            // list of the files in the version: we discard the getFileIdsInVersionFromSolr.
+            getFileIdsInVersionFromSolr(workingVersion.getId(), this.fileLabelSearchTerm);
             // Since the search results should include the full set of fmds if all the
             // terms/facets are empty, setting them to null should just be
-            // an optimization for the loop below
+            // an optimization to skip the loop below
             searchResultsIdSet = null;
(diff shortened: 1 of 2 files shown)


ErykKul:
@pdurbin @kcondon 
This pull request fixes a bug introduced in my PR #9558, I have missed a side effect there and the facet were not set when opening a dataset page. The change is very limited: only 2 lines of code, the rest is comments. Sorry for the late notice.

qqmyers (approved):
makes sense. QA should be able to verify the fix works given the problem description.

ErykKul:
@landreev 
I think that attribute `rel="nofollow"` should do the trick:
image
https://developers.google.com/search/docs/crawling-indexing/qualify-outbound-links

If it is not what is desired, let me know and I will revert it.

kcondon:
@ErykKul @qqmyers This works but a minor issue is that the filter doesn't appear right away when a draft is created from a published dataset -page needs to be refreshed. Is this because async not available at time of page init? I can merge as is since it is minor, just wanted to mention it.

landreev:
Thanks. I've been experimenting with "nofollow", among other things. There's some evidence that googlebot doesn't always honor it (and of course, there are other bots that don't honor any rules). But it definitely shouldn't hurt, so please leave it in place, thank you!

ErykKul:
The `rel="nofollow"` attribute does not impact the user experience. I think the engines should not crawl the filters. From what I understand, Yahoo would still crawl the link "nofollow", other engines would not...

landreev:
>  the filter doesn't appear right away when a draft is created from a published dataset 

I haven't verified yet, but I would guess that this has always been the case.

kcondon:
@ErykKul @landreev I just read a bit on nofollow and saw this: Nofollow for internal links. You shouldn’t use nofollow for internal links. If you don’t want a certain page to be crawled or indexed, there are better ways to achieve it, such as [robots meta tags](https://www.semrush.com/blog/robots-meta/).

kcondon:
> > the filter doesn't appear right away when a draft is created from a published dataset
> 
> I haven't verified yet, but I would guess that this has always been the case.

I tested on demo, it works there.

kcondon:
OK, I'm good to merge, fine with trying nofollow as well. Can always change if it doesn't work?

landreev:
Seeing how publishing has always been asynchronous - not just since #9558 - I would be very surprised if there were a consistent change in this behavior since then. So my money would still be on some combination of timing and coincidence, as an explanation of the observed behavior on demo... but I don't have time to test this hypothesis heavily, so not a large sum of money. 😄

landreev:
Trying to read on googlebot rules is a rabbit hole... But I am with Eryk in that I don't think it can hurt.

kcondon:
I'm happy to take your money ;) Yeah, also saw comments that google now treats nofollow as a hint rather than as a rule.

PR #9669, opened 2023-06-21
IQSS/9506 thumbnail failure tracking and other performance improvements

### LINKED ISSUE #9506: Add a "no thumbnail" flag to mark problematic images (to avoid extra generation attempts)
Being able to generate a thumbnail for an image file isn't guaranteed. Occasionally it fails, more frequently for certain image formats. Or for files that were mistakenly assigned image mime types. This results in a condition where the application will try to generate and cache the thumbnail, causing a delay, on every page load. 
We have a boolean database flag indicating that the thumbnails are already generated for a specific file. Let's add a "bad image, don't even try" flag as well. 

(an existing admin workaround for this condition of course is to generate and cache the files with the right `.thumb*` extensions outside of the application; whether the actual thumbnails, when possible, or of some "broken image" icon, if it's not a valid image file. changing the mime type in the latter case is also an option)w

qqmyers:
**What this PR does / why we need it**: This PR makes several changes to improve the performance of thumbnail generation and retrieval, covering the issue #9506 (also reported via QDR) and the issue raised by Bikramjit Singh/Borealis when using relatively slow S3 storage (email between Borealis and @scolapasta, @pdurbin and @qqmyers ) and additional issues discovered while investigating:
 - Caches the isThumbnailAvailable response in the ThumbnailServiceWrapper for the dataset file table in edit/view modes
 - Switches to returning a download URL (versus a base64-encoded copy) in the main dataset search display
 - Sets the dataset id in datasets returned in search results to enable the existing caching in the ThumbnailServiceWrapper.dvobjectThumbnailsMap (the lack of an id meant the caching map wasn't being populated)
 - Implements a ~~previewshavefailed~~ previewimagefail flag that is set the first time an attempt to create a thumbnail for a given file fails which is then used to avoid retrying the thumbnail creation process every time a thumbnail is requested (or isThumbnailAvailable() was called). Adds api calls to reset this flag globally or per file (to allow retrying)
 - Switches to using streams (vs channels) in pdf thumb generation for the temp file case since the channel.transferFrom method is not guaranteed to transfer all bytes (and can transfer 0 bytes) whereas the InputStream.transferTo method blocks until all bytes are transferred.
 - Refactors to remove duplicate code
 - Sets the preview available flag false when the attempt to copy temporary previews during Ingest fails

- Closes #9506 

This PR doesn't completely remove using a base64 encoded thumb URL, e.g. on the dataset and file page where one base64 image is displayed. Once could also remove it in that case, but the performance issues related to base64 generation are primarily when many images have to be created before a page can be rendered.

This may also be useful for the SPA?

**Suggestions on how to test this**: Assure that thumbnails appear for image/pdf files as before (regression), that initial page load is faster/DV server is not making multiple S3 calls from the server to render initial root  collection page, that files where thumbnail generation fails get marked with the previewshavefailed flag and subsequent accesses don't attempt to recreate the thumbnail, that using the api call to reset the flag for a/all files results in a new attempt to create a thumbnail.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: thumbnails on the main page now load ~asynchronously (versus the page not loading at all until all thumbs are available).

**Additional documentation**: admin API docs


bikramj:
Thank you so much @qqmyers for implementing it, this will solve the slow page load issue for us and anyone using custom S3 endpoints.

bikramj:
Hi @qqmyers, 
Will this go in `6.1` release?

qqmyers:
It's been marked for the 6.1 release so assuming that decision isn't changed, it will be.

pdurbin (commented):
The Flyway script needs to be renamed.

src/main/resources/db/migration/V5.13.0.2__9506-track-thumb-failures.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE dvobject ADD COLUMN IF NOT EXISTS previewshavefailed BOOLEAN DEFAULT FALSE;
```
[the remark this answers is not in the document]
sekmiller: I think I mentioned in standup yesterday that I have 0.5 on the harvesting id script so it will be a race.

src/main/java/edu/harvard/iq/dataverse/ThumbnailServiceWrapper.java
```diff
@@ -5,19 +5,27 @@
  */
 package edu.harvard.iq.dataverse;
 
+import edu.harvard.iq.dataverse.api.Datasets;
 import edu.harvard.iq.dataverse.dataaccess.DataAccess;
 import edu.harvard.iq.dataverse.dataaccess.StorageIO;
```
sekmiller, line 10: There are a bunch of unused imports here. Some new; some existing.

src/main/java/edu/harvard/iq/dataverse/ThumbnailServiceWrapper.java
```diff
@@ -46,6 +57,7 @@ public class ThumbnailServiceWrapper implements java.io.Serializable  {
     
     private Map<Long, String> dvobjectThumbnailsMap = new HashMap<>();
     private Map<Long, DvObject> dvobjectViewMap = new HashMap<>();
+    private Map<Long, Boolean> hasThumbMap = new HashMap<>();
 
     private String getAssignedDatasetImage(Dataset dataset, int size) {
```
sekmiller, line 62: Any reason to keep this method around as it doesn't seem to be used anywhere?

qqmyers: flyway updated, unused imports/method removed.

sekmiller (approved):
Looks good. thanks for the updates.

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
```diff
@@ -1380,7 +1380,7 @@ public void setDisplayCardValues() {
                 result.setImageUrl(thumbnailServiceWrapper.getDataverseCardImageAsBase64Url(result));
             } else if (result.getType().equals("datasets")) {
                 if (result.getEntity() != null) {
-                    result.setImageUrl(thumbnailServiceWrapper.getDatasetCardImageAsBase64Url(result));
+                    result.setImageUrl(thumbnailServiceWrapper.getDatasetCardImageAsUrl(result));
```
pdurbin, line 1383: I'm not sure why I'm seeing a broken thumbnail on the search card:

Screenshot 2023-12-04 at 3 49 10 PM

Nothing in server.log. Works fine on demo. Here are the two files I uploaded:

9568-homer
[Architecture ideas - Sheet1.pdf](https://github.com/IQSS/dataverse/files/13551225/Architecture.ideas.-.Sheet1.pdf)

In short, if you set PNG (of Homer) as the dataset thumbnail, it looks weird, as above. A workaround is to remove it so that people see the default dataset icon.

qqmyers: Hmm, just tested at QDR, which has the original version of the code and it works OK as far as I can tell: 
image
. I can look to see if there are code differences and try to set up this branch somewhere. (FWIW: I can't get your exact homer image - github says "This private-user-images.githubusercontent.com page can’t be found" when I try to copy it/it's link.

qqmyers: The pdf works OK too.

pdurbin: @qqmyers I'm sorry but QDR is a fork and not a good test. I just switched to develop and thumbnails work fine. Then I switched back to this branch and I'm getting 404s:

Screenshot 2023-12-05 at 9 06 14 AM

Let's pick which server to test on (I've been testing on Docker on my laptop). Either dataverse-internal or a newly spun up EC2 instance. I'll reach out in Slack.

pdurbin: This seems to be fixed as of 7349ed9f75.

pdurbin:
There are merge conflicts and the SQL script needs to be bumped.

pdurbin:
Found a regression, which Jim fixed (thank you!). I confirmed that I can still generate thumbnails for images and PDFs. For PDFs I had to install ImageMagick. See also this issue:

- https://github.com/GlobalDataverseCommunityConsortium/dataverse-ansible/pull/110

PR #9684, opened 2023-06-28
Faster combined query for retrieving datasets via API

### LINKED ISSUE #9683: Large amount of queries when getting a dataset with API
Retrieve a large dataset (containing thousands of files) through the API

* When does this issue occur?
It is the same for all datasets, but is more problematic for larger datasets.

* Which page(s) does it occur on?
API calls

* What happens?
When we check the query log, we see multiple queries for each file of the dataset.

All users.

One larger query that retrieves all the necessary data at once.

Develop.

- #9557 

A better solution would be to have an API that retrieves only the dataset and a separate API calls for retrieving the file metadata in a paginated way, as proposed by @Kris-LIBIS. However, many existing applications already use the retrieve dataset API call, including all file metadata in one call, therefore, making it more efficient should be beneficial too.

ErykKul:
It reduces the number of queries needed for retrieving a dataset via API. Especially for large dataset, but even for smaller datasets when there is large traffic on the server, this makes the usage of resources more efficient.

Closes #9683 

Observed time for retrieving a dataset with 10000 files went from 1 minute to 35 seconds with this PR. Real life applications would be less spectacular, but still useful.


ErykKul:
I have run the integration tests locally, and they have passed. I have no idea why the tests failed on jenkins. Can someone check it for me? Thanks!

ErykKul:
Tests passed, it looks OK now.

ErykKul:
I have closed it: it is better to go for a real solution, this one is more a workaround. Also, it is not essential and does require merging...

ErykKul:
Reopened and merged.

ErykKul:
@pdurbin 
I did not investigate deeply, but it looks to me that this pull request might be still useful. Little bit unexpected since the linked issue #9763 is closed. If it is still useful, let me know and I will merge the latest develop branch to it, and resolve the conflicts. If I am missing something and it can be closed, feel free to close this without merging.

pdurbin:
@ErykKul I'm not the strongest with databases so when I see `eclipselink.left-join-fetch` my inclination is to let another developer here take a look: @scolapasta @landreev or @sekmiller

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -390,31 +392,48 @@ protected DataverseLinkingDataverse findDataverseLinkingDataverseOrDie(String da
...
-            if (dataset == null) {
-                throw new WrappedResponse(notFound(BundleUtil.getStringFromBundle("find.dataset.error.dataset.not.found.persistentId", Collections.singletonList(persistentId))));
+            GlobalId globalId;
+            try {
+                globalId = PidUtil.parseAsGlobalID(persistentId);
+            } catch (IllegalArgumentException e) {
+                throw new WrappedResponse(
+                    badRequest(BundleUtil.getStringFromBundle("find.dataset.error.dataset.not.found.bad.id", Collections.singletonList(persistentId))));
+            }
+            datasetId = dvObjSvc.findIdByGlobalId(globalId, DvObject.DType.Dataset);
```
qqmyers, line 414: Skipping the call to datasetSvc.findByGlobalId(String) changes the functionality - that call does a lookup both by the PID and the alternatePID. I think we still want that.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -390,31 +392,48 @@ protected DataverseLinkingDataverse findDataverseLinkingDataverseOrDie(String da
...
+            if (datasetId == null) {
+                throw new WrappedResponse(
+                    badRequest(BundleUtil.getStringFromBundle("find.dataset.error.dataset_id_is_null", Collections.singletonList(PERSISTENT_ID_KEY.substring(1)))));
```
qqmyers, line 417: Why change to badRequest from notFound? (cut and paste error?) The find.dataset.error.dataset_id_is_null seems to be about mot sending in a persistentID at all (as on line 405 above) versus a dataset with the DOI not being found.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -127,7 +127,8 @@ public Dataset findDeep(Object pk) {
             .setHint("eclipselink.left-join-fetch", "o.files.dataFileTags")
             .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas")
             .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas.fileCategories")
-            //.setHint("eclipselink.left-join-fetch", "o.files.guestbookResponses")
+            .setHint("eclipselink.left-join-fetch", "o.files.fileMetadatas.varGroups")
```
qqmyers, line 130: Is this part of the optimization here? I would guess this isn't often accessed, so I'm not sure how much of a performance help it would be.

qqmyers (changes requested):
Overall, looks OK. I requested a couple changes and there are merge conflicts to resolve, but the overall change to get the 'deep' version of a dataset for the GET api/datasets/{id} looks fine. Once this is updated, should be ready to go.

pdurbin:
@ErykKul heads up that there are merge conflicts.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
ErykKul: I had seen in in the query log even when it was not used. For large datasets it does make a difference (one query less for each file).

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
ErykKul: fixed.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
ErykKul: I added the alt global id lookup.

ErykKul:
@jp-tosca I just tested with a dataset with 10000 files, like I did for other PRs. Getting that dataset on the current develop branch took 24 seconds. With this PR it took 8 seconds, 3 times faster! I think this PR aged well (it is from 10 months ago). I will try the `{id}/versions/{versionId}/files` call with the new parameters (limit and offset). Is this one used by the SPA?

ErykKul:
The files API got worse: for 10 files from a 10000 files dataset it went from 100 ms to 3 seconds. For getting all 10000 files it went from 22 seconds to 25 seconds. It looks like it adds 3 seconds to the result (I guess it is the time needed for find deep to finish?). I will remove the find deep from that method, and leave it only for getting the dataset.

ErykKul:
It looks great! It would be very useful for some of the frequently used calls, like getting datasets. Probably out of scope for this PR. I am not sure why integration tests fail now, but I have seen on Slack that there might be a problem with Jenkins. Other than that, I think that this PR can go to QA?

jp-tosca:
> @jp-tosca I just tested with a dataset with 10000 files, like I did for other PRs. Getting that dataset on the current develop branch took 24 seconds. With this PR it took 8 seconds, 3 times faster! I think this PR aged well (it is from 10 months ago). I will try the `{id}/versions/{versionId}/files` call with the new parameters (limit and offset). Is this one used by the SPA?

I would ask maybe @GPortas or @ekraffmiller to check what method is used by the SPA 😄

ErykKul:
Jenkins failed, I am trying to figure out how to check why (I remember that there was a way to see the build log).

ErykKul:
I have checked the action log, it looks like all tests passed, unit and integration tests (maybe I am missing something):
image

qqmyers (approved):
This looks OK. For some reason, the last run couldn't launch the AWS machine so the IT tests didn't run, so perhaps a rerun should get triggered before this passes QA.

PR #9688, opened 2023-06-30
5.14 release notes

landreev:
**What this PR does / why we need it**:

This is the release note for 5.14. 
It is mostly complete, as of writing this, with a few exceptions:
- A few (3?) PRs are still in QA; their release notes will need to be added to the main release note. Please note that I have already included the note for #9388. I could have added the others too, but figured we should wait, in case anything in them changes before the PRs are merged. 
- There's a couple of TODO: notes in the note itself. Please make sure they are resolved before the release goes live. (for example, one is for changing the guides preview urls to the permanent location on `guides.dataverse.org`)
- I experimented some with the order of the sub-sections. I put the Installation and Upgrade at the very top, for anyone who's eager to just go ahead and upgrade. The progressively verbose sections are further down. If anyone actively dislikes this arrangement, it should be a simple matter of cut-and-paste to address. 
- Please review carefully at the very least the sections corresponding to your PRs, making sure nothing got broken. 
- I made utterly arbitrary calls as to what got a feature section of its own, and what ended up on the "Changes and fixes [...] not already mentioned above" list. If you feel your feature or fix deserves its own section, please make it so. 
- Similarly, the order in which things are highlighted is random and does not imply the order of importance. But if you feel like something needs to be made more prominent, please rearrange as needed. 

**Which issue(s) this PR closes**:

Closes #9670

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #9702, opened 2023-07-12
Update 5.14-release-notes.md

qqmyers:
Standardizing the language around the new options (all are microprofile, which can be set as JVM options or other via other mp config providers)


pdurbin:
Most of this feedback had been added already. In 87fd4d25c7 I pushed the last bit to the release notes PR:

- #9688

Closing.

PR #9705, opened 2023-07-14
Prevent "DRAFT" from appearing in citation file downloaded from a published dataset

### LINKED ISSUE #8952: Cite Dataset: "DRAFT VERSION" instead of version number
For a published dataset that has an active draft version, go to the published version and click "Cite Dataset" and select a format for the citation export (e.g. BibTex.) The exported file will display "DRAFT VERSION" instead of the version number. It does not matter whether the user is logged in or not

This behavior was seen in V5.9 anf V5.11.1,

Discussion about this bug in google groups: https://groups.google.com/g/dataverse-community/c/kOSGz6DV4ZQ

okaradeniz:

When a citation is downloaded for a dataset from the cite dataset dropdown, the latest version is cited regardless of the version currently displayed or the format selected. This fix allows the citations to be downloaded with the selected version number, displayed in the URL such as "version=1.0"

- Closes #8952


okaradeniz:
Thanks, you're right. This only addresses  #8952
<https://github.com/IQSS/dataverse/issues/8952>, fixing just the citation
download via UI, so it's not testable via API.

pdurbin:
@okaradeniz uh oh. There are merge conflicts. Would you be able to resolve them?

okaradeniz:
@pdurbin thanks! I resolved the conflict and pushed the changes to this branch. Please let me know if anything else to do about this.

pdurbin:
Huh, https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9705/4/consoleFull is showing this:

fatal: [localhost]: FAILED! => {"changed": false, "msg": "Source /opt/dataverse/dataverse/target/dataverse-6.1.war not found"}

I'm not sure why. 🤔

qqmyers:
Looks like one issue may be an old pom - bad merge: https://github.com/okaradeniz/dataverse/blob/2113412552c2cadec30a93dc7e3ef3be2382cd1d/modules/dataverse-parent/pom.xml#L134

pdurbin:
Bah. Thanks,  @qqmyers !

@okaradeniz what do you think? Do you want to make a fresh pull request?

okaradeniz:
ooh ok, I see now. I'll do the merge again

okaradeniz:
> Bah. Thanks, @qqmyers !
> 
> @okaradeniz what do you think? Do you want to make a fresh pull request?

Yeah, I think that'll be easier. Thanks!

src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
```diff
@@ -386,6 +386,16 @@ public void downloadDatasetCitationXML(Dataset dataset) {
         downloadCitationXML(null, dataset, false);
     }
 
+    public void downloadDatasetCitationXML(DatasetVersion version) {
```
qqmyers, line 389: As far as I can tell, it looks like this method is now used on place of the only prior use of downloadDatasetCitationXML - can that method be deleted? Further, since that method was the only one calling downloadCitationXML with a non-null dataset, can we remove that parameter and associated logic?

Same questions for the other RIS/Bibtex formats - can we remove the method/param that isn't used anymore?

qqmyers (changes requested):
Looks good. I'm requesting some cleanup of now unused code (unless I've missed a use somewhere).
Otherwise, I think it looks Ok. There was a comment about creating a new PR but I think this one looks clean at this point, so I don't think that's needed.

src/main/java/edu/harvard/iq/dataverse/FileDownloadServiceBean.java
qqmyers: Looks good - thanks.

qqmyers:
@okaradeniz - Can you merge with the latest develop branch? It looks like this PR builds a v6.1 war file instead of the now expected 6.2 war so the IT tests don't run.

qqmyers (approved):
Looks good. I see the api.MoveIT test failed but I think that is one of the timing sensitive ones and isn't related to this PR.

okaradeniz:
Thanks! I've looked at the MoveIT test but couldn't find how these changes
are failing it. Please let me know if I need to make any changes.

jp-tosca:
The test works fine locally, I just sent another build. Hopefully, this one will be successful. 🤔

jp-tosca:
I was able to check that this was the case on develop and this change fixes it, thanks a lot @okaradeniz

Issue #9672, opened 2023-06-21
v5.14 Performance testing


kcondon:
Perform performance testing on release candidate

PR #9711, opened 2023-07-17
reduce logging and fix typo

pdurbin:

Logs are too chatty after #9558 was merged. See https://iqss.slack.com/archives/C010LA04BCG/p1689620939737689

None

Found during perf testing:

- #9672

The logging was added a looong time ago: a129c9a. I assume it was just for troubleshooting.

Why is it different now that 5.14? If you look at 5.14 at https://github.com/IQSS/dataverse/blob/v5.13/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L1871 you'll see that `if (persistentId != null) {` was the first in a series of `if else` clauses. Presumably that first `if` was executed a lot.

Now instead of `} else if (this.getId() != null) {`...
... we have `if (this.getId() != null) {` with no `else` so it gets executed a lot more.  This change was made in 64743bf as part of this PR:

- #9558

In Docker, without the reduced logging, I see this when I load a dataset:

```
dev_dataverse>  retreived version: id: 1, state: RELEASED|#]
```

I've been testing by loading a published dataset landing page.

No

No

**Additional documentation**:

None.


qqmyers (approved):
Looks good!

PR #9716, opened 2023-07-19
Iqss/8889 limit file pids management

### LINKED ISSUE #8889: Feature Request: Enable file PIDs at Dataverse Collection Level

qqmyers:
**What this PR does / why we need it**: This PR adds restriction to who can enable file PIDs for a given collection and when. Changing the setting via the API now requires a superuser and is only allowed if the global :FilePIDsEnabled setting is set (to true or false, but not undefined/non-existent).

The PR also tightens the admin API calls to register a file, all files in a collection, or all files globally to follow the per-collection setting for file PIDs.

Docs have been updated.

Also one minor change to the batch file PID calls - the sleep now only applies when a file is registered (when the register command is called) and not for files that are skipped/in a draft version, etc.

Closes #8889 

**Suggestions on how to test this**: Regression testing should assure the collection-level file PID functionality still works. The changes here can be tested by calling the change collection attributes API call to change the file pids setting - as a superuser/normal user, when the :FilePIDsEnabled setting is set (true or false)/isn't set. (The former in each pair should succeed, the latter should not.)

**Is there a release notes update needed for this change?**: I think the release notes are generic w.r.t. these changes, - could possibly say the new collection-level file pid functionlity is optional (versus always enabled).

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -248,7 +248,7 @@ this provider.
 - :ref:`:Shoulder <:Shoulder>`
 - :ref:`:IdentifierGenerationStyle <:IdentifierGenerationStyle>` (optional)
 - :ref:`:DataFilePIDFormat <:DataFilePIDFormat>` (optional)
-- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to true)
+- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to false)
 
 .. _pids-handle-configuration:
 
@@ -297,7 +297,7 @@ Here are the configuration options for PermaLinks:
 - :ref:`:Shoulder <:Shoulder>`
 - :ref:`:IdentifierGenerationStyle <:IdentifierGenerationStyle>` (optional)
 - :ref:`:DataFilePIDFormat <:DataFilePIDFormat>` (optional)
-- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to true)
+- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to false)
 
 .. _auth-modes:
 
@@ -2775,14 +2775,20 @@ timestamps.
 :FilePIDsEnabled
 ++++++++++++++++
 
-Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be true. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
+Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be false. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
 
-If you don't want to register file-based PIDs for your installation, set:
+It is possible to override the installation-wide setting for specific collections, but only if it is set to true or false (and not left undefined). For example, registration of PIDs for files can be enabled in a specific collection when it is disabled instance-wide. Or it can be disabled in specific collections where it is enabled by default. See :ref:`collection-attributes-api` for details. 
+
+To enable file-level PIDs for the entire installation::
+
+``curl -X PUT -d 'true' http://localhost:8080/api/admin/settings/:FilePIDsEnabled``
+
+
+If you don't want to register file-based PIDs for your entire installation, but do want to allow them to be enabled for a given collection set:
 
 ``curl -X PUT -d 'false' http://localhost:8080/api/admin/settings/:FilePIDsEnabled``
 
 
-It is possible to override the installation-wide setting for specific collections. For example, registration of PIDs for files can be enabled in a specific collection when it is disabled instance-wide. Or it can be disabled in specific collections where it is enabled by default. See :ref:`collection-attributes-api` for details. 
 
 .. _:IndependentHandleService:
 
--- a/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
@@ -1514,6 +1514,9 @@ public class Admin extends AbstractApiBean {
             User u = getRequestUser(crc);
             DataverseRequest r = createDataverseRequest(u);
             DataFile df = findDataFileOrDie(id);
+            if(!systemConfig.isFilePIDsEnabledForCollection(df.getOwner().getOwner())) {
+                return forbidden("PIDs are not enabled for this file's collection.");
+            }
             if (df.getIdentifier() == null || df.getIdentifier().isEmpty()) {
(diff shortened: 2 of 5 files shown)


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -2775,14 +2775,20 @@ timestamps.
 :FilePIDsEnabled
 ++++++++++++++++
 
-Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be true. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
+Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be false. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
 
-If you don't want to register file-based PIDs for your installation, set:
+It is possible to override the installation-wide setting for specific collections, but only if it is set to true or false (and not left undefined). For example, registration of PIDs for files can be enabled in a specific collection when it is disabled instance-wide. Or it can be disabled in specific collections where it is enabled by default. See :ref:`collection-attributes-api` for details.
```
pdurbin, line 2780: Should we explain why `:FilePIDsEnabled` has to be set to override it with a collection-specific setting?

I'm not even 100% sure I know myself. Is it that we want installations to be thoughtful about this setting? We want them to make an active decision about if they want installation-wide file PIDs or not before they start enabling file PIDs for this or that collection?

pdurbin (commented):
I left a few comments and questions but I'll defer to others to decide when this is ready for QA.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -248,7 +248,7 @@ this provider.
 - :ref:`:Shoulder <:Shoulder>`
 - :ref:`:IdentifierGenerationStyle <:IdentifierGenerationStyle>` (optional)
 - :ref:`:DataFilePIDFormat <:DataFilePIDFormat>` (optional)
-- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to true)
+- :ref:`:FilePIDsEnabled <:FilePIDsEnabled>` (optional, defaults to false)
```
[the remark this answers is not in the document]
qqmyers: ~Yes. We could also use flyway to change things appropriately - might still need a note somewhere but we could keep the behavior from changing.

doc/sphinx-guides/source/installation/config.rst
qqmyers: As ~coded, setting this flag turns on the ability to set filePIDs per collection and it's value is the global default. The main reason to have not set imply a global default of false and no ability to set at the collection level is to assure that no one can turn on file PIDs when the system admin doesn't expect it (and therefore incur unexpected costs). Given that the collection level change now requires a superuser, we may not need to turn off that ability completely (and we could remove code related to that). Alternately, we could split the setting into two: the global default as to whether to use file PIDs, and a second to decide whether collection-level settings are allowed. In this case, we wouldn't have to change the meaning of not having a global default, but we still could (I think we all agree it is better to start with them off).

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1537,11 +1540,18 @@ public Response registerDataFileAll(@Context ContainerRequestContext crc) {
         Integer alreadyRegistered = 0;
         Integer released = 0;
         Integer draft = 0;
+        Integer skipped = 0;
         logger.info("Starting to register: analyzing " + count + " files. " + new Date());
         logger.info("Only unregistered, published files will be registered.");
         for (DataFile df : fileService.findAll()) {
             try {
                 if ((df.getIdentifier() == null || df.getIdentifier().isEmpty())) {
+                    if(!systemConfig.isFilePIDsEnabledForCollection(df.getOwner().getOwner())) {
```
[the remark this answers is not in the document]
qqmyers: Yes, as coded, these should flip (see other comments).

doc/sphinx-guides/source/installation/config.rst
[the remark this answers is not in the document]
scolapasta: I think adding a flyaway script make sense, so that behavior stays the same, and a quick note to explain it.

I also think that installations that have it currently off should be set to null. But open to other ideas.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -297,7 +297,7 @@ Here are the configuration options for PermaLinks:
...
```
scolapasta, line 300: just so I understand this - defaults to false, instead of null?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -2775,14 +2775,20 @@ timestamps.
 :FilePIDsEnabled
 ++++++++++++++++
 
-Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be true. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
+Toggles publishing of file-level PIDs for the entire installation. By default this setting is absent and Dataverse Software assumes it to be false. If enabled, the registration will be performed asynchronously (in the background) during publishing of a dataset.
```
scolapasta, line 2778: again, shouldn't absent mean "null" instead of false? or am I just confused?

doc/sphinx-guides/source/installation/config.rst
scolapasta: Yes, the idea is preemptive for if/when we allow non superusers to turn this on.

scolapasta (commented):
I added a few comments; overall looks good.

One small note to see if we care; right now if global setting is null, you cannot set the value for a collection. But if it were true or false, you could set, then set global value to null. Do we care at all? (the logic should do an "AND"* with the global value, correct?)

* if the global value is true, can a (super)user set it to false for a collection? Or is it actally an "OR" and not an "AND"

doc/sphinx-guides/source/installation/config.rst
qqmyers: Optional, if not set the default is filePIDs are not allowed (globally or per collection)

qqmyers:
Re: setting :FilePIDsEnabled, changing a collection, and then removing :FilePIDsEnabled - I think right now the code leaves the collection setting active, just the API to make further changes is disabled. The logic to change that would be to check that the setting is set/non-null and AND that with collection setting. (versus trying to get the boolean value for the setting and having null become false).

PR #9730, opened 2023-07-24
Minor improvements for the external metadata filtering

### LINKED ISSUE #9727: "External Metadata Filtering" improvements.
Opening the issue per conversation with @cmbz and @siacus. 
This is based on the tech. discussion in the spike issue in the local queue: https://github.com/IQSS/dataverse.harvard.edu/issues/221

Per conversations with the IQSS support team, we are going to implement some changes to how the external metadata checks are performed, to make their life easier and have fewer upset users contact them. Most of these changes will be done on the side of the filter, i.e. outside of the Dataverse source tree. But this issue is for any hooks that may need to be added to the application code.

landreev:

This PR adds a couple of extra bits of information to the json that the application passes to the external metadata filter. This is for better record keeping and managing white lists on the filter side. 

Closes #9727

All this PR does is add 2 extra fields to the json exports for datasets and collections that are generated for the external filter. 
The easiest way to test would be to configure the filtering by setting both `:DataverseMetadataValidatorScript` and `:DatasetMetadataValidatorScript` to `/usr/bin/true`; then publish a collection and a dataset, and inspect the metadata exports left in `/tmp`. Both should have the 2 extra fields at the end: `userIdentifier` and `parentAlias`. 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
@@ -524,7 +524,7 @@ public class DatasetUtil {
             jsonMetadata = json(ds).add("datasetVersion", json(ds.getLatestVersion()))
                     .add("sourceAddress", sourceAddressLabel)
                     .add("userIdentifier", userIdentifier)
-                    .add("colAlias", ds.getOwner().getAlias())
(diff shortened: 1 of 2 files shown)


pdurbin (approved):
I didn't test this but the changes make sense

landreev:
Added the milestone 5.14, per #9727.

PR #9703, opened 2023-07-12
#9700 increase wait time for indexing on failing test

sekmiller:
**What this PR does / why we need it**:
The test for moving linked datasets is failing intermittently. This PR doubles the wait time for the required reindexing to occur.

**Which issue(s) this PR closes**:

Closes #9700 Fix Jenkins tests before 5.14 release

**Special notes for your reviewer**: This is the only test that has caused build failures over the last couple of days, but it's obviously not a comprehensive fix.

**Suggestions on how to test this**: No functional changes to review. continue to monitor the Jenkins build for improvements or at least reduced incidence of failures

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:
Link to build status: https://jenkins.dataverse.org/job/IQSS-dataverse-develop/

PR #9741, opened 2023-07-27
increase "universe" in variablemetadata table from varchar 255 to text

### LINKED ISSUE #9728: Universe field in variablemetadata table
Recently we needed to import datasets with variable file level metadata and we discovered these files have variable file level metadata that has field "universe" for a variable that is much bigger than 255 character. Therefore update of var metadata for these files failed. Presently all the string fields in variablemetadata table have type **text** except the "universe". It would be helpful for our import to change universe field from 255 characters to text type as the rest of the fields in the table.

lubitchv:
**What this PR does / why we need it**: It increases field universe in variablemetadata table from varchar 255 to text.

**Which issue(s) this PR closes**: #9728 

Closes #9728 

**Suggestions on how to test this**: use data curation tool to  insert in "universe" field string bigger than 255 and save to dataverse. Open Data Curation or Data Explorer to see that field was saved.


qqmyers (approved):
Looks good. Should the VariableMetadata class have a @Column(columnDefinition="TEXT") annotation for universe as well? (Presumably that's not necessary but would make the code clearer/match other fields.)

qqmyers:
FWIW: The test failure is for testMoveLinkedDataset – edu.harvard.iq.dataverse.api.MoveIT and is presumably related to test issues we've been having. Can't really be related to this PR.

pdurbin:
Yes, @lubitchv if you merge the latest from develop, I assume you'll pick up #9703 which should fix the MoveIT test.

pdurbin:
Also, I hope you don't mind but I gave this issue a new title.

I'm guessing you don't think it's worth adding a release note snippet but please feel free to add one if you like.

lubitchv:
> Yes, @lubitchv if you merge the latest from develop, I assume you'll pick up #9703 which should fix the MoveIT test.

It is already should be there and if there are still integration errors it did not fix it.

PR #9747, opened 2023-07-31
Fixed incorrect behavior of :BlockedApiPolicy #9742

### LINKED ISSUE #9742: Incorrect behavior of :BlockedApiPolicy when set to localhost-only and when using a load balancer
The upstream in the load balancer is defined to localhost, the setting dataverse-useripaddresssourceheader is not configured and the API block policy is localhost-only. If the setting is not configured, [this condition](https://github.com/IQSS/dataverse/blob/727df9794857f73fe1af35b2e404cb48414eeb5e/src/main/java/edu/harvard/iq/dataverse/engine/command/DataverseRequest.java#L68) is not fulfilled  and the source IP address will be retrieved from [this line](https://github.com/IQSS/dataverse/blob/727df9794857f73fe1af35b2e404cb48414eeb5e/src/main/java/edu/harvard/iq/dataverse/engine/command/DataverseRequest.java#L132C33-L132C33) which always will have localhost as its value, if the upstream in the load balancer is configured localhost.

* When does this issue occur?
When using a load balancer, for example nginx, and the API block policy localhost-only.

API endpoints that should be blocked by localhost-only API policy.

* What happens?
Dataverse will always consider that every API request have localhost as origin and the API policy will fail, allowing anyone to access, for example, the admin and builtin-users API endpoints.

All users.

API endpoints defined in the setting `:BlockedApiEndpoints` should only allow requests with origin from localhost.

5.13

**Possible Fix**
There are two possible solutions to fix this bug:

1. Configure in load balancer the upstream for the private IP address instead of localhost;
2. Configure the setting dataverse-useripaddresssourceheader with the command `./asadmin create-jvm-options '-Ddataverse.useripaddresssourceheader=X-Forwarded-For'` and have the load balancer configured to correctly set the header `X-Forwarded-For`.

**Screenshots:**

*

cesarpferreira:
Documentation from our findings when solving an issue related to incorrect behavior of the policy to block API access.

Closes #9742 

No.

No.

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -34,6 +34,12 @@ It is very important to keep the block in place for the "admin" endpoint, and to
 
 It's also possible to prevent file uploads via API by adjusting the :ref:`:UploadMethods` database setting.
 
+If you are using a load balancer or a reverse proxy, there are some additional considerations. If no additional configurations are made and the upstream is configured to redirect to localhost, the API will be accessible from the outside, as your installation will register as origin the localhost for any requests to the endpoints "admin" and "builtin-users". To prevent this, you have two options:
+
+- If your upstream is configured to redirect to localhost, you will need to set the :ref:`JVM option <useripaddresssourceheader>` to one of the following values ``%client.name% %datetime% %request% %status% %response.length% %header.referer% %header.x-forwarded-for%`` and configure from the load balancer side the chosen header to populate with the client IP address.
+
+- Another solution is to set the upstream to the client IP address. In this case no further configuration is needed.
+
 Forcing HTTPS
 +++++++++++++
 


pdurbin (approved):
https://dataverse-guide--9747.org.readthedocs.build/en/9747/installation/config.html#blocking-api-endpoints looks good to me! Approved.

Here's a screenshot:

Screen Shot 2023-07-31 at 12 00 59 PM

donsizemore (commented):
is this specific to nginx? if so, should we make a note of that?

cesarpferreira:
> is this specific to nginx? if so, should we make a note of that?

@donsizemore that is a good question. We were doing some tests with haproxy in one installation, but it had the host IP address so it was working correctly. I can try to change it to localhost and check if this happens

cesarpferreira:
> > is this specific to nginx? if so, should we make a note of that?
> 
> @donsizemore that is a good question. We were doing some tests with haproxy in one installation, but it had the host IP address so it was working correctly. I can try to change it to localhost and check if this happens

Just confirmed, it also happens with haproxy and this situation is most common when you only have one server behind the reverse proxy service and it is easier to set up the backend as localhost.

pdurbin:
Does that mean we're ok with the pull request as-is? Or should @cesarpferreira add an additional note? (@cesarpferreira please feel free to add more if you like!)

PR #9756, opened 2023-08-04
5.14 Infinite loop in DDI export with tabular tag

qqmyers:
**What this PR does / why we need it**: As discovered immediately after release, there is an infinite loop in the DDI export when a tabular tag exists. The fix is self explanatory.

Closes #

**Suggestions on how to test this**: Upload a tabular file, wait for ingest. Select the file in the dataset file table. Hit the table level Edit/Metadata option and add a tabular tag. Then publish. With 5.14, this will result in a rapidly growing tmp file that will crash the machine when the ddi export starts to be generated. WIth the fix, this should not happen and you should get a ddi export.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
@@ -2086,7 +2086,7 @@ public class DdiExportUtil {
 
                 if (fileJson.containsKey("tabularTags")) {
                     JsonArray tags = fileJson.getJsonArray("tabularTags");
-                    for (int j = 0; j < tags.size(); i++) {
+                    for (int j = 0; j < tags.size(); j++) {
                         xmlw.writeStartElement("notes");
                         writeAttribute(xmlw, "level", LEVEL_FILE);
                         writeAttribute(xmlw, "type", NOTE_TYPE_TAG);


landreev (approved):
AI bot says it's ok!

pdurbin (approved):
Looks good, thanks.

pdurbin:
I was able to test this bug with this:

```
diff --git a/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java b/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
index 866524a260..a2c3a19c39 100644
--- a/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
@@ -2391,6 +2391,13 @@ createDataset = UtilIT.createRandomDatasetViaNativeApi(dataverse1Alias, apiToken
 
         assertTrue("Failed test if Ingest Lock exceeds max duration " + pathToFile, UtilIT.sleepForLock(datasetId.longValue(), "Ingest", authorApiToken, UtilIT.MAXIMUM_INGEST_LOCK_DURATION));
 
+        JsonObjectBuilder json = Json.createObjectBuilder()
+                .add("dataFileTags", Json.createArrayBuilder().add("Genomics"));
+
+        Response addTabularTag = UtilIT.updateFileMetadata(fileId, json.build().toString(), authorApiToken);
+        addTabularTag.prettyPrint();
+        addTabularTag.then().assertThat().statusCode(OK.getStatusCode());
+
         Response restrictFile = UtilIT.restrictFile(fileId, true, authorApiToken);
         restrictFile.prettyPrint();
         restrictFile.then().assertThat().statusCode(OK.getStatusCode());
```

Running `du -sh docker-dev-volumes` showed the volume growing continuously.

I stopped the containers, switched to this PR, and the bug is gone.

Issue #9714, opened 2023-07-19
[Spike - API] [2/2] Extend the Files API to support filters


GPortas:
**Overview of the Feature Request**

The logic to be implemented will extend https://github.com/IQSS/dataverse/issues/9692 so that the API is also capable of filtering files by file type, access type and tags.

**What kind of user is the feature intended for?**
API User

**What inspired the request?**
- https://github.com/IQSS/dataverse-frontend/issues/150

**What existing behavior do you want changed?**
N/A

**Any brand new behavior do you want to add to Dataverse?**
Extend Files API

**Any open or closed issues related to this feature request?**
- https://github.com/IQSS/dataverse-frontend/issues/150

pdurbin:
Closing because this PR was merged:

- #9783

PR #9783, opened 2023-08-16
Dataset files API extension for filters

GPortas:

We need this API extension to support dataset files tab filters in the new frontend, as well as extend the capabilities of the API for other use cases.

In particular, this API extension includes:

- New optional query parameter `accessStatus` to `getVersionFiles` endpoint.

- #9714 

The QueryDSL library has been added to Dataverse, which makes it much easier to build queries dynamically. This library integrates perfectly on top of the existing JPA + Jakarta setup. This library can be used from now on for other different use cases within Dataverse.

Other similar libraries have been studied but have been discarded in favor of QueryDSL, due to its simplicity and size.

The getVersionFiles endpoint integration test has been extended to cover all possible filter values, but manual tests can be done with curl:

Example:

`curl -H "X-Dataverse-Key: <YOUR_API_KEY>" -X GET "http://localhost:8080/api/v1/datasets/:persistentId/versions/:latest/files?persistentId=<DATASED_PID>&accessStatus=Restricted"`

Will retrieve all restricted files in the specified dataset.

`accessStatus` query parameter supports the possible values:

* ``Public``
* ``Restricted``
* ``EmbargoedThenRestricted``
* ``EmbargoedThenPublic``

No

Yes

**Additional documentation**:
- http://querydsl.com/static/querydsl/latest/reference/html/ch02.html#jpa_integration


landreev (commented):
This looks good, based on a very quick look. I will take a closer look and add a more formal review. 
I like the idea of adding a dedicated service bean (DataverseVersionFilesServiceBean). I'm wondering if we want to also add a dedicated command - ListVersionFilesCommand? - for this, mirroring the existing ListVersionsCommand.

scolapasta:
@GPortas  I'll move this milestone 6.1 issue to "ready for Review", but I'm assigning you first so that you can handle the 6.0 merge and address any EE10 issues.

landreev (approved):

landreev:
I moved the PR into QA. 
I hope it can be merged soon-ish. Once that's done, I will need to resolve further merge conflicts in my #9883, but that was expected.

kcondon:
@GPortas I've tested this and it is ready to merge but has merged conflicts. Would you take a look? Thanks.

GPortas:
@kcondon 

I'm analyzing the behavior of JSF following your feedback on https://github.com/IQSS/dataverse/pull/9820#issuecomment-1739889082 about the search text filtering.

I have some considerations on this part of your comment: "It also supports file extension but couldn't get it to support file type as displayed on the page".

I have tested the files tab search in JSF for a published dataset and was unable to filter by the display file type. Maybe I'm missing something, but I don't think it is supported. I have recorded the behavior for a published dataset:

searchfiledisplay

File extension should be supported as it is part of the file name. An unexpected fact is that when testing JSF, I have realized that searching by the file extension does not work:

searchbyextension

As you can see in the gif above, it seems to only work when you include part of the first characters of the file name, I think that is why it doesn't work for the file extension.

PR #9817, opened 2023-08-22
New handlnet library 9.3.1 from maven central (9805)

### LINKED ISSUE #9805: move from local_lib handle-8.1.1.jar to Maven Central version 9.x
We'll need to upgrade to version 9.x: https://repo1.maven.org/maven2/net/handle/handle-client/

landreev:

Self-explanatory. 
Tested basic functionality (creating and publishing a dataset with a handle), appears to be working. 

Closes #9805

Our QA handlenet instance appears to be running and working fine. 
In order to test, the following jvm options are needed: 
```
<jvm-options>-Ddataverse.handlenet.admcredfile=/usr/local/dvn-admin/hdl/admpriv.bin</jvm-options>
<jvm-options>-Ddataverse.pid.handlenet.key.passphrase=</jvm-options>
```
The file path in the first one is where the key file is stored on dataverse-internal. The second option is probably not necessary at all (since it's empty), but we had it in the old config for whatever reason. 

The persistent identifier-related database settings should be as follows: 
```
    ":Protocol": "hdl"
    ":Authority": "20.500.12050"
    ":Shoulder": ""
```
Let me know if you run into any problems with this setup. 

**Additional documentation**:


pdurbin (approved):
The change looks good. Tests are passing. I didn't do any testing of my own. Approved.

kcondon:
This works but there is some weird, noisy logging for a single dataset handle:
Note that informational entries are listed as SEVERE and there is a comment about a deprecated auth method.
[hdl_create_pub_log.txt](https://github.com/IQSS/dataverse/files/12442554/hdl_create_pub_log.txt)

landreev:
Hmm. It really is chatty - and it is logging all that seemingly debug-level stuff as `SEVERE`. I'll need to check their docs and see if it's a thing, that we need to disable explicitly. (?)

landreev:
OK, I just realized that the SEVERE level may be for a reason - it may be warning us that some cert used to talk to the handlenet server is about to expire: 
```
sending HDL-HTTP request (version=2.10; oc=1; rc=0; snId=0 caCrt noAuth expires:Sat Aug 26
```
And @kcondon confirmed that this is not unique to this 9.3.1 version from maven central (going back to the local_lib version 8.* results in these same messages in the log). 
We'll revisit it on Mon.

landreev:
I took a quick look at the handles-client library code. They are simply dumping these messages into stderr: 
```
System.err.println("  sending HDL-TCP request (" + req + ") to " + Util.rfcIpPortRepr(addr, port));
```
It is possible that this is something new to p6 (or even p6 under jdk17), that stderr output automatically gets logged as level SEVERE in server.log. Although we would need to check to confirm how it behaved under p5 if we wanted to know for sure.

kcondon:
The logging was a preexisting issue, shows up in 5.14 too on p5. Not an easy way to fix currently.

PR #9840, opened 2023-08-28
add note about :MDCLogPath, Payara 5 vs 6 #9822

### LINKED ISSUE #9822: Payara 6: Mention other examples of app-specific resource paths that may need to be updated to p6.
Some resources, logos, custom homepage, paths to logs (:MDCLogPath) and potentially others based on examples in guides that include payara5/6 in the example path may no longer be accessible or writeable. We should try to characterize these and add it as a note to the upgrade process. Also a suggestion to preserve the former app server dir and possibly db as backup in case an issue needs correcting?

I've seen so far:
1. A broken root dataverse logo (https://dataverse-internal.iq.harvard.edu/logos/1/konqi_svg.png)
2. An error in the logs writing to the old, payara5 path to :MDCLogPath (/usr/local/payara5/glassfish/domains/domain1/logs)

In the case of MDCLogPath, that showed an example command line pointing to p5. It has been updated to p6 but wondering whether all such examples might be potential candidates for review on upgrading? Likely on the minor side of issues but affects fit and finish and the MDC logging is important if you are relying on that.

pdurbin:

This PR lets people upgrading to Payara 6 know that :MDCLogPath might be pointing at their old Payara 5 directory.

- Closes #9822

I poked around in the docs and couldn't find other cases of this.

Nothing to test, really. It's just a note.

No.

No, this edit is part of the main Payara 6 release note.

**Additional documentation**:

None.

### DIFF
--- a/doc/release-notes/9340-payara5to6.md
+++ b/doc/release-notes/9340-payara5to6.md
@@ -76,6 +76,10 @@
 
    `sudo -u dataverse vi /usr/local/payara6/glassfish/domains/domain1/config/jhove.conf`
 
+1. If you are using Make Data Count (MDC), edit :MDCLogPath
+
+   Your `:MDCLogPath` database setting might be pointing to a Payara 5 directory such as `/usr/local/payara5/glassfish/domains/domain1/logs`. If so, edit this to be Payara 6. You'll probably want to copy your logs over as well.
+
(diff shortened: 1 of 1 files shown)


donsizemore (approved):
Looks good!

PR #9856, opened 2023-08-30
A static 404 page (9770)

### LINKED ISSUE #9770: Error Handling: 404 page thrown as a result of trying to load missing/incorrect Dataverse results in large stack trace but otherwise functions.
This was found as part of payara 6 branch testing: https://github.com/IQSS/dataverse/pull/9685 and was the first issue.
The way to reproduce is to go to a non existent dataverse alias: /dataverse/foobar in the browser. In p5 the large stack trace isn't present so maybe a result of err handling in p6.

Since this doesn't appear to affect the user and it fairly narrow, we're spinning it out from the payara 6 branch to work on separately and not block p6.

There is more discussion around this issue but the basics are above. An example statement stack trace: 
[faces_context_large_stack_trace (1).txt](https://github.com/IQSS/dataverse/files/12314756/faces_context_large_stack_trace.1.txt)

From a slack discussion:
Oliver Bertuch
I think what happens is this: JSF sends a 404. The servlet picks it up and redirects to /404.xhtml. This is within the JSF realm (/), so JSF kicks in and sees "wait, I'm already done with this request", handing out the exception.

Oliver Bertuch

I suppose this is the problematic line: [https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdf[…]5/src/main/java/edu/harvard/iq/dataverse/NavigationWrapper.java](https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdfd26e5/src/main/java/edu/harvard/iq/dataverse/NavigationWrapper.java#L101-L101)
[NavigationWrapper.java](https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdfd26e5/src/main/java/edu/harvard/iq/dataverse/NavigationWrapper.java)
        context.responseComplete();

Oliver Bertuch

It is called in sendError(), which is triggered by the wrappers around permission and navigation when the error response is built from here: [https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdf[…]d26e5/src/main/java/edu/harvard/iq/dataverse/DataversePage.java](https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdfd26e5/src/main/java/edu/harvard/iq/dataverse/DataversePage.java#L330-L330)
[DataversePage.java](https://github.com/poikilotherm/dataverse/blob/0133009d5e705af3d9ea05ac12ac9f59fdfd26e5/src/main/java/edu/harvard/iq/dataverse/DataversePage.java)

landreev:

This static page replaces the way-too-dynamic 404.xhtml page that's been throwing errors under p6 on account of some conflict with prettyfaces rewrite. 
It's a rather confusing affair.  I will add some details to the issue, spelling out just how confusing it is. Which will take more time than it took to put together this replacement page - which by itself should be a good reason to go with it. 

Closes #9770

**Additional documentation**:

### DIFF
--- a/src/main/webapp/404static.xhtml
+++ b/src/main/webapp/404static.xhtml
@@ -72,7 +72,7 @@
             <nav id="navbarFixed" class="navbar navbar-default navbar-fixed-top" role="navigation">
                 <div class="container">
                     <div class="navbar-header">
-                        <img src="harvarddataverselogo.png" class="navbar-brand custom-logo" alt="Harvard Dataverse"/>
+                        <h:outputFormat class="navbar-brand custom-logo" value="Dataverse Project"/>
                     </div>
                     <div class="collapse navbar-collapse" id="topNavBar">
                         <ul class="nav navbar-nav navbar-right">
--- a/src/main/webapp/404static.xhtml
+++ b/src/main/webapp/404static.xhtml
@@ -1,0 +1,110 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" lang="en">
+    <head>
+        <title>Dataverse - Not Found</title>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+        <meta http-equiv="Content-Language" content="en" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <meta name="description" content="Share, archive, and get credit for your data. Find and cite data across all research fields." />
+        
+        <link type="image/png" rel="icon" href="https://cdn.rawgit.com/IQSS/dataverse/src/main/webapp/resources/images/favicondataverse.png.xhtml" />
+
+        <link type="image/png" rel="image_src" href="https://cdn.rawgit.com/IQSS/dataverse/src/main/webapp/resources/images/dataverseproject.png.xhtml" />
+        
+        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css" />
+
+        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap-theme.min.css" />
+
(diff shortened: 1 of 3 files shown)


src/main/webapp/404static.xhtml
```diff
@@ -0,0 +1,110 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" lang="en">
+    <head>
+        <title>Dataverse - Not Found</title>
```
pdurbin, line 5: I'm seeing some hard coded English on this page. Should we be using something like..

```
#{bundle['error.404.page.title']}
```

here? And other bundle entries from the old 404.xhtml?

src/main/webapp/404static.xhtml
```diff
@@ -0,0 +1,110 @@
...
+            <!-- div class="jumbotron" style="margin-top:20px;" -->
+            <div class="alert alert-danger" role="alert" style="margin-top:3em;">
+                <h:outputFormat value="#{bundle['error.404.message']}" escape="false" />
+            </div>
+        </div>
+        <div id="footer">
+            <div class="container">
+                <div class="row">
+                    <div class="col-sm-8 small">
+                        <p>Copyright &#169; 2023, The President &#38; Fellows of Harvard College | <a href="http://best-practices.dataverse.org/harvard-policies/harvard-privacy-policy.html" target="_blank">Privacy Policy</a>
```
pdurbin, line 97: I guess we can't have anything dynamic like this:

```
value="#{settingsWrapper.get(':ApplicationPrivacyPolicyUrl')}"
```

pdurbin (commented):
Just some quick feedback. I don't mean to hold up this PR. I think it can be tested as-is.

landreev:
Here's what it looks like: 
Screen Shot 2023-08-30 at 11 55 11 AM

src/main/webapp/404static.xhtml
landreev: Sure, why not. 
The words "copyright ..." etc. should be ok to stay hard-coded in English though.

pdurbin (approved):
I didn't run this but the code makes sense. Approved.

Issue #9773, opened 2023-08-11
Payara 6: Embargo selection fails from Dataverse page, works from file page.


kcondon:
Found while testing Payara 6 branch: https://github.com/IQSS/dataverse/pull/9685 
Selecting embargo from the dataset page, either from the individual file kebab or the edit files menu, results in a popup that says selected files are already published and cannot be embargoed, along with a stack trace. The popup and stack trace appear to be the same in both cases.
<img width="596" alt="Screen Shot 2023-08-11 at 9 31 55 AM" src="https://github.com/IQSS/dataverse/assets/4983816/289ab663-013d-4ffe-bbb2-a3adb3688c71">
[embargo_fail_datasetpg_editfiles.txt](https://github.com/IQSS/dataverse/files/12322026/embargo_fail_datasetpg_editfiles.txt)
[embargo_fail_datasetpg_file_kebab.txt](https://github.com/IQSS/dataverse/files/12322027/embargo_fail_datasetpg_file_kebab.txt)

Selecting embargo from the files page, edit files button, works but shows the same stack trace in the logs
[embargo_success_filepg_editfile.txt](https://github.com/IQSS/dataverse/files/12322031/embargo_success_filepg_editfile.txt)

So, it seems like the failure is initializing the popup due to null, but for some reason the error handling from the file page allows it to proceed anyway but from the dataset page it seems to go to an incorrect, perhaps default error?

scolapasta:
I thing I noticed is that if you then click the embargo a 2nd time, it does show the "correct" popup.  but this also happens if you click on a file that is already published - that is it shows the popup that would be correct for the "previous" file. So something is getting initiallzed late. (possibly due to the error in the stack trace?)

PR #9857, opened 2023-08-30
Update file-edit-button-fragment.xhtml

scolapasta:
What this PR does / why we need it:

fixes the issue with embargo on Payara6

Which issue(s) this PR closes:

Closes https://github.com/IQSS/dataverse/issues/9773

note that the deleted action was never defined in the .xhtml files that included this fragment nor did any refresh Embargo methods ever exist in the associated java classes (likely that was copied from refreshTags and erroneously left behind)

refreshTags does have some logic that may be desired for embargo, but that is out of scope as it dod not exist in 5.14 (or ever) and we can handle in the SPA

Suggestions on how to test this:

Make sure embargo now works as expected (including not allowing for published files)

Does this PR introduce a user interface change? If mockups are available, please link/include them here:

no

Is there a release notes update needed for this change?:

no

Additional documentation:

### DIFF
--- a/src/main/webapp/file-edit-button-fragment.xhtml
+++ b/src/main/webapp/file-edit-button-fragment.xhtml
@@ -95,8 +95,7 @@
             <p:commandLink      update="@([id$=fileEmbargoPopup])" 
                                 onclick="if (!(#{fileMetadata!=null} || testFilesSelected()))
                                             return false;" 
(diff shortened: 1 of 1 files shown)


qqmyers (approved):
Really looks good this time - changing the embargo part! :-)

PR #9863, opened 2023-08-30
reduce logging for download file guestbook/terms bug

### LINKED ISSUE #9806: Payara 6: When downloading a file with tou/guestbook enabled, works but throws server log error.
When downloading a file with tou/guestbook enabled, when popup appears, throws server log error but it appears to work:
[gb_tou_popup_log_err.txt](https://github.com/IQSS/dataverse/files/12381876/gb_tou_popup_log_err.txt)

pdurbin:

The excessive logging will drive them mad. 😵‍💫 

- Closes #9806

Yes, it's a bummer that we have to add the same asadmin command in two places.

See if the logging goes away.

Confirm that download still works.

Confirm that guestbooks still work.

No.

Yes, added.

**Additional documentation**:

No.

### DIFF
--- a/doc/release-notes/9340-payara5to6.md
+++ b/doc/release-notes/9340-payara5to6.md
@@ -130,3 +130,19 @@
    `sudo -u dataverse /usr/local/payara6/bin/asadmin stop-domain`
 
    `sudo -u dataverse /usr/local/payara6/bin/asadmin start-domain`
+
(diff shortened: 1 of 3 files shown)


pdurbin:
@landreev yes, that's my understanding. Closing. We can always re-open it or steal parts of it for a new PR.

PR #9867, opened 2023-08-31
9806 fix issue with noisy error in logs when downloading a file with tou/guestbook enabled

### LINKED ISSUE #9806: Payara 6: When downloading a file with tou/guestbook enabled, works but throws server log error.

scolapasta:
reduces noise in log 🎉 

Closes #9806 

Please make sure I didn't miss any of the "actions". (I missed two of them in my first commit)

try to download a file with guestbook enabled and confirm no error in log. There were 10 places the code changed, so ideally confirm the different paths - these include tabular vs non tabular, globus, package, explore tools, etc. Each way has a different link. I'm pretty sure the same fix is correct for all, so alternatively, also look at code to confirm I got them all.

no

no

**Additional documentation**:

### DIFF
--- a/src/main/webapp/file-download-button-fragment.xhtml
+++ b/src/main/webapp/file-download-button-fragment.xhtml
@@ -74,7 +74,7 @@
                          styleClass="btn-download"
                          process="@this"
                          disabled="#{(fileMetadata.dataFile.ingestInProgress or lockedFromDownload) ? 'disabled' : ''}" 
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I didn't test this but Gustavo said it works and the change seems reasonable.

PR #9885, opened 2023-09-06
fix command for only building configbaker image

pdurbin:

As discussed at [#containers > ✔ iterating on configbaker @ 💬](https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/.E2.9C.94.20iterating.20on.20configbaker/near/385284844) the current command for building just the configbaker images doesn't work. This PR corrects the docs.

Building configbaker separately from Dataverse is much, much faster.

None

None.

I edited `modules/container-configbaker/scripts/bootstrap.sh` (to increase the timeout) but any configbaker script on config would do.

No.

No

**Additional documentation**:

None.


poikilotherm (approved):
Good catch! Don't compile the nasty Java package every time! :gun: :see_no_evil:

PR #9892, opened 2023-09-07
A quick xhtml-only fix for the "template with custom terms of use" bug

### LINKED ISSUE #9825: Trying to create dataset in collection with certain default dataset templates returns error page

* When does this issue occur?
On a collection that has a default template with certain metadata fields filled, click on Add Data, and New Dataset. For example, the collection at https://dataverse.harvard.edu/dataverse/worldfish has a default dataset template. When I try to create a dataset in that collection, the issue occurs. I've also done more testing on Demo Dataverse with collections and different dataset templates with the same results.

Dataverse collection pages

* What happens?
I'm taken to an error page:

  <kbd>Screenshot 2023-08-23 at 12 48 45 PM</kbd>

Users with permission to create datasets in the collection

We're taken to the page where we can add metadata and upload files.

  Or maybe the error page lets us know what's wrong with the dataset template that the dataset is trying to use so that we can fix the template. Maybe this has something to do with how certain fields in the template are filled or not filled?

v5.14

None that I could find

landreev:

This fixes, or works around the bug by disabling **everything** inside the file-download-popup-fragment when the page is in CREATE mode. 

Closes #9825

Going forward, we may want to consider adding some null checks inside the java code to prevent this; it's not entirely impossible that another developer will add some code that will try to call this method on a dataset that doesn't yet have a persistent id for whatever reason - ?

see the linked issue for the steps to reproduce the bug. 

edit: would also be prudent to confirm that the file download popup is still working when you are actually downloading files from a dataset that has terms of use. I have tested - but worth confirming. 

**Additional documentation**:

### DIFF
--- a/src/main/webapp/file-download-popup-fragment.xhtml
+++ b/src/main/webapp/file-download-popup-fragment.xhtml
@@ -8,6 +8,8 @@
                 xmlns:jsf="http://xmlns.jcp.org/jsf"
                 xmlns:iqbs="http://xmlns.jcp.org/jsf/composite/iqbs">
 
+    <ui:fragment rendered="#{DatasetPage.editMode != 'CREATE'}">
+
         <o:importFunctions type="edu.harvard.iq.dataverse.util.MarkupChecker" />
         <p:focus context="guestbookUIFragment"/>
         <p class="help-block">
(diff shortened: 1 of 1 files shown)


scolapasta (approved):

kcondon:
@landreev It does do something weird but likely not related to this bug or fix: if you download a file from the files page with custom terms enabled, it does show the terms download popup and accepting it downloads the file and the popup disappears but there is a copy of the popup in the file preview iframe, also with an accept button you can click, though it doesn't seem to do anything. Weird. I think this fix is good and the other maybe a separate bug?

PR #9895, opened 2023-09-08
Adding the .gitattributes file so it can make it easier for Windows u…

### LINKED ISSUE #9894: Request for .gitattributes file to handle CRLF vs LF for SH Files
For reference: https://groups.google.com/g/dataverse-community/c/dGduT2T8F7Y

Adding the .gitattributes file so it can make it easier for Windows users to get going on the code by just adding a file to the Git repository. 

Users of Docker

The use of docker and the push to welcome Windows users. 

As it is mentioned in the documentation for [[Windows]](https://guides.dataverse.org/en/latest/developers/windows.html) users can set their core.autocrlf to input or false according to the [GitHub documentation](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf) but there is a way, so each individual doesn't have to do this step and also travels with the repository.

If a file .gitattributes is added to the root of the repository with the following line: 

*.sh text eol=lf

It will ignore the settings on the user for conversion of CRLF/LF line endings on .sh files, here is the [documentation](https://www.git-scm.com/docs/gitattributes) on the .gitattributes file.

Can remove the step of each user to deactivate their auto conversion feautre

N/A

jp-tosca:

Adding the .gitattributes file so it can make it easier for Windows users to get going on the code by just adding a file to the Git repository.

- Closes #9894

https://groups.google.com/g/dataverse-community/c/dGduT2T8F7Y 

Download the project from the original repo
Create a windows installation and set the core.autocrlf to false.
The development of the containers should fail. 
Delete the instalation
Set core.autocrlf=input to input or true
The development of the containers should be successful. 
set the core.autocrlf to false again
download from this fork
the development should be successful 

No

Change on the documentation at https://guides.dataverse.org/en/latest/developers/windows.html this step would not be necessary 

**Additional documentation**:
https://www.git-scm.com/docs/gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# https://www.git-scm.com/docs/gitattributes
+
+# This set mandatory LF line endings for .sh files preventing from windows users to having to change the value of their git config --global core.autocrlf to 'false' or 'input'
+*.sh text eol=lf


pdurbin:
@jp-tosca thanks for the pull request!

While you're in here, would you be able to hack on https://github.com/jp-tosca/9894-gitattributes-add/blob/develop/doc/sphinx-guides/source/developers/windows.rst and add anything you think would be helpful to Windows developers?

We are very Mac-heavy at IQSS so we don't have a great understanding of the Windows world! 😅 

Also, dumb question. To test this I need a Windows box, right? I'm on a Mac.

jp-tosca:
Thanks @pdurbin for sure I will get more familiar with the Dataverse and add as much as I can for Windows development. 

You know what they say right? "The only dumb question is the one we don't ask". Yes, a Windows installation would be necessary for testing, I am not sure what is the current situation with Windows on Mac since I think they can't boot Windows since the new chip came out (I am probably wrong) but in the past, I used to use something called parallels to run windows on Mac. 

I also can stop by the center one of these days and bring my machine 😅

jp-tosca:
@pdurbin Was checking this fork

pdurbin:
Closed in favor of this one:

- #9905

Issue #9717, opened 2023-07-19
Postgres 15: permissions errors on public schema


donsizemore:
Public schema permissions are much stricter in Postgres 15; we'll need to modify the installer and/or guides to accommodate this change.

I bumped FlyWay to 9.20.1 to explicitly support Postgres 15 for good measure, but:

```
Caused by: org.flywaydb.core.internal.sqlscript.FlywaySqlScriptException: Migration  failed
-----------------
SQL State  : 42501
Error Code : 0
Message    : ERROR: permission denied for schema public
  Position: 14
Location   :  ()
Line       : 1
Statement  : CREATE TABLE "public"."flyway_schema_history" (
    "installed_rank" INT NOT NULL,
    "version" VARCHAR(50),
    "description" VARCHAR(200) NOT NULL,
    "type" VARCHAR(20) NOT NULL,
    "script" VARCHAR(1000) NOT NULL,
    "checksum" INTEGER,
    "installed_by" VARCHAR(100) NOT NULL,
    "installed_on" TIMESTAMP NOT NULL DEFAULT now(),
    "execution_time" INTEGER NOT NULL,
    "success" BOOLEAN NOT NULL
)
Caused by: org.postgresql.util.PSQLException: ERROR: permission denied for schema public
  Position: 14

Caused by: org.flywaydb.core.internal.sqlscript.FlywaySqlScriptException: Migration  failed
-----------------
SQL State  : 42501
Error Code : 0
Message    : ERROR: permission denied for schema public
  Position: 14
Location   :  ()
Line       : 1
Statement  : CREATE TABLE "public"."flyway_schema_history" (
    "installed_rank" INT NOT NULL,
    "version" VARCHAR(50),
    "description" VARCHAR(200) NOT NULL,
    "type" VARCHAR(20) NOT NULL,
    "script" VARCHAR(1000) NOT NULL,
    "checksum" INTEGER,
    "installed_by" VARCHAR(100) NOT NULL,
    "installed_on" TIMESTAMP NOT NULL DEFAULT now(),
    "execution_time" INTEGER NOT NULL,
    "success" BOOLEAN NOT NULL
)
Caused by: org.postgresql.util.PSQLException: ERROR: permission denied for schema public
  Position: 14
```

To resolve, we probably want the installer (and Dataverse-Ansible) to:
`GRANT ALL ON SCHEMA public TO dvndb` unless we just want to grant `CREATE`?

PR #9901, opened 2023-09-12
Revert "#9717 grant CREATE instead of ALL per pdurbin"

pdurbin:
This reverts commit f71274e7c7a4d47ab7fb973320bcfdb7e6822fbd.

Just a test for now...


donsizemore (approved):
according to my local tests, this fixes the installer on Postgres 13 and 15.

PR #9909, opened 2023-09-13
Allow merging of accounts that are members of the same group

### LINKED ISSUE #9284: Error when merging two accounts with same group
Hello,

When we try to merge two accounts belonging to the same group, an error is reported: 
(  ERROR: duplicate key value violates unique constraint \"explicitgroup_authenticateduser_pkey\"\\n  Detail: Key (explicitgroup_id, containedauthenticatedusers_id)=(16, 72) already exists.)

Is that the expected behavior or it is a Dataverse bug please?

Sirine

stevenferey:

- Closes #9284 

**Additional documentation**:


src/test/java/edu/harvard/iq/dataverse/api/UsersIT.java
```diff
@@ -206,15 +207,15 @@ public void testMergeAccounts(){
...
         String displayName = "Group for " + dataverseAlias;
         String user2identifier = "@" + usernameConsumed;
+        String target2identifier = "@" + targetname;
         Response createGroup = UtilIT.createGroup(dataverseAlias, aliasInOwner, displayName, superuserApiToken);
         createGroup.prettyPrint();
         createGroup.then().assertThat()
                 .statusCode(CREATED.getStatusCode());
 
-        String groupIdentifier = JsonPath.from(createGroup.asString()).getString("data.identifier");
+        // String groupIdentifier = JsonPath.from(createGroup.asString()).getString("data.identifier");
```
bencomp, line 216: I suggest to remove code that is no longer needed, instead of leaving it commented out. The codebase has lots of commented code already, which is very rarely removed.

src/test/java/edu/harvard/iq/dataverse/api/UsersIT.java
```diff
@@ -206,15 +207,13 @@ public void testMergeAccounts(){
...
         Response createGroup = UtilIT.createGroup(dataverseAlias, aliasInOwner, displayName, superuserApiToken);
         createGroup.prettyPrint();
         createGroup.then().assertThat()
                 .statusCode(CREATED.getStatusCode());
 
-        String groupIdentifier = JsonPath.from(createGroup.asString()).getString("data.identifier");
-
-        List<String> roleAssigneesToAdd = new ArrayList<>();
-        roleAssigneesToAdd.add(user2identifier);
+        List<String> roleAssigneesToAdd = Arrays.asList(user2identifier, target2identifier);
```
sekmiller, line 216: @stevenferey does it make sense to leave in the original test (that is, create another group and only add the "consumed" user) - just so we don't lose the original test?

jeromeroucou: The original test have been remove because if the test is valid for merged accounts in the same group, I think it's difficult to not to be valid with accounts not in the same group. However, I don't think the reverse is true.

But I can keep the original test and add a specific test for accounts in the same group.

sekmiller: No, it's OK - thanks. Just wanted to make sure it was considered.

sekmiller (approved):

PR #9937, opened 2023-09-20
Extend Datasets API to support dataset deaccession and recovery of files and file counts of deaccessioned datasets

### LINKED ISSUE #9852: [Spike - API] Extend the Files API to return files for deaccessioned datasets
Currently the getVersionFiles endpoint does not allow getting the files for a deaccessioned dataset version, returning a not found error instead. 

Since the SPA needs to display deaccessioned dataset versions (As JSF), including its files, we need to extend the API to return files for deaccessioned dataset version.

API User

- https://github.com/IQSS/dataverse-frontend/issues/134

Support deaccessioned dataset versions in the Files API.

Return files for deaccessioned dataset versions via API.

- https://github.com/IQSS/dataverse-frontend/issues/134

GPortas:

This PR extends the following endpoints related to dataset version files:

- getVersionFiles ("{id}/versions/{versionId}/files")
- getVersionFileCounts ("{id}/versions/{versionId}/files/counts")

The above endpoints now accept a new boolean optional query parameter "includeDeaccessioned", which, if enabled, causes the endpoint to consider deaccessioned versions when searching for versions to obtain files or file counts. To do this, the following existing commands were modified:

- GetLatestAccessibleDatasetVersionCommand
- GetSpecificPublishedDatasetVersionCommand
- GetLatestPublishedDatasetVersionCommand

This logic is required for the SPA to display file metadata for deaccessioned dataset versions.

Additionally, a new endpoint has been developed to support version deaccessioning through API (Given a dataset and a version): "{id}/versions/{versionId}/deaccession".

The version identifiers ":latest", ":latest-published" and ":draft" have been extracted to API constants (ApiConstants.java), since these strings were hardcoded throughout the application and this way it is easier to manage them.

Finally, the DataFile API payload has been extended to add the field "friendlyType", as described in https://github.com/IQSS/dataverse-frontend/pull/169#issuecomment-1727018970

- Closes #9852

1) Create a dataset version and publish it

2) Deaccession dataset version via curl:

Create a deaccession.json file, which should like this:

```
{
  "deaccessionReason": "Description of the deaccession reason.",
  "deaccessionForwardURL": "https://demo.dataverse.org"
}

```
Export the following variables:

```
export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
export SERVER_URL=http://localhost:8080 (or your instance URL)
export ID=<DATASET_ID>
export VERSIONID=1.0
export FILE_PATH=deaccession.json
```

Then, call the deaccession endpoint:

`  curl -H "X-Dataverse-key:$API_TOKEN" -X POST "$SERVER_URL/api/datasets/$ID/versions/$VERSIONID/deaccession" -H "Content-type:application/json" --upload-file $FILE_PATH
`

3) Get version files for the deaccessioned dataset version:

`curl "$SERVER_URL/api/datasets/$ID/versions/$VERSIONID/files?includeDeaccessioned=true"`

4) Get version file counts for the deaccessioned dataset version:

`curl "$SERVER_URL/api/datasets/$ID/versions/$VERSIONID/files/counts?includeDeaccessioned=true"`

5) Repeat previous operations but removing includeDeaccessioned query param to check how the deaccessioned version is not found (old behavior).

No

Yes


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1344,6 +1364,26 @@ The fully expanded example above (without environment variables) looks like this
...
+  export VERSIONID=1.0
+
+  curl -H "X-Dataverse-key:$API_TOKEN" -X PUT "$SERVER_URL/api/datasets/$ID/versions/$VERSIONID/deaccession"
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X PUT "https://demo.dataverse.org/api/datasets/24/versions/1.0/deaccession"
+
```
sekmiller, line 1386: I think we need to describe the json file that goes along with this for providing values for deaccessionReason and deaccessionForwardURL

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -3917,4 +3923,32 @@ public Response getDatasetVersionCitation(@Context ContainerRequestContext crc,
         return response(req -> ok(
                 getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers).getCitation(true, false)), getRequestUser(crc));
     }
+
+    @PUT
+    @AuthRequired
+    @Path("{id}/versions/{versionId}/deaccession")
+    public Response deaccessionDataset(@Context ContainerRequestContext crc, @PathParam("id") String datasetId, @PathParam("versionId") String versionId, String jsonBody, @Context UriInfo uriInfo, @Context HttpHeaders headers) {
+        if (DS_VERSION_DRAFT.equals(versionId) || DS_VERSION_LATEST.equals(versionId)) {
```
sekmiller, line 3931: If the "latest" is published would this be ok?

doc/sphinx-guides/source/api/native-api.rst
GPortas: You are right, I missed that part! Thank you.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
GPortas: Since ":latest" considers daft version (If you have permissions), I opted for restricting ":latest" and considered only ":latest-published". 

Conceptually I think it makes sense, and makes the underlaying logic simpler since otherwise we would have to add additional logic in the dataset version recovery commands (for latest) when we search for a dataset for deaccessioning.

If you think it is relevant enough, I can update it and add such logic.

sekmiller: Just to make it easier on the user you could allow them to use "latest", do the getDraftVersionOrDie first and check to see if that version is draft. Also, do we want to check to see if the version is already deaccessioned? Maybe it's ok and we want to allow them to update the reason/url.

sekmiller: With respect to re-deaccessioning, it's not allowed in the UI.

pdurbin:
Since this PR isn't against develop here's a compare link: https://github.com/IQSS/dataverse/compare/9851-datafile-payload-extension...9852-files-api-deaccessioned-datasets

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1020,7 +1020,17 @@ Usage example:
 
 Please note that both filtering and ordering criteria values are case sensitive and must be correctly typed for the endpoint to recognize them.
 
-Keep in mind that you can combine all of the above query params depending on the results you are looking for.
+By default, deaccessioned dataset versions are not supported by this endpoint and will be ignored in the search when applying the :latest or :latest-published identifiers. Additionally, when filtering by a specific version tag, you will get a not found error if the version is deaccessioned and you do not enable the option described below.
```
pdurbin, line 1023:
```suggestion
By default, deaccessioned dataset versions are not included in the search when applying the :latest or :latest-published identifiers. Additionally, when filtering by a specific version tag, you will get a "not found" error if the version is deaccessioned and you do not enable the ``includeDeaccessioned`` option described below.
```

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1020,7 +1020,17 @@ Usage example:
...
+
+If you want to consider deaccessioned dataset versions, you must specify this through the ``includeDeaccessioned`` query parameter.
```
pdurbin, line 1025:
```suggestion
If you want to consider deaccessioned dataset versions, you must set ``includeDeaccessioned`` query parameter to ``true``.
```

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1048,6 +1058,16 @@ The fully expanded example above (without environment variables) looks like this
 
   curl "https://demo.dataverse.org/api/datasets/24/versions/1.0/files/counts"
 
+By default, deaccessioned dataset versions are not supported by this endpoint and will be ignored in the search when applying the :latest or :latest-published identifiers. Additionally, when filtering by a specific version tag, you will get a not found error if the version is deaccessioned and you do not enable the option described below.
```
pdurbin, line 1061: Same suggestion as above: not included instead of not supported.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1344,6 +1364,27 @@ The fully expanded example above (without environment variables) looks like this
...
+
+Given a version of a dataset, updates its status to deaccessioned.
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+  export ID=24
+  export VERSIONID=1.0
+  export JSON='{"deaccessionReason":"Description of the deaccession reason.", "deaccessionForwardURL":"https://demo.dataverse.org"}'
```
pdurbin, line 1378: What are the possible deaccession reasons? Is there a enum of reasons (FOO, BAR, BAZ) or are they English strings? Can you send any English string? What about Spanish or whatever?

src/main/java/edu/harvard/iq/dataverse/api/ApiConstants.java
```diff
@@ -12,4 +12,9 @@ private ApiConstants() {
 
     // Authentication
     public static final String CONTAINER_REQUEST_CONTEXT_USER = "user";
+
+    // Dataset
+    public static final String DS_VERSION_LATEST = ":latest";
+    public static final String DS_VERSION_DRAFT = ":draft";
+    public static final String DS_VERSION_LATEST_PUBLISHED = ":latest-published";
```
pdurbin, line 19: Thank you for adding all these constants! 🚀

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: I'm fine with only supporting ":latest-published". Only published datasets can be deaccessioned, after all. If there's a strong need, we could always add ":latest" later.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1344,6 +1364,27 @@ The fully expanded example above (without environment variables) looks like this
 
   curl -H "X-Dataverse-key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X DELETE "https://demo.dataverse.org/api/datasets/24/versions/:draft"
 
+Deaccession Dataset
+~~~~~~~~~~~~~~~~~~~
+
+Given a version of a dataset, updates its status to deaccessioned.
```
pdurbin, line 1370: As far as I understand, the code allows you to re-deaccession a dataset that has already been deaccessioned, allowing you to update the `deaccessionReason`, for example. Can we mention this here in the guides?

pdurbin (commented):
I didn't run the code but I left some comments, questions, and suggestions. Overall, I'm very excited about this PR! As I mentioned, we haven't been able to deaccession datasets via API since DVN 3! https://guides.dataverse.org/en/3.6.2/dataverse-api-main.html#deaccession-a-study-released-studies-only

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
GPortas: The drawback I see with :latest is that the dataset it refers to depends on the user's permissions. If the user can access the draft version, :latest will correspond to draft (Current behavior), otherwise it will search for the latest published version if there is one.

The problem is that you can't deaccession a draft version. Therefore it can be a bit confusing for the user. An alternative I thought of is to never point to the draft version for this particular endpoint when using :latest, but this can also create confusion for the user as the other endpoints would continue to support draft for :latest.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -3917,4 +3923,32 @@ public Response getDatasetVersionCitation(@Context ContainerRequestContext crc,
...
+            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers, false);
+            try (StringReader stringReader = new StringReader(jsonBody)) {
+                JsonObject jsonObject = Json.createReader(stringReader).readObject();
+                datasetVersion.setVersionNote(jsonObject.getString("deaccessionReason"));
+                String deaccessionForwardURL = jsonObject.getString("deaccessionForwardURL", null);
+                if (deaccessionForwardURL != null) {
+                    try {
+                        datasetVersion.setArchiveNote(deaccessionForwardURL);
+                    } catch (IllegalArgumentException iae) {
+                        return error(Response.Status.BAD_REQUEST, "Invalid deaccession forward URL: " + iae.getMessage());
```
pdurbin, line 3944: Is it possible that the SPA would want an error like this (unable to set URL due to length, for example) in a language other than English? Otherwise, all the SPA knows is 400 BAD REQUEST and none of the details.

pdurbin (commented):
One more thought about English vs. other languages for error handling.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
sekmiller: My suggestion was to allow the use of latest and if that happens to be draft to give an error response. We can leave it as only allow latest-published.

doc/sphinx-guides/source/api/native-api.rst
GPortas: Resolving since we discussed about it on Friday. We allow any string.

doc/sphinx-guides/source/api/native-api.rst
GPortas: Just added a note about this to the docs. As we discussed on Friday, re-deaccessioning is not supported.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1344,6 +1364,29 @@ The fully expanded example above (without environment variables) looks like this
...
+
+  curl -H "X-Dataverse-key:$API_TOKEN" -X PUT "$SERVER_URL/api/datasets/$ID/versions/$VERSIONID/deaccession" -d "$JSON"
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X PUT "https://demo.dataverse.org/api/datasets/24/versions/1.0/deaccession" -d '{"deaccessionReason":"Description of the deaccession reason.", "deaccessionForwardURL":"https://demo.dataverse.org"}'
+
+.. note:: You cannot deaccession a dataset more than once. If you call this endpoint twice for the same dataset version, you will get a not found error on the second call, since the dataset you are looking for will no longer be public since it is already deaccessioned.
```
pdurbin, line 1388:
```suggestion
.. note:: You cannot deaccession a dataset more than once. If you call this endpoint twice for the same dataset version, you will get a not found error on the second call, since the dataset you are looking for will no longer be published since it is already deaccessioned.
```

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: We discussed this. Yes, someday the SPA might want more information but we can revisit this when we implement a deaccessioning API in the context of the SPA actually using it.

pdurbin (commented):
Just a tiny doc suggestion.

pdurbin:
Looking good to me. I just made a tiny doc fix. I'm unassigning myself so we don't have too many cooks!

GPortas:
I just updated the PR with the changes we discussed on Friday:

- Using POST instead of PUT since the deaccessionDataset endpoint, even though it is an update, is not idempotent.
- Using string bundle for error handling
- Integration tests extended with more cases

@sekmiller @pdurbin 

Let me know if you think something is missing. Thanks!

sekmiller (approved):
Thanks for the updates! passing to QA

GPortas:
We have conflicts here due to the revert. Let's wait for #9972  to be merged before resolving the conflicts and merging this PR.

@kcondon

GPortas:
@kcondon Conflicts fixed.

kcondon:
@GPortas In how to test you list a deaccession api but it is incomplete, doesn't include data. So, I used the example in your doc and it failed:
curl -H "X-Dataverse-key:xyz" -X POST "http://localhost:8080/api/datasets/336086/versions/1.0/deaccession" -d '{"deaccessionReason":"Description of the deaccession reason.", "deaccessionForwardURL":"https://demo.dataverse.org"}'
{"status":"ERROR","message":"Error parsing Json: Invalid token=EOF at (line no=1, column no=0, offset=-1). Expected tokens are: [CURLYOPEN, SQUAREOPEN, STRING, NUMBER, TRUE, FALSE, NULL]"

GPortas:
@kcondon 

I apologize. I again forgot to update the description with the latest status of changes (like in other PR). Anyway, I had to update the documentation because the example curl calls were not correct, as you explained. I've updated both the docs and the example calls in the PR description so you can test it now.

Regarding the getVersionFiles and getVersionFileCounts calls with the includeDeaccessioned option and without sending credentials (guest user), they were indeed not working correctly.

The EditDataset permission check was missing for the GetSpecificPublishedDatasetVersionCommand command. I've fixed this and covered that case with integration tests.

kcondon:
@GPortas Darn, just noticed merge conflicts. Can you take a look? Otherwise good to merge. Thanks!

GPortas:
@kcondon Thanks!

I run the failing tests on my localhost and they don't fail. Can you give me more information about those failing tests? Maybe some trace from the Jenkins output? I currently don't have access to Jenkins and can't see it.

Now working on resolving conflicts.

PR #9941, opened 2023-09-20
#9940 - fixed various issues with generated urls of authors for signposting

### LINKED ISSUE #9940: Signposting Authors links issues
Hello there, 

By working on [Signposting parameters (dataverse.signposting.level1-author-limit and dataverse.signposting.level1-item-limit)](https://guides.dataverse.org/en/latest/installation/config.html?highlight=signposting#dataverse-signposting-level1-author-limit) and crawling the Java code, we found 2 issues :

- When author type and identifier fields are correctly filled, level 1 does not display a correct link 
    Ex: Given `type=ORCID` and `idvalue=0000-0002-3202-0263` it prints `Link: <0000-0002-3202-0263>;rel="author"` instead of `Link: <https://orcid.org/0000-0002-3202-0263>;rel="author"`
- When you have more authors than the limit, level 2 does not display links for authors and it should display them all unlike level 1

Also there is comment [//If no type and there's a value, assume it is a URL (is this reasonable?)](https://github.com/IQSS/dataverse/blob/d674d6687f8ade869b285c4853d8722b6112a240/src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java#L215) : I would say that it's reasonable if it's a valid URL. As [FAIR Signposting Profile](https://signposting.org/FAIR/) suggest it must be a link and looks at the end at least like "http(s)://(domain)/(author-id)" not only "(author-id)".

PR is already done.

luddaniel:

- Level 2 is no longer limited in authors number
- Level 1 show better authors links, using type and id or using id as url if valid

- Closes #9940

I tested a lot of cases, I suggest the same approach :  
- check both level 1 and level 2 in every case scenario
- add different kind of data on authors (empty identifier, good and bad url/id as identifier, with and without type)
- cross the limit of 5 good possible links

I would say yes

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
@@ -19,6 +19,8 @@ import edu.harvard.iq.dataverse.dataset.DatasetUtil;
 import jakarta.json.Json;
 import jakarta.json.JsonArrayBuilder;
 import jakarta.json.JsonObjectBuilder;
+import org.apache.commons.validator.routines.UrlValidator;
+
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/util/SignpostingResources.java
```diff
@@ -164,12 +166,11 @@ private List<String> getAuthorURLs(boolean limit) {
...
             logger.fine(String.format("idtype: %s; idvalue: %s, affiliation: %s; identifierUrl: %s", da.getIdType(),
                     da.getIdValue(), da.getAffiliation(), da.getIdentifierAsUrl()));
-            String authorURL = "";
-            authorURL = getAuthorUrl(da);
+            String authorURL = getAuthorUrl(da);
             if (authorURL != null && !authorURL.isBlank()) {
                 // return empty if number of visible author more than max allowed
                 // >= since we're comparing before incrementing visibleAuthorCounter
-                if (visibleAuthorCounter >= maxAuthors) {
+                if (limit && visibleAuthorCounter >= maxAuthors) {
```
qqmyers, line 173: Nice catch - we had a limit param and didn't use it!

qqmyers (approved):
Looks good - thanks for the fixes.

cmbz:
**2923/12/13**
Added to 6.2 Proposals due to linkage with #9940

cmbz:
2024/01/08: Moved into sprint read during prioritization meeting.

PR #9960, opened 2023-09-26
New API option to get the download size of the dataset version files without the original tabular files size

GPortas:
…oadSize datasets API endpoint

Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -2947,25 +2947,32 @@ public class Datasets extends AbstractApiBean {
         String nullCurrentMonth = null;
         return getMakeDataCountMetric(idSupplied, metricSupplied, nullCurrentMonth, country);
     }
-    
+
     @GET
     @AuthRequired
     @Path("{identifier}/storagesize")
-    public Response getStorageSize(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf, @QueryParam("includeCached") boolean includeCached,
-                                   @Context UriInfo uriInfo, @Context HttpHeaders headers) throws WrappedResponse {
-
+    public Response getStorageSize(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf, @QueryParam("includeCached") boolean includeCached) {
         return response(req -> ok(MessageFormat.format(BundleUtil.getStringFromBundle("datasets.api.datasize.storage"),
-                execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), includeCached, GetDatasetStorageSizeCommand.Mode.STORAGE, null)))), getRequestUser(crc));
+                execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), includeCached, true, GetDatasetStorageSizeCommand.Mode.STORAGE, null)))), getRequestUser(crc));
     }
-    
+
     @GET
     @AuthRequired
     @Path("{identifier}/versions/{versionId}/downloadsize")
-    public Response getDownloadSize(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf, @PathParam("versionId") String version,
-                                    @Context UriInfo uriInfo, @Context HttpHeaders headers) throws WrappedResponse {
-
-        return response(req -> ok(MessageFormat.format(BundleUtil.getStringFromBundle("datasets.api.datasize.download"),
-                execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), false, GetDatasetStorageSizeCommand.Mode.DOWNLOAD, getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers))))), getRequestUser(crc));
+    public Response getDownloadSize(@Context ContainerRequestContext crc,
+                                    @PathParam("identifier") String dvIdtf,
+                                    @PathParam("versionId") String version,
+                                    @QueryParam("ignoreOriginalTabularSize") boolean ignoreOriginalTabularSize,
+                                    @Context UriInfo uriInfo,
+                                    @Context HttpHeaders headers) {
+        return response(req -> {
+            Long datasetStorageSize = execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), false, !ignoreOriginalTabularSize, GetDatasetStorageSizeCommand.Mode.DOWNLOAD, getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers)));
+            String message = MessageFormat.format(BundleUtil.getStringFromBundle("datasets.api.datasize.download"), datasetStorageSize);
+            JsonObjectBuilder jsonObjectBuilder = Json.createObjectBuilder();
+            jsonObjectBuilder.add("message", message);
+            jsonObjectBuilder.add("storageSize", datasetStorageSize);
+            return ok(jsonObjectBuilder);
+        }, getRequestUser(crc));
     }
 
     @GET
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetStorageSizeCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetStorageSizeCommand.java
@@ -7,7 +7,6 @@ package edu.harvard.iq.dataverse.engine.command.impl;
 
 import edu.harvard.iq.dataverse.Dataset;
 import edu.harvard.iq.dataverse.DatasetVersion;
-import edu.harvard.iq.dataverse.Dataverse;
 import edu.harvard.iq.dataverse.authorization.Permission;
 import edu.harvard.iq.dataverse.engine.command.AbstractCommand;
 import edu.harvard.iq.dataverse.engine.command.CommandContext;
@@ -15,6 +14,7 @@ import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
 import edu.harvard.iq.dataverse.engine.command.RequiredPermissions;
 import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
 import edu.harvard.iq.dataverse.util.BundleUtil;
+
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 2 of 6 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetStorageSizeCommand.java
```diff
@@ -58,21 +58,20 @@ public GetDatasetStorageSizeCommand(DataverseRequest aRequest, Dataset target, b
...
     @Override
     public Long execute(CommandContext ctxt) throws CommandException {
-        logger.fine("getDataverseStorageSize called on " + dataset.getDisplayName());
-
         if (dataset == null) {
             // should never happen - must indicate some data corruption in the database
             throw new CommandException(BundleUtil.getStringFromBundle("datasets.api.listing.error"), this);
         }
 
+        logger.fine("getDataverseStorageSize called on " + dataset.getDisplayName());
```
GPortas, line 66: I have changed this order to avoid NPE when logging.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2947,25 +2947,34 @@ public Response getMakeDataCountMetricCurrentMonth(@PathParam("id") String idSup
...
-                execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), false, GetDatasetStorageSizeCommand.Mode.DOWNLOAD, getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers))))), getRequestUser(crc));
+    public Response getDownloadSize(@Context ContainerRequestContext crc,
+                                    @PathParam("identifier") String dvIdtf,
+                                    @PathParam("versionId") String version,
+                                    @QueryParam("ignoreOriginalTabularSize") boolean ignoreOriginalTabularSize,
+                                    @Context UriInfo uriInfo,
+                                    @Context HttpHeaders headers) {
+        return response(req -> {
+            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers);
+            Long datasetStorageSize = ignoreOriginalTabularSize ? DatasetUtil.getDownloadSizeNumeric(datasetVersion, false)
```
landreev, line 2970: Hmm.  Backward compatibility is important... although I honestly have no idea how popular this api is (nobody seems to be using it on our instance here), and if it was originally implemented that way (to count both types) because somebody specifically requested it/had a reason to count both, or if it was just implemented like that unintentionally. 

I'm ok with leaving its default behavior backward compatible by default...
However, I'm assuming that the new UI will work the same way as the current page - with the user having a choice between downloading either of the 2 types, but never both at the same time, correct? In this case we need this api to also be able to *only* count the original sizes. So, maybe instead of the `ignoreOriginalTabularSize` boolean, something like `mode=(all|original|archival)`, with `all` being the default?

landreev:
Aside from my earlier comment, my bigger concern - at least in the longer term is the performance. 
Both the command and the DatasetUtil method rely on `version.getFileMetadatas();` The way EJB normally handles this underneath, this is a very expensive operation for a dataset with many thousands of files. When this is done in the context a page bean, at least we can initialize the version and all its filemetadatas and datafiles once, and then iterate through them as needed. But in the context of an API call, it will be initializing all that stuff from scratch, just to produce this size. 

I feel like the best solution would be to use custom database queries instead. I'm not insisting that this needs to be done in this PR, but we need to somehow document this as a todo that needs to be addressed. I'm assuming it will need to be addressed for the MVP - we do have monstrous datasets in our prod. with 25K and 40+K files.

GPortas:
@landreev

Thank you for your detailed comments.

Maybe I have been too strict with backwards compatibility, which on this endpoint is not so important.

I agree with what you mention about performance and it is something that I have not considered at all and that I appreciate you mention it. I like the approach of using database queries + mode=(all|original|archival) and would like to at least evaluate the implementation for this PR. I will let you know.

Thank you!

pdurbin:
> added by request from an outside project

Yeah. The use case still makes sense at least:

"I'm an integrating system (Whole Tale, etc) and I want to provide some expectation in my UI about how long something will take when pulling files from Dataverse"

From https://github.com/IQSS/dataverse/issues/6524#issuecomment-574814237

landreev:
Just want to mention here that, instead of creating custom database queries for calculating these sizes, another alternative would be to try and utilize the Eclipse left join "hints", to avoid the followup datatbase queries on the datafiles and datatables; so,  instead of doing `version.getFileMetadatas()` etc., executing something along the lines of:
```
TypedQuery<FileMetadata> query = em.createQuery("SELECT fm FROM FileMetadata fm WHERE ...", FileMetadata.class)
   .setParameter("datasetVersionId", versionId)
   .setHint("eclipselink.left-join-fetch", "fm.dataFile")
   .setHint("eclipselink.left-join-fetch", "fm.dataFile.dataTables");
```
etc. And that would look up all the datafiles and datatables in one query, without the N additional queries... And then we can iterate through the list as before. (There's a chance that you don't need the `.setHint` for `fm.dataFile` there - the second one may be enought to look up both). 
In other words, the same approach as in `DatasetServiceVersionBean.findDeep()`, but only looking up the bare minimum of what we need for the filesize calculation. 

The only way to find out which approach is better would be test and time it on a dataset with 10Ks of files. A custom query would almost certainly be faster; but the difference may or may not be significant. The left-join approach above may feel a little cleaner, but I'm assuming you pay an extra cost in memory.

landreev:
I may need to take back the "bare minimum" part though - if we ever use that left join hint approach, we *have* to add hints for any other 1:1 relationships there - or else we are back to the EJB executing N additional queries for them...
Regardless, that was not really a practical suggestion for this specific use case, just wanted to document this alternative for other optimization situations.

GPortas:
Thank you very much for your suggestions @landreev.

Finally I have built the queries through QueryDSL. You can see the implementation in the latest commits. There are two new queries: one for the original file sizes and another for the archival sizes. Depending on the value of FileDownloadSizeMode (All, Archival, Original) one or both will be executed.

I have enabled SQL logging through these config lines (https://github.com/IQSS/dataverse/blob/develop/src/main/resources/META-INF/persistence.xml#L18) to check the SQL query that is produced by QueryDSL.

queries_filedownloadsize

FWIW - I have not analyzed the generated queries too much or benchmarked them, but you can see that JOINS are not used. 

As you say, the performance of this solution should be tested with a large dataset. In any case, this solution must be more optimal than the previous one without custom queries.

I have created [this issue](https://github.com/IQSS/dataverse-frontend/issues/184) to do a performance test of the SPA once the entire Dataset page in view mode is completed. Within the scope of that issue we can set up a heavy dataset scenario so we can test the changes in this PR in action, as well as pagination or any other critical performance aspect.

landreev:
> FWIW - I have not analyzed the generated queries too much or benchmarked them, but you can see that JOINS are not used.

I'm still curious to see what the SQL query looks like underneath when the `.where(fileMetadata.dataFile.dataTables.isEmpty())` clause is added. The second query in the screenshot simply counts *all* the `dataFile.filesize` in the version (i.e., `mode=Archival`). 
I can of course build your PR and check myself. I am actually not sure how to implement this case (count the sizes, except for tab. files) elegantly without using JOINs.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -3630,4 +3630,99 @@ public void deaccessionDataset() {
...
+        getDownloadSizeResponse.then().assertThat().statusCode(OK.getStatusCode())
+                .body("data.storageSize", equalTo(expectedTextFilesStorageSize));
+
+        // Upload test tabular file
+        String pathToTabularTestFile = "src/test/resources/tab/test.tab";
+        Response uploadTabularFileResponse = UtilIT.uploadFileViaNative(Integer.toString(datasetId), pathToTabularTestFile, Json.createObjectBuilder().build(), apiToken);
+        uploadTabularFileResponse.then().assertThat().statusCode(OK.getStatusCode());
+
+        // Get the original tabular file size
+        int tabularOriginalSize = Integer.parseInt(uploadTabularFileResponse.getBody().jsonPath().getString("data.files[0].dataFile.filesize"));
```
landreev, line 3672: Minor thing, but why not just `= 157`? (the known size of `src/test/resources/tab/test.tab`). 
It's a tiny file that should get ingested almost instantly. There may be a non-zero chance of the size in the response above already being the size of the generated archival file (I'm not sure... but it should be safest to use the known size, I think).

landreev:
Sorry, my bad - if you have read it already, please ignore my last comment about the query - I missed that `AND (COUNT(...) = ...)` clause in it. Will edit.

src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
```diff
@@ -230,4 +262,29 @@ private void applyOrderCriteriaToGetFileMetadatasQuery(JPAQuery<FileMetadata> qu
...
         }
     }
+
+    private long getOriginalTabularFilesSize(DatasetVersion datasetVersion) {
+        JPAQueryFactory queryFactory = new JPAQueryFactory(em);
+        Long result = queryFactory
+                .from(fileMetadata)
+                .where(fileMetadata.datasetVersion.id.eq(datasetVersion.getId()))
+                .from(dataTable)
+                .where(fileMetadata.dataFile.dataTables.isNotEmpty().and(dataTable.dataFile.eq(fileMetadata.dataFile)))
```
landreev, line 272: Yes, `.where(dataTable.dataFile.eq(fileMetadata.dataFile))` 
should be sufficient - and it simplifies the resulting query a bit.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
GPortas: I agree. If it makes it safer, it's worth to update it.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2947,25 +2947,40 @@ public Response getMakeDataCountMetricCurrentMonth(@PathParam("id") String idSup
...
-                execCommand(new GetDatasetStorageSizeCommand(req, findDatasetOrDie(dvIdtf), false, GetDatasetStorageSizeCommand.Mode.DOWNLOAD, getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers))))), getRequestUser(crc));
+        return response(req -> {
+            DatasetVersionFilesServiceBean.FileDownloadSizeMode fileDownloadSizeMode;
+            try {
+                fileDownloadSizeMode = mode != null ? DatasetVersionFilesServiceBean.FileDownloadSizeMode.valueOf(mode) : DatasetVersionFilesServiceBean.FileDownloadSizeMode.All;
+            } catch (IllegalArgumentException e) {
+                return error(Response.Status.BAD_REQUEST, "Invalid mode: " + mode);
+            }
+            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, version, findDatasetOrDie(dvIdtf), uriInfo, headers);
+            long datasetStorageSize = datasetVersionFilesServiceBean.getFilesDownloadSize(datasetVersion, fileDownloadSizeMode);
```
landreev, line 2977: Hey, I'm wondering if I gave you bad advice, to move this size calculation out of the command. The command provided one extra function - the authorization (requiring the `ViewUnpublishedDataset` permission on draft versions); which we no longer appear to be enforcing in the above (?). 
I can't imagine this information, this download size, being truly sensitive... but we apparently used to require that permission, so at the very least we need to confirm if it is necessary or not. 
Sorry for the extra confusion.

landreev: Uh, never mind? That authorization check is still being performed, inside the `getDatasetVersionOrDie()`.

GPortas:
> @GPortas I am seeing the same download size for all variations of the endpoint.

@kcondon Sorry! I forget to update the PR description with the testing instructions with latest changes.

PR #9959, opened 2023-09-26
Faster redeployment for developers, skip deploy, remove version from war in dev

poikilotherm:
**What this PR does / why we need it**:
Containers provide a bad DX at the moment with editing XHTML files. This PR enables transfering modified files from the host into the running deployment inside the container.

**Which issue(s) this PR closes**:

- Closes #9590

**Special notes for your reviewer**:
Simply start the containers and, when using Netbeans, edit an XHTML file. After a page reload the change should be picked up without any manual intervention. It does not require you to run Payara from Netbeans.

Note: as this touches the base image, you will need to package this locally first and build your app image on top of it!

**Suggestions on how to test this**:
See above.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Nope

**Is there a release notes update needed for this change?**:
Nope

**Additional documentation**:
https://dataverse-guide--9959.org.readthedocs.build/en/9959/container/dev-usage.html

Issue #9590, opened 2023-05-12
Faster turn-around times for devs using containers


poikilotherm:
Part of the [containerization strategy](https://docs.google.com/document/d/14DHDB24Cp_kzpYqhHCKCtnzOw8_WuLOOONyqJHSsaYM/edit#), [milestone D](https://docs.google.com/document/d/14DHDB24Cp_kzpYqhHCKCtnzOw8_WuLOOONyqJHSsaYM/edit#heading=h.ysp9fn6ri34v).

With #9584 we will recommend stopping, repackaging and start all containers again after code changes. This takes, depending on your local hardware, 60-120 seconds. Investigate in making this faster is the task of this issue.

pdurbin:
We talked about this about 45 minutes into today's container meeting: https://ct.gdcc.io

pdurbin:
I'm just noting that today (and last Friday), I've been hacking around in xhtml and the feedback loop is unbearable with containers. I'm going to install Payara 6 directly on my Mac so I can hack on xhtml and have it deployed by Netbeans basically instantaneously.

pdurbin:
@beepsoft just started a related thread at https://dataverse.zulipchat.com/#narrow/stream/375812-containers/topic/Payara.20admin.20console.20redirects.20to.20https/near/399050958

PR #9961, opened 2023-09-26
allow fast xhtml edits in Docker

pdurbin:

It is a nightmare to edit xhtml files in our containerized dev environment without a fix like this. You have to recompile the war file and redeploy. It's extremely slow.

None, but relates to:

- #9590

@poikilotherm has what is probably a better fix here:

- #9959

However, I haven't looked into it. I'm making this pull request because it's small and only touches our docker compose file. I used this fix in anger this week while hacking on xhtml.

I went with a refresh period of 1 second because that's what we document over at https://guides.dataverse.org/en/6.0/developers/classic-dev-env.html#configure-your-development-environment-for-gui-edits

The default is -1. Here are the docs from https://jakarta.ee/specifications/faces/3.0/jakarta-faces-3.0.html#a6088

> jakarta.faces.FACELETS_REFRESH_PERIOD — When a page is requested, what interval in seconds should the compiler check for changes. If you don’t want the compiler to check for changes once the page is compiled, then use a value of -1. Setting a low refresh period helps during development to be able to edit pages in a running application.The runtime must also consider the facelets.REFRESH_PERIOD param name as an alias to this param name for backwards compatibility with existing facelets tag libraries.

p.s. The failing test at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-9961/1/consoleFull MoveIT.testMoveLinkedDataset which is known to be flaky. Changing the docker compose file doesn't have any effect on Jenkins tests anyway.

- Uncomment the line about volumes and keep it at 6.0 or bump as needed to the current version.
- `mvn -Pct clean package docker:run` per the quickstart: https://guides.dataverse.org/en/6.0/developers/dev-environment.html#quickstart
- Edit an xhtml page.
- Refresh your browser.
- The change should be visible, like magic! 🪄 ✨ 

No.

No. Maybe I'll mention it on the dev list.

**Additional documentation**:

None.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -32,7 +32,7 @@ services:
     volumes:
       - ./docker-dev-volumes/app/data:/dv
       - ./docker-dev-volumes/app/secrets:/secrets
-      # Uncomment for changes to xhtml in Netbeans to be deployed immediately.
+      # Uncomment for changes to xhtml to be deployed immediately (if supported your IDE or toolchain).
       # Replace 6.0 with the current version.
       # - ./target/dataverse-6.0:/opt/payara/deployments/dataverse
     tmpfs:
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -18,6 +18,7 @@ services:
       - DATAVERSE_AUTH_OIDC_CLIENT_ID=test
       - DATAVERSE_AUTH_OIDC_CLIENT_SECRET=94XHrfNRwXsjqTqApRrwWmhDLDHpIYV8
       - DATAVERSE_AUTH_OIDC_AUTH_SERVER_URL=http://keycloak.mydomain.com:8090/realms/test
+      - DATAVERSE_JSF_REFRESH_PERIOD=1
     ports:
       - "8080:8080" # HTTP (Dataverse Application)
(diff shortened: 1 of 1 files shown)


docker-compose-dev.yml
```diff
@@ -27,6 +28,9 @@ services:
     volumes:
       - ./docker-dev-volumes/app/data:/dv
       - ./docker-dev-volumes/app/secrets:/secrets
+      # Uncomment for changes to xhtml in Netbeans to be deployed immediately.
```
johannes-darms, line 31: Could we comment a bit more broader for none netbeans users? Something like:
`Uncomment for changes to xhtml files on your local development file system to be mounted, read and deployed immediately to the container.`

pdurbin: I guess but @poikilotherm made it sound like what Netbeans does is somewhat unique.

What if I change it to this?

`# Uncomment for changes to xhtml to be deployed immediately (if supported your IDE or toolchain).`

pdurbin: @johannes-darms I didn't hear an objection so I switched to my suggestion: 5fa3f46ec6

donsizemore (approved):

PR #9964, opened 2023-09-28
Always add validity field when indexing a dataset

### LINKED ISSUE #9836: MyData metadata validity facet violates Don't Make Me Think
I was just playing around with the "Creating datasets with incomplete metadata through API" feature developed in this PR:

- #8940

I'm on Dataverse 5.14.

It's a cool feature, but I think MyData suffers a bit from it.

Imagine the following scenario:

- a user creates a collection (dataverse), a dataset, and uploads a file
- time passes
- the user is looking for their dataset and files

Below is what might happen.

## User clicks MyData 

Screen Shot 2023-08-28 at 12 37 05 PM

## User sees "sorry no, results are found"

Screen Shot 2023-08-28 at 12 37 19 PM

## Now what? 🤔 

Realistically, at this point the user probably contacts support. Where's my data? Help!

A user with time on their hands and a penchant for clicking things might explore further.

## What if I uncheck "incomplete metadata"?

Nope, nothing.

Screen Shot 2023-08-28 at 12 39 04 PM

What if I uncheck "incomplete metadata" AND "valid"?

Finally, we are getting somewhere! There's my dataset and my collection (dataverse)! But where's my file?

Screen Shot 2023-08-28 at 12 39 14 PM

## Check files to see files

As usual, files are not selected by default but if you check "files" you can see them. This is out of scope for this issue.

Screen Shot 2023-08-28 at 12 39 19 PM

## In conclusion

When a user visits MyData they should see their data, not "Sorry, no results were found." They shouldn't have to uncheck boxes to see their data.

ErykKul:
Right now, the validity field is only added to solr documents when indexing datasets if the allow-incomplete-metadata feature is on. As a result, each time when the feature is turned on, reindexing of datasets is needed, since datasets added when the feature was turned off do not have that field present and are filtered out from my data page (regardless of the show-validity-filter setting value). With this PR the field is always added, and the datasets do not need to be reindexed each time the feature is turned on.

- Closes #9836 

Before applying this PR, turn the allow-incomplete-metadata feature OFF, reindex the datasets, turn that feature ON and go to my data page. All datasets are filtered out unless some checkboxes are unchecked or all datasets are reindexed again with the feature ON. Do the same with this PR and the datasets should be visible.


qqmyers (approved):
LGTM

PR #9966, opened 2023-09-28
Submit Incomplete Dataset doc clarification regarding :MetadataLanguages

### LINKED ISSUE #9942: Clarification for the “Submit Incomplete Dataset” feature
_entrepot.recherche.data.gouv.fr team_

* What happens?

Version 5.14 of Dataverse allows the creation of partial dataverse with the API with only the author's information:

https://guides.dataverse.org/en/5.14/api/native-api.html#submit-incomplete-dataset

It's not always the case !
When the setting ":MetadataLanguages"
is configured and the dataverse has the value "Dataset Metadata Language" to "Chosen at Dataset Creation":

https://guides.dataverse.org/en/5.14/installation/config.html?#allowing-the-language-used-for-dataset-metadata-to-be-specified

Capture d’écran du 2023-09-20 16-55-43

In this case, a language that is part of the ":MetadataLanguages" list must be declared in the incomplete dataset.

To users wanting to create a partial dataset with the API

Is this the expected behavior?
If so, add clarification in the documentation for this scenario?

5.14

ErykKul:
It clarifies dataverse.api.allow-incomplete-metadata documentation in the case when the ":MetadataLanguages" setting is used. It also reminds of the necessity of reindexing the datasets after updating of the Solr schema.

- Closes #9942

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -525,10 +525,16 @@ Submit Incomplete Dataset
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 **Note:** This feature requires :ref:`dataverse.api.allow-incomplete-metadata` to be enabled and your Solr
-Schema to be up-to-date with the ``datasetValid`` field.
+Schema to be up-to-date with the ``datasetValid`` field. If not done yet with the version upgrade, you will
+also need to reindex all dataset after enabling the :ref:`dataverse.api.allow-incomplete-metadata` feature.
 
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
https://dataverse-guide--9966.org.readthedocs.build/en/9966/api/native-api.html#submit-incomplete-dataset looks good to me, formatting-wise. I'm trusting @ErykKul and @stevenferey that the behavior described is accurate.

p.s. @ErykKul I'm reminded that you put "Submit Incomplete Dataset" (the weird case) ahead of "Submit Dataset" (the normal case). Would you be willing to create a future PR to put the normal case first?

PR #9967, opened 2023-09-28
9913 template preview - fix display of Terms of Access

### LINKED ISSUE #8781: Collapse not working properly for Restricted Files and Terms of Access panel
In looking at code for other reasons, I ran across https://github.com/IQSS/dataverse/blob/365e7beb944afce8c44b241e98b5ba03ebc509ed/src/main/webapp/dataset-license-terms.xhtml#L248 where it looked like a typo using the lowercase datasetPage variable (which is a boolean) should be causing problems. However, it doesn't cause the page to fail. Investigating further, it appears that the class is never being set because jsf:class isn't correct/valid. Apparently, the fact that jsf:class isn't valid stops the logic from ever being called.

In the UI, this manifests as the panel within the Restricted Files and Terms of Access not having any collapse info on it, resulting in the panel being open to start and the first click not closing it. (Once there's a user click, the collapse class is added and the toggle starts to work).

If I replace jsf:class with class, I then get a page error due to datasetPage not having a hasValidTermsOfAccess method. 

If I then replace datasetPage with DatasetPage to make it work, the div gets class "collapse" when there are valid terms which causes the panel (with terms you  should see) to start closed. 

I think the correct logic would be to have valid terms result in setting the class to "collapse in" and otherwise to just have the class be "collapse" - valid to toggle but closed to start.

If I change to that, I think the dataset page then works correctly/at least reasonably to me. However, this fragment gets called from the template pages as well (where DatasetPage is not defined) so making the fix outlined here causes templates to fail.

I think adding additional logic for that, resulting in a final:

`<div id="panelCollapseTOA" class="#{!datasetPage or DatasetPage.hasValidTermsOfAccess ? 'collapse in' : 'collapse'}">`

makes both dataset and template pages work correctly.

Overall - a minor issue that I stumbled across while trying to find an example of calling DatasetPage from this xhtml fragment. I think the fix here works (assuming I understand the original intent) - I mostly wanted to investigate so I was sure I understood enough about why this wasn't working and wasn't causing an error (aside from the UI issue) to be able to make other changes. 

If this makes sense, I can make a one-line PR (or someone else can).

**Which version of Dataverse are you using?** develop branch

### LINKED ISSUE #8918: Template view issues
In working through #8917, I noticed

- that a template with terms of access defined does not show the ToA in the template view mode - due to [line 267](https://github.com/IQSS/dataverse/blob/1e528090e37d0236fcc3c69e041763b56a711834/src/main/webapp/dataset-license-terms.xhtml#L267),
-  the collapse state for the TofA pane in template view isn't initially set correctly due to use of[ jsf:class instead of class](https://github.com/IQSS/dataverse/blob/1e528090e37d0236fcc3c69e041763b56a711834/src/main/webapp/dataset-license-terms.xhtml#L248), 
- there is no display of template fileaccessrequest status in the template view - the existing html isn't rendered in the template view and, if it were, the logic inside appears to check the dataset setting and, due to issues like those described in #8917, the template view would always default to fileaccessrequest not set due to a silent null. 
- [the check re: displaying the restricted file count](https://github.com/IQSS/dataverse/blob/1e528090e37d0236fcc3c69e041763b56a711834/src/main/webapp/dataset-license-terms.xhtml#L257) appears unnecessary as the div enclosing it isn't shown unless you are on the dataset page and there are restricted files.

I didn't have time to look for further issues, so might suggest a bit of a spike to have someone check that template edit/view works as expected after #8917 and see if there are more before addressing the things above. I think most are straight forward to fix though the collapse state one may take some work as both the collapse state and the up/down chevron have to be synched and it wasn't clear to me if the design was to start with that panel open or closed.

**What steps does it take to reproduce the issue?** as described

* Which page(s) does it occurs on? Viewing templates.

develop before/after #8917

### LINKED ISSUE #9913: Dataset template preview/view: the template preview should display prefilled texts for "restricted files & terms of access" section

Create a template and prefill the "restricted access and terms of access" section and save. Use the "preview/view" icon on the template success page to view text that was entered

* When does this issue occur?

After you complete the TERMS section and click on the view icon

On the preview/view page after successfully creating a template

* What happens?
The text that was filled into the template is not visible for preview/view for this section: "restricted files & terms of access" 

All users

Expected to see the text that was populated into this field

Most recent version

None

**Screenshots:**

Text that was entered into template: 
Screen Shot 2023-09-13 at 5 30 05 PM

What is seen with the template preview icon:
Screen Shot 2023-09-13 at 6 29 16 PM

sekmiller:
Allows Terms of Access to display on template view/preview

Closes #9913 template preview should display terms of access, etc.
Closes #8918 Template View issues.
Closes #8781 Collapse not working for Restricted files TOA

Add terms of access and see that it is displayed with preview on the manage templates page


qqmyers (approved):
Haven't tested but I did walk through code to identify that the DatasetPage.restrictedFileCount > 0 was what was shutting this off in the template preview and this fix looks like it should work - easy to see in QA.

sekmiller:
I'm going to pull this back because Gustavo asked me to take a look at related issues to see if they would be fixed by this or if they are low-hanging fruit to add to this, but thanks for the quick action @qqmyers

sekmiller:
@qqmyers can you take another look at this in light of #8918 ? Thanks!

src/main/webapp/dataset-license-terms.xhtml
```diff
@@ -277,18 +280,18 @@
...
-                       <div class="form-group" jsf:rendered="#{!publicStore and ( datasetPage == true and DatasetPage.hasRestrictedFiles)}">
+                       <div class="form-group" jsf:rendered="#{!publicStore and ((datasetPage == true and DatasetPage.hasRestrictedFiles) or managePage==true)}">
                            <label for="datasetForm:tabView:metadata_RequestAccessSelected" class="col-sm-3 control-label">
                                #{bundle['file.dataFilesTab.terms.list.termsOfAccess.requestAccess']}
                                <span class="glyphicon glyphicon-question-sign tooltip-icon"
                                      data-toggle="tooltip" data-placement="auto right" data-original-title="#{bundle['file.dataFilesTab.terms.list.termsOfAccess.requestAccess.title']}"></span>
                            </label>
                            <div id="metadata_RequestAccessSelected" class="col-sm-9">
-                               <p jsf:rendered="#{!DatasetPage.workingVersion.draft}">
+                               <p jsf:rendered="#{datasetPage == true and !DatasetPage.workingVersion.draft}">
```
qqmyers, line 290: does this need `or managedPage == true`?

qqmyers (commented):
Looks good - one question about the request access selected entry showing on the template page.

src/main/webapp/dataset-license-terms.xhtml
sekmiller: No, because it's getting the values from the Dataset Page so managedPage would never be true. - But we need it down below (294) because the values are coming from the termsOfUseAndAccess which is passed in as part of the include.

qqmyers (approved):

PR #9923, opened 2023-09-15
9314 Get JSON object from JsonUtil

bencomp:
**What this PR does / why we need it**:

Work around a potential resource leak by using the existing `JsonUtil` (as suggested by @qqmyers)

**Which issue(s) this PR closes**:

Closes #9314 

**Special notes for your reviewer**:
You're welcome, @pdurbin.

**Suggestions on how to test this**:
See Sonarcloud's number of issues go down.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
No

**Is there a release notes update needed for this change?**:
No

**Additional documentation**:
Perhaps add a note to the developer docs to make sure this doesn't happen again?

Issue #9846, opened 2023-08-29
Set up Sonar


pdurbin:

bencomp:
See also (linking in the other direction)

- #775
- #3950

I wonder what Sonar is going to be used for. There's a checkstyle workflow (that could be updated to highlight more issues) and similar actions for other tools exist.

(I shouldn't get involved... I'm just happy that old issues are not completely forgotten.)

pdurbin:
Hey, @bencomp, thanks for stopping by. 😄 

I just upgraded Netbeans from 14 to 18 and it was surprisingly easy to set up SolarLint this time. (At least it seems to be going well.)

The plugin was easy to find this time:

![Screen Shot 2023-08-29 at 4 54 41 PM](https://github.com/IQSS/dataverse/assets/21006/f4d744a0-4518-46ea-acff-02afd88823dc)

I tried this "analyze" button...

![Screen Shot 2023-08-29 at 4 59 33 PM](https://github.com/IQSS/dataverse/assets/21006/21b0570c-6e1c-41af-b7db-9ab8c2c834fd)

... and stuff started happening immediately:

![Screen Shot 2023-08-29 at 5 02 42 PM](https://github.com/IQSS/dataverse/assets/21006/802f758e-d89d-46c5-9b92-0cf91988d00e)

Maybe this is because we already set up https://sonarcloud.io/project/overview?id=IQSS_dataverse earlier. 🤔 

I'm sort of surprised it's already analyzing and doing stuff. I assumed I'd have to enter some credentials for my new sonarcloud.io account or set up SonarCube server. 🤔

poikilotherm:
As I've mentioned before - the connection is necessary to download and apply customized rulesets. Either disabling / reconfiguring or adding new rules.

pdurbin:
I still want Sonar to be set up but I'm closing this and will let someone else be the champion.

poikilotherm:
As per discussed during tech hour, let's do this!

PR #9320, opened 2023-01-24
Sonarcube Fixes

qqmyers:
**What this PR does / why we need it**: This is a branch/draft PR to collect tech-debt fixes for sonarcube.

**Which issue(s) this PR closes**:

Closes #9280
Closes #9282

**Special notes for your reviewer**:

**Suggestions on how to test this**: Regression testing plus running through Sonarcube to assure the matching issues are gone.
#9280 is in the Files API {id}/metadata call (trivial fix replacing code with the same code in the JsonUtil class).
#9282 is a change in the tabular download subset functionality - not sure if there's already a test or guides example for that call. 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

Issue #9314, opened 2023-01-23
In JsonParser.java use try-with-resource for JsonReader jsonReader = Json.createReader(new StringReader(jsonString)); in method remapGeographicCoverage(...)


rtreacy:

mreekie:
Sizing:
- If this is a single file,  then this should be a 3.
- simple fix.

qqmyers:
Using the JsonUtil.getJsonObject() or getJsonArray methods would be better than just adding a try with resources.

pdurbin:
@bencomp are you interested in creating a PR to fix this issue? As of 4e1bc5bccd this is the line in question:

```
diff --git a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
index febb785cd9..8f3ffd2671 100644
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
@@ -682,7 +682,7 @@ public class JsonParser {
         // convert DTO to datasetField so we can back valid values.
         Gson gson = new Gson();
         String jsonString = gson.toJson(geoCoverageDTO);
-        JsonReader jsonReader = Json.createReader(new StringReader(jsonString));
+        JsonReader jsonReader = Json.createReader(new StringReader(jsonString)); // HERE
         JsonObject obj = jsonReader.readObject();
         DatasetField geoCoverageField = parseField(obj);
``` 

We have a snapshot from Sonarcloud from a couple weeks ago that also shows it: https://sonarcloud.io/code?id=IQSS_dataverse&selected=IQSS_dataverse%3Asrc%2Fmain%2Fjava%2Fedu%2Fharvard%2Fiq%2Fdataverse%2Futil%2Fjson%2FJsonParser.java&line=685

![Screen Shot 2023-09-15 at 10 44 47 AM](https://github.com/IQSS/dataverse/assets/21006/ad7d3c23-559a-4610-844d-e8b71eba2132)

![Screen Shot 2023-09-15 at 10 44 01 AM](https://github.com/IQSS/dataverse/assets/21006/307a107f-56f2-434e-97f6-e0804c94cfe5)

PR #9980, opened 2023-10-05
Use JsonUtil.getJsonObject to prevent resource leaks

bencomp:
This is more or less a follow-up to #9923 and #9320, fixing additional instances of potentially unclosed `StringReader`s and/or `JsonReader`s.

None; I did not create a separate issue for these files. They would have been copies of #9314.

#9314 includes the context for the problem.
I added new *TODOs*, because the flow of the code (before and after these changes) may cause `NullPointerException`s. I have asked others not to do this in PRs, but I needed to get out of the rabbit hole. I may open a separate issue pointing to the TODOs and refer back to this PR.

The same way as #9923.

No

No

**Additional documentation**:
https://sonarcloud.io/organizations/bencomp/rules?open=java%3AS2095&rule_key=java%3AS2095 is the SonarCloud rule that triggered me to start this PR. I look forward to seeing #9846 resolved 😃


qqmyers (approved):
Looks good!

PR #9984, opened 2023-10-06
External Controlled Vocab: add missing db constraints, add cvoc transaction and try/catch

### LINKED ISSUE #9983: Missing constraints allow duplicate values for external CVoc and oaiset
The ExternalVocabularyValue and OaiSet classes have an @Column annotation intended to put a unique constraint on the columns in the externalvocabularyvalue (uri column) and oaiset (spec column) tables. For the externalvocabularyvalue table, parallel edits of different datasets including the same external vocab uri can result in duplicate values in the table which then causes failures in retrieval when getSingleValue() is used with the query.

The tables should have these constraints on them. Further, in the case of external vocab values, it makes sense to isolate the update of this table in its own transaction - this minimizes the time during when parallel dataset edits can actually cause a problem, and assures that if an attempt to write a duplicate value is made (and now fails due to the constraint) that the overall dataset transaction can succeed.

**Which version of Dataverse are you using?** 5.14/6.0 - the problem should not be new but would only be seen when parallel edits of datasets are made with external controlled vocabulary values configured and with the same uri appearing in more than one dataset.

PR to follow.

qqmyers:
**What this PR does / why we need it**: Per the issue, it adds missing db constraints and improves the code for registering new external vocab values by isolating the update in a small transaction and stopping any attempt to add a duplicate value from killing the overall dataset edit/other transaction being done.

Closes #9983 

**Special notes for your reviewer**: As of now, we don't know why the constraints are not there. There is some dv-tech channel slack discussion about it but no conclusions.
If the db has duplicate values now, the current flyway script would cause install failure. However, it appears very unlikely that either table would have duplicates now. The code around both tables checks for existing values before saving a new one, so only race conditions can create duplicates. Problems with the external vocab table immediately start causing edit failures, so in addition to a race condition being unlikely (parallel changes and ext. vocab enabled), the problem would be immediately noticed. For the oaiset table, it seems even more unlikely that duplicates would exist as the code there also appears to check for an existing value with the same spec and there isn't any parallel case where one would expect multiple oaisets to be being created at once. It isn't clear to me that we can automate removing duplicates if they do exist, especially for oaiset where the descriptions and other fields may differ. My guess is that handling this with a release note instruction might be enough as problems should be extremely rare.

Regression: verify that OAI sets can be created/used and that external vocab values can be added to datasets and their metadata appears in the exported metadata. IN either case it appears to be hard to test the improvement from having the constraints as one would have to run multiple adds of an OAISet/multiple edits to datasets with the same ext. vocab value terms to be able to trigger the problem. (Prior to the PR, parallel changes would add duplicates to the database which you could see there and, at least for ext. vocab, you'd see failures related to multi-value exceptions when trying to retrieve ext. vocab values. After the PR, you shouldn't ever have duplicate values. For OAISets you might get errors when trying to run parallel ops. For the external vocab case, you'd see a fine message in the log about an attempt to write a duplicate value but the overall edit/metadata change would succeed.)

One other thing to test - have a dataset that has uses a type with controlled vocabulary values and then change the definition of the type to not use CVV. Make sure that this does not adversely affect the functionality (display, indexing, etc) of the existing dataset. (adding this because if the change to line 287 of DatasetFieldType.java.

**Additional documentation**:


PaulBoon:
@qqmyers Can this be merged into the next release or is there something blocking it, besides the review?

cmbz:
2024/01/30
- Prioritized and moved to Sprint Ready after Slack conversation with @qqmyers and @scolapasta

doc/release-notes/9983-unique-constraints.md
```diff
@@ -0,0 +1,14 @@
...
+and
+
+SELECT spec, count(*) FROM oaiset group by spec;
+
+and then removing any duplicate rows (where count>1).
+
+
+
+
+TODO: Add note about reloading metadata blocks after upgrade.
```
scolapasta, line 14: is this a todo that we want before we qa / merge this?

scripts/api/data/metadatablocks/astrophysics.tsv
```diff
@@ -2,13 +2,13 @@
 	astrophysics		Astronomy and Astrophysics Metadata													
 #datasetField	name	title	description	watermark	 fieldType	displayOrder	displayFormat	advancedSearchField	allowControlledVocabulary	allowmultiples	facetable	displayoncreate	required	parent	metadatablock_id
 	astroType	Type	The nature or genre of the content of the files in the dataset.		text	0		TRUE	TRUE	TRUE	TRUE	FALSE	FALSE		astrophysics
-	astroFacility	Facility	The observatory or facility where the data was obtained. 		text	1		TRUE	TRUE	TRUE	TRUE	FALSE	FALSE		astrophysics
```
scolapasta, line 5: I assume these relate to the todo? What is the reason for changing these values? (I probably just missed it, now I'm just curious)

src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
```diff
@@ -284,7 +284,7 @@ public void setDisplayOnCreate(boolean displayOnCreate) {
     }
     
     public boolean isControlledVocabulary() {
-        return controlledVocabularyValues != null && !controlledVocabularyValues.isEmpty();
+        return allowControlledVocabulary;
```
scolapasta, line 287: is this change because we think the definition is enough (and more important than actual values). Not sure, but could we ever get a case where a field changes FROM allow controlled vocab to disallowing, but we still have old datasets with CVV?

scolapasta (commented):
Overall, looks good, had a few basic questions.

doc/release-notes/9983-unique-constraints.md
qqmyers: I think that is just a note that the main release notes should include the reload block instructions for these blocks.

scripts/api/data/metadatablocks/astrophysics.tsv
qqmyers: Not related to the todo. I think these are just inconsistencies - the block says they are controlled vocab fields but there are no vocab values for them in the block.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
qqmyers: @janvanmansum may know more here but my understanding is that this is primarily being used to decide if you display the controlled vocab list for metadata input or not. So with the change if you set the field to not be cvv, it will now show as a text input, whereas before, if you ever had cvv values in the db, the field shows as cvv going forward. I don't think it affects things if there are specific field values in datasets that have to display, its just about the type.

scripts/api/data/metadatablocks/astrophysics.tsv
scolapasta: That makes sense - I just wonder how why they were ever true in the first place?

src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
scolapasta: @qqmyers that makes sense. My concern would be if there are other places that use it for either display or indexing tasks and how this would affect a dataset that had controlled vocabulary values before the type was changed (rare, but could happen. I'll just add a note in the testing section to see if that can be looked at and make sure things look fine.

scolapasta (approved):

src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
qqmyers: Looking at code outside the PR- there is an isAllowControlledVocabulary() method above that returns allowControlledVocabulary (so a duplicate now). Perhaps the xhtml or other places where we need this change could just switch to use that method.

FWIW: It does seem to be an odd case to have former CVVs - we're potentially relying on the metadatablock loader to keep old values/ people to keep the old values in the tsv when the field no longer uses them, etc. (I can hardly wait until they make the field CVV again and want new values and not have the old ones show up as choices ...)

scolapasta: I do agree it's an edge case. It's possible any time we would switch from cvv to not, we'd write a db script to convert the values anyway?

doc/release-notes/9983-unique-constraints.md
jp-tosca: Do you guys think we should clarify that whoever makes the release notes has to do this? I am not sure it is clear enough with the **_TODO:_** probably we could just mention it on Slack but we run the risk of someone missing the message 🤷🏼 @scolapasta @qqmyers

PR #9991, opened 2023-10-10
Fix code coverage reports after introducing QueryDSL

### LINKED ISSUE #9990: QueryDSL APT processing break JaCoCo code coverage reporting
As can be seen in the action logs ([here's an example](https://github.com/IQSS/dataverse/actions/runs/6469750216/job/17564690918), see under "Deposit Code Coverage"), the Coveralls plugin cannot find sources for classed generated by QueryDSL.

Usually these should be ignored by JaCoCo, but they are present as the default annotation [`javax.annotation.processing.Generated`](https://docs.oracle.com/en/java/javase/17/docs/api/java.compiler/javax/annotation/processing/Generated.html) has retention scope `SOURCE` and is no longer visible to JaCoCo.

[CoverAll Maven plugin doc says: do not include generated sources.](https://github.com/trautonen/coveralls-maven-plugin#faq) (Which makes sense - why would we count that...)

poikilotherm:
The Coveralls report is broken because JaCoCo's XML report file contains generated source files (and cannot be made to exclude them). The Coveralls Maven Plugin does not skip these classes and breaks of trying to read nonexisting source files.

Closes #9990

- [ ] Figure out a way how to either exclude the class references or
- [ ] Figure out a way how to make Coveralls skip ignored files (that have been excluded by some JaCoCo mechanism) or
- [ ] Migrate to some other reporting of coverage that is compatible with this...

None

Nope

Nope

**Additional documentation**:
None

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,14 +252,7 @@
             <artifactId>expressly</artifactId>
             <scope>provided</scope>
         </dependency>
-
-        <dependency>
-            <groupId>com.querydsl</groupId>
-            <artifactId>querydsl-apt</artifactId>
-            <version>${querydsl.version}</version>
-            <classifier>jakarta</classifier>
-            <scope>provided</scope>
-        </dependency>
+        
         <dependency>
             <groupId>com.querydsl</groupId>
             <artifactId>querydsl-jpa</artifactId>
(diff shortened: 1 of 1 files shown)


poikilotherm:
Without the changes here I definitely see the QueryDSL generated classes in the HTML report:

image

poikilotherm:
Superseded by #10041

PR #10004, opened 2023-10-11
fix S3 provider priorities

### LINKED ISSUE #10003: #9206 appears to break non-AWS S3 storage when role-based access is used
At QDR, we use storJ as one S3 provider. In 5.14+, it appears that access to storJ is broken. Tracing the issue, it appears that adding the role-based provider option back in as the preferred default in #9206 is the cause. Since we have role-based access set up for AWS S3 stores, it is taking precedence, even for our storJ S3 store which is configured to use profile based access, custom-endpoint, etc.

I'm submitting a PR that will allow role-based access to be turned off per store which will allow those stores to use profiles or static creds.

* When does this issue occur?
When role-based access is configured and some stores, e.g. those from other providers, can't use it.

* What happens?
All accesses to the affected store(s) fail - so file uploads/downloads, thumb accesses, etc. are affected

5.14+

#9205, #9206
Attn: @orazionelson

qqmyers:
**What this PR does / why we need it**: This PR addresses priority issues that existed between the 3 types of credential providers used by S3 stores. Before - the global role-based creds, if set, would override the per store profile or instance credentials. Now they do not, and static creds won't be ignored if/when a default profile exists in the Unix account running Payara.
The PR also includes removal of a redundant check that the bucket exists that would run for every file access.

Closes #10003 

**Special notes for your reviewer**: New priority rules are listed in a source comment. As this is a bug fix, it looks to me like the current guides say the right thing about priorities. So I don't see anything to change in the docs (without doing a big rewrite of the whole S3 setup section!). Similarly, since this is a bug fix for a reasonably rare case (role-based access was broken from 4.20 to 5.13, and you have to be using two different S3 providers), I don't know that anything is needed in the release notes.

Regression testing that S3 setup works. Depending on how far you want to go, you could check all the combinations of EC2 role based access is/isn't set, the S3 store has a .profile set or not, (the profile actually exists in the .aws/credentials file or not), and the static microprofile config is/isn't set and verify that setting a profile or static creds wins over the role based method, profile wins over static when both are set, setting nothing results in the default profile in the .aws/credentials file being used, etc. (I have tested at QDR that this fix works in the case where role-based creds exist and we have an S3 store that has to use a non-default profile instead.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
+++ b/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
@@ -4,6 +4,7 @@ import com.amazonaws.AmazonClientException;
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.HttpMethod;
 import com.amazonaws.SdkClientException;
+import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.auth.AWSCredentialsProviderChain;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.BasicAWSCredentials;
@@ -57,9 +58,11 @@ import java.nio.channels.WritableByteChannel;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 import java.util.function.Predicate;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1181,29 +1183,59 @@ private static AmazonS3 getClient(String driverId) {
...
+            // use, you should not use env vars or system properties to provide these, but 
+            // use the secrets config source provided by Payara.
+            Optional<String> accessKey = config.getOptionalValue("dataverse.files." + driverId + ".access-key", String.class);
+            Optional<String> secretKey = config.getOptionalValue("dataverse.files." + driverId + ".secret-key", String.class);
+            if (accessKey.isPresent() && secretKey.isPresent()) {
+                allowInstanceCredentials = false;
+                AWSStaticCredentialsProvider staticCredentials = new AWSStaticCredentialsProvider(
+                        new BasicAWSCredentials(
+                                accessKey.orElse(""),
+                                secretKey.orElse("")));
```
poikilotherm, line 1213:
```suggestion
                                accessKey.get(),
                                secretKey.get()));
```

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1181,29 +1183,59 @@ private static AmazonS3 getClient(String driverId) {
...
+                providers.add(staticCredentials);
+            } else if (s3profile == null) {
+                //Only use the default profile when it isn't explicitly set for this store when there are no static creds (otherwise it will be preferred).
+                s3profile = "default";
```
poikilotherm, line 1217: Why not do this above as else to the `!= null`?

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1181,29 +1183,59 @@ private static AmazonS3 getClient(String driverId) {
...
+            }
             // Add all providers to chain - the first working provider will be used
-            // (role-based is first in the default cred provider chain, so we're just
+            // (role-based is first in the default cred provider chain (if no profile or
+            // static creds are explicitly set for the store), so we're just
             // reproducing that, then profile, then static credentials as the fallback)
-            AWSCredentialsProviderChain providerChain = new AWSCredentialsProviderChain(instanceCredentials, profileCredentials, staticCredentials);
+
+            // As the order is the reverse of how we added providers, we reverse the list here
+            Collections.reverse(providers);
```
poikilotherm, line 1235: Instead of reversing, wouldn't it make sense to do the addition in correct order? (so start with role, then go on)

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1181,29 +1183,59 @@ private static AmazonS3 getClient(String driverId) {
...
+            }
+            if (s3profile != null) {
+                ProfileCredentialsProvider profileCredentials = new ProfileCredentialsProvider(s3profile);
+                providers.add(profileCredentials);
+            }
+
+            if (allowInstanceCredentials) {
+                // Add role-based provider as in the default provider chain
+                InstanceProfileCredentialsProvider instanceCredentials = InstanceProfileCredentialsProvider.getInstance();
+                providers.add(instanceCredentials);
```
poikilotherm, line 1227:
```suggestion
                providers.add(InstanceProfileCredentialsProvider.getInstance());
```

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1181,29 +1183,59 @@ private static AmazonS3 getClient(String driverId) {
...
+                                accessKey.orElse(""),
+                                secretKey.orElse("")));
+                providers.add(staticCredentials);
+            } else if (s3profile == null) {
+                //Only use the default profile when it isn't explicitly set for this store when there are no static creds (otherwise it will be preferred).
+                s3profile = "default";
+            }
+            if (s3profile != null) {
+                ProfileCredentialsProvider profileCredentials = new ProfileCredentialsProvider(s3profile);
+                providers.add(profileCredentials);
```
poikilotherm, line 1221:
```suggestion
                providers.add(new ProfileCredentialsProvider(s3profile));
```

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
qqmyers: Since we have to know if the profile and/or static creds exist before we could add the instance creds at the start of the array, I thought it was cleaner to just reverse than to split the code where the cred providers are looked up from where they are actually created/added to the array.

pdurbin:
It would be nice to review and merge our new S3 tests...

- #10044

... before reviewing, testing, and merging this PR.

pdurbin (approved):
I haven't tested this but I don't see anything obviously wrong with it. Approved.

jp-tosca:
> It would be nice to review and merge our new S3 tests...
> 
> * [add S3 tests, LocalStack, MinIO #10044](https://github.com/IQSS/dataverse/pull/10044)
> 
> ... before reviewing, testing, and merging this PR.

Do you still think this is the case? @pdurbin

pdurbin:
It's not a hard requirement or anything.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -108,14 +111,13 @@ public S3AccessIO(T dvObject, DataAccessRequest req, String driverId) {
...
-            } catch (SdkClientException sce) {
-                throw new IOException("ERROR: S3AccessIO - Failed to look up bucket "+bucketName+" (is AWS properly configured?): " + sce.getMessage());
-            }
+
+            // FWIW: There used to be a check here to see if the bucket exists.
+            // It was very redundant (checking every time we access any file) and didn't do
+            // much but potentially make the failure (in the unlikely case a bucket doesn't
+            // exist/just disappeared) happen slightly earlier (here versus at the first
+            // file/metadata access).
+
```
landreev, line 120: Good call.

landreev:
Looking good.

PR #10016, opened 2023-10-16
RO-Crate metadata mime type detection

### LINKED ISSUE #10015: better detection of RO-Crate files
Currently, we can specify mime types for specific files recognized by name in the ```MimeTypeDetectionByFileName.properties```
However, this only works for files without extensions. Some files that have a specific name require an extension, e.g., `ro-crate-metadata.json`, which is the file containing the RO-Crate metadata. Allowing adding extensions in the ```MimeTypeDetectionByFileName.properties``` would then make it possible to recognize these files and give them the correct mime type (e.g., ```application/ld+json; profile="http://www.w3.org/ns/json-ld#flattened http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate"``` for the RO-Crate metadata).

Depositor, Superuser, Sysadmin

RO-Crate previewer (https://github.com/gdcc/dataverse-previewers/pull/41) relies on correct recognition of the RO-Crate metadata files associated with that previewer.

```MimeTypeDetectionByFileName.properties``` would also be used for recognition of files having extensions.

### PR DESCRIPTION
**Pull request by ErykKul**
This PR adds support for RO-Crate metadata detection. It does so by extending `MimeTypeDetectionByFileName.properties` to accept filenames with extensions and adding the two possible RO-Crate metadata file names:
```
ro-crate-metadata.json=application/ld+json; profile="http://www.w3.org/ns/json-ld#flattened http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate"
ro-crate-metadata.jsonld=application/ld+json; profile="http://www.w3.org/ns/json-ld#flattened http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate"
```

Closes #10015 

Add a file named `ro-crate-metadata.json` to a dataset. The displayed mime type should be `application/ld+json` instead of `application/json` (the default for `JSON` files):
image

When you add the RO-Crate previewer (https://github.com/gdcc/dataverse-previewers/pull/41) to the external tools, it should then preview the metadata file as RO-Crate.

Yes.

### CODE REVIEW

**Comment by pdurbin** [comment]
> The displayed mime type should be `application/ld+json` instead of `application/json` (the default for `JSON` files):

This is a nice improvement! Wouldn't the next logical step be to edit the following files...

- src/main/java/propertyFiles/MimeTypeDisplay.properties
- src/main/java/propertyFiles/MimeTypeFacets.properties

... so that we can show `RO-Crate` rather than `application/ld+json` to the user? Here's a typical example of when we edit these files:

- #9366

Also, some tests would be nice. And a release note. 😄

**Review by qqmyers — APPROVED** [review]
Looks good.

**Comment by ErykKul** [comment]
@pdurbin 
Can you check my release note? Is it fine like this?
I added the test, I also needed to tweak the code a little bit to make it work with the profile in the mime type (RO-Crate is `ld+json`, but not all `ld+json` are RO-Crate).
It was not easy to get that cumbersome mime type as key in the property files, but Unicode escaping worked fine. It only looks strange, the test confirms that it works.
I have set the MimeTypeDisplay property of the RO-Crate as `RO-Crate metadata` and the MimeTypeFactes as `Metadata`. If you have a suggestion to change them, I will be happy to do it. Especially the facet is tricky, maybe already existing facet like `Data` or `Document` would be better here?
Thanks!

**Comment by pdurbin** [comment]
> I have set the MimeTypeDisplay property of the RO-Crate as `RO-Crate metadata`

Sounds good. In line with https://www.researchobject.org/ro-crate/1.1/terminology.html

> and the MimeTypeFactes as `Metadata`.

No objection from me. It's the first of these but it doesn't seem to fit with anything more popular:

```
  55 Data
  47 Code
  26 Image
  21 Archive
  15 Text
  13 Document
  10 Shape
   9 Audio
   8 Video
   2 FITS
   1 Unknown
   1 Tabular Data
   1 Network Data
   1 Metadata
```

@jggautier do you have an opinion on this?

Also, @ErykKul the release notes look good. I'm glad tests are passing. I haven't really looked at the code yet but I'm giving this a small effort size for review and QA.

**Comment by jggautier** [comment]
Thanks for asking. I don't have an informed opinion.

File: src/test/java/edu/harvard/iq/dataverse/util/FileUtilTest.java [hunk 1]
```diff
@@ -371,4 +371,33 @@ public void testHdf4File() throws IOException {
...
+    public void testDetermineFileTypeROCrate() {
+        final String roCrateContentType = "application/ld+json; profile=\"http://www.w3.org/ns/json-ld#flattened http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate\"";
+        final DataFile rocrate = new DataFile(roCrateContentType);
+        
+        assertEquals(roCrateContentType, rocrate.getContentType());
+        assertEquals("RO-Crate metadata", FileUtil.getUserFriendlyFileType(rocrate));
+        assertEquals("Metadata", FileUtil.getIndexableFacetFileType(rocrate));
+
+        final File roCrateFile = new File("src/test/resources/fileutil/ro-crate-metadata.json");
+        if (roCrateFile.exists()) {
```
qqmyers, line 384: Do we need to check for existence and have a sample file for this test? It should return just from seeing the name.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -545,6 +562,11 @@ public static String determineFileType(File f, String fileName) throws IOExcepti
     }
 
     public static String determineFileTypeByNameAndExtension(String fileName) {
+        final ResourceBundle bundle = BundleUtil.getResourceBundle("MimeTypeDetectionByFileName");
```
qqmyers, line 565: It looks like this obsoletes some of lookupFileTypeFromPropertiesFile - that ends up trying to check the same file again after checking based on extension. Can that logic be removed/otherwise refactored to just prioritize the check by full name?

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -421,6 +433,11 @@ public static String retestIngestableFileType(File file, String fileType) {
     }
     
     public static String determineFileType(File f, String fileName) throws IOException{
+        final ResourceBundle bundle = BundleUtil.getResourceBundle("MimeTypeDetectionByFileName");
+        if (bundle.keySet().contains(fileName)) {
+            return bundle.getString(fileName);
+        }
+
```
qqmyers, line 440: Same comment as below - this addition now makes the call on line 500 partially redundant unless that call drops checking for whole name again.

qqmyers (changes requested):
Looks good. I left comments about dropping some now duplicate code and getting rid of the example file (which I think is not needed). Otherwise, it looks ready to go. I see that the build failed way back when but the results are no longer available, so we should check when this rebuilds after changes.

cmbz:
Related to https://github.com/IQSS/dataverse-pm/issues/175. Added related GREI tag.

src/test/java/edu/harvard/iq/dataverse/util/FileUtilTest.java
ErykKul: I removed the file and adjusted the test, it works. Thanks!

ErykKul:
@qqmyers, I have refactored the lookup logic. It looks better now and is slightly more readable.

ErykKul:
@qqmyers I am not sure why the test was failing, but if it is not related to this PR, it can go to "ready for QA"

PR #10017, opened 2023-10-16
Fix Datasets API getVersionFiles endpoint content type filtering

GPortas:

DatasetsIT.getVersionFiles fails intermittently as it sometimes finds out unordered files when sorting by content type.

A second orderBy condition has been added to force that when sorting by content type, files are also sorted by name.

New order example:

- test_5.png
- test_1.txt
- test_2.txt
- test_3.txt

Get dataset files via curl using contentType query param:

`curl "https://demo.dataverse.org/api/datasets/24/versions/1.0/files?contentType=image/png"`


landreev (approved):
Perfect, thank you for a quick fix!

PR #10020, opened 2023-10-17
Lookup dataverse by alias or ID when downloading guestbook responses via API

### LINKED ISSUE #9619: input database ID for Retrieve Guestbook Responses for a Dataverse Collection
The API call “[Retrieve Guestbook Responses for a Dataverse Collection](https://guides.dataverse.org/en/5.12.1/api/native-api.html#id47)” doesn’t work with the Database Id. With the alias it works.
With alias it works:
curl -H X-Dataverse-key:xxx https://data.aussda.at/api/dataverses/autnes/guestbookResponses 
With Id it doesn’t work
curl -H X-Dataverse-key:xxx https://data.aussda.at/api/dataverses/84/guestbookResponses 

* What happens?
Possible reason, Jim Myers: “I see the same thing – an empty response when using an id. My guess is that this was broken about two years ago when this line was added: https://github.com/IQSS/dataverse/blob/bef00db2fdf34c5460286cd22e7ac82a43f228fa/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java#L1129”
https://groups.google.com/g/dataverse-community/c/O9jF5o1FxTw/m/YVR9ILLrAAAJ?utm_medium=email&utm_source=footer

API users

v. 5.12.1

bencomp:
Replaces a dataverse by-alias lookup with `findDataverseOrDie` to allow looking up a dataverse by both its alias and its ID, as the [documentation for 5.12.1](https://guides.dataverse.org/en/5.12.1/api/native-api.html#retrieve-guestbook-responses-for-a-dataverse-collection) mentions. The [same section for the latest version](https://guides.dataverse.org/en/latest/api/native-api.html#retrieve-guestbook-responses-for-a-dataverse-collection) says the same.

- Closes #9619 

**Special notes for your reviewer**: I would like this PR to count towards my Hacktoberfest goal, so if this is useful, please add the hacktoberfest-accepted label or an approving review.
My editor VSCodium with Java extensions made me handle the `WrappedResponse` that `findDataverseOrDie` may throw. Throwing a `WebApplicationException` is appropriate for the context, according to its Javadoc description.

**Suggestions on how to test this**: check that the commands in the issue work after applying the change.

I suppose this may fix a bug in the API, but I don't consider this a big change otherwise.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
@@ -1194,10 +1194,15 @@ public class Dataverses extends AbstractApiBean {
             public void write(OutputStream os) throws IOException,
                     WebApplicationException {
 
-                Dataverse dv = dataverseService.findByAlias(dvIdtf);
+                Dataverse dv;
+                try {
+                    dv = findDataverseOrDie(dvIdtf);
+                } catch (WrappedResponse wr) {
+                    throw new WebApplicationException(wr.getResponse());
+                }
                 Map<Integer, Object> customQandAs = guestbookResponseService.mapCustomQuestionAnswersAsStrings(dv.getId(), gbId);
                 Map<Integer, String> datasetTitles = guestbookResponseService.mapDatasetTitles(dv.getId());
-                
+
                 List<Object[]> guestbookResults = guestbookResponseService.getGuestbookResults(dv.getId(), gbId);
                 os.write("Guestbook, Dataset, Dataset PID, Date, Type, File Name, File Id, File PID, User Name, Email, Institution, Position, Custom Questions\n".getBytes());
                 for (Object[] result : guestbookResults) {
--- a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
@@ -1173,8 +1173,9 @@ public class Dataverses extends AbstractApiBean {
     public Response getGuestbookResponsesByDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf,
             @QueryParam("guestbookId") Long gbId, @Context HttpServletResponse response) {
 
+        Dataverse dv;
         try {
-            Dataverse dv = findDataverseOrDie(dvIdtf);
+            dv = findDataverseOrDie(dvIdtf);
             User u = getRequestUser(crc);
             DataverseRequest req = createDataverseRequest(u);
             if (permissionSvc.request(req)
@@ -1194,12 +1195,6 @@ public class Dataverses extends AbstractApiBean {
             public void write(OutputStream os) throws IOException,
                     WebApplicationException {
 
-                Dataverse dv;
-                try {
-                    dv = findDataverseOrDie(dvIdtf);
-                } catch (WrappedResponse wr) {
-                    throw new WebApplicationException(wr.getResponse());
-                }
                 Map<Integer, Object> customQandAs = guestbookResponseService.mapCustomQuestionAnswersAsStrings(dv.getId(), gbId);
                 Map<Integer, String> datasetTitles = guestbookResponseService.mapDatasetTitles(dv.getId());
 
@@ -1208,7 +1203,6 @@ public class Dataverses extends AbstractApiBean {
                 for (Object[] result : guestbookResults) {
                     StringBuilder sb = guestbookResponseService.convertGuestbookResponsesToCSV(customQandAs, datasetTitles, result);
                     os.write(sb.toString().getBytes());
-
                 }
             }
         };
--- a/src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
@@ -375,7 +375,7 @@ public class UtilIT {
 
     static Response getGuestbookResponses(String dataverseAlias, Long guestbookId, String apiToken) {
         RequestSpecification requestSpec = given()
-                .auth().basic(apiToken, EMPTY_STRING);
+                .header(API_TOKEN_HTTP_HEADER, apiToken);
         if (guestbookId != null) {
             requestSpec.queryParam("guestbookId", guestbookId);
         }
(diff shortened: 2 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1194,10 +1194,15 @@ public Response getGuestbookResponsesByDataverse(@Context ContainerRequestContex
             public void write(OutputStream os) throws IOException,
                     WebApplicationException {
 
-                Dataverse dv = dataverseService.findByAlias(dvIdtf);
+                Dataverse dv;
+                try {
+                    dv = findDataverseOrDie(dvIdtf);
+                } catch (WrappedResponse wr) {
+                    throw new WebApplicationException(wr.getResponse());
```
pdurbin, line 1201:
```suggestion
                    return wr.getResponse());
```

Untested. Can you please give this a try? It's our normal pattern.

pdurbin (commented):
@bencomp thanks for the PR! 🎉 I left a comment about WrappedResponse.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
bencomp: Indeed, I have seen this pattern before (e.g. on line 1188). But void methods cannot return a value, the IDE helpfully says 😄 

My gut tells me this lookup should never fail, because the same lookup is performed a few lines up. Unfortunately I couldn't find a way to reference the Dataverse object from within this `write` method to circumvent the whole lookup with try-catch pattern alltogether.

pdurbin: Ah, now that I'm in front of my IDE I see what you mean. 😄 

I have the same gut feeling but how to do you feel about writing some API tests? I'd suggest adding a method for downloading guestbook responses to UtilIT.java (I couldn't find one) and then calling it from DataversesIT.java.

pdurbin: > couldn't find a way to reference the Dataverse object

My IDE doesn't complain about this:

```
diff --git a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
index 30c1453525..3fd17c517b 100644
--- a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
@@ -1173,8 +1173,9 @@ public class Dataverses extends AbstractApiBean {
     public Response getGuestbookResponsesByDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf,
             @QueryParam("guestbookId") Long gbId, @Context HttpServletResponse response) {
 
+        Dataverse dv;
         try {
-            Dataverse dv = findDataverseOrDie(dvIdtf);
+            dv = findDataverseOrDie(dvIdtf);
             User u = getRequestUser(crc);
             DataverseRequest req = createDataverseRequest(u);
             if (permissionSvc.request(req)
@@ -1194,7 +1195,6 @@ public class Dataverses extends AbstractApiBean {
             public void write(OutputStream os) throws IOException,
                     WebApplicationException {
 
-                Dataverse dv = dataverseService.findByAlias(dvIdtf);
                 Map<Integer, Object> customQandAs = guestbookResponseService.mapCustomQuestionAnswersAsStrings(dv.getId(), gbId);
                 Map<Integer, String> datasetTitles = guestbookResponseService.mapDatasetTitles(dv.getId());
```

bencomp: > My IDE doesn't complain about this

Mine doesn't either! Nice!

bencomp: > how to do you feel about writing some API tests?

Writing API tests is a bit too much at this moment, sorry. Maybe later this month.

pdurbin (commented):
Do you feel like writing some tests? If so, I'd suggest adding a method for downloading guestbook responses to UtilIT.java (I couldn't find one) and then calling it from DataversesIT.java.

bencomp:
Working on it.

bencomp:
I couldn't test locally – maybe because I have en_GB as locale – and the Maven integration test job doesn't run the new test either. Does that require something else to happen, @pdurbin?

pdurbin:
@bencomp thanks for adding tests! To run tests locally, I recommend running Dataverse in Docker using https://guides.dataverse.org/en/6.0/developers/dev-environment.html#quickstart and then running just the tests in that class with `mvn test -Dtest=DataversesIT`. You could get even more granular with `mvn test -Dtest=DataversesIT#testGetGuestbookResponses`.

Jenkins is failing but not due to your code. I just kicked off a manual run at https://github.com/gdcc/api-test-runner/actions/runs/6554508063 to get a second opinion.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -372,7 +372,16 @@ static Response showDataverseContents(String alias, String apiToken) {
                 .header(API_TOKEN_HTTP_HEADER, apiToken)
                 .when().get("/api/dataverses/" + alias + "/contents");
     }
-    
+
+    static Response getGuestbookResponses(String dataverseAlias, Long guestbookId, String apiToken) {
+        RequestSpecification requestSpec = given()
+                .auth().basic(apiToken, EMPTY_STRING);
```
pdurbin, line 378:
```suggestion
                .header(API_TOKEN_HTTP_HEADER, apiToken);
```

pdurbin (commented):
This should help resolve the 403.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
bencomp: Seriously?! I've been stuck on this for a long time.

pdurbin:
I just looked over https://guides.dataverse.org/en/6.0/api/native-api.html#retrieve-guestbook-responses-for-a-dataverse-collection and it seems fine. It already claims to support a collection alias, which is what @bencomp is enabling in this PR.

bencomp:
Well, now it *works on my machine* 🎉 
I would say this is done 😌

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
bencomp: I guess I should not have started from a SWORD call 😞

pdurbin:
Ok, it passed. @scolapasta can we put this on https://github.com/orgs/IQSS/projects/2 ?

PR #10029, opened 2023-10-20
disable s3 tagging JVM option

### LINKED ISSUE #10022: Support for s3 without tagging (e.g., s3 api on top of swift)
When using an s3 implementation that does not support tagging, e.g., [s3 api](https://github.com/openstack/swift/tree/master/swift/common/middleware/s3api) of the [openstack swift](https://github.com/openstack/swift), you cannot use the upload-redirect feature. The uploads are prevented with "501: Not implemented" error.

An error is thrown when uploading a file with upload-redirect feature turned on. This happens only on a Dataverse instance using an s3 implementation that does not support tagging. I have tested it with s3 on top of the openstack swift.

File upload

**What happens?**
 "501: Not implemented" error.

All users

I expect the upload to succeed.

Latest.

#9130
Since the solution to this problem is removing the tag header (I am working on a pull request for this issue that would add an option to disable tagging for uploads), a cleanup mechanism is needed to removed partially uploaded files, etc., since the `temp` tag would not be there anymore. That mechanism is already implemented in #9132 as a solution to the linked issue.

**Screenshots:**

image

ErykKul:
Some S3 implementations do not support tagging. This might prevent the upload-redirect from working, with an error on the `x-amz-tagging` header. This PR adds an JVM option that allows disabling the tagging.

Closes #10022

** Testing - do a direct upload and check in the browser to see that the generated URLs do not have the tag when the flag is true. Direct uploads should work on a store that supports tags regardless of the state of the flag. Unless we have a known S3 store that doesn't support tags available for QA, I think we can skip testing that case (presumably @ErykKul already has in developing this.)
 
Yes.

**Additional documentation**:
The new option is listed in the JVM settings for s3 store section.


qqmyers (approved):
Looks good. I made minor edits to the notes/guides text and added a //STORAGE DRIVER header in the JvmSettings (since this is the first of presumably many such settings - they just haven't been converted yet.).
It's likely that this will be an s3 specific setting which may mean we'll want a dataverse.files.{id}.s3.disable-tagging setting (extra s3 scope) at some point, but I think it is overkill to change it at this point since we haven't done the other storage driver settings yet.

pdurbin:
@ErykKul I haven't tested this yet but I worked on the docs a bit. I hope you approved. Among other things, I edited some existing docs and the paragraph a section header: https://dataverse-guide--10029.org.readthedocs.build/en/10029/developers/big-data-support.html#s3-tags

ErykKul:
@pdurbin It looks great! Thanks!

pdurbin:
@ErykKul @qqmyers in e473d53 I pushed a doc change because I saw this error when I disabled tagging:

```
<Error>
  <Code>AccessDenied</Code>
  <Message>There were headers present in the request which were not signed</Message>
  <RequestId>25ff2bb0-13c7-420e-8ae6-3d92677e4bd9</RequestId>
  <HostId>9Gjjt1m+cjU4OPvX9O9/8RuvnG41MRb/18Oux2o5H5MY7ISNTlXN+Dz9IG62/ILVxhAGI0qyPfg=</HostId>
  <HeadersNotSigned>x-amz-tagging</HeadersNotSigned>
</Error>
```

I assume this was the right thing to do. Otherwise, please feel free to revert or further edit.

That is to say, of course, that I can at least tell that something is happening when I disable tagging.

I must say that I don't know how to do this: "do a direct upload and check in the browser to see that the generated URLs do not have the tag when the flag is true". If someone can help, I'm happy to try. Otherwise, I'm ready to merge as I don't think this pull request does any harm.

Something I am able to check is the URL generated from the `mvn test -Dtest=S3AccessIT#testDirectUpload`. The "before" URL with tagging enabled (JVM option unset) has "x-amz-tagging" in it:

`url: http://localstack:4566/mybucket/10.5072/FK2/8HBFB9/18eed42827c-73a9c09a1ac1?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20240417T181223Z&X-Amz-SignedHeaders=host%3Bx-amz-tagging&X-Amz-Expires=3599&X-Amz-Credential=default%2F20240417%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Signature=8e31e272ba6a6f889858b693cc41319bc618e2df3d264ee743b752f974f318ca`

The "after" URL (tagging disabled), does not:

`url: http://localstack:4566/mybucket/10.5072/FK2/9H00VE/18eed4f74e4-6f2d4788a4c6?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20240417T182631Z&X-Amz-SignedHeaders=host&X-Amz-Expires=3599&X-Amz-Credential=default%2F20240417%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Signature=9dbd1a640b70c3d6970a49ea92eded04e6f57a4bf02947aae81ec1350996c8be`

So, I've tested what I understand how to test. I don't actually see the "temp" tag anywhere. If you can explain how, I'm happy to keep digging. Otherwise, like I said, I'm ready to merge. Please let me know.

p.s. This is more than a 3. My bad.

qqmyers:
What you saw is basically it - Dataverse's job is either to allow that tag to go through or not (so it's not the tag itself in the URL but having that header allowed as part of the call). The temp tag itself is on the object in S3 so you could use the aws cli to go look for it, but if you can't send the tag, it can't be there. Your doc update to not include the header when the flag is off is a good catch.

pdurbin:
@qqmyers thanks, that's good enough for me. Merging!

ErykKul:
@pdurbin, thanks for catching that and merging!

PR #10050, opened 2023-10-26
9635 solr improvements

### LINKED ISSUE #9635: Investigate solr performance, diagnose bottlenecks and develop optimizations
This primarily concerns busy instances with large amounts of indexed metadata. Such as the IQSS prod. installation. There's some evidence of Solr becoming the performance bottleneck during high service load. Facet queries are especially suspect. But this has the potential to benefit any busy instance, hence the issue in the main repo. 

@kcondon got the ball rolling on this and has already found and applied some useful  info. He has started a google doc documenting his work. We may link it here and we will try to document the most useful things we find here (and in the guides, eventually). Kevin's out today, but we had a conversation and decided I'll go ahead an open the issue to make sure this work is scheduled and tracked; striking it while it's hot, etc.

landreev:

The following application-side improvements/workarounds are being added as part of this PR:

- Adding app-side support for the Solr-side circuit breakers. Enabling those makes Solr start dropping requests with 503s (instead of dying) under heavy load. As currently implemented however, the page does not detect that condition and doesn't handle it in any special way, and simply shows zero search results/empty collection page. It needs to show something intelligent, along the lines of "search engine is experiencing heavy load, please try again later". 
- Optimizing the second search that the page repeats solely for the purpose of populating the type counts facets for the _unchecked_ types. There is absolutely no need to repeat the **same exact search**, with all its super expensive parts, like all the other facets. There's no need to search for anything other than these one or two total counts; or to even run this second search, if all 3 type facets happen to be checked. 
- Adding `nofollow` to all the facet links and such, that we don't want crawlers/bots to touch. This will obviously only help with the nicer bots that take hints and follow rules, but better than nothing. 
- An off switch for facets support on the dataverse page. Again, facets are super expensive; it should be far better to be able to display the page with just the search cards and "facets are temporarily unavailable because of load issues" in the left column, than to have an empty page/have a "solr down" error page/have the page load take an hour. 

Closes #9635

**Additional documentation**:


landreev:
I just added a "closes" tag for #9635 to the PR. Once the pr is merged, I want to hot-patch our prod. with the improvements here, and then close the issue, at least for now. As I said before, there's definitely more to research and investigate and I want to get back to it, but there may be more pressing things r/n.

landreev:
OK, one test is legitimately failing in this PR, will fix.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -822,6 +822,12 @@ public List<FacetLabel> getFileTagsFacetLabels() {
      * @return boolean
      */
     public boolean isThisVersionSearchable() {
+        // Just like on the collection page, facets on the Dataset page can be
+        // disabled instance-wide by an admin:
+        if (settingsWrapper.isTrueForKey(SettingsServiceBean.Key.DisableSolrFacets, false)) {
+            return false;
+        }
+
```
qqmyers, line 830: re: the following code - is it worth it to query solr just to see if you can query solr? Would a check of the last indexed time be a reasonable alternative? (Not sure how much this would save given it's a simple query).

landreev: Yeah, this looks like it's yet another thing we can save/economize on. 
We are already checking that this is the latest released version...  so, yes, comparing the version published date against the indexed time stamp should be reliable enough. 
These queries that just search for the version id cannot be that expensive; but there's definitely a lot of them in the logs.

src/main/java/edu/harvard/iq/dataverse/GuestbookPage.java
```diff
@@ -288,19 +288,22 @@ public String save() {
...
+            // reindexing of the underlying datasets). - L.A.
             if (editMode == EditMode.CREATE || editMode == EditMode.CLONE ) {
                 guestbook.setCreateTime(new Timestamp(new Date().getTime()));
-                guestbook.setUsageCount(new Long(0));
+                guestbook.setUsageCount(Long.valueOf(0));
                 guestbook.setEnabled(true);
                 dataverse.getGuestbooks().add(guestbook);
-                cmd = new UpdateDataverseCommand(dataverse, null, null, dvRequestService.getDataverseRequest(), null);                
-                commandEngine.submit(cmd);
+                logger.info("Returned from command");
```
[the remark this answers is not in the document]
landreev: I actually noticed it, and meant to drop it. 😄

qqmyers (approved):
Looks good. I haven't tested everything, but I have cherry-picked the nofollow links to QDR (and have the changes to the Guestbook page running there as well. The rest of the changes make sense.

landreev:
> Looks good. I haven't tested everything, but I have cherry-picked the nofollow links to QDR (and have the changes to the Guestbook page running there as well. The rest of the changes make sense.

Cool, thank you. I will refactor the "version searchable" test above, and drop the .info log line quickly.

sekmiller:
Seeing some weird results on the counts for dvobject facets in child Dataverses when those types are unselected.
Screen Shot 2023-11-15 at 10 42 07 AM
Screen Shot 2023-11-15 at 10 41 32 AM

sekmiller:
You can see the above on Dataverse internal at https://dataverse-internal.iq.harvard.edu/dataverse/sek11-15

(I added a couple of objects while facets were unavailable and it worked fine)

landreev:
@sekmiller 
Just to confirm that yes, this is clearly a bug. The counts shown for the unchecked types are the totals for the *root* collection, not the current sub-collection. 
... But I'm still working on a fix.

landreev:
@sekmiller 
Sorry for the delay. The fix was quite simple actually.

pdurbin:
@sekmiller and I talked and I just kicked this off for a second opinion on API tests: https://github.com/gdcc/api-test-runner/actions/runs/6935079103

landreev:
I still need to fix the API test above.

landreev:
The tests are passing now.

PR #10058, opened 2023-10-27
IQSS/10057 ddi export bugs

### LINKED ISSUE #10057: DDI Export bugs introduced in 5.14
The change allow external exporters in 5.14 introduced some bugs in translating variable level metadata to an intermediate JSON 'transfer object' form and then using that JSON to create the DDI XML. So far these have primarily related to optional fields where either the JSON creation or DDI code fails with a null pointer. 

Two in particular:
QDR has some cases where the CategoryVariable frequency is null and Odum has files where the DataVariable label is null.

Either of these causes an export failure for the DDI export (and hence the HTML view of DDI as well that uses it).

PR to follow.

qqmyers:
**What this PR does / why we need it**: Fixes two bugs causing DDI metadata export to fail in some cases.

Closes #10057

**Special notes for your reviewer**: FWIW - there has been some slack discussion about whether this is worth a point release and/or if an external DDI exporter with the fixes could be built for 5.14/6.0. So far, these appear rare but Harvard hasn't analyzed its holding's export failures yet afaik.

**Suggestions on how to test this**:  The SPSS binary file in https://dataverse.theacss.org/dataset.xhtml?persistentId=doi:10.25825/FK2/4LAWJN has one bug and https://data.qdr.syr.edu/file.xhtml?persistentId=doi:10.5064/F6MW2F2K/V3SPLV&version=1.0 has the other. Verifying that you can get a DDI export from a dataset published with these files in it would prove the fixes work.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
@@ -1824,7 +1824,7 @@ public class DdiExportUtil {
                 xmlw.writeCharacters(dvar.getString("label"));
                 xmlw.writeEndElement(); //labl
             }
-        } else if (vm != null && vm.containsKey("label")) {
+        } else {
             xmlw.writeStartElement("labl");
             writeAttribute(xmlw, "level", "variable");
             xmlw.writeCharacters(vm.getString("label"));
--- a/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
@@ -1818,10 +1818,12 @@ public class DdiExportUtil {
 
         // labl
         if ((vm == null || !vm.containsKey("label"))) {
-            xmlw.writeStartElement("labl");
-            writeAttribute(xmlw, "level", "variable");
-            xmlw.writeCharacters(dvar.getString("label"));
-            xmlw.writeEndElement(); //labl
+            if(dvar.containsKey("label")) {
+                xmlw.writeStartElement("labl");
+                writeAttribute(xmlw, "level", "variable");
+                xmlw.writeCharacters(dvar.getString("label"));
+                xmlw.writeEndElement(); //labl
+            }
         } else if (vm != null && vm.containsKey("label")) {
             xmlw.writeStartElement("labl");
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
```diff
@@ -1818,10 +1818,12 @@ private static void createVarDDI(XMLStreamWriter xmlw, JsonObject dvar, String f
 
         // labl
         if ((vm == null || !vm.containsKey("label"))) {
-            xmlw.writeStartElement("labl");
-            writeAttribute(xmlw, "level", "variable");
-            xmlw.writeCharacters(dvar.getString("label"));
-            xmlw.writeEndElement(); //labl
+            if(dvar.containsKey("label")) {
```
bencomp, line 1821: Could you combine this if with the enclosing if?

bencomp (commented):
This is more a question than a comment... :nerd_face:

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
qqmyers: Decided to remove the redundant 'if' on 1825 below instead.

bencomp: If I'm not mistaken, you could still combine this `if`'s condition with the `(vm == null || !vm.containsKey("label"))` and have the same logic.

```
if ((vm == null || !vm.containsKey("label")) && dvar.containsKey("label")) {
```
(and remove the `}` on line 1826)

qqmyers: if (vm == null || !vm.containsKey("label")) is true and dvar.containsKey("label") is false, the else would be triggered, causing vm.getString("label") to be called (line 1830) when it doesn't contain that key.

bencomp: Of course! 🤦

landreev (approved):
The IQSS prod. instance does have a low 3-digit number of nightly export failures. So yes, I would like to patch our servers too. 
Thank you for the PR, Jim!

landreev:
I'll volunteer to QA and merge this PR.

PR #10096, opened 2023-11-03
IQSS/10095-download estimate bug fix

### LINKED ISSUE #10095: Bug in download estimate function changed for Guestbook-at-request functionality
Seen at QDR: An error in the estimateGuestBookResponseTableSize function added to postgres was introduced in #9599. The new code to estimate what fraction of guestbook table entries relate to downloads and not to access requests incorrectly give the type bigint to a number that is a fraction. See https://github.com/IQSS/dataverse/blob/6cfe43a0a6aae9df55679ff320d774e567269243/src/main/resources/db/migration/V6.0.0.1__9599-guestbook-at-request.sql#L51. Changing the type to float resolves the issue.

* When does this issue occur?
When the number of guestbook entries is large enough and guestbook-at-request functionality has been enabled.

Homepage showing the download count

* What happens?
Causes a 500 error

develop

PR to follow

qqmyers:
**What this PR does / why we need it**: As described in the issue, the prior method tried to assign a fraction to a bigint, causing  a failure. This fixes the issue by using the type float

Closes #10095

**Special notes for your reviewer**: The bug is only in the dev branch. The PR adds a new flyway script rather than replacing the old one. Just changing the old one would work except for developers having to remove the last couple of flyway entries. I can change the PR to that if it is desirable (avoids recording the bug in every flyway table forever.)

**Suggestions on how to test this**: This only appears once the table size is large enough (relpages >10), guestbook-at-request has been enabled somewhere and a response for an at-request guestbook has been entered, and table stats are up to date. I'm not sure what the best way to recreate that is although I expect using a copy of the real Harvard DB and setting up a guestbook-at-request and using it would work. Nominally all of the select statements in the query could be performed manually to verify that the test case is hitting the changed line. Happy to discuss. I can confirm that the new query is working at QDR (and that the old one was failing/same conditions).

**Additional documentation**:

### DIFF
--- a/src/main/resources/db/migration/V6.0.0.3__10095-guestbook-at-request2.sql
+++ b/src/main/resources/db/migration/V6.0.0.3__10095-guestbook-at-request2.sql
@@ -1,0 +1,34 @@
+-- This creates a function that ESTIMATES the size of the
+-- GuestbookResponse table (for the metrics display), instead
+-- of relying on straight "SELECT COUNT(*) ..."
+-- It uses statistics to estimate the number of guestbook entries
+-- and the fraction of them related to downloads,
+-- i.e. those that weren't created for 'AccessRequest' events.
+-- Significant potential savings for an active installation.
+-- See https://github.com/IQSS/dataverse/issues/8840 and 
+-- https://github.com/IQSS/dataverse/pull/8972 for more details
+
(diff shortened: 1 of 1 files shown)


sekmiller (approved):
Looks good. And I'm ok with having multiple scripts as long as the final one works, which it does.

Issue #9898, opened 2023-09-10
System creates unwanted API token


philippconzett:
During the past two years or so I've experienced several times that an API token was created for my user (superuser) without me clicking on the Create Token button.

To illustrate the latest occurrence of this behavior: This morning I created an API token which I needed to run a regular API script. After I ran the script, I revoked the token. I even clicked on other buttons on the page, refreshed the page, and navigated back to check whether no token was defined anymore. The token I used in the script started with "7d2a". When I checked this evening, a new token starting with "2c4d" was displayed.

**What steps does it take to reproduce the issue?**
1. Click on your username on the top right of the Dataverse page.
2. Click on API token.
3. Create a token and run an API call.
4. Revoke the token.
5. After a while, return to the API token page.

* When does this issue occur?
Sporadically, as described above.

* Which page(s) does it occurs on?
The API Token page.

* What happens?
See description above.

* To whom does it occur (all users, curators, superusers)?
I've only tested this as a superuser.

* What did you expect to happen?
The token being permanently revoked and no new token being created until I create a new one.

**Which version of Dataverse are you using?**
5.13

**Any related open or closed issues to this bug report?**
No.

**Screenshots:**
No.

qqmyers:
FWIW: 
Right now, tokens get created when you use something like a previewer that would require one to exist. As we switch to using signedUrls for external tools, it would be easier to change this behavior. If we changed it today, previewers/explore/config tools for files in draft versions and restricted files would fail unless/until you create an API key manually.

philippconzett:
Ah, thanks. I guess this explains the behavior described above. Should we close the issue?

jggautier:
Since it's not easier to change this behavior now, would it be possible to let users know that this might happen? For example, if someone revokes thier api token, they're told that it the repository will recreate the token when they use certain external tools?

Otherwise the user doesn't know it's misleading. And maybe a security issue?

pdurbin:
I do think we should somehow inform users that an API token is being created for them. Here's how we talk about these tokens in the User Guide (i.e. treat them with care!):

[How Your API Token Is Like a Password](https://guides.dataverse.org/en/6.0/user/account.html#id23)

In many cases, such as when depositing data, an API Token is required to interact with Dataverse Software APIs. The word “token” indicates a series of letters and numbers such as c6527048-5bdc-48b0-a1d5-ed1b62c8113b. Anyone who has your API Token can add and delete data as you so you should treat it with the same care as a password.

[...]

Issue #9405, opened 2023-02-23
Signed URLs: If target user account does not yet have an API Token, requestURL fails, even though request made with valid token.


kcondon:
When requesting a signed url as a super user but for permissions for a new non super user, it fails with "Do not have a valid user with apiToken". In this case it is referring to the the target user who has not yet created an api token for their account.

Discussed with Jim and he thought the request should create a token if there was not yet one and so a bug.

Relevant command lines:

JSON='{"url":"https://dataverse-internal.iq.harvard.edu/api/v1/datasets/:persistentId/?persistentId=doi:10.70122/FK2/AH9UIE","timeOut":35,"user":"kc022223"}' 

curl -k -H "X-Dataverse-key:$API_KEY" -H 'Content-Type:application/json' -d "$JSON" http://localhost:8080/api/admin/requestSignedUrl

{"status":"ERROR","message":"Do not have a valid user with apiToken"}

PR #10098, opened 2023-11-03
signed URLs with private url users, autogenerate API token

### LINKED ISSUE #10045: Auto generate USER API Token to enable External Tool GET/POST metadata
Hello

We are working on an external tool(TurboCurator by ICPSR) that will enable the user to look at recommendations/suggestions for title/summary/keywords.
Currently this tool is available on beta.dataverse.org and it is not fully functional yet. We are actively developing and testing on the beta site.
For this tool to pull dataset-metadata from dataverse, we need a user API token to be present already for the signed URLs to work. But that is not the case for all users. So, for this tool to work properly we would like for DATAVERSE to create an API token for the User if there is not one already created.

This API token creation can happen after they click on our external tool link in the "Access Dataset" dropdown or whenever you think its appropriate to do so.

Thanks
Raghunath Ravi
Sr. Developer
ICPSR-CNS
University of Michigan

### LINKED ISSUE #10093: SignedUrls don't work for privateUrlUsers
The way privateUrlUser credentials are converted into an ApiToken in the Files API class doesn't provide a username for use in generating signedUrls (and username is required). This breaks generating signedUrls for use with externalTools/Previewers configured for signedUrls when accessed by a PrivateUrlUser.

5.14 - develop

PR to follow.

qqmyers:
**What this PR does / why we need it**: This PR addresses issues related to using signedUrls as a PrivateUrlUser and, as part of refactoring/reducing duplicate code, also adds generating an API token if needed to launch a config tool.

For PrivateUrlUsers, signedUrls weren't working for a couple reasons: 
* The signedUrl generator was not being given a username for PrivateUrlUsers which is required in Dataverse
* PrivateUrlUsers have names of the form "PREFIX + datasetId" where PREFIX was '#'. Using this name in a signedUrl is problematic as '#' indicates the end of a URL (and start of an anchor) unless escaped.
* The SignedUrl validation mechanism assumed only AuthenticatedUsers were valid and did not have logic to look up the tokens for PrivateUrlUsers.

- Closes #10093
- Closes #10045

Re: the PrivateUrlUser PREFIX - I chose to change this to '!' rather than use a character that has to be escaped (given that other rounds of escaping/unescaping may be happening). The only place this PREFIX was recorded is in the roleassignment table, and a flyway update script is given.  I don't think these were exposed anywhere - perhaps in some api call? Alternately, the username used for SignedUrl purposes could be different than what is used in the db, but that seemed to be less straight-forward. (One could go without a PREFIX in signedUrl usernames, but it seems useful to be able to distinguish PrivateUrlUsers without having to see if the rest of the name can be parsed as a datasetId.)
W.r.t. #10045/autogenerating api keys - we do this in some places now, This PR adds it for dataset-level config tools. With this PR enabling signedUrls for PrivateUrlUsers and a planned PR to convert the existing Previewers to use signedUrls, it seems like were close to a model where we could keep the API key hidden unless the user wanted to use it explicitly. That may be a good way to address the concerns in #9898.

Configure a file previewer to use signedUrls and verify that the signedUrls provided work on a draft file (or restricted file) for a normal user and for a PrivateUrlUser, both in the file page and when launched as a separate page.
The signedUrls can be verified manually or by using one of the previewers now enabled for signedUrls (new PR being added to https://github.com/gdcc/dataverse-previewers). Manual verification involves looking at the URL used to launch the previewer in the browser console/network tab, getting the callback parameter, base64 decoding it, using that signedUrl in the browser to retrieve the params and allowedApiCall signedUrls and then verify that they also work. (I can do a quick walkthrough for anyone.)
Re: #10045 - this is needed for TurboCurator and using the TurboCurator configuration currently set up on the beta machine would be easiest (i.e. use the same configuration json as their to register the TurboCurator tool on your test instance.) As with previewers, manually verifying that the signedURLs returned work would be enough. (afaik, TurboCurator is not yet publicly visible.)

**Is there a release notes update needed for this change?**: I've included some notes - not sure what rises to the level of being called out. I also went ahead and made the notes reference new Previewers that can use signed Urls. While those aren't in this PR and aren't part of a 6.1 release per se, they will work with 6.1 (and the PrivateUrl fix here means that 6.1 is the first version where they will work for PrivateUrl users. They should work for normal authenticated users in 5.14/6.0 as well.)

**Additional documentation**:

https://dataverse-guide--10098.org.readthedocs.build/en/10098/api/external-tools.html


pdurbin:
Does this PR fix the following issue?

- #9405

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -931,14 +939,45 @@ public List <WorkflowComment> getWorkflowCommentsByAuthenticatedUser(Authenticat
...
+     * @param user
+     * @return a token or null (i.e. if the user is not an AuthenticatedUser or PrivateUrlUser)
+     */
+
+    public ApiToken getValidApiTokenForUser(User user) {
+        ApiToken apiToken = null;
+        if (user instanceof AuthenticatedUser) {
+            apiToken = getValidApiTokenForAuthenticatedUser((AuthenticatedUser) user);
+        } else if (user instanceof PrivateUrlUser) {
+            PrivateUrlUser privateUrlUser = (PrivateUrlUser) user;
```
pdurbin, line 972: Netbeans is suggesting the fancy new Java 14 instanceof automatic casting feature ( https://blogs.oracle.com/javamagazine/post/pattern-matching-for-instanceof-in-java-14 ), but I'll leave it alone.

src/main/java/edu/harvard/iq/dataverse/authorization/users/PrivateUrlUser.java
```diff
@@ -12,7 +12,7 @@
  */
 public class PrivateUrlUser implements User {
 
-    public static final String PREFIX = "#";
+    public static final String PREFIX = "!";
```
pdurbin, line 15: I don't remember why I picked `#` but I think it's fine to change it (I'm glad there's a flyway script).

src/main/java/edu/harvard/iq/dataverse/api/auth/SignedUrlAuthMechanism.java
```diff
@@ -45,26 +50,35 @@ private String getSignedUrlRequestParameter(ContainerRequestContext containerReq
...
         UriInfo uriInfo = containerRequestContext.getUriInfo();
         String userId = uriInfo.getQueryParameters().getFirst(SIGNED_URL_USER);
-        AuthenticatedUser targetUser = authSvc.getAuthenticatedUser(userId);
-        ApiToken userApiToken = authSvc.findApiTokenByUser(targetUser);
+        User targetUser = null; 
+        ApiToken userApiToken = null;
+        if(!userId.startsWith(PrivateUrlUser.PREFIX)) {
+        targetUser = authSvc.getAuthenticatedUser(userId);
+        userApiToken = authSvc.findApiTokenByUser((AuthenticatedUser)targetUser);
+        } else {
```
pdurbin: Indentation is off. I'll push a fix.

doc/release-notes/10093-signedUrl_improvements.md
```diff
@@ -0,0 +1,5 @@
+A new version of the standard Dataverse Previewers from https://github/com/gdcc/dataverse-previewers is available. The new version supports the use of signedUrls rather than API keys when previewing restricted files (including files in draft dataset versions). Upgrading is highly recommended.
```
pdurbin, line 1: I'm a little confused by this. When Dataverse 6.1 is released, will Dataverse Previewers 1.4 be released?

pdurbin (commented):
The release note is a little confusing. Otherwise, the code is looking good. I haven't tested it but API tests are passing.

pdurbin:
> The signedUrls can be verified manually or by using one of the previewers now enabled for signedUrls (new PR being added to https://github.com/gdcc/dataverse-previewers).

Has this new PR been added yet? I can't find it.

pdurbin:
@qqmyers can you please profile a tool and manifest to test with? I tried simply modifying the existing text previewer but it doesn't work. Here's the manifest I used:

```
curl -X POST -H 'Content-type: application/json' http://localhost:8080/api/admin/externalTools -d \
'{
  "displayName":"Signed URL Read Text",
  "description":"Read the text file.",
  "toolName":"textPreviewer",
  "scope":"file",
  "types":["preview"],
  "toolUrl":"http://localhost:8080/dataexplore/dataverse-previewers/previewers/v1.3/TextPreview.html",
  "toolParameters": {
      "queryParameters":[
        {"fileid":"{fileId}"},
        {"siteUrl":"{siteUrl}"},
        {"datasetid":"{datasetId}"},
        {"datasetversion":"{datasetVersion}"},
        {"locale":"{localeCode}"}
      ]
    },
  "allowedApiCalls": [
    {
      "name":"retrieveDataFile",
      "httpMethod":"GET",
      "urlTemplate":"/api/v1/access/datafile/{fileId}",
      "timeOut":270
    }
  ],
  "contentType":"text/plain"
}'
```

The URL in my browser (if it helps): http://localhost:8080/dataexplore/dataverse-previewers/previewers/v1.3/TextPreview.html?callback=aHR0cDovL2xvY2FsaG9zdDo4MDgwL2FwaS92MS9maWxlcy84L21ldGFkYXRhLzcvdG9vbHBhcmFtcy8yP3VudGlsPTIwMjMtMTEtMjFUMjA6Mjc6NTMuOTU3JnVzZXI9ZGF0YXZlcnNlQWRtaW4mbWV0aG9kPUdFVCZ0b2tlbj01NTJiYzZkNmUwZTk1NTNmNWI4ZmI2YTU4NTg5MWYyZmE1ZTdiM2JiMDlhN2JmZThmNjUyMmNmZmJmYmI2OWJjYjAyNzdjYjEzZGQ1NGUyYzBlODYwY2E4ZDM4ZWM5ZjVkMDc0MWJmNDUwMWNhYzVjZWUyZmY4MzU5ZDQwMjZiMQ==&locale=en

What I see in the browser:

Screenshot 2023-11-21 at 3 27 57 PM

qqmyers:
The new previewers aren't out yet, so using the manual method described in how to test would be the best option. It looks like you have a valid config - the response you get has the base64 encoded callback param, which, when decoded, shows:
`http://localhost:8080/api/v1/files/8/metadata/7/toolparams/2?until=2023-11-21T20:27:53.957&user=dataverseAdmin&method=GET&token=552bc6d6e0e9553f5b8fb6a585891f2fa5e7b3bb09a7bfe8f6522cffbfbb69bcb0277cb13dd54e2c0e860ca8d38ec9f5d0741bf4501cac5cee2ff8359d4026b1`
If you had used that before the timeout, you would get json that includes json with the specific signed URL you requested. That URL should also work. The PR was primarily about making sure this also works with a privateURL user and should also cover the dataset-level config tool issue (though I'm not sure I tested that).

doc/release-notes/10093-signedUrl_improvements.md
qqmyers: That is my intent. Feel free to hedge.

Issue #9495, opened 2023-04-05
Experimental dataset JSON GET API doesn't return multi-valued fields as arrays


vera:
**What steps does it take to reproduce the issue?**

1. Create dataset with multi-valued field, e.g. `subject` in the citation block
2. Use experimental GET API to retrieve dataset https://guides.dataverse.org/en/latest/developers/dataset-semantic-metadata-api.html#get-dataset-metadata e.g. `GET /api/datasets/42/metadata`

* What happens?

The multi-valued field values are not returned as arrays, e.g. `'subject': 'Medicine, Health and Life Sciences'`

Side note, the non-experimental API https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-dataset does not have this bug: `{'typeName': 'subject', 'multiple': True, 'typeClass': 'controlledVocabulary', 'value': ['Medicine, Health and Life Sciences']}`

* What did you expect to happen?

The multi-valued field values to be returned as arrays, e.g. `'subject': ['Medicine, Health and Life Sciences']`

**Which version of Dataverse are you using?**

v5.13

**Any related open or closed issues to this bug report?**

/

**Screenshots:**

/

qqmyers:
FWIW: That API (and the ORE export) suppresses using an array when there is only one value, but it does use an array when there is more than one value. The relevant line is https://github.com/IQSS/dataverse/blob/97310b3e33db4657e28cdffb89e55b6e9055a798/src/main/java/edu/harvard/iq/dataverse/util/bagit/OREMap.java#L440. To change it you'd need to keep track of whether the field allows multiple values or not (since single value fields still shouldn't use an array). Any PR for this should include a release note about the backward incompatibility being introduced.

pdurbin:
I'm in favor of the predictability of the subject always being an array. We could always put this in new field (rather than changing the field that is sometimes a string an sometimes an array) for backward compatibility.

Issue #9494, opened 2023-04-05
Dataset JSON GET API returns numbers as strings


vera:
**What steps does it take to reproduce the issue?**

1. Enable metadata block with fields with `fieldType` `int` or `float`
2. Create dataset
3. Use the JSON GET API to retrieve the dataset https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-dataset e.g. `GET /api/datasets/42?exporter=dataverse_json`

* What happens?

The values for the numerical fields are returned as strings, e.g. `{'typeName': 'my_int_field', 'multiple': False, 'typeClass': 'primitive', 'value': '5'}`

* What did you expect to happen?

The numerical fields to be returned as integers/floats, e.g. `{'typeName': 'my_int_field', 'multiple': False, 'typeClass': 'primitive', 'value': 5}`

**Which version of Dataverse are you using?**

v5.13

**Any related open or closed issues to this bug report?**

-

**Screenshots:**

-

vera:
I would be happy to submit a PR (I assume this would not be very hard to fix?!) if someone could point me to the right part of the code.

qqmyers:
FWIW: This wouldn't be hard to change (I think the code is all in util.JsonPrinter) but there would be many places where that format is used that would have to change - most of the metadata exporters parse that JSON, previewers and external tools use it, probably pyDataverse, community-developed applications, and we expect to be able to read that format back in, so all of the code to import the Json could have to be changed to accept this (hopefully keeping backward compatibility as well). I know some JSON APIs will work as long as the value can be converted to the requested type, but I think there are a lot of places where a change like this could have an effect.
 
There is some discussion going on w.r.t. using a JSON Schema (e.g. in #9463) which could involve changes to the JSON output that would be significant enough to perhaps be a version 2 of the API and where changes like this might be something to incorporate.

PR #9213, opened 2022-12-08
#8944 - Extend 'metadatablocks/{block_id}' endpoint JSON output

JR-1991:
_This is a continuation of PR #9091 due to a change of the branch name._

**What this PR does / why we need it**:

This PR extends the API endpoint ```/api/metadatablocks/{block_id}``` to include the following fields:

- ```multiple```: Whether or not multiple instances of this field can exist
- ```isControlledVocabulary```:  Whether or not this field has a controlled vocabulary. Important to derive the ```typeClass``` later on
- ```controlledVocabularyValues ``` - If a vocabulary is given, includes all possible values.

With this extension it is possible to build compatible Dataverse JSON files for upload and download from any installation without having access to the TSV files. EasyDataverse would greatly benefit from this and allow users to "connect" to a Dataverse installation without the need of generating specific code.

**Which issue(s) this PR closes**:

- Closes #8944
- Closes #1510

**Special notes for your reviewer**:

None

**Suggestions on how to test this**:

Add or extend a fixture to test if the new fields are given upon execution.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

**Is there a release notes update needed for this change?**:

The API endpoint ```/api/metadatablocks/{block_id}``` has been extended to include the following fields:

- ```multiple```: Whether or not multiple instances of this field can exist
- ```isControlledVocabulary```:  Whether or not this field has a controlled vocabulary.
- ```controlledVocabularyValues ``` - If a vocabulary is given, includes all possible values.

**Additional documentation**:

 None

PR #10109, opened 2023-11-09
JSON Schema creator and validator

### LINKED ISSUE #6978: Query Dataverse for mandatory metadata fields via API
RSpace ELN uses the Dataverse API to submit research data to Dataverse. It has a minimal UI for metadata fields such as title, subject, description, authors, contacts and this works on various Dataverses till now.

One of our RSpace customers has their own Dataverse as well - 4.19. They have configured Dataverse to require additional metadata when submitting a Dataset. RSpace doesn't know these fields are mandatory and submission fails:

`Deposit failed: ERROR2020-06-09T10:58:26ZProcessing failedCouldn't update dataset edu.harvard.iq.dataverse.engine.command.exception.IllegalCommandException: Validation Failed: Producer Name is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ]), Distributor Name is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ]), Description Date is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ]), Keyword Term is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ]), Deposit Date is required. (Invalid value:edu.harvard.iq.dataverse.DatasetField[ id=null ]).`

This corresponds exactly to what are the required properties as sent by the Dataverse admin, that are not set by RSpace
- Author - Name
- Contact - Name
- Contact - Email
- Description - Text
- Description - Date
- Keyword - Term
- Producer - Name
- Distributor - Name (In our default templates, this is always the name of the (sub-)dataverse. I'm not sure how this should be handled when a dataset is created from RSpace.)
- Deposit Date (In Dataverse, this is generated by the system.)

If this list never changes, then RSpace could develop a solution where it reads a list of mandatory fields from a configuration file. But if it does change from time to time, it would be great if there was an API method in Dataverse to get a list of mandatory metadata fields .Then, a client could programmatically generate input fields for these properties so that the end-user could make a valid submission.

### LINKED ISSUE #9464: Dataset Json Validation: Update the code to use this schema
This is the second step in addressing: https://github.com/IQSS/dataverse-pm/issues/26

Use this schema to update the code. We currently have a library that checks the JSON schema (wc3provenance schema). As part of this solution we should not feel limited to using our current library. (On the page about the JSON schema there are multiple java library options available for validation.)(10)

### LINKED ISSUE #9465: Dataset Json Validation: Create API endpoints to retrieve a JSON schema
This is the third step in addressing: https://github.com/IQSS/dataverse-pm/issues/26

Create API endpoints to retrieve a JSON schema that also has info about the required meta-data so that we can create a full-fledged validation of a dataset. That API end point has a paramenter that is the dataverse collection that you are targetting for deposit. (10)

sekmiller:
**What this PR does / why we need it**: Creates a custom json schema for a given Dataverse Collection that can be used to validate a dataset.json file prior to uploading to create the dataset

Closes #9464 Dataset json validation update code to use schema
Closes #9465 Create api.endpoint for Dataset json validation.
Closes #6978 Query Dataverse for mandatory metadata fields via API 

**Special notes for your reviewer**: The api returns a json string that includes the escape characters. wasn't sure if that that is OK or if I should strip them out before returning

**Is there a release notes update needed for this change?**: Release notes are included

**Additional documentation**: doc updated to include the new endpoints

https://dataverse-guide--10109.org.readthedocs.build/en/10109/api/native-api.html#retrieve-a-dataset-json-schema-for-a-collection


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ValidateDatasetJsonCommand.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author stephenkraffmiller
+ */
+@RequiredPermissions(Permission.AddDataset)
+public class ValidateDatasetJsonCommand extends AbstractCommand<String> {
+    
+    private static final Logger logger = Logger.getLogger(GetDatasetSchemaCommand.class.getCanonicalName());
```
landreev, line 24: Copy-and-paste from the other command, probably.

landreev:
I'm not seeing this discussed, in the pr or the linked issues - sorry if I missed it - but was there a specific reason to only use this validation code in the dedicated "validate" api? - and not always use it in the "create" api as well, before attempting to import the json uploaded by the user? 
(are we just assuming that the attempt to import is validation enough, so there's no reason to spend extra cycles pre-validating? ... and if the user needs to know more details about why it failed, then they can run the validate api? - I may have answered my own question already, but please let me know what you think)

landreev:
Something about the fact that the schema is always created on the fly makes me a little uncomfortable.  It should only change when a) metadatablocks configuration changes for the collection and b) when metadatablocks themselves are updated... so it should be possible to cache; and in many practical cases it will never need to be refreshed. 

I'm not sure where we would cache it and how. So this is probably outside the scope for now.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -505,6 +505,51 @@ The fully expanded example above (without environment variables) looks like this
 
 .. note:: Previous endpoints ``$SERVER/api/dataverses/$id/metadatablocks/:isRoot`` and ``POST https://$SERVER/api/dataverses/$id/metadatablocks/:isRoot?key=$apiKey`` are deprecated, but supported.
 
+.. _get-dataset-json-schema:
+
+Retrieve a JSON schema for a Collection
```
pdurbin, line 510:
```suggestion
Retrieve a JSON Schema for a Collection
```

pdurbin (commented):
> Something about the fact that the schema is always created on the fly makes me a little uncomfortable.

I feel the same way. While it's very cool that Dataverse can generate a JSON Schema on the fly for a specific collection, we should supply a static copy of generic JSON Schema in our API Guide. By generic I'm thinking the five standard required fields.

landreev (approved):
Looks good! (and the Jenkins tests are now passing).

pdurbin:
@sekmiller I'm just repeating these comments from me and Leonid in case you missed them:

> Something about the fact that the schema is always created on the fly makes me a little uncomfortable.

I feel the same way. While it's very cool that Dataverse can generate a JSON Schema on the fly for a specific collection, we should supply a static copy of generic JSON Schema in our API Guide. By generic I'm thinking the five standard required fields.

sekmiller:
@pdurbin, yeah I just added a base schema file yesterday. I will add some logic to see if the collection has any extra required fields before generating on the fly.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -232,6 +233,39 @@ public Response addDataverse(@Context ContainerRequestContext crc, String body,
...
+    @GET
+    @AuthRequired
+    @Path("{identifier}/datasetSchema")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getDatasetSchema(@Context ContainerRequestContext crc, @PathParam("identifier") String idtf) {
+        User u = getRequestUser(crc);
+
+        try {
+            String datasetSchema = execCommand(new GetDatasetSchemaCommand(createDataverseRequest(u), findDataverseOrDie(idtf)));
+            return ok(datasetSchema);
```
pdurbin, line 261: @sekmiller and were talking abouthow we probably want to return just the JSON instead of escaped JSON in our normal "ok... data" data structure.

Jim did this recently in this commit:  9953 - don't wrap linkset in a data element 3a4d8f98053ff726c617a45c5ad15d2f3059c138

pdurbin: @sekmiller and I decided to return just the JSON Schema. See 2d3f7ab. This way instead of the \n characters and data...message, you get just what you want:

```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "$defs": {
        "field": {
            "type": "object",
            "required": [
                "typeClass",
...
```

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -919,5 +932,262 @@ public List<Object[]> getDatasetTitlesWithinDataverse(Long dataverseId) {
...
+        }
+    }
+    
+    private  String datasetSchemaPreface = 
+    "{\n" +
+    "    \"$schema\": \"http://json-schema.org/draft-04/schema#\",\n" +
+    "    \"$defs\": {\n" +
+    "    \"field\": {\n" + 
+    "        \"type\": \"object\",\n" +
+    "        \"required\": [\"typeClass\", \"multiple\", \"typeName\"],\n" +
```
pdurbin, line 1104: We could use the fancy new text blocks here: https://docs.oracle.com/en/java/javase/21/text-blocks/index.html

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -919,5 +932,262 @@ public List<Object[]> getDatasetTitlesWithinDataverse(Long dataverseId) {
...
+            }
+        }
+        if (mdbHasReqField){
+        retval  += startOfMDB.replace("blockName", mdb.getName());
+        
+        retval += minItemsTemplate.replace("numMinItems", Integer.toString(requiredThisMDB.size()));
+        int count = 0;
+        for (DatasetFieldType dsft:requiredThisMDB ){
+            count++;
+            String reqValImp = reqValTemplate.replace("reqFieldTypeName", dsft.getName());
```
pdurbin, line 1052: I'm sort of wondering why we're starting with a string and doing `replace()` here and there rather than building up a JSON object.

pdurbin (approved):
Tests are passing. @landreev already approved this. I'm approving it too.

I did leave a question/comment about how we're constructing the JSON. We seem to start with a template (a String) and do `replace` on it here and there. Another approach could be to create a JSON object programatically. Not sure what's best.

I did also add some tests and tweaks some docs. I didn't do any testing apart from running the new tests.

I'm wondering if it's possible to mark fields as required via API. If not, the SPA will surely need this sooner or later.

pdurbin:
@JR-1991 are you thinking the values themselves (Chemistry, Law, etc.) should be returned in the JSON Schema?

That would certainly be more convenient than hitting another API to get them. Thanks again for adding that in this PR:

- #9213

Also, I'm wondering if this PR closes the following issue:

- #6978

I'll go ask. 😄

johannes-darms:
I'm not sure if we should address this in this MR. However, IMHO its crucial that the json schema encodes the correct datatypes as defined within the metadatablock and the API returns data that is valid wrt. to the schema. So it would be nice if these two issues #9494. #9495 were somehow addressed.

pdurbin:
> crucial that the json schema encodes the correct datatypes as defined within the metadatablock

A good example of this is that the geospatial block expects floats for north, south, east, west (latitude and longitude). See this related pull request by @stevenwinship and @landreev merged yesterday:

- #10142

JR-1991:
> @JR-1991 are you thinking the values themselves (Chemistry, Law, etc.) should be returned in the JSON Schema?

Sorry for missing your comment earlier. I believe this addition would be valuable and allow for client-side controlled vocabulary validation.

PR #10115, opened 2023-11-10
Single retry when datafile checksum validation fails

### LINKED ISSUE #10114: Publish fails for large datasets on some s3 implementations
When using s3 as storage, publish a large dataset where checksum verification takes longer than 30 minutes.

* When does this issue occur?
When publishing datasets with many large files where checksum validation takes longer than 30 minutes. This happens consistently on our installation with OpenStack s3 implementation.

Dataset publish.

* What happens?
S3 closes the connection unexpectedly.

```
java.lang.RuntimeException: java.net.SocketException: Connection reset by peer (Write failed)
  at edu.harvard.iq.dataverse.util.FileUtil.calculateChecksum(FileUtil.java:742)
  at edu.harvard.iq.dataverse.util.FileUtil.validateDataFileChecksum(FileUtil.java:1902)
  at edu.harvard.iq.dataverse.engine.command.impl.FinalizeDatasetPublicationCommand.validateDataFiles(FinalizeDatasetPublicationCommand.java:325)
  at edu.harvard.iq.dataverse.engine.command.impl.FinalizeDatasetPublicationCommand.execute(FinalizeDatasetPublicationCommand.java:89)
  at edu.harvard.iq.dataverse.engine.command.impl.FinalizeDatasetPublicationCommand.execute(FinalizeDatasetPublicationCommand.java:50)
  at edu.harvard.iq.dataverse.EjbDataverseEngineInner.submit(EjbDataverseEngineInner.java:36)
```

Users with publish permissions.

Dataset gets published.

5.14, 6.0, current develop branch, ...

No.

**Screenshots:**

image

ErykKul:
It fixes the problem when publishing large datasets, and the s3 server closes the connection. On some installations, this happens every 20 to 30 minutes. In that case, publishing of large datasets is not possible without disabling the validation.

Closes #10114 

No

No

**Additional documentation**:
The fix implements a single retry when datafile validation fails. If the fail was due to a lost connection, the checksum gets recomputed and the publishing succeeds.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
@@ -1469,37 +1469,24 @@ public class FileUtil implements java.io.Serializable  {
         }
 
         StorageIO<DataFile> storage = dataFile.getStorageIO();
-        InputStream in = null;
+        String recalculatedChecksum = null;
 
-        try {
-            in = getOriginalFileInputStream(storage, dataFile.isTabularData());
+        try (InputStream inputStream = getOriginalFileInputStream(storage, dataFile.isTabularData())) {
+            recalculatedChecksum = FileUtil.calculateChecksum(inputStream, checksumType);
         } catch (IOException ioex) {
-            in = null;
-        }
-
-        if (in == null) {
             String info = BundleUtil.getStringFromBundle("dataset.publish.file.validation.error.failRead", Arrays.asList(dataFile.getId().toString()));
             logger.log(Level.INFO, info);
             throw new IOException(info);
-        }
-
-        String recalculatedChecksum = null;
-        try {
-            recalculatedChecksum = FileUtil.calculateChecksum(in, checksumType);
         } catch (RuntimeException rte) {
             logger.log(Level.SEVERE, "failed to calculated checksum, one retry", rte);
             recalculatedChecksum = null;
-            IOUtils.closeQuietly(in);
+        }
+
+        if (recalculatedChecksum == null) { //retry once
             storage = dataFile.getStorageIO();
-            try {
-                in = getOriginalFileInputStream(storage, dataFile.isTabularData());
-                recalculatedChecksum = FileUtil.calculateChecksum(in, checksumType);
-            } catch (RuntimeException rte2) {
-                logger.log(Level.SEVERE, "failed to calculated checksum, no retry", rte2);
-                recalculatedChecksum = null;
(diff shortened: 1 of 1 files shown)


ErykKul:
Link to the conversation on the Google Groups:
https://groups.google.com/g/dataverse-community/c/cQ3k8OvhHcI/m/vhDreGXTAAAJ

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1449,6 +1449,16 @@ public static S3AccessIO getS3AccessForDirectUpload(Dataset dataset) {
     	return s3io;
     }
     
+    private static InputStream getInputStream(StorageIO<DataFile> storage, boolean isTabularData) throws IOException {
```
qqmyers, line 1452: Perhaps call this getOriginalFileInputStream to make it clearer? Also - any reason not to put the storage.open call in this method?

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -1484,7 +1487,18 @@ public static void validateDataFileChecksum(DataFile dataFile) throws IOExceptio
         try {
             recalculatedChecksum = FileUtil.calculateChecksum(in, checksumType);
         } catch (RuntimeException rte) {
+            logger.log(Level.SEVERE, "failed to calculated checksum, one retry", rte);
             recalculatedChecksum = null;
+            IOUtils.closeQuietly(in);
+            storage = dataFile.getStorageIO();
+            try {
+                storage.open(DataAccessOption.READ_ACCESS);
```
qqmyers, line 1495: If storage.open is part of the get stream call, could you use try with resources here? (I think your code works but having the outer try close the stream you open in the inner try's catch seems fragile. Plus we're slowly shifting to use try with resources in general.)

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
ErykKul: Done.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
ErykKul: I refactored the try-catch blocks to the try-with-resources. I think the code is not only safer but also more readable right now. Thanks!

qqmyers (approved):
Looks good

landreev:
This is great, merging.

PR #10122, opened 2023-11-15
make BagIT source organization info (name, address, email) configurable

### LINKED ISSUE #8760: Dataset Archiving: BagIt bag-info.txt has some fields that are populated from bundle.properties, could be doc'd and maybe made a setting?
While testing a recent bag-it, dataset archiving change, noticed the bag-info.txt file had some filler, replacement text in it. In discussing with @qqmyers it was identified as an earlier attempt at populating the fields and that they could be populated in a more standardized and documented way:

cat bag-info.txt
Contact-Name: kc052622b, kc052622b
Contact-Email: [kc052622b@mailinator.com](mailto:kc052622b@mailinator.com)
Source-Organization: Dataverse Installation (<Site Url>)
Organization-Address: <Full address>
Organization-Email: <Email address>
External-Description: This is my description
Bagging-Date: 2022-05-26
External-Identifier: doi:10.70122/FK2/DJNPSO
Bag-Size: 109.77 KB
Payload-Oxum: 109765.3
Internal-Sender-Identifier: Dataverse Internal:kc052622b

from @qqmyers: 

they ~are filled in, with the values from https://github.com/IQSS/dataverse/blob/552dfc8c7af7a0b0e0c3948e6de1a8321af98c37/src/main/java/propertyFiles/Bundle.properties#L2269-L2271 

if you want an issue to move these things, or document that they exist (not sure if they are in the guides or not), feel free to add one and I'll take a look.
[Bundle.properties](https://github.com/IQSS/dataverse/blob/552dfc8c7af7a0b0e0c3948e6de1a8321af98c37/src/main/java/propertyFiles/Bundle.properties)
bagit.sourceOrganization=Dataverse Installation (<Site Url>)
bagit.sourceOrganizationAddress=<Full address>
bagit.sourceOrganizationEmail=<Email address>

pdurbin:
## **What this PR does / why we need it**:

Currently values are hard-coded in Bundle.properties.

## **Which issue(s) this PR closes**:

- Closes #8760

## **Special notes for your reviewer**:

I tested this with MPCONFIG in Docker. I would add an API test for it but I don't think it's possible. I opened this issue about that:

- #10128

The following note is old. I left it in but the bottom line is that I suggesting `,` instead of `\n` per the RFC.

While testing with these values...

```
DATAVERSE_BAGIT_SOURCEORG_NAME=LibraScholar
DATAVERSE_BAGIT_SOURCEORG_ADDRESS=123 Wisdom Way\nCambridge, MA\nUSA
DATAVERSE_BAGIT_SOURCEORG_EMAIL=hello@dataverse.librascholar.edu
```

... I get this output...

```
Source-Organization: LibraScholar
Organization-Address: 123 Wisdom Way\nCambridge, MA\nUSA
Organization-Email: hello@dataverse.librascholar.edu
```

Do we want the address on all one line? Maybe with commas?

Update: the example at https://datatracker.ietf.org/doc/html/rfc8493 uses commas (`1 Main St., Cupertino, California, 11111`) so I'm going with that and mentioned it in our guides.

## **Suggestions on how to test this**:

- use the new config (either as JVM options with MPCONFIG)
- export a bag
- unzip it
- confirm values are correct in bag-info.txt

## **Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

## **Is there a release notes update needed for this change?**:

Yes, included.

## **Additional documentation**:

Included: https://dataverse-guide--10122.org.readthedocs.build/en/10122/installation/config.html#bag-info-txt


src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -2425,5 +2426,27 @@ public Response getSignedUrl(@Context ContainerRequestContext crc, JsonObject ur
...
+        try {
+            AuthenticatedUser user = getRequestAuthenticatedUserOrDie(crc);
+            if (!user.isSuperuser()) {
+                return error(Response.Status.FORBIDDEN, "Superusers only.");
+            }
+        } catch (WrappedResponse wr) {
+            return wr.getResponse();
+        }
+        try {
+            return ok(new FileInputStream(pathToFile));
```
qqmyers, line 2446: even as a protected admin call, I'd suggest limiting this to some subdirectory

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -150,6 +150,13 @@ public enum JvmSettings {
     SCOPE_NETCDF(PREFIX, "netcdf"),
     GEO_EXTRACT_S3_DIRECT_UPLOAD(SCOPE_NETCDF, "geo-extract-s3-direct-upload"),
 
+    // BAGIT SETTINGS
+    SCOPE_BAGIT(PREFIX, "bagit"),
```
qqmyers, line 154: We already have a Brand name for the repository and associated email address. Should those be used here instead of new settings? If not, I don't know that these should be in a bagit scope - they probably could/should go in other exports, be available via API, etc. as generally usable info.

pdurbin: I had that thought... I'm ok with reusing existing settings.

Should "mailing address" also be a more global setting? Not BagIt-specific, I mean?

qqmyers: Yeah - not sure how settings will be organized but I'd think the address could be a dataverse.owner.address or something like that (dataverse.organization.address?)

pdurbin: I'm asking the community what they think: https://groups.google.com/g/dataverse-community/c/E3Vfrs757cU/m/HWNvu86yCAAJ

pdurbin: @shlake thanks for the feedback! ❤️ 

Two things:

- I like the way you think because it means what I've already coded up should work for you. 😜 You seem ok with dedicated BagIt settings.
- It's interesting that you put "University of Virginia" under "Organization-Address". Technically [RFC 8493](https://datatracker.ietf.org/doc/html/rfc8493) says it's for the "mailing address of the source organization". The example is comma-separated (`Organization-Address: 1 Main St., Cupertino, California, 11111`) so that's probably what I'll suggest in our documentation. (That way I don't have to complicate our code to support multi-line addresses.) As for addresses changing, you're absolutely right! Please use the field in whatever way you feel is best!

Thanks again!

pdurbin: I just documented the new BagIt settings: f0f5ac7

That is to say, I'm going with what @shlake wants and letting these settings have their own life independent of `:InstallationName`, `:SystemEmail`, etc.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
pdurbin: Hmm, yeah, I can easily download `/etc/passwd` which seems pretty suboptimal. 😬

pdurbin:
I added the 6.1 milestone because the issue has a 6.1 milestone.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
pdurbin: In fa6f850 I limited the API to downloading files from `/tmp`.

sekmiller (approved):

PR #10124, opened 2023-11-16
 JDD Metrics: Label KO #10123

### LINKED ISSUE #10123: JDD Metrics: Label KO
In the display within the metric block under "download," there is a parenthesis that appears. Please refer to the attached screenshot for clarification. Thank you.
Capture3

sbondka:
In the display within the metric block under "download," there is a parenthesis that appears. Please refer to the attached screenshot for clarification. Thank you.
Capture3

- closes #10123

### DIFF
--- a/src/main/webapp/dataset.xhtml
+++ b/src/main/webapp/dataset.xhtml
@@ -571,13 +571,13 @@
                                                     </h:outputFormat>
                                                     <span class="glyphicon glyphicon-question-sign tooltip-icon"
                                                           data-toggle="tooltip" data-placement="auto top" data-original-title="#{bundle['metrics.dataset.downloads.makedatacount.tip']}"></span>
-                                                    <span jsf:rendered="#{settingsWrapper.getMDCStartDate()!=null}">        
+                                                    <span jsf:rendered="#{settingsWrapper.getMDCStartDate()!=null and guestbookResponseServiceBean.getDownloadCountByDatasetId(DatasetPage.dataset.id, settingsWrapper.getMDCStartDate()) > 0 }">        
                                                         <h:outputFormat value="#{bundle['metrics.downloads.nonMDC']}">
                                                             <f:param value="#{guestbookResponseServiceBean.getDownloadCountByDatasetId(DatasetPage.dataset.id, settingsWrapper.getMDCStartDate())}"/>
                                                         </h:outputFormat>
-                                                        <span jsf:rendered="#{guestbookResponseServiceBean.getDownloadCountByDatasetId(DatasetPage.dataset.id, settingsWrapper.getMDCStartDate()) > 0}" class="glyphicon glyphicon-question-sign tooltip-icon"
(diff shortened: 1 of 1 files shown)


qqmyers:
I see the difference at QDR now. That closing paren ) is only intended to show when there are pre-MDC counts available. Right now the [span that includes the paren](https://github.com/IQSS/dataverse/blob/8d8f35f919794ae6650e17f5e7e6e792a1064cfa/src/main/webapp/dataset.xhtml#L574) will show when there are no counts while the [span that would show an opening paren ( and the number of pre-MDC counts](https://github.com/IQSS/dataverse/blob/8d8f35f919794ae6650e17f5e7e6e792a1064cfa/src/main/webapp/dataset.xhtml#L578) is ~correctly suppressed. At QDR, I've moved the render condition on the inner span (#{`guestbookResponseServiceBean.getDownloadCountByDatasetId(DatasetPage.dataset.id, settingsWrapper.getMDCStartDate()) > 0`) up to the outer one. That keeps the closing paren from appearing when there are no counts but leaves it in place when there are.

pdurbin:
@qqmyers I'm a little confused. Is either fix valid? The commit by @sbondka in this PR and the one you described? Or to ask another way, is there anything wrong with this PR as-is?

qqmyers:
Yes, as is the PR will remove the closing paren in the case where it isn't needed (issue #10123) as well as when it is (when there are pre-MDC counts).

pdurbin:
I see, thanks. @sbondka are you able to adjust your PR based on this feedback? Thanks!

sbondka:
Hello @qqmyers,

Thank you for your response. However, I need more details regarding the placement of the closing parenthesis. Should it be just before or after the question mark? If it's before, the correction will be in the Bundle.properties ( metrics.downloads.nonMDC ); if it's after, it will be in the dataset.xhtml file just before the </span> ( 
 data-toggle="tooltip" data-placement="auto top" [data-original-title="#{bundle['metrics.dataset.downloads.premakedatacount.tip']}">)</span>)  ).
 Please review the two screenshots.
MicrosoftTeams-image (3)
MicrosoftTeams-image (2)


Thank you.

qqmyers:
This is how it is at QDR: 
image
- the paren is after the help text (the help text there is about the number in the parenthesis).

sbondka:
Hello,

With the condition `settingsWrapper.getMDCStartDate() != null` alone,
we were entering this block, and the display of ")" persisted in all cases
which was our problem.

To fix this, I add the condition
`guestbookResponseServiceBean.getDownloadCountByDatasetId(DatasetPage.dataset.id, settingsWrapper.getMDCStartDate()) > 0`
along with `settingsWrapper.getMDCStartDate()!=null` .

Thanks.

qqmyers (approved):
Looks good - thanks!

DS-INRA:
Closes #10123
should be added so that the issue closes too when merged :)

pdurbin:
@DS-INRA added, thanks (it has to go into the description)

pdurbin (approved):
I didn't test this but the code change makes sense. We're simply moving some conditional logic higher up. Approved.

PR #9115, opened 2022-11-01
Deleting cruft under the watchful eyes of Phil

mreekie:
**What this PR does / why we need it**:

**Which issue(s) this PR closes**:

Closes #

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #10129, opened 2023-11-17
fix beta server, remove cruft: mdc logs

pdurbin:
The beta server is down. See https://help.hmdc.harvard.edu/Ticket/Display.html?id=352376

@donsizemore and I have a working theory that when `V6.0.0.4__10017-failure-with-long-custom-question.sql` was added in 1652764b44abbdd887a2957ec42ad31e108dd864 and then removed in 2424a55235cc015cfba283146dc2f97d2b8699af it's causing problem. Not sure.

I'm making this pull request to test the beta deploy workflow. Rather than a no-op, I thought I'd do something useful, so I'm deleting some cruft in the root of the git repo. I first make a swing at getting rid of it here:

- #9115


pdurbin:
Yes, exactly, a new script. This was the error:

Caused by: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Detected resolved migration not applied to database: 6.0.0.3.
To ignore this migration, set -ignoreMigrationPatterns='*:ignored'. To allow executing this migration, set -outOfOrder=true.

pdurbin:
Fixed and running the latest commit (1343de7)! Thanks, @donsizemore !

Screenshot 2023-11-17 at 11 27 58 AM

PR #10131, opened 2023-11-20
Remove the authentication annotation from info API

### LINKED ISSUE #9466: Evolve API authentication to omit it on endpoints intended to be open

After #9293 refactor, it has become more visible that there are endpoints that, despite being intended to be open, triggers user authentication, when it is not required.

This behavior already existed before the Auth Filter refactor, but it is now more visible by having the `@AuthRequired` annotation. Before the refactor, credential filtering was executed via the `AbstractApiBean.response(DataverseRequestHandler hdl)` method, which was called from several endpoints, a method which in turn called the `findUserOrDie` method.

These methods no longer exist, since the logic is now moved to the Auth Filter, and the same endpoints which used those methods now they are wrapped by the Auth Filter.

The goal of this issue is to simplify authentication by omitting the auth filter on endpoints that do not require user authentication. This makes the API code more understandable for developers and improves performance by bypassing the auth filter when it's not needed.

Example endpoint:  `/api/info/version`.

API User, developers

Slack discussion about confusion when seeing endpoint `/api/info/version` marked with `AuthRequired` 

API authentication

No

- #9293

jp-tosca:

After the changes that happened on #9303 and #9360  the `@AuthenticationRequired` was present on some API methods that are public and some additional code that is not required, this change:

-Overloads `AbstractAPIBean.response` requiring only a `DataverseRequestHandler` and passing a null `User`. 
-Removes the `@AuthenticationRequired` from the following info endpoints: {`version, server and apiTermsOfUse`}
-it also removes the `@Context ContainerRequestContext` from the signature of the method and makes a new call to the new re-introduced signature. This signature was deprecated on #9303 and removed on #9360.

Closes #9466 

This method signature was marked as deprecated on #9303, probably the intention of the author was to deprecate it while the functionality was migrated to the new authentication mechanism and removed once all the changes were done. 

-Run the APIs and integration tests.

N/A

N/A

**Additional documentation**:
N/A

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
@@ -636,6 +636,16 @@ public abstract class AbstractApiBean {
         }
     }
 
+    /***
+     * The preferred way of handling a request from an open API.
+     *
+     * @param hdl handling code block.
+     * @return HTTP Response appropriate for the way {@code hdl} executed.
+     */
+    protected Response response(DataverseRequestHandler hdl) {
+        return response(hdl, null);
+    }
+
     private Response handleDataverseRequestHandlerException(Exception ex) {
         String incidentId = UUID.randomUUID().toString();
         logger.log(Level.SEVERE, "API internal error " + incidentId +": " + ex.getMessage(), ex);
--- a/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
@@ -636,16 +636,6 @@ public abstract class AbstractApiBean {
         }
     }
 
-    /***
-     * The preferred way of handling a request from an open API.
-     *
-     * @param hdl handling code block.
-     * @return HTTP Response appropriate for the way {@code hdl} executed.
-     */
-    protected Response response(DataverseRequestHandler hdl) {
-        return response(hdl, null);
-    }
-
     private Response handleDataverseRequestHandlerException(Exception ex) {
         String incidentId = UUID.randomUUID().toString();
         logger.log(Level.SEVERE, "API internal error " + incidentId +": " + ex.getMessage(), ex);
--- a/src/main/java/edu/harvard/iq/dataverse/api/Info.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Info.java
@@ -35,30 +35,27 @@ public class Info extends AbstractApiBean {
     }
 
     @GET
-    @AuthRequired
     @Path("version")
-    public Response getInfo(@Context ContainerRequestContext crc) {
+    public Response getInfo() {
         String versionStr = systemConfig.getVersion(true);
         String[] comps = versionStr.split("build",2);
         String version = comps[0].trim();
         JsonValue build = comps.length > 1 ? Json.createArrayBuilder().add(comps[1].trim()).build().get(0) : JsonValue.NULL;
 
         return response( req -> ok( Json.createObjectBuilder().add("version", version)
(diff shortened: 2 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -636,6 +636,16 @@ protected Response response(DataverseRequestHandler hdl, User user) {
...
     }
 
+    /***
+     * The preferred way of handling a request from an open API.
+     *
+     * @param hdl handling code block.
+     * @return HTTP Response appropriate for the way {@code hdl} executed.
+     */
+    protected Response response(DataverseRequestHandler hdl) {
+        return response(hdl, null);
```
GPortas, line 646: This passes a null user to the [response(DataverseRequestHandler hdl, User user)] method (https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L629) which will create the DataverseRequest object with a null user, and the injected HttpServletRequest.

Until now I have always seen that the DataverseRequest class is instantiated with a user, even if it is GuestUser, in order to perform permission checks on the actions to be carried out on authenticated API endpoints.

Without the calling user, the DataverseRequest object would only contain the injected HttpServletRequest, which as far asI know its purpose there is to know the user IP address, which I'm not sure is useful information for unauthenticated operations like those in _Info_ (if I'm not mistaken, these address checks are only used in certain areas and are not common in Dataverse).

If there are no specific reasons that I am not aware of for persisting the HttpServletRequest data and therefore use the DataverseRequest class in unauthenticated scenarios, I would not create a DataverseRequest for unauthenticated operations, and would call the required logic directly like [here](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Info.java#L72).

I'm almost certain that null users are never passed when instantiating a DataverseRequest object. Maybe other people can confirm this. I also think that allowing null users can cause future accidental NPE issues.

GPortas: If you look at the method, which you mention is reintroduced, the one introduced is not the same as the previous one. Please take a look at the old one here: https://github.com/IQSS/dataverse/blob/v5.12/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L705

You can see then old method always passed a non-null user to the DataverseRequest object, coming from the old findUserOrDie method.

jp-tosca: >as far as I know its purpose there is to know the user IP address, which I'm not sure is useful information for unauthenticated operations like those in Info

Probably we can get more input about this on the next stand-up meeting. Right now, it is not used on these public endpoints, but should we keep it so we can have an IP for log/additional functionality on the future?  

> If you look at the method, which you mention is reintroduced, the one introduced is not the same as the previous one. Please take a look at the old one here

The new method is not intended to be the same than the older one, I added it as a way to expose the response method without the need of the user. My initial thought process was that this was not done previously because there were dependencies to `response(DataverseRequestHandler hdl)` in the past while the other APIs were migrated to the new functionality defined on `response(DataverseRequestHandler hdl, User user)`.

pdurbin: Alternatively, we could delete the new `response()` method and rewrite the three methods to not use it.

pdurbin: @GPortas @jp-tosca and I discussed this and decided to remove the new `response()` method in 8e69c6d.

pdurbin:
I ran the tests with the API runner and they passed: https://github.com/gdcc/api-test-runner/actions/runs/7021121336

PR #10133, opened 2023-11-21
Allows the dataverse administrator to set a dataverse user to respond to file access requests.

lwo:
Allows the dataverse administrator to set a dataverse user to respond to file access requests.

Sometimes employees to leave the organisation and no longer are able to approve access requests.
Or the organisation has entirely outsourced file approval tasks to a data officer.
In those cases, the organisational data manager needs to receive notifications.
This can be set with the DatasetManagerForFileAccessNotification set to the username of the 
datamanager's dataverse username.

Closes #

Default behavior: leave DatasetManagerForFileAccessNotification unset and there is no change for curent dataverse users.
Set DatasetManagerForFileAccessNotification to a dataverse username and make a file access request. Only this
user will then receive the access request.

No

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -3215,6 +3215,17 @@ This curl command...
 
 See also :doc:`oauth2`.
 
+.. _:DatasetManagerForFileAccessNotification
+
+:DatasetManagerForFileAccessNotification
+++++++++++++++++++++++++++++++++++++++++
+
+The identifier for the data manager's dataverse username. This is a dataverse account for the dataset manager of the organisation responsible
+for receiving file access requests. Use this in case owners of datasets are not responsible for approving access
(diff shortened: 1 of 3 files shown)


qqmyers:
FWIW: Right now, email is sent to everyone with ManageDatasetPermissions on the dataset. Could you manage this use case by setting the permission? I.e. remove the original author/whoever is no longer able to handle file access requests and add an admin account like your datasetmanager account to have that permission on those datasets (or assign that permission at the root collection level so it applies to all datasets). ManageDatasetPermissions is broader than just handling the file access request emails, but if someone has left, presumably they don't need permission to edit the dataset either. 

If they do need other access, it is still possible to just mute the file access request notification/email for that user (see https://guides.dataverse.org/en/latest/admin/user-administration.html#letting-users-manage-notifications - I'm not sure there's a way for an admin to change muting for a given user though).

scolapasta:

PR #10135, opened 2023-11-22
6630 lastlogintime not updated

### LINKED ISSUE #6630: lastlogintime - not updated for users when logged in via web
We were trying to use the authenticated users and list-user APIs to generate the list of all users in Dataverse. What I found is that the lastlogintime is not updated even after we login. I think this is a bug. 

Please see the details that I got when using the authenticated users API. (I have removed sensitive details from the following list).
  "id": 23,
  "identifier": "@venki",
  "displayName": "Kannadasan Venkatachalam",
  "emailLastConfirmed": "2017-07-18T09:58:00Z",
  "createdTime": "2017-07-18T09:58:00Z",
  "lastLoginTime": "2017-07-18T09:58:00Z",
  "lastApiUseTime": "2020-01-31T06:25:45Z",
  "authenticationProviderId": "shib"

This is my details and the lastLoginTime was updated in 2017 but the lastApiUseTime was updated on 31st Jan since I used the API.

NTU Dataverse is running on v4.16.

stevenferey:

Closes #6630 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Shib.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Shib.java
@@ -257,9 +257,9 @@ public class Shib implements java.io.Serializable {
                 return;
             }
             state = State.REGULAR_LOGIN_INTO_EXISTING_SHIB_ACCOUNT;
-            au = authSvc.updateLastLogin(au);
             logger.fine("Found user based on " + userPersistentId + ". Logging in.");
             logger.fine("Updating display info for " + au.getName());
+            authSvc.updateLastLogin(au);
             authSvc.updateAuthenticatedUser(au, displayInfo);
             logInUserAndSetShibAttributes(au);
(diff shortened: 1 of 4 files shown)


qqmyers (commented):
Looks like a straight-froward fix. The one question I have is whether it would be better to just include     
```
    @EJB
    UserServiceBean userService;
``` 
to be able to call userService.updateLastLogin rather than creating a pass-though method in the AuthenticationServiceBean. Not a major concern but I think just using @EJB would be cleaner unless there's a reason to avoid that.

stevenferey:
Hello @qqmyers ,

Yes you are right, the method was initially there to include other calls
I update the PR.

qqmyers (approved):
Looks good - thanks!

qqmyers (approved):
@stevenferey  - another one where I think all that's needed is merging with develop.

PR #10140, opened 2023-11-24
Avoiding slow bucket acl requests

ErykKul:
It is only a suggestion, up for discussion. As it turns out, each time an S3AccessIO object is constructed, a call is made to verify if the corresponding bucket exists (`s3.doesBucketExistV2(bucketName)`). This call uses internally `getBucketAcl(String)`, that, dependently on the s3 implementation, can be slow. For example, on our s3 the call takes longer time if the bucket contains more files. Since we keep adding files to our Dataverse, the response time grew to more than 500 miliseconds. This call is made for each dataset in the search results shown in a dataverse/collection page in the context of retrieving of the thumbnails. This happens for every dataset, regardless if the thumbnail is present or not, if not present, then thumbnail is not retrieved. Since there are 10 datasets by default in the dataverse page, 10 calls are made for a total time of 5 seconds, regardless if there are any thumbnails in these datasets or not. As a result, the dataverse page was very slow. Also, many other operations were slow, since this call is made very often. Eliminating this call made our dataverse installation much faster. We are also investigating if a caching can be enabled for these calls on the s3 side such that we won't have to rely on this pull request.

It is related to #9506 and to the pull request #9669

There might be a better way of addressing it, we are investigating it. Nevertheless, the check if the bucket exists looks not very usefull, since if it does not exist, retrieving files and other operations would fail anyway. Since the call takes long time in some cases, dropping it might be a good idea.

This issue is only relevant when the bucket ACL call takes long time, it might be not easy to test on a very fast s3 implementation. Alternatively, you can add `sleep` in the code to see the impact.

It makes our dataverse faster ;-)

No.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
+++ b/src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
@@ -110,7 +110,7 @@ public class S3AccessIO<T extends DvObject> extends StorageIO<T> {
             }
             //Not sure this is needed but moving it from the open method for now since it definitely doesn't need to run every time an object is opened.
             try {
-                if (bucketName == null || !s3.doesBucketExistV2(bucketName)) {
+                if (bucketName == null) {
                     throw new IOException("ERROR: S3AccessIO - You must create and configure a bucket before creating datasets.");
                 }
             } catch (SdkClientException sce) {


qqmyers:
FWIW - I think I have ~the same fix in #10004 which should go into 6.1.

ErykKul:
Great! I am closing this PR.

Issue #9916, opened 2023-09-14
Let contributors know why test are failing


pdurbin:
Currently we show them a red x with no details of which tests failed.

pdurbin:
> This is because https://jenkins.dataverse.org is behind a firewall.

The other day @donsizemore mentioned that when this server moves there's a chance we can open it up.

pdurbin:
@DS-INRAE @bencomp @johannes-darms @julian-schneider @lubitchv and everyone else, there's a new PR from @srmanda-cs to check out!

- #12368

pdurbin:
I just merged #12368!

Contributors, please try merging "develop" into your branch and tell me if you still can't tell why tests are failing! I hope you have enough permission. Also, we should probably follow up with some docs to explain where to look! 😅

ashmanda-cs:
Please feel free to tag me directly here or on Zulip if the test failures are not showing up as intended

pdurbin:
Yes! Here's the topic on Zulip we've been using: [#containers > goodbye-jenkins-you-did-an-amazing-job](https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/goodbye-jenkins-you-did-an-amazing-job/with/597906930)

PR #10141, opened 2023-11-25
Make superuser API call idempotent

### LINKED ISSUE #9887: idempotent API call for making a user a superuser
Hello! I like to report an issue in the docker installation process, User Administration does not exist. Issue related: https://github.com/IQSS/dataverse-docker/issues/78

Dataverse Docker installation
* When does this issue occur?
login with the admin account

* What happens?
The admin account works fine, but it doesn't have administration permissions. 

super user. 

login with admin and have admin permission.  (red color user)

v. 5.13 build 1244-79d6e57

https://github.com/IQSS/dataverse-docker/issues/78

In the thread, there are a solution: 

https://github.com/IQSS/dataverse-docker/issues/78#issuecomment-1549662669

mr-loop-1:
The api to make user a superuser is currently toggle based. The changes will make it a silent call.

Closes #9887 

`POST http://$SERVER/api/admin/superuser/$identifier` will make normal user a superuser but no effect on already superuser

No

Yes

**Additional documentation**:
Related issue https://github.com/IQSS/dataverse-docker/issues/78

### DIFF
--- a/doc/release-notes/9887-superuser-api-is-now-idempotent.md
+++ b/doc/release-notes/9887-superuser-api-is-now-idempotent.md
@@ -1,4 +1,3 @@
-The Admin api to make/remove a user as superuser (`POST http://$SERVER/api/admin/superuser/$identifier`
-), which was previously toggle based, is now a idempotent or silent.
+The Admin api to make/remove a user as superuser (`POST http://$SERVER/api/admin/superuser/$identifier`), which was previously toggle based, is now a idempotent or silent.
 The api is now only to make a user as superuser.
 Calling the api multiple times or on already superuser user has no effect.
--- a/doc/release-notes/9887-superuser-api-is-now-idempotent.md
+++ b/doc/release-notes/9887-superuser-api-is-now-idempotent.md
@@ -1,3 +1,4 @@
-The Admin api to make/remove a user as superuser (`POST http://$SERVER/api/admin/superuser/$identifier`), which was previously toggle based, is now idempotent or silent call.
-The api is now only to make user as superuser.
-Calling the api multiple times or on already superuser user has no effect.
+The admin api to make/remove a user as superuser (`POST http://$SERVER/api/admin/superuser/$identifier`), which was previously toggle based, is now idempotent or silent call.
+
+- The api is now only to make user as superuser.
+- Calling the api multiple times or on already superuser user has no effect.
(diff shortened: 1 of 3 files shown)


mr-loop-1:
Hi @pdurbin ,
I have a concern that with these changes, wouldn't the admin not be able to remove superuser access from a user. If there is a use case for making someone a superuser, then there conversely would be use cases to remove them.

mr-loop-1:
interestingly, I found an instance of this call here
https://github.com/IQSS/dataverse/blob/454e0bb9613e7984b5159fc37b6b6c85168a1959/src/test/java/edu/harvard/iq/dataverse/api/filesystem/FileRecordJobIT.java#L208-L211
where right before destroying a dataset, the superuser api is being called.
I investigated the destroy api and found this
https://github.com/IQSS/dataverse/blob/454e0bb9613e7984b5159fc37b6b6c85168a1959/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L359-L369

So, the user had to made superuser before calling the destroy. 
But before this PR, the superuser api was toggle, so I am not sure how this was supposedly working.

pdurbin (commented):
@mr-loop-1 thanks for the pull request!

First of all, it's awesome that all of our API tests are passing. I know that you can't see this (#9916), but for others: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10141/4/testReport/edu.harvard.iq.dataverse.api/

I guess this means that none of our tests removes superuser status. Otherwise, we'd see a failure, right?

We'll want to preserve the ability to remove superuser status. I'm sorry I wasn't clear about this before.

Would you be willing to iterate on the code some more? Or even create a new pull request if that's easier? Let's leave the existing toggle in place (mostly for backward compatibility, even if we don't exercise the "remove" functionality in our tests). And let's add a new API endpoint. I'm thinking one where we pass true or false with PUT. (See `putSetting` in Admin.java for an example.)

Does that make sense? Please feel free to ask questions here or in #dev at https://chat.dataverse.org

Thanks again! 🎉 ❤️ 🚀

mr-loop-1:
@pdurbin yes that's what I was thinking, that it would be a big change to remove that completely.

pdurbin:
@mr-loop-1 hi, just checking in. Need anything?

mr-loop-1:
@pdurbin i think this can be closed now since no longer needed

pdurbin:
@mr-loop-1  Oh, sorry, I meant do you need anything to move forward. The API itself is still needed. 😄 But no worries, I'll close this and hope another contributor will come along or that the issue will be prioritized for the core team to work on. Thanks!

PR #10142, opened 2023-11-27
Add validation for entered coordinate values in Geospatial metadata

### LINKED ISSUE #9547: Add validation for entered coordinate values in Geospatial metadata (bounding boxes, etc)
`TL;DR`: We allow users to enter junk on the geo metadata form. That ends poorly. 

This came in as a support request, somebody published a dataset, but it was not showing up on the collection page. Meaning, indexing failed, so according to the search engine, it was still an unpublished draft. 
On closer inspection, indexing was failing because it didn't like the geo bounding box coordinates: 
```
org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at ...
Error adding field 'geolocation'='ENVELOPE(- 81.06667,-75.19694,-1.363889,-5.015556)' 
msg=Unable to parse shape given formats "lat,lon", "x y" or as WKT because 
java.text.ParseException: java.lang.NumberFormatException: For input string: "-" input:
```
(note that the above was from server.log; zero diagnostic info back from the index api)

The only problem there was the space character in "`- 81.06667`". For this user it was addressed by fixing the field value in the database and reindexing. But we clearly need to validate the values that are being entered, on the edit form or via the api.

stevenwinship:
**What this PR does / why we need it**: Prevents the user input of invalid geospatial bounding box coordinates

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/9547

Closes #9547

**Suggestions on how to test this**: try adding non numbers/null values and values outside the norm for coordinates

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: error messages were added

**Additional documentation**:


landreev:
I checked in a quick change to the geospatial block tsv and .properties. Added extended labels for the bbox fields on the edit form as discussed. 
Also added some this bit of formatting for the *display* form. 
To show this: 
Screen Shot 2023-11-29 at 5 42 54 PM
instead of this:
Screen Shot 2023-11-29 at 5 46 34 PM

Also added a 2 sentence release note telling them to update the block.

landreev:
@stevenwinship and everybody
I retested extra carefully the scenario where there are some preexisting datasets with invalid bounding box values, and you try to reindex everything. Confirmed that these datasets are now getting indexed properly (minus the bad bboxes that are dropped), instead of failing quietly. 

We reindexed our prod. catalogue from scratch after the last deployment (6.0), with a new version of Solr installed (older versions were less picky about these bboxes being correctly defined). We discovered in the process that we had a few thousand automatically-generated datasets that had the North/South values swapped, and they all just quietly disappeared from the search catalog. 

I'm going to declare this PR merged and done for real now.

jggautier:
"Prod. catalogue" is Harvard Dataverse right, and the thousand automatically-generated datasets are datasets published in Harvard Dataverse? If so, what does it mean that datasets disappeared from the catalog? Sounds like they're not searchable anymore 😬

landreev:
@jggautier There was a very long slack thread about the whole disaster, I'll look it up. 

> "Prod. catalogue" is Harvard Dataverse right, and the thousand automatically-generated datasets are datasets published in Harvard Dataverse? If so, what does it mean that datasets disappeared from the catalog? Sounds like they're not searchable anymore 😬

jggautier:
Thanks.  And sorry; I try to keep up with the conversation. Sometimes things go over my head

PR #9509, opened 2023-04-11
removed copyFile method as it is not being used anywhere; used try-wi…

rtreacy:
…th-catch for handling autoclosable resources in determineFileType method and isGraphMLFile method; also fixed a couple of catch blocks in those methods to have minimal logging, rather than hiding the exceptions via empty catch blocks

**What this PR does / why we need it**:
Better handling of resources

**Which issue(s) this PR closes**

Closes #9312 

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #10144, opened 2023-11-27
Collection Quotas (8549)

### LINKED ISSUE #8549: Add mechanism for collection-wise storage size quotas
There is a number of *old* open issues with requests for storage quotas, in different scenarios: 

#4339 - requesting per-user quotas;
#3939 - requesting per-dataset quotas;
there's even an issue with a 3-digit number where collection-wise (well, "dataverse"-wise) quotas are mentioned: 
#938 

While I do believe we should consider all of the above, I am opening this narrowly-defined issue specifically for collection quotas, since this is becoming increasingly important. 

Perhaps there should be the installation-wide size default. With a mechanism for assigning custom limits to specific collections. In other words this should be handled the same way custom stores are handled now. At a minimum, this could be another storage option in the collection settings (available to the superusers only). Or we could add a dashboard UI.

landreev:
Please note the draft nature of this PR. 
I'm still working, on the tests and extra documentation, etc. But making a draft PR because I want to start the process of reviewing the implementation approach. 

Closes #8549

I'd like to double-check/have a second and third opinions etc. on the viability of this approach as described below. 

The main idea behind the implementation is to maintain a real-time record of storage use for all the DvObjectContainers (Datasets and Collections), since we cannot possibly afford to calculate it on the fly for every file upload. We record storage use for the entire hierarchy of nested objects. I.e., adding a file to the dataset increments the storage use for the parent dataset, and all the parent collections up to the root. 

This information is stored in its own dedicated table, StorageUse. As opposed to storing it as an extra column in the DvObject table, since that would require updating all the parent DvObjects every time we upload a file, likely causing merge conflicts on a busy instance/large collections. 

These updates are performed via native recursive queries in new transactions, avoiding any cascading updates. 

These recursive updates need to be performed not just every time a new file is uploaded, but every time a file is successfully ingested (this adds the size of the archival tab-delimited file to the total), when an unpublished file is deleted and when a tabular file is un-ingested. 

I'm only counting the sizes of the main datafiles, and the produced tab-delimited versions for the ingested tabular files for these purposes. I.e., I consider these to be the archival "payload" of datasets and collections, as opposed to all the file- and dataset-level auxiliary files, such as the resized image thumbnails and metadata exports, under the assumption that those are transitive, i.e., can be deleted and automatically re-generated. If necessary, we can extend the scheme to also count the sizes of (some, select?) auxiliaries going forward.

I chose to have this system of keeping the real-time record of storage use automatically enabled, regardless of whether quota enforcement is configured on the instance. With the rationale being that this information can be very useful regardless. (The existing commands for calculating storage sizes perform this in real time and are thus less efficient). Also, the idea is that these updates via direct native queries should be fairly efficient and not result in any serious overhead. A somewhat bulky flyway script is provided to populate this map for the existing objects. I tested it on a copy of the prod. db and the performance of the current version is adequate. But, like with everything else in the PR, I would appreciate another developer taking a quick look at it. 

I'm assuming there may be cases when these recorded totals get out of sync with the real storage allocation. My idea is to assume that this may happen, and to provide an API that would recalculate the values for an affected DvObjectContainer as a remedy. 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -139,39 +143,6 @@ public class DataFileServiceBean implements java.io.Serializable {
      */
     public static final String MIME_TYPE_PACKAGE_FILE = "application/vnd.dataverse.file-package";
     
-    public class UserStorageQuota {
```
landreev, line 142: This class used during the phase one/proof of concept development.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -225,224 +244,277 @@ public List<DataFile> saveAndAddFilesToDataset(DatasetVersion version,
...
+                        logger.warning("file size over quota limit, skipping");
+                        // @todo: we need to figure out how to better communicate
+                        // this (potentially partial) failure to the user.  
+                        //throw new FileExceedsStorageQuotaException(MessageFormat.format(BundleUtil.getStringFromBundle("file.addreplace.error.quota_exceeded"), bytesToHumanReadable(confirmedFileSize), bytesToHumanReadable(storageQuotaLimit)));
+                    } else {
+                        // Update storage use for all the parent dvobjects: 
+                        // @todo: Do we want to do this after after *each* file is saved? - there may be 
+                        // quite a few files being saved here all at once. We could alternatively
+                        // perform this update only once, after this loop is completed (are there any
+                        // risks/accuracy loss?)
```
pdurbin, line 388: Good that we're thinking about the "many files" case.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -225,224 +244,277 @@ public List<DataFile> saveAndAddFilesToDataset(DatasetVersion version,
...
-						 */
-						dataAccess.savePath(tempLocationPath);
-
-						// Set filesize in bytes
-						//
-						dataFile.setFilesize(dataAccess.getSize());
-						savedSuccess = true;
-						logger.fine("Success: permanently saved file " + dataFile.getFileMetadata().getLabel());
-
-                                            // TODO: reformat this file to remove the many tabs added in cc08330
```
pdurbin, line 247: Oh good, tabs are being removed, I think, I hope.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -754,6 +754,31 @@ The following attributes are supported:
 * ``affiliation`` Affiliation
 * ``filePIDsEnabled`` ("true" or "false") Restricted to use by superusers and only when the :ref:`:AllowEnablingFilePIDsPerCollection <:AllowEnablingFilePIDsPerCollection>` setting is true. Enables or disables registration of file-level PIDs in datasets within the collection (overriding the instance-wide setting).
 
+.. _collection-storage-quotas:
+  
+Collection Storage Quotas
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
pdurbin, line 760: Any UI impact? Should we add to the long, dynamic list of rules? Or defer until we're off JSF? Screenshot below:

Screenshot 2023-11-28 at 12 02 31 PM

pdurbin (commented):
First pass review. Overall, looks good. I hope performance is good enough! It'll be a great feature, one that's often asked for!

doc/sphinx-guides/source/api/native-api.rst
landreev: Yes, the remaining quota info will be shown to the user on the file upload page, above. The code for that is already in the develop branch, it was added in the proof of concept/phase one pr (#9509).

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
landreev: Yes, this is still one of the work-in-progress/still experimenting with parts. I may have a better solution in the works).

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -754,6 +754,31 @@ The following attributes are supported:
...
+
+.. code-block:: 
+
+  curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/dataverses/$ID/storage/quota"
+
+Will output the storage quota allocated (in bytes), or a message indicating that the quota is not defined for the collection.
```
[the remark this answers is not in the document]
landreev: Yes, there's definitely going to be more documentation, and more tests. I'm still working on these parts, and this is the main reason this is still a draft PR. But I really wanted other developers to look at the mechanics of the implementation.

doc/release-notes/8549-collection-quotas.md
```diff
@@ -0,0 +1,3 @@
+This release adds support for defining storage size quotas for collections. Please see the API guide for details. This is an experimental feature that has not yet been used in production on any real life Dataverse instance, but we are planning to try it out at Harvard/IQSS.
+Please note that this release includes a database update (via a Flyway script) that will calculate the storage sizes of all the existing datasets and collections on the first deployment. On a large production database with tens of thousands of datasets this may add a couple of extra minutes to the deployments.
```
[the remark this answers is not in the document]
landreev: I committed your typo fix, but I will also add a clarifying adjective - "to the *first* depolyment" or "to the *initial* deployment".

doc/sphinx-guides/source/api/native-api.rst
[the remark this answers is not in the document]
scolapasta: I think (correct me if I'm wrong - that that comment is less about a default global quota, and more that since collections can contain collection you have to go up the tree.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1396,28 +1367,38 @@ public Embargo findEmbargo(Long id) {
...
+     * ancestor Collections further up have a configured quota. If it finds one, 
+     * it will retrieve the current total content size for that specific ancestor 
+     * dvObjectContainer and use it to define the quota limit for the upload
+     * session in progress. 
+     * 
+     * @param parent - DvObjectContainer, Dataset or Collection
+     * @return upload session size limit spec, or null if quota not defined on 
+     * any of the ancestor DvObjectContainers
+     */
+    public UploadSessionQuotaLimit getUploadSessionQuotaLimit(DvObjectContainer parent) {
```
scolapasta, line 1383: is it possible to define a quote of say 100 GB for a collection, and a larger quote for a child collection (say 1 TB)? in this case, does the child size effectively override the parent?

src/main/java/edu/harvard/iq/dataverse/DvObject.java
```diff
@@ -467,4 +484,14 @@ public void setStorageIdentifier(String storageIdentifier) {
...
     @OneToMany(mappedBy = "definitionPoint",cascade={ CascadeType.REMOVE, CascadeType.MERGE,CascadeType.PERSIST}, orphanRemoval=true)
     List<RoleAssignment> roleAssignments;
+    
+    /**
+     * Should only be used in constructors for DvObjectContainers (Datasets and 
+     * Collections), to make sure new entries are created and persisted in the 
+     * database StorageUse table for every DvObject container we create.
+     * @param storageUse 
+     */
+    public void setStorageUse(StorageUse storageUse) {
```
scolapasta, line 494: if only possible for Dataverse and Dataset, is it possible to move this to DvObjectContainer? Or is that not possible for the db table to be properly created / connected to? Partially asking because I see the imports in DvObjectContaine, but no other references, so maybe you tried it there, but had to move it here? (if so, we should clean up those imports).

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateNewDataFilesCommand.java
```diff
@@ -696,10 +712,19 @@ public CreateDataFileResult execute(CommandContext ctxt) throws CommandException
...
 
-            // Update quota (may not be necessary in the context of direct upload - ?)
+            // Update the quota definition for the *current upload session*
+            // This is relevant for the uploads going through the UI page 
+            // (where there may be an appreciable amount of time between the user
+            // uploading the files and clicking "save". The file size should be 
+            // available here for both direct and local uploads via the UI. 
+            // It is not yet available if this is direct-via-API - but 
+            // for API uploads the quota check will be enforced during the final 
+            // save.
```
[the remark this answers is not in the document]
scolapasta: presumably it always happens in final save, and this here is just redundant, no? If not, I could imagine that a crafty user could exceed the quota by doing some properly timed concurrent uploads?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteCollectionQuotaCommand.java
```diff
@@ -0,0 +1,49 @@
...
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+import edu.harvard.iq.dataverse.engine.command.exception.IllegalCommandException;
+import edu.harvard.iq.dataverse.storageuse.StorageQuota;
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author landreev
+ */
+@RequiredPermissions(Permission.ManageDataversePermissions)
```
scolapasta, line 22: wondering about permissions, here and generally - i.e. these should be superuser, no? otherwise anyone with a collection could just change or remove their quota?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteCollectionQuotaCommand.java
```diff
@@ -0,0 +1,49 @@
...
+    public void executeImpl(CommandContext ctxt) throws CommandException {
+        if (targetDataverse == null) {
+            throw new IllegalCommandException("", this);
+        }
+        
+        StorageQuota storageQuota = targetDataverse.getStorageQuota();
+        
+        if (storageQuota != null && storageQuota.getAllocation() != null) {
+            storageQuota.setAllocation(null);
+            ctxt.em().merge(storageQuota);
```
scolapasta, line 44: normally, I *think* we tried to have the calls to the entity manager in the ejbs and not the commands directly? though I don't think that's critical, and it's possibly we've broken that practice multiple times. But something to consider, at least. (also I could be wrong... :))

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetCollectionStorageUseCommand.java
```diff
@@ -0,0 +1,49 @@
...
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author landreev
+ */
+@RequiredPermissions(Permission.ManageDataversePermissions)
+// alternatively, we could make it dynamic - public for published collections
+// and Permission.ViewUnpublishedDataverse required otherwise (?)
```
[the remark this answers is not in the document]
scolapasta: I'd tentatively agree, since users can contribute who can't manage permissions and may want to have a sense of the quota?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SetCollectionQuotaCommand.java
```diff
@@ -0,0 +1,62 @@
...
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+import edu.harvard.iq.dataverse.engine.command.exception.IllegalCommandException;
+import edu.harvard.iq.dataverse.storageuse.StorageQuota;
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author landreev
+ */
+@RequiredPermissions(Permission.ManageDataversePermissions)
```
scolapasta, line 23: same comment as with delete (superuser only?)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SetCollectionQuotaCommand.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+        
+        if (allocation == null) {
+            throw new IllegalCommandException("", this);
+        }
+        
+        StorageQuota storageQuota = dataverse.getStorageQuota();
+        
+        if (storageQuota != null) {
+            storageQuota.setAllocation(allocation);
```
scolapasta, line 51: and save related to entity manager and ejb save()...

scolapasta (commented):
Overall looks great! I left several comments, mostly questions that are likely straightforward / just want confirmation. A couple of clean up suggestions based on what I think our practices have been. 

two medium questions:
 • left comments on hte commands, but shouldn't the permissions be superuser only for setting and deleting? I don't think we want collection admins to be able to alter what we set up for them.
• so does this replace what was there then for (not used,, I know) user quotas - I saw several places where there seemed to be some straghtforward replacing and figure it;s just easoer to ask here.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
landreev: Yes, I should have mentioned this in the PR, when describing my implementation (will add): the quota defined on the nearest ancestor applies, if any, in all cases. 
Answering your specific question, yes, it is possible to give a child collection a quota larger than the parent's, which it will override. 

But the way my implementation works, once the child collection goes over the quota limit defined for the parent,  that will set put the parent's storage use over their limit, so nobody will be able to add any content to the datasets immediately under the parent, or in any other subcollections thereof that don't have their own quota defined. So I should be telling site admins not to do this in the guide. 

(Is this a problem?)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteCollectionQuotaCommand.java
landreev: On the previous question - not entirely redundant, no; in part because in the context of uploading through the UI there may be some appreciable time between uploading and deciding to click 'save', yes. Performing the check during the final save only doesn't seem perfect either, since it just feels wrong to allow the user to complete that UI upload, when they are clearly going over. 

On the question of the permissions, yes, of course, that's clearly wrong - should be readable to the owners, but setting/unsetting should be admin only... just haven't gotten to fixing - will do shortly.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
landreev: ... Maybe not telling them "not to do this", but rather, spell it out that you only want to do this if you want to specifically make a collection read-only/closed for more uploads, while keeping one or more of the sub-collections open for business - ?

landreev: > so does this replace what was there then for (not used,, I know) user quotas - I saw several places where there seemed to be some straghtforward replacing and figure it;s just easoer to ask here.

It uses/builds on top of the mechanics of quota enforcement added in that phase one PR. The object that was called "UserStorageQuota" is still being used, but it was renamed to UploadSessionQuota... to better reflect what it does (it was moved out of DataFileServiceBean also). The only way to actually define a quota in that phase one pr was to set an instance-wide limit, which then applied to all users; AND "applied" in a sense of counting against the sum of filesizes by `dvobject.creator_id`. A limit defined like that had no practical value obviously; it was only used to test the mechanisms of enforcement/the UI etc. It was removed from this branch. 
Does this make sense?

landreev: I should add that, if we want to add real per-user quotas going forward, I even left a commented out relation to AuthenticatedUser in the new StorageQuota class. So, the same entity can be used for that as well (but we would need to figure out how/what to count as storage use by specific users). There's also a comment saying that nothing should be stopping us from changing the DvObject relation there to `@ManyToOne`, and then be able to define quotas on specific collections/datasets AND for specific users... but I don't know if we ever want to go there.

landreev:
Writing the Admin Guide entry now. I'm satisfied with the state of the PR aside from that. 
I have some concerns and ideas (will share), but I believe they can be tested and addressed in QA.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateNewDataFilesCommand.java
```diff
@@ -696,10 +712,19 @@ public CreateDataFileResult execute(CommandContext ctxt) throws CommandException
...
+            // Update the quota definition for the *current upload session*
+            // This is relevant for the uploads going through the UI page 
+            // (where there may be an appreciable amount of time between the user
+            // uploading the files and clicking "save". The file size should be 
+            // available here for both direct and local uploads via the UI. 
+            // It is not yet available if this is direct-via-API - but 
+            // for API uploads the quota check will be enforced during the final 
+            // save. 
             if (fileSize > 0 && quota != null) {
+                logger.info("Setting total usage in bytes to " + (quota.getTotalUsageInBytes() + fileSize));
```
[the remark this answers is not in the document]
landreev: Sure, but I usually leave stuff like this in for QA, then change to fine before we merge.

src/main/java/edu/harvard/iq/dataverse/DvObject.java
landreev: To be clear, there was actually a reason I ended up putting it in DvObject - because the column in StorageUse needs an entity to point to, and DvObjectContainer is not an entity... But it was still possible to move the above to DvObjectContainer, I just had to keep this in StorageUse: 
```
@OneToOne
@JoinColumn(nullable=false)
private DvObject dvObjectContainer; 
```

scolapasta (approved):

stevenwinship:
I'm doing the QA now. What if anything should happen if the quota is set to a value less than the current size of the existing data?

PR #10146, opened 2023-11-27
9686 move harvesting client id

### LINKED ISSUE #9686: Move harvestingclient_id from the dataset to dvobject and use it directly for files
**Description**
- Move the column `harvestingclient_id` from the dataset table in the database to the common `dvobject`. 
- This decision arrived at during a tech hour where it was decided not to redesign the concept of a harvested file, but instead make it easier to distinguish between harvested vs. 'real' files. See comment: 
https://github.com/IQSS/dataverse/issues/8629#issuecomment-1585099765

- Closes spike issue: https://github.com/IQSS/dataverse/issues/8629

sekmiller:
**What this PR does / why we need it**: Moves harvesting client id from the dataset table to the dvobject table to make it easier to distinguish between harvested vs. 'real' files.

Closes #9686 Move the harvesting client id from the dataset table to the dvobject table.

**Special notes for your reviewer**: Pretty straight forward. Additional changes made in the Metrics service bean to update custom queries that relied on dataset.harvestingclient_id. Also the migration script includes a drop column for the harvesting client id from the dataset table. It's currently commented out because reluctant to lose that data just in case something goes wrong in the upgrade. Do we ever just have a the drop column update happen in a subsequent release or is there some way to make it "safer" to drop the column in a single upgrade?

**Suggestions on how to test this**: Make sure that harvesting still works as expected and take a look at the results from the Metrics API

**Is there a release notes update needed for this change?**: probably not

**Additional documentation**:


src/main/resources/db/migration/V6.0.0.5__9686-move-harvestingclient-id.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE dvobject ADD COLUMN IF NOT EXISTS harvestingclient_id BIGINT;
+
+update dvobject dvo set harvestingclient_id = s.harvestingclient_id from
+(select id, harvestingclient_id from dataset d where d.harvestingclient_id is not null) s
+where s.id = dvo.id; 
+
+--ALTER TABLE dataset drop COLUMN IF EXISTS harvestingclient_id;
```
scolapasta, line 7: the query looks fine and we will have fully tested (and have backups); this can be uncommented back in

src/test/java/edu/harvard/iq/dataverse/api/MetricsIT.java
```diff
@@ -30,7 +30,7 @@ public static void cleanUpClass() {
 
     @Test
     public void testGetDataversesToMonth() {
-        String yyyymm = "2018-04";
+        String yyyymm = "2023-11";
```
scolapasta, line 33: if it's important for these dates to be current, should we have it instead get the values from Date.now()?

scolapasta (changes requested):
Looks good, go ahead and uncomment that script line (and see my other minor comment)

src/test/java/edu/harvard/iq/dataverse/api/MetricsIT.java
sekmiller: Yeah, I was just being a little lazy - trying to get somewhat meaningful results without changing too much. I fixed it.

scolapasta:
After review, we'd like the scope of the solution to include the changes to harvested files.

PR #10159, opened 2023-12-01
Fix Datasets API getVersionFiles endpoint content type filtering (JPA Criteria)

GPortas:

DatasetsIT.getVersionFiles fails intermittently as it sometimes finds out unordered files when sorting by content type.

A second orderBy condition has been added to force that when sorting by content type, files are also sorted by name.

New order example:

test_5.png
test_1.txt
test_2.txt
test_3.txt

Get dataset files via curl using contentType query param:

`curl "https://demo.dataverse.org/api/datasets/24/versions/1.0/files?contentType=image/png"`

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
@@ -260,22 +260,27 @@ public class DatasetVersionFilesServiceBean implements Serializable {
         return criteriaBuilder.and(predicates.toArray(new Predicate[]{}));
     }
 
-    private Order createGetFileMetadatasOrder(CriteriaBuilder criteriaBuilder,
-                                              FileOrderCriteria orderCriteria,
-                                              Root<FileMetadata> fileMetadataRoot) {
+    private List<Order> createGetFileMetadatasOrder(CriteriaBuilder criteriaBuilder,
+                                                    FileOrderCriteria orderCriteria,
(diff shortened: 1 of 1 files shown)


landreev:
@GPortas May be a stupid question, but why not add "by id" as the final clause, to make it really bulletproof? - Or do we ever call that method on files that are not yet saved in the database? 

Also, should the same additional "by label" clause be added to the "by size" criteria?

landreev:
The integration tests passed passed successfully on Jenkins. And I ran `mvn test -Dtest=DatasetsIT#getVersionFiles` on my own build a few times in a row. Merging. 

The questions I asked about making the sort order even more consistent should be considered out of scope. I don't know whether there is a use case where this is specifically important, that the files in the same version are always listed in the same exact order. 
But if such case exists, we need to keep in mind that this is not going to be true necessarily, neither for the current "by size" nor for "by type" or "by label" criteria. As we allow non-unique file names ("labels") in some situations - when they are in different folders.

pdurbin:
Thanks for fixing and merging. I deleted the branch.

GPortas:
@landreev What you say about adding "by label" to "by size criteria" makes sense. Even adding "by id" as the final clause would make sense too, to make it really bulletproof. 

I didn't add "by label" to "by size criteria" since I found files of the same size to be less common than the same type (It is quite common to have many files of the same type).

This isn't really based on any sorting requirements for the files tab, so I'm not sure if it makes sense to spend more time on this, but we can discuss more about it at the frontend weekly meeting.

PR #10174, opened 2023-12-11
8549 collection quotas

landreev:

This is a followup PR to #10144 - a small refinement I remembered to add after the fact. 
It's for counting the sizes of any new "registered aux. files" - like the ones that an external tool like OpenDP would deposit, etc.; it makes sense to consider these real data payload, unlike the throwaway aux. files like exports and thumbnails. 
This was discussed on slack, outside of the issue/PR, so slipped through the cracks. 

If it's too late to merge, it's probably just fine. It's extremely unlikely that these would ever amount to a statistically important amount of storage on any instance actively using aux. files. 

A simple and quick way to test this would be to upload the netcdf file from our source tree `src/test/resources/netcdf/ICOADS_R3.0.0_1662-10.nc` and check that the storage use for the parent collection got incremented by a few bytes (via `curl -H "X-Dataverse-key: xxx" http://localhost:8080/api/dataverses/<COLLECTIONALIAS>/storage/use` or a db query). The special treatment of NetCDF is one scenario in our own prod. where registered aux. files are created. 

**Additional documentation**:


qqmyers (approved):

landreev:
I waited for the Jenkins test to pass. I tested it some too. Will just go ahead and merge it - ?

Issue #93, opened 2014-07-09
File "thumbnails" - preview/summary functionality for the datafile display on the dataset page


eaquigley:
---

Author Name: **Leonid Andreev** (@landreev)
Original Redmine Issue: 3502, https://redmine.hmdc.harvard.edu/issues/3502
Original Date: 2014-02-03
Original Assignee: Elda Sotiri

---

In 4.0 we want to provide a more elaborate preview functionality for various types of files - in addition to the thumbnails we provide for graphic images in 3.*. 

The following types of preview are being considered so far: 

Image thumbnail (essentially unchanged from 3.6; optimization functionality - currently, imagemagick - needs to be reviewed and possible alternatives considered);

Data file summary - depends on the "summary generator" software James and Vito are working on. For the 4.0a: some generic thumbnail representing a tabular file will be used. 

Fits file: similarly, some generic FITS file icon will be used; the existing 3.\* summary will be shown. Eventually, we will investigate generating a preview image somehow summarizing the contents of a FITS file (?)

PR #10184, opened 2023-12-12
bump DuraCloud to 8.0.0

donsizemore:

See dataverse-security #93

Closes dataverse-security #93

Sorry to bump a major version, but 7.1.4 didn't address these.

Duracloud is only used by TDL. They have been alerted that this PR is getting merged and may impact their Duracloud archiving. Since IQSS doesn't have a Duracloud account/setup to test this, just regression testing is probably all that is needed.

No

No

**Additional documentation**:

None

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -466,7 +466,7 @@
         <dependency>
             <groupId>org.duracloud</groupId>
             <artifactId>common</artifactId>
-            <version>7.1.1</version>
+            <version>8.0.0</version>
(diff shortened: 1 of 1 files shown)


qqmyers (approved):
Seems OK to me. At this point only TDL/@CCMumma are set up to test the DuraCloud archiver, so I think in general we should keep up to date and alert TDL when changes are made that might break that archiver.

PR #10191, opened 2023-12-14
return deaccessioned version of dataset in more cases

### LINKED ISSUE #10164: API only returns Deaccessioned Dataset Version if user has edit privileges on the version
1. Create a deaccessioned version
2. from the API, try to get the version using another account that doesn't have edit privileges on that version:

http://localhost:8000/api/v1/datasets/:persistentId/versions/1.0?persistentId=doi:10.5072/FK2/VV0A0P&includeDeaccessioned=true&includeFiles=false

3. The API returns an error message instead of the dataset version
{"status":"ERROR","message":"Dataset version 1.0 of dataset 2 not found"}

* When does this issue occur?
Using API from SPA to view Deaccessioned version

SPA Dataset View Page

* What happens?
It shows the latest published page, or page not found

all

Return the Dataset Version

The latest unstable version from Docker Hub:
https://hub.docker.com/layers/gdcc/dataverse/unstable/images/sha256-d552309321453989fc55ddabe12c8ff996e039db77738353fe15246dc85c6aca?context=explore

no

**Screenshots:**

Screenshot 2023-12-02 at 4 57 01 PM
Screenshot 2023-12-02 at 4 55 32 PM

jp-tosca:
API only returns Deaccessioned Dataset Version if user has edit privileges

Closes #10164 

We agreed to change the parameter name from includeFiles with excludeFiles, this is the default and the lack of the parameter will default to false so files are expected to be included by default if this parameter is not provided.

`getVersionFiles`, `getVersionFileCounts` and `getDownloadSize` act exactly as before, this check would execute every time includeDeaccessioned was set to true, `getDatasetVersionCitation` will skip this test when includeDeaccessioned is true and `getVersion` will check for when includeDeaccessioned is true and excludeFiles is false. This change also replaced includeFiles with excludeFiles which is the default behaviour. 

This change required to update Datasets.getDatasetVersionOrDie which impacted the following methods:

- **getVersion `{id}/versions/{versionId`:**  Will check file permissions when files when excludeFiles and includeDeaccessioned are true. 
- **getVersionFiles `{id}/versions/{versionId}/files`:**  Will check file permissions when includeDeaccessioned is true. 
- **getVersionFileCounts `{id}/versions/{versionId}/files/counts`:** Will check file permissions when includeDeaccessioned is true.
- **getDownloadSize `{identifier}/versions/{versionId}/downloadsize`:** Will check file permissions when includeDeaccessioned is true.
- **getDatasetVersionCitation `{id}/versions/{versionId}/citation`:** Won't check file permissions.
- **listVersions `{id}/versions`:** replaced includeFiles with excludeFiles which is the default.

Have a Dataset with one deaccessioned version, here is the test matrix that I used and a flow chart that i made:

Deaccesioned   dataset | Include   Deacessioned | Exclude files | User can edit | Result
-- | -- | -- | -- | --
N | Y | Y | Y | OK
N | Y | Y | N | OK
N | Y | N | Y | OK
N | Y | N | N | OK
N | N | Y | Y | OK
N | N | Y | N | OK
N | N | N | Y | OK
N | N | N | N | OK
Y | Y | N | Y | OK
Y | Y | N | N | ERR
Y | Y | Y | Y | OK
Y | Y | Y | N | OK
Y | N | N | Y | ERR
Y | N | N | N | ERR
Y | N | Y | Y | ERR
Y | N | Y | N | ERR

image


doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -5,25 +5,26 @@ API Changelog
     :local:
     :depth: 1
 
-v6.1
-----
+6.2
```
pdurbin, line 8: I put these "v"s in here for a reason. The anchor links don't look right otherwise.

doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -5,25 +5,26 @@ API Changelog
...
-New
-~~~
-- **/api/dataverses/{id}/datasetSchema**: See :ref:`get-dataset-json-schema`.
-- **/api/dataverses/{id}/validateDatasetJson**: See :ref:`validate-dataset-json`.
+6.1
+---
 
 New
 ~~~
 - **/api/admin/clearThumbnailFailureFlag**: See :ref:`thumbnail_reset`.
```
pdurbin, line 20: Why is there any "New" at all in this file? There shouldn't be. We took it out: https://guides.dataverse.org/en/6.1/api/changelog.html

Does "develop" need to be merged into this branch?

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -718,11 +703,83 @@ public void testDatasetVersionsAPI() {
...
+            then().assertThat().statusCode(404);
+        excludeFiles = false;
+        UtilIT.getDatasetVersion(datasetPid, DS_VERSION_LATEST_PUBLISHED, apiToken, excludeFiles, includeDeaccessioned).
+            then().assertThat().statusCode(404);
+        UtilIT.getDatasetVersion(datasetPid, DS_VERSION_LATEST_PUBLISHED, apiTokenNoPerms, excludeFiles, includeDeaccessioned).
+            then().assertThat().statusCode(404);
+
+      
+
+
```
pdurbin, line 776: While we're testing deaccessioning, would it be possible to add some tests asserting search behavior? Please see https://github.com/IQSS/dataverse/issues/10164#issuecomment-1854151124

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -1881,6 +1881,9 @@ static Response getDatasetVersions(String idOrPersistentId, String apiToken, Int
             requestSpecification = given()
                     .header(UtilIT.API_TOKEN_HTTP_HEADER, apiToken);
         }
+        String io = "/api/datasets/" + idInPath + "/versions" + optionalQueryParam;
+        System.out.println(io);
```
pdurbin, line 1885: Do we need to keep these debugging lines?

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -668,13 +644,22 @@ public void testDatasetVersionsAPI() {
...
         unpublishedDraft.then().assertThat()
                 .body("data.files", equalTo(null))
                 .statusCode(OK.getStatusCode());
 
+        excludeFiles = false; 
+        unpublishedDraft = UtilIT.getDatasetVersion(datasetPid, DS_VERSION_DRAFT, apiToken, excludeFiles, false);
+        unpublishedDraft.prettyPrint();
+        unpublishedDraft.then().assertThat()
+                .body("data.files.size()", equalTo(1))
+                .statusCode(OK.getStatusCode());
```
pdurbin, line 659: statusCode should be checked before body.

Otherwise you don't get a meaningful error when statusCode is an unexpected value.

Basically, statusCode should always be the first line.

sekmiller:
@landreev can you hear me now?

pdurbin (commented):
I haven't wrapped my head around the substantive change in this PR but I'm leaving comments about docs and tests, at least.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
jp-tosca: I need to remove those, thanks!

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
jp-tosca: I am fixing this, thanks! I am also going to fix it on other places since this was code I copied. 😓

doc/sphinx-guides/source/api/changelog.rst
jp-tosca: The file was just updated with the latest version and changes 😄

doc/sphinx-guides/source/api/changelog.rst
jp-tosca: The file was just updated with the latest version and changes 😄

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
jp-tosca: I have removed those, thanks!

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
jp-tosca: I into this but this should address the issue with the API

jp-tosca:
I would still like to check with @ekraffmiller about the citation API, on dev branch I was able to access to the citation of a deaccessioned dataset when the parameter is provided, but I found a bug(?) that this parameter will also give the citation of a draft.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
jp-tosca: I did a quick dive into the Search API. It seems to be unrelated to this one and if I get the question right, they want to use the search API to include a deaccessioned dataset but I can't see any parameters similar to `includeDeaccessioned`, probably would be a feature request but still not sure.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2770,10 +2770,11 @@ public static <T> T handleVersion(String versionId, DsVersionHandler<T> hdl)
     }
 
     private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers) throws WrappedResponse {
-        return getDatasetVersionOrDie(req, versionNumber, ds, uriInfo, headers, false);
+        //The checkPerms was added to check the permissions ONLY when the dataset is deaccessioned.
+        return getDatasetVersionOrDie(req, versionNumber, ds, uriInfo, headers, false, false);
```
qqmyers, line 2774: I'd suggest making true (the safer option) the default, which would also avoid you having to add true on the methods that you aren't otherwise changing, e.g. line 577.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2770,10 +2770,11 @@ public static <T> T handleVersion(String versionId, DsVersionHandler<T> hdl)
...
     }
 
-    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers, boolean includeDeaccessioned) throws WrappedResponse {
+    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers, boolean includeDeaccessioned, boolean checkPerms) throws WrappedResponse {
```
qqmyers, line 2777: How about checkPermsOnDeaccessioned /something more descriptive. This doesn't shut off all permission checks, only when deciding whether to include a deaccessioned version.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4026,7 +4027,7 @@ public Response deaccessionDataset(@Context ContainerRequestContext crc, @PathPa
             return badRequest(BundleUtil.getStringFromBundle("datasets.api.deaccessionDataset.invalid.version.identifier.error", List.of(DS_VERSION_LATEST_PUBLISHED)));
         }
         return response(req -> {
-            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers, false);
+            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers, false, false);
```
qqmyers, line 4030: FWIW: This shouldn't ever need a checkPerms when deaccessioned false - because includeDeaccessioned is false. Might be clearer if this were just getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers);

qqmyers (changes requested):
A couple minor suggestions for changes. Overall - looks good. Very clear testing instructions.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: That sounds like a good change. I am just afraid that if I default it to true, someone will use this method and this bug could happen again. 

As an example for `{id}/versions/{versionId}/citation` on line `4023`; it needs to be set to false. What do you think? I sill can create a default signature that doesn't require and defaults it to true, I agree it looks cleaner but still we need the exception in this case. 

I have the change as you suggested ready but wanted to double check. ☕

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: Good catch! 🐟

pdurbin (commented):
@jp-tosca can you please resolve merge conflicts?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: I pushed these changes, probably will add something on the comments just so ppl (if they read the comments 😆 ) know they have to be careful about using the default.

jp-tosca: I pushed more changes with the comments. I think this should be enough. 😃

qqmyers: It's pretty limited in that it only affects deaccessioned versions to begin with, but my thought was to set the default so that someone wouldn't accidentally make info public when it shouldn't be (versus -as you say - make it so that someone might accidentally block a subset of info about a deaccessioned version that should be public).

qqmyers (approved):
Looks good - just the merge issues to fix.

jp-tosca:
Merge conflicts have been solved. 🥳

PR #10162, opened 2023-12-01
Globus Store implementation

qqmyers:
**What this PR does / why we need it**: Implements new Globus functionality to handle

- use of a standard Globus endpoint managed by dataverse
- referencing files in remote Globus endpoints, and
- supporting use of an S3 store with the Globus S3 Connector (which existed before) as a store rather than a one-per-instance option (with better security)

The overall functionality requires use of the [Borealis Dataverse-Globus app](https://github.com/scholarsportal/dataverse-globus) which is being updated to work with the functionality added in this PR.
 
**Which issue(s) this PR closes**:

Closes #9123

**Special notes for your reviewer**:
~code complete - working on handling guestbook-at-download with Globus, some docs but more to follow. [External doc](https://docs.google.com/document/d/15ARPROjMoCM6E1kK7Vrqr3PdYe0Ia2oamo3dRS84lCY/edit?usp=sharing) includes more info.
**Suggestions on how to test this**:
Without the app, testing requires some manual steps. Basically, Dataverse launches the app like an external tool to support upload (transfer to Dataverse) or download (transfer from Dataverse), so to test one can look at the URL used to launch the app and manually make the Dataverse API calls it would do, along with initiating the Globus transfer it would do (via the standard Globus app). As it sounds, this is tedious, and it requires a properly configured Dataverse instance. 

For dev, I have an AWS instance (with associated Globus endpoints) that can be used for testing. Getting on a zoom call is probably the easiest way to walk through it all.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: It adds Globus-related functionality for upload/download if /when the Globus functionality is enabled.

**Is there a release notes update needed for this change?**: yes - tbd

**Additional documentation**:

PR #10201, opened 2023-12-21
IQSS/10200 - Create dirs needed for Globus transfer to managed stores

### LINKED ISSUE #10200: Globus Transfer in fails for new datasets and non-privileged users
**What steps does it take to reproduce the issue?** Missed in testing: When transfer to a managed dataset is initiated, the Globus endpoint does not have the subdirectories required to store datafiles for it. A non-privileged user is given permission to write to that (non-existent) directory, but that permission doesn't allow the non-privileged user to create the directory. (Once the dir exists, e.g. from a privileged user (one who has permissions to create the directory independent of Dataverse) performing a transfer, or the dir being created manually, a non-privileged user can make a success transfer of files to Dataverse. These are potential work-arounds until the bug is fixed.)

v6.1

PR to follow - adding code to create the nested directories required dynamically

qqmyers:
**What this PR does / why we need it**: Creates/verifies the existence of the required directories on a managed Globus endpoint for data files to be added to a dataset. 

- Closes #10200

**Special notes for your reviewer**: FWIW - I've verified that Globus will not create parent dirs if they are missing, so the code tries to create the dir for the authority and then the subdir(s) for the dataset identifier, e.g. the path 10.5072/FK2/ABCDEF. If the dirs exist, Globus responds with 502. If not a 202 indicates successful creation. 

**Suggestions on how to test this**: Perform the manual testing associated with the initial Globus PR #10162, assuring that you're using a new dataset and using a Globus principal that does not have any permissions on the endpoint.  Could test variants where the authority/shoulder dirs already exist (or not). Tests can be repeated on the same dataset by using a privileged principal to delete dirs. On the test instance I have set up, there are endpoints where I can grant others permission to be able to do this or I can help with resetting.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
@@ -240,7 +240,7 @@ public class GlobusServiceBean implements java.io.Serializable {
         MakeRequestResponse result = null;
         String body = "{\"DATA_TYPE\":\"mkdir\",\"path\":\"" + dir + "\"}";
         try {
-            logger.info(body);
+            logger.fine(body);
             URL url = new URL(
                     "https://transfer.api.globusonline.org/v0.10/operation/endpoint/" + endpoint.getId() + "/mkdir");
             result = makeRequest(url, "Bearer", endpoint.getClientToken(), "POST", body);
--- a/src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
@@ -134,7 +134,7 @@ public class GlobusServiceBean implements java.io.Serializable {
      * @param globusLogger - a separate logger instance, may be null
      */
     public void deletePermission(String ruleId, Dataset dataset, Logger globusLogger) {
-        globusLogger.info("Start deleting rule " + ruleId + " for dataset " + dataset.getId());
+        globusLogger.fine("Start deleting rule " + ruleId + " for dataset " + dataset.getId());
         if (ruleId.length() > 0) {
             if (dataset != null) {
                 GlobusEndpoint endpoint = getGlobusEndpoint(dataset);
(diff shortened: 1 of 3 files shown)


landreev (approved):
Appears to be straightforward.

pdurbin:
@qqmyers can you please send me the information on how to test? Thanks!

qqmyers:
To test with a new server, a managed globus store based on a file endpoint need to be configured [as described in the guides](https://guides.dataverse.org/en/latest/installation/config.html#globus-storage).

Once a store is configured, the manual test process involves using [the Dataverse Globus Transfer API](https://guides.dataverse.org/en/latest/developers/globus-api.html). Before the "Adding Files to the Dataset" step, one needs to actually transfer files with the names specified in the response from the prior requestGlobusUpload api call. For those who don't have access to an existing Globus endpoint, a personal Globus endpoint can be used as the source location for the files. 

Using [the standard online Globus transfer app](https://app.globus.org/file-manager) is the easiest way to initiate the transfer, although the Globus API could be used. In the online app, one should name/rename the files on the source endpoint to the names in the requestGlobusUpload response, navigate to the Dataverse endpoint in the other pane, navigate to the dataset's directory (the path given in the requestGlobusUpload response) and then transfer all files at once. The Globus app can then be used to find the taskId for the transfer as needed to complete the "Adding Files to the Dataset" step. (Note that with the API and the in-progress Dataverse-Globus app, the files don't need to be renamed on the source endpoint - that's only needed due to standard app not handling a transfer with name change as one step (afaik)).

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -179,25 +179,95 @@ public JsonObject requestAccessiblePaths(String principal, Dataset dataset, int
...
+            }
+        }
+        return response;
+    }
+    
+    private int makeDir(GlobusEndpoint endpoint, String dir) {
+        MakeRequestResponse result = null;
+        String body = "{\"DATA_TYPE\":\"mkdir\",\"path\":\"" + dir + "\"}";
+        try {
+            logger.info(body);
```
pdurbin, line 243: Do we want this at "info" level?

pdurbin (commented):
just a quick comment

pdurbin:
While we're in here, maybe we could fix the bullets.

Screenshot 2024-01-05 at 5 08 19 PM

You just have to add a line. I tried but my commit was rejected:

```
commit dc18eb74cc864e6ba63d63152767a6efa38f6344
Author: Philip Durbin <philip_durbin@harvard.edu>
Date:   Fri Jan 5 17:02:57 2024 -0500

    make into bullets instead of a single line #10200

diff --git a/doc/sphinx-guides/source/developers/globus-api.rst b/doc/sphinx-guides/source/developers/globus-api.rst
index de9df06a79..2f922fb1fc 100644
--- a/doc/sphinx-guides/source/developers/globus-api.rst
+++ b/doc/sphinx-guides/source/developers/globus-api.rst
@@ -2,6 +2,7 @@ Globus Transfer API
 ===================
 
 The Globus API addresses three use cases:
+
 * Transfer to a Dataverse-managed Globus endpoint (File-based or using the Globus S3 Connector)
 * Reference of files that will remain in a remote Globus endpoint
 * Transfer from a Dataverse-managed Globus endpoint
```

pdurbin:
I fixed a couple typos in the guides.

@qqmyers helped me on the Globus side and with understanding the workflow generally. Adding a file worked fine. Thanks!

Issue #3797, opened 2017-04-26
OAI-PMH responses indicating errors should be processable by OAI-PMH clients


tkgesis:
A valid OAI request such as
https://dataverse.unc.edu/oai?verb=ListIdentifiers&set=census&metadataPrefix=dc
should return a response processable for OAI-PMH clients or HTTP 5XX.

pdurbin:
@tkgesis thanks but when I run `curl -i 'https://dataverse.unc.edu/oai?verb=ListIdentifiers&set=census&metadataPrefix=dc'` I see `HTTP/1.1 500 Internal Server Error` rather than `200`. That said, I do agree that it should return a processable response! Maybe we just need to change the subject/title of this issue.

donsizemore:
[#|2017-04-26T13:29:06.671-0400|WARNING|<snip>
  StandardWrapperValve[OAIServlet]: Servlet.service() for servlet OAIServlet threw exception
javax.servlet.ServletException: Unknown servlet exception in Get.

tkgesis:
'dc' is not a supported metadataPrefix, while 'oai_dc' is.

tkgesis:
According to http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions 
a XML response indicating the badArgument would do.

pdurbin:
> 'dc' is not a supported metadataPrefix, while 'oai_dc' is.

Ah, so you get the same error at https://dataverse.unc.edu/oai?verb=ListIdentifiers&set=census&metadataPrefix=foobar presumably.

donsizemore:
[#|2017-04-27T06:00:08.092-0400|WARNING|glassfish 4.1|edu.harvard.iq.dataverse.harvest.server.web.servlet.OAIServlet|_ThreadID=48;_ThreadName=jk-connector(2);_TimeMillis=1493287208092;_LevelValue=900;|
  Unknown exception in Get; null|#]

[#|2017-04-27T06:00:08.093-0400|WARNING|glassfish 4.1|javax.enterprise.web|_ThreadID=48;_ThreadName=jk-connector(2);_TimeMillis=1493287208093;_LevelValue=900;|
  StandardWrapperValve[OAIServlet]: Servlet.service() for servlet OAIServlet threw exception
javax.servlet.ServletException: Unknown servlet exception in Get.
        at edu.harvard.iq.dataverse.harvest.server.web.servlet.OAIServlet.doGet(OAIServlet.java:262)
[...]

PR #10205, opened 2024-01-04
OAI-PMH error handling: meaningful errors in XML for verb processing

### LINKED ISSUE #9275: Malformed OAI-PMH requests generate an internal server error

Try a malformed request, like, e.g., without any keyword:

https://demo.dataverse.org/oai

Or with a bad keyword:

https://demo.dataverse.org/oai?key=value

* When does this issue occur?

Always

OAI-PMH protocol

* What happens?

The bad request generates an "Internal Server Error"

All users

The client should get an XML response with an error message from the OAI-PMH protocol.

This used to work on old version of Dataverse, example:

https://entrepot.recherche.data.gouv.fr/oai

https://entrepot.recherche.data.gouv.fr/oai?key=value

(The installation in the example runs Dataverse 5.3) 

5.12, 5.12.1

**Screenshots:**

*

pdurbin:

Currently, certain OAI requests, especially related to verbs or the absence of a verb, result in a 500 error and no meaningful information for the end user. Plus the poor sysadmin gets an ugly stacktrace in server.log.

As of this PR, we send XML responses with meaningful errors, like this:

## /oai?foo=bar will show "No argument 'verb' found"

```
<OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
  <responseDate>2024-01-04T21:29:01Z</responseDate>
  <request>http://localhost:8080/oai</request>
  <error code="badArgument">No argument 'verb' found</error>
</OAI-PMH>
```

## /oai?verb=foo&verb=bar will show "Verb must be singular, given: '[foo, bar]'"

```
<OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
  <responseDate>2024-01-04T21:29:01Z</responseDate>
  <request>http://localhost:8080/oai</request>
  <error code="badArgument">Verb must be singular, given: '[foo, bar]'</error>
</OAI-PMH>
```

- Closes #9275

[In Slack](https://iqss.slack.com/archives/C03R1E7T4KA/p1704400519526429) I checked with @landreev and he's fine with the 200 response. "You have successfully queried the OAI server, and here’s your answer" even though the answer is "there was an error." This follows an established pattern set by (at least) "no set found", which I added a test for.

Due to how the xoai library is written, I have to return `<error code="badArgument">` rather than `<error code="badVerb"> as the spec requires. ( http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions says "Value of the verb argument is not a legal OAI-PMH verb, the verb argument is missing, or the verb argument is repeated."). However, Leonid and I are ok with deferring this as it requires work on the library side.

I updated the docker compose file so I could test this.

Related:

- #3797

Play around with variations on these:

- /oai?foo=bar will show "No argument 'verb' found"
- /oai?verb=foo&verb=bar will show "Verb must be singular, given: '[foo, bar]'"

Confirm I didn't break anything (HarvestingServerIT passes for me locally.)

No.

Yes, included.

**Additional documentation**:

None.


src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -256,9 +259,18 @@ private void processRequest(HttpServletRequest httpServletRequest, HttpServletRe
...
+
+            RawRequest rawRequest = null;
+            try {
+                rawRequest = RequestBuilder.buildRawRequest(httpServletRequest.getParameterMap());
+            } catch (BadVerbException bve) {
+                // Verb.Type is required. Hard-code one.
+                rawRequest = new RawRequest(Verb.Type.Identify);
+                // Ideally, withError would accept a BadVerbException.
+                BadArgumentException bae = new BadArgumentException(bve.getLocalizedMessage());
+                rawRequest.withError(bae);
```
landreev, line 271: On a second thought, there should be a less hacky way to handle this, when a rawRequest cannot be created, that does not rely on creating one with a hard-coded valid verb and .withError() etc. 
Note that DataProvider also has a version of the .handle() method that takes a map:
```
  public OAIPMH handle(Map<String, String[]> queryParameters) {
```
https://github.com/gdcc/xoai/blob/xoai-5.2.0/xoai-data-provider/src/main/java/io/gdcc/xoai/dataprovider/DataProvider.java

Also,
https://github.com/gdcc/xoai/blob/xoai-5.2.0/xoai-data-provider/src/test/java/io/gdcc/xoai/dataprovider/DataProviderTest.java#L49 etc.

landreev (changes requested):

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
pdurbin: Awesome, I'll take a look. Thanks!

pdurbin: Ok, fixed in 88af3d4ed1.

One thing I'll note is that I believe we can continue refactoring even further. I tried replacing `catch  BadVerbException` with `catch OAIException` and the tests still pass. Let me know.

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
```diff
@@ -256,10 +258,16 @@ private void processRequest(HttpServletRequest httpServletRequest, HttpServletRe
...
-            
-            OAIPMH handle = dataProvider.handle(rawRequest);
+
+            Map<String, String[]> params = httpServletRequest.getParameterMap();
+            OAIPMH handle;
+            try {
+                RawRequest rawRequest = RequestBuilder.buildRawRequest(params);
+                handle = dataProvider.handle(rawRequest);
+            } catch (BadVerbException bve) {
+                handle = dataProvider.handle(params);
```
landreev, line 268: awesome!

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
landreev: Hmm, that I would not remember off the top of my head - if it was possible to encounter any other OAIException there; but I'm guessing that anything with a valid verb, and any combination of other invalid values should be properly handled by the existing code.  
Catching `BadVerbException` solves the problem at hand, so I'm happy to approve as is.

landreev (approved):
Perfect!

src/main/java/edu/harvard/iq/dataverse/harvest/server/web/servlet/OAIServlet.java
pdurbin: Make sense. Thanks for approving!

pdurbin:
@landreev from https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10205/3/testReport/ we're seeing this:

Screenshot 2024-01-12 at 3 41 35 PM

Any insight on this? I could swear I've seen this switch from noRecordsMatch to noSetHierarchy and back again, but that doesn't make sense. It should be one or the other, right?

pdurbin:
I just merged develop to kick off another Jenkins run to get another data point. I'm not sure if the error is flipping back and forth or not. Let's see.

pdurbin:
In 30e357b I changed the "no such set" test to expect noSetHierarchy rather than noRecordsMatch.

landreev:
If the set does not exist, then I would expect it to consistently return "no set hierarchy". I can't really think of a timing issue or anything like that, that would be causing a "no records match" intermittently.

landreev:
Wait, I was clearly wrong about the above. 
You only get the "no set hierarchy" when there are NO sets on this server, neither the requested set, nor any others. 
So yes, the response will be different, depending on what else the developer has in their database/whether other tests have already run. 

You know, I would go as far as to suggest to drop this test from the PR. Because
a) it's a bit out of scope - since the issue was specifically about a 500 for a bad verb request
b) I have another PR currently in review/QA, #10228 that changes the behavior of ListSets slightly that may actually conflict with this.

landreev:
... I should've said, I have 2 PRs in the works, #10222 and #10228 that change the behavior of both ListIdentifiers and ListSets, that may conflict with this test.

landreev:
... Although feel free to just modify the test, to expect either of the 2 error codes, "no sets" or "no records matching"; up to you. 
Also, should we pick a weirder set name than "census" - in case a developer running the tests happens to have a set named that? 😄

landreev:
(sorry for the confusion; too many OAI PRs going on in parallel...)

pdurbin:
> So yes, the response will be different, depending on what else the developer has in their database/whether other tests have already run.

That's enough to convince me to simply drop the test, which I did in edd6fc8.

Thanks for all the feedback, @landreev

pdurbin:
Tests are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10205/7/testReport/edu.harvard.iq.dataverse.api/

@landreev there's probably no need for you to be assigned to this PR. Thanks for the feedback. I'll remove you.

PR #10214, opened 2024-01-05
9686 move harvesting client

### LINKED ISSUE #9686: Move harvestingclient_id from the dataset to dvobject and use it directly for files

sekmiller:
**What this PR does / why we need it**: Moves Harvesting Client Id from the Dataset record to the dvObject. Also adds the harvesting client id to harvested files and updates dataset-related metrics

Closes #9686 Move harvesting client ID and use it for Files

**Special notes for your reviewer**: I didn't add any metrics related methods for files because I wasn't sure what was needed. There wasn't any guidance in the ticket

**Suggestions on how to test this**: verification is available In the dvobject table. you can see the harvesting client id in the dvobject records of harvested files

[edit: from @landreev A thorough way to qa this would be to test both the migration of any existing harvested files and datasets, and harvesting going forward. For example:
*Before* deploying this branch, make sure you have some harvested files in your instance. You can harvest this small set `controlTestSet` from demo, **in the `dataverse_json` format** - you'll get 6 datasets with some files in them. When you build and deploy this branch, the flyway script in it should properly migrate these harvested dvobjects. So you should be able to see some non-zero counts for both of these:
`SELECT COUNT(id) FROM dvobject WHERE dtype='Dataset' AND harvestingclient_id IS NOT null`
`SELECT COUNT(id) FROM dvobject WHERE dtype='DataFile' AND harvestingclient_id IS NOT null`
Delete the harvesting client in the dashboard (this will delete the harvested content associated with it), re-create it and re-harvest the set from scratch. Check the queries above to confirm that the newly-harvested dvobjects have the `harvestingclient_id` properly set.]

**Is there a release notes update needed for this change?**: none

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/imports/ImportServiceBean.java
```diff
@@ -332,6 +332,11 @@ public Dataset doImportHarvestedDataset(DataverseRequest dataverseRequest, Harve
             
             Dataset existingDs = datasetService.findByGlobalId(ds.getGlobalId().asString());
 
+            //adding the harvesting client id to harvested files #9686
+            for (DataFile df : ds.getFiles()){
+                df.setHarvestedFrom(harvestingClient);
```
landreev, line 337: 👍

src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE dvobject ADD COLUMN IF NOT EXISTS harvestingclient_id BIGINT;
+
+update dvobject dvo set harvestingclient_id = s.harvestingclient_id from
+(select id, harvestingclient_id from dataset d where d.harvestingclient_id is not null) s
+where s.id = dvo.id; 
+
```
landreev, line 6: We also need to populate this harvestingclient_id for any existing harvested file dvobjects here. ... and it should be a relatively simple query, now that all the dataset-type dvobjects have the harvestingclient_id populated.

landreev (changes requested):
Just one extra small thing needed in the flyway scripts, otherwise looks great.

src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
sekmiller: @landreev the script update is in. I also added a release note which may not be necessary or could use your elaboration if you'd like. Thanks!

landreev (approved):
Looks (and works) great. 
I added an extended possible example of a QA test under "how to test".

landreev:
P.S. Would also be prudent to check why the last Jenkins run failed. Probably another fluke, but just in case.

doc/release-notes/9686-move-harvesting-client-id.md
```diff
@@ -0,0 +1 @@
+With this release the harvesting client id will be available for harvested files. A database update will copy the id to previously harvested files./
```
landreev, line 1: Thinking about it, maybe we could add something like `This will make looking up harvested vs. local files in the database simpler: ... But make sure to update any existing database scripts you may have`.
I'll think about it/we can revisit this when putting together the notes for the release.

src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
jp-tosca: I made a PR into this PR so we can add the FK @sekmiller, I spoke earlier with @landreev and it makes sense to add it since it was previously.

jp-tosca:
Beside the conversation about the FK, I am seeing some exceptions on the back:

`Internal Exception: org.postgresql.util.PSQLException: ERROR: column d.harvestingclient_id does not exist
Position: 76
Error Code: 0
Call: SELECT d.id, h.archiveDescription FROM harvestingClient h, dataset d WHERE d.harvestingClient_id = h.id AND d.id IN (5, 6, 7, 8, 9, 10, 11)
Query: DataReadQuery(sql="SELECT d.id, h.archiveDescription FROM harvestingClient h, dataset d WHERE d.harvestingClient_id = h.id AND d.id IN (5, 6, 7, 8, 9, 10, 11)")`

I will investigate a little bit more and get back @sekmiller @landreev

sekmiller:
@jp-tosca - as I mentioned, that error you noted above causes an issue with the homepage "cards" of harvested datasets and files. With the fix I put in you should see a note about the harvested objects directly on the homepage cards.

jp-tosca:
We reviewed some issues today regarding the DB scripts. 💾

PR #10215, opened 2024-01-08
10202 extend get version files api to include total file count

### LINKED ISSUE #10202: Extend getVersionFiles API endpoint to include the total file count, which is required for pagination

To paginate the results of getVersionFiles it is necessary to know the total file count and thus know the pagination limit. Currently a separate endpoint (getVersionFileCounts) is used to know the total file count. 

The goal is to add the total file count in the getVersionFiles response so we don't have to use an additional endpoint to know the total file count and avoid an additional API call.

The Solr search API implements pagination in this way, returning {"total_count", "items"} in the response payload, so this extension will make the API experience more consistent between search endpoints.

The getVersionFileCounts endpoint is necessary to know the different file counts for each filter option, so it is necessary in any case and should remain in the API.

API User

- https://github.com/IQSS/dataverse-client-javascript/pull/107#discussion_r1434858529

- None

- New field with total file count in getVersionFiles API endpoint response

- https://github.com/IQSS/dataverse-client-javascript/issues/109
- https://github.com/IQSS/dataverse-client-javascript/pull/107

stevenwinship:
Extend getVersionFiles API endpoint to include the total file count, which is required for pagination

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/10202

Closes #10202

**Suggestions on how to test this**: call getVersionFiles. response should have total_count in addition to the status and data fields

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -661,7 +635,15 @@ protected Response ok( JsonArrayBuilder bld ) {
             .add("data", bld).build())
             .type(MediaType.APPLICATION_JSON).build();
     }
-    
+
+    protected Response ok( JsonArrayBuilder bld , long totalCount) {
+        return Response.ok(Json.createObjectBuilder()
+                        .add("status", ApiConstants.STATUS_OK)
+                        .add("total_count", totalCount)
```
GPortas, line 642: Although the search endpoint uses underscore for "total_count", I would use camel case here ("totalCount") as it is the convention we normally use in most endpoints.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -546,7 +480,9 @@ public Response getVersionFiles(@Context ContainerRequestContext crc,
             } catch (IllegalArgumentException e) {
                 return badRequest(BundleUtil.getStringFromBundle("datasets.api.version.files.invalid.access.status", List.of(accessStatus)));
             }
-            return ok(jsonFileMetadatas(datasetVersionFilesServiceBean.getFileMetadatas(datasetVersion, limit, offset, fileSearchCriteria, fileOrderCriteria)));
+            // TODO: should we count the total every time or only when offset = 0?
```
GPortas, line 483: Since the goal is to homogenize pagination endpoints (search API and files API), I would count and return it every time, as the search endpoint does. In this way we also guarantee that we always get the actual total count, even if there have been deletions during user paging (Not very likely).

The only reason I would consider doing it only when _offset=0_ would be because of a performance bottleneck in the count query, but I don't think that's the case.

GPortas (commented):
It looks good to me!

I just added a couple of comments related to the code.

I think we should mention the new total count response field in the documentation. See: `native-api.rst` (List Files in a Dataset). There are some examples of calling the endpoint using pagination params. It might be a good place to mention it.

Please also add a release note describing the change.

GPortas:
I just realized that Jenkins step is failing, but I can't see the job (network permissions).

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
stevenwinship: Fixed

stevenwinship:
> I just realized that Jenkins step is failing, but I can't see the job (network permissions).

This was a Jenkins issue that was fixed

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1066,7 +1066,9 @@ The fully expanded example above (without environment variables) looks like this
  
   curl "https://demo.dataverse.org/api/datasets/24/versions/1.0/files"
 
-This endpoint supports optional pagination, through the ``limit`` and ``offset`` query parameters:
+This endpoint supports optional pagination, through the ``limit`` and ``offset`` query parameters.
+To aid in pagination the Json response also includes the total number of rows (totalCount) available.
+Usage example:
```
GPortas, line 1071:
```suggestion
This endpoint supports optional pagination, through the ``limit`` and ``offset`` query parameters.

To aid in pagination the JSON response also includes the total number of rows (totalCount) available.

Usage example:
```

GPortas (commented):
Looks good @stevenwinship, thank you for applying the changes.

I just added a suggestion with minor formatting changes for the documentation.

PR #10222, opened 2024-01-09
3437 OAI records for datasets that still exist, but are temporarily un-indexed should not be marked deleted

### LINKED ISSUE #3437: Harvest: Set export can temporarily be adversely affected by full (clean) reindex.
Set scope is defined by the results of a solr query in the set definition. When the scope changes, such as additions made or removals, the set is updated and subsequent harvests by clients may get additions or delete notifications.

This can accidentally happen if a clean index is performed and the nightly set exporter runs while the index is relatively blank, not yet finished. The net effect would be to mark most of the sets as deleted and any client harvesting before the next nightly set export or a manual reexport, will update their local harvest content by deleting those we have not finished indexing at the time of export. This would be temporary in principle if the client is schedule to run again at a later date -it would get the updated set once our index has finished and export was rerun.

landreev:

A rare case of an issue with a very low number that's still relevant. 

Our OAI sets are defined by search (solr) queries *) therefore when a set is re-exported and the query no longer finds a certain dataset, its OAI record is marked as "deleted" - to communicate to remote clients who may have harvested it that they need to purge it. 
A side effect of this is that if an admin decides to clear their index and run a full reindex overnight - and it can take a while on a large database - if it coincides with a set re-export, many existing datasets can end up marked as "deleted". 

As Kevin explains in the opening comment, like many OAI problems this one is not fatal, but transient: once the holdings are reindexed, the next set re-export will undo this and mark all such datasets as active again. However, a remote client may be forced to delete and re-harvest a large number of our datasets unnecessarily in the process. So we may as well fix this. 

The PR fixes this by adding an extra database check - if the dataset that has disappeared from the search is still there, published and exported properly, the record will be kept in the set as active. 

*) This is true for all sets, except for the main, default, "everything" set - which is defined by a database query. The code in the PR recognizes this case, and avoids unnecessary secondary database checks when the main set is re-exported. 

- Closes #3437

Note that I have added a simple Index API endpoint - `-X DELETE .../api/admin/index/datasets/<id>` - for clearing an individual dataset from Solr. I needed it to help create a meaningful API test for this functionality; I could use the existing API for clearing Solr completely, but this felt cleaner, and I feel like it could be useful on its own. 

Create an OAI set in the harvesting server dashboard. It can be a set with a single published dataset in it (can be defined with a query like `dsPersistentId:B9CEYI` (for example). 
Export the set. Run `http://localhost:8080/oai?verb=ListIdentifiers&set=YOURSETNAME&metadataPrefix=oai_dc`, make sure the record is there. 
Clear your solr index. Re-export the set from the dashboard. Run the above again, confirm that the record is NOT marked as deleted. 

As a control, destroy the dataset, and re-export the set again. This time, the record should be showing with
`<header status="deleted">`

**Additional documentation**:


landreev:
I know why the Jenkins run failed. The test I just added destroyed a dataset that another test in that class relied on. ☹️ 
Will fix.

doc/sphinx-guides/source/admin/solr-search-index.rst
```diff
@@ -86,6 +86,16 @@ To re-index a dataset by its database ID:
 
 ``curl http://localhost:8080/api/admin/index/datasets/7504557``
 
+Clearing a Dataset from Solr
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This API will clear the Solr entry for the dataset specified. It can be useful if you have reasons to stop showing a published dataset in search results and/or on Collection pages, but don't want to destroy and purge it from the database just yet. 
+
+``curl -X DELETE http://localhost:8080/api/admin/index/datasets/<DATABASE_ID>``
```
pdurbin, line 94: A release note to advertise this new API would be nice.

pdurbin (commented):
Just a quick unofficial review of docs.

landreev:
The tests are passing now.

doc/sphinx-guides/source/admin/solr-search-index.rst
landreev: It should also be "if you have reasons *to want* to stop...", probably.

landreev:
My other OAI-PMH PR got merged yesterday; the testing class HarvestingServerIT.java was modified in that PR as well, but a different part of it - so there are no merge conflicts. I've synced the branch, Jenkins running now.

scolapasta (approved):

PR #10228, opened 2024-01-11
3322 Empty OAI sets should not be listed in ListSets 

### LINKED ISSUE #3322: Harvest: Do not list a set as available in ListSets until it has been successfully exported at least once.
The IQSS set in production is hung up on export but it appears in the list of available sets. When I harvest from it I get 0 datasets. It would be better not to list it until it has been successfully exported.

landreev:

Another OAI issue that's 7 years old, but still valid. 
Pretty straightforward: when an OAI set is created, it should not be shown in ListSets until it is successfully exported and contains 1 or more dataset records. 

Closes #3322

Slightly out of scope, but while I was at it, I noticed that the harvesting server page had some calls requiring potentially non-trivial database lookups in the rendering logic rules. This is likely making our own prod. harvesting server page take 7 times longer to load. 
Made the page cache these numbers (7 times = the number of times rendering logic rules are evaluated)

Create a set or two, via the API or the UI, experiment. 
When creating a set on the harvesting server dashboard page, it is exported right away. So the easiest way to test this would be to create a set that doesn't match any records. With a query like `dsPersistentId:NOSUCHID`. Confirm that the set isn't shown in the output of the OAI ListSets.

**Additional documentation**:


landreev:
Please note that the HarvestingServer API tests require the following jvm options: 
```
<jvm-options>-Ddataverse.oai.server.maxidentifiers=2</jvm-options>
<jvm-options>-Ddataverse.oai.server.maxrecords=2</jvm-options>
```
This info is in the guide, but may be a bit buried.

sekmiller (approved):

landreev:
(I just kicked off a Jenkins run manually)

landreev:
Checked in a better definition for the test in the PR. Test is passing locally, Jenkins run is in progress.

Issue #10000, opened 2023-10-11
MPCONFIG: move JvmSettings to SPI (was: allow to retrieve all values used for a placeholder scoped setting)


poikilotherm:
**Overview of the Feature Request**
We need to move the JVM settings stuff into the SPI package, so people writing exporters and other modular stuff can reuse the configuration lookup things in their modules.

~~For JVM options that contain placeholders, we usually provide the substitution value in some way to lookup the configuration value.

It is currently not possible to retrieve a list of used values for a placeholder, determined from settings provided by a sysadmin.~~

It might no longer be necessary to add this, as we seem to come to a consensus that it's OK to have a list of things to look for. See the new PID settings.

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
~~This is currently necessary for storage configuration (#9998), but is likely to be used in the future, when we enable multiple instances of PID providers (#3623).~~ No longer true. One of the inspirations to move these things is now #10048 

**What existing behavior do you want changed?**
See above.

**Any brand new behavior do you want to add to Dataverse?**
Nope

**Any open or closed issues related to this feature request?**
#7000 #3623 #9998

PR #10234, opened 2024-01-12
IQSS/3623 - Multiple PID Provider support

### LINKED ISSUE #3623: Support flexible option for DOI and Handle registration, in same installation
At Scholars Portal (and possibly other library hosted installations) we'd like to have the choice to assign a handle to a dataset that we don't have "authority" over, or, when there are multiple copies of that data across various access repositories (e.g. licensed data for example). This would be in accordance with DOI best practice for data copies.  

Support for both handle and doi as a PID in the same installation would be great. Possibly the user can choose which PID to assign upon deposit. 
Related, but different than https://github.com/IQSS/dataverse/issues/2943

qqmyers:
**What this PR does / why we need it**: Continues the refactoring of the PID Providers to now support 

- instantiating multiple providers, of the same or different types (DataCite, EZId, Handle, Fake, PermaLink), with associated protocol, authority, shoulder, separator (for Perma)
- per provider lists of additional PIDs that are managed by or excluded from the set defined by the auth/separator/shoulder
- basic support for adding additional PidProvider types (classes not yet separated into the SPI)
- per collection setting of the PidProvider to use for new PIDs (datasets, or files created in an existing dataset when it is not possible to create additional PIDs with the dataset's protocol/authority/separator/shoulder)
- api calls to get provider info and see what provider a dataset will use to mint datafile pids, and superuser calls to set the dataset pid generator.

Closes #3623

**Special notes for your reviewer**: See [explanatory doc](https://docs.google.com/document/d/1yCewV6qlNQ7tCgxGXqoldpxOeMcKO18xkLPxrxm4eyc/edit?usp=sharing). 

I've gone ahead with a partial refactoring of the DataCite Provider, removing the old cache class, etc. There's still more that could be done, but I think that's better for future PRs.

I've also tweaked the installer to use the fake provider rather than putting in 10.5072 as an authority and default DataCite username/passwords and then swapping to use :DoiProvider=FAKE at a later step (i.e. for the Jenkins build). The docs are in synch and the installer text as well. 

**Suggestions on how to test this**:  The IT tests should handle legacy config and single provider cases. The unit tests check the basic logic of working with multiple providers. The main area to test would be in configuring multiple providers via microprofile and testing relevant scenarios, e.g. using a test DataCite provider and a PermaLink provider, etc. 

There's also a semi-automated test of the DataCite provider. See the DataCiteProviderTest class - you basically have to export username/password/authority/shoulder for a valid DataCite test account and then run the test - it will go though creating, updating, publishing, deaccessioning the DOI at DataCite (leaving one non-deletable DOI in the 'register' state in your account).

**Is there a release notes update needed for this change?**: yes

**Additional documentation**: included


src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProvider.java
```diff
@@ -1,19 +1,16 @@
-package edu.harvard.iq.dataverse;
```
poikilotherm, line 1: I see why you did not move `PidProvider` and `PidProviderFactory` to the SPI package - we would need to deal with a data model in there, too. Good choice to prepare as much as possible so this next step is well-prepared.

src/main/java/edu/harvard/iq/dataverse/DvObjectServiceBean.java
```diff
@@ -389,4 +395,18 @@ public String generateNewIdentifierByStoredProcedure() {
         return (String) query.getOutputParameterValue(1);
     }
     
+    /** Backward-compatibility method to get the effective pid generator for a DvObjectContainer.
+     * If the dvObjectContainer method fails, this method will check for the old global default settings.
+     * 
+     * If/when those are no longer supported, this method can be removed.
+     * 
+     */
+    public PidProvider getEffectivePidGenerator(DvObjectContainer dvObjectContainer) {
```
poikilotherm, line 404: This is a backward compatibility method. IMHO it should be flagged `@Deprecated` and have the deprecation Javadoc added, so it gets highlighted for later removal.

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProviderFactoryBean.java
```diff
@@ -0,0 +1,230 @@
...
+                            + providerFactory.getClass().getClassLoader().getClass().getCanonicalName());
+        });
+    }
+
+    private void loadProviders() {
+        try {
+            String[] providers = JvmSettings.PID_PROVIDERS.lookup(String[].class);
+            for (String name : providers) {
+                String type = JvmSettings.PID_PROVIDER_TYPE.lookup(name);
+                if (pidProviderFactoryMap.containsKey(type)) {
```
poikilotherm, line 118: The `else` branch is missing. We rely on string matching here: a typo in the type would render the configuration useless without letting the admin know. IMHO this is crucial and potentially should fail deployment or at least trigger a big fat warning.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -91,30 +91,69 @@ public enum JvmSettings {
...
+
+    // PROVIDER HANDLE (legacy - single provider setting)
+    SCOPE_LEGACY_PID_HANDLENET(SCOPE_PID, "handlenet"),
+    LEGACY_HANDLENET_INDEX(SCOPE_LEGACY_PID_HANDLENET, "index", "dataverse.handlenet.index"),
+    SCOPE_LEGACY_PID_HANDLENET_KEY(SCOPE_LEGACY_PID_HANDLENET, "key"),
+    LEGACY_HANDLENET_KEY_PATH(SCOPE_LEGACY_PID_HANDLENET_KEY, "path", "dataverse.handlenet.admcredfile"),
+    LEGACY_HANDLENET_KEY_PASSPHRASE(SCOPE_LEGACY_PID_HANDLENET_KEY, "passphrase", "dataverse.handlenet.admprivphrase"),
+    
+    // PROVIDER EZID - these settings were formerly kept together with DataCite ones
+    SCOPE_PID_EZID(SCOPE_PID_PROVIDER, "ezid"),
```
poikilotherm, line 132: For this and the other providers: I'm not yet sure I like this sub-scoping. It wasn't necessary for `dataverse.files.*`, maybe it isn't necessary here, too? It would be more aligned if not, lot less typing, etc. How large is the risk of collisions?

That said: the `dataverse.files.*` MPC-enablement is still a TODO. If you feel we should do more sub-scoping for these, too, this is fine!

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -91,30 +91,69 @@ public enum JvmSettings {
...
+    PID_PROVIDER_LABEL(SCOPE_PID_PROVIDER, "label"),
+    PID_PROVIDER_AUTHORITY(SCOPE_PID_PROVIDER, "authority"),
+    PID_PROVIDER_SHOULDER(SCOPE_PID_PROVIDER, "shoulder"),
+    PID_PROVIDER_IDENTIFIER_GENERATION_STYLE(SCOPE_PID_PROVIDER, "identifier-generation-style"),
+    PID_PROVIDER_DATAFILE_PID_FORMAT(SCOPE_PID_PROVIDER, "datafile-pid-format"),
+    PID_PROVIDER_MANAGED_LIST(SCOPE_PID_PROVIDER, "managed-list"),
+    PID_PROVIDER_EXCLUDED_LIST(SCOPE_PID_PROVIDER, "excluded-list"),
+
+    // PROVIDER DATACITE (legacy - single provider setting)
+    SCOPE_LEGACY_PID_DATACITE(SCOPE_PID, "datacite"),
```
poikilotherm, line 108: Having all these legacy settings is a bummer, but not easily avoidable without modifying the alias config source. Do you feel we should enable non-placeholder lookups for aliases for settings with placeholders?

IMHO the legacy settings should be marked `@Deprecated` and removed with Dataverse 7.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -91,30 +91,69 @@ public enum JvmSettings {
...
     SCOPE_PID(PREFIX, "pid"),
-    
-    // PROVIDER EZID (legacy) - these settings were formerly kept together with DataCite ones
-    SCOPE_PID_EZID(SCOPE_PID, "ezid"),
-    EZID_API_URL(SCOPE_PID_EZID, "api-url", "doi.baseurlstring"),
-    EZID_USERNAME(SCOPE_PID_EZID, "username", "doi.username"),
-    EZID_PASSWORD(SCOPE_PID_EZID, "password", "doi.password"),
+    PID_PROVIDERS(SCOPE_PID, "providers"),
+    PID_DEFAULT_PROVIDER(SCOPE_PID, "default-provider"),
+    PROVIDERS_DIRECTORY(SCOPE_PID, "provider-implementations-directory"),
```
poikilotherm, line 96: For exporters, we created `dataverse.spi.exporters.directory`. Shouldn't we use a similar scope und `spi` for PID providers, too?

poikilotherm (changes requested):
I very much like this long overdue reorganization of the code and the modularity that comes with it! Left some pointers. 

One thing I can't really add to a single file: I would appreciate moving all of the PID provider files into subpackages of `edu.harvard.iq.dataverse.pidproviders` instead of having them in that package itself. (There already is a subpackage for `datacite`, just use that pattern for all)

poikilotherm:
Addendum: I did not look into tests very much. But it would be great if we could enable automated testing for this.

I dunno if it would be in scope to talk about remote JVM option manipulation at runtime. Let me know if you feel that would help in getting this done. If you think we should create some Testcontainers based test and/or go for some WireMock-ing when multiple PIDs would be activated, let's talk!

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -91,30 +91,69 @@ public enum JvmSettings {
 
     // PERSISTENT IDENTIFIER SETTINGS
     SCOPE_PID(PREFIX, "pid"),
-    
-    // PROVIDER EZID (legacy) - these settings were formerly kept together with DataCite ones
-    SCOPE_PID_EZID(SCOPE_PID, "ezid"),
-    EZID_API_URL(SCOPE_PID_EZID, "api-url", "doi.baseurlstring"),
-    EZID_USERNAME(SCOPE_PID_EZID, "username", "doi.username"),
-    EZID_PASSWORD(SCOPE_PID_EZID, "password", "doi.password"),
+    PID_PROVIDERS(SCOPE_PID, "providers"),
```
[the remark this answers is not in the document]
qqmyers: Yeah, I had all the stores in mind with this too. It sounds like you can do a scan through all properties to find things in dataverse.files.* etc. but I didn't see any efficient way to do it except to have another setting like providers here. That does make it easier to define a provider or store and then turn it on/off by editing the property listing them, so perhaps it's useful on its own.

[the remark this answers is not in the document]
poikilotherm: Scanning through the available options is what #10000 is about. I have some ideas to solve this properly (not as easy as it sounds!), also enabling env vars etc.

But: if we can agree we should just make people add a list of names, that makes things sooooo much easier!

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProviderFactoryBean.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private void loadProviderFactories() {
+        /*
+         * Step 1 - find the PROVIDERS dir and add all jar files there to a class loader
+         */
+        List<URL> jarUrls = new ArrayList<>();
+        Optional<String> providerPathSetting = JvmSettings.PROVIDERS_DIRECTORY.lookupOptional(String.class);
+        if (providerPathSetting.isPresent()) {
+            Path exporterDir = Paths.get(providerPathSetting.get());
```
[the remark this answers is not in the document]
qqmyers: re: duplication - possible, but a little tricky I think given that creating the HashMap involves different methods on Exporters and PID providers and, as is, the code checks which class loader was used. Perhaps if/when Providers go in the SPI?

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProviderFactoryBean.java
```diff
@@ -0,0 +1,230 @@
...
+
+        /*
+         * Step 2 - load all PidProcviderFactories that can be found, using the jars as
+         * additional sources
+         */
+        loader = ServiceLoader.load(PidProviderFactory.class, cl);
+        /*
+         * Step 3 - Fill pidProviderFactoryMap with type as the key, allow external
+         * factories to replace internal ones for the same type. FWIW: From the logging
+         * it appears that ServiceLoader returns classes in ~ alphabetical order rather
```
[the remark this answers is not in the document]
qqmyers: Maybe, but there's no fall through here - either the internal or external one wins (or the highest priority) and none of the others ever get called.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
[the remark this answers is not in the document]
qqmyers: I'll vote for easy. I do think the ability to flip these on/off with one setting change is useful. Probably not done often, but I could see someone wanting to try a globus store, and then turn it off temporarily after testing without having to delete all the secrets, etc.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
qqmyers: I'm not set on it, but to me the benefit is not in avoiding collisions, it's in making it clear which entries you need for a given type of provider. EZID doesn't have mds and rest api URLs, Permas don't have username/password, nothing but handles get a key, etc. I think we have the same issue with stores - direct up/download only apply to S3 stores, etc.

qqmyers: BTW - I think you gave me the idea with https://github.com/IQSS/dataverse/blob/4b5cbe865d215e4f4c8dcf58c28b5cf3af98d414/src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java#L115

qqmyers:
> Addendum: I did not look into tests very much. But it would be great if we could enable automated testing for this.
> 
> I dunno if it would be in scope to talk about remote JVM option manipulation at runtime. Let me know if you feel that would help in getting this done. If you think we should create some Testcontainers based test and/or go for some WireMock-ing when multiple PIDs would be activated, let's talk!

Right now there are two types of tests - new tests in PidUtilTest that cover reading the settings to create the set of PidProviders and then verifying that various PIDs are matched to the right providers in multipid setups, that the allow/disallow lists work, etc., and the existing *IT tests that exercise a broad range of functionality with the single provider legacy settings. While I agree automated testing of a multipid configuration would be useful, it's beyond the scope I thought I could accomplish for 6.2, and, as you imply, is going to require expertise in setting up the testing environment, not to mention that we don't test all of the individual PIDProvider types now, so I suggest handling it as a separate issue(s).

On the plus side, while it might not be obvious due to the refactoring, this PR doesn't really change the internals of any of the PID Provider types, so since they work now, they should continue to. (I had thought to do some simplification to the DataCite provider since it has some caching that isn't used/isn't too useful, etc., but I thought that might be better as a separate future item as well (potentially something that could leverage the [new CrossRef provider code](https://github.com/IQSS/dataverse/pull/10235)).

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProviderFactoryBean.java
```diff
@@ -0,0 +1,230 @@
...
+        if (protocol != null && authority != null && shoulder != null && provider != null) {
+            logger.warning("Found legacy settings: " + protocol + " " + authority + " " + shoulder + " " + provider);
+            if (PidUtil.getPidProvider(protocol, authority, shoulder) == null) {
+                PidProvider legacy = null;
+                // Try to add a legacy provider
+                String identifierGenerationStyle = settingsService
+                        .getValueForKey(SettingsServiceBean.Key.IdentifierGenerationStyle, "random");
+                String dataFilePidFormat = settingsService.getValueForKey(SettingsServiceBean.Key.DataFilePIDFormat,
+                        "DEPENDENT");
+                switch (protocol) {
```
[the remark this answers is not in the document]
qqmyers: FWIW: I've reformatted in Eclipse, but that is leaving the switch and case statements at the same level. We have that style in many other parts of the code as well.

qqmyers:
@poikilotherm - FWIW: I've removed myself as an assignee and don't have anything more planned unless you have further feedback.

I did end up refactoring the DataCite Provider somewhat and added a @Disbaled test that can be run with valid credentials for a DataCite test account. (It does leave a DOI in the undeletable 'registered' state in that account.)

src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProviderFactoryBean.java
```diff
@@ -0,0 +1,230 @@
...
+            for (String name : providers) {
+                String type = JvmSettings.PID_PROVIDER_TYPE.lookup(name);
+                if (pidProviderFactoryMap.containsKey(type)) {
+                    PidProvider provider = pidProviderFactoryMap.get(type).createPidProvider(name);
+                    provider.setPidProviderServiceBean(this);
+                    PidUtil.addToProviderList(provider);
+                }
+            }
+        } catch (NoSuchElementException e) {
+            logger.warning("No PidProviders configured");
```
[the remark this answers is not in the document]
qqmyers: FWIW: I went ahead and added to ConfigCheckService, with the @DependsOn in the other direction so it tries to load the PidProviders first. This initially broke the build as we were using legacy settings that were added via API after Dataverse was loaded. I've changed the build to use the new JVM settings so it works and the ConfigCheck works. The one thing this won't allow is for a new installation to try and use the legacy settings. As long as there's a valid Provider added via the new settings (and the installer sets up FAKE as before), it's all good. I think that's OK - not allowing new instances to use the old db settings.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -178,56 +178,383 @@ Persistent Identifiers and Publishing Datasets
...
+.. _dataverse.pid.default-provider:
+
+dataverse.pid.default-provider
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ID of the default PID provider to use.
+
+.. _dataverse.pid.provider-implementations-directory:
+
+dataverse.pid.provider-implementations-directory
```
poikilotherm, line 276: **issue**: This seems to be outdated, now that it has been moved to the `dataverse.spi.pidproviders.directory` setting.

**suggestion**: Replace with new name.

src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -52,14 +52,14 @@ dataverse.oai.server.maxsets=100
 
 # PERSISTENT IDENTIFIER PROVIDERS
```
poikilotherm, line 53: **suggestion**: As the ids might not be the ones folks will use, I suggest we simply delete these defaults here and migrate them into the provider code.

poikilotherm (approved):
Looks good to me! Much appreciated!

I only have to small things left, see below.

src/main/resources/META-INF/microprofile-config.properties
qqmyers: OK - removed, and the code will use these as defaults if they aren't set.

doc/sphinx-guides/source/installation/config.rst
qqmyers: done

stevenwinship:
FYI src/main/resources/db/migration/V6.1.0.4__* already exists. Please rename :)

stevenwinship:
Putting back in ready for QA since I was unable to get the docker started

[#|2024-03-04T15:23:53.419+0000|SEVERE|Payara 6.2023.8|javax.enterprise.system.core|_ThreadID=35;_ThreadName=RunLevelControllerThread-1709565816776;_TimeMillis=1709565833419;_LevelValue=1000;|
 Exception while loading the app : jakarta.ejb.CreateException: Initialization failed for Singleton ConfigCheckService
jakarta.ejb.CreateException: Initialization failed for Singleton ConfigCheckService
	at com.sun.ejb.containers.AbstractSingletonContainer.createSingletonEJB(AbstractSingletonContainer.java:598)

pdurbin:
@stevenwinship can you please provide the full output? It will help me and @poikilotherm troubleshoot. Thanks!

qqmyers:
@stevenwinship - try that. The basic issue is that Dataverse will refuse to launch if a valid pid provider isn't set up using the new microprofile/jvm options. I've added those to the docker/composer.yml file, which I hope is what's used for the docker launch. If so, hopefully this clears your error. If not, we'll need to find the right place to add these settings.

pdurbin:
There's also docker-compose-dev.yml in the root of the repo. Pick your poison. 😄

qqmyers:
OK - I've added there too - thanks.

docker/compose/demo/compose.yml
```diff
@@ -18,6 +18,11 @@ services:
         -Ddataverse.files.file1.type=file
         -Ddataverse.files.file1.label=Filesystem
         -Ddataverse.files.file1.directory=${STORAGE_DIR}/store
+        -Ddataverse..pid.providers=fake
```
stevenwinship, line 21:
```suggestion
        -Ddataverse.pid.providers=fake
```

docker-compose-dev.yml
```diff
@@ -46,6 +46,11 @@ services:
         -Ddataverse.files.minio1.download-redirect=false
         -Ddataverse.files.minio1.access-key=4cc355_k3y
         -Ddataverse.files.minio1.secret-key=s3cr3t_4cc355_k3y
+        -Ddataverse..pid.providers=fake
```
stevenwinship, line 49:
```suggestion
        -Ddataverse.pid.providers=fake
```

qqmyers:
Sorry for the typos!

docker-compose-dev.yml
```diff
@@ -46,6 +46,11 @@ services:
         -Ddataverse.files.minio1.download-redirect=false
         -Ddataverse.files.minio1.access-key=4cc355_k3y
         -Ddataverse.files.minio1.secret-key=s3cr3t_4cc355_k3y
+        -Ddataverse.pid.providers=fake
```
stevenwinship, line 49:
```suggestion
        -Ddataverse.pid.providers=fake
        -Ddataverse.pid.default-provider=fake
```

PR #10247, opened 2024-01-19
Intellij IDEA webapp hotreload

beepsoft:

This is related to https://github.com/IQSS/dataverse/pull/9959.

When running Dataverse in Docker we still want to be able to just edit things under `src/main/webapp` using IntelliJ IDEA and then just reload the web page to see the changes. To do this in this PR:

1. Mapped Payara `/opt/payara/appserver/glassfish/domains/domain1/applications` folder to `./docker-dev-volumes/glassfish/applications`
2. Added `watchers.xml` File watcher configuration, which can be imported into  IDEA to ...
3. ... run `cpwebapp.sh` to copy changed files under src/main/webapp to `./docker-dev-volumes/glassfish/applications/dataverse-{current version}`

None. 

You need IntelliJ IDEA Ultimate edition for File watchers support.

1. Run Dataverse in Docker https://guides.dataverse.org/en/latest/container/index.html. 
2. In IntelliJ IDEA, import `scripts/intellij/watchers.xml` into File watchers (https://www.jetbrains.com/help/idea/settings-tools-file-watchers.html).
3. Setup Payara remote run config and deploy Dataverse in debug mode (https://dataverse-guide--9959.org.readthedocs.build/en/9959/container/dev-usage.html)
4. Open Dataverse in browser.
5. Edit and save any XHTML/CSS/etc. file under `src/main/webapp`.
6. Reload browser to see changes to the UI.

See this here in action: https://harvard.zoom.us/rec/share/EUOqzaq4NzsH4N8xX7kg9II8w292MEB6eqbyzPMlZOACsErKV-Z20hHBj1HerFH2.bGFh04EdU1Qnu9ic

No.

Yes.

**Additional documentation**:

The documentation for this is practically what is written in `Suggestions on how to test this`.


docker-compose-dev.yml
```diff
@@ -60,6 +60,8 @@ services:
     volumes:
       - ./docker-dev-volumes/app/data:/dv
       - ./docker-dev-volumes/app/secrets:/secrets
+      # Map the glassfish applications folder so that we can update webapp resources using scripts/intellij/cpwebapp.sh
+      - ./docker-dev-volumes/glassfish/applications:/opt/payara/appserver/glassfish/domains/domain1/applications
```
pdurbin, line 64:
```suggestion
      # Uncomment to map the glassfish applications folder so that we can update webapp resources using scripts/intellij/cpwebapp.sh
      # - ./docker-dev-volumes/glassfish/applications:/opt/payara/appserver/glassfish/domains/domain1/applications
```

Should we comment it out, like the example below?

pdurbin (commented):
I left a comment about commenting. 😄

docker-compose-dev.yml
beepsoft: Yes, perhaps. The only problem is that this file is checked into git and if users edit it, they may commit their (local) changes. This is another reason why we'd need a more modular docker compose approach in the future.

pdurbin: Sure, makes sense. I haven't uncommented the "./target" one in a few months (I haven't needed to hack on xhtml, thankfully) but it suffers from the same problem.

I vote we comment it out, merge it, update PR #9959 to include it. And then think more about it while we work on #9959.

pdurbin: @beepsoft I see the 👍 . Do you want to go ahead and click "accept suggestion" to get the commit in?

pdurbin (approved):
Looks good. As discussed in the container meeting this morning, we'll use these files as raw material as we work on the new PR:

- #9959

PR #10248, opened 2024-01-19
Add missing FK harvestingclient_id to dvobject

jp-tosca:
I spoke with @landreev about the missing FK so we would need to add it to the FlyAway script

### DIFF
--- a/src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
+++ b/src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
@@ -1,5 +1,7 @@
 ALTER TABLE dvobject ADD COLUMN IF NOT EXISTS harvestingclient_id BIGINT;
 
+ALTER TABLE dvobject ADD CONSTRAINT fk_dvobject_harvestingclient_id FOREIGN KEY (harvestingclient_id) REFERENCES harvestingclient (id);
+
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V6.1.0.1__9686-move-harvestingclient-id.sql
```diff
@@ -1,5 +1,7 @@
 ALTER TABLE dvobject ADD COLUMN IF NOT EXISTS harvestingclient_id BIGINT;
 
+ALTER TABLE dvobject ADD CONSTRAINT fk_dvobject_harvestingclient_id FOREIGN KEY (harvestingclient_id) REFERENCES harvestingclient (id);
```
qqmyers, line 3: FWIW: We usually try to make Flyway scripts idempotent, and usually try to avoid changing scripts that have gone into develop. For relatively new ones, we've sometimes made changes and then just send a note to the dev mailer (that if you've deployed develop already, you'll have to delete this flyway entry if you deploy again after the PR).

To make this idempotent - look at the past scripts using 'CONSTRAINT' - they either drop the constraint first if it exists, or they catch the error if adding it is a duplicate.

jp-tosca:
This PR was meant to be on [this branch](https://github.com/IQSS/dataverse/tree/9686-move-harvesting-client-id)

PR #10258, opened 2024-01-23
Fix the issue with the thumbnail size

### LINKED ISSUE #10220: Update to 6.1 possibly caused loss of dataset thumbnails
Dear Dataverse Developers,

It seems that after updating 6.0 to 6.1there are several issues with thumbnails on our system. For example the manually created dataset thumbnails are gone and automatically created ones are low resolution. If I manually add large thumbnails, they are also large in the result list, destroying the web page. Here detailed cases with links to our test-system:

Case 1:
Disappeared icon: https://bonndata-test.hrz.uni-bonn.de/dataset.xhtml?persistentId=doi:10.80624/FK2/9S7RJF
Icon added in 6.0, disappeared after update to 6.1. The icon however it still there if I open edit the thumbnail of the dataset. It does not appear even after editing the metadata and re-publishing

Case 2:
Auto-generated: https://bonndata-test.hrz.uni-bonn.de/dataset.xhtml?persistentId=doi:10.80624/FK2/WOX4YV
The thumbnail was automatically generated from pdf of the dataset files in 6.0. After update to 6.1 it became tiny resolution. Generally all automatically generated icons have tiny resolution now....

Case 3:
Dataset with manually added large thumbnail in 6.1. https://bonndata-test.hrz.uni-bonn.de/dataset.xhtml?persistentId=doi:10.80624/FK2/0OBMAR
The thumbnail appear ok in the dataset page, but in search results it is remaining huge. https://bonndata-test.hrz.uni-bonn.de/dataverse/sandbox

 
Since I am not sure whether there could be other source of error (I could not find any yet) I would like just to ask those who did update to check if they encounter the same..

PS: At the moment, I rolled back our productive installation back to 6.0 and try to find the cause on the test system.

Best regards
Sergej

jp-tosca:
It was reported on this issue that a thumbnail will be showing the full resolution image on the search/listing of the datasets. During the testing @landreev discovered that there was an issue also with a discrepancy of the resolution of a published vs a draft dataset. 

Closes #10220

This PR does not address another issue discovered that sometimes a customized thumbnail will not show, this will be investigated on a separate issue.  Probably to test be sure to have some datasets with custom thumbnails from a previous installation. 

It should fix the issues caused to the UI reported on #10220


landreev (commented):
I would prefer to at least figure out what's going on with the 3rd known bug, before making this PR. To check whether it's new in 6.1; and to at least get an idea of how difficult it would be to fix. 
I understand that we may want to open a new issue for it, if it's something difficult/time-consuming - but we don't even know that yet.

landreev:
> another issue discovered that sometimes a customized thumbnail will not show

Just to be clear, it's not really "sometimes", as in, it's not a mystery of any kind - it appears to be a very clearly/narrowly defined condition: custom logos don't work unless there is at least one image Datafile in the dataset.

landreev:
(meant to add this here:)
It is new as of 6.1. Can be easily tested using our own prod.:
Screen Shot 2024-01-23 at 2 41 48 PM

jp-tosca:
I have right now some Datasets that have images and still don't show the thumbnail, see:

dsthumbn

It is still not clear to me exactly why or how this happens, since we have a solution and identified these 2 bugs I thought it would be a good idea just to get rid of them and then come with a separate solution. This way we can start doing research and asses what would it take and why did the 3rd bug is happening. 

Let's keep this PR, right now I will do some final QA on another PR and then come back to this and do some research. For my testing this not only happens when there is no image on the dataset.

landreev:
Everything is happening too fast in that gif, lol 😄 
But yeah, it does look like DS4 has BOTH image Datafiles, AND the custom logo (?)... 
If it's really looking like that 3rd bug is NOT as straightforward as we thought it was, then that could be a reason to treat it as a separate issue. 
But please do open an issue for it. Are you saying that you are don't know for the fact how to reproduce/get to this condition that DS4 is in?

jp-tosca:
That is correct as of now is not clear to me why or when the third bug happens. At first glance doesn't seem as straight forward as we initially thought and part of the reason why I thought on separating this from this issue. 😄

landreev:
@jp-tosca When you have a moment, could you check and post the output of `ls -l dataset_logo*` in the storage directory of the dataset DS4 above - the one where the custom thumbnail isn't showing?
I'm working on too many other things at once r/n, but very curious about it by now, so can't help myself.

jp-tosca:
Sadly I deleted that installation already and I don't have it anymore since I am testing another PR and I needed to clear the environment multiple times but I will keep everyone updated tomorrow with my testing about this issue.

landreev:
@jp-tosca To reiterate, if we are not fixing the other bug(s) in #10220 in this PR, please open a separate github issue for that asap. Otherwise we'll merge this PR and that will mark the issue as resolved without any trace left of the remaining problems.

jp-tosca:
Just to follow up the Issue for the third bug was created and @qqmyers already submitted a PR for it 😃

PR #10269, opened 2024-01-25
IQSS/10268 fix thumbnail selection logic

### LINKED ISSUE #10268: Dataset thumbnail not applied in some cases

in #10220 was reported the thumbnails were not working properly after 6.0 and 3 bugs were identified during the testing and 2 of them were addressed on #10258. The remaining issue was initially identified by @landreev [on this comment](https://github.com/IQSS/dataverse/issues/10220#issuecomment-1896191664) but after some testing, I noticed that this would also happen in some cases where the Dataset had already images, more info [here](https://github.com/IQSS/dataverse/issues/10220#issuecomment-1905071001) it was clear to me that this may require to look a bit deeper on the code @qqmyers provided information [here](https://github.com/IQSS/dataverse/pull/10258#issuecomment-1907047985) that can be useful to solve this bug. 

* When does this issue occur?
Still not clear, initially was suspected that this would happen only when a dataset had no other images attached but I was able to reproduce this on Datasets with images. 

Dataset result page.

* What happens?
A custom thumbnail will not show up. 

Everyone

A dataset should show the custom thumbnail when defined.

This bug was discovered during a report after 6.1

#10220
#10258

**Screenshots:**
fix

qqmyers:
**What this PR does / why we need it**: See the issue for the description of the bug being fixed.

Closes #10268 

**Special notes for your reviewer**: FWIW - I don't think we have any other way to see of there's an uploaded logo than to check and see if it exists. If there's any db mechanism, we could check that instead.

**Suggestions on how to test this**: See the issue -create  a new dataset with no datafiles/datafiles that only use default icons, upload a separate dataset logo. The old code would only show the default dataset icon in the list of datasets on the front page, with this PR you'll see the logo.

**Additional documentation**:


landreev (approved):
Makes sense. 
Thank you, Jim!

PR #10270, opened 2024-01-25
Revert "9686 move harvesting client"

landreev:
Reverts IQSS/dataverse#10214

Pulling the plug on the PR, at least until we figure out how to reconcile rearranging these tables with the already released script V6.0.0.5 that counts on finding harvestingclient_id in the old place, and thus failing on a brand new database created by a deployment of a post 6.1 war file. 
Another problem with the script as merged, the missing constraint (#10248) was never added to it.


sekmiller (approved):
OK. let's do this

landreev:
There may not be an absolutely transparent and painless solution. But @qqmyers' proposal does not sound outright unreasonable:  we re-organize this flyway script as to only execute the UPDATE block if the harvestingclient_id is still in the dataset table; i.e., if this is indeed an upgrade, and not a new deployment. We then similarly reorganize V6.0.0.5. Since that one is in 6.1, we will need to add a sql change of the flyway_schema_history table to the upgrade steps for those upgrading from 6.1. A bit hacky, but survivable. And Jenkins runs will just keep working, because those are always from-scratch builds. 
(this is all of course up for discussion still; there is also a potential solution of not doing anything and simply canceling the plan of rearranging these tables]

PR #10274, opened 2024-01-26
#9748 delete tools only added by tests

### LINKED ISSUE #9748: Installed External Tools deleted by running Integration Test
NOTE: This may not be worth fixing, but it did cause a minor annoyance while working on the Ask The Data tool PR. (I didn't put in the time then so as not to slow down the 5.14 release.) 

When you run ExternalToolsIT, the first thing it does is delete all of the existing external tools. You have to re-run the api to get your tools back.

* When does this issue occur?
Running the ExternalToolsIT

it's manifested in the files list on the dataset page. Tool icons disappear after the test is run - and the external tools table is wiped out.

* What happens?

developers

it would be great if the tools didn't get wiped out, but that would probably make the existing tests fail - so fix the test so that the tools don't get wiped out.

dev/5.14

unknown

**Screenshots:**

*

sekmiller:
**What this PR does / why we need it**: Updates ExternalToolsIT so that existing tools are not deleted

Closes #9748 Installed External tools deleted by running ExternalToolsIT

**Special notes for your reviewer**: Going in I thought that deleting the tools was a requirement to get the tests to pass. That turned out not to be the case which made the fix easier. The tests were modified to delete only those tools added via the tests.

**Suggestions on how to test this**: Install an external tool in your test environment. Run ExternalToolsIT
verify that your pre-installed tool(s) exist after running the test. And that there are no additional tools.
api for getting a list of installed tools:
 curl http://localhost:8080/api/admin/externalTools

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/ExternalToolsIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/ExternalToolsIT.java
@@ -40,21 +40,6 @@ public class ExternalToolsIT {
     @Test
     public void testFileLevelTool1() {
 
-        // Delete all external tools before testing.
-        Response getTools = UtilIT.getExternalTools();
-        getTools.prettyPrint();
-        getTools.then().assertThat()
-                .statusCode(OK.getStatusCode());
-        String body = getTools.getBody().asString();
-        JsonReader bodyObject = Json.createReader(new StringReader(body));
-        JsonArray tools = bodyObject.readObject().getJsonArray("data");
-        for (int i = 0; i < tools.size(); i++) {
-            JsonObject tool = tools.getJsonObject(i);
-            int id = tool.getInt("id");
-            Response deleteExternalTool = UtilIT.deleteExternalTool(id);
-            deleteExternalTool.prettyPrint();
-        }
-
         Response createUser = UtilIT.createRandomUser();
         createUser.prettyPrint();
         createUser.then().assertThat()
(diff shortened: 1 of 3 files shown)


src/test/java/edu/harvard/iq/dataverse/api/ExternalToolsIT.java
```diff
@@ -145,26 +130,14 @@ public void testFileLevelTool1() {
                 .statusCode(OK.getStatusCode())
                 // No tools for this file type.
                 .body("data", Matchers.hasSize(0));
+        
+        //Delete the tool added by this test...
+        Response deleteExternalTool = UtilIT.deleteExternalTool(toolId);
```
scolapasta, line 135: @sekmiller for the other delete, you follow with an:
     deleteExternalTool.then().assertThat()
For consistency can you add that here?

scolapasta (changes requested):

src/test/java/edu/harvard/iq/dataverse/api/ExternalToolsIT.java
sekmiller: sure

scolapasta (approved):

pdurbin:
There's a test failure at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10274/1/testReport/edu.harvard.iq.dataverse.api/ExternalToolsIT/testFileLevelTool1/

Seems like an easy fix:

```
1 expectation failed.
JSON path data[0].displayName doesn't match.
Expected: AwesomeTool
  Actual: Data Explorer
```

sekmiller:
Weird. It's not failing for me locally.

pdurbin (approved):
Tests are passing. Code looks good. Will merge.

PR #10299, opened 2024-02-06
Update get file endpoint to add a datasetVersion optional query parameter and extend its payload

### LINKED ISSUE #10280: Update get file endpoint to add a datasetVersion optional query parameter and extend its payload
### Overview of the Feature Request

Update the [Get JSON Representation of a File](https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-file) endpoint to include a `datasetVersion` optional query parameter. This parameter would allow users to specify the datasetVersion of the file they wish to retrieve, in the same way that happens when using the [Get JSON Representation of a Dataset](https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-dataset) endpoint 

Example `/api/files/:persistentId/datasetVersion/3.0?persistentId=doi:10.70122/FK2/QJANR3/J2MR4Q`

A part from the new `datasetVersion` query parameter we also want to include the dataset version information in the payload. This will allow the SPA to display the File Page with all the dataset version data that used to appear in the JSF UI.

This is the dataset version information needed in the SPA:

```
versionId: number
displayName: string
versionNumber: {majorNumber?: number, minorNumber?: number}
publishingStatus: string
citation: string
isLatest: boolean
isInReview: boolean
latestVersionPublishingStatus: string
```

API User

This request is inspired by the need  to access specific versions of files metadata to display them in the SPA.

This conversation inspired the request:
- https://github.com/IQSS/dataverse-client-javascript/pull/119#discussion_r1469714748

The 'Get JSON Representation of a File' endpoint

Allow API users to specify the dataset version of the file they wish to retrieve. 

Modify the get file endpoint to match the structure of the get dataset endpoint

- https://github.com/IQSS/dataverse-client-javascript/issues/115

GPortas:

The API endpoint `api/files/{id}` has been extended to support the following optional query parameters:

- `includeDeaccessioned`: Indicates whether or not to consider deaccessioned dataset versions in the latest file search. (Default: `false`).
- `returnDatasetVersion`: Indicates whether or not to include the dataset version of the file in the response. (Default: `false`).

A new endpoint `api/files/{id}/versions/{datasetVersionId}` has been created. This endpoint returns the file metadata present in the requested dataset version. To specify the dataset version, you can use ``:latest-published``, or ``:latest``, or ``:draft`` or ``1.0`` or any other available version identifier.

The endpoint supports the `includeDeaccessioned` and `returnDatasetVersion` optional query parameters, as does the `api/files/{id}` endpoint.

- Closes #10280 

DataFile's `getLatestPublishedFileMetadata` and `getLatestFileMetadata` methods have been refactored to reduce duplication and improve readability and logic reusability.

As I mentioned in the frontend weekly meeting, I discovered that these methods include deaccessioned dataset versions in the lookup, regardless of the user permissions, which may expose file metadata of deaccessioned dataset versions publicly.

It is important to review this behavior in the future, as we should only expose deaccessioned file metadatas if the user has permissions and requires deaccessioned versions explicitly ("includeDeaccessioned" parameter).

In order not to overextend this PR, I finally chose not to fix the issue in this PR, keeping this behavior as it is for now, since the mentioned methods are used in many parts of the code and modifying them would require a more exhaustive review, since there are cases in which obtaining latest files in a deaccessioned state may be necessary (for example when handling new file versions or editing them).
 
So although I have refactored the `getLatestFileMetadata` and `getLatestPublishedFileMetadata` methods, I am not using them in this API extension. I am using commands instead, where the user permissions are checked and the includeDeaccessioned optional query param is managed, as in get dataset commands.

------------------------

**`files/{id}/draft` endpoint is no longer available** in favor of the new endpoint `files/{id}/versions/{datasetVersionId}`, which can use the version identifier `:draft` (`files/{id}/versions/:draft`) to obtain the same result.

Follow the new docs to call the extended API endpoint. For example:

**api/files/{id}:**

`curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" "http://localhost:8080/api/files/:persistentId/?persistentId=doi:10.5072/FK2/J8SJZB&returnDatasetVersion=true&includeDeaccessioned=true"`

**api/files/{id}/versions/{datasetVersionId}:**

`curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" "http://localhost:8080/api/files/:persistentId/versions/:draft?persistentId=doi:10.5072/FK2/J8SJZB&returnDatasetVersion=true&includeDeaccessioned=true"`

None

Yes. Added.

**Additional documentation**:

None


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2790,6 +2792,106 @@ The fully expanded example above (without environment variables) looks like this
...
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" "https://demo.dataverse.org/api/files/:persistentId/?persistentId=doi:10.5072/FK2/J8SJZB&returnDatasetVersion=true"
+
+Get JSON Representation of a File given a Dataset Version
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note:: Files can be accessed using persistent identifiers. This is done by passing the constant ``:persistentId`` where the numeric id of the file is expected, and then passing the actual persistent id as a query parameter with the name ``persistentId``.
```
qqmyers, line 2834: "When a file has been assigned a persistent identifier, it can be used in the api "...  - just something to point out that files may not have PIDs

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLatestPublishedFileMetadataCommand.java
```diff
@@ -0,0 +1,25 @@
...
+        super(request, dataFile, includeDeaccessioned);
+    }
+
+    @Override
+    public FileMetadata execute(CommandContext ctxt) throws CommandException {
+        return dataFile.getFileMetadatas().stream().filter(fileMetadata -> {
+            DatasetVersion.VersionState versionState = fileMetadata.getDatasetVersion().getVersionState();
+            return (!versionState.equals(DatasetVersion.VersionState.DRAFT)
+                    && !(versionState.equals(DatasetVersion.VersionState.DEACCESSIONED) && !includeDeaccessioned)
+                    && isDatasetVersionAccessible(fileMetadata.getDatasetVersion(), dataFile.getOwner(), ctxt));
```
qqmyers, line 22: Doesn't this include the deaccession check on the line above?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLatestAccessibleFileMetadataCommand.java
```diff
@@ -0,0 +1,41 @@
...
+        this.includeDeaccessioned = includeDeaccessioned;
+    }
+
+    @Override
+    public FileMetadata execute(CommandContext ctxt) throws CommandException {
+        FileMetadata fileMetadata = null;
+
+        if (ctxt.permissions().requestOn(getRequest(), dataFile.getOwner()).has(Permission.ViewUnpublishedDataset)) {
+            fileMetadata = ctxt.engine().submit(
+                    new GetDraftFileMetadataIfAvailableCommand(getRequest(), dataFile)
```
qqmyers, line 29: this is going to go through all the command setup and recheck the permission, just to call DataFile.getLatestFileMetadata() and see if it's draft - would it be easier/more efficient to just repeat those two lines here (or even make a 1 line DataFile.getDraftFileMetadata() call to use in both?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLatestAccessibleFileMetadataCommand.java
```diff
@@ -0,0 +1,41 @@
...
+            );
+        }
+
+        if (fileMetadata == null) {
+            fileMetadata = ctxt.engine().submit(
+                    new GetLatestPublishedFileMetadataCommand(getRequest(), dataFile, includeDeaccessioned)
```
qqmyers, line 35: I guess the same question here. There's no extra overall permission check but if the one line in the other command were just called here (or this time wrapped in an AbstractGetPublishedFileMetadata. getLatestPublishedFileMetadata() method) we could avoid the extra submit().

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
```diff
@@ -1403,68 +1400,271 @@ public void testDataSizeInDataverse() throws InterruptedException {
...
                 .body("data.dataFile.filesize", equalTo(8361))
                 .statusCode(OK.getStatusCode());
-        
-        getFileDataResponse = UtilIT.getFileData(dataFileId, apiTokenRegular);
+
+        // Regular user should not get to see draft file data
+        getFileDataResponse = UtilIT.getFileData(dataFileId, regularApiToken);
         getFileDataResponse.then().assertThat()
-                .statusCode(BAD_REQUEST.getStatusCode());
+                .statusCode(UNAUTHORIZED.getStatusCode());
```
qqmyers, line 1436: If you're sending a valid api token, I think this should be FORBIDDEN (403) rather than UNAUTHORIZED (401).

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
```diff
@@ -1403,68 +1400,271 @@ public void testDataSizeInDataverse() throws InterruptedException {
...
         getFileDataResponse.then().assertThat()
                 .statusCode(OK.getStatusCode());
 
-        //cleanup
-        Response destroyDatasetResponse = UtilIT.destroyDataset(datasetId, apiToken);
-        assertEquals(200, destroyDatasetResponse.getStatusCode());
+        // Regular user should not get to see draft file data
+        getFileDataResponse = UtilIT.getFileData(dataFileId, regularApiToken, DS_VERSION_DRAFT);
+        getFileDataResponse.then().assertThat()
+                .statusCode(UNAUTHORIZED.getStatusCode());
```
qqmyers, line 1480: same here - 403

qqmyers (changes requested):
Looks good. I left a few comments w.r.t. some changes, but nothing major.

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
GPortas: I agree, but this is a core behavior in the API. When the user does not have sufficient permissions to execute the commands associated with an endpoint, a 401 is always returned. 

See `execCommand` method from `AbstractApiBean` (https://github.com/IQSS/dataverse/blob/develop/src/main/java/ edu/harvard/iq/dataverse/api/AbstractApiBean.java#L604).

Changing the error code should be a general change IMO to be consistent across the API.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLatestAccessibleFileMetadataCommand.java
GPortas: Makes sense, I just added `DataFile.getDraftFileMetadata()`. Thanks.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLatestAccessibleFileMetadataCommand.java
GPortas: This is really nice. I just refactored it as you describe.

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
qqmyers: I created #10340 for a more general fix.

PR #10303, opened 2024-02-06
10297 dataset metadata endpoint fix

### LINKED ISSUE #10297: api/datasets/{id}/metadata defaults to :draft causing a server error when no DRAFT version exist

When trying to access the endpoint `api/datasets/{id}/metadata` it will try by default to get the draft version and if this endpoint is used on a dataset with no draft it will return an error specifying that there is no draft version accompanied by an exception on the server side.

All

**What did you expect to happen?
We discussed this on slack today, the endpoint should default to the latest (draft if exist, published if there is no draft)

6.1

jp-tosca:

When trying to access the endpoint api/datasets/{id}/metadata it will try by default to get the draft version and if this endpoint is used on a dataset with no draft it will return an error specifying that there is no draft version accompanied by an exception on the server side.

Closes #10297 

**Special notes for your reviewer**: It was discussed on slack that this was the intended behavior of the endpoint. Tests were added to validate when multiple versions of the dataset exist.

Try to access the endpoint at `api/datasets/{id}/metadata`

If a draft exists and the user has access to it it should be prioritized, otherwise, a published version will provided if exists.

Yes.

### DIFF
--- a/doc/release-notes/10297-metadata-api-fix.md
+++ b/doc/release-notes/10297-metadata-api-fix.md
@@ -1,0 +1,1 @@
+The API endpoint `api/datasets/{id}/metadata` has been changed to default to the latest version of the dataset that the user has access.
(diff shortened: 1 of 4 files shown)


qqmyers (approved):
Looks good.

pdurbin:
Tests are passing at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10303/1/testReport/

Looks good. Lots of tests added. I don't think it's worth adding this to the API Changelog, honestly. Merging.

PR #10306, opened 2024-02-07
Geospatial properties name fix 

### LINKED ISSUE #5645: Attribute naming of bounding-box wrong (North and South Longitude)
I was working on a harvester to grab data/metadata from Dataverse and noticed that North and South Latitude fields in the geospatial metadata are mislabeled as "Longitude". The web interface titles the fields as "Latitudes" so this doesn't break anything other than for developers who are trying to parse metadata coming out of Dataverse. The error is both in actual metadata coming out of Dataverse and in the field names in the Dataverse Metadata v4.x spreadsheet, although the spreadsheet has the correct labels in the title.

jp-tosca:

The metadata block for Geospatial Data had 2 fields that were incorrectly labeled as `North Longitude` and `South Longitude`. There were some efforts to partially address this previously to this PR but this change will permanently deprecate `North Longitude` and `South Longitude` for `North Latitude` and `South Latitude`

Closes #5645 

- This will cause that endpoints that use "northLongitude" and "southLongitude" to stop working.
- Some tests were updated to reflect the changes on the API
- In the case of the OpenAire test had to change since the XML is generated with a Map and the order of the fields switched fields. 

I tested this by importing data with the API and also creating data and then migrating. This also was tested on a fresh installation but let me know if you have any other ideas of how to test this. I also tested other things like harvesting. 

N/A

Yes, this change will require users to stop using references to `North Longitude` and `South Longitude`.

Docs: https://dataverse-guide--10306.org.readthedocs.build/en/10306/api/changelog.html


sekmiller (commented):
When it comes time for the actual release we'll need to include instructions for updating schema.xml for installations with and without custom metadata blocks. - maybe make a note of that here.

sekmiller (changes requested):
Looks good overall. We need to put a note into the release notes about updating the schema.xml

sekmiller (approved):
OK. thanks for the update

PR #10322, opened 2024-02-14
10286 add owner info

### LINKED ISSUE #10286: Extend the API to retrieve owner info for files, datasets and collections

For each resource: datafile, dataset or collection, we need to get the owner information and the owners above. This is necessary so that the SPA can implement breadcrumb navigation.

The returned info must contain:

- Identifier (PID / alias /numeric)
- Resource type: file, dataset or collection
- Display name (will be used in the breadcrumb)
- Owner: object including the same parameters to have the hierarchy

API User.

- SPA requirements

- None

- Extend the API to retrieve owner info for files, datasets and collections

https://github.com/IQSS/dataverse-frontend/issues/265

sekmiller:
**What this PR does / why we need it**: Adds info about the owners of an object for use by the SPA to display breadcrumbs

Closes #10286 Extend the API to include owner information for dv objects

I added a boolean of "includeOwners" to the json printers of the various objects. Per a suggestion from Guillermo I didn't include the object itself in the owners array.

**Suggestions on how to test this**: add a query param of "returnOwners=true" to any of the get object apis. You should see an array of each owning object up to the root dv collection.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -186,11 +186,12 @@ public interface DsVersionHandler<T> {
     @GET
     @AuthRequired
     @Path("{id}")
-    public Response getDataset(@Context ContainerRequestContext crc, @PathParam("id") String id, @Context UriInfo uriInfo, @Context HttpHeaders headers, @Context HttpServletResponse response) {
+    public Response getDataset(@Context ContainerRequestContext crc, @PathParam("id") String id, @Context UriInfo uriInfo, @Context HttpHeaders headers, @Context HttpServletResponse response,  @QueryParam("returnOwners") Boolean returnOwners) {
```
MellyGray, line 189: For the frontend, we will need the same query parameter but in a different endpoint, getVersion, as this is the one we are using to retrieve a dataset. Thx!

```
@Path("{id}/versions/{versionId}")
 public Response getVersion
```

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -303,6 +306,46 @@ public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts) {
         }
         return jsonArrayOfContacts;
     }
+    
+    public static JsonArrayBuilder getOwnersFromDvObject(DvObject dvObject) {
+        
+        List <DvObject> ownerList = new ArrayList();
```
MellyGray, line 312: I've noticed a potential improvement in the data structure format we're using. The current implementation may not intuitively indicate the hierarchical relationship between "root" and "First Dataset" unless one already understands the intended structure. This could lead to misunderstandings, especially for those unfamiliar with the context.

Here's the current payload structure:
```javascript
ownerArray = [
  {
    type: 'DATASET',
    identifier: 'doi:10.5072/FK2/WTBMGC',
    displayName: 'First Dataset'
  },
  { type: 'DATAVERSE', identifier: 'root', displayName: 'Root' }
];
```
To enhance clarity, I suggest adopting a nested object approach, where each entity has an 'owner' property.

Proposed payload structure:
```javascript
ownerArray = {
  type: 'DATASET',
  identifier: 'doi:10.5072/FK2/WTBMGC',
  displayName: 'First Dataset',
  owner: {
    type: 'DATAVERSE',
    identifier: 'root',
    displayName: 'Root'
  }
};
```
I'd appreciate your thoughts on this proposed adjustment.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -303,6 +306,46 @@ public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts) {
...
+                }
+                if (dvo.isInstanceofDataset()){
+                    ownerObject.add("type", "DATASET");
+                }
+                 if (dvo.isInstanceofDataFile()){
+                    ownerObject.add("type", "DATAFILE");
+                }
+                if (dvo.isInstanceofDataverse()){
+                    Dataverse in = (Dataverse) dvo;
+                    ownerObject.add("identifier", in.getAlias());
```
MellyGray, line 334: It would be beneficial to include the version as well to the ownerObject, because for a file, its owner might be the dataset doi:10.5072/FK2/WTBMGC with version=2.0, but its parent is not the same dataset at an earlier version such as 1.0. I hope I'm making myself clear

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -186,11 +186,12 @@ public interface DsVersionHandler<T> {
...
         return response( req -> {
             final Dataset retrieved = execCommand(new GetDatasetCommand(req, findDatasetOrDie(id)));
             final DatasetVersion latest = execCommand(new GetLatestAccessibleDatasetVersionCommand(req, retrieved));
-            final JsonObjectBuilder jsonbuilder = json(retrieved);
+            Boolean includeOwners = returnOwners == null ? false : returnOwners;
```
GPortas, line 193: You can directly use the primitive boolean type: `@QueryParam("returnOwners") boolean returnOwners` 

And thus avoid the null check. If the query param is not set in the URL, `false` is the default value.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -610,10 +610,12 @@ private Dataset parseDataset(String datasetJson) throws WrappedResponse {
     @GET
     @AuthRequired
     @Path("{identifier}")
-    public Response viewDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String idtf) {
+    public Response viewDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String idtf, @QueryParam("returnOwners") Boolean returnOwners) {
```
GPortas, line 613: To be consistent with method naming in other places, why not rename the endpoint method from "viewDataverse" to "getDataverse"?

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
GPortas: I also think the structure @MellyGray mentions would improve readability and make the response more intuitive and self-explanatory.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -303,6 +306,46 @@ public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts) {
...
+        while (dvObject != null) {
+            ownerList.add(dvObject);
+            dvObject = dvObject.getOwner();
+        } 
+        
+        JsonArrayBuilder jsonArrayOfOwners = Json.createArrayBuilder();
+        
+        for (DvObject dvo : ownerList){
+                JsonObjectBuilder ownerObject = jsonObjectBuilder();
+                if (dvo.isInstanceofDataverse()){
```
GPortas, line 323: I would extract the type check to a separate method to improve readability, and try to use a `switch` statement instead of these `ifs`. As far as I know a `dvObject` can't be of more than one type, and by having these `ifs` without `else ifs` it seems that a dvObject can have multiple types which is confusing.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -1887,6 +1887,47 @@ public void testDeleteDatasetWhileFileIngesting() {
                 .statusCode(FORBIDDEN.getStatusCode());
 
     }
+    
+    @Test
+    public void testGetIncludeOwnerArray() {
```
GPortas, line 1892:
```suggestion
    public void testGetDatasetOwners() {
```

Just to be consistent with the other test method names.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
qqmyers: also - using "isPartOf" to mirror https://schema.org/isPartOf might be better than "owner" - that's what is used in the OREMap file.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -303,6 +306,46 @@ public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts) {
...
+                 if (dvo.isInstanceofDataFile()){
+                    ownerObject.add("type", "DATAFILE");
+                }
+                if (dvo.isInstanceofDataverse()){
+                    Dataverse in = (Dataverse) dvo;
+                    ownerObject.add("identifier", in.getAlias());
+                }
+                if (dvo.isInstanceofDataset() || dvo.isInstanceofDataFile() ){
+                    if (dvo.getIdentifier() != null){
+                        Dataset ds = (Dataset) dvo;
```
qqmyers, line 338: I can get here if dvo is a DataFile and it has a global ID right? so the cast might fail. I don't think it is needed though.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
sekmiller: I simplified it here because since we're ignoring the object itself we can only have a Dataverse or a Dataset

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
sekmiller: @MellyGray do we also need it for getPrivateUrlDatasetVersion @Path("privateUrlDatasetVersion/{privateUrlToken}")?

MellyGray: @sekmiller Yes, I think so. Thank you for noting this change!

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
sekmiller: I added the logic for including a version number, but since we're omitting the base object from the isPartOf it can only apply to Files, however there's no endpoint for getting a file by its version like there is for dataset - the current result is that if I use the getFileData (@Path("{id}")) it gets me the latest available version

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
sekmiller: Fixed the logic so that it doesn't need to cast or decide whether there's a dataset or datafile - basically for the purposes of "isPartOf" we're ignoring the object itself so we're never dealing with files.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -303,6 +306,46 @@ public static JsonArrayBuilder json(List<DataverseContact> dataverseContacts) {
...
+                        ownerObject.add("identifier", ds.getGlobalId().asString());
```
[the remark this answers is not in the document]
sekmiller: @GPortas I was just trying to keep it consistent across types - like I use identifier for Dataverses even though it's really 'alias' - so if there's no global/persistent id should I call it 'id' as in table id?

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
sekmiller: @MellyGray @GPortas I think I need https://github.com/IQSS/dataverse/issues/10280 to be completed before I can finish this the way it needs to be finished.

sekmiller: Sure. I can do that. The reason I had put them together is that the url for the dataset page currently looks like this: http://localhost:8080/dataset.xhtml?persistentId=doi:10.5072/FK2/UE0DKV&version=2.0

MellyGray: Got it, @sekmiller! I see where you're coming from, but we're aiming to follow an API-first approach to decouple the API implementation from the UI use cases, ensuring our API remains flexible for various applications. We don't know who is going to consume this endpoint, so we should keep these two pieces of information separate to allow them to be used independently. Thanks for implementing the suggestion

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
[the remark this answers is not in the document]
sekmiller: As discussed I will add persistentIdentifier to the dataset object. For collections I will assign 'alias' to identifier and for datasets I will assign the table id to 'identifier'

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -421,6 +421,7 @@ public Response getVersion(@Context ContainerRequestContext crc,
                                @PathParam("versionId") String versionId,
                                @QueryParam("excludeFiles") Boolean excludeFiles,
                                @QueryParam("includeDeaccessioned") boolean includeDeaccessioned,
+                               @QueryParam("returnOwners") boolean includeOwners,
```
GPortas, line 424: Minor detail: rename includeOwners to returnOwners to match the query param name as in other endpoints

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -610,10 +610,12 @@ private Dataset parseDataset(String datasetJson) throws WrappedResponse {
     @GET
     @AuthRequired
     @Path("{identifier}")
-    public Response viewDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String idtf) {
+    public Response getDataverse(@Context ContainerRequestContext crc, @PathParam("identifier") String idtf, @QueryParam("returnOwners") Boolean returnOwners) {
+        Boolean includeOwners = returnOwners == null ? false : returnOwners;
```
GPortas, line 614: You can avoid this using boolean primitive, as I mentioned in the first review.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
GPortas, line 424: I would rename the parameter `includeOwners` to `returnOwners` to match the query param, as in other endpoints.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2892,6 +2916,25 @@ The fully expanded example above (without environment variables) looks like this
...
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+
+  curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/files/:persistentId/versions/$DATASET_VERSION?persistentId=$PERSISTENT_IDENTIFIER&returnOwners=true"
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" "https://demo.dataverse.org/api/files/:persistentId/versions/:draft?persistentId=doi:10.5072/FK2/J8SJZB&returnOwners=true"
+
```
GPortas, line 2937: Updated docs for `getPrivateUrlDatasetVersion` are missing, if I am not wrong.

GPortas (changes requested):
Thanks for applying the requested changes.

Although overall it looks good, I have left a couple of comments.

I have tested the following endpoints and they work as expected:

** Files **
_curl "http://localhost:8080/api/v1/files/5?returnOwners=true"_

The payload includes the _isPartOf_ object, which looks good:

```
"isPartOf":{
  "type":"DATASET",
  "persistentIdentifier":"doi:10.5072/FK2/JSKMZS",
  "identifier":2,
  "version":"1.0",
  "displayName":"test",
  "isPartOf":{
     "type":"DATAVERSE",
     "identifier":"root",
     "displayName":"Root"
  }
}
```

** Dataset version **
_curl "http://localhost:8080/api/v1/datasets/2/versions/:latest?includeFiles=false&returnOwners=true"_

The _isPartOf_ object looks good:

```
"isPartOf": {
      "type": "DATAVERSE",
      "identifier": "root",
      "displayName": "Root"
},
```

** Collections **

First, I requested the root collection (`curl "http://localhost:8080/api/v1/dataverses/root?returnOwners=true"`).

No owners retrieved, which makes sense.

Then, I created a sub collection root called "child", and a sub collection in "child" called "child1", and called the endpoint for "child1". (`curl "http://localhost:8080/api/v1/dataverses/child1?returnOwners=true"`)

The _isPartOf_ looks good:

```
"isPartOf": {
      "type": "DATAVERSE",
      "identifier": "child",
      "displayName": "child",
      "isPartOf": {
        "type": "DATAVERSE",
        "identifier": "root",
        "displayName": "Root"
      }
    },

```

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
```diff
@@ -1667,6 +1667,75 @@ public void testGetFileInfo() {
...
+
+        getFileDataResponse.prettyPrint();
+        getFileDataResponse.then().assertThat()
+                .body("data.label", equalTo("trees.png"))
+                .body("data.dataFile.filename", equalTo("trees.png"))
+                .body("data.dataFile.contentType", equalTo("image/png"))
+                .body("data.dataFile.filesize", equalTo(8361))
+                .statusCode(OK.getStatusCode());
+        
+        getFileDataResponse.then().assertThat().body("data.dataFile.isPartOf.identifier", equalTo(datasetId));
```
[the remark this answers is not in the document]
sekmiller: missed this one, sorry.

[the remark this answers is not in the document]
sekmiller: OK. it's in now...

GPortas (approved):
LGTM. Thank you for applying the changes @sekmiller.

Issue #9992, opened 2023-10-10
Feature Request/Idea: Harvest metadata values out of controlled values


DS-INRA:
**Overview of the Feature Request**
When harvesting, the controlled values field (e.g. Subject) prove very limiting as the value expected for this field have to be in the list determined by the tsv.
It would be best if for harvested datasets these constraints were lifted.

**What kind of user is the feature intended for?**
(Example users roles: API User, Curator, Depositor, Guest, Superuser, Sysadmin)

**What inspired the request?**
Errors harvesting other data repositories. Example (with KindOfData field) :
```
<message>Exception processing getRecord(), oaiUrl=https://repo-int.ortolang.fr/api/oai, identifier=oai:ortolang.fr:e5bb1019-df55-41b8-bfb1-657dd9c37e3a, edu.harvard.iq.dataverse.api.imports.ImportException, Failed to import harvested dataset: class edu.harvard.iq.dataverse.util.json.ControlledVocabularyException (Value 'corpus' does not exist in type 'kindOfData')</message>
```
**What existing behavior do you want changed?**
Allow harvested datasets to have values out of the controlled list from the tsv. 

**Any brand new behavior do you want to add to Dataverse?**
no

**Any open or closed issues related to this feature request?**
no ?

pdurbin:
Related:

- #6559

cmbz:
2023/12/19: Sized at 33. Some discussion about how best to implement the approach.

landreev:
This is my recollection of the recent discussion of this at a tech hour: 

- This is for harvesting only (i.e., a site should be able to configure this as an option - "accept values for such and such field even when they are not in the controlled vocab, when harvesting";
- We already have this notion, of relaxing our metadata requirements when importing harvested content - in the method `ImportServiceBean.doImportHarvestedDataset()`, some constraint violations are relaxed compared to "normal" imports; 
- The way our metadata fields work, it should in fact be possible, to simply save an arbitrary field value for a field that has a controlled vocab., so this part should just work; 
- Important to note that harvested metadata fields are only used for indexing in solr and nothing else.

stevenwinship:
RE: > * The way our metadata fields work, it should in fact be possible, to simply save an arbitrary field value for a field that has a controlled vocab., so this part should just work;

I've tried this but one issue was that it created a new value that gets added to the UI dropdown list. The value wasn't added to the db table but it could be if needed. If we want to add it, it would require re-ordering the list or adding it to the end. 
[...]

PR #10323, opened 2024-02-15
Add harvesting feature to handle missing controlled values

stevenwinship:
What this PR does / why we need it: Allows datasets to be harvested with Controlled Vocabulary Values that existed in the originating Dataverse Project but are not in the harvesting Dataverse Project.

Which issue(s) this PR closes: https://github.com/IQSS/dataverse/issues/9992

Closes https://github.com/IQSS/dataverse/issues/9992
Suggestions on how to test this: See HarvestingClientsIT.java for examples.

Does this PR introduce a user interface change? No

Is there a release notes update needed for this change?: in 9992-harvest-metadata-values-not-in-cvv-list.md

Additional documentation:


src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
```diff
@@ -931,30 +932,30 @@ private String jsonValueToString(JsonValue jv) {
...
                     }
-                    // Only add value to the list if it is not a duplicate 
-                    if (strValue.equals("Other")) {
-                        System.out.println("vals = " + vals + ", contains: " + vals.contains(cvv));
-                    }
-                    if (!vals.contains(cvv)) {
-                        vals.add(cvv);
+                    if (cvv != null) {
+                        // Only add value to the list if it is not a duplicate
+                        if (strValue.equals("Other")) {
```
qqmyers, line 951: Maybe it's time to remove this, or change to logger.fine.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
```diff
@@ -963,11 +964,14 @@ public List<ControlledVocabularyValue> parseControlledVocabularyValue(DatasetFie
...
                 String strValue = json.getString("value", "");
                 ControlledVocabularyValue cvv = datasetFieldSvc.findControlledVocabularyValueByDatasetFieldTypeAndStrValue(cvvType, strValue, lenient);
-                if (cvv == null) {
+                if (cvv == null && !allowHarvestingMissingCVV) {
                     throw new ControlledVocabularyException("Value '" + strValue + "' does not exist in type '" + cvvType.getName() + "'", cvvType, strValue);
                 }
-                return Collections.singletonList(cvv);
+                if (cvv != null) {
+                    vals.add(cvv);
+                }
```
qqmyers, line 972: It's clear that the changes above allow a cvv==null to propagate instead of throwing an exception and the if clause here just skips adding a null value to vals. So this should just ignore non-CVV values w/o blowing up.

I think the intended scope would be to also keep the non-CVV val for indexing which would be ~like an else clause here to add a DatasetFieldValue (as in https://github.com/IQSS/dataverse/blob/f456e515e90e15eed1f1ef6f089e6b764c61df66/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java#L876-L877) which would allow you to keep the value so it could be indexed (without adding a new CVV value to the defined ones which has the side effect you note of adding this term as a new CVV value that can be added in non-harvested datasets.) 

An else clause here wouldn't work as DatasetFieldValue and ControlledVocabularyValue don't have a common subclass (and aren't saved in the same array in the DatasetField class). What might work instead, would be to just call the parsePrimitiveValue method instead of this one when it is a Harvested dataset and allowHarvestingMissingCVV is true. This would make the DatasetField have only DatasetFieldValues, even for the ones where a CVV did exist. The hope is that since these fields are only used for indexing in harvested datasets, there would not be any side effects of having DatasetFieldValues instead of CVVs in this case. I don't think that's proven yet, but I think this is something you could/should try. If it gets messy (i.e. if we make assumptions in the indexing code about whether a DatasetField for a CVV DatasetFieldType has to have CVV values and you can't easily add the same allowHarvestingMIssingCVV flag there), it may be worth reconsidering the scope/priority.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
stevenwinship: Dome

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
stevenwinship: restructured the code to keep the non CVVs in the DatasetFieldValues list instead of skipping them so that the can be indexed later

src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
```diff
@@ -54,6 +55,16 @@ public static void setUpClass() {
         setupCollection();
         
     }
+    @AfterEach
+    public void cleanup() {
```
landreev, line 59: So, this will delete all the clients that the tests were leaving behind? - nice.

landreev:
I'm leaning towards just approving this, not seeing anything to nitpick on. 
@qqmyers thanks for the detailed input on the parsing/indexing earlier. Reading the current code there, it appears to be doing what we want, but please take a quick look/give it a thumbs up, just in case.

src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
stevenwinship: Yes. It needs to be done after each test since there are now 2 harvesting tests using the same data.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java
```diff
@@ -1001,6 +1000,7 @@ public String parseHarvestingClient(JsonObject obj, HarvestingClient harvestingC
         harvestingClient.setMetadataPrefix(obj.getString("metadataFormat",null));
         harvestingClient.setHarvestingSet(obj.getString("set",null));
         harvestingClient.setCustomHttpHeaders(obj.getString("customHeaders", null));
+        harvestingClient.setAllowHarvestingMissingCVV(obj.getBoolean("allowHarvestingMissingCVV", false));
```
landreev, line 1003: @stevenwinship One quick thing I can think of - we could use an equivalent line in JsonPrinter.java, to show this setting in the output of the GET version of the API, when present.

stevenwinship: added

landreev (approved):
Looks great.

sekmiller:
Do we care that there's not a UI change for this? That is when I set up a new client via the UI we aren't allowed to set the new param, nor do we see its value in the in the list of completed clients. Works fine when setting up client via api or if you modify the param directly in the db.

PR #10325, opened 2024-02-15
IQSS/10324-fix granting file access w/o request

### LINKED ISSUE #10324: When granting file permissions (for a group), only one file permission is granted

Start with a datset with >1 restricted files (it can be a draft)
Then go to file permissions and grantt access to multiple files to a user. 
UI does not refect success, then reloading page shows only access to 1 file (and not all).

Confirmed in 6.1 (well ~6.1.1)

qqmyers:
**What this PR does / why we need it**:  To fix granting file access when there was no access request.

At QDR, this appears to be due to an npe: ```
Caused by: java.lang.NullPointerException: Cannot invoke "edu.harvard.iq.dataverse.FileAccessRequest.setStateGranted()" because "far" is null
        at edu.harvard.iq.dataverse.ManageFilePermissionsPage.grantAccess(ManageFilePermissionsPage.java:425)

```

Digging in the code a bit, it looks like a bug introduced with the guestboook-at-request PR: As of that PR, requests are not deleted when granted, they are now given the state granted. At the line in question, the code is looking for a request in the created state associated with the file, so if no request exists, it fails. That should mean that when granting access for one file w/o a request existing (didn't test), it would succeed but the success message/page update would fail. With more than one file, the files after the first access isn't granted. 

Closes #10324

**Suggestions on how to test this**: See the issue and above. Regression testing - it should also work if someone requested access first.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
@@ -422,7 +422,10 @@ public class ManageFilePermissionsPage implements java.io.Serializable {
                     // set request(s) granted, if they exist
                     for (AuthenticatedUser au : roleAssigneeService.getExplicitUsers(roleAssignee)) {
                         FileAccessRequest far = file.getAccessRequestForAssignee(au);
-                        far.setStateGranted();
+                        //There may not be a request, so do the null check
+                        if (far != null) {
+                            far.setStateGranted();
+                        }
                     }
                     datafileService.save(file);
                 }


pdurbin (approved):
I haven't tested the code but this is a very straightforward null check. Approved.

stevenwinship:
I did verify that I had access to all files even though I saw the error in the UI for the file that was previously granted access

qqmyers:
@stevenwinship - is the error you see also in the main codebase? It's hard to see how this PR could change things, but I could image it is an existing bug. If it is unique to this PR I can look again. If it is also in develop already, I'd suggest we start a new issue.

PR #10336, opened 2024-02-22
IQSS 9375 - Retention period

### LINKED ISSUE #9375: Feature Request/Idea: Retention period.
### Retention period functionality overview: 
It should be possible to specify a retention period with the result that the data wil be automatically made unavailable after the period. The motivation is GDPR related; privacy sensitive data in files should be retended. 

### Original motivation or use-case:
In our Dataverse instance (DataverseNL), we have some datasets on medical topics from the Dutch university medical centres. Due to our national legislation, these medical datasets should be deleted after a defined period. (10 or 15 years). Therefore it would be useful to have a metadata field to store information about the retention period, so the datasets that are reaching the end of the retention period can be easily found and proper measures on file level can be taken. 

### Proposed functionality:
The following sections describe the proposed behaviour in the chronological consecutive stages: 

- Retention specification
Allow to specify a retention period for a file. Similar to embargo we should be able to specify the period on a file. Using the File options dropdown.
We should have the 'Retention' option possibly just below the 'Embargo' but before the 'Delete' option. There will then be an 'Edit Retention' dialog (similar to embargo), where you select a day for the retention and an optional reason.
The data is made unavailable on (and after) the day being specified.
When set in different versions, the earliest date 'wins' and will override the others.  

- Pre removal information
Like with an embargo there is information available about the file that has a retention set, but that it is not removed yet. On the File page and in search result the Status message will indicate the retention date. Also the Dataset status will indicate that there is some retention set.  

- File removal/deaccession
When the file is made unavailable it can not be downloaded or previewed. The deletion of the data file on storage (which would make it irreversible) could be later. If the file is 'retended' it is made unavailable for every version. The consequence is that if it is specified in an earlier version it will also be  made unavailable in the later version.  
The files won't be deleted from storage automatically by dataverse, instead there should be another process (manual or otherwise) to do the actual deletion if needed.
It should be possible to configure some kind of action that is automatically executed (like the pre- and post publish workflows) when a file becomes 'retended'. The action itself could be sending a simple email to some sysadmin or update some database, whatever it does is up to the implementer. 

- Post removal information
There should be a landing page (thombstone like) for the files being removed. Especially if a Dataverse has 'PID for files' configured it would be a real problem if we did not provide a decent landing page. The file metadata should remain visible and also the retention information; date and reason.

PaulBoon:
The minimal implementation of the retention feature, which will make a file unavailable after the retention period (or end date) has expired. This is opposite to the embargo feature, because that makes it available. 

Closes #9375

**Additional documentation**:

Work breakdown

- [x] Persistence; database additions for the retention.
- [x] Setting for minimal retention period
- [x] API for set and unset of the retention
- [x] GUI - display of retention
- [ ] GUI - input forms and dialogs
- [ ] documentation
- [ ] unit tests
- [ ] integration test


src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1366,7 +1366,10 @@ public Embargo findEmbargo(Long id) {
         DataFile d = find(id);
         return d.getEmbargo();
     }
-    
+
+    public boolean isActivelyRetended(FileMetadata fm) {
```
qqmyers, line 1370: maybe isExpired? isNoLongerRetained()? (I'm not sure retended is a word and actively rentended/under retension could be confusing as, unlike embargo, retention is the period it's available and the file is not /is no longer retained when this method is true.

PaulBoon: I think the word 'expired' is better than 'no longer', so if I understand you correctly I have to implement  the following:

- While it is 'under retention' the file is  available and in a  'retained' state. 
- After the retention period, the file is unavailable and that is the 'retentionExpired' state. 
- The data that is being set for the retention is then the retention 'end-date' and the file is available on that date but not after that.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1115,6 +1115,8 @@ public SolrInputDocuments toSolrDocs(IndexableDataset indexableDataset, Set<Long
             }
             LocalDate embargoEndDate=null;
             LocalDate end = null;
+            LocalDate retentionEndDate=null;
+            LocalDate start = null;
```
qqmyers, line 1119: A bug in the embargo code? Do start, and end (line 117) need to be defined inside the loop? Looks like they get set when the retention/embargo exist for a file but then aren't ever returned to null for the next file. It looks like that would cause files w/o retention/embargo to get their retention/embargo end dates set (~lines 1167-1172).

src/main/java/propertyFiles/astrophysics.properties
```diff
@@ -40,7 +40,7 @@ datasetfieldtype.coverage.Spectral.Wavelength.description=The minimum and maximu
 datasetfieldtype.coverage.Spectral.MinimumWavelength.description=The minimum wavelength of the spectral bandpass, in meters.
 datasetfieldtype.coverage.Spectral.MaximumWavelength.description=The maximum wavelength of the spectral bandpass, in meters.
 datasetfieldtype.coverage.Temporal.description= Time period covered by the data.
-datasetfieldtype.coverage.Temporal.StartTime.description=Dataset Start Date
+datasetfieldtype.coverage.Temporal.StartTime.description=Dataset End Date
```
qqmyers, line 43: A mistake -  it should stay as Start?

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -4094,6 +4094,18 @@ public void getVersionFiles() throws IOException, InterruptedException {
...
+        // Test Access Status Retention
+        UtilIT.setSetting(SettingsServiceBean.Key.MinRetentionDurationInMonths, "-1");
+        String retentionEndDate = LocalDate.now().plusMonths(240).format(DateTimeFormatter.ofPattern("yyyy-MM-dd"));
+
+        // Create retention for test file 2 (Retention and Public)
+        Response createFileRetentionResponse = UtilIT.createFileRetention(datasetId, Integer.parseInt(testFileId2), retentionEndDate, apiToken);
+        createFileRetentionResponse.then().assertThat()
+                .statusCode(OK.getStatusCode());
+
+        // TODO something with the search criteria...  retentionExpired, no just public because retention has not expired ?
```
qqmyers, line 4107: I think that makes sense - but I see you haven't added any retention related values the FileAccessStatus enum. Perhaps one new RetentionPeriodExpiredvalues/text could be added? If that's done, then adding an unexpired restriction would still show the file in public (or other category) and a Retention expired one would make it unavailable/listed with that filter.

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -735,6 +735,14 @@ Once a dataset with embargoed files has been published, no further action is nee
 
 As the primary use case of embargoes is to make the existence of data known now, with a promise (to a journal, project team, etc.) that the data itself will become available at a given future date, users cannot change an embargo once a dataset version is published. Dataverse instance administrators do have the ability to correct mistakes and make changes if/when circumstances warrant.
 
+Retentions
```
qqmyers, line 738: "Retention" isn't as common a term as "embargo" so I'd suggest changing the Guides info to talk about "retention period". If you want, I can either make suggestions here or make a PR against your branch with changes.

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -735,6 +735,14 @@ Once a dataset with embargoed files has been published, no further action is nee
...
+==========
+
+Support for file-level retentions can also be configured in a Dataverse instance. Retentions make file content inaccessible after the retention end date. This means that file previews and the ability to download files will be blocked. The effect is similar to when a file is restricted except that the retention will end at the specified date without further action and after the retention, requests for file access cannot be made.
+
+Retentions are intended to support use cases where files must be made unavailable (or even destroyed) after a certain period or date.
```
qqmyers, line 743: I'd strengthen this - and in most cases destroyed, e.g. to meet legal requirements

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -4316,6 +4316,18 @@ can enter for an embargo end date. This limit will be enforced in the popup dial
...
 ``curl -X PUT -d 24 http://localhost:8080/api/admin/settings/:MaxEmbargoDurationInMonths``
 
+.. _:MinRetentionDurationInMonths:
+
+:MinRetentionDurationInMonths
++++++++++++++++++++++++++++++
+
+This setting controls whether retentions are allowed in a Dataverse instance and can limit the minimum duration users are allowed to specify. A value of 0 months or non-existent
+setting indicates retentions are not supported. A value of -1 allows retentions of any length. Any other value indicates the minimum number of months (from the current date) a user
+can enter for a retention end date. This limit will be enforced in the popup dialog in which users enter the retention date. For example, to set a ten year minimum:
```
qqmyers, line 4326: Perhaps we should mention the 1000 year hardcoded max? (Same for embargo?)

Should we note that when also allowing embargoes and allowing embargoes longer than the min retention period, files would never be available? (Real question - it's true but not sure it needs to be stated.)

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2498,7 +2498,38 @@ The API call requires a Json body that includes the list of the fileIds that the
...
+``/api/datasets/$dataset-id/files/actions/:set-retention`` can be used to set a retention on one or more files in a dataset. Retentions can be set on files that are only in a draft dataset version (and are not in any previously published version) by anyone who can edit the dataset. The same API call can be used by a superuser to add a retention to files that have already been released as part of a previously published dataset version.
+
+The API call requires a Json body that includes the retention's end date (dateUnavailable), a short reason (optional), and a list of the fileIds that the retention should be set on. The dateUnavailable must be after the current date and the duration (dateUnavailable - today's date) must be more than the value specified by the :ref:`:MinRetentionDurationInMonths` setting. All files listed must be in the specified dataset. For example:
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+  export PERSISTENT_IDENTIFIER=doi:10.5072/FK2/7U7YBV
+  export JSON='{"dateUnavailable":"2051-12-31", "reason":"Standard project embargo", "fileIds":[300,301,302]}'
```
qqmyers, line 2514: Change embargo in reason example

qqmyers (changes requested):
Overall, looks good - fairly parallel to embargo (as intended). I saw one potential bug in embargo that's copied here and made some other suggestions for small changes additions.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -1159,10 +1167,14 @@ public SolrInputDocuments toSolrDocs(IndexableDataset indexableDataset, Set<Long
...
-                    
+                    if(start!=null) {
+                        datafileSolrInputDocument.addField(SearchFields.RETENTION_END_DATE, start.toEpochDay());
+                    }
                     /* Full-text indexing using Apache Tika */
                     if (doFullTextIndexing) {
-                        if (!dataset.isHarvested() && !fileMetadata.getDataFile().isRestricted() && !fileMetadata.getDataFile().isFilePackage()) {
+                        if (!dataset.isHarvested() && !fileMetadata.getDataFile().isRestricted()
+                                && !fileMetadata.getDataFile().isFilePackage()
+                                && fileMetadata.getDataFile().getRetention() == null) {
```
qqmyers, line 1177: This seems somewhat overprotective, especially if we're talking about 10 year + retention periods. In most cases, reindexing is likely to occur fairly soon after. Without inventing a new event trigger mechanism I could see a couple options to try handling this:
* an optional flag to allow fulltext indexing for unexpired files
* a check of the last indexed date whenever  isRetentionExpired is checked - if it hasn't been indexed after the expire date, add it for indexing. This only slows things when expired files are involved and just involves one additional date check. It does mean that the file would remain full-text indexed until the first hit
* add an API call to reindex newly expired files that could be called by a cron script.

Is any of this worth doing now?

doc/sphinx-guides/source/installation/config.rst
PaulBoon: That 1000 years is a hardcoded default value, I don't think that it makes sense to mention it. 
The possibility that you have embargo and retention setup in a way that make the file never accessible for public seems unlikely. We could mention that there can also be an embargo and you have to check that the retention expires after that, but I think people will understand this.

doc/sphinx-guides/source/user/dataset-management.rst
PaulBoon: I will try to change it, and some extra eyeballing is much appreciated

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -735,6 +735,14 @@ Once a dataset with embargoed files has been published, no further action is nee
...
+Actual destruction is not automatically handled, but would have to be done on the storage if needed.
```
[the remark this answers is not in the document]
PaulBoon: These suggestions are beyond this 'minimal implementation' PR, but I do think we should have it in the follow-up.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
PaulBoon: I am extremely cautious when there are legal issues involved, so I would not go for the second point, because it seems not `waterproof`. 

The optional flag together with the added API endpoint would be a good thing for a follow-up PR, but not right now.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
PaulBoon: Ok, if this is a bug then we need to fix it and all Dataverse instances with embargo enabled need to do a reindex.

PaulBoon: embargoEndDate and retentionEndDate must be outside the loop because those track the overall dates for the dataset; the latest embargo and the earliest retention expiration. 
But end and start should be inside the loop, as you suggest.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
PaulBoon: Not sure how to add to that FileaccessStatus; RetentionPeriodExpired makes sense and the fact that is was public or not is not very relevant I think. For the not-expired we might end up with all possible combinations with public,restricted and embargo.

PaulBoon: Just checked that I do have an 'Acces' filter with 'Retention Expired (1)' at the left side when I make a file retention period expire. If I select it I get that one file with the search breadcrumb: 'Access: Retention Expired' above the results. This is without that FleAccessStatus enum.

PaulBoon: I am tempted to also add 'RetentionNotExpired' resulting in only those that have a retention set, but that has not expired. Would be useful in the Integration Test, or when you want to know which ones are near to expiring if you sort the result.

doc/sphinx-guides/source/user/dataset-management.rst
PaulBoon: @qqmyers Do we need to change 'Retention Expired" to "Retention Period Expired" in all GUI strings as well?

qqmyers: Yeah, I think that would be better.

PaulBoon: @qqmyers There are still  some lines that I am no sure about; when we have 'under retention' as in "Is under retention until", should that be changed to 'retained'?

src/main/webapp/file.xhtml
```diff
@@ -552,9 +553,9 @@
                                                         <td>#{FilePage.file.publicationDateFormattedYYYYMMDD}
                                                         </td>
                                                     </tr>
-                                                    <tr id="fileEPublicationDateBlock" jsf:rendered="#{FilePage.file.released or !(empty FilePage.file.embargo)}">
+                                                    <tr id="fileEPublicationDateBlock" jsf:rendered="#{!(empty FilePage.file.embargo)}">
```
PaulBoon, line 556: Removed because the Embargo End date was also rendered when there was no embargo, but it was published (released)

doc/sphinx-guides/source/user/dataset-management.rst
PaulBoon: Changed it to 'retained' seems best

src/main/webapp/file.xhtml
qqmyers: I don't understand this change. The Publication Date is supposed to show in both cases. It will either be the real publication date or will show the embargo end date if it is set.

PaulBoon: The publication date is shown above in the element with `id="metadataReleaseDateBlock"`. Also the element with `id="fileEPublicationDateBlock"` has the label that states it is about embargo, so why should it be shown if there is no embargo. 
When I tested the retention period I got both retention and embargo elements, just showing the publication date, which was weird. Therefore I changed it, but I wonder why I was the only one seeing it.

qqmyers: It looks like you've changed the label in the line 558 below - it would normally say publication date. I think the design guidance was to always show the metadata release date and the publication date all the time. With no embargo, those dates are the same. With an embargo, the pub date is delayed.

PaulBoon: Just changed it, but the composition of all those options in one element makes it a bit complicated; al lot of those `#{}` combined.

qqmyers (approved):
Looks good. Everything has been addressed.

src/main/webapp/file.xhtml
```diff
@@ -552,11 +553,16 @@
...
                                                         <th scope="row">
                                                             #{bundle['file.metadataTab.fileMetadata.publicationDate.label']}
                                                         </th>
-                                                        <td>#{!(empty FilePage.file.embargo) ? FilePage.embargoPhrase: ''} #{!(empty FilePage.file.embargo) ? FilePage.file.embargo.dateAvailable : FilePage.file.publicationDateFormattedYYYYMMDD}
+                                                        <td>#{!(empty FilePage.file.embargo) ? FilePage.embargoPhrase: ''} #{!(empty FilePage.file.embargo) ? FilePage.file.embargo.dateAvailable : ''}
+                                                            #{!(empty FilePage.file.embargo) and !(empty FilePage.file.retention) ? '; ': ''}
+                                                            #{!(empty FilePage.file.retention) ? FilePage.retentionPhrase: ''} #{!(empty FilePage.file.retention) ? FilePage.file.retention.dateUnavailable: ''}
```
qqmyers, line 562: Can the next 3-4 lines be combined? They all check !(empty FilePage.file.retention) and add a sequence of 
`FilePage.retentionPhrase FilePage.file.retention.dateUnavailable , bundle['retention.enddateinfo']` if true and '' if not.

PaulBoon: I used `concat` to avoid that duplication

sekmiller:
Hey @PaulBoon Can I trouble you to update this branch from our Development branch? Specifically you are missing a couple of sql update scripts and also you will need to rename the script in this branch as described here https://guides.dataverse.org/en/6.2/developers/sql-upgrade-scripts.html#renaming-sql-upgrade-scripts Please note that our next release will be version 6.3 so your script should take that into account. Thanks!

PaulBoon:
@sekmiller hope it's OK now.

sekmiller:
Thanks @PaulBoon. I was away yesterday, but I will make time today to take a look.

sekmiller:
Screen Shot 2024-04-16 at 10 43 45 AM

When I tried to shorten the retention period I got this error in the popup. That date range seems to be out of whack.

sekmiller:
@PaulBoon One thing that I noticed - if I use the multiple file download button when one of the files I have selected is no longer available due to the retention expiration the warning message refers to "Restricted Files" 
Screen Shot 2024-04-16 at 4 07 13 PM

sekmiller:
Also can you add a release note that briefly describes the feature and notes that the schema.xml must be updated in the upgrade process. 

Thanks

PaulBoon:
> @PaulBoon One thing that I noticed - if I use the multiple file download button when one of the files I have selected is no longer available due to the retention expiration the warning message refers to "Restricted Files" Screen Shot 2024-04-16 at 4 07 13 PM

@sekmiller The best solution I can think of is to change these messages in a way that they also cover these files with an expired retention. A multiple selection can contain both types and it would be difficult to adapt the message to that dynamically. 
 In both cases the file is not available, but for a different reason.

qqmyers:
FWIW: There's a similar issue w.r.t. files that are only available for Globus transfer and I've made changes to these messages and display logic in #10451. I'd suggest adding retention period expiration into those changes messages rather than tweaking the existing ones. (That PR is also in QA - not sure if it can get merged and picked up here in a reasonable time or not.

qqmyers:
@PaulBoon - FYI: we had a discussion in the daily standup mtg and think the way to go is to get #10451 merged and have you update this PR to include it from develop, and you or I can then make changes. For the changes themselves, I think the idea was to go with the simpler title of "Inaccessible Files Selected" and then include retention period expired as one of the options in the descriptive text.

PaulBoon:
@qqmyers Just added the release notes, and as with most of this PR I reused yours. I will wait for that other PR to be merged while I am on a short vacation (just a week).

sekmiller:
@PaulBoon Hi Paul, I have a few observations/change requests
1) For the api success message can we say "File(s) retention information updated" instead of "Files were retentioned"
2) If I put in a date that isn't a date (February 30) in the api, it doesn't update but there's no error response. (response is just {})
3) If I'm logged in as a superuser I'm not able to set or update the retention of a published file - which I am able to do via the api - did you consider allowing this via the UI?

PaulBoon:
> Screen Shot 2024-04-30 at 3 42 03 PM
> 
> On the dataset page file list - It looks like you should add a space or punctuation between the deposit date and "Draft:..."

This is strange, because I don't see that happening. Maybe your CSS is not the same? If I inspect the page I see that this 'Draft: .." string is inside a `span` with `class=retentionRight`, and that has a `margin-left` of `10px;`.

PaulBoon:
> @PaulBoon Hi Paul, Thanks for the updates. I have a few observations/change requests
> 
>     1. For the api success message can we say "File(s) retention information updated" instead of "Files were retentioned"
> 
>     2. If I put in a date that isn't a date (February 30) in the api, it doesn't update but there's no error response. (response is just {})
> 
>     3. If I'm logged in as a superuser I'm not able to set or update the retention of a published file - which I am able to do via the api - did you consider allowing this via the UI?

@sekmiller About the observations:
1. I agree that those messages need improving; will make it "File(s) retention period has been set or updated". 
2. Will have to look into that, I did see this kind of behavior with other API endpoints, so maybe it was difficult to handle properly. 
3. This behavior was copied from the embargo implementation, not only to be efficient with the limited development time, but also to keep similar features behave similar. 
If we discover that this happens a lot and we want to have it in the UI, we could handle this in an separate issue/PR.

PaulBoon:
@sekmiller Just commited additions to the input handling of that set-retention API end point. 
It now gives informative responses when: 
1. The JSON is invalid
2. The dateUnavailable is not specified
3. The dateUnavailable is not a string
4. The dateUnavailable has not a valid date format

sekmiller:
@PaulBoon Thanks for the updates they look good. Your error message for improper json led me to try leaving some data out of a well-formed json object. This is what I found:

If I use the api with a json object that doesn't include "fileIds", I get a success message and a record is written to the retention table, but of course there are no records in the datafile table with that retention_id.

If I leave out "reason" I get an empty response {} and nothing seems to happen to the retention table.

If I leave out "dateUnavailable" I get a proper error message - no change needed here.

PaulBoon:
At some point the Dataverse APIs should have some formal specification which we can hopefully use to do some automated 'schema' validation against. This sort of 'manually' handling of all possible issues with the input is cumbersome. But until then...

sekmiller:
Thanks for hanging in there and making the updates. I will merge it shortly.

PR #10354, opened 2024-03-05
Dataset version API change - Adds latestVersionPublishingStatus

### LINKED ISSUE #10330: Add latestVersionPublishingStatus field to dataset version payload

This parameter is very necessary in many use cases. By including it in the payload of a specific Dataset version we avoid to make an extra call to the API.

API User

Frontend devs discussion

None

Add latestVersionPublishingStatus field to dataset version payload

- https://github.com/IQSS/dataverse/issues/10280

jp-tosca:

SPA request for change on the API

Closes #10330

Keep an eye on **DRAFT** and _DEACCESSIONED_ versions of a dataset, these should show regardless of the user permissions. 

`curl "$SERVER_URL/api/datasets/$ID/versions/$VERSION?includeFiles=false"`


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4432,8 +4459,12 @@ public Response getDatasetVersionCitation(@Context ContainerRequestContext crc,
                                               @QueryParam("includeDeaccessioned") boolean includeDeaccessioned,
                                               @Context UriInfo uriInfo,
                                               @Context HttpHeaders headers) {
+        boolean checkFilePerms = false;
+        boolean checkUserPerms = true;
```
GPortas, line 4463: This variable _checkUserPerms_ is never used.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4432,8 +4459,12 @@ public Response getDatasetVersionCitation(@Context ContainerRequestContext crc,
                                               @QueryParam("includeDeaccessioned") boolean includeDeaccessioned,
                                               @Context UriInfo uriInfo,
                                               @Context HttpHeaders headers) {
+        boolean checkFilePerms = false;
```
GPortas, line 4462: We can simplify this and remove checkFilePerms as it is always `false`.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2712,33 +2729,43 @@ public static <T> T handleVersion(String versionId, DsVersionHandler<T> hdl)
      * includeDeaccessioned default to false and checkPermsWhenDeaccessioned to false. Use it only when you are sure that the you don't need to work with
      * a deaccessioned dataset.
      */
-    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers) throws WrappedResponse {
+    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds,
+            UriInfo uriInfo, HttpHeaders headers) throws WrappedResponse {
         //The checkPerms was added to check the permissions ONLY when the dataset is deaccessioned.
-        return getDatasetVersionOrDie(req, versionNumber, ds, uriInfo, headers, false, false);
+        boolean checkFilePerms = false;
+        boolean includeDeaccessioned = false;
```
GPortas, line 2736: Same here, I would remove these variables and make the call simpler.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -432,6 +438,10 @@ public static JsonObjectBuilder json(DatasetVersion dsv, List<String> anonymized
                 .add("publicationDate", dataset.getPublicationDateFormattedYYYYMMDD())
                 .add("citationDate", dataset.getCitationDateFormattedYYYYMMDD());
 
+        if(latestDsv != null) {
+            bld.add("latestVersionPublishingStatus", latestDsv.getVersionState().name()).add("versionNote", latestDsv.getVersionNote());
```
GPortas, line 442: What is the purpose of `versionNote` here? I'm asking to know better the intention here.

If you are referring to the latest version note, I would rename it to `latestVersionNote` to avoid confusion.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2712,33 +2729,43 @@ public static <T> T handleVersion(String versionId, DsVersionHandler<T> hdl)
...
     
     /*
      * checkPermsWhenDeaccessioned default to true. Be aware that the version will be only be obtainable if the user has edit permissions.
      */
-    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers, boolean includeDeaccessioned) throws WrappedResponse{
-        return getDatasetVersionOrDie(req, versionNumber, ds, uriInfo, headers, includeDeaccessioned, true);
+    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds,
+            UriInfo uriInfo, HttpHeaders headers, boolean includeDeaccessioned) throws WrappedResponse {
+        boolean checkFilePerms = true;
+        return getDatasetVersionOrDie(req, versionNumber, ds, uriInfo, headers, includeDeaccessioned, checkFilePerms);
```
GPortas, line 2746: Same here, I would remove the `checkFilePerms` variable.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2712,33 +2729,43 @@ public static <T> T handleVersion(String versionId, DsVersionHandler<T> hdl)
...
     }
 
     /*
      * Will allow to define when the permissions should be checked when a deaccesioned dataset is requested. If the user doesn't have edit permissions will result in an error.
      */
-    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds, UriInfo uriInfo, HttpHeaders headers, boolean includeDeaccessioned, boolean checkPermsWhenDeaccessioned) throws WrappedResponse {
+    private DatasetVersion getDatasetVersionOrDie(final DataverseRequest req, String versionNumber, final Dataset ds,
```
GPortas, line 2752: I would try to avoid these random changes to the code indentation, even more so if there are no changes to the logic.

Also, if your intention is to shorten the parameter line, I would break the line like we do here: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L451. There is no strict consensus on styling, as far as I know, but this is the style I see in the code more often.

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -947,6 +947,7 @@ public Response getFileCitationByVersion(@Context ContainerRequestContext crc, @
             DataverseRequest req = createDataverseRequest(getRequestUser(crc));
             final DataFile df = execCommand(new GetDataFileCommand(req, findDataFileOrDie(fileIdOrPersistentId)));
             Dataset ds = df.getOwner();
+
```
GPortas, line 950: I would avoid these blank additions that are not related to any logic change.

GPortas (commented):
@jp-tosca 

I still have to test the endpoint, but I have already left comments related to the code.

Additionally, integration tests are missing. Could you add some to cover the new changes and different possible scenarios?

Thank you!

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -429,18 +429,35 @@ public Response getVersion(@Context ContainerRequestContext crc,
...
+                    checkPerms);
+            
+            DatasetVersion latestDatasetVersion = null;
+            boolean deaccesionedLookup = true;
+            //Check perms is false since we are never going to retrieve files, we are just getting the status of the version.
+            checkPerms = false;
+            if(versionId != DS_VERSION_LATEST){                
+                latestDatasetVersion = getDatasetVersionOrDie(req, DS_VERSION_LATEST, dst, uriInfo, headers,
+                        deaccesionedLookup,
+                        checkPerms);
```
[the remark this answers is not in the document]
jp-tosca: The checkPerms is only false in the case that the request skip the files, the default is true since the default behavior of the API is to bring the files for the user.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: I am removing this, this was a leftover from when I tried to skip checking the user permissions which as the conversation in slack is not what is intended.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: In this case is false because this endpoint never accesses the files, I added the variable just to add context when reading the code so people don't have to go to the definition of the method to know what true/false does.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: I added the variable just to add context when reading the code so people don't have to go to the definition of the method to know what true/false does. (same from above, I can remove them but I like to give some context of what a parameter does without you having to dig into the method)

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
jp-tosca: I am not fully aware of the versionNote, this is how the regular version is attached to the JSON and since the latestVersion is retrieved in the same way I decided to leave it exactly than what is on 431 probably @qqmyers can enlighten us a little bit more here 😆

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: Same, these are just to make it more readable.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
jp-tosca: I will change it that way 👍🏼

jp-tosca:
I will check if there are any changes required to the integration tests.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -429,18 +429,35 @@ public Response getVersion(@Context ContainerRequestContext crc,
...
+            } else {
+                latestDatasetVersion = requestedDatasetVersion;
+            }
+  
+            if (requestedDatasetVersion == null || requestedDatasetVersion.getId() == null) {
```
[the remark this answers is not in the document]
jp-tosca: I had to remove the condition because I found a bug when people would ask for the latest version specifically with the files depending on the permissions but besides that with the last update the code should work as intended, I added a few comments on the integration tests to explain the specific situations but I can explain more in detail, the logic is a bit complicated.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
jp-tosca: Resolved after our latest slack call, thanks! 👍🏼

GPortas:
I just met with @jp-tosca to discuss the comments!

Will accept the PR once the changes are pushed.

GPortas (approved):

jp-tosca:
-Branch updated on March 28th

PR #10370, opened 2024-03-12
IQSS/10369-support legacy PID settings

### LINKED ISSUE #10369: Multipid ignores legacy settings/isn't backward compatible
**What steps does it take to reproduce the issue?** Deploying on an existing instance with legacy (pre-multipid) PID db settings fails and shouldn't. PR to follow.

dev branch

qqmyers:
**What this PR does / why we need it**: A late reshuffle of the code added a return after looking for the new PID provider settings, bypassing the check of legacy settings. This PR removes the return so that legacy settings are checked again. 

Closes #10369

**Suggestions on how to test this**: Deploy to an existing db with legacy settings. Prior to the PR the ConfigCheck will cause the deploy to fail. With the PR, it works. (If you don't have an existing setup, run a new setup, add the legacy settings, e.g. for a FAKE provider, and then remove the -Ddataverse.pid.providers= and -Ddataverse.pid.default-provider= jvm options. Trying to restart at that point will fail, but after the PR you can deploy.)

**Additional documentation**:


pdurbin (changes requested):
This PR has 121 commits to remove one line.

Can we please have a fresh PR (or force push this one) to have one commit?

qqmyers:
Closed in favor of #10374. FWIW, the additional commits in this PR are just the ones from the multipid branch which are already merged into develop.

Issue #9276, opened 2023-01-11
Feature Request/Idea:Implement mechanism to support integration with OntoPortal


jonquet:
**Overview of the Feature Request**

This issue is a connection to the correspondent issue in the OntoPortal repo: https://github.com/ontoportal/ontoportal-project/issues/16

We would like a way to ease the selection of vocabulary/ontology term in Dataverse using OntoPortal instances. 
(e.g., BioPortal or AgroPortal)
For the moment the current interface requires 4 fields (example from DataINRAE):

![image](https://user-images.githubusercontent.com/7379183/211806905-c6749469-b89f-4417-86ef-9da890dbc46e.png)

In [Data INRAE](https://entrepot.recherche.data.gouv.fr/dataverse/inrae) we can replace the fields above with only one, the concept search below:

<img width="484" alt="image" src="https://user-images.githubusercontent.com/7379183/211807828-803c2a8b-2839-4e79-80db-0d799f674e59.png">

**What existing behavior do you want changed?**

- [ ] Implement a mechanism to change the 4 fields above by only one field with autocompletion with selected vocabularies hosted in an OntoPortal instance
- [ ] Implement a mechanism to call the correspondant OntoPortal Annotator on the title/abstract fields of the dataset to pre-propose some ontology/vocabulary terms in the UI that the user can only keep or discard.

**What kind of user is the feature intended for?**
Depositor
The key idea is to ease depositon going from n*4 fields for n terms to only one field with autocompletion

**What inspired the request?**
Tons of request for this. 
This is being done in the context of Horizon Europe project FAIR-IMPACT T4.5

**Any related open or closed issues to this feature request?**
https://github.com/ontoportal/ontoportal-project/issues/16

jonquet:
Colleagues on OntoPortal side: @syphax-bouazzouni and Dataverse (at INRAE side) @DS-INRA

qqmyers:
I'd suggest developing an [external vocabulary script ](https://github.com/gdcc/dataverse-external-vocab-support)for this. The compoundDemo field in the examples there seems to be close to what you want - retrieving values from a Skosmos server rather than OntoPortal but it shows the user one field and fills in the four required by Dataverse.

luddaniel:
Hi everyone, as we started to work on implementing Ontoportal support for external controlled vocabulary, we wanted to start a list of points that was need to face to successed.

- [ ] Having the capability to use an apikey on backend side (`registerExternalTerm` function) 
[...]

PR #10371, opened 2024-03-13
10288 Add keywordTermURI metadata in keyword block

### LINKED ISSUE #10288: Feature Request/Idea: Add "Term URI" metadata in Keyword block
The proposal of the issue is to integrate a new metadata "Term URI" to the keyword block to facilitate the integration of controlled vocabulary services. 

Example with entrepot.recherche.data.gouv.fr:

capture_keyword_block

All depositors

This is related to issue #9276 and the [Google group discussion](https://groups.google.com/g/dataverse-community/c/4eTd3kfJeBU/m/gcgjM1ZXAAAJ)

Adding a “Term URI” metadata to the “keyword” block

- Issue #9276
- PR #10145 
- [Google group discussion](https://groups.google.com/g/dataverse-community/c/4eTd3kfJeBU/m/gcgjM1ZXAAAJ)

stevenferey:

To continue on the subject of [OntoPortal integration
](https://github.com/IQSS/dataverse/issues/9276)

Closes #10288 

In addition to the addition of the `keywordTermURI` field, we added a script allowing administrators to optionally migrate data from `keywordValue` to `keywordTermURI` when the data starts with "http".

- The keyword metadata block is modified:

Capture d’écran du 2024-02-28 17-51-00

- Before data migration, here is an example with a `keywordValue` containing data starting with "http" :
Capture d’écran du 2024-03-13 10-48-35

- After optional data migration :
Capture d’écran du 2024-03-13 10-48-52

Yes, a proposal is visible in the file doc/release-notes/10288-add-term_uri-metadata-in-keyword-block.md


jggautier:
Heya @stevenferey. Have you had a chance to check out the guidelines for metadata text at https://docs.google.com/document/d/1tY5t3gjrIgAGoRxVMWQSCh46fnbSmnFDLQ7aLkNLhJ8?

src/main/java/propertyFiles/citation.properties
```diff
@@ -22,9 +22,10 @@ datasetfieldtype.dsDescriptionValue.title=Text
 datasetfieldtype.dsDescriptionDate.title=Date
 datasetfieldtype.subject.title=Subject
 datasetfieldtype.keyword.title=Keyword
+datasetfieldtype.keywordTermURI.title=Term URI
 datasetfieldtype.keywordValue.title=Term
 datasetfieldtype.keywordVocabulary.title=Controlled Vocabulary Name
-datasetfieldtype.keywordVocabularyURI.title=Controlled Vocabulary URL
+datasetfieldtype.keywordVocabularyURL.title=Controlled Vocabulary URL
```
qqmyers, line 28: Why does this need to change to URL from URI? (While you're adding Term `URI`?) Minimally I think we should be consistent and, unless we're sure that there are no URNs in use, sticking with URI might be the better option. (I do see that we focus on URLs in the descriptions and the idea of web locations for things instead of talking about URIs as identifiers that might also be resolvable on the web. I'm not sure what the best way to resolve that is, or that it needs to be handled in this PR.) @jggautier - any thoughts?

jggautier: I'm not sure and might not understand this enough. But maybe I could help see if there _are_ URNs in use?

In case this info is helpful, I think we used "keywordVocabularyURI" for the database name because [DDI Codebook uses "vocabURI"](https://ddialliance.org/Specification/DDI-Codebook/2.5/XMLSchema/field_level_documentation_files/schemas/codebook_xsd/elements/keyword.html). And maybe we thought that most people would be more familiar with the term "URLs", so that's what we used for the label people see on the page and in the tooltips. Of course this was all before I joined Dataverse.

luddaniel: Update is done @qqmyers :)

src/main/resources/db/migration/V6.2.0.1__10288-update_keywordVocabularyURI_metadata.sql
```diff
@@ -0,0 +1,5 @@
+-- update of the "keywordVocabularyURI" metadata to make it consistent with its name "Controlled Vocabulary URL"
```
qqmyers, line 1: This is obsolete now?

qqmyers: The PR description also still mentions changing from URI to URL (not in the code, but might confuse in review or QA)

qqmyers:
@jggautier - any reason this can't get on the board now to move forward?

jggautier:
Hmm maybe. This could get on the board in the next sprint's "This Sprint" column while we wait to hear back from @stevenferey about the tooltip text and how it does or doesn't follow [the guidelines for metadata text](https://docs.google.com/document/d/1tY5t3gjrIgAGoRxVMWQSCh46fnbSmnFDLQ7aLkNLhJ8)? Or should that be resolved first?

I'd like to make sure that those guidelines are easy to find and are helpful when folks are working on the text associated with metadata, like the field labels and tooltips shown in the UI.

@stevenferey, I asked about this a few weeks ago because the current tooltip text for this Term URI field makes me think that  you haven't seen those guidelines because they're hard to find or that the guidelines could be clearer.

stevenferey:
Sorry for the late response,
I was able to read the document after your message.

Is this description better?
`Keyword URI points to the web presence of the term. Enter an absolute URI, e.g. starting by https://`

We listen to your suggestions for improvement.

jggautier:
Hey @stevenferey. Thanks for taking a look!

The more important thing is to avoid adding instructions in the tooltip for how depositors should add metadata, since people looking for data also see those tooltips and that information isn't relevant for them. And we hope that the watermark and the field validation will ensure that depositors are entering what you expect.

So I think we should remove the second sentence about entering an absolute URI. And the rest of the changes I suggested below are mostly style changes for consistency:

A URI that points to the web presence of the Keyword Term

jggautier:
It would be cool if we could encourage folks to use a tool that would apply our style guidelines to their text. 

The paid versions of Grammarly and a less popular tool called [ProWritingAid](https://prowritingaid.com) say they can do this. Maybe their are free tools, too!

stevenferey:
Thank you for this feedback,

I pushed a label update to follow the recommendation.

jggautier:
@qqmyers after you asked me last if this could get on the board to move forward, I asked if it could while we waited to hear back about the tooltip text or if that should be addressed first.

I think the issue with that text's been addressed, so I can't think of any reason why this couldn't get on the board to move forward.

src/main/java/edu/harvard/iq/dataverse/export/openaire/OpenAireExportUtil.java
```diff
@@ -446,26 +446,32 @@ public static void writeSubjectsElement(XMLStreamWriter xmlw, DatasetVersionDTO
...
-                                    schemeURI = next.getSinglePrimitive();
+                                    keywordVocabURI = next.getSinglePrimitive();
                                 }
                             }
 
                             if (StringUtils.isNotBlank(subject)) {
                                 subject_check = writeOpenTag(xmlw, "subjects", subject_check);
-                                writeSubjectElement(xmlw, subjectScheme, schemeURI, subject, language);
+                                // we prioritize the keywordTermURI metadata to populate schemeURI
+                                writeSubjectElement(xmlw, subjectScheme, StringUtils.isNotBlank(keywordTermURI) ? keywordTermURI : keywordVocabURI, subject, language);
```
qqmyers, line 474: @stevenferey  - The DataCite Schema includes both a schemeURI and valueURI attributes - should this now include both rather than swapping the term URI into the schemeURI attribute?

stevenferey: @qqmyers, The branch is updated with the modification of the OpenAire export. Thanks a lot

qqmyers:
@stevenferey - I think the change looks good. There is a (probably unrelated) failing test though that may be related to destroying a dataset - can you merge with develop again to pick up #10566.  Details:  expected: <404> but was: <500>
Stacktrace
org.opentest4j.AssertionFailedError: expected: <404> but was: <500>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:531)
	at edu.harvard.iq.dataverse.api.HarvestingServerIT.testSingleRecordOaiSet(HarvestingServerIT.java:630)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)

luddaniel:
Hello @qqmyers, as @stevenferey is in holiday for a full month, I will handle the next exchanges.
This branch is already up to date, and by jumping into debug mode I can tell that the expected should be <500>.
Screenshot from 2024-05-29 15-19-51
`Caused by: jakarta.persistence.NoResultException: getSingleResult() did not retrieve any entities.` handled error will throw `.add("code", 500)` response. 
By fast crawling git blame I would say that no code has been modified for the last 3 months so I don't know when this IT code started to fail.
I did go ahead and fixed the error in the last commit a79cdbf, tell me if you want to do something else.

qqmyers:
@luddaniel - thanks. We're seeing the test fail in other branches, one with only docs changes, so definitely unrelated. That said, we're looking to see if we can understand why it started failing recently and will probably want to pick up your try/catch fix. I'll let you know if there's anything else to do.

pdurbin:
> will probably want to pick up your try/catch fix

Yes, I cherry-picked it here:

- #10601

Thanks, @luddaniel!

pdurbin:
@stevenferey we cherry-picked a commit from this PR into #10601 and merged it. Can you please merge develop into this PR? Thanks.

qqmyers (approved):
I think this is done at this point. The last build failed with the AWS instance not starting but probably not related. (Should rerun at some point.)

stevenwinship:
Issue with "Data migration to the new `keywordTermURI` field"
No validation takes place until someone tries to edit the metadata in the UI

Before update:
Screenshot 2024-06-14 at 9 55 51 AM
After Update:
Screenshot 2024-06-14 at 9 57 24 AM
After trying to update the metadata:
Screenshot 2024-06-14 at 9 58 33 AM

qqmyers:
I think this is how all fields of type url work, i.e. not related to the one field added in the PR.

PR #10373, opened 2024-03-13
log file with a different instance root directory

### LINKED ISSUE #10372: log file with a different instance root directory
If you have a different instance root, you cannot execute the commands as you describe. (https://guides.dataverse.org/en/5.14/admin/metadataexport.html#batch-exports-through-the-api)
Because the log directory has been permanently programmed.
For example here:
https://github.com/IQSS/dataverse/blob/ec1b174dbc89edd4e40c9c61760523e800497754/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java#L702
(I have already made an improvement and I will upload it right away.)

* When does this issue occur?
curl http://localhost:8080/api/admin/metadata/exportAll
curl http://localhost:8080/api/admin/metadata/reExportAll
curl http://localhost:8080/api/admin/metadata/clearExportTimestamps

* What happens?
It is not executed at all and an error occurs.

It is the same source code in 5.13 to 6

**Screenshots:**

*

BenediktMeierUIT:
log file with a different instance root directory
Closes #10372 

**Additional documentation**:


pdurbin:
@BenediktMeierUIT thanks for the pull request. Out of curiosity, what is your instance root?

Louis-wr:
> @BenediktMeierUIT thanks for the pull request. Out of curiosity, what is your instance root?

Benedikt as been working with us on dataverse.no

pdurbin:
Cool. I'm still wondering what you use as an instance root, but no rush. 😄 Oh! Maybe you're using Docker? 🤔 

We recently make a similar change at 8efb964 and as discussed [in Slack](https://iqss.slack.com/archives/C010LA04BCG/p1690921171172649?thread_ts=1690912135.247139&cid=C010LA04BCG), we know there are more places to fix, as you've done. Thanks!

BenediktMeierUIT:
Yes, this is a Docker.
And I solved the errors straight away. 
https://github.com/IQSS/dataverse/commit/8efb96487d181f83726ac5a338a13692b1710850

I'm not quite sure about another point. Why it was done differently.
https://github.com/IQSS/dataverse/blob/ec1b174dbc89edd4e40c9c61760523e800497754/src/main/java/edu/harvard/iq/dataverse/DatasetVersionServiceBean.java#L896

https://github.com/IQSS/dataverse/blob/ec1b174dbc89edd4e40c9c61760523e800497754/src/main/java/edu/harvard/iq/dataverse/batch/util/LoggingUtil.java#L70

(I don't have access to Slack :-( )

pdurbin:
Sorry, can you please spell out a little more how it's done differently? From a quick look it seems more or less the same to me.

BenediktMeierUIT:
Sorry... 
It's the same, but it's inconsistent.
My question was/is, what is the guideline and how do you want it?

BenediktMeierUIT:
At that point, it's done differently again:
https://github.com/IQSS/dataverse/blob/5caf436c6321a8a26749d3dbf8e9ea4dfcdde5f7/src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvesterServiceBean.java#L91
I was going to change the line like this:
`public static final String DATAVERSE_HARVEST_STOP_FILE="stopharvest_";`
and in the place:
https://github.com/IQSS/dataverse/blob/5caf436c6321a8a26749d3dbf8e9ea4dfcdde5f7/src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvesterServiceBean.java#L402
`String stopFileName = System.getProperty("com.sun.aas.instanceRoot") + File.separator + "logs" + File.separator + DATAVERSE_HARVEST_STOP_FILE + harvestingClient.getName() + "." + pid; `

poikilotherm:
This should be a JVM option defaulting to the system property.

pdurbin:
It's funny, just today I was in DatasetServiceBean.java debugging something and I copied the fix from this PR because I'm in Docker and was hitting a problem without it.

I'd say we definitely want this fix.

And yes, a JVM option/MicroProfile Config setting eventually. In this PR or later.

Louis-wr:
Yes, A JVM option/MicroProfile Config  would be nice in the future but for now we should avoid overreaching this pull request and  keep it confine to the bug fix that it is.
In the near future once this is fixed it will be easy to replace all the `System.getProperty("com.sun.aas.instanceRoot") `with the appropriate newly created JVM option/MicroProfile Config.

pdurbin:
Yes, sounds good to defer adding a JVM/MPConfig option until later.

poikilotherm:
Adding the option is almost a one liner. Do you want me to create a small PR to this PR with it? @pdurbin if you leave a review after, I can do QA and we ship this one fast.

BenediktMeierUIT:
If you give me the JVM/MPConfig name, I can change the format to what you want.

pdurbin:
Sure, I'm happy to do a review and let @poikilotherm do the QA.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -699,7 +699,7 @@ public void exportAllDatasets(boolean forceReExport) {
         Integer countError = 0;
         String logTimestamp = logFormatter.format(new Date());
         Logger exportLogger = Logger.getLogger("edu.harvard.iq.dataverse.harvest.client.DatasetServiceBean." + "ExportAll" + logTimestamp);
-        String logFileName = "../logs" + File.separator + "export_" + logTimestamp + ".log";
+        String logFileName = System.getProperty("com.sun.aas.instanceRoot") + File.separator + "logs" + File.separator + "export_" + logTimestamp + ".log";
```
pdurbin, line 702: I can vouch for the need for this change. Without it, I couldn't get some code I was playing with to execute. I was exporting some JSON for troubleshooting: https://github.com/pdurbin/dataverse/commit/8adb28d67522936b7e3c6f0aa98e8ab720a1982a

poikilotherm: Sounds like within a container, the application has a different working directory than in a classic deployment. Which means `../logs` may point to some place that does not exist / the app user has no permissions to write to.

poikilotherm (changes requested):
With the explanation via DM and in https://github.com/IQSS/dataverse/pull/10373/files#r1535665937, I understood better why this change is necessary. 

As we should be thorough with this change, can you fix the missing occurrences @BenediktMeierUIT ?

Those would be in `dataverse.authorization.AuthFilter` and `dataverse.harvest.client.HarvesterServiceBean`.

BenediktMeierUIT:
That's right... 
I have changed it.

poikilotherm (approved):
Looks good to me now, thanks for the change! @pdurbin stage is yours :smile:

pdurbin:
Tests are passing as as I mentioned above these changes are very similar to changes I made in https://github.com/pdurbin/dataverse/commit/8adb28d67522936b7e3c6f0aa98e8ab720a1982a

Merging.

PR #10377, opened 2024-03-13
Doc Update: Fix old DataCite 3.1 reference

qqmyers:
The Guides have a reference to DataCite 3.1 that wasn't updated when we changed to support DataCite 4.0.


jggautier:
I checked other parts of the guides to see if other versions of the DataCite schema are mentioned or not mentioned.

On this same appendix page we say that the Geospatial Metadata is "compliant... with DataCite":

> [Geospatial Metadata](https://docs.google.com/spreadsheet/ccc?key=0AjeLxEN77UZodHFEWGpoa19ia3pldEFyVFR0aFVGa0E#gid=4) ([see .tsv version](https://github.com/IQSS/dataverse/blob/master/scripts/api/data/metadatablocks/geospatial.tsv)): compliant with DDI Lite, DDI 2.5 Codebook, DataCite, ...

Should we say "DataCite 4.0" there, too?

qqmyers:
I added that and made the DDI Lite and DDI Codebook entries for the other blocks links as in the Citation block line.

jggautier:
Ah okay 👍🏽

pdurbin (approved):
I tried all the new links and they work fine. I'll go ahead and merge this. I tested at https://dataverse-guide--10377.org.readthedocs.build/en/10377/user/appendix.html

The same links are repeated over and over. In Markdown you can define a link once and reuse it. Probably you can do the same with rst but I'd suggest we switch this page to Markdown some day. It should make it easier to maintain.

PR #10388, opened 2024-03-19
semaphore for async indexing and sync index in transaction after publish

### LINKED ISSUE #10381: Index after publish
As @scolapasta described, If you go to Harvard Dataverse right now and look at the datasets on the front page, most of them do not show the filters (on the currently released versions). This is because (at least from past investigation), the index time is getting set to an earlier time than the publish time. (which is why running a manual reindex fixes it for a particular dataset).

* When does this issue occur?*
After publishing a new dataset.

* Which page(s) does it occurs on?*
All pages.

* What happens?*
The dataset has no index that is more recent than the publishing time.

* To whom does it occur (all users, curators, superusers)?*
All users.

* What did you expect to happen?*
Dataset is indexed.

5.14, 6.0, 6.1

#9199

ErykKul:
Hopefully this fixes indexing issues after publishing of datasets and lowers the risk of solr getting overwhelmed.

Closes #10381 

I do not have much experience with Java semaphores yet, I used them extensively in other languages. We need to be sure it works as intended before merging. I tried to balance this solution and not to move the problem to having too many threads sleeping and waiting on semaphore i.s.o. relying on solr to cope with the load. The new parameter for max running async indexes should be then tuned for Harvard Dataverse, I think (guesstimate) 4 is OK for most instances, but Harvard might want a bigger number.

Ideally, under realistic load in an environment similar to production: publish and edit many datasets concurrently. Also, edit some collections/dataverses, as that appears to be the worst case scenario: after editing a large collection, all datasets get reindexed asynchronously causing large load on the server and solr. This might be the root cause of the problems on Harvard Dataverse.

No.

Yes, it includes new property.

**Additional documentation**:
Needs testing. I hope it will pass unit and integration tests without problems. I will also try testing it on my local installation. Any help in testing it in realistic setting is welcome.


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -2352,6 +2352,15 @@ when using it to configure your core name!
 
 Can also be set via *MicroProfile Config API* sources, e.g. the environment variable ``DATAVERSE_SOLR_PATH``.
 
+dataverse.concurrency.max-async-indexes
```
pdurbin, line 2355: Can we please have a release note snippet where this new config option is mentioned?

pdurbin (commented):
Just a quick doc review.

doc/sphinx-guides/source/installation/config.rst
poikilotherm, line 2355: The setting name is not consistent with the code.

poikilotherm (changes requested):

doc/sphinx-guides/source/installation/config.rst
ErykKul: Note added.

ErykKul: I hope I got it right this time: ```dataverse.solr.concurrency.max-async-indexes```

qqmyers (approved):
Looks good. I guess the proof is in the QA to see how well this handles the issue.

doc/sphinx-guides/source/installation/config.rst
pdurbin: @ErykKul thank you! ❤️

poikilotherm:
**thought**: Do we need a counter and expose it as a metric to see if there are threads pilling up, waiting for their turn? It's  hard to get a feeling for the right number here. Ideally you want to see in your monitoring system a correlation between the setting, the number of threads waiting and the loadavg on Solr...

landreev:
Sorry, what is the status of this PR? - I see that it's been approved; but then it is still sitting in the "Ready for Review" column. And it's showing the red "Merging is blocked" and "Changes Requested" banners - but the latter appears to be pointing to the change @poikilotherm requested the other day, that appears to have been addressed ... Oliver, is it just waiting for you to approve the change? (or do you actually want more to be done, in response to the question about threads?)
This is somehow more confusing than it should be. 🙂 ... please drag into "ready for QA" one way or another.

pdurbin:
Huh. Jim did approve the PR but maybe it didn't automatically move to "ready for QA" because Oliver requested changes?

It looks like we can dismiss Oliver's request for changes, if necessary:

Screenshot 2024-03-20 at 9 08 58 PM

Maybe that would unlock the merge button? Not sure.

poikilotherm (approved):

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -247,6 +247,12 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         
         logger.info("Successfully published the dataset "+readyDataset.getGlobalId().asString());
         readyDataset = ctxt.em().merge(readyDataset);
+
+        try {
+            ctxt.index().indexDataset(readyDataset, true);
+        } catch (SolrServerException | IOException e) {
+            throw new CommandException("Indexing failed: " + e.getMessage(), this);
+        }
```
landreev, line 255: I am trying to understand what's going on here. I understand why we may want to change indexing from asynchronous to synchronous. But why are we moving the indexing from `onSuccess()` and back into the `execute()` method? 
Is this just to address the original issue with timestamps (#10381)? 
Are we positive we want to do this? - If I understand correctly, this would bring us back to a failure to index resulting in a failure to publish. Which I thought we specifically wanted to avoid. 
I may easily be missing something of course.

ErykKul: @landreev @scolapasta  I do not fully understand the issue myself. Moving the indexing back to execute makes the indexing to happen earlier, not later. Also making it synchronous instead of async makes it also happen earlier, not later. If the issue is that the index did happen, but its timestamp is before the publish moment, this PR would make the problem worse, not better. I just do not see yet why the publish date would be greater than the index moment, unless the index was interrupted (e.g., by a circuit breaker), or a wrong version of dataset was indexed (e.g., the draft and not the published version). I will go again through the code to see if I am missing something, and maybe wrong version gets indexed, etc. As its turns out, we just migrated to the version 6.1 (from 5.14) and recently published dataset has the same problem. I think that the core cause is still not addressed. I will keep digging.

poikilotherm:
May I suggest taking a look at https://github.com/ErykKul/dataverse/pull/10 which adds timers? It's a very simple change but will provide a better way to understand if there actually are any load issues. These might be used in production as well as benchmarking, as it is a common "standard" for metrics exposure.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
ErykKul: @landreev @scolapasta I did not see any errors on solr, no failure log in payara, etc. Also, the index looks fine, with the right version etc. I assume that indexing went fine after publish, only the "indextime" was not updated on the DB. I reverted to async update in finalize after publish, and added ```em.flush()``` after the ```indextime``` update, which should fix the problem.

ErykKul:
@poikilotherm Thanks! I have just merged your code, it looks great! I used prometheus with Grafana in the past, It would be cool to have something like that. It would help finding other issues as well.

poikilotherm:
> @poikilotherm Thanks! I have just merged your code, it looks great! I used prometheus with Grafana in the past, It would be cool to have something like that. It would help finding other issues as well.

Splendid! Thanks for merging! Two more questions:

1. Should we add some more details to the docs (guides/release notes) about the availability of these metrics?
2. Are there other metrics we should add now, as a part of this PR? Like histograms of sizes? XY data of size vs time, etc?

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -402,7 +445,16 @@ public void asyncIndexDataset(Dataset dataset, boolean doNormalSolrDocCleanUp) {
     @Asynchronous
     public void asyncIndexDatasetList(List<Dataset> datasets, boolean doNormalSolrDocCleanUp) {
         for(Dataset dataset : datasets) {
-            asyncIndexDataset(dataset, true);
+            try {
+                ASYNC_INDEX_SEMAPHORE.acquire();
```
poikilotherm, line 449:
```suggestion
               acquirePermitFromSemaphore();
```

I forgot to fix this one!

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
ErykKul: Now I think that index was done, the correct ```indextime``` was written, but then it got overwritten by the ```merge``` in
```
try {
            ExportService instance = ExportService.getInstance();
            instance.exportAllFormats(dataset);
            dataset = ctxt.datasets().merge(dataset); 
        }
```
I moved the indexing to the bottom, I am more confident now that it is fixed.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
ErykKul: I just added it and committed.

ErykKul:
@poikilotherm histogram would be more useful than raw data. Also, documentation would be great! Do we have any on how access metrics? If it is a new thing, it might be out of scope of this PR. I think it deserves a new section in the documentation specifically on metrics. We should also add more metrics at other places.

poikilotherm:
There is `doc/sphinx-guides/source/admin/monitoring`. We could add a few words there that can be reformatted / moved / extended later on.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
landreev: Trying to process everything in the comments above... 
But, I am already happy to see the indexing back in `onSuccess()`. I agree that it was unlikely to make things better. (Generally, what happens in our prod. now when solr is crapping the bed - datasets get published without getting properly indexed; these cases can be identified by looking at the index time stamps and datasets can be reindexed after the fact. If we were to move indexing into `execute()` and make a failure fatal, we would be ending up with datasets that are neither published, nor indexed - likely causing more heartbreak to users.

landreev:
I started testing the effects of the semaphores on our test instance running off a copy of the prod. database clone; with somewhat inconclusive results just yet. But I'm only doing it in parallel with having been gang-pressed into working on something else prod. and admin here.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
poikilotherm: While hacking on the timers, I was thinking that we might want a list of datasets for which indexing failed and try them again later at random. (Like CSMA has a retry mechanism with exponentially growing random retry times)

landreev: Yeah, we've been doing this (asynchronously reindexing missed datasets after the fact) outside of the application; but an internal mechanism as you described would help. I would maybe add a provision for identifying the cases where a dataset fails to index for a reason other than the load - something in the metadata that solr doesn't like;  does happen too - so that we don't waste any time trying to reindex these repeatedly, at least until the next time the version is updated.

ErykKul: I think the best check would be controlling if the `releasetime` is lower than `indextime`, without calling solr for extra load. However, the indexing in `onSuccess()` happens asynchronously, so it is not possible to know which finishes first: `onSuccess()` or the `index` operation. I do not see how we could implement the check that is consistent in either of them, unless we lock them to wait until both finish, which might not be ideal. I see it more like a batch that runs a query on db and detects which datasets need reindexing (`indextime < releasetime`, and maybe `releasetime` < `getDate() - 1Day`?) and reindexes them. This could happen once a day at low traffic moments in the night time, if you have any on your server...

Do we have something to run cron jobs inside payara? I recently used the [Quartz Job Scheduling Library](https://www.quartz-scheduler.org/) on a Tomcat server and I am happy with it. If we do not have anything yet, it might be a good idea. Using that we can introduce batch jobs for cleaning up stuff using existing code and from within the war, what would make writing jobs way easier. It could look something like the stuff I wrote for the project I am working on:

image

If it is a direction we want to go to, I can pick that up.

poikilotherm: Two standard options without additional dependencies:

1. EJB Timers: https://www.baeldung.com/scheduling-in-java-enterprise-edition
2. Concurrency: https://jakarta.ee/specifications/concurrency/3.0/apidocs/jakarta/enterprise/concurrent/managedscheduledexecutorservice

Personally, I would like to avoid EJB timers if we ever want to get rid of EJB and move to a more lightweight stack.
That said, [there are folks you think otherwise](https://dzone.com/articles/managedscheduledexecutorservice-vs-ejb-timer) and we do use EJB timers in Dataverse as well. Look for `@Scheduled` in the codebase.

We also have some limited timer things in the codebase already: `edu.harvard.iq.dataverse.timer.DataverseTimerServiceBean`. I'm not very fond of it, but that's where we are.

In the future, I would love to use more `ScheduledExecutor` things, as [we can use Hazelcast](https://docs.hazelcast.com/hazelcast/5.3/computing/scheduled-executor-service) to coordinate the execution to spread the load or just make sure it is executed only once. Currently this works by defining one node as primary.

ErykKul: I agree on avoiding adding new dependencies and streamlining the approach. Having a GUI to have an overview on what is running and how often would be a nice addition, especially if you could edit it, but this is more in a long term then. In the shorter term, I think that adding a job for fixing the missing indexes with one of the techniques you described (`ScheduledExecutor` looks more familiar to me than the timers), might be a good idea. For the current sprint, I would leave it as it is now, and plan a new task for this.

ErykKul: @Kris-LIBIS Can you take a look at the schema.xml and config.xml differences between the 5.14 and 6.1, if you have some spare time? Maybe you can see something that makes solr work harder on the new version? This might have nothing to do with the config, but might be worth checking: [diff](https://github.com/IQSS/dataverse/compare/v5.14...v6.1#diff-09a593592430aa226691f22630744fa71ae278a5baf8ffe559259bec8b7900f2)

landreev: @ErykKul I'm trying to understand a couple of things from the discussion above:

> I think the best check would be controlling if the releasetime is lower than indextime, without calling solr for extra load. 

Yes, that would be the same check that the DatasetPage.java is performing: 

```
(workingVersion.getDataset().getIndexTime() != null) && 
workingVersion.getDataset().getIndexTime().after(workingVersion.getReleaseTime());
```

We used to actually call solr to check if the version was indexed in that page; then switched to performing the quick check above, for performance reasons. But, we now have a problem of a false negative, when the version is indexed, but the indexing time stamp was not properly set. This is the issue that started this (#10381). ... Apologies if I'm restating the obvious, just trying to make sure we are on the same page. 

> However, the indexing in onSuccess() happens asynchronously, so it is not possible to know which finishes first: onSuccess() or the index operation. ... 

But, why is it relevant at all, when onSuccess() completes? - Shouldn't the update of the DatasetVersion (and the ReleaseTime value) have already been finalized after execute() completed? - wasn't that the whole point of onSuccess(), to run it after the fact? (or are they both in the same transaction somehow?? - that's what I meant yesterday when I said I wanted to revisit how these methods were executed in the command engine). 

But, what about your last comment yesterday: 

> Now I think that index was done, the correct ```indextime``` was written, but then it got overwritten by the ```merge``` in [the export code that was executed later in the same ```onSuccess()``` method 

that actually made sense to me, and I assumed that that was indeed the problem, and that you fixed it, by simply rearranging the index and export methods there, and the testing I did yesterday appeared to confirm it...Do you no longer think that it was the case?

landreev: ... and if it is indeed the case, that the updates made to the version and the dataset in the body of the `execute()` method are not finalized in the database until the completion of the `onSuccess()` method - and therefore the new indexing time stamp can be overwritten during that finalization, if the async. indexing has happened too fast, before `onSuccess()` finished... well, seeing how indexing is now the very last thing in the onSuccess() method, something as dumb as inserting a couple of seconds of sleep in that indexing method will ensure that it never finishes before `onSuccess()` itself - ?

landreev: > I do not see how we could implement the check that is consistent in either of them, unless we lock them to wait until both finish, which might not be ideal. I see it more like a batch that runs a query on db and detects which datasets need reindexing (indextime < releasetime, and maybe releasetime < getDate() - 1Day?) and reindexes them. This could happen once a day at low traffic moments in the night time, if you have any on your server...


I've been trying to do precisely that in our prod. - detect such unindexed datasets and reindex them late at night (while sleeping generously between the calls - otherwise we may be back to solr getting overloaded ☹️ ). A crontab does this in a pinch, but we could definitely try and do this internally. As @poikilotherm already pointed out, we already have a timer service, that's already doing the equivalent of the above for exports - tries to detect any datasets that for one reason or another didn't get to exported and reexport them - so we may as well add a job for reindexing there. I understand Oliver's dislike of that service; but, as he points out, that would be one quick way of achieving this. 
I'm not sure this should be done within this PR though; feels like a bit of a scope creep. I also agree with Oliver in that if we were to implement that internal detection of unindexed datasets, it would be nice to implement it in a smarter way - for example, the idea of increasing sleep intervals when these reindexing calls get rejected because of load.

landreev: Oh, and as for this idea specifically: 

> ... and maybe ```releasetime < getDate() - 1Day``` - that was one of the first ideas that came to my mind, as a quick way to resolve the original problem in #10381, just add a few hours of padding to that check. If that results in a false positive, the worst thing that's going to happen we will waste making a solr call from DatasetPage, which will come back empty if the version is not indexed. (We will NOT run into a problem of getting stale results back from solr, from a version that's no longer the latest published - because the numeric id of the version is prat of the solr query).

ErykKul: It looks to me that it is safe to merge the current code of this PR and leave it like this until we see how it will behave in production. I saved the link to this issue as it contains some interesting discussion, especially on running batches and metrics. We can revisit these ideas at some point. As for adding the sleep calls, I am against that since this would makes the queue holding indexing jobs longer than necessary. I like the idea of adding the index logic to the timer for fixing exports, but I agree it is out of scope for this PR. We will pick it up later, if needed.

landreev: Could it be the move from Solr 8 to 9 itself that changed something dramatically? 
From a brief look, I'm not seeing any information online suggesting 9 is dramatically slower than 8 in general. ... But could be something Dataverse specifically relies on when indexing that is behaving differently.

ErykKul: I moved the indexed time in the `isIndexedVersion` test by 3 hours as suggested. This should fix the problem for already published datasets with `indexedtime` before `releasetime`.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -790,11 +790,15 @@ public boolean isIndexedVersion() {
...
         }
         // If this is the latest published version, we want to confirm that this 
-        // version was successfully indexed after the last publication 
-        
+        // version was successfully indexed after the last publication
+        // We add 3 hours to the indexed time to prevent false negatives
+        // when indexed time gets overwritten in finalizing the publication step
+        // by a value before the release time
+        final long duration = 3 * 60 * 60 * 1000;
+        final Timestamp movedIndexTime = new Timestamp(workingVersion.getDataset().getIndexTime().getTime() + duration);
```
landreev, line 798: Sorry to be a pest, but the null check below may need to happen before this line - ?

ErykKul: That was an ugly bug. Thanks for catching it!

ErykKul:
I added a mention in the `doc/sphinx-guides/source/admin/monitoring.rst` about the newly added metrics. I am not sure if it is sufficient and I think it will need some cleaning up later on.

landreev:
The guide entry for the metrics is perfect. 
I'm going to merge the PR now. Will share some observations from running tests on a large database when I have a moment. But definitely looking forward to seeing the effects, and the metrics in real production use. 
Thank you again, @ErykKul and @poikilotherm for this effort!

PR #10391, opened 2024-03-19
IQSS 10390: Multipid UI Fix

### LINKED ISSUE #10390: MultiPid UI selector incorrect for sub-collections
The existing code doesn't handle the case where a non-default inherited PID Provider exists - the entry shows the default which is not what will be used. This is only an issue in sub-collections (where an inherited PID provider can exist).

* When does this issue occur?

* What happens?

As with other UI selections, the UI should show what will be used (the default or the inherited value if it exists), e.g. 
image

develop

**Screenshots:**

*

qqmyers:
**What this PR does / why we need it**: Fixes the issue by adding logic to determine what option will be inherited by a collection - the global default or the value inherited from a parent collection. The PR also addresses the issue with the GET /api/datasets/id/pidGenerator call sometime returning "default" instead of the PidProvider that will be used (as documented in the Guides).

Closes #10390

**Special notes for your reviewer**: Suggesting we put this in 6.2 to avoid a bug going out.

**Suggestions on how to test this**: Have more than one pid provider configured, create a subcollection and verify that the menu of options on the Dataverse/Edit/GeneralInfo pane shows the options, including one that is either the (default) - if no parent collection has a non-default selection - or (inherited...) option - when a parent/grandparent has an explicit PIDProvider setting. 
Can also call the API call above and verify that it returns a PidProvider id in all cases.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: see screenshot in inssue.

**Is there a release notes update needed for this change?**: no - fixes a bug that hasn't yet been released.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -39,6 +39,7 @@ import edu.harvard.iq.dataverse.ingest.IngestServiceBean;
 import edu.harvard.iq.dataverse.makedatacount.*;
 import edu.harvard.iq.dataverse.makedatacount.MakeDataCountLoggingServiceBean.MakeDataCountEntry;
 import edu.harvard.iq.dataverse.metrics.MetricsUtil;
+import edu.harvard.iq.dataverse.pidproviders.PidProvider;
 import edu.harvard.iq.dataverse.pidproviders.PidUtil;
 import edu.harvard.iq.dataverse.privateurl.PrivateUrl;
 import edu.harvard.iq.dataverse.privateurl.PrivateUrlServiceBean;
@@ -4575,6 +4576,12 @@ public class Datasets extends AbstractApiBean {
         }, getRequestUser(crc));
     }
     
+    /**
+     * Get the PidProvider that will be used for generating new DOIs in this dataset
+     *
+     * @return - the id of the effective PID generator for the given dataset
+     * @throws WrappedResponse
+     */
     @GET
     @AuthRequired
     @Path("{identifier}/pidGenerator")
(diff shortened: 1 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4588,7 +4595,12 @@ public Response getPidGenerator(@Context ContainerRequestContext crc, @PathParam
         } catch (WrappedResponse ex) {
             return error(Response.Status.NOT_FOUND, "No such dataset");
         }
-        String pidGeneratorId = dataset.getPidGeneratorId();
+        PidProvider pidProvider = dataset.getEffectivePidGenerator();
+        if(pidProvider == null) {
+            //This is basically a config error, e.g. if a valid pid provider was removed after this dataset used it
+            return error(Response.Status.NOT_FOUND, "No PID Generator found for the give id");
```
sekmiller, line 4601: nitpick here - should be "given" and maybe put it in the bundle. Also should we add a test to PidUtilTest?

sekmiller (changes requested):
Just an error label update, and maybe an additional test - see comment.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: done

sekmiller (approved):
Thanks for the updates

pdurbin:
Works fine but I didn't like how (inherited from enclosing Dataverse) and (Default) are squished...

Screenshot 2024-03-27 at 11 34 37 AM
Screenshot 2024-03-27 at 11 36 16 AM

... so I pushed a fix to add spaces in b196f54

Here's the config I tested with:

        -Ddataverse.pid.providers=fake,foo
        -Ddataverse.pid.default-provider=fake
        -Ddataverse.pid.fake.type=FAKE
        -Ddataverse.pid.fake.label=FakeDOIProvider
        -Ddataverse.pid.fake.authority=10.5072
        -Ddataverse.pid.fake.shoulder=FK2/
        -Ddataverse.pid.foo.type=FAKE
        -Ddataverse.pid.foo.label=Foo
        -Ddataverse.pid.foo.authority=10.1234

Tests are passing. Will merge.

PR #10401, opened 2024-03-20
Fix files not uploading when there is a required selectone component

### LINKED ISSUE #10385: File Upload: Changing the HOST collection on HDV is interfering with file upload

Navigate to dataverse.harvard.edu and select the **"add data" "new dataset" option at the TOP of the page**
change the host from "Harvard Dataverse" to collection  [Climate Change and Health Research Coordinating Center (CAFE) Collection](https://dataverse.harvard.edu/dataverse/CAFE)

fill in required fields, upload files, and select save.

File doesn't show up on save.

* When does this issue occur?

When using "add data" and "new dataset" at the TOP of the production page, within a collection
*inside CAFE collection, but using "add data" "new dataset" on top page toolbar

add dataset deposit page when inside a collection but using top level "add data-new dataset" option

* What happens?
Files disappear when you hit save and don't show up in the saved draft

All

Uploaded files are ingested/uploaded on save

**Screenshots:**

*

scolapasta:
Fixes an issue with dataset field types that use controlled vocab, are only single select, and are required

Closes #10385

So it investigating this, I found that the selectOne simple fields had a required attribute, whereas no other fields did. Removing it fixed the issue AND validation still works correctly. 

Looked in history, found it was added 5 years ago, but not to other fields - not sure why. Likely it was not actually needed (as validation is handled differently).

Here is what the attribute looked like before removing:
```required="#{dsf.required and datasetPage}"```

Use the CAFE metadatablock that has dataset field types that use controlled vocab, are only single select, and are required (or create a new one for testing). Add a new dataset and try to upload a file before filling out the metadata. Make sure it gets added to the list, and then also fill out metadata, and make sure everything gets saved correctly.

Also confirm that the field still acts as required (won't save without value, gets validation message)

No

No, simple bug fix

**Additional documentation**:

### DIFF
--- a/src/main/webapp/metadataFragment.xhtml
+++ b/src/main/webapp/metadataFragment.xhtml
@@ -285,7 +285,7 @@
                                                     <div class="form-group dataset-field-values">
                                                         <div class="form-col-container col-sm-9 edit-field">
                                                             <p:selectOneMenu value="#{dsf.singleControlledVocabularyValue}" converter="controlledVocabularyValueConverter" style="width: auto !important; max-width:100%; min-width:200px;" styleClass="form-control primitive"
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
Seems like a reasonable fix. I haven't tested it.

PR #10331, opened 2024-02-20
CVOC : add HTTP request headers (Ontoportal integration)

jeromeroucou:
**What this PR does / why we need it**:

Add HTTP headers customization for [:CVocConf](https://guides.dataverse.org/en/latest/installation/config.html#cvocconf) settings. This is needed for Ontoportal integration as a external vocabulary services, which required API KEY provided by HTTP request headers.

**Which issue(s) this PR closes**:

Closes #10316 

**Suggestions on how to test this**:

Add `:CVocConf` to a external vocabulary service, with `header` customization : 
```
"headers": {
    "Accept": "application/xml",
    "Cache-Control": "no-cache"
}
```
Replace `XXX` below by the external vocabulary service protocol configured.

Verify if `span[data-cvoc-protocol="XXX"][data-cvoc-headers]` in advanced search page and dataset page (with metadata tab open) have the configuration previously configured. Verify if `input[data-cvoc-protocol="XXX"][data-cvoc-headers]` in edition dataset page have also the configuration.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No

Docs:

 - https://dataverse-guide--10331.org.readthedocs.build/en/10331/installation/config.html#cvocconf
 - https://github.com/gdcc/dataverse-external-vocab-support/pull/19/files

PR #10404, opened 2024-03-21
CVOC : allow flexible params in retrievalUri (Ontoportal integration)

luddaniel:
@qqmyers Following agreements on next steps of #10145 here is a way to add flexibility in retrievalUri.
We can now add named fields (child or siblings) of `term-uri-field` and capacity to encodeUri value when usefull.
Example below : 

cvoc_conf.json contains : 

```json
"field-name": "keyword",
"term-uri-field": "keywordTermURL",
"retrieval-uri": "https://data.agroportal.lirmm.fr/ontologies/{keywordVocabulary}/classes/{encodeUrl:keywordTermURL}",
"managed-fields": {
        "vocabularyName": "keywordVocabulary",
        "termName": "keywordValue",
        "vocabularyUri": "keywordVocabularyURI"
    },
```

Resulting in retrieval-uri = `https://data.agroportal.lirmm.fr/ontologies/AGROVOC/classes/http%3A%2F%2Faims.fao.org%2Faos%2Fagrovoc%2Fc_331529`

This code was tested with a classic skomos `https://skosmos.loterre.fr/rest/v1/data?uri={0}` and it's still working.

Refactoring has been made regarding the use of `registerExternalTerm` function around dataset imports / updates methods where it was used while parsing json fields meaning imcomplete related fields. Now `registerExternalTerm` is called after dataset metadata are ready to be commited and directly as step of `Command` of dataset create and update.


qqmyers (changes requested):
Looks good. Nice to have the register part out of the DatasetPage and in the commands (where it can be run with the SPA). 

It would be good to add a release note (could be a combined one for this and #10331). It would also be helpful to have some testing instructions and/or a test: nominally existing scripts should still work, so those could be tested with the examples in the https://github.com/gdcc/dataverse-external-vocab-support/ repo. If there isn't an ontoportal example yet, perhaps just verifying that the substitution works in replaceRetrievalUriParam in a unit test would be helpful.

I'll also note that this is matched by a docs update at https://github.com/gdcc/dataverse-external-vocab-support/pull/20.

luddaniel:
Hello @qqmyers Thanks for the review,
- Release note has been added for both PR here (#10404 + #10331) .
- PR synchronized with develop
- https://github.com/gdcc/dataverse-external-vocab-support/pull/20 updated

For testing / unit tests / integration tests, we decided to go for unit tests of methods of DatasetFieldServiceBean, coming in another PR. Other kind of testing looks complicated for now. I'll talk to you on slack for this topic.

luddaniel:
Decided with @qqmyers, the unit tests for all PR for [Ontoportal / CVOC](https://github.com/IQSS/dataverse/pulls?q=is%3Apr+%28Ontoportal+integration%29+) will be pushed after the merging of PR as it will require merged codes.

qqmyers:
@luddaniel - There's 1 test failing with this PR - a 500 error returned instead of a 200 response at edu.harvard.iq.dataverse.api.EditDDIIT.testUpdateVariableMetadata(EditDDIIT.java:153) which is caused ultimately by an attempt to save a DatasetVersion that doesn't have a createTime set. From the lengthy stack trace, I'm guessing that is due to the call to ctxt.dsField().registerExternalVocabValues(df); which at some point calls em.flush() which must be causing the DatasetVersion itself to be flushed to the db. This happens before the date gets set in UpdateDatasetVersionCommand line ~261 (setting the last update time also sets the createTime if it was null). 

If that's all correct, I think adding a CVOC value to a published version (causing a new version to be created) in the UI would trigger the issue. Assuming you can repeat the problem, it might be that we no longer need the flush(), or that setting the date can be moved earlier in the method, etc.

doc/release-notes/9276-allow-flexible-params-in-retrievaluri-cvoc.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Updates on Support for External Vocabulary Services
+
+#### HTTP Headers
+
+You are now able to add HTTP request headers required by the service you are implementing (#10331)
+
+#### Flexible params in retrievalUri
+
+You can now use `managed-fields` field names `term-uri-field` field name as parameters of `retrieval-uri`.
```
qqmyers, line 11: Suggested change:
You can now use `managed-fields` field names as well as the `term-uri-field` field name as parameters in the 'retrieval-uri' when configuring an external vocabulary service. '{0}' as an alternative to using the 'term-uri-field' name is still supported for backward compatibility.

luddaniel:
@qqmyers your tips with https://github.com/IQSS/dataverse/pull/7334/commits/da87bd1cc5d2aacb8bd0c1c6a5c5750b8a2f9d10 were correct, I just needed to set the created time before `registerExternalVocabValuesIfAny()` function for whatever reason...
Tested without `@TransactionAttribute(TransactionAttributeType.REQUIRES_NEW)` it seems to work just fine.

Suggested change on release note was added

qqmyers (approved):
Looks good, test is passing.

PR #9764, opened 2023-08-07
Upgrade to Java 17

pdurbin:
**What this PR does / why we need it**:

Upgrade from Java 11 to 17

**Which issue(s) this PR closes**:

- Closes #8094

**Special notes for your reviewer**:

The biggest change (in terms of number of files) has to do with REST Assured. I had to upgrade it to get it working on Java 17. The upgrade required a namespace change so I had to fix all the classes that import REST Assured.

The following tests were failing...

```
Error:  Errors: 
Error:    DataAccessTest.testCreateNewStorageIO_createsFileAccessIObyDefault:59 » NullPointer
Error:    DatasetUtilTest.testGetThumbnailCandidates:33 » NullPointer
Error:    DatasetUtilTest.testGetThumbnailNullDataset:51 » NullPointer
```

... and I fixed the DatasetUtilTest failures by cherry picking 23cc3b5 and DataAccessTest by my setting the dataset authority in a mock (c1b57a168c).

I've set up an API test running on the side at https://github.com/pdurbin/dataverse-api-test-runner/actions/workflows/java17.yml but I'd be happy to put this into the main code base.

Jenkins tests are failing until we switch Jenkins to Payara 6.

**Suggestions on how to test this**:

Test with Java 17

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

**Is there a release notes update needed for this change?**:

Included.

**Additional documentation**:

Included.

PR #10405, opened 2024-03-21
Documentation update to java requirements

jp-tosca:
Documentation update to java requirements

Java version is 17 not 11


pdurbin (approved):
As I was telling @jp-tosca I updated 11 to 17 in this PR:

- #9764

But then we renamed the page and the change got lost. Whoops. Will merge. Good catch!

PR #10407, opened 2024-03-21
The commits that didn't make it. #9356

### LINKED ISSUE #9356: Add execution rate metering to the command engine
The proposed idea is to modify the command engine as to make it possible to keep accurate counts of executed commands per specific time intervals. We would want to count the totals for all commands, as well as specific commands (and, possibly, classes of commands?) as well as the counts and rates for individual users. 
 
Note that this can be achieved without adding any new data structures (the values can be obtained by counting the entries in the ActionLogRecord table). However, that table is quite unwieldy on any busy installation, so we almost certainly want to add some efficient way of caching and updating the counts in the database. 

Then we want to add a system of configurable rate limits for specific commands, in a way that would allow setting different limits for different users/groups of users, etc. 
In addition to limits by the number of executed commands - for example, "certain number of  UpdateDatasetVersionCommand per minute allowed for an otherwise unprivileged logged-in user" - Action Log records how much time it took to execute each command as well. So we may want to consider limiting use by that measure too (?). For example, if a specific user keeps making edits, but for whatever reason the UpdateDatasetVersionCommand on their dataset is especially cpu- and time-consuming - maybe that could trigger some red flags too. 

A simplest use of this functionality would be to set a rate limit on, say, Get*Version commands for the anonymous (guest) users, and that would address a somewhat common case of a scripted crawler plowing through our holdings; past a certain call in too short a period of time the API starts giving them "too busy, try again later". 

More care will need to be taken to make our pages communicate this "try again later" message to the UI users. We don't want anyone to lose a page-worth of edited metadata they are trying to save, etc.

landreev:

These are the late commits that were left out of the merge yesterday, unintentionally. 
Some of it purely cosmetic - I added `Command` to the names of the 2 new commands, for consistency with all the existing commands. But the other changes are important, without them rate-limiting the 2 workhorse pages results in 500/Internal Server Error shown to the user, which is not ideal. 

To reiterate, these changes were properly tested yesterday. Specifically re-tested even seconds before I clicked the Merge button. ☹️ 

Closes #9356 (well, not unlike that dead horse - you can't close it any more than it's closed already, but ...)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -24,7 +24,7 @@ import edu.harvard.iq.dataverse.datavariable.VariableServiceBean;
 import edu.harvard.iq.dataverse.engine.command.Command;
 import edu.harvard.iq.dataverse.engine.command.CommandContext;
 import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
-import edu.harvard.iq.dataverse.engine.command.impl.CheckRateLimitForDatasetPage;
+import edu.harvard.iq.dataverse.engine.command.impl.CheckRateLimitForDatasetPageCommand;
 import edu.harvard.iq.dataverse.engine.command.impl.CreatePrivateUrlCommand;
 import edu.harvard.iq.dataverse.engine.command.impl.CuratePublishedDatasetVersionCommand;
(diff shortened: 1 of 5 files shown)


qqmyers (approved):
LGTM

stevenwinship (approved):

landreev:
The new harvesting test - interesting. (I'm seeing Jenkins is running again)

PR #10408, opened 2024-03-21
Fix PUBLICATION_STATUS facet order and visibility

### LINKED ISSUE #10338: Move puvlication status facet from bottom to where it used to be
In one of our recent releases, the publication status facet location was inadvertently moved to the bottom of the facet list. Let's move it back to where it used to be.

jp-tosca:

It was reported that the order of the facets was unintentionally changed, this PR adds the publication status facet back to the original place, it also was discussed on Slack that users can see objects even if they are not published if the owner grants access to the group. 

Closes #10338

- If you are logged in and you have objects with published, draft and/or unpublished status you should see this facet on the top of the left side.
- If you create a dataset unpublished and give access to everyone you should be able to see this facet on the top while not being logged in.
- If you have a dataset on draft and give access to everyone you should be able to see this facet on the top while not being logged in.

Yes

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
@@ -214,16 +214,17 @@ public class SearchServiceBean {
         }
         
         List<DataverseMetadataBlockFacet> metadataBlockFacets = new LinkedList<>();
-        
-        /*
-         * We talked about this in slack on 2021-09-14, Users can see objects on draft/unpublished 
-         *  if the owner gives permissions to all users so it makes sense to expose this facet 
-         *  to all users. The request of this change started because the order of the facets were 
-         *  changed with the PR #9635 and this was unintended.
-         */
-         solrQuery.addFacetField(SearchFields.PUBLICATION_STATUS);
 
         if (addFacets) {
+
+            /*
+            * We talked about this in slack on 2021-09-14, Users can see objects on draft/unpublished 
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
```diff
@@ -222,6 +214,14 @@ public SolrQueryResponse search(
...
         
         List<DataverseMetadataBlockFacet> metadataBlockFacets = new LinkedList<>();
+        
+        /*
+         * We talked about this in slack on 2021-09-14, Users can see objects on draft/unpublished 
+         *  if the owner gives permissions to all users so it makes sense to expose this facet 
+         *  to all users. The request of this change started because the order of the facets were 
+         *  changed with the PR #9635 and this was unintended.
+         */
+         solrQuery.addFacetField(SearchFields.PUBLICATION_STATUS);
```
qqmyers, line 224: Should still be inside the if(addFacets)?

jp-tosca: Done! Thanks a lot, good catch 🐟

qqmyers (approved):
Looks like it should resolve the issue

pdurbin:
Looks good to me:

Screenshot 2024-03-22 at 2 36 03 PM

Screenshot 2024-03-22 at 2 36 19 PM

Screenshot 2024-03-22 at 2 36 41 PM

PR #10416, opened 2024-03-22
fixing bug with tabs in log line for MDC

stevenwinship:
Log lines for MDC must be tab delimited but in some cases the data in the lines contain tabs which throws off the columns.
Replace the tabs with spaces in MakeDataCountLoggingServiceBean.

In this example the authors column(14) contains 2 tabs adding 2 additional columns causing an error in counter_processor
ex:
2024-03-21 15:17:49 importing log/counter_2024-02-01.log
line is wrong: 2024-02-01T00:13:30-0500 98.121.236.139 - - :guest https://dataverse.harvard.edu/api/v1/datasets/export?exporter=schema.org&persistentId=doi%3A10.7910%2FDVN%2FJ1UD6S doi:10.7910/DVN/J1UD6S - - python-requests/2.28.2 Main Cities grid tbd 00_metadata 00_metadata 100% 10 Digital Map Database of China 已启用屏幕阅读器支持。 Digital Map Database of China 2020-02-15T21:35:14Z 1 - https://dataverse.harvard.edu/api/v1/datasets/export?exporter=schema.org&persistentId=doi%3A10.7910%2FDVN%2FJ1UD6S 2020

Which issue(s) this PR closes:

Needed for https://github.com/IQSS/dataverse.harvard.edu/issues/3

Suggestions on how to test this:
Make sure the log file has no extra tabs

Does this PR introduce a user interface change? If mockups are available, please link/include them here: No

Is there a release notes update needed for this change?: No

Additional documentation: None

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/makedatacount/MakeDataCountLoggingServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/makedatacount/MakeDataCountLoggingServiceBean.java
@@ -46,6 +46,12 @@ public class MakeDataCountLoggingServiceBean {
     public String getLogFileName() {
         return "counter_"+new SimpleDateFormat("yyyy-MM-dd").format(new Timestamp(new Date().getTime()))+".log";
     }
+
+    // Sanitize the values to a safe string for the log file
+    static String sanitize(String in) {
+        // Log lines are tab delimited so tabs must be replaced. Replacing escape sequences with a space.
+        return in != null ? in.replaceAll("\\s+", " ") : null;
+    }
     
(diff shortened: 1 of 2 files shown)


qqmyers (approved):
Makes sense - gets rid of tabs that cause issues and I don't think any of this info is used outside counter-processor so any transform that doesn't result in false matches should be OK (and we wouldn't expect anything differing only by whitespace to really be different as far as c-p cares.).

landreev (approved):
Makes sense. Was going to approve, and ask @qqmyers to take a look anyway, just in case... But he appears to have beat me to it.

PR #10418, opened 2024-03-25
Extend the MetadataBlocks API endpoints for the Create Dataset Feature

### LINKED ISSUE #10389: Extend the API to add the displayOnCreate flag in the DatasetFieldType and MetadataBlock payloads

We need to add this flag in the DatasetFieldType and MetadataBlock API payloads so that we can print the required fields in the SPA dataset creation form.

API User

Frontend dev team discussion

None

Extend the API payloads to add the displayOnCreate flag on DatasetFieldType and MetadataBlock

- https://github.com/IQSS/dataverse-client-javascript/issues/135

GPortas:

New optional query parameters added to _api/metadatablocks_ and _api/dataverses/{id}/metadatablocks_ endpoints:

- `returnDatasetFieldTypes`: Whether or not to return the dataset field types present in each metadata block. If not set, the default value is false.
- `onlyDisplayedOnCreate`: Whether or not to return only the metadata blocks that are displayed on dataset creation. If `returnDatasetFieldTypes` is true, only the dataset field types shown on dataset creation will be returned within each metadata block. If not set, the default value is false.

Added new `displayOnCreate` field to the MetadataBlock and DatasetFieldType payloads.

- Closes #10389 

None

Call the modified endpoints using the new optional query params. You can test the endpoints by sending one of the new parameters, both, or none.

**Metadatablocks:**

`curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/metadatablocks?returnDatasetFieldTypes=true&onlyDisplayedOnCreate=true"
`

**Dataverse Metadatablocks:**

`curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/dataverses/$ID/metadatablocks?returnDatasetFieldTypes=true&onlyDisplayedOnCreate=true"
`

No.

Yes, attached.

**Additional documentation**:

None


src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -710,14 +710,20 @@ public Response deleteDataverseLinkingDataverse(@Context ContainerRequestContext
...
-            for (MetadataBlock mdb : blocks) {
-                arr.add(brief.json(mdb));
-            }
-            return ok(arr);
+            final List<MetadataBlock> metadataBlocks = execCommand(
+                    new ListMetadataBlocksCommand(
+                            createDataverseRequest(getRequestUser(crc)),
+                            findDataverseOrDie(dvIdtf),
+                            onlyDisplayedOnCreate,
+                            metadataBlockSvc
```
qqmyers, line 723: Rather than passing in a service, the MetadataBlockServiceBean should probably get added to the [CommandContext class](src/main/java/edu/harvard/iq/dataverse/engine/command/CommandContext.java) - that's the standard pattern for accessing services in commands.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -580,6 +580,14 @@ public static JsonObjectBuilder json(MetadataBlock block, List<DatasetField> fie
         return blockBld;
     }
 
+    public static JsonArrayBuilder json(List<MetadataBlock> metadataBlocks, boolean returnDatasetFieldTypes, boolean onlyDisplayedOnCreate) {
+        JsonArrayBuilder arrayBuilder = Json.createArrayBuilder();
+        for (MetadataBlock metadataBlock : metadataBlocks) {
+            arrayBuilder.add(returnDatasetFieldTypes ? json(metadataBlock, onlyDisplayedOnCreate) : brief.json(metadataBlock));
```
qqmyers, line 586: If returnDatasetFieldTypes is false and onlyDisplayedOnCreate is true, this will still return all metadatablocks, not just those displayed on create? Is that a bug or feature?

qqmyers: I see that in use only displayed on create blocks are handed to this method when onlyDisplayedOnCreate is true - perhaps a comment here that that's a requirement?

qqmyers (changes requested):
Looks good. Asking for a comment and to switch to passing the MetadataBlockServiceBean to the command via the CommandContext. Once those are done, it's good to go.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
GPortas: Actually the `onlyDisplayedOnCreate` param in this method was referring to the dataset field types that we want to print, not the metadata blocks, that's why I just renamed the parameter to `printOnlyDisplayedOnCreateDatasetFieldTypes`. Tell me if you think this makes the behavior clearer, if not I can add a comment as you suggest.

In practice, the method could be invoked for any metadata block, which is why I personally prefer not to add the comment which would relate its use to a specific scenario.

qqmyers: That's works

PR #10420, opened 2024-03-25
Fix missing environment variable for IntelliJ on docker at the dev-usage guide

jp-tosca:

There is a missing environment variable on this guide and will cause an incompatibility when running docker from the CLI and when using IJ.

After the discussion on this thread, we have opted for not updating the guide but changing the .env file so PG16 is used by default when using IJ.


doc/sphinx-guides/source/container/dev-usage.rst
```diff
@@ -263,7 +263,7 @@ Hotswapping methods requires using JDWP (Debug Mode), but does not allow switchi
 
         .. image:: img/intellij-compose-add-new-config.png
 
-        Give your configuration a meaningful name, select the compose file to use (in this case the default one), add the environment variable ``SKIP_DEPLOY=1``, and optionally select the services to start.
+        Give your configuration a meaningful name, select the compose file to use (in this case the default one), add the environment variables ``SKIP_DEPLOY=1`` and ``POSTGRES_VERSION=16``, optionally select the services to start.
```
poikilotherm, line 266: The root cause for this is the `.env` file wasn't updated. It still has POSTGRES_VERSION=13 in it. @donsizemore changed the POM, but the .env file change wasn't included.

poikilotherm: Wondering if we should stop including this file and make it generated by Maven (even without the `-Pct`)

poikilotherm: Maybe we should start a Chores Github Action that simply collects all of these tedious task that are easy to forget and apply checks like "is the version in these files matching?".

jp-tosca: Should we bump it? I am just afraid it will hit somehow any non-docker development environment, what do you guys think? @poikilotherm @pdurbin

jp-tosca: I just saw this on one of the patch-notes "Postgres 13 remains the version used with automated testing." 🤔

poikilotherm: The .env file is solely used for containers. Safe to upgrade.

jp-tosca: What do you guys think about this? @poikilotherm @pdurbin Since the guide is shipping with 6.2 this will help that the instructions provided work. In theory, we could just do the previously suggested change for the guide.

pdurbin: Developers should probably get used to looking at https://preview.guides.gdcc.io for the latest and greatest. 😄

poikilotherm (approved):
Looks good to me. Will merge as trivial.

PR #9262, opened 2023-01-05
standardize license configuration

philippconzett:
**What this PR does / why we need it**:
This PR is part of the work to make license information from Dataverse installations harvestable in a way that complies with DataCite recommendations as described in issue #8512 Standardize standard license configuration.

**Which issue(s) this PR closes**:
Together with other PRs, this PR closes #8512.

**Special notes for your reviewer**:
See my comment from January 5, 2023 to issue #8512.

**Suggestions on how to test this**:
In a test environmen:
1. Install the JSON license files.
2. Create a dataset.
3. Select a standard license.
4. Export DataCite metadata.
5. Verify that the license information is as recommended by DataCite.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
No.

**Is there a release notes update needed for this change?**:
Yes. See my comment from January 5, 2023 to issue #8512.

**Additional documentation**:
The longer we wait implementing this change, the more installations will have to do clean-ups. Therefore, I hope this PR can be prioritized.

PR #10426, opened 2024-03-25
Add MIT license file and guidance for more license files

### LINKED ISSUE #10425: Add MIT License request so can be added to HDV

The request to include on HDV the MIT License was created on https://github.com/IQSS/dataverse.harvard.edu/issues/248, This PR adds a JSON file so the license can be added.

jp-tosca:

The request to include on HDV the MIT License was created on https://github.com/IQSS/dataverse.harvard.edu/issues/248, This PR adds a JSON file so the license can be added.

Closes #10425

You can add the license and check that the link is working


scripts/api/data/licenses/licenseMIT.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "MIT License",
+  "uri": "https://spdx.org/licenses/MIT.html",
+  "shortDescription": "MIT License (MIT).",
```
pdurbin, line 4: I'm just wondering where this short description comes from.

scripts/api/data/licenses/licenseMIT.json
pdurbin, line 1: Can we please add a new section called "Adding Software Licenses" and add this as the first license? In this area: https://guides.dataverse.org/en/6.1/installation/config.html#adding-licenses

pdurbin (commented):
some quick feedback

scripts/api/data/licenses/licenseMIT.json
jp-tosca: I am open to suggestions 🤣 

` Massachusetts Institute of Technology License (MIT).`?

pdurbin:
I asked the community for feedback on this pull request: https://groups.google.com/g/dataverse-community/c/_UUKZT4RrmM/m/HCtul3VlAQAJ

pdurbin:
There's some discussion going on at https://dataverse.zulipchat.com/#narrow/stream/379673-dev/topic/first.20software.20license.20in.20guides.3A.20MIT/near/429826659

pdurbin:
@jp-tosca and I just pushed 5e0c73f to update the MIT license and add guidance on how to add additional licenses:

Screenshot 2024-04-11 at 12 44 09 PM

Here's a preview of the docs: https://dataverse-guide--10426.org.readthedocs.build/en/10426/installation/config.html#contributing-to-the-collection-of-standard-licenses-above

At list point we should solicit more input from the community, especially on the guidance above. The MIT license we're adding is slightly different (different URI, at least) than the one @DieuwertjeBloemen mentioned adding in https://groups.google.com/g/dataverse-community/c/_UUKZT4RrmM/m/IxQaA7ycAQAJ

Also, I'm interested in what @philippconzett thinks since he's been leading the charge on license standardization in these issues and PRs:

- #8512
- #9262

I'll try to dig up the right threads on the google group to have more people look. For now, I guess I'll leave myself as a reviewer.

philippconzett:
Thanks all for driving standardized license information forward! I have a couple of questions:
1. Does this PR cover #8512?
2. What is the rationale for using the actual URL that the SPDX license link (in some/most cases) redirects to as value in the uri field? I wonder whether using the SPDX license link would spare us from monitoring link rot?
3. Does this PR include adapting the database to include table fields where all the SPDX values (name, description, uri, ...) are stored?
4. Should the PR include scripts for how to clean up / align licence information in legacy datasets, so that the new approach is applied to the entire Dataverse installation?

DieuwertjeBloemen:
@pdurbin Look great! I think my MIT uri is the faulty one, as the url in the SPDX list has the lower-case variant.

@philippconzett 

1. I think it covers a lot if not all of #8512 though it doesn't provide a set of standard licenses straight away, but rather guidelines on how to contribute new JSONs/Licenses to ensure they are standardized. This will probably make these JSONs grow over time. (we could already do an initial push of some JSONs we have at KU Leuven that are pretty much in line with this, I'll just check them once the guidelines are approved).
2. We decided not to use the SPDX landing page because we looked for the uri's that harvesters expected (e.g. for the creative commons, that's where you find the one [OpenAire expects](https://guidelines.openaire.eu/en/latest/data/field_rights.html#rightsuri-ma) and [DataCite](https://datacite-metadata-schema.readthedocs.io/_/downloads/en/4.5/pdf/) (page 40) base URL-wise). I think the only real way to prevent link-rot would be to have a local version of each license text, but I don't think anyone wants to maintain or do that ;) We're also not guaranteed of the SPDX landing page URLs infinite availability.

philippconzett:
Thanks, @DieuwertjeBloemen. I had to revisit my issue/PR once more and now realize that the main difference between #8512 and #10426 is that #8512 is based on the DataCite recommendations, whereas #10426 is still based on the setup Dataverse uses currently, with some modifications. I think you clearly see the difference when you compare how the JSON file for CC BY 4.0 looks like in the two approaches:

JSON according to #8512:
{
  "rightsName": "CC BY 4.0",
  "rightsURI": "https://creativecommons.org/licenses/by/4.0/",
  "rightsIdentifier": "CC-BY-4.0",
  "rightsIdentifierScheme": "SPDX",
  "schemeURI": "https://spdx.org/licenses/",
  "rightsShortDescription": "Creative Commons Attribution 4.0 International.",
  "rightsIconUrl": "https://licensebuttons.net/l/by/4.0/88x31.png",
  "rightsActive": true
}

JSON according to #10426:
{
  "name": "CC-BY-4.0",
  "uri": "http://creativecommons.org/licenses/by/4.0",
  "description": "CC BY 4.0",
  "iconUrl": "https://licensebuttons.net/l/by/4.0/88x31.png",
  "active": true,
  "sortOrder": 2
}

I guess you want to add the MIT license as soon as possible, for which #10426 seems to be a feasible way. At DataverseNO, we still would like to be able to deliver license metadata to DataCite in line with their recommendations, which will mean implementing #8512, which will take some more ressources, I guess, because fields need to be added and renamed in the database, among other things.

pdurbin:
Hi @philippconzett and @DieuwertjeBloemen thanks for your comments!

Yes, this PR (#10426) is quite small, only adding the MIT license using existing database columns/tables and adding some new documentation/guidance on adding new licenses moving forward (still using the existing columns/tables).

As for letting the SPDX link resolve or not, I'm happy to reverse the stance we've taken and declare that we should use the SPDX link as-is without redirection. Mostly I just wanted to capture the (frustrating) fact that redirection is going on and to pick one way or the other (as-is or redirected). More on this below.

Yes, I think we should leave #8512 open to think about adding additional database columns and further improving how we store licenses in the database and standardize them.

As for SQL migration scripts to handle existing licenses that are not in compliance with the guidance we've written up (CC0 and friends) @jp-tosca and talked about out but decided this work is out of scope for this issue. You may have noticed that we added this note to the guidance: `Note that prior to Dataverse 6.2, various license above have been added that do not adhere perfectly with this procedure. For example, the name for the CC0 license is CC0 1.0 (no dash) rather than CC0-1.0 (with a dash). We are keeping the existing names for backward compatibility. For more on standarizing license configuration, see https://github.com/IQSS/dataverse/issues/8512`. Basically, we talked briefly about how it would be a fair amount of work to write these scripts so we'd rather defer this until #8512.

As for providing additional licenses, yes, sure, we're open to more. After this PR (#10426) gets finalized and merged, @DieuwertjeBloemen you're welcome to add more. Thanks!

So! In the interest of keeping things moving, it sounds like we're all more or less in agreement of the scope of this PR (#10426) as well as its content, with the possible exception of this line...

`- For the ``uri`` field, go to the SPDX landing page for the license and click on the link under "other web pages for this license". Let any redirection happen and then copy the URL (e.g. ``https://opensource.org/license/mit``) into the ``uri`` field.`

If we change the language to use the exact URL as shown on the SPDX landing page (rather than letting redirection happen), we would change...

`"uri": "https://opensource.org/license/mit",`

to

`"uri": "https://opensource.org/license/mit/",`

Again, I don't feel strongly about this. Does anyone?

qqmyers:
CC-BY-4.0 is an interesting one: https://spdx.org/licenses/CC-BY-4.0.html points you to https://creativecommons.org/licenses/by/4.0/legalcode where the CC folks tell you the canonical URL is  https://creativecommons.org/licenses/by/4.0/. There are no redirects. 

In general, I'd think we'd want the canonical URL as defined by the license provider (versus wherever spdx points if that's different) but I agree the question of a trailing slash is painfully trivial (especially when CC and MIT appear to choose opposite conventions!). I wouldn't be surprised if people trying to parse these can handle that much difference, but who knows.

DieuwertjeBloemen:
Yesterday I was also wondering if maybe an example (one of the CC options perhaps) should be chosen in addition to the MIT license for the documentation that examplifies if the "name" field in the JSON is with or without the dash, because MIT is not an example that makes this explicit. But that's just a minor detail that I thought could perhaps be improved in the above-mentioned documentation.

pdurbin:
@DieuwertjeBloemen good idea.

@jp-tosca in the docs, can you please switch from MIT to another license as the example? Please feel free to add an additional license in the process, one that exercises the rules a little more thoroughly.

@DieuwertjeBloemen as we wrote in the guidance above, we are considering the existing CC0 licenses grandfathered in: `Note that prior to Dataverse 6.2, various license above have been added that do not adhere perfectly with this procedure. For example, the name for the CC0 license is CC0 1.0 (no dash) rather than CC0-1.0 (with a dash). We are keeping the existing names for backward compatibility.` What do you think?

DieuwertjeBloemen:
@pdurbin It think it makes sense to 'grandfather' it for now. If someone at some points wants to make it compatible with the rest and figure out how to update it on existing datasets, then that can always be done at a later stage.

pdurbin:
@DieuwertjeBloemen yeah, updating the CC license to comply with the new thinking will require an SQL migration script (in Flyway). I'm glad you're ok with this being out of scope for this PR.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1786,6 +1786,31 @@ JSON files for `Creative Commons licenses <https://creativecommons.org/about/ccl
...
+
+If you do not find the license JSON you need above, you are encouraged to contribute it to this documentation. Following the Dataverse 6.2 release, we have standardized on the following procedure:
+
+- Look for the license at https://spdx.org/licenses/
+- ``cd scripts/api/data/licenses``
+- Copy an existing license as a starting point.
+- Name your file using the SPDX identifier. For example, if the identifier is ``Apache-2.0``, you should name your file ``licenseApache-2.0.json``.
+- For the ``name`` field, use the "short identifier" from the SPDX landing page (e.g. ``Apache-2.0``).
+- For the ``description`` field, use the "full name" from the SPDX landing page (e.g. ``Apache License 2.0``).
+- For the ``uri`` field, go to the SPDX landing page for the license and remove the extension ``.html`` from your browser and then copy the URL (e.g. ``https://spdx.org/licenses/Apache-2.0``) into the ``uri`` field.
```
pdurbin, line 1808: @jp-tosca at standup this morning @qqmyers said something about going to DataCite Fabrica and doing a search for the license using its SPDX identifier. And from there we could/should use the URL that DataCite uses (presumably from any of the returned results).

I'm not sure if I'm doing this right but if you type `rightsList.rightsIdentifierScheme:SPDX AND rightsList.rightsIdentifier:Apache-2.0` at https://commons.datacite.org you'll find a number of works that are licensed under Apache 2.

Maybe this is a starting point? Or the wrong direction? I'm not sure. For Apache 2, DataCite is giving back two different URLs when you export the metadata.

qqmyers: I meant going to https://doi.test.datacite.org/ (with a valid account) and creating a DOI using their form. They show a list from spdx and then pick a specific URI for you (the rightUri field is greyed out/non-editable if you pick a license from the list): 
image
which is what eventually shows in their DataCite XML. They are not using the spdx URI itself for Apache2, CC or MIT at least.

pdurbin: @qqmyers thank. I had a play with it. Check this out. They seem to simply return [the first](https://github.com/datacite/bracco/blob/670f34a0d872cb90492cbb641caee5bab3c6f122/app/components/doi-rights.js#L29C9-L29C59) in an array...

`this.fragment.set('rightsUri', rights.seeAlso[0]);`

So for the [EFL-2.0](https://github.com/datacite/bracco/blob/670f34a0d872cb90492cbb641caee5bab3c6f122/app/spdx.js#L1385C1-L1398C7) license...

    {
      reference: './EFL-2.0.html',
      isDeprecatedLicenseId: false,
      isFsfLibre: true,
      detailsUrl: 'http://spdx.org/licenses/EFL-2.0.json',
      referenceNumber: '7',
      name: 'Eiffel Forum License v2.0',
      licenseId: 'EFL-2.0',
      seeAlso: [
        'http://www.eiffel-nice.org/license/eiffel-forum-license-2.html',
        'https://opensource.org/licenses/EFL-2.0'
      ],
      isOsiApproved: true
    },

... they don't show the opensource.org URL (like they do for MIT):

Screenshot 2024-04-23 at 2 14 56 PM

I guess I'm left wondering... is this array curated by DataCite to put the best URL first? Or it is fairly arbitrary?

jp-tosca: I think this one is an interesting example @pdurbin while searching on google to see what would be the canonical URL, searching for "Eiffel forum license 2.0" shows the first result as https://www.gnu.org/licenses/eiffel-forum-license-2.html, search for "eiffel license" then the first result is "https://www.eiffel.com/eiffelstudio/licensing/" so this URL "http://www.eiffel-nice.org/license/eiffel-forum-license-2.html" is not between the first results.

jp-tosca: Also as I mentioned on the call the second link that comes out of the JSON from Datacite is expired "https://opensource.org/licenses/EFL-2.0" shows an  **Oops! That page can’t be found.**

qqmyers: Looks like SPDX isn't being kept up to date (or picked the wrong URL to start) - opensource now has that license at https://opensource.org/license/ver2_eiffel-php.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1858,6 +1858,31 @@ JSON files for `Creative Commons licenses <https://creativecommons.org/about/ccl
...
+
+If you do not find the license JSON you need above, you are encouraged to contribute it to this documentation. Following the Dataverse 6.2 release, we have standardized on the following procedure:
+
+- Look for the license at https://spdx.org/licenses/
+- ``cd scripts/api/data/licenses``
+- Copy an existing license as a starting point.
+- Name your file using the SPDX identifier. For example, if the identifier is ``Apache-2.0``, you should name your file ``licenseApache-2.0.json``.
+- For the ``name`` field, use the "short identifier" from the SPDX landing page (e.g. ``Apache-2.0``).
+- For the ``description`` field, use the "full name" from the SPDX landing page (e.g. ``Apache License 2.0``).
+- For the ``uri`` field, we encourage you to use the same resource that DataCite uses, most of the times this will be the same than the first provided resource on SPDX, but please get in contact with us if you have any questions on the cominity group.
```
qqmyers, line 1880: Right now the new licenses don't strictly follow this advice. The easiest change might be to just use the DataCite URIs for both Apache 2 and MIT licenses below. Alternately, if there's interest in using other forms (e.g. using https for the Apache URL and license/MIT - see below) perhaps this could be softened: 
we encourage you to use the same resource that DataCite uses, which is often the same as the first "Other web pages for this license" on the SPDX page for the license. When these differ, or there are other concerns about the URI DataCite uses, please reach out to the community to see if a consensus can be reached.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1858,6 +1858,31 @@ JSON files for `Creative Commons licenses <https://creativecommons.org/about/ccl
...
+- For the ``active`` field, put ``true``.
+- For the ``sortOrder`` field, put the next sequention number after checking previous files with ``grep sortOrder scripts/api/data/licenses/*``.
```
qqmyers, line 1882: typo : sequential number

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1858,6 +1858,31 @@ JSON files for `Creative Commons licenses <https://creativecommons.org/about/ccl
...
+
+Note that prior to Dataverse 6.2, various license above have been added that do not adhere perfectly with this procedure. For example, the ``name`` for the CC0 license is ``CC0 1.0`` (no dash) rather than ``CC0-1.0`` (with a dash). We are keeping the existing names for backward compatibility. For more on standarizing license configuration, see https://github.com/IQSS/dataverse/issues/8512
```
qqmyers, line 1884: If we don't use the DataCite URIs for these two, this should drop the "prior to Dataverse 6.2": Note that not all of the example licenses follow the recommended procedure above. These represent legacy cases or cases where community discussion has led the contributors to deviate from the recommendation. Given that internal consistency across Dataverse installations is important for interoperability and creating commons tools for exporting and parsing Dataverse metadata, we encourage you to use the provided .json files rather than creating new ones.

qqmyers (approved):
Looks good.

PR #10443, opened 2024-03-28
#10442 fix null from template

### LINKED ISSUE #10442: Datasets created Via Template are not indexed
Create any dataset via the UI and a template. The draft is created successfully but it is not indexed

* When does this issue occur? On dataset create

* Which page(s) does it occurs on? Create Dataset

* What happens? created draft does not appear on home page

* To whom does it occur (all users, curators, superusers)? anyone who can create a dataset

* What did you expect to happen? I expect the draft to be indexed/findable

**Which version of Dataverse are you using?** current dev

**Any related open or closed issues to this bug report?** unknown 

**Screenshots:**

*

sekmiller:
**What this PR does / why we need it**: Fixes an issue where new datasets created in UI with template are failing to index

Closes #10442 Datasets created via template are not indexed

**Special notes for your reviewer**: This is a BandAid(tm) fix. Needed for 6.2. Should probably revisit templates to see why there are nulls getting into the controlled vocab values.

**Suggestions on how to test this**: This only seems to effect templates that were created prior to the current release/dev environment, so to test you need to go to Dataverse-internal and create a dataset using the template named  "check23".  A dataset created from this template will not be indexed if you are in the current dev branch, but it will be indexed in this PR branch. Create a dataset with that template. It should be indexed properly.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
@@ -1016,7 +1016,7 @@ public class IndexServiceBean {
                              */
                             if (dsf.getControlledVocabularyValues().isEmpty()) {
                                 for (DatasetFieldValue dfv : dsf.getDatasetFieldValues()) {
-                                    if (dfv.getValue().equals(DatasetField.NA_VALUE)) {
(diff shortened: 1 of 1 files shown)


pdurbin:
Related PR:

- #10323

scolapasta (approved):

pdurbin:
I just tested it on dataverse-internal (before and after the fix). Works great. Merging.

PR #10444, opened 2024-03-28
fix bad curl url in doc

stevenwinship:
Fix curl url in doc by adding /makeDataCount


pdurbin (approved):
Looks good. Approved. Sorry I didn't catch this in the last go 'round. Will merge.

Issue #10217, opened 2024-01-09
Add Source/Server name to OAI-PMH Harvesting clients


DS-INRAE:
**Overview of the Feature Request**
As a superuser, I can use a new field to document the source repository in OAI-PMH harvesting clients :
![image](https://github.com/IQSS/dataverse/assets/46443753/f571e302-4cf0-4ef4-9b2b-5ccd940394fe)

In order to (see related issues) filter content from various harvested repositories with a search facet.

One source/server name should be able to be used for several clients (e.g. to dispatch several sets from the same repository in different dataverse collection).

**What kind of user is the feature intended for?**
(Example users roles: API User, Curator, Depositor, Guest, Superuser, Sysadmin)
Superuser

**What inspired the request?**
Needed for our harvested repositories.

**Any open or closed issues related to this feature request?**
- #10195
- Related discussion: https://dataverse.zulipchat.com/#narrow/stream/375707-community/topic/add.20Server.20Name.20to.20harvesting.20client/near/452663590

jp-tosca:
Hi @DS-INRAE thanks for reaching out, it has been a while since we looked at this issue but we can bring it up to the team. This was mentioned on #10464 and reading through the comment I wonder if this last [comment](https://github.com/IQSS/dataverse/pull/10464#discussion_r1556261875) from @landreev is where we can pick up from this.

DS-INRAE:
Yes,
> What would happen if someone has multiple clients harvested into the same collection? 🤔
To be sure to understand is it the case where we have something akin to this ?
- Harvesting client "A" with Server name "Source_1" in collection Z
- Harvesting client "B" with Server name "Source_1" in collection Z
- Harvesting client "C" with Server name "Source_2" in collection Z
In this case (hopefully it's the right one), it should be 
Metadata Sources :
- Source_1 : datasets harvested from A+B
- Source_2 : datasets harvested from C

Do you want me to also copy this in the comment so it stays in the thread ?

luddaniel:
Hi @jp-tosca, are you still on the subject or can we develop it ? (not much code normally)

jp-tosca:
Sorry @luddaniel I see that I was still assigned to this but I removed myself. I will bring it up again to the team when I see Leo to see if this can be planned or prioritized.

PR #10464, opened 2024-04-04
Add Harvesting Client name to the Metadata Source facet

### LINKED ISSUE #10298: Add Harvesting Source to search facets
Following the addition of a source name for harvesting clients :
- #10217
This information should be useable as a search facet to filter results.

API User, Guests

 Needed for our harvested repositories.

Modify the current search facet "Metadata Source" to include the list of Sources from harvesting clients.

- #10195 (macro issue for centralization)
- #10217

jp-tosca:

This PR will change the "Metadata Source" facet. It will stop showing all the Harvested datasets inside the same category and group them by the nickname of the client. ›

**Which issue(s) this PR closes**: 10298

Closes #10298

We may need to update this when #10217 gets done with that value. 

Create multiple harvesting clients and you should be able to test the change on the facet.

***
**Before**: 

image

***
**After**: 
image

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
@@ -897,7 +897,8 @@ public class IndexServiceBean {
 
         if (dataset.isHarvested()) {
             solrInputDocument.addField(SearchFields.IS_HARVESTED, true);
-            solrInputDocument.addField(SearchFields.METADATA_SOURCE, HARVESTED);
+            solrInputDocument.addField(SearchFields.METADATA_SOURCE,
+                                        dataset.getHarvestedFrom() != null ? dataset.getHarvestedFrom().getName() : HARVESTED);
         } else {
             solrInputDocument.addField(SearchFields.IS_HARVESTED, false);
             solrInputDocument.addField(SearchFields.METADATA_SOURCE, rdvName); //rootDataverseName);
--- a/src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
@@ -15,9 +15,7 @@ import static jakarta.ws.rs.core.Response.Status.UNAUTHORIZED;
 import static jakarta.ws.rs.core.Response.Status.ACCEPTED;
 import static jakarta.ws.rs.core.Response.Status.OK;
 import static org.hamcrest.CoreMatchers.equalTo;
-import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.*;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * This class tests Harvesting Client functionality. 
@@ -272,6 +270,12 @@ public class HarvestingClientsIT {
         } while (i<maxWait); 
         
         System.out.println("Waited " + i + " seconds for the harvest to complete.");
+
+        Response searchHarvestedDatasets = UtilIT.search("metadataSource:" + nickName, normalUserAPIKey);
+        searchHarvestedDatasets.prettyPrint();
+        searchHarvestedDatasets.then().assertThat()
+                                .statusCode(OK.getStatusCode())
+                                .body("data.total_count", equalTo(expectedNumberOfSetsHarvested));
         
         // Fail if it hasn't completed in maxWait seconds
         assertTrue(i < maxWait);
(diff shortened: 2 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -897,7 +897,8 @@ public SolrInputDocuments toSolrDocs(IndexableDataset indexableDataset, Set<Long
 
         if (dataset.isHarvested()) {
             solrInputDocument.addField(SearchFields.IS_HARVESTED, true);
-            solrInputDocument.addField(SearchFields.METADATA_SOURCE, HARVESTED);
+            solrInputDocument.addField(SearchFields.METADATA_SOURCE,
```
qqmyers, line 900: Looks good. Probably need a release note about this feature that also notes that (async/background) reindexing is needed to populate the facet.

qqmyers (changes requested):
Looks good. Asked for a release note. The tests haven't completed yet so I'm not sure if there's any Harvesting/search related test that needs to be updated. (Can one be added - do we already have harvested test datasets where we could search on this field?)

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
jp-tosca: Hi @qqmyers 👋🏼 

Thanks! I will add the release note in a few just looking at some other PR right now but I will add it ASAP. Regarding the tests, I am not sure I will also look at this. 

Best,
Juan

jp-tosca: Note added and test condition added to the harvesting test to search by the new collection name. 😃

qqmyers (approved):
Looks good!

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
landreev: Putting what I said during standup in writing: 
Rather than using the HarvestingClient's nickname for this facet, we should probably use the name of the local collection into which the client is harvesting. 
The clear advantages of doing that: 
1) This mirrors what we are doing for the local datasets: ```solrInputDocument.addField(SearchFields.METADATA_SOURCE, rootDataverse.getName());```
2) The name of the collection is likely to be more descriptive/better-looking to a human user
3) While both the client nickname and the name of the local collection are chosen by the local admin, it is far easier to change the latter.  The former is not editable at all. With the current implementation, if a prod. instance admin realizes that they named the harvesting client ```oai_3``` and that's what's going to show up in the facet, the only way for them to address it would be to delete the client (and all the content associated with it), and re-create it with a better-looking nickname, then re-harvest. 
4) This will make it unnecessary to add an extra field with a descriptive label to the client class (as was mentioned during standup).

landreev:
(I'm generally ready to merge this)

landreev:
@pdurbin Since we're still going to be showing the total number of local datasets, it should still be fairly clear how many harvested datasets total there are - because, math ("datasets" facets - total local); even if we are not showing the exact total number. 
IMO, this would be useful to the users implemented like this. But, if anyone complains - we can revisit it and add a quick setting for reversing the behavior.

jp-tosca:
We are having a meeting today at 11:00 EST with @DS-INRA, I will post if there are any updates.

src/test/java/edu/harvard/iq/dataverse/api/HarvestingClientsIT.java
```diff
@@ -272,6 +270,12 @@ private void harvestingClientRun(boolean allowHarvestingMissingCVV)  throws Inte
         } while (i<maxWait); 
         
         System.out.println("Waited " + i + " seconds for the harvest to complete.");
+
+        Response searchHarvestedDatasets = UtilIT.search("metadataSource:" + nickName, normalUserAPIKey);
```
landreev, line 274: Let's keep an eye on it - will not surprise me if it needs an extra second here, occasionally, for the indexing of all the harvested datasets to complete.

landreev: @jp-tosca @stevenwinship 
Hah! This appears to be exactly what's happening now, since I merged this PR yesterday. 
The new, weird failures - ```data.total_count doesn't match. Expected: <8>  Actual: <5>```

landreev: JP, this was a bad call on my part - I should've asked you to run Jenkins 3 times in a row on the PR before merging it.

PR #10498, opened 2024-04-16
Extended Dataverse metadatablocks API to retrieve metadata fields configured as required in the General Information pag page

GPortas:
…fields displayed on create in a dataverse

Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Dataverse.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataverse.java
@@ -411,6 +411,14 @@ public class Dataverse extends DvObjectContainer {
         return dataverseFieldTypeInputLevels;
     }
 
+    public boolean isDatasetFieldTypeRequiredAsInputLevel(Long datasetFieldTypeId) {
+        for(DataverseFieldTypeInputLevel dataverseFieldTypeInputLevel : dataverseFieldTypeInputLevels) {
+            if (dataverseFieldTypeInputLevel.getDatasetFieldType().getId().equals(datasetFieldTypeId)) {
+                return true;
+            }
+        }
+        return false;
+    }
 
     public Template getDefaultTemplate() {
         return defaultTemplate;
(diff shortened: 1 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseInputLevelsCommand.java
```diff
@@ -0,0 +1,50 @@
+package edu.harvard.iq.dataverse.engine.command.impl;
```
sekmiller, line 1: Instead of adding this new command can you use the original UpdateDataverseCommand and pass in the inputlevel list?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseInputLevelsCommand.java
```diff
@@ -0,0 +1,50 @@
...
+        this.inputLevelList = new ArrayList<>(inputLevelList);
+    }
+
+    @Override
+    public Dataverse execute(CommandContext ctxt) throws CommandException {
+        if (inputLevelList == null) {
+            throw new CommandException("Exception while updating dataverse input levels: Input level list cannot be null", this);
+        }
+        addInputLevelMetadataBlocks();
+        return ctxt.engine().submit(new UpdateDataverseCommand(dataverse, null, null, getRequest(), inputLevelList));
```
sekmiller, line 33: Also you need to set isMetadataBlockRoot to true so that the Dataverse knows that it not inheriting its MetadataBlocks from its parent - that is it has a different input level, etc.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseInputLevelsCommand.java
GPortas: The `UpdateDataverseInputLevelsCommand` command wraps the `UpdateDataverseCommand` command by adding the metadata block update logic, which is not included in the `UpdateDataverseCommand`.

I explain this in more detail in the PR description section: _Special notes for your reviewer_

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseInputLevelsCommand.java
GPortas: I just added that line and also check that the input list is not empty

sekmiller (changes requested):
There's one change that needs to be made with respect to updating isMetadataBlockRoot. Also suggesting that you might not need the new command.

sekmiller (approved):

PR #10503, opened 2024-04-18
CVOC : Adding hidden metadata fields (Ontoportal integration)

luddaniel:
Following agreements on next steps of https://github.com/IQSS/dataverse/pull/10145

Dataset metadata view of a CVOC field is by default only `term-uri-field` value of the `CVocConf:`
Historically, this field is an idenfier (ORCID ID or skomos permalink), allowing to reconstruct data depending on user language (ex. [skosmos.js#L31](https://github.com/gdcc/dataverse-external-vocab-support/blob/main/scripts/skosmos.js#L31)) while loading the related javascript file (`js-url`) of `CVocConf:`.
Ontoportal uses 2 fields to be able to reconstruct the data. In order to minimize code modifications, we decided to add missing child fields as hidden `<span>`.

**HTML Before PR :**

```html
<span data-cvoc-headers="{&quot;Authorization&quot;:&quot;apikey token=xxx-xxx-xxx&quot;}" data-cvoc-service-url="https://data.agroportal.lirmm.fr/" data-cvoc-protocol="ontoportal" lang="fr">http://purl.obolibrary.org/obo/OBT_000193</span>
```
**HTML After PR :**

```html
<span class="hidden" data-cvoc-metadata-name="keywordValue" hidden="hidden">animal</span>
<span data-cvoc-headers="{&quot;Authorization&quot;:&quot;apikey token=xxx-xxx-xxx&quot;}" data-cvoc-service-url="https://data.agroportal.lirmm.fr/" data-cvoc-protocol="ontoportal" lang="fr">http://purl.obolibrary.org/obo/OBT_000193</span>
<span class="hidden" data-cvoc-metadata-name="keywordVocabulary" hidden="hidden">ONTOBIOTOPE</span>
<span class="hidden" data-cvoc-metadata-name="keywordVocabularyURI" hidden="hidden">https://agroportal.lirmm.fr/ontologies/ONTOBIOTOPE</span>
```

Allowing to do this in Javascript : 

```javascript
var ontology = parent.find('[data-cvoc-metadata-name="keywordVocabulary"]').text();
let termName = parent.find('[data-cvoc-metadata-name="keywordValue"]').text();
let url = cvocUrl.replace("data.", "") + "ontologies/"+ontology+"/classes/"+encodeURIComponent(id);
```

Full javascript : [ontoportal.js (POC)](https://github.com/Recherche-Data-Gouv/dataverse-external-vocab-support/blob/poc_ontoportal/scripts/ontoportal.js)

Check html fields on Dataset metadata view using `CVocConf:` that uses `managed-fields`.

**Does this PR introduce a user interface change?**

No

### DIFF
--- a/doc/release-notes/10503-cvoc-hidden-html-fields.md
+++ b/doc/release-notes/10503-cvoc-hidden-html-fields.md
@@ -4,5 +4,8 @@
 
 #### Hidden HTML Fields
 
-To enlarge javascript possibilities of [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) configured service, the child fields of the configured field had been added to the dataset metadata view.
-Those values are hidden and can be found with the html attribute `data-cvoc-metadata-name`. (#10503)
+External Controlled Vocabulary scripts, configured via [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf), can now access the values of managed fields as well as the term-uri-field for use in constructing the metadata view for a dataset.
+
+Those values are hidden and can be found with the html attribute `data-cvoc-metadata-name`.
+
+For more information, see [#10503](https://github.com/IQSS/dataverse/pull/10503).
--- a/doc/release-notes/10503-cvoc-hidden-html-fields.md
+++ b/doc/release-notes/10503-cvoc-hidden-html-fields.md
@@ -1,0 +1,8 @@
+## Release Highlights
+
+### Updates on Support for External Vocabulary Services
+
+#### Hidden HTML Fields
+
+To enlarge javascript possibilities of [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) configured service, the child fields of the configured field had been added to the dataset metadata view.
+Those values are hidden and can be found with the html attribute `data-cvoc-metadata-name`. (#10503)
--- a/src/main/webapp/metadataFragment.xhtml
+++ b/src/main/webapp/metadataFragment.xhtml
@@ -159,6 +159,17 @@
                                                                       <f:passThroughAttribute name="data-cvoc-protocol" value="#{cvocConf.get(dsf.datasetFieldType.id).getString('protocol','')}" />
                                                                       <f:passThroughAttribute name="data-cvoc-managedfields" value="#{cvocConf.get(dsf.datasetFieldType.id).get('managedfields').toString()}" />
                                                         </h:outputText>
+                                                        <!-- Cvoc for others fields -->
+                                                        <h:outputText value="#{cvPart.key.value}"
+                                                                      escape="#{cvPart.key.datasetFieldType.isEscapeOutputText()}"
+                                                                      rendered="#{(cvocOnDsf or cvocOnCvPart) and not (
+                                                                            cvPart.key.datasetFieldType.name.equals(cvocConf.get(cvPart.key.datasetFieldType.id).getString('term-uri-field'))
+                                                                            or cvPart.key.datasetFieldType.name.equals(cvocConf.get(dsf.datasetFieldType.id).getString('term-uri-field')))}"
+                                                                      styleClass="hidden">
+                                                            <f:passThroughAttribute name="hidden" value="hidden" />
+                                                            <f:passThroughAttribute name="data-cvoc-metadata-name"
+                                                                                    value="#{cvPart.key.datasetFieldType.name}" />
+                                                        </h:outputText>
                                                     </ui:repeat>
                                                     <ui:fragment rendered="#{not compoundValuesstatus.last}">
                                                         <br/>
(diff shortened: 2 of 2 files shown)


luddaniel:
_Release note is coming._

doc/release-notes/10503-cvoc-hidden-html-fields.md
```diff
@@ -0,0 +1,8 @@
+## Release Highlights
+
+### Updates on Support for External Vocabulary Services
+
+#### Hidden HTML Fields
+
+To enlarge javascript possibilities of [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) configured service, the child fields of the configured field had been added to the dataset metadata view.
```
qqmyers, line 7: Suggested change: External Controlled Vocabulary scripts, configured via [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf), can now access the values of managed fields as well as the term-uri-field for use in constructing the metadata view for a dataset.

src/main/webapp/metadataFragment.xhtml
```diff
@@ -159,6 +159,17 @@
...
+                                                                      escape="#{cvPart.key.datasetFieldType.isEscapeOutputText()}"
+                                                                      rendered="#{(cvocOnDsf or cvocOnCvPart) and not (
+                                                                            cvPart.key.datasetFieldType.name.equals(cvocConf.get(cvPart.key.datasetFieldType.id).getString('term-uri-field'))
+                                                                            or cvPart.key.datasetFieldType.name.equals(cvocConf.get(dsf.datasetFieldType.id).getString('term-uri-field')))}"
+                                                                      styleClass="hidden">
+                                                            <f:passThroughAttribute name="hidden" value="hidden" />
+                                                            <f:passThroughAttribute name="data-cvoc-metadata-name"
+                                                                                    value="#{cvPart.key.datasetFieldType.name}" />
+                                                        </h:outputText>
```
qqmyers, line 172: I think this works. A possible alternate design would be to use a data-cvoc-managed-fields element in the term uri span that has a json object with managed field names: values. That might be more SPA friendly. Not sure how how much effort it might take on the Java side to construct the value or whether changing is worth the effort at this point. @luddaniel - any thoughts?

doc/release-notes/10503-cvoc-hidden-html-fields.md
jeromeroucou: Suggested change accepted and commit

src/main/webapp/metadataFragment.xhtml
jeromeroucou: In order to avoid introducing any blocking changes, we have opted to add hidden metadata. So existing Javascript files like [skosmos.js](https://github.com/gdcc/dataverse-external-vocab-support/blob/main/scripts/skosmos.js) or [ror.js](https://github.com/gdcc/dataverse-external-vocab-support/blob/main/scripts/ror.js) don't have to be modified to keep working.

I imagine that for SPA, metadata will be retrieved via API, probably using an endpoint providing the JSON representation of a datasetversion (based on its ID ?). But maybe I'm wrong ?

qqmyers: I don't think there's any backward compatibility issue in either design - the old scripts are either ignoring extra hidden DOM elements or extra attributes in the DOM element they normally look at.

My point about SPA-friendly may be incorrect - I was assuming, like I think you are, that the SPA would be getting the list of fields/values via some API, as JSON, and then guessing that dumping the JSON for the necessary fields into an attribute would be less work than generating more DOM elements. That may or may not be the case. I wouldn't be surprised if the SPA needs breaking changes in the cvoc mechanism anyway, so this could be something we discuss when the SPA cvoc design starts.

Overall, I'm raising this for two reasons - one is just to write down the possible alternative somewhere, and two is to see if you thought this was worth making a change now. If not (doesn't sound like it) I don't think it's a big enough issue to hold up the PR/ go discuss in a larger group, etc., so I can go ahead and approve the PR as is.

luddaniel: Hello @qqmyers , we discussed with @jeromeroucou to understand the depth of your point, here are our views :
- There is no simple path to anticipate the idea of preparing key/values fetching of cvoc metadata (most of it is done in JSF, cvoc is a static setting, fields are dynamics and can be repeatables, [Get JSON Representation of a Dataset](https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-dataset) requires ATM API Token so it cannot be use by javascript for now...)
- We agree to treat this out of this PR
- We agree with you that we must anticipate SPA portability (we discused about it in this meeting https://groups.google.com/g/dataverse-community/c/4eTd3kfJeBU/m/fPQix_bUAAAJ) and I have mentionned it here [#9276#issuecomment-1830172630](https://github.com/IQSS/dataverse/issues/9276#issuecomment-1830172630) _Import subject : what will happen of CVOC feature under SPA Dataverse architecture ?_ but no one seems to have created the related issue in https://github.com/IQSS/dataverse-frontend
- First idea we have got was to add in [Get JSON Representation of a Dataset](https://guides.dataverse.org/en/latest/api/native-api.html#get-json-representation-of-a-dataset) endpoint the id of the related cvoc conf to the field when applicable then offer an new endpoint to fetch cvoc details of the field... We could take time to specify this before the SPA version is released.

qqmyers: FWIW: My initial thought would just be to have a get DatasetPage.getCvocManagedFields(DatasetField dsf) that would get called on line 160 and generate the serialized json for the data-cvoc-managedfields attribute. There shouldn't be any problem navigating from the dsf to the required values for the managed child fields in Java - I agree it would be hard to generate it in the JSF. How the SPA would assemble the info is tbd, but it seems like a cleaner solution to avoid extra hidden fields that have to be generated and that the Ontoportal JS has to go find.

In any case, not too hard to add to the attribute at some point and deprecate the hidden fields if they are used in many scripts.

qqmyers (approved):
Looks good.

Issue #10317, opened 2024-02-12
Some tiff files don't generate thumbnails


juancorr:
**What steps does it take to reproduce the issue?**
Upload a tiff file in a dataset using the normal procedure . You can download examples from here: https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/LNCJWK

* When does this issue occur?
This happens with some tiff files and not with others. 

* What happens?
Thumbnails are not generated sometimes uploading tiff files. The XXXXXXXXX-XXXXXXXX.thumbYY files are generated but with size 0. The alternative text is shown from the parent dataverse and not image is shown in the file preview. This problem is not dependent on the file size or the image file size limits in the database.

* To whom does it occur (all users, curators, superusers)?
All users.

**Which version of Dataverse are you using?**
v. 6.1 build iqss-prod-patch  (demo Dataverse) and v.5.10.1

**Any related open or closed issues to this bug report?**
None related

**Screenshots:**
Tiff files can't be uploaded, but here are two screenshots

![imagen](https://github.com/IQSS/dataverse/assets/325896/4256ec6c-b1d9-41ed-91b9-4999d8928981)

![imagen](https://github.com/IQSS/dataverse/assets/325896/4f915584-ef8a-4d63-9188-ec662bab3ffe)

juancorr:
If had a problem generating tiff thumbnails with the native java library, a possible solution could be use the external Imagemagick convert tool as pdf thumbnails are generated. We had generated the tiff thumbnails in the filesystem running the convert tool from the command line here: https://edatos.consorciomadrono.es/dataset.xhtml?persistentId=doi:10.21950/KTALA8

pdurbin:
@juancorr thanks! I'm uploading the test files here as a zip because we clear out the demo server from time to time: 
[Archive.zip](https://github.com/IQSS/dataverse/files/14241257/Archive.zip)

pdurbin:
@juancorr you should expect some different behavior as of this PR (a default image):

- #10509

pdurbin:
@juancorr sure, you're welcome to close this issue if you like. Generally speaking, please keep any eye on tiff support and feel free to open new issues in the future!

There's also this issue to watch:

- #10351

juancorr:
Closing this issue because the solution it being developed with #10509 and #10351.

juancorr:
[...]

PR #10509, opened 2024-04-19
catch failures with tiff file thumbnail generation

### LINKED ISSUE #10352: We need a more graceful failure mode: TIFF images  failure is not detected and a size-0 thumbnail is cached on s3 - and that results in the broken page view.
Normally when a thumbnail cannot be generated Dataverse detects the failure and then Jim's recently improved code marks the image as thumbnail-less. In this case, with these TIFF images for whatever reason the failure is not detected and a size-0 thumbnail is cached on s3 - and that's what results in the broken page view. (edit: even if there are somehow no exceptions thrown, we could detect that this was a failure by checking the size of the thumbnail produced)

reference
https://github.com/IQSS/dataverse.harvard.edu/issues/250

stevenwinship:
Handles failures with TIFF file thumbnail failures by preventing broken images as thumbnails or previews

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/10352

- Closes #10352

**Suggestions on how to test this**: Use some of the tiff files in question from [[
](https://github.com/IQSS/dataverse.harvard.edu/issues/250)](https://github.com/IQSS/dataverse.harvard.edu/issues/250)

When testing with existing bad tiff files you will see the broken images first since the ui gets the dataset(s) first and doesn't know the images failed. After loading the dataset the db is updated so each time the dataset is loaded going forward the images should look correct (ie default thumbnails with no preview).

Before:
Screenshot 2024-04-19 at 10 08 43 AM
Screenshot 2024-04-19 at 10 09 36 AM

After:
Screenshot 2024-04-19 at 10 16 55 AM


qqmyers (changes requested):
The code changes look reasonable - finding places where the failure wasn't getting flagged and making sure the 0 length file is deleted if there's a failure. Also some nice cleanup.

The one add that would be nice is to get an IT test with a tiff file known to fail - is that possible?

Also - I see that jenkins is failing in a new way - with a file not found during the build. Doesn't seem related to this PR.

src/test/java/edu/harvard/iq/dataverse/api/ThumbnailsIT.java
```diff
@@ -35,4 +42,65 @@ public void testDatasetThumbnail() {
...
+        // test set logo with badTiff and tiff to large (goodTiff)
+        Response uploadLogoResponse = UtilIT.uploadDatasetLogo(datasetPersistentId, badTiff, apiToken);
+        uploadLogoResponse.prettyPrint();
+        uploadLogoResponse.then().assertThat().statusCode(FORBIDDEN.getStatusCode());
+        uploadLogoResponse.then().assertThat().body("message", equalTo("In setNonDatasetFileAsThumbnail could not generate thumbnail from uploaded file."));
+
+        uploadLogoResponse = UtilIT.uploadDatasetLogo(datasetPersistentId, goodTiff, apiToken);
+        uploadLogoResponse.prettyPrint();
+        uploadLogoResponse.then().assertThat().statusCode(FORBIDDEN.getStatusCode());
+        uploadLogoResponse.then().assertThat().body("message", equalTo("File is larger than maximum size: 500000."));
```
qqmyers, line 103: FWIW: Including the size makes the test dependent on the default max size.

qqmyers (approved):
Thanks for the test

src/test/java/edu/harvard/iq/dataverse/api/ThumbnailsIT.java
stevenwinship: fixed

qqmyers:
@stevenwinship - Heads up that your change causes a failure in a SearchIT test which is checking for the exact string (looks like you dropped a period in going to the bundled string).

pdurbin:
Does this PR also fix the following issue? Should we put it through QA as well?

- #10317

stevenwinship:
> Does this PR also fix the following issue? Should we put it through QA as well?
> 
> * [Some tiff files don't generate thumbnails #10317](https://github.com/IQSS/dataverse/issues/10317)

The files will now show up with a default image but the tiff files will not generate an actual thumbnail image. I agree that we should switch to another image conversion library, but I don't think it's part of this bug. I'll create another issue to address this.

stevenwinship:
Created new feature issue for enhancing the code
https://github.com/IQSS/dataverse/issues/10515

qqmyers:
Hmm - I thought the code would delete the 0 size file. In any case, https://guides.dataverse.org/en/latest/api/native-api.html#reset-thumbnail-failure-flags are API calls to remove the fail flags. Not sure if that's enough with the changes in this PR to get the thumbs completely reprocessed or not.

landreev:
> Hmm - I thought the code would delete the 0 size file. In any case, https://guides.dataverse.org/en/latest/api/native-api.html#reset-thumbnail-failure-flags are API calls to remove the fail flags. Not sure if that's enough with the changes in this PR to get the thumbs completely reprocessed or not.

I may be missing something, but the api above resets the failure flag; it's the other way around with a preexisting case - the failure flag is false, the previewImageAvailable flag is true. Plus zero-size thumbnail files on disk. 

In my experiments the workaround required BOTH setting ```previewimageavailable=FALSE``` AND deleting the zero-size caches, then reloading the page for the fixes in the PR to take effect. But that also makes me think that perhaps all that's needed is a zero-size check added to the method that checks if a cached copy exists (and then setting ```previewimagefail=TRUE``` if it is zero).

landreev:
Appears to be working as advertised - a pre-existing "bad" thumbnail still looking broken on the first page load, fixed going forward. 
Will test just a bit more, but so far it's looking ready to be merged.

landreev:
@stevenwinship OK, I have one request: 
As of now, everything is working as is should for "bad" TIFFs, but the 0-size caches are kept in storage. (For pre-existing cases, this is true for multiple sizes; for new uploads, it's only the `.thumb64` version). 

Once again, this is not a problem for as long as the images are bad, thumbnails generation-wise. But it does become a problem if they become processable. If/when, for example, we upgrade the image libraries to handle the extra flavors of TIFF. Then the presence of the 0-size cache prevents the re-generation, even after the `previewimagefail` flag is reset. 

I tested this by replacing one of the existing bad tiffs on disk with a tiff known to be processable, adjusting the filesize in the db accordingly and resetting `previewimagefail` to false, then reloading the page. 

There should be an easy way to both not leave that 0-size cache behind, AND delete any existing 0-size caches when detected by the new code from the last commits - is there? 

**Please use your judgement re: whether this is worth any extra dev. effort. If this is remotely non-trivial for whatever reason, and you feel we should instead say that it's the responsibility of the admin to manually delete these 0-size caches when processing fixes become available, I can accept that**.

landreev:
@stevenwinship 
P.S. in that future situation where we expand our TIFF and/or other formats coverage and tell the admins to try to generate the thumbnails again - we will already be asking them to do something manually for the files affected: we will have to tell them to identify the files affected, then use the reset API (`/api/admin/clearThumbnailFailureFlag`) to clear the `previewimagefail` flag on them... so, that would be a reason to argue that adding "... and you need to delete any 0-size `*.thumb*` files in storage and delete them..." to the release note would not be that big of a big deal, possibly? - I'm seriously open to feedback on this.

landreev:
@stevenwinship 
probably unnecessary at this point, since I clarified this part during standup, but this was my hacky way to test the "image becomes processable" case: 
```
ls -l ~/Downloads/known_good_image.tiff
-rw-r--r--@ 1 landreev  staff  2266171 Apr 29 10:27 /Users/landreev/Downloads/known_good_image.tiff
# replace an existing "bad" tiffs:
cp ~/Downloads/known_good_image.tiff /usr/local/payara6/glassfish/domains/domain1/files/10.70122/FK2/HCTR3R/18f2a0f4c3d-c90ea25532ae
```
... then update the file size in the db (I _think_ this is necessary for the whole thing to work) and reset the flag (could've used the API, but did so in the db): 
```
UPDATE datafile SET filesize=2266171 WHERE id=9045
UPDATE dvobject SET previewimagefail=FALSE WHERE id=9045
```
... then reload the dataset page

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
```diff
@@ -463,8 +473,18 @@ public static InputStream getLogoAsInputStream(Dataset dataset) {
...
+                boolean origImageFailed = thumbnailFile.isPreviewImageFail();
+                InputStreamIO isIO = ImageThumbConverter.getImageThumbnailAsInputStream(thumbnailFile.getStorageIO(),
+                        ImageThumbConverter.DEFAULT_DATASETLOGO_SIZE);
+                if (!origImageFailed && thumbnailFile.isPreviewImageFail()) {
+                    // We found an older 0 length thumbnail. Newer image uploads will not have this issue.
+                    // Once cleaned up, this thumbnail will no longer have this issue
+                    // ImageThumbConverter fixed the DataFile
+                    // Now we need to update dataset since this is a bad logo
+                    DatasetServiceBean datasetService = CDI.current().select(DatasetServiceBean.class).get();
+                    datasetService.removeDatasetThumbnail(dataset);
```
landreev, line 485: @stevenwinship Just to follow up on the slack conversation: 
 `datasetService.removeDatasetThumbnail(dataset);` does NOT, and is not supposed to, remove the cached thumbnails associated with a specific DataFile. We may want to come up with a better, and potentially less confusing name for the method. Something like `datasetService.removeDatasetThumbnail(dataset);`. Because that's what it does - the purpose of the method is to make sure the Dataset no longer has a designated thumbnail. If that designated thumbnail was a "custom logo", it will attempt to delete the physical file (because, once we stop using it as the dataset thumbnail, there's no other use for it); but if it was one of the DataFile-level thumbnail, it will only clear the dataset-level  thumbnail designation, but it normally would not have any reason to attempt to delete the caches, because normally they would still be useful elsewhere on the page. 

I do believe that it is entirely safe to just leave the code above as is; I don't think you need to do anything special to delete these 0-size caches here, in the dataset-level thumbnail method; instead, those should be deleted in the datafile-level method, in ```ImageThumbConverter.getImageThumbnailAsInputStream(thumbnailFile.getStorageIO(),ImageThumbConverter.DEFAULT_DATASETLOGO_SIZE);```
where you are checking for that ```size == 0```. 

I hope this makes sense!

landreev: Please note that I corrected an important typo in the above. When I was talking about changing the method name, I meant to say I wanted to name it ```clearDatasetThumbnail()```, instead of ```remove...``` 
I just typed the same name twice earlier ...

landreev: (I'll go ahead and do that)

landreev:
@qqmyers I offered @stevenwinship to add a couple of last minute improvements yesterday. I'm fairly confident they are not going to break anything, but for the sake of due process, if you have a moment before standup, please take a look at the commit https://github.com/IQSS/dataverse/pull/10509/commits/6d1f70a41b09d2601bf8efbcadefe96530076bc7 and see if anything seems off. I'm planning to merge it otherwise. 

The only functional changes are for removing broken/zero size caches. The rest are comments and such.

qqmyers (approved):
Still looks good.

landreev:
@qqmyers @stevenwinship 
This isn't directly relevant to the QA of the fixes in the PR, but I just wanted to share this, as an illustration of how fragile and unpredictable things are in the thumbnail-generating framework: 
The bug that started this, with the unprocessable TIFFs still being marked as `previewimageavailable=TRUE` etc. - it was NOT happening with direct upload: the `previewimagefail` flag was properly set then. So this difference in the workflow - whether the rescaling was first attempted before, or after the file was permanently saved, was enough to account for the flags to be properly set, and/or saved in one case but not in the other. (I'm _guessing_ that was the difference, that is). 

I'll take any further discussion of things like this to the refactoring issue #10515. 
(merging the PR)

Issue #5336, opened 2018-11-21
Code formatting issues


pkiraly:
I just found that Dataverse already have a checkstyle.xml. Most of the rules are commented out, but even the remaining produce 200+ reported violation of the ruleset.

The command to check errors:
```
mvn clean checkstyle:checkstyle
```

This generates the HTML page 'target/site/checkstyle.html'.

It is clear if you open any Java code, that the number of commiters working with different coding style, so the result is not uniformed. Clearer code does not improve the software, but helps lowering the burden of maintenance and making the code more readable. My past experiences with code clearing efforts suggest, that this is an ongoing process: you have to run checks again and again, and clearing code constantly.

Code cleaning on the other hand is an ideal task if you, as a novice in the community just would like to test the water and try to understand the structure of the code. It doesn't require of deep knowledge of the code, and still, you can contribute with some useful thing.

I see three subtasks here:

* run check, and submit pull requests for the clearer code
* extend the checks one by one, increasing the number of checkstyle ruleset
* improving the documentation about the coding style in Dataverse manual (http://guides.dataverse.org/en/4.9.4/developers/coding-style.html)

pkiraly:
Meantime I've started working with the current checkstyle settings, and fixed all the problems. It covers about 200 Java source files.

pdurbin:
@pkiraly um. I'm not sure I want all of that change at once. It's very disruptive to our git history. Can we please talk about this more?

pdurbin:
@pkiraly oh. You already made pull request #5337. I'll drag it to code review at https://waffle.io/IQSS/dataverse and we'll talk about it at standup.

pkiraly:
@pdurbin I am open to discuss it. If this approach doesn't fit for your workflow, we can do differently. The thing is that approx. 200 files have either tabs in it instead of spaces (most of them has sporadic tabs, and only a couple of files has only tabs) for indentation. The other type of issue the checkstyle check is the usage of new lines before right curly braces, such as

```
public someMethod()
{
    ...
}
```

pdurbin:
@pkiraly thanks. Please be advised that standup was cancelled today due to light attendance and the next two days are a holiday (Thanksgiving). Thank you for helping us make the code more consistent. I'm curious about your thoughts on keeping it consistent over time.

poikilotherm:
[...]

PR #10524, opened 2024-04-23
provide guidance to Java devs on wildcard imports

pdurbin:
As discussed in tech hours today.

Relates to:

- #5336


bencomp (approved):
That sounds unhelpful, although in a sense "not caring" is a choice too.

pdurbin:
Yep, the consensus was that we don't have a strong preference either way.

It comes up frequently in code review because we used to lean toward discouraging wildcard imports. 

Now this PR clarifies the current stance.

By writing it down can save time during code review.

qqmyers (approved):

Issue #8814, opened 2022-06-24
bad link on dataverse.org/best-practices/data-citation to old IQSS Data Science team website


pdurbin:
https://dataverse.org/best-practices/data-citation links to http://datascience.iq.harvard.edu but this website doesn't exist anymore.

In context:

>For nearly 20 years, members of [IQSS](http://www.iq.harvard.edu/) and its [Data Science](http://datascience.iq.harvard.edu/) team, who work on the Dataverse Project, have played an active part in the the work to standardize data citation

pdurbin:
Actually the link to https://www.force11.org/datacitation is bad as well.

In context:

> Illustrated in the figure below, is an example of how the data citation is formulated in a Dataverse repository, using the [Joint Declaration of Data Citation Principles (2014)](https://www.force11.org/datacitation) : a synthesis of all previously existing principles and initiatives on data citation.

pdurbin:
Somebody fixed these links. Great! Closing.

PR #10535, opened 2024-04-30
10527 404static.xhtml has an old date in the footer (2023 is hard-coded) and update URLs

### LINKED ISSUE #10527: 404static.xhtml has an old date in the footer (2023 is hard-coded)
If you call up a flat page, an error page "404static.xhtml" appears. This still has an old date in the footer. Maybe you should adjust something there too?

**Screenshots:**
https://github.com/IQSS/dataverse/blob/develop/src/main/webapp/404static.xhtml

Screenshot from 2024-04-24 10-19-09

BenediktMeierUIT:

Closes #10527

JavaScript and update URLs

**Additional documentation**:

### DIFF
--- a/src/main/webapp/404static.xhtml
+++ b/src/main/webapp/404static.xhtml
@@ -1,109 +1,118 @@
 <?xml version='1.0' encoding='UTF-8' ?>
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" lang="en">
-    <head>
-        <title>#{bundle['error.404.page.title']}</title>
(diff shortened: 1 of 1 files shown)


pdurbin (commented):
@BenediktMeierUIT thanks for the pull request.

From a quick look the date looks ok.

Why are there so many other changes? Are they important? Did you consider only changing the date?

BenediktMeierUIT:
I made several changes because otherwise, the page would not be correct, and I would have to report multiple errors. But I can change that if it's better for you.

pdurbin (approved):
I haven't tested it but it looks like a good fix to me.

The "best practices" link doesn't work but let's consider that out of scope. There's a related issue here:

- #8814

pdurbin:
This small fix is a good candidate for 6.4.

PR #10543, opened 2024-05-07
Json dataset validation

### LINKED ISSUE #10169: Feature Request/Idea: JSON Schema Enhancements
In version 6.1 we added a json schema for validating dataset json upload files. The first release only verifies the presence of valid json formatting as well as required elements and required fields (customized by collection). Enhancements which have been requested or should be contemplated.

- Controlled vocabulary support - get valid responses and verify that it/they are present (validate that there may be only one response or multiple responses as specified by the datasetfieldtype)
- field type checking - validate that the values provided conform to the type specified by the datasetfieldtype)
- enhanced error messages - the current validation library does not provide helpful error messages for every exception. See if there are other ways to provide useful information to the user.

Users with edit dataset permission who wish to validate their dataset json prior to upload.

responses to the first release of the dataset schema which were made too late to be included in 6.1

the get dataset schema api and validator

- #9464

stevenwinship:
Overview of the Feature Request
In version 6.1 we added a json schema for validating dataset json upload files. The first release only verifies the presence of valid json formatting as well as required elements and required fields (customized by collection). Enhancements which have been requested or should be contemplated.

- Controlled vocabulary support - get valid responses and verify that it/they are present (validate that there may be only one response or multiple responses as specified by the datasetfieldtype)
- field type checking - validate that the values provided conform to the type specified by the datasetfieldtype)
- enhanced error messages - the current validation library does not provide helpful error messages for every exception. See if there are other ways to provide useful information to the user.

What kind of user is the feature intended for?
Users with edit dataset permission who wish to validate their dataset json prior to upload.

What inspired the request?
responses to the first release of the dataset schema which were made too late to be included in 6.1

What existing behavior do you want changed?
the get dataset schema api and validator
**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/10169

Closes #10169

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
+++ b/src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
@@ -1,0 +1,207 @@
+package edu.harvard.iq.dataverse.validation;
+
+import com.mashape.unirest.http.JsonNode;
+import edu.harvard.iq.dataverse.DatasetFieldServiceBean;
+import edu.harvard.iq.dataverse.DatasetFieldType;
+import edu.harvard.iq.dataverse.util.BundleUtil;
+import jakarta.enterprise.inject.spi.CDI;
+import org.everit.json.schema.Schema;
+import org.everit.json.schema.ValidationException;
+import org.json.JSONArray;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class JSONDataValidation {
+    private static DatasetFieldServiceBean datasetFieldService = null;
+    private static Map<String,List<String>> schemaDTOMap = new ConcurrentHashMap<>();
+
+    /**
+     *
+     * @param schema Schema file defining the JSON objects to be validated
+     * @param jsonInput JSON string to validate against the schema
+     * @throws ValidationException
+     */
+    public static void validate(final Schema schema, String jsonInput) throws ValidationException {
+        if (datasetFieldService == null) {
+            datasetFieldService = CDI.current().select(DatasetFieldServiceBean.class).get();
+        }
+        if (schemaDTOMap.isEmpty()) {
+            // TODO: load from a config file
+            schemaDTOMap.put("datasetContact", Collections.EMPTY_LIST);
+            schemaDTOMap.put("datasetContact.required", List.of("datasetContactName"));
+            schemaDTOMap.put("datasetContact.allowed", List.of("datasetContactName", "datasetContactEmail","datasetContactAffiliation"));
+            schemaDTOMap.put("dsDescription", Collections.EMPTY_LIST);
+            schemaDTOMap.put("dsDescription.required", List.of("dsDescriptionValue"));
+            schemaDTOMap.put("dsDescription.allowed", List.of("dsDescriptionValue", "dsDescriptionDate"));
+        }
+        JsonNode node = new JsonNode(jsonInput);
+        if (node.isArray()) {
+            JSONArray arrayNode = node.getArray();
+            validateObject(schema, "root", arrayNode.toList());
+        } else {
+            node.getObject().toMap().forEach((k,v) -> {
+                validateObject(schema, k, (v instanceof JSONArray) ? ((JSONArray) v).toList() : v);
+            });
+        }
+    }
+
+    /*
+     * Validate objects recursively
+     */
+    private static void validateObject(final Schema schema, String key, Object value) {
+        if (value instanceof Map<?,?>) {
+            validateSchemaObject(schema, key, (Map) value);
+
+            ((Map<?, ?>) value).entrySet().forEach(e -> {
+                validateObject(schema, (String) e.getKey(), e.getValue());
+            });
+        } else if (value instanceof List) {
+            ((List<?>) value).listIterator().forEachRemaining(v -> {
+                validateObject(schema, key, v);
+            });
+        }
+    }
+
+    /*
+     * Validate objects specific to a type. Currently only validating Datasets
+     */
+    private static void validateSchemaObject(final Schema schema, String key, Map valueMap) {
+        if (schema.definesProperty("datasetVersion")) {
+            validateDatasetObject(schema, key, valueMap);
+        }
+    }
+
+    /*
+     * Specific validation for Dataset objects
+     */
+    private static void validateDatasetObject(final Schema schema, String key, Map valueMap) {
+        if (valueMap != null && valueMap.containsKey("typeClass")) {
+            validateTypeClass(schema, key, valueMap, valueMap.get("value"), "dataset");
+        }
+    }
+
+    /*
+     * key: The name of the parent object
+     * valueMap: Map of all the metadata of the object
+     * value: The value field of the object
+     * messageType: Refers to the parent: if this is an object from a dataset the messageType would be 'dataset'
+     *              This needs to match the Bundle.properties for mapping the error messages when an exception occurs
+     *
+     *  Rules for typeClass:
+     *      The contents of value depend on the field attributes
+     *      if single/primitive, value is a String
+     *      if multiple, value is a JsonArray
+     *         multiple/primitive: each JsonArray element will contain String
+     *         multiple/compound: each JsonArray element will contain Set of FieldDTOs
+     */
+    private static void validateTypeClass(Schema schema, String key, Map valueMap, Object value, String messageType) {
+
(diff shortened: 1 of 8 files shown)


src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
```diff
@@ -0,0 +1,190 @@
...
+                    requiredFields.remove(childTypeName);
+                }
+            }
+        });
+        if (!requiredFields.isEmpty()) {
+            throwValidationException(messageType, "required.missing", List.of(typeName, requiredFields.stream().collect(Collectors.joining(", ")), typeName));
+        }
+    }
+    private static void throwValidationException(String key, List<String> argList) {
+        throw new ValidationException(BundleUtil.getStringFromBundle("schema.validation.exception." + key, argList));
```
sekmiller, line 181: When I am trying to figure out why an error message is appearing I often search the code with the full bundle key implied by the error message. This kind of split, hardcoding only part of the key instead of the whole key makes that difficult.

doc/release-notes/10169-JSON-schema-validation.md
```diff
@@ -0,0 +1,4 @@
+### JSON Schema for datasets
+
+Enhanced JSON schema validation with checks for required and allowed child objects, Type checking for field types including: ''primative''; ''compound''; and ''controlledVocabulary'' . More user-friendly error messages to help pinpoint the issues in the Dataset JSON.  Rules are driven off the database schema, so no manual configuration is needed. See [Retrieve a Dataset JSON Schema for a Collection](https://guides.dataverse.org/en/6.1/api/native-api.html#retrieve-a-dataset-json-schema-for-a-collection) in the API Guide and PR #10169.
```
sekmiller, line 3: should the link here reference the 6.2 or 6.3 version of the guide?

src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
stevenwinship: The key is dynamic. It could be a generic error message regardless of the specific field in question or it could be specific to a field.
ie. schema.validation.exception.value.missing is specific to the 'value' field where schema.validation.exception.dataset.required.missing is for any required field.

doc/release-notes/10169-JSON-schema-validation.md
stevenwinship: It should be 6.3. this was a cut and paste flaw. Do we have a link that always posts to the latest? Since 6.3 endpoint doesn't exist yet.

stevenwinship: modified to 6.3 since I saw that in another md pre release file

src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
sekmiller: No, I get why you are doing it this way - it's just that sometimes when I find myself debugging a production issue I will start with an error message or UI prompt and then find that verbiage in the bundle. Then I will search the code with the entire bundle key to see where in the code the message is coming from. When you split it like that the search will come up empty and I'll have to guess how much of the key to search on. Also some keys might end the same way. We can leave it like this, I'll just have to remind myself that there are times when I'll need to search with a partial bundle key.

sekmiller (approved):
Looks good.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -3025,7 +3025,18 @@ api.errors.invalidApiToken=Invalid API token.
 api.ldninbox.citation.alert={0},<br><br> The {1} has just been notified that the {2}, <a href=''{3}''>{3}</a>, cites "<a href=''{4}/dataset.xhtml?persistentId={5}''>{6}</a>" in this repository.
 api.ldninbox.citation.subject={0}: A Dataset Citation has been reported!
 
+#Schema Validation
+schema.validation.exception.value.missing=Invalid data for key:{0} typeName:{1}. 'value' missing.
```
pdurbin, line 3029: Overall, I like how the errors look. From the IT test run:

- "Controlled vocabulary for key:fields typeName:language value:'badlang' is not valid."
- "Invalid data for key:fields typeName:title. Fields with 'multiple' set to true must be a list."
- "Invalid data for key:fields typeName:language. Found value as list but 'multiple' is set to false."
- "Compound value datasetContactName must match typeName of the object. Found datasetContactNme"
- "Invalid data for key:datasetContact typeName:datasetContactNotAllowed. Only datasetContactName, datasetContactAffiliation, datasetContactEmail allowed."


However, the double space shown in all these errors bothers me a bit (after `with the following error:`):

```
"message": "The Dataset JSON provided failed validation with the following error:  Controlled...
"message": "The Dataset JSON provided failed validation with the following error:  Invalid data...
```

Can we reduce it down to a single space?

src/main/java/edu/harvard/iq/dataverse/validation/JSONDataValidation.java
pdurbin: Yeah, I agree with this. I'm much more likely to search for the full `schema.validation.exception.value.missing` than the partial `value.missing` when hunting down a bug.

Raman used to do the same thing, used to not use the full key.

I can live with it for this PR but maybe we should write it up in our coding style page.

pdurbin (commented):
I ran the IT tests and I think this PR adds a lot of value. I'm looking forward to real users providing feedback on it.

However, while testing, I found a couple things I think we should fix before merging.

Also, I'm a bit confused. Shouldn't the JSON Schema output (which we also capture at doc/sphinx-guides/source/_static/api/dataset-schema.json for the default case) get longer and more nuanced as we better define our rules? I thought the logic would go into the sample JSON Schema file itself. Maybe I'm missing something fundamental.

And we already have validation when datasets are created. Is it flawed? (Probably.) Should we replace it with the new logic in this pull request?

I'll keep playing a bit in the few minutes I have before I go on vacation but I'm going to kick this back to "in progress" so @stevenwinship can take a look. No objections if someone merges this (even as-is) when I'm gone!

src/test/java/edu/harvard/iq/dataverse/validation/JSONDataValidationTest.java
```diff
@@ -0,0 +1,350 @@
...
+    @Mock
+    static DatasetFieldServiceBean datasetFieldServiceMock;
+    @Mock
+    static DatasetFieldType datasetFieldTypeMock;
+    static ControlledVocabularyValue cvv = new ControlledVocabularyValue();
+    static Map<String, Map<String, List<String>>> schemaChildMap = new HashMap<>();
+    static String rawSchema() {
+        return """
+         {
+            "$schema": "http://json-schema.org/draft-04/schema#",
```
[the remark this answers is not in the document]
stevenwinship: fixed to read from the existing file

JR-1991:
To me, it looks great already! I have two small points that could be beneficial for integration into external tools/libs:

The message is good, but it is currently limited to a human-readable format. Adding a JSONPath or any other path that displays the exact location would allow other libraries to do more with the validation result. Furthermore, adding different error types could help. For instance, if a type validation fails, this could be indicated. 

If I could imagine a response example it would look something like this:

*Paths are not accurate*

```json
{
  "is_valid": "yes",
  "errors": [
    {
      "location": "citation/fields/0/value",
      "error_type": "required",
      "message": "The title field is required."
    },
    {
      "location": "citation/fields/1/value",
      "error_type": "invalid",
      "message": "The description must be a string."
    }
  ]
}
```

Is it possible to derive such a format from your validator?

stevenwinship:
> To me, it looks great already! I have two small points that could be beneficial for integration into external tools/libs:
> 
> The message is good, but it is currently limited to a human-readable format. Adding a JSONPath or any other path that displays the exact location would allow other libraries to do more with the validation result. Furthermore, adding different error types could help. For instance, if a type validation fails, this could be indicated.
> 
> Is it possible to derive such a format from your validator?

Unfortunately the validator stops once it finds the first issue so displaying a list would not be possible. 
It could be possible to add more information to the error to help pinpoint the exact location of the issue found.

landreev:
The PR looks great, I meant to merge it last night. 
I didn't go out of my way trying things to break it; but all the valid json files I tried were accepted. And all the broken ones were recognized as invalid, with meaningful error messages. 
The error message for the case where required fields are missing could be prettier, tbh. - maybe list the specific fields that must be populated? - but it didn't feel serious enough to complain.
Merging.

PR #10547, opened 2024-05-08
Solr: Try Soft Commit on Indexing

qqmyers:
This is a quick test to use softindexing/commit within functionality to see if it improves Solr performance. If it works, we might want COMMIT_WITHIN to be configurable

Note that the PR contains code updates to use commitWithin and avoid explicit hard commits, and a solrconfig.xml update to use autoSoftCommit and increate the autoHardCommit time. To test both, you have to install DV and swap the solrconfig.xml file in and restart Solr. Nominally they could be tried separately.

Closes #

**Suggestions on how to test this**: There are two parts to this PR - one is a change to turn on autoSoftCommit and increase the length of the hard commit time, both in solrConfig.xml and the other is changes in the code to use add/delete with a COMMIT_WITHIN parameter and to not have a separate commit() statement for indexing. These could be tested separately. Nominally the only affect should be on performance, so the main testing should be to verify that each part independently speeds things up. The solrConfig.xml change can be deployed to an existing machine and tested with a solr restart. The code changes have to be deployed via the war.

There is one ~unrelated test change in this PR - for the CacheFactoryBeanTest. While testing, I observed a failure where the testAuthenticatedUserGettingRateLimited() failed at line 169 when the final count was 122 instead of 120. I'm guessing this was due to some other test running in parallel and contributing to the count. I added an @ResourceLock(value = "cache") on the test to stop this. I haven't seen a repeat since but it is clearly rare overall, so I'm not sure how to test for QA.

**Additional documentation**:


qqmyers:
One issue with this is tests - I see several failing - guessing that this might be because the changes add a small delay. That might be acceptable in practice, but might require adding delays in the tests.  (Could also be a real problem/bug, not sure yet.)

landreev:
If I were to experiment with/benchmark these code changes, should I be using this branch, or solr-soft-commit2?

qqmyers:
This branch has just the autoSoftCommit related changes. The other branch currently has that and changes to avoid deleting file docs unless needed. I'm pretty sure the changes here are all good. I haven't done much testing of the delete changes yet.

qqmyers:
Though this branch doesn't have the adds to timing in the tests to make them work yet, so OK to benchmark, but not ready to merge.

landreev:
Could you please sync the branch with develop - thanks.

pdurbin:
@qqmyers is going to take a look at failing tests (thanks).

landreev:
Stupid question - I may not fully understand how this works - but, say, I'm publishing a dataset. With this softAutoCommit set, is the dataset really not guaranteed to be searchable for 15 more seconds after the indexed timestamp has been set?

qqmyers:
Yes, but it's now set a 1 second with hardCommit at 30. So searchable within 1 second and written within 30.

jp-tosca:
Hi @qqmyers 👋🏼!

I did some more testing and added a wait of 5 seconds before: 

https://github.com/IQSS/dataverse/blob/1f9a682e95e5dcbe01a138e02ca7bc09ef57d3a7/src/test/java/edu/harvard/iq/dataverse/api/SearchIT.java#L917

worked for me, no matter how long a wait is before or after:

https://github.com/IQSS/dataverse/blob/1f9a682e95e5dcbe01a138e02ca7bc09ef57d3a7/src/test/java/edu/harvard/iq/dataverse/api/SearchIT.java#L918

Produce the same error.

I can dig a bit more into what is happening in the test but adding a wait here worked for me but needs to be done before/with: `searchPublishedSubtreeSuper`

jp-tosca:
One change that would make this test pass would be just swapping these 2 lines 117-118 since the sleep is done on 118 but from what I got yesterday from the comments is that the super user shouldn't have to wait?

qqmyers:
I think this is OK to go forward. For some reason, it appears that testDeepLinks test is significantly slower with this PR although [the overall effect appears to be more than a factor of 2 in speed improvement when reindexing ~225 datasets/4400 files.](https://iqss.slack.com/archives/C010LA04BCG/p1715188295908219). [@jp-tosca reports ](https://github.com/IQSS/dataverse/pull/10547#issuecomment-2136236975) some sensitivity to the order of two searches that exists in the current develop branch which might be part of the increase (since I've swapped the order to do the search that requires all indexing to complete before the one that would succeed without everything being complete.) (Has work been done to avoid repeat indexing of collections like @ErykKul did for datasets?).

In any case, tests are now passing (except for the confirmEmail one fixed in a separate PR) so I think review/QA can proceed/I'll remove myself as an assignee.

jp-tosca:
👋🏼 @qqmyers @pdurbin @landreev 

I was doing some reading on the soft/hard commit and particularly I was wondering if we need to do something additional, [here](https://cwiki.apache.org/confluence/display/solr/NearRealtimeSearch) is some of the documentation: 


> A common configuration might be to 'hard' auto-commit every 1-10 minutes and 'soft' auto-commit every second. 
With this configuration, new documents will show up within about a second of being added, and ***if the power goes out, you will be certain to have a consistent index up to the last 'hard' commit***.

qqmyers:
line 293 in the solrconfig file should be doing a hard commit at a maximum of 30 seconds after a given addition.

qqmyers:
lines 266-269 in the solrconfig file turn on transaction logging which is what is required to not lose info during a crash (see comments starting at 253). At startup, the transaction log is played to update the indexes - basically doing a hard commit. Nominally, hard commits can be less frequent than 30 seconds, but the longer it is/the bigger the log, the more time required after a crash to restart.

jp-tosca:
Thanks! That makes sense. FWIW for whoever tests this, I killed my instance before the indexing and the item was indexed just as the instance went back up! 👍🏼

jp-tosca (approved):
👍🏼 I reviewed the changes and left some questions about the changes but everything seems to be fine. All tests are passing after the changes and the last update from develop.

pdurbin:
Something else to keep in mind is that the Solr data is just a copy of the data in PostgreSQL. So while it would be sad to lose some Solr transactions, we can always reindex, if necessary.

stevenwinship:
I'm not sure if this is good or not. I'm running a test multiple times against this branch and 2 others, including develop. This branch is taking 5 times longer to search for a newly created dataset. It's also taking 20 times longer to realize the dataset was deleted.

qqmyers:
The intent of the PR is not to decrease the latency of any given indexing operation - it makes changes to reduce the overall cost of indexing while keeping latency 'acceptable'. Right now, the autoSoftCommit in solrConfig.xml is set to 1 second, so the maximum latency should be ~1 second - consistent with what your test shows. If that's not acceptable, the autoSoftCommit time can be reduced. (Same for the autoHardCommit - it can be changed from 30 seconds if the time to restart (which causes the server to process the n seconds since the last hard commit) is not acceptable.)

The key test is probably the full reindex time on a significantly sized database. On a QDR test box with only thousands of files, the two parts of this PR made the full reindex more than twice as fast.

landreev:
Happy to see this PR merged!
🎉

landreev:
> The key test is probably the full reindex time on a significantly sized database. On a QDR test box with only thousands of files, the two parts of this PR made the full reindex more than twice as fast.

Just want to put it on record that re-indexing a copy of the IQSS prod. database, the result is essentially the same: the speedup is just a bit under 2X.

dsmiley:
If you can differentiate a full "reindexing" (aka "zero day" scenario, aka bulk updates) from regular _incremental_ updates, then you could eschew the auto soft commits in the solrconfig.xml and instead have only the incremental update requests add commitWithin since only those should be expected to be found within a second.

PR #10565, opened 2024-05-15
Banner IT improvements after resolving issues on rocky.

### LINKED ISSUE #10312: AdminIT.testBannerMessages failing on Rocky 9
This test has failed consistently enough on Rocky 9 that I feel it's worth reporting:

`"[ERROR]   AdminIT.testBannerMessages:863 expected: <OK> but was: <ERROR>"`

Unfortunately I don't find anything more helpful in the stack of server.logs generated during end-to-end tests.

Related to #10271

jp-tosca:
After debugging this test and the endpoint to resolve the issues on #10312 I made a few changes:

1. This test was relying heavily on having any of the specified locales in the JSON but with no reason. This dependency was cause due the inability of the API to return the ID of the created banner and the need to search the id by the display message, see this [comment]( https://github.com/IQSS/dataverse/issues/10312#issuecomment-2118306934). To avoid this, the endpoint `api/admin/bannerMessage` has been extended so the ID is returned when created, this will allow to delete a specific banner with ID after creation.  This is not a breaking change since is not removing anything from the JSON.
2. The method [UtilIT.getBannerMessageIdFromResponse](https://github.com/IQSS/dataverse/blob/da3dd95184846da6bd04fe37526c3785ab6e8ddc/src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java#L3376C19-L3376C49) is not necessary anymore, it also was hardcoded a return to 0 if the string was not found but the id could be some other number.

Closes #10312

Yes.


src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -2398,10 +2403,19 @@ public Response deactivateBannerMessage(@PathParam("id") Long id) throws Wrapped
...
     public Response getBannerMessages(@PathParam("id") Long id) throws WrappedResponse {
 
-        return ok(bannerMessageService.findAllBannerMessages().stream()
-                .map(m -> jsonObjectBuilder().add("id", m.getId()).add("displayValue", m.getDisplayValue()))
-                .collect(toJsonArray()));
+        List<BannerMessage> messagesList = bannerMessageService.findAllBannerMessages();
+
+        for (BannerMessage message : messagesList) {
+            if ("".equals(message.getDisplayValue())) {
+               return error(Response.Status.INTERNAL_SERVER_ERROR, "No banner messages found for this locale.");
```
pdurbin, line 2410: Should we return the locale?

Also, should we add something to the guides to explain how to prevent this error?

pdurbin (commented):
I took a quick look.

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
jp-tosca: I added to the docs a warning: 

image

I also added a makefile that would make it easier to build the guides by just simply typing "make guides" let me know what you think.

makefile
```diff
@@ -0,0 +1,5 @@
+all::
+	PWD := $(shell pwd)
+	
+guides:
```
pdurbin, line 4:
```suggestion
html-docker:
```

Maybe this should have HTML in the name since it only builds that? Should we also have a way to build the PDF in Docker in the future?

jp-tosca: > Maybe this should have HTML in the name since it only builds that? Should we also have a way to build the PDF in Docker in the future?

How about `docs-html` `docs-pdf` `docs-epub` `docs-all`

makefile
```diff
@@ -0,0 +1,5 @@
+all::
+	PWD := $(shell pwd)
```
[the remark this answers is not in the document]
jp-tosca: Fixed

makefile
```diff
@@ -0,0 +1,12 @@
+
+docs-html:
+	docker run -it --rm -v $(shell pwd):/docs sphinxdoc/sphinx:7.3.6 bash -c "cd doc/sphinx-guides && pip3 install -r requirements.txt && make clean && make html"
```
pdurbin, line 3: Can we factor out 7.3.6 to a variable? Or better yet, get it from `doc/sphinx-guides/requirements.txt`? It should actually be 7.2.6.

pdurbin (commented):
I left another comment on the Makefile.

makefile
jp-tosca: Done @pdurbin 😄

jp-tosca:
I had to change the make `docs-all` and `docs-pdf` since I didn't noticed these requires `sphinx-latexpdf` over `sphinx`. They work fine now.

sekmiller (approved):

PR #10566, opened 2024-05-16
Delete Dataset fix

qqmyers:
**What this PR does / why we need it**: With two or more files in a draft dataset, the Delete Dataset option in the dataset page menu fails. (The user is directed to the collection page, but the draft dataset still exists.

Closes #

**Special notes for your reviewer**: There is a [long discussion in the core team's slack](https://iqss.slack.com/archives/C010LA04BCG/p1715794658270009). What changed to start causing this/why explicitly deleting the filemetadata list from the version is now a fix is not completely clear. (~makes sense why you'd need to get rid of the fm references in the version, but why wasn't that needed before?)

**Suggestions on how to test this**: Add a dataset with two or more files. Verify that delete works as expected. Could/should also test the api/datasets/id/destroy call (same command) and that harvesting still works (harvesting calls destroy dataset if a dataset with that DOI already exists).

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DestroyDatasetCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DestroyDatasetCommand.java
@@ -2,6 +2,7 @@ package edu.harvard.iq.dataverse.engine.command.impl;
 
 import edu.harvard.iq.dataverse.DataFile;
 import edu.harvard.iq.dataverse.Dataset;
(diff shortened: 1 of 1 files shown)


landreev (approved):
Looks great. 
Somewhat unsettling, the fact that it's not clear at all why it wasn't failing before. 
A dumb guess - could it be that the rate limit check added in the command engine has added just enough delay to the command execution to trigger this? 
That would mean that the issue is present in stock 6.2 as well, but we just lucked out and didn't encounter it, due to somewhat intermittent nature of the bug.

PR #10567, opened 2024-05-17
Reconcile the PID of a Dataset (If Multiple PID Providers Are Enabled)

### LINKED ISSUE #10501: Reconcile the PID of a Dataset (If Multiple PID Providers Are Enabled)
There is a superuser API/UI to move a dataset from one dataverse to another. Suppose both dataverses have different PID providers configured. Which PID provider do we expect to mint the identifier? IMHO it should be the one configured in the target dataverse, i.e. the one it was moved to. In the current implementation, PID properties are not changed during a move, so the old PID configuration remains active. This makes sense for PIDs that have already been published, even if it is counterintuitive for local administration and configuration! The situation is different for unpublished (i.e. draft) records, which are not yet visible outside the instance and can therefore obey the configuration of their dataverse/collection. Therefore, the PID properties of a dataset must be changed during the move process.

**What kind of user is the feature intended for?** Superuser
**What inspired the request?**  Managing a dataverse with two collections, one with a PermaLink provider and one with a DoiProvider. If the user has accidentally created a dataset in the wrong dataverse, we cannot change the PID configuration using the move operation.

 The PID properties of a dataset must be changed during the move process and a notification must be sent to the user about the change. 

Adaption of [MoveDatasetCommand](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java) for draft datasets to invoke [delete identifier](https://github.com/IQSS/dataverse/blob/d9a79228ef6776aff155f8d8a03349eb4f06751d/src/main/java/edu/harvard/iq/dataverse/pidproviders/PidProvider.java#L45) for the configured PID and later [register a new PID](https://github.com/IQSS/dataverse/blob/d9a79228ef6776aff155f8d8a03349eb4f06751d/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java#L154) with new PID provider. Further, addition of a user notification about the move action

Sample text for the move UI:

> "This function can be used to transfer a dataset from one dataverse to another. The PID settings of the target dataverse become active when an unpublished (i.e. draft) dataset is moved to another dataverse. This invalidates the existing PID and creates a new one. If the PID has already been in use outside the system, it will have to be adjusted. The PID configuration is not adjusted for dataset that have already been published."

Sample text for the user notification for moved draft datasets:

> "The record {Dataset title} was moved from {previous host datverse name} to {new host dataverse name}. The intended PID ({PID prior move}) was changed to {new PID after move}.  If the intended PID was already in use outside the system, it must be adapted."

Sample text for the user notification for moved published datasets:
> "The record {Dataset title} was moved from {previous host datverse name} to {new host dataverse name}."

johannes-darms:

Please refer to this issue:

- #10501

- Closes #10501

**Special notes for your reviewer/ Suggestions on how to test this**:**:
Configure two PID Provider via docker-compose-file
Alter this line:
```
  -Ddataverse.pid.providers=fake,perma
```
add those lines to the dataverse env:
```
      
        -Ddataverse.pid.perma.type=perma
        -Ddataverse.pid.perma.label=PermaProvider
        -Ddataverse.pid.perma.permalink.base-url=https://example.org/
        -Ddataverse.pid.perma.permalink.separator=-
        -Ddataverse.pid.perma.authority=identifier
```
Start up dataverse.
Create a new dataset with a datafile.
Then change the pidProvider of the dataset.
```
curl -X PUT --location "http://localhost:8080/api/v1/datasets/{id}/pidReconcile/perma" \
    -H "X-Dataverse-key: $KEY" -d 'perma'
```
Then reconcile the pid via API call:
```
curl -X PUT --location "http://localhost:8080/api/v1/datasets/{id}/pidReconcile" \
    -H "X-Dataverse-key: $KEY"
```
Verify that the PID changes in the UI and a notification apprears.

No.

Included.

Preview docs at https://dataverse-guide--10567.org.readthedocs.build/en/10567/api/native-api.html#reconcile-the-pid-of-a-dataset-if-multiple-pid-providers-are-enabled


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -151,9 +144,9 @@ protected void validateOrDie(DatasetVersion dsv, Boolean lenient) throws Command
      * @param ctxt
      * @throws CommandException
      */
-    protected void registerExternalIdentifier(Dataset theDataset, CommandContext ctxt, boolean retry) throws CommandException {
+    protected void registerExternalIdentifier(DvObject theDataset, CommandContext ctxt, boolean retry) throws CommandException {
         if (!theDataset.isIdentifierRegistered()) {
-            PidProvider pidProvider = PidUtil.getPidProvider(theDataset.getGlobalId().getProviderId());
+            PidProvider pidProvider = theDataset.getEffectivePidGenerator();
```
johannes-darms, line 149: @qqmyers I've updated this line to make it easier to test. I'm not sure if the use of the utilities is intentional and I should revert this line...

src/main/java/edu/harvard/iq/dataverse/DvObject.java
```diff
@@ -493,6 +494,7 @@ public void setStorageQuota(StorageQuota storageQuota) {
      */
     public abstract boolean isAncestorOf( DvObject other );
     
+    public abstract PidProvider getEffectivePidGenerator();
```
johannes-darms, line 497: @qqmyers I've pulled this method up to the first common ancestor of the dataset and datafile. The only reason for this is that I wanted to reuse the `registerExternalIdentifier' method without much change.

src/main/java/edu/harvard/iq/dataverse/DataFile.java
```diff
@@ -1111,4 +1112,8 @@ private boolean tagExists(String tagLabel) {
         }
         return false;
     }
+    @Override
+    public PidProvider getEffectivePidGenerator() {
+        return getOwner().getEffectivePidGenerator();
```
johannes-darms, line 1117: I hope this transitive implementation of the property is in line with planned developments for more advanced filePIDs.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4613,7 +4613,42 @@ public Response getCanDownloadAtLeastOneFile(@Context ContainerRequestContext cr
             return ok(permissionService.canDownloadAtLeastOneFile(req, datasetVersion));
         }, getRequestUser(crc));
     }
-    
+
+    @PUT
+    @AuthRequired
+    @Path("{identifier}/pidReconcile/{pididentifier}")
+    public Response reconcilePid(@Context ContainerRequestContext crc, @PathParam("identifier") String datasetId,
+                                 @PathParam("pididentifier") String generatorId) throws WrappedResponse {
```
qqmyers, line 4621: I'd suggest not having this param and always using the effectivePidGenerator. That can be set at the dataset level, so you could pick any PidProvider by setting it for the dataset first, but the more usual situation would be that the effectivePidGenerator is the one set for the collection, so you'd automatically get the provider for the new collection after a move.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+            getDataset().setGlobalIdCreateTime(null);
+            getDataset().setIdentifierRegistered(false);
+        } catch (Exception e) {
+            logger.log(Level.WARNING, "Identifier deletion was not successful:", e.getMessage());
+        }
+
+        if(ctxt.systemConfig().isFilePIDsEnabledForCollection(getDataset().getOwner())) {
+            reconcileFilePids(ctxt, currentPidProvider);
+        }
+        getDataset().setPidGenerator(this.newPidProvider);
```
qqmyers, line 87: This is setting the effective Pid Generator, which you probably don't want (or should do via the existing API call if you do).

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+
+        GlobalId oldId = getDataset().getGlobalId();
+        if (oldId == null) {
+            throw new IllegalStateException("Dataset without a global identifier, cannot alter!");
+        }
+        logger.fine("Reconciling dataset( id =`" + getDataset().getId() + ")` - removing globalId `" + getDataset().getGlobalId() + '`');
+        // remove dataset PID
+        try {
+            if (currentPidProvider.alreadyRegistered(getDataset())) { //if not registered with PIDProvider than there is no need to delete it...
+                currentPidProvider.deleteIdentifier(getDataset()); // delete it externally
```
qqmyers, line 75: Perhaps it would be better to not delete this, so it remains reserved and can't be used again for new datasets? (Doesn't help for PermaLinks, but there (or for all providers) you could extend the local PID lookup to also check the alternate id field.)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+        if(getDataset().getAlternativePersistentIndentifiers()!=null) {
+            getDataset().getAlternativePersistentIndentifiers().add(api);
+        }else{
+            getDataset().setAlternativePersistentIndentifiers(Set.of(api));
+        }
+        // We keep the old persistent identifier as AlternativePersistentIdentifier with storageLocationDesignator true.
+        // This keep the link the object store intact, without altering the files.
+        // IMHO: This command should also update the storage. First, maintenance of the storage becomes a mess if there is a counterintuitive layout.
+        // Second, it could occur that another object is minted with the old identifier and consequently we have a conflict in our storage system.
+        // We accept this risk for now since a super user can update the storage manually from old file path to the new one, and remove the AlternativeIdentifer form the database.
```
qqmyers, line 110: FWIW: If changes above don't remove this risk (by making sure you still find the alternateId when checking to see if a newly generated one already exists), then a manual work-around would be to add the alternateID to the prohibited list for the PID provider.

src/main/java/edu/harvard/iq/dataverse/DvObject.java
qqmyers: See other comments - I don't think files should have an effectivePidGenerator - you can always find the right one by looking up which provider can manage the file's globalId.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcilePIDCommandTest.java
```diff
@@ -0,0 +1,354 @@
...
+        when(pidProvider.getProtocol()).thenReturn("PROTO");
+        when(pidProvider.getAuthority()).thenReturn("PROTO");
+        when(pidProvider.getShoulder()).thenReturn("PROTO");
+        when(pidProvider.getSeparator()).thenReturn("PROTO");
+        when(pidProvider.registerWhenPublished()).thenReturn(true);
+
+        // given a dataset linked with the PIDprovider
+        Dataset ds = MocksFactory.makeDataset();
+        ds.setHarvestedFrom(mock(HarvestingClient.class));
+        ds.setPidGenerator(pidProvider);
```
qqmyers, line 343: This shouldn't be needed at all for the test.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcilePIDCommandTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+        // given a linked dataset with a givendataverse and
+        Dataverse dv = MocksFactory.makeDataverse();
+        dv.setPidGenerator(pidProvider);
```
qqmyers, line 347: This should only be needed if you make the change suggested earlier to drop the pidProvider param for the API call/command and instead pick up the dataset's effective Pid Provider, which, if it isn't set directly on the dataset will be the PidGenerator set on it's parent collection.

qqmyers:
@johannes-darms - I made several related comments about avoiding mixing the effective Pid generator and PidProvider for the existing PID concepts. If that's not clear, let's have a call or slack, etc.

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcilePIDCommandTest.java
```diff
@@ -0,0 +1,354 @@
...
+        PidProvider pidProvider = mock(PidProvider.class);
+        when(pidProvider.getProtocol()).thenReturn("PROTO");
+        when(pidProvider.getAuthority()).thenReturn("PROTO");
+        when(pidProvider.getShoulder()).thenReturn("PROTO");
+        when(pidProvider.getSeparator()).thenReturn("PROTO");
+        when(pidProvider.registerWhenPublished()).thenReturn(false);
+
+
+
+        // given a dataset linked with the PIDprovider
```
qqmyers, line 108: FWIW - the PidUtilTest has code to setup/register PidProviders - it might not be too hard to make those callable here as well. With the change to not use effectivePidGenerator in the registerExternalIdentifier call, you'll need to at least call PidUtil.addToProviderList and have a parseable PID to get a working test.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+        api.setDvObject(getDataset());
+        api.setStorageLocationDesignator(true);// cf. Dataset#getIdentifierForFileStorage()
+        if(getDataset().getAlternativePersistentIndentifiers()!=null) {
+            getDataset().getAlternativePersistentIndentifiers().add(api);
+        }else{
+            getDataset().setAlternativePersistentIndentifiers(Set.of(api));
+        }
+        // We keep the old persistent identifier as AlternativePersistentIdentifier with storageLocationDesignator true.
+        // This keep the link the object store intact, without altering the files.
+        // IMHO: This command should also update the storage. First, maintenance of the storage becomes a mess if there is a counterintuitive layout.
```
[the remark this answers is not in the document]
johannes-darms: I agree with you and therefore leave the files in place.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: I believe I need to update this. Otherwise the dataset is an inconsistent state and new datafiles will not use the new PidGenerator.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
johannes-darms: >there are two concepts: the PidProvider associated with the object based on it's globalId, and the PidProvider that will be used if/when this DvObjectContainer is used in creating child objects (a dataset in a a collection or files in a dataset)
Thanks for the explanation, I wasn't aware of those two concepts.
>In non-move cases, they would usually be the same.
They should also be in sync after the reconciliation. 
>Whenever you create a new one, you'd use the effectiveGenerator.
Hence, I need to update the effectiveGenerator so new child objects (datafiles) us it, right? 
This is basically the response the other comment related to setting the effectiveGenerator below.

src/main/java/edu/harvard/iq/dataverse/DvObject.java
johannes-darms: Thanks, I'll revert those change and adapt my changed to obtain the PIDProvider via GlobalID.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
johannes-darms: That's a good idea!

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: No you are right. This means the effective generator is obtained via the transitive closure

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: While this would resolve some of our problems we end up with a bunch of registered PIDs that will never be used. I'd rather accept the risk (I'm not sure if many are using this feature) and document that an administrator must complete the reconciliation offline, i.e. move the files and delete the alternativeIdentifier from the database. 

> you could extend the local PID lookup to also check the alternate id field.
Sure do, I believe this is needed anyhow.

johannes-darms:
> I made several related comments about avoiding mixing the effective Pid generator and PidProvider for the existing PID concepts. If that's not clear, let's have a call or slack, etc.

Thanks for the explanation I wasn't aware of the distinction. I'll update the code accordingly.

johannes-darms:
@qqmyers : I've adapted the PR according to your feedback. Could you have another look?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractDatasetCommand.java
```diff
@@ -153,7 +156,7 @@ protected void validateOrDie(DatasetVersion dsv, Boolean lenient) throws Command
      * @param ctxt
      * @throws CommandException
      */
-    protected void registerExternalIdentifier(Dataset theDataset, CommandContext ctxt, boolean retry) throws CommandException {
+    protected void registerExternalIdentifier(DvObject theDataset, CommandContext ctxt, boolean retry) throws CommandException {
```
qqmyers, line 159: FOr this change, the theDataset parameter name should probably change as well. That said, files don't currently (before #7334) get PIDs prior to publication, so file related code could be dropped for now.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    public Dataset execute(CommandContext ctxt) throws CommandException {
+        // ensure that only superuser can execute the command.
+        if (!(getUser() instanceof AuthenticatedUser) || !getUser().isSuperuser()) {
+            throw new PermissionException(BundleUtil.getStringFromBundle("admin.api.auth.mustBeSuperUser"), this,
+                    Collections.singleton(Permission.EditDataset), getDataset());
+        }
+        // Datast must be unreleased! This means there is only one version!
+        if (getDataset().isReleased()) { //@TODO: Clarify whether this is the best check...
```
qqmyers, line 54: seems OK - covers deaccessioned as well

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
qqmyers: Still needs to be documented

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+            registerExternalIdentifier(getDataset(), ctxt, true); // this updates GlobalIdCreateTime and IdentifierRegistered
+        }
+        // keep old Id als alternative identifier
+        AlternativePersistentIdentifier api;
+        api = new AlternativePersistentIdentifier();
+        api.setProtocol(oldId.getProtocol());
+        api.setAuthority(oldId.getAuthority());
+        api.setIdentifier(oldId.getIdentifier());
+        api.setDvObject(getDataset());
+        api.setStorageLocationDesignator(true);// cf. Dataset#getIdentifierForFileStorage()
```
qqmyers, line 101: rare case - if there's already an alt PID that is the storageLocationDesignator

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        return getDataset();
+    }
+
+    private void reconcileFilePids(CommandContext ctxt) {
+        // remove datafile PIDs
+            try {
+                for (DataFile df : getDataset().getFiles()) {
+                    GlobalId oldPid=df.getGlobalId();
```
qqmyers, line 133: Files may not have PIDs and, pre-#7334, shouldn't have PIDs before publication. Minimally need to have a null check if you want to keep the code ready for when files get registered PIDs at upload.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+                    df.setGlobalIdCreateTime(null);
+                    df.setIdentifierRegistered(false);
+
+                    AlternativePersistentIdentifier api;
+                    api = new AlternativePersistentIdentifier();
+                    api.setProtocol(oldPid.getProtocol());
+                    api.setAuthority(oldPid.getAuthority());
+                    api.setIdentifier(oldPid.getIdentifier());
+                    api.setDvObject(df);
+                    api.setStorageLocationDesignator(true); // cf. Dataset#getIdentifierForFileStorage()
```
qqmyers, line 149: same rare issue - if an altPID already exists that is the storage location designator

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
```diff
@@ -0,0 +1,181 @@
...
+            } catch (Exception e) {
+                logger.log(Level.WARNING, "Reconcilation of the datafile persistent identifier was not successful:", e.getMessage());
+            }
+    }
+
+    @Override
+    public boolean onSuccess(CommandContext ctxt, Object r) {
+        //update search index with the state
+        ctxt.index().asyncIndexDataset(getDataset(), true);
+        //invalidate all existing Export caches
```
qqmyers, line 170: I'd suggest removing - we have this code elsewhere for if/when we need it.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -292,6 +293,8 @@ notification.typeDescription.WORKFLOW_FAILURE=External workflow run has failed
 notification.typeDescription.STATUSUPDATED=Status of dataset has been updated
 notification.typeDescription.DATASETCREATED=Dataset was created by user
 notification.typeDescription.DATASETMENTIONED=Dataset was referenced in remote system
+notification.typeDescription.PIDRECONCILED=The Persistent identifier of dataset has been updated
+
```
qqmyers, line 297: is a notification.email.pidreconciled needed as well (somewhere around line 800)? I think sendNotification does try to send email unless this notification is muted.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
qqmyers: How about adding a second call to a new AlternativePersistentIdentifier.findByProtocolIdentifierAuthority query in the DVObjectServiceBean.isGlobalIdLocallyUnique method (only if the first query would return true)? (Query could be modelled on DvObject.findIdByAlternativeGlobalId query but avoiding the join since we don't care which dvObject has that alt PID). That would assure that the old id is never reused. (Seems like a generally good idea even though your code may be the first automated way to end up with an alt PID from an active provider. It's possible that someone migrating in data from elsewhere could be reusing

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2956,6 +2956,26 @@ The API can also be used to reset the dataset to use the default/inherited value
 
 The default will always be the same provider as for the dataset PID if that provider can generate new PIDs, and will be the PID Provider set for the collection or the global default otherwise.
 
+Reconcile the PID ofa a dataset (If multiple PIDProvider are Enabled)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Dataverse can be configured to use multiple PID Providers (see the :ref:`pids-configuration` section for more information).
+It is possible to update the used PIDProvider of a dataset. This API not not modifies the PIDProvider, but
+reconciles the PID with the current configured provider if it has not already been published.
+This means that the identifier of a previous configured PIDProvider is removed and a new one is created using the current PIDProvider.
+Note that this change does not affect the storage repository where the old identifier is still used.
```
qqmyers, line 2965: (An administrator could move the files manually and set the storagelocationdesignator to false on the entry in the alternativepersistentidentifier table for the old identifier if this is a concern. This is not necessary for Dataverse to function correctly.)

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2956,6 +2956,26 @@ The API can also be used to reset the dataset to use the default/inherited value
 
 The default will always be the same provider as for the dataset PID if that provider can generate new PIDs, and will be the PID Provider set for the collection or the global default otherwise.
 
+Reconcile the PID ofa a dataset (If multiple PIDProvider are Enabled)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Dataverse can be configured to use multiple PID Providers (see the :ref:`pids-configuration` section for more information).
+It is possible to update the used PIDProvider of a dataset. This API not not modifies the PIDProvider, but
```
qqmyers, line 2962: Suggested change: It is possible to update the PIDProvider used by a dataset and assign the dataset a new PID that uses the protocol/authority/separator/shoulder of the new PIDProvider. This API 'reconciles' a draft dataset's PID by creating a new PID supported by the PidProvider and assigning the original PID as an alternativePersistentIdentifier for the dataset. The API is restricted to datasets that have not already been published. (It does not make any changes to any PidProvider.)

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -2956,6 +2956,26 @@ The API can also be used to reset the dataset to use the default/inherited value
...
+
+To reconcile the PID of the dataset:
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+  export PERSISTENT_IDENTIFIER=doi:10.5072/FK2/YD5QDG
+
+  curl -X PUT -H "X-Dataverse-key:$API_TOKEN"  "$SERVER_URL/api/datasets/$PERSISTENT_IDENTIFIER/pidReconcile/"
```
qqmyers, line 2977: I don't think this is correct - it should be ~ /api/datasets/:persistentId/pidReconcile?persistentId=$PERSISTENT_IDENTIFIER

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: I removed the comment about export caches. However, the call to reindex is left in place.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: I've added needed code and added a testcase.

src/main/java/propertyFiles/Bundle.properties
johannes-darms: Thanks for hint. I wasn't aware the system notifications and email are handled differently. I've added the needed adaptions. Shall I add a testcase for the email construction or is the attached screenshot OK?
Screenshot 2024-08-05 at 14 53 08

doc/sphinx-guides/source/api/native-api.rst
johannes-darms: This is the API call I use to test the implementation. Shall I change the structure?

src/test/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcilePIDCommandTest.java
```diff
@@ -0,0 +1,398 @@
...
+        when(pidProvider.parsePersistentId(eq("PROTO"),eq("PROTO"),eq("OLD-ID"))).thenReturn(pid);
+
+        // given a dataset linked with the PIDprovider
+        Dataset ds = MocksFactory.makeDataset();
+        ds.setGlobalId(pid);
+        ds.setGlobalIdCreateTime(Timestamp.from(Instant.now()));
+        ds.setIdentifierRegistered(true);
+        // set file
+        ds.setFiles(ds.getFiles().subList(0,1));
+        DataFile df = ds.getFiles().get(0);
```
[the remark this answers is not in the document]
johannes-darms: Done

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: If added a null check.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: Right, the current solution is the addition of the old id as an alternativeIdentifier and the integration of alternativeIdentifers in the `isGlobalIdLocallyUnique` check of the PIDproivder.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ReconcileDatasetPidCommand.java
johannes-darms: Done as suggested, alternatives identifiers are not included in the `isGlobalIdLocallyUnique` check of the PIDprovider.

johannes-darms:
@qqmyers it took a while, but I finally managed to implement all your change requests. Would you like to look over it again?

pdurbin:
@johannes-darms heads up that there are merge conflicts.

doc/sphinx-guides/source/api/native-api.rst
vera: I've updated the URL in 10f3b5d2c4ec6a55f78ecdc86664986c8d0e6b22.

qqmyers:
@vera - I'm seeing the following in the log. If it doesn't make sense, let me know and I can dig for more info:
```
[INFO] Running edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest
Feb 05, 2025 4:09:14 PM edu.harvard.iq.dataverse.DataCitation getDateFrom
WARNING: Unable to find citation date for datasetversion: null
Feb 05, 2025 4:09:14 PM edu.harvard.iq.dataverse.DataCitation getDateFrom
WARNING: Unable to find citation date for datasetversion: null
[ERROR] Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.132 s <<< FAILURE! -- in edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest
[ERROR] edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreationAllFieldsMultipleGeoLocations -- Time elapsed: 0.039 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "name" is null
	at edu.harvard.iq.dataverse.pidproviders.PidUtil.getPidProvider(PidUtil.java:229)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:107)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:84)
	at edu.harvard.iq.dataverse.pidproviders.doi.datacite.DOIDataCiteRegisterService.getMetadataFromDvObject(DOIDataCiteRegisterService.java:169)
	at edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreationAllFieldsMultipleGeoLocations(XmlMetadataTemplateTest.java:313)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)

[ERROR] edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreation -- Time elapsed: 0.002 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "name" is null
	at edu.harvard.iq.dataverse.pidproviders.PidUtil.getPidProvider(PidUtil.java:229)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:107)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:84)
	at edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreation(XmlMetadataTemplateTest.java:177)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)

[ERROR] edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreationAllFields -- Time elapsed: 0.014 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "name" is null
	at edu.harvard.iq.dataverse.pidproviders.PidUtil.getPidProvider(PidUtil.java:229)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:107)
	at edu.harvard.iq.dataverse.pidproviders.doi.XmlMetadataTemplate.generateXML(XmlMetadataTemplate.java:84)
	at edu.harvard.iq.dataverse.pidproviders.doi.datacite.DOIDataCiteRegisterService.getMetadataFromDvObject(DOIDataCiteRegisterService.java:169)
	at edu.harvard.iq.dataverse.pidproviders.doi.datacite.XmlMetadataTemplateTest.testDataCiteXMLCreationAllFields(XmlMetadataTemplateTest.java:248)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

vera:
Thanks! It seems there was some interaction between the new tests and some old tests, because the new tests didn't remove the PID providers they created. I've pushed a commit cleaning up the tests so they should now pass.

By the way, since the test steps for testing the API in the initial comment are outdated, here are some updated steps:

1. Start up Dataverse with at least two configured PID providers (e.g. by editing `docker-compose-dev.yaml` as described in the initial comment)
2. Create collection configured for one of the PID providers, e.g. Permalinks
3. Create dataset
4. Use API to reconcile PID of that dataset (expected: no effect)

   `$ curl -X PUT "http://localhost:8080/api/datasets/:persistentId/pidReconcile?persistentId=$PERSISTENT_ID" -H "X-Dataverse-key: $API_KEY"`

   `{"status":"ERROR","message":"PID Provider perma is same as configured. This Operation has no effect!"}`

5. Use API to set PID generator of that dataset (to another PID provider than the collection, e.g. fake DOIs)

   `curl -X PUT -H "X-Dataverse-key: $API_TOKEN" -H 'Content-type: application/json' -d fake "http://localhost:8080/api/datasets/:persistentId/pidGenerator?persistentId=$PERSISTENT_ID"`

6. Use API to reconcile PID of that dataset (expected: dataset receives new PID)

   `curl -X PUT "http://localhost:8080/api/datasets/:persistentId/pidReconcile?persistentId=$PERSISTENT_ID" -H "X-Dataverse-key: $API_TOKEN"`

   `{"status":"OK","data":{"message":"doi:10.5072/FK2/CR3K6U"}}`

6. Repeat steps 3-6, but publish dataset before trying to reconcile PID (expected: no effect)

   `$ curl -X PUT "http://localhost:8080/api/datasets/:persistentId/pidReconcile?persistentId=$PERSISTENT_ID" -H "X-Dataverse-key: $API_KEY"`

   `{"status":"ERROR","message":"Dataset already published, cannot alter PID Provider"}`

qqmyers:
@vera = FWIW: I think I found the cause of the most recent test fails - see #11225.

doc/sphinx-guides/source/api/native-api.rst
pdurbin: Looks right to me now. Thanks.

pdurbin:
For the record, this run yesterday was successful: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10567/12/testReport/ . All API tests passed in about 17 minutes.

pdurbin:
Ok, yes, in the latest run at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10567/13/console I'm seeing this:

```
[ERROR] Errors: 
[ERROR]   XmlMetadataTemplateTest.testDataCiteXMLCreationAllFields:248 » NullPointer Cannot invoke "String.equals(Object)" because "name" is null
[ERROR]   XmlMetadataTemplateTest.testDataCiteXMLCreationAllFieldsMultipleGeoLocations:313 » NullPointer Cannot invoke "String.equals(Object)" because "name" is null
```

So yeah, same territory as this PR:

- #11225

pdurbin (commented):
Just some quick feedback. For context, I'm responding to @johannes-darms saying that this PR is high priority (2 of 3) in the discussion on Zulip about which PRs we should consider for Dataverse 6.6: https://dataverse.zulipchat.com/#narrow/channel/375707-community/topic/Release.206.2E6.20Timeline/near/498207634

I gave this PR a size of 20 rather than 10 because I think it still needs some thinking and design work buy in from the core team. Maybe I'll move it down once I understand the PR better. I'm leaving some questions below.

Overall it seems like a nice feature! I'm curious... what's the real world use case? I assume it's something like "We give all of our datasets Permalinks by default but along the way we decide that some datasets deserve an actual DOI. We want to move these datasets to a collection where DOIs are configured. Once moved, we give them a DOI (and keep the old Permalink around as an alternative PID)." Is that close?

johannes-darms:
> Overall it seems like a nice feature! I'm curious... what's the real world use case? I assume it's something like "We give all of our datasets Permalinks by default but along the way we decide that some datasets deserve an actual DOI. We want to move these datasets to a collection where DOIs are configured. Once moved, we give them a DOI (and keep the old Permalink around as an alternative PID)." Is that close?

We offer different PID systems for our users, but asking them to choose one upfront often leads to confusion and may result in the data capture or publication process being abandoned. To address this, we have moved the selection to the final step before publication. At this stage, users can choose the appropriate PID system, after which the system moves the dataset into the corresponding collection, updates the PID, and initiates the publication request.

Initially, I considered this as a reconciliation operation that should be part of the move process to ensure dataset consistency with the collection configuration. However, after discussing it with Jim, I revised the approach. Now, the PR focuses solely on reconciliation without moving datasets.

Another use case would be test or demo instances that start with a fake DOI or permalink and later need to transition to a proper PID system. Currently, this is not possible because certain safeguards prevent the reconciliation of already published records. However, this functionality could be easily implemented if needed.

pdurbin:
@johannes-darms phew! 6.6 is out! It looks like @vera merged the latest from develop (thanks!) but some of the questions from my last review are unanswered. Can you please take a look? Thanks!

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -3381,6 +3381,29 @@ The API can also be used to reset the dataset to use the default/inherited value
...
+Reconcile The PID of a Dataset (If Multiple PID Providers Are Enabled)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Dataverse can be configured to use multiple PID Providers (see the :ref:`pids-configuration` section for more information).
+It is possible to update the PID Provider used by a dataset and assign the dataset (and any PIDs currently assigned to
+the dataset's files) a new PID that uses the protocol/authority/separator/shoulder of the new PID Provider. This API 'reconciles'
+a draft dataset's PID by creating a new PID supported by the PID Provider and assigning the original PID as an
+alternativePersistentIdentifier for the dataset. The API is restricted to datasets that have not already been published.
+(It does not make any changes to any PID Provider.) Note that this change does not affect the storage repository where the
+old identifier is still used. (An administrator could move the files manually and set the storagelocationdesignator to
```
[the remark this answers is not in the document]
pdurbin: Better, let me try to iterate on what you wrote:

"Warning: This change does not affect the storage repository, where the old PID is still used in the name of where files are stored for the dataset. If you want to remove the PID from the name used in storage, you could manually move the files offline and remove the old identifier from the database (by setting storagelocationdesignator to false for the old identifier in the alternativepersistentidentifier table). However, this step is not required for Dataverse to function correctly."

I dunno, maybe your version is better. 😅 I do like having more explanation of some sort.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -3529,6 +3529,32 @@ The API can also be used to reset the dataset to use the default/inherited value
...
+
+Dataverse supports configuration with multiple Persistent Identifier (PID) providers (refer to the :ref:`pids-configuration` section for further details).
+This API endpoint assigns new PIDs to a draft Dataset - and, if applicable, to its Datafiles (cf. :ref:`:AllowEnablingFilePIDsPerCollection <:AllowEnablingFilePIDsPerCollection>`) —
+using the currently configured PIDProvider. In cases where the active PIDProvider differs from the one initially used to mint the dataset’s original PID, this API call facilitates reconciliation.
+It ensures consistency by reassigning a PID that aligns with the current provider’s specifications. More specifically, for a draft dataset,
+a new PID is minted through the active provider, and the previously assigned PID is preserved as an alternativePersistentIdentifier.
+The same procedure applies to associated datafiles, provided that DataFile PIDs are enabled. (Note: If the currently configured PID provider is identical to the one originally used, this API call has no effect. )
+
+The API is restricted to superusers and to datasets that have not already been published.(It does not make any changes to any PID Provider.)
```
pdurbin, line 3542:
```suggestion
The API is restricted to superusers and to datasets that have not already been published. (It does not make any changes to any PID Provider.)
```

doc/release-notes/10567-feat-reconcilepid.md
```diff
@@ -0,0 +1,4 @@
+Added a new API for persistent identifier reconciliation. An unpublished dataset can be updated with a new
```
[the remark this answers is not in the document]
pdurbin: Let's stick with reconcile. "Reassign PID" might also work. 🤷

doc/release-notes/10567-feat-reconcilepid.md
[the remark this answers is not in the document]
pdurbin: The additional docs you added help a lot. Thanks.

doc/sphinx-guides/source/api/native-api.rst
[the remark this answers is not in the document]
pdurbin: The extra docs are great. Thanks.

pdurbin (approved):
Test are passing. I haven't run the code myself but it makes sense and I'm moving this to "ready for QA".

I found one typo in the doc and made a tweak to the release notes.

Also, the issue was called "[-]Feature Request/Idea: Move draft Datasets should update their PID configuration" but I renamed it and the PR to "Reconcile the PID of a Dataset (If Multiple PID Providers Are Enabled)" because the PR is closing the issue and the PR does not do what the original issue title was asking for. You have to move the dataset and then run the reconcile command.

PR #10571, opened 2024-05-17
Test fix for confirmEmail test failure

qqmyers:
**What this PR does / why we need it**: Quick draft to see if this is the problem. The server.log in Jenkins shows an error related the the dvObject being null in a permission check coming from the call modified in this PR (which was recently updated). It occurs in the log right after the noTokenCheck for confirmEmailTest. The next test appears to fail with a  500 but I see no log errors except this. I'm not sure how a searchIncludeFragment issue would get triggered from a confirmemail.xhtml call, but trying this fix to see if it helps and to log what search result might have a null entityId.

Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
@@ -1495,15 +1495,15 @@ public class SearchIncludeFragment implements java.io.Serializable {
     }
 
     public boolean isValid(SolrSearchResult result) {
-        Long id = result.getEntityId();
-        
-        DvObject obj = dvObjectService.findDvObject(id);
-        if(obj != null && obj instanceof Dataset) {
-                
-        return result.isValid(x -> permissionsWrapper.canUpdateDataset(getDataverseRequest(), datasetService.find(x.getEntityId())));
-        }
-        logger.fine("isValid called for dvObject that is null (or not a dataset), id: " + id + "This can occur if a dataset is deleted while a search is in progress");
-        return result.isValid(x -> true);
+        return result.isValid(x -> {
+            Long id = x.getEntityId();
+            DvObject obj = dvObjectService.findDvObject(id);
+            if(obj != null && obj instanceof Dataset) {
+                return permissionsWrapper.canUpdateDataset(getDataverseRequest(), (Dataset) obj);
+            }
+            logger.fine("isValid called for dvObject that is null (or not a dataset), id: " + id + "This can occur if a dataset is deleted while a search is in progress");
+            return true;
+        });
     }
     
     public enum SortOrder {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
```diff
@@ -1495,7 +1495,15 @@ private DataverseRequest getDataverseRequest() {
     }
 
     public boolean isValid(SolrSearchResult result) {
+        Long id = result.getEntityId();
+        
+        DvObject obj = dvObjectService.findDvObject(id);
+        if(obj != null && obj instanceof Dataset) {
+                
         return result.isValid(x -> permissionsWrapper.canUpdateDataset(getDataverseRequest(), datasetService.find(x.getEntityId())));
```
ErykKul, line 1503: Passing the function instead of boolean indicating if the user has an update permission is for optimization reasons, where the cheap checks are done first (e.g., the dataset is valid and the boolean is already in the search result, which should cover most common cases), before a query is executed to find the DV object. The query and the permission check happen only when the search object is a dataset with incomplete metadata, is not draft and the boolean for displaying incomplete labels for published datasets is set to true, which should be an uncommon situation. I am not sure where the nullpointer comes from, but putting the null check inside the lambda should be way more efficient, something like this:

```java
    public boolean isValid(SolrSearchResult result) {
        return result.isValid(x -> {
            Long id = x.getEntityId();
            DvObject obj = dvObjectService.findDvObject(id);
            if(obj != null && obj instanceof Dataset) {
                return permissionsWrapper.canUpdateDataset(getDataverseRequest(), (Dataset) obj);
            }
            logger.fine("isValid called for dvObject that is null (or not a dataset), id: " + id + "This can occur if a dataset is deleted while a search is in progress");
            return true;
        });
    }
```

ErykKul (changes requested):

src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java
qqmyers: change made - thanks!

ErykKul (approved):

landreev:
No, it's not obvious to me either, why search include fragment is loaded in that scenario... but the Confirm Email test has passed 5 times in a row now, so I'm just going to merge it. 
I ran the full test suite 2 more times. On one of the runs our old friend Harvesting Clients test failed again. It (obviously) has nothing to do with this PR; but it it must be on account of async. harvesting; I'm just guessing the second of sleep we gave it recently is not always enough and we should increase it to 3. 
The last ran passed.

qqmyers:
The confirmEmail.xhtml does a redirect to the main dataverse.xhtml page so it loads the list of current search results before exiting. With the tests, we're deleting datasets in other tests, so the results returned can be out of date.

PR #10481, opened 2024-04-10
8243 improve language controlled vocab

landreev:
**What this PR does / why we need it**:

Please note the Draft nature of the PR, not ready just yet, but will be shortly

**Which issue(s) this PR closes**:

Closes #8243

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #10578, opened 2024-05-22
Support the full ISO 639-3 list of languages

### LINKED ISSUE #8578: Figure out whether, or how to support the extended ISO 639-3 list of languages
After version 5.4, things have improved regarding language mapping problems.
Some codes are still not managed. In the cases encountered, frm (Medieval French) and fro (Old French).
Would it be possible to include all codes in the Dataverse source?

Try to harvest from https://repository.ortolang.fr/api/oai/?verb=ListRecords&set=producer:atilf&metadataPrefix=oai_dc
6 datasets are not harvested, 4 due to language mapping issues. 

**What happens?**
Mapping errrors documented in the harvest log : 
Exception processing getRecord(), oaiUrl=https://repository.ortolang.fr/api/oai, identifier=oai:ortolang.fr:0c2017f1-7c3b-473a-b75d-ad97b4e09bd0, edu.harvard.iq.dataverse.api.imports.ImportException, Failed to import harvested dataset: class edu.harvard.iq.dataverse.util.json.ControlledVocabularyException (Value 'fro' does not exist in type 'language')"
I'm attaching the server.log relevant extract and the harvest log.

[harvest_ortolang3_2022-04-04T15-34-00.log](https://github.com/IQSS/dataverse/files/8417423/harvest_ortolang3_2022-04-04T15-34-00.log)
[server.log](https://github.com/IQSS/dataverse/files/8417425/server.log)

5.10

- IQSS/dataverse#7638

stevenwinship:
**What this PR does / why we need it**: Some codes are still not managed. In the cases encountered, frm (Medieval French) and fro (Old French).

**Which issue(s) this PR closes**: 8578

Closes #8578

**Suggestions on how to test this**: See setup instructions in ISO639IT.java test

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: No. Only new languages in the list of languages

**Is there a release notes update needed for this change?**: Yes. to be included in this PR

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
```diff
@@ -256,6 +260,10 @@ public Response loadDatasetFields(File file) {
                         case "#controlledVocabulary":
                             header = HeaderType.CONTROLLEDVOCABULARY;
                             break;
+                        case "#controlledVocabularyExt":
```
pdurbin, line 263: Can you please document `#controlledVocabularyExt` in source/admin/metadatacustomization.rst?

doc/release-notes/8578-support-extended-iso-639-languages.md
```diff
@@ -0,0 +1,16 @@
...
+
+` citation.tsv: Dutch nld dut nl`
+
+` language_ISO_639.tsv: Dutch, Flemish dut nld nl`
+
+` Updated in DB as: Dutch, Flemish nld dut nl`
+
+```
+wget https://raw.githubusercontent.com/IQSS/dataverse/v6.3/scripts/api/data/metadatablocks/language_ISO_639.tsv
+curl http://localhost:8080/api/admin/datasetfield/load -H "Content-type: text/tab-separated-values" -X POST --upload-file language_ISO_639.tsv
```
pdurbin, line 14: Why do we have a separate language_ISO_639.tsv file instead of putting the extra languages in citation.tsv? Are we worried some installations won't want all the extra languages?

For someone reading the Installation Guide, how would they discover that extra languages are available in the language_ISO_639.tsv file? Can we have something in source/installation/config.rst?

src/test/java/edu/harvard/iq/dataverse/api/AdminIT.java
```diff
@@ -890,7 +889,41 @@ public void testDownloadTmpFile() throws IOException {
                 .body("status", equalTo("ERROR"))
                 .body("message", equalTo("Path must begin with '/tmp' but after normalization was '/etc/passwd'."));
     }
+    @Test
+    public void testLoadISO_639_controlledVocabularyExt() {
+        Response createUser = UtilIT.createRandomUser();
+        String apiToken = UtilIT.getApiTokenFromResponse(createUser);
+        byte[] updatedContent = null;
+        try {
+            updatedContent = Files.readAllBytes(Paths.get("scripts/api/data/metadatablocks/language_ISO_639.tsv"));
```
pdurbin, line 898: Once this gets loaded in my installation, is there any way to remove it?

pdurbin (commented):
Just some high-level, doc, and test questions.

doc/release-notes/8578-support-extended-iso-639-languages.md
stevenwinship: There were concerns about the size of the list. Also, this feature allows installations the ability to merge new tsv files and re-order the list as opposed to adding new values to the end of the list. The Ext tag can be used by other CVV types and not just languages.

I will update the installation guide

stevenwinship: Looks like the merging of languages isn't going to work. I see there is a citation.properties file which I believe hold the translations for the language name to the local language. This file would not be touched so the merged languages would not be found. Merged languages would always default to English representations in the list.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
stevenwinship: "#controlledVocabularyExt" removed

doc/release-notes/8578-support-extended-iso-639-languages.md
stevenwinship: No longer relevant

src/test/java/edu/harvard/iq/dataverse/api/AdminIT.java
stevenwinship: no longer relevant

doc/release-notes/8578-support-extended-iso-639-languages.md
```diff
@@ -0,0 +1,12 @@
+The Controlled Vocabulary Values list for the metadata field Language in the Citation block has been extended.
+Roughly 300 ISO 639 languages added.
```
qqmyers, line 2: Why only 300? Where did that subset come from? Are these 639-2 codes (guessing from zho which is in 639-2 and as far as I can see not in 639-3)?

stevenwinship: The list came from the Library of Congress. Yes, they are 639-2 codes. Once I merged all the codes and removed the duplicates the list increased by ~300.

landreev: The PR says that it "closes #8578". That issues is specifically for figuring out how to support ISO **639-3** list (*thousands* of languages). All the recent discussion in that issue was centered around that - whether it was feasible to just add that many to the CVV, whether the pulldown menu was still going to be usable in the UI and/or whether we wanted to consider using an external CV instead, etc. 
I can understand it if we decide to handle this via incremental improvements, and first extend the list to the full 639-2, before proceeding with the 639-3. But I'm not seeing any discussion of this approach neither in #8578 nor in this PR. And, at the very least, I don't think this PR should be closing that issue. 

Please note that there is real interest in the full ISO 639-3 support, see, for example, #10481, where a user is specifically requesting it.

stevenwinship: I can pull this feature back and load the full 639-3 languages. The file seems like a lot but if I remember correctly it's one line per language and we combine the language codes so the final list would be shorter. Let me take another look at this.
Moving back to in progress

stevenwinship:
Includes 7900+ languages with degridation to the UI.

doc/release-notes/8578-support-extended-iso-639-languages.md
```diff
@@ -0,0 +1,11 @@
+The Controlled Vocabulary Values list for the metadata field Language in the Citation block can now be extended to include roughly 7920 ISO 639-3 values.
+
+To be added to the 6.4 release instructions:
+
+Update the Citation block, to incorporate the additional controlled vocabulary for languages:
```
pdurbin, line 5: This should be in the guides somewhere as well, not just in a release note.

stevenwinship: added more

landreev:
@stevenwinship There are 3 extra .tab files in `scripts/api/data/metadatablocks/iso-639-3_Code_Tables_20240415` (`iso-639-3-macrolanguages.tab` etc.). Are these checked in for reference purposes? (or, are they checked in on purpose?)

stevenwinship:
> @stevenwinship There are 3 extra .tab files in `scripts/api/data/metadatablocks/iso-639-3_Code_Tables_20240415` (`iso-639-3-macrolanguages.tab` etc.). Are these checked in for reference purposes? (or, are they checked in on purpose?)

They were part of the origin zip file. I don't think we need them so I'll delete them

landreev (commented):
Glad to hear that the extended list of languages does not affect the performance of the UI. 
I am ok with/like the idea of not adding the entire CV to the citation block .tsv as distributed, and instead treating it as an "extension pack" that can be added optionally. 
As for the implementation, specifically the fact that extending this CV is handled in a very unique and customized way (using the Lib. of Congress tabular file format instead of our normal .tsv format; and adding a dedicated API for extending just this specific CV), I am ok with it, I think. (I can actually see the advantage in being able to use the LoC-distributed list directly whenever they update it going forward). However, I do have a reputation within this team for being fairly cavalier in my willingness to adopt non-standard solutions and hacks. So I would like to ask for a second opinion. @qqmyers What do you think?

qqmyers:
Honestly, I'm not a fan of having a new mechanism and an API call dedicated to parsing this particular language-specific file format, particularly since it relies on one citation.properties file that has to cover both (plus the language-specific properties file variants.) I also think the current code has problems with the implementation of the merge of existing and new entries (see comments). Some of those are probably fixable in code (not sure about all - see the Pular example), but wouldn't exist if we didn't try an auto-merge.

It may also be a pain when the citation block is updated. If that happens and you reload the block (say due to some non-language change to some unrelated metadata field), then reloading the block will restore just the alternates from the block and drop any from the ISO file. So unless we add the step of using the new API to update from the ISO file as well to the release notes, we'd have unintended changes.

I do like the idea of this being optional/not a change to the citation block that everyone has to adopt, but I'm not sure doing it with a separate API and merging is worth it (versus, for example, a Javascript that might allow filtering to common/complete lists for selection).

qqmyers (commented):
See inline comments.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
```diff
@@ -543,4 +536,86 @@ public static String getDataverseLangDirectory() {
...
+                    if (cvv != null) {
+                        cvv.getControlledVocabAlternates().stream().forEach(alt -> {
+                            codes.add(alt.getStrValue());
+                        });
+                    }
+
+                    // Now call parseControlledVocabulary to create/update the Controlled Vocabulary Language
+                    // values: unused, type, displayName, identifier, display order, alt codes...
+                    String displayName = cvv != null ? cvv.getStrValue() : name;
+                    String displayOrder = String.valueOf(cvv != null ? cvv.getDisplayOrder(): lineNumber + offset);
```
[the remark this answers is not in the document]
stevenwinship: It's hard to set the display order to remain alphabetical. Especially if the Dataverse installation isn't using English. There are 180 ish languages in the original tsv file. A sql script could be created to re-order them forcing the "Not applicable" last. The ISO639-3 file actually has a "zxx No linguistic content" entry at the end. It really isn't that noticable in the UI since it's filtered on the partial string you enter. I did reorder the tab file from the LoC since that too was ordered by the code and not the display string. I thought it was best to not mess with the display order since that could be set by the admins of the instalation. Maybe it would be better to bump up the offset to 1000.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
```diff
@@ -543,4 +536,86 @@ public static String getDataverseLangDirectory() {
...
+                    String name = fields[6];
+                    // get controlledvocabularyvalue by name or by codes
+                    ControlledVocabularyValue cvv = datasetFieldService.findControlledVocabularyValueByDatasetFieldTypeAndStrValue(dsv, name, true);
+                    // if not found, find using main code and alternate codes
+                    Iterator<String> codesIterator = codes.iterator();
+                    while (cvv == null && codesIterator.hasNext()) {
+                        cvv = datasetFieldService.findControlledVocabularyValueByDatasetFieldTypeAndStrValue(dsv, codesIterator.next(), true);
+                    }
+
+                    // if it is found we need to merge the alternate codes since the next step will delete all existing alternate codes before adding the new ones
```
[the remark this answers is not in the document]
stevenwinship: The merge does not remove the original codes since they may not be in the new file. This is why the original alts get saved and merged to the new list

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
```diff
@@ -543,4 +536,86 @@ public static String getDataverseLangDirectory() {
...
+    public Response mergeLanguagePropertyFile(File file) {
+        ActionLogRecord alr = new ActionLogRecord(ActionLogRecord.ActionType.Admin, "mergeLanguageList");
+        alr.setInfo( file.getName() );
+        JsonArrayBuilder responseArr = Json.createArrayBuilder();
+
+        BufferedReader br = null;
+        String line = null;
+        String splitBy = "\t";
+        int lineNumber = 0;
+        int offset = 200; // number of existing languages // TODO: get the number from db
```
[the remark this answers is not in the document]
stevenwinship: I think there are 187. This just gives wiggle room. also the display order is actually redundant if you always want to order the list alphabetically. A simple sql script can be written to re-order the list

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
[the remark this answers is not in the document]
qqmyers: OK - so this is a problem for Pular for example.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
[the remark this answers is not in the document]
qqmyers: FWIW: Although I don't think it is necessary give the current code, we've so far tried to keep these in sequence w/o gaps.  I still think the logic has a problem though - if the ISO file adds a line in the middle, you'll try to add the new entry at offset + line number and there will already be one with that displayOrder number and it won't change, so two will have the same displayOrder.

src/main/java/edu/harvard/iq/dataverse/api/DatasetFieldServiceApi.java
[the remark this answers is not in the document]
qqmyers: The ISO file itself seemed to be alphabetical. After one run of the code here, it looks like all of the current citation.tsv entries come first and then everything from the ISO file is in the order from that file. If there are changes in the ISO file though, the current code won't maintain it's internal order (either a duplicate displayOrder number as is, or the new entry will come after all existing entries if you update the code to query the db for number of entries (or max displayOrder) as in the todo.) Having the display order depend on when you run the api/if you ran it for intermediate changes in the ISO file, doesn't seem optimal. (Versus, for example, having a separate script to generate an always in sequence version of what would go in the citation.tsv file, or even just using only the line number from the ISO file as the displayOrder (or do we have citation.tsv entries that are not in the ISO file?)).

landreev:
In the light of having confirmed that having the full list in the CV does not make the UI unusable, I would at least consider adding it in full to citation.tsv (and give up on the _optional_ aspect of the expansion). My $0.02?

qqmyers:
This is being replaced by #10762, so this should close.

PR #10505, opened 2024-04-18
CVOC : Indexed field accuracy (Ontoportal integration)

luddaniel:
Following agreements on next steps of #10145

**What this PR does / why we need it**:

This PR allows admins to specify which indexed fields must be use for a [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) configured service.

**Suggestions on how to test this**:

Related documentation : https://github.com/gdcc/dataverse-external-vocab-support/pull/21 
Ontoportal CVocConf example : [cvoc-agroportal-ontoportal-conf.json](https://github.com/Recherche-Data-Gouv/dataverse-external-vocab-support/blob/poc_ontoportal/examples/config/cvoc-agroportal-ontoportal-conf.json)

**Default behaviour (with `indexIn`)**:

Screenshot from 2024-04-18 16-31-45

**With `indexIn`**:

Screenshot from 2024-04-18 16-28-38

**Does this PR introduce a user interface change?**
Indirectly, it changes search result snippet.

PR #10603, opened 2024-05-31
CVOC: Fix NPE

luddaniel:

Fix NPE in complex case of CVOC : free text + partial data using flexible params in retrievalUri 
- Original and merged PR : #10404 

Code is tested and will fix already merged code to develop.

_I'm fixing my own code here_

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
@@ -42,6 +42,7 @@ import jakarta.persistence.TypedQuery;
 
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.httpclient.HttpException;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpResponseInterceptor;
 import org.apache.http.client.methods.HttpGet;
@@ -560,6 +561,10 @@ public class DatasetFieldServiceBean implements java.io.Serializable {
     }
 
     private String replaceRetrievalUriParam(String retrievalUri, String paramName, String value) {
+
+        if(StringUtils.isBlank(paramName) || StringUtils.isBlank(value)) {
+            return retrievalUri;
+        }
 
         if(retrievalUri.contains("encodeUrl:" + paramName)) {
             retrievalUri = retrievalUri.replace("{encodeUrl:"+paramName+"}", URLEncoder.encode(value, StandardCharsets.UTF_8));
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
```diff
@@ -561,6 +562,10 @@ public void process(HttpResponse response, HttpContext context) throws HttpExcep
 
     private String replaceRetrievalUriParam(String retrievalUri, String paramName, String value) {
 
+        if(StringUtils.isBlank(paramName) || StringUtils.isBlank(value)) {
+            return retrievalUri;
+        }
+
```
qqmyers, line 568: It seems like line 501 above is the only thing that could cause the NPE? And in that case the paramName can't be null but the value could be? I wonder if checking on line 571 to see if retrievalUri.contains(paramName) would be a better check - if the param name is in the Uri and it isn't replaced, that's a problem you wouldn't want to silently ignore.

luddaniel: CVOC is sooo complexe...

- you are right the only case where there is NPE can happen is : L501 `retrievalUri = replaceRetrievalUriParam(retrievalUri, f.getDatasetFieldType().getName(), f.getValue());` when value is null which can happen with termUri filled but when a parameter set in retrievalUri has no value.
- _retrievalUri.contains(paramName) would be a better check - if the param name is in the Uri and it isn't replaced, that's a problem you wouldn't want to silently ignore_ if param is null, empty or wrong I just do nothing, either it cannot be done, data are wrong or json setting is wrong. But that being said, a "if paramName does not exist in retrievalUri then logger.warning" can be done.
- _the code shouldn't just call the remote service with a broken URL, so maybe the NPE should be caught and used to avoid the call to the service instead?_ hum, you are right, for a wrong parameter we could simply `return;` instead of performing the call.

luddaniel: @qqmyers  79b8d89 modifications have been made, if you can review / approve :)

qqmyers (approved):
Looks good - thanks for the update.

qqmyers:
@luddaniel - I ran across another NPE when trying to use the latest dev branch with a skomos and ror script. I was getting a case where the value of a datasetfieldvalue was null. Either of the two changes in https://github.com/IQSS/dataverse/commit/a520799fce6a2ecc6872421853aef2e1847e365d would fix it, but I was surprised to see that a DatasetFieldValue would return a null value (versus there not being a DatasetFieldValue at all). 

I guess two questions - any insight into why there's a null value? Assuming no ideas about a deeper cause, could you confirm that this fix (one of the two checks in the commit) makes sense and perhaps add it to your PR so it can all get tested together?

luddaniel:
@qqmyers I don't know how this null value came along with Dataverse releases or different works on CVOC. I did came across the same issue one day as I fixed this issue here : https://github.com/IQSS/dataverse/pull/10505/files#diff-171c256be8744ca0742a9570ed33a3e335eae0544d08be04545ef31fdd5328d1R485

PR #10610, opened 2024-06-05
Fix DOI/ID message when deaccession a Dataset

### LINKED ISSUE #10309: Correct success message of deaccessioning API endpoint when dataset PID is used
When using the API endpoint for deaccessioning a dataset, documented at https://guides.dataverse.org/en/6.1/api/native-api.html#deaccession-dataset, while passing in a dataset's PID instead of its database ID

**What happens?**
The success message reads:
`{'status': 'OK', 'data': {'message': 'Dataset :persistentId deaccessioned for version 1.0'}}`

All users

In place of ":persistentId", the success message would include either the PID of the dataset or its database ID.

  v6.1

  https://github.com/IQSS/dataverse/issues/778, https://github.com/IQSS/dataverse/issues/8132

jp-tosca:
It was reported on #10309 that this parameter was not substituted by the value on the success message.

Closes #10309

1. Create a Dataset version
2. Deaccession it with `http://localhost:8080/api/datasets/:persistentId/versions/:versionID/deaccession?key=API_KEY&persistentId=doi:10.5072/FK2/RGFWNO`
3. the new message should show as `"message": "Dataset 10.5072/FK2/RGFWNO deaccessioned for version 9"`
4. Check the database id still works `http://localhost:8080/api/datasets/6/versions/10/deaccession?key=API_KEY`
5. message should show as `"message": "Dataset 6 deaccessioned for version 10"`

No

**Additional documentation**:
No

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -4778,7 +4778,8 @@ public class Datasets extends AbstractApiBean {
             return badRequest(BundleUtil.getStringFromBundle("datasets.api.deaccessionDataset.invalid.version.identifier.error", List.of(DS_VERSION_LATEST_PUBLISHED)));
         }
         return response(req -> {
-            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, versionId, findDatasetOrDie(datasetId), uriInfo, headers);
+            Dataset dataset = findDatasetOrDie(datasetId);
+            DatasetVersion datasetVersion = getDatasetVersionOrDie(req, versionId, dataset, uriInfo, headers);
             try {
                 JsonObject jsonObject = JsonUtil.getJsonObject(jsonBody);
                 datasetVersion.setVersionNote(jsonObject.getString("deaccessionReason"));
@@ -4791,7 +4792,10 @@ public class Datasets extends AbstractApiBean {
                     }
                 }
                 execCommand(new DeaccessionDatasetVersionCommand(req, datasetVersion, false));
-                return ok("Dataset " + datasetId + " deaccessioned for version " + versionId);
+
+                return ok("Dataset " + 
+                        (":persistentId".equals(datasetId) ? dataset.getAuthority() + "/" + dataset.getIdentifier() : datasetId) + 
+                        " deaccessioned for version " + versionId);
             } catch (JsonParsingException jpe) {
                 return error(Response.Status.BAD_REQUEST, "Error parsing Json: " + jpe.getMessage());
             }
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -4793,7 +4793,7 @@ public class Datasets extends AbstractApiBean {
                 execCommand(new DeaccessionDatasetVersionCommand(req, datasetVersion, false));
                 
                 return ok("Dataset " + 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4791,7 +4791,10 @@ public Response deaccessionDataset(@Context ContainerRequestContext crc, @PathPa
                     }
                 }
                 execCommand(new DeaccessionDatasetVersionCommand(req, datasetVersion, false));
-                return ok("Dataset " + datasetId + " deaccessioned for version " + versionId);
+                
+                return ok("Dataset " + 
+                        (":persistentId".equals(datasetId) ? datasetVersion.getDataset().getAuthority() + "/" + datasetVersion.getDataset().getIdentifier() : datasetId) +
```
pdurbin, line 4796: Won't this be missing "doi:"? What if it's "hdl:" instead?

Can we please have a test?

pdurbin (commented):
I'm pretty sure this won't show the full PID.

jp-tosca:
> I'm pretty sure this won't show the full PID.

It won't, I removed it. Previously I was using the "protocol" would that be the right thing to use to prepend? @pdurbin

jp-tosca:
Thanks @qqmyers! I updated this PR to use ***asString*** from the GlobalID.

I also added to the tests to check when it is deleted by a specified ID to be contained on the message and a function to delete by global. @pdurbin

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -3858,6 +3858,19 @@ static Response deaccessionDataset(Integer datasetId, String version, String dea
...
+        if (deaccessionForwardURL != null) {
+            jsonObjectBuilder.add("deaccessionForwardURL", deaccessionForwardURL);
+        }
+        String jsonString = jsonObjectBuilder.build().toString();
+        return given()
+                .header(API_TOKEN_HTTP_HEADER, apiToken)
+                .body(jsonString)
+                .post("/api/datasets/:persistentId/versions/" + version + "/deaccession?persistentId=" + globalId);
+    }
+
```
pdurbin, line 3873: I guess it's ok to add a whole new method but if you want to pass a PID (I would suggest this term over GlobalID), you could use this pattern you'll see elsewhere in UtilIT:

```
String idInPath = datasetIdOrPersistentId; // Assume it's a number.
String optionalQueryParam = ""; // If idOrPersistentId is a number we'll just put it in the path.
if (!NumberUtils.isCreatable(datasetIdOrPersistentId)) {
    idInPath = ":persistentId";
    optionalQueryParam = "?persistentId=" + datasetIdOrPersistentId;
}
```

pdurbin (commented):
This is coming along. I'm adding a suggestion for how to avoid adding an entire new method to UtilIT.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
jp-tosca: I made some changes, please let me know if this is better.

pdurbin: It's better, but did you find a datasetIdOrPersistentId example in UtilIT? Do you see what I mean, how we can reduce code duplication?

jp-tosca: I will change that in a bit 😅

jp-tosca: Ok, I made some changes, let me know what you think. 😄

qqmyers (approved):
Looks good. Tests pass. Looks like all of the prior review comments have been addressed.

Issue #9401, opened 2023-02-21
When using Chrome to download guestbook file of a Dataverse collection whose name includes comma(s), file is saved without .csv extension and with truncated collection name


jggautier:
**What steps does it take to reproduce the issue?**

* When does this issue occur?
  When using the Chrome browser to download the guestbook of a Dataverse collection that has a comma in the collection's name, such as the collection named "Representation, Treatment, Outcome" at https://dataverse.harvard.edu/dataverse/representationtreatmentoutcome

* Which page(s) does it occurs on?
The Manage Dataset Guestbook page

* What happens?
  The file is downloaded and its name should be the name of the collection, but the name is truncated. The first comma and everything after that is missing, including the .csv file extension. 

  So when using Chrome to download the guestbook for the collection "Representation, Treatment, Outcome", file is named Representation.

  This might make it confusing to open the file. Users can add the .csv extension to the file name. Or they can use a different browser.

* To whom does it occur (all users, curators, superusers)?
Dataverse collection managers who use Chrome

* What did you expect to happen?
The .csv file extension is included. I think it's less important that the collection's full name is used, although that depends on where the comma is.

**Which version of Dataverse are you using?**
v5.13, v6.1, v6.2

**Any related open or closed issues to this bug report?**
Cannot download guestbook in Chrome if Dataverse has a "," in the name (https://github.com/IQSS/dataverse/issues/4512)

jggautier:
When using latest versions of Firefox, the guestbook CSV file is named as expected.

PR #10613, opened 2024-06-05
adding fix to file name for download guestbook responses

stevenwinship:
When using the Chrome browser to download the guestbook of a Dataverse collection that has a comma in the collection's name, such as the collection named "Representation, Treatment, Outcome" the file name is truncated to Representation.

Closes https://github.com/IQSS/dataverse/issues/9401

**Suggestions on how to test this**: test by creating a collection with commas in the name and setting up guestbook. Then, download using Chrome, Safari, and Firefox.

**Additional documentation**: Also removes other chars that could cause issues when used in a file name.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ManageGuestbooksPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ManageGuestbooksPage.java
@@ -220,7 +220,8 @@ public class ManageGuestbooksPage implements java.io.Serializable {
        // The fix below replaces any spaces in the name of the dataverse with underscores;
        // without it, the filename was chopped off (by the browser??), and the user 
        // was getting the file name "Foo", instead of "Foo and Bar in Social Sciences.csv". -- L.A.
-       return  dataverse.getName().replace(' ', '_') + "_GuestbookReponses.csv";
+       // Also removing some chars that have been reported to cause issues with certain browsers
+       return  dataverse.getName().replace(' ', '_').replaceAll("[\\\\/:*?\"<>|,;]", "") + "_GuestbookResponses.csv";
     }
     
     public void deleteGuestbook() {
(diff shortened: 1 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/ManageGuestbooksPage.java
```diff
@@ -220,7 +220,8 @@ private String getFileName(){
        // The fix below replaces any spaces in the name of the dataverse with underscores;
        // without it, the filename was chopped off (by the browser??), and the user 
        // was getting the file name "Foo", instead of "Foo and Bar in Social Sciences.csv". -- L.A.
-       return  dataverse.getName().replace(' ', '_') + "_GuestbookReponses.csv";
+       // Also removing some chars that have been reported to cause issues with certain browsers
+       return  dataverse.getName().replace(' ', '_').replaceAll("[\\\\/:*?\"<>|,;]", "") + "_GuestbookResponses.csv";
```
pdurbin, line 224: Since these two regex's are the same, should they be defined once as a constant?

Also, can we have some tests?

pdurbin (commented):
I took a quick look.

src/main/java/edu/harvard/iq/dataverse/ManageGuestbooksPage.java
stevenwinship: There in 2 different classes.

stevenwinship: And neither of these classes have any tests :(

stevenwinship: moving this method to FileUtil and adding tests

pdurbin: So much more readable! Thanks!

qqmyers (approved):
Looks good. Earlier comments were addressed.

PR #10620, opened 2024-06-07
Deprecate rsync: remove from guides, mark code as deprecated

### LINKED ISSUE #8985: remove rsync from guides (deprecated)
Currently, rsync and package files are featured prominently in the User Guide:

https://guides.dataverse.org/en/5.11.1/user/dataset-management.html#rsync-ssh-upload

Screen Shot 2022-09-20 at 4 45 29 PM

https://guides.dataverse.org/en/5.11.1/user/find-use-data.html#downloading-a-dataverse-file-package-via-url

Screen Shot 2022-09-20 at 4 46 19 PM

pdurbin:

Support for rsync has never been strong and no one is using it in production. At https://groups.google.com/g/dataverse-community/c/TMtVpYcqqFM/m/z0t4isZlHQAJ we asked the community if anyone minds if we deprecate it and there were no objections.

This pull request removes rsync and related content from the guides. It also marks various Java code and settings as deprecated but stops short of removing any code.

- Closes #8985

~~I went ahead and deprecated :PublicInstall which was mostly added for rsync. It does interact a bit with Swift as well but I adjusted the doc accordingly.~~ Nevermind.

Nothing to test. Just look at the diff and see if it makes sense.

No.

Yes, included.

**Additional documentation**:

You can preview the changes at https://dataverse-guide--10620.org.readthedocs.build/en/10620/


doc/release-notes/8985-deprecate-rsync.md
```diff
@@ -0,0 +1,9 @@
+Support for rsync has been deprecated. Information has been removed from the guides for rsync and related software such as Data Capture Module (DCM) and Repository Storage Abstraction Layer (RSAL). You can still find this information in [older versions](https://guides.dataverse.org/en/6.2/developers/big-data-support.html#data-capture-module-dcm) of the guides.
+
+The following related database settings have been deprecated as well:
+
+- :DataCaptureModuleUrl
+- :DownloadMethods
+- :LocalDataAccessPath
+- :PublicInstall
```
landreev, line 8: Was this `:PublicInstall` setting specific to/only used by this Rsync/DCM functionality?

src/main/java/edu/harvard/iq/dataverse/FilePage.java
```diff
@@ -1546,6 +1546,7 @@ public String getIngestMessage() {
     }
     
     //Determines whether this File uses a public store and therefore doesn't support embargoed or restricted files
+    @Deprecated(forRemoval = true, since = "2024-07-07")
```
landreev, line 1549: I'm not sure that we want to deprecate this method. Even if we deprecate the instance-wide `:PublicInstall` setting, note the code in the method defaults to `StorageIO.isPublicStore(...)`; i.e. individual stores can still be "public".

doc/release-notes/8985-deprecate-rsync.md
landreev: ... I actually thought it was added for the early "compute" experiments, not for Rsync. (hence the overlap with swift - MOC cluster used it for storage back then). 
It also looks like the concept of "public" storage has been ported to individual stores since then, where it's used for "remote" stores (I believe) at least. 
In other words, I would double-check before removing/deprecating this one.

src/main/java/edu/harvard/iq/dataverse/FilePage.java
pdurbin: Ah, good point. I got a bit overzealous here. Will revert.

doc/release-notes/8985-deprecate-rsync.md
pdurbin: Yeah, you're probably right. I probably need to but `:PublicInstall` back in the docs and revert the deprecation annotations. It's a bummer because I was hoping we could simplify the code in the future but like you said it seems to be used for stuff beyond rsync.

pdurbin: You're right. It was added for early MOC experiments in 3a573338d90aea74410dd546b791755da608c5a0

Will revert.

pdurbin: Fixed in a00d2700d7.

src/main/java/edu/harvard/iq/dataverse/FilePage.java
pdurbin: Fixed in a00d2700d7.

landreev (approved):
Looks good! 
Thanks for going to the trouble of going through the old code and deprecating things properly (as opposed to just dropping the text from the gudies).

pdurbin:
I went ahead and remove rsync from our list of features at https://dataverse.org/software-features

Here's how it looked before:

Screenshot 2024-06-12 at 4 31 09 PM

PR #10627, opened 2024-06-12
ignore shapefiles if they are under a hidden directory in the zip file

### LINKED ISSUE #8945: SPIKE: Improve how Dataverse labels shapefiles to prevent mislabelling of zip files that aren't shapefiles
A depositor uploaded a double zipped file into a dataset in the Harvard Dataverse Repository and the file has been incorrectly labelled as a "Shapefile as ZIP Archive".

The file is in the published dataset at https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/HWVUER. 

There are no shape files in the zip file and the depositor wrote that it isn't a shapefile. The depositor also wrote that they used the UI (their Chrome browser) to upload the file (and not the Dataverse API). The email conversation with the depositor is at https://help.hmdc.harvard.edu/Ticket/Display.html?id=322790.

The file needs to be correctly labelled as a "ZIP Archive". Having it labelled as a "Shapefile as ZIP Archive" might be confusing to anyone looking to download the data.

stevenwinship:
**What this PR does / why we need it**:  Zip files containing shape files under hidden directories should labelled as a "ZIP Archive". Having it labelled as a "Shapefile as ZIP Archive" might be confusing to anyone looking to download the data.

**Which issue(s) this PR closes**: SPIKE: Improve how Dataverse labels shapefiles to prevent mislabelling of zip files that aren't shapefiles #8945

Closes #8945

zip test.zip src/test/resources/hiddenShapefiles.zip
upload this file which contains shapefile data under a hidden directory. Was showing as 'Shapefile as ZIP Archive'. Now shows 'ZIP Archive'
Upload double zip file with shapefiles in visible directory and see that it shows as 'Shapefile as ZIP Archive'.


pdurbin (commented):
Looks good. I'm including one suggestion.

jggautier:
Hi @stevenwinship. After this improvement makes its way to Harvard Dataverse, would I be able to change the label of a file that was labelled as "Shapefile as ZIP Archive", like the file in the dataset at https://doi.org/10.7910/DVN/HWVUER?

Maybe with the [redetect file type API endpoint](https://guides.dataverse.org/en/5.10.1/api/native-api.html#redetect-file-type)?

stevenwinship:
> Hi @stevenwinship. After this improvement makes its way to Harvard Dataverse, would I be able to change the label of a file that was labelled as "Shapefile as ZIP Archive", like the file in the dataset at https://doi.org/10.7910/DVN/HWVUER?
> 
> Maybe with the [redetect file type API endpoint](https://guides.dataverse.org/en/5.10.1/api/native-api.html#redetect-file-type)?

Yes. I just tested the redetect endpoint and after exiting the ui and going back in the label changed to 'ZIP Archive'. Not sure why I had to exit and come back in but at least it looks correct.

qqmyers (approved):
Looks good and should fix the specific issue - as shown by the test. FWIW: I suspect we could exclude more cases, i.e. if we don't detect shape file component files at the top or one dir down, the zip isn't a shapefile, but the problem is probably rare enough that changes can probably wait until there's a reported problem.

PR #10634, opened 2024-06-17
adding sql script to convert custom license CC0 Waiver to CC0 1.0

### LINKED ISSUE #9081: Multiple license update turned the terms of some datasets with only CC0 licenses into "custom licenses"

* When does this issue occur?
  After a Dataverse installation that's running a Dataverse software version before 5.10 is updated to version 5.10 or later.

  The dataset page on the Terms tab and in the metadata exports that include license and terms metadata

* What happens?
  
  The update from pre-v5.10 to 5.10+ adjusts the terms metadata of datasets in the installation. The datasets that have a CC0 waiver plus text in any other field in the old "Terms of Use" accordion are considered to have "custom licenses".

  But some datasets with CC0 waivers and no text in fields in the old "Terms of Use" accordions also appear, after the update, to have "custom licenses", such as https://doi.org/10.7910/DVN/00ROYZ and https://doi.org/10.15787/h6vv-ts37. The only text in the custom license is "CC0 waiver":

  Screen Shot 2022-10-18 at 1 53 03 PM

Depositors, curators, anyone responsible for or looking at datasets whose terms metadata has changed after the Dataverse installation's software version was updated 

The dataset would have a CC0 waiver instead of a "custom license"

  v5.10 or later

stevenwinship:
**What this PR does / why we need it**: The update from pre-v5.10 to 5.10+ adjusts the terms metadata of datasets in the installation. The datasets that have a CC0 waiver plus text in any other field in the old "Terms of Use" accordion are considered to have "custom licenses". But some datasets with CC0 waivers and no text in fields in the old "Terms of Use" accordions also appear, after the update, to have "custom licenses", such as https://doi.org/10.7910/DVN/00ROYZ and https://doi.org/10.15787/h6vv-ts37. The only text in the custom license is "CC0 waiver":

**Which issue(s) this PR closes**:Multiple license update turned the terms of some datasets with only CC0 licenses into "custom licenses" #9081

Closes #9081 

**Suggestions on how to test this**: SQL Script

**Is there a release notes update needed for this change?**: Yes. Included

**Additional documentation**:


src/main/resources/db/migration/V6.2.0.2.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE termsofuseandaccess SET license_id = (SELECT license.id FROM license WHERE license.name = 'CC0 1.0'), termsofuse = 'CC0 Waiver'
```
qqmyers, line 1: termsofuse should be null after the update

stevenwinship: I thought it would be nice to be able to find the ones changed after the fact. There are many rows with termsofuse not null with license_id set. I can change it if you want.

qqmyers: Hmm - in the UI if you have something in terms of use and select a license, the terms of use are nulled. I don't think there should be a way to have a license and terms of use or related fields- they terms of use aren't displayed when a license is in use. (Terms of Access and fields related to that are fine). I'd suggest putting null in with this PR and perhaps opening an issue to address any other cases that exist (legacy and if there's a way to create new datasets with a license and terms of use we should close that hole.).

stevenwinship: changed to null

qqmyers (approved):
Looks good.

PR #10642, opened 2024-06-19
Fix metadata field type display condition in dataverses/{id}/metadatablocks API endpoint

### LINKED ISSUE #10637: Fix dataverses/{id}/metadatablocks API returned dataset field types

The fields returned for each metadata block when returnDatasetTypes query parameter is true are incorrect because the display condition based on dataset field types input levels is not working as expected.

API User

- Working on SPA create collection form
- Working on SPA edit dataset form

Current dataverses/{id}/metadatablocks API endpoint returned dataset field types

- https://github.com/IQSS/dataverse-client-javascript/issues/156
- https://github.com/IQSS/dataverse-frontend/issues/406

GPortas:

dataverses/{id}/metadatablocks API endpoint has been fixed, since the fields returned for each metadata block when returnDatasetTypes query parameter is set to true was not correct.

Closes #10637 

The goal of the changes is for the endpoint to return the correct blocks and field types for both the create dataset form (onlyDisplayedOnCreate=true) and the edit dataset form (onlyDisplayedOnCreate=false).

The conditions on the input levels, which are configured when creating or editing the collection, were not working as expected.

New IT tests have been added, but for manual testing, follow these steps:

Create different collections with required or included fields (input levels) using the checkboxes in the metadata fields section.

Screenshot 2024-06-20 at 09 22 44

Then, call the dataverses/{id}/metadatablocks API endpoint using returnDatasetFieldTypes=true, for each collection you created.

First, add onlyDisplayedOnCreate=true to return the fields that should be displayed in the create dataset form. This should match what you see on JSF:

`curl -H "X-Dataverse-key:XXXXX" "http://localhost:8080/api/dataverses/{ID}/metadatablocks?returnDatasetFieldTypes=true&returnDatasetFieldTypes=true"`

Then, add onlyDisplayedOnCreate=false, to return the fields you see in the edit dataset form. Again, this should match what you see on JSF:

`curl -H "X-Dataverse-key:XXXXX" "http://localhost:8080/api/dataverses/{ID}/metadatablocks?returnDatasetFieldTypes=false&returnDatasetFieldTypes=true"`

No

Yes, attached.

**Additional documentation**:

None

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -632,7 +632,7 @@ public class JsonPrinter {
     }
 
     public static JsonObjectBuilder json(MetadataBlock metadataBlock, boolean printOnlyDisplayedOnCreateDatasetFieldTypes, Dataverse ownerDataverse) {
-        JsonObjectBuilder jsonObjectBuilder = Json.createObjectBuilder();
+        JsonObjectBuilder jsonObjectBuilder = jsonObjectBuilder();
         jsonObjectBuilder.add("id", metadataBlock.getId());
         jsonObjectBuilder.add("name", metadataBlock.getName());
         jsonObjectBuilder.add("displayName", metadataBlock.getDisplayName());
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -647,8 +647,11 @@ public class JsonPrinter {
             boolean requiredAsInputLevel = isInputLevel && ownerDataverse.isDatasetFieldTypeRequiredAsInputLevel(datasetFieldTypeId);
             boolean includedAsInputLevel = isInputLevel && ownerDataverse.isDatasetFieldTypeIncludedAsInputLevel(datasetFieldTypeId);
 
+            DatasetFieldType parentDatasetFieldType = datasetFieldType.getParentDatasetFieldType();
+            boolean isRequired = parentDatasetFieldType == null ? datasetFieldType.isRequired() : parentDatasetFieldType.isRequired();
+
             boolean displayCondition = printOnlyDisplayedOnCreateDatasetFieldTypes
-                    ? (datasetFieldType.isDisplayOnCreate() || datasetFieldType.isRequired() || requiredAsInputLevel)
+                    ? (datasetFieldType.isDisplayOnCreate() || isRequired || requiredAsInputLevel)
                     : !isInputLevel || includedAsInputLevel;
 
             if (displayCondition) {
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -643,12 +643,12 @@ public class JsonPrinter {
 
         for (DatasetFieldType datasetFieldType : datasetFieldTypes) {
             Long datasetFieldTypeId = datasetFieldType.getId();
-            boolean isInputLevel = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeInInputLevels(datasetFieldTypeId);
             boolean requiredAsInputLevel = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeRequiredAsInputLevel(datasetFieldTypeId);
             boolean includedAsInputLevel = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeIncludedAsInputLevel(datasetFieldTypeId);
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -640,18 +640,26 @@ public static JsonObjectBuilder json(MetadataBlock metadataBlock, boolean printO
...
+            boolean isInputLevel = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeInInputLevels(datasetFieldTypeId);
+            boolean requiredAsInputLevel = isInputLevel && ownerDataverse.isDatasetFieldTypeRequiredAsInputLevel(datasetFieldTypeId);
+            boolean includedAsInputLevel = isInputLevel && ownerDataverse.isDatasetFieldTypeIncludedAsInputLevel(datasetFieldTypeId);
+
+            DatasetFieldType parentDatasetFieldType = datasetFieldType.getParentDatasetFieldType();
+            boolean isRequired = parentDatasetFieldType == null ? datasetFieldType.isRequired() : parentDatasetFieldType.isRequired();
+
+            boolean displayCondition = printOnlyDisplayedOnCreateDatasetFieldTypes
+                    ? (datasetFieldType.isDisplayOnCreate() || isRequired || requiredAsInputLevel)
+                    : !isInputLevel || includedAsInputLevel;
```
qqmyers, line 655: This is the trickiest one - this is not quite the same as` ownerDataverse==null || includedAsInputLevel`?  - isInputLevel could be false if the ownerDataverse is null or if owner is non-null and the field is not in the inputLevels - do you want to include a field with an owner that is not in the inputLevels (along with ones that are in InputLevels and inputLevel.isInclude() from the second clause)? If the logic is correct as is, I guess the isDatasetFieldTypeInInputLevels is needed, but maybe only has to be called here?

GPortas: The case where ownerDataverse is null is controlled and only occurs when the method is invoked from api/metadatablocks API, not from api/dataverses API. There will be no case where fields associated with a Dataverse and fields not associated with a Dataverse are mixed. It's either one or the other, but not both.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -640,18 +640,26 @@ public static JsonObjectBuilder json(MetadataBlock metadataBlock, boolean printO
...
         Set<DatasetFieldType> datasetFieldTypes = new TreeSet<>(metadataBlock.getDatasetFieldTypes());
+
         for (DatasetFieldType datasetFieldType : datasetFieldTypes) {
-            boolean requiredInOwnerDataverse = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeRequiredAsInputLevel(datasetFieldType.getId());
-            boolean displayCondition = !printOnlyDisplayedOnCreateDatasetFieldTypes ||
-                    datasetFieldType.isDisplayOnCreate() ||
-                    requiredInOwnerDataverse;
+            Long datasetFieldTypeId = datasetFieldType.getId();
+            boolean isInputLevel = ownerDataverse != null && ownerDataverse.isDatasetFieldTypeInInputLevels(datasetFieldTypeId);
+            boolean requiredAsInputLevel = isInputLevel && ownerDataverse.isDatasetFieldTypeRequiredAsInputLevel(datasetFieldTypeId);
```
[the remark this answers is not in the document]
GPortas: I just refactored the code a little.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
qqmyers: Not sure I understand. Line 655 says the (owner is null or the owner exists and there is no inputlevel for this field) or (the owner exists, there is an input level for this field, and the includeLevel  for that entry is true).  Should the (owner exists and there is no inputlevel for this field) condition result in this being true?

GPortas: You are right, sorry. I was not correctly applying the Morgan's laws. Indeed, the logic which I think should be the correct one and I just committed, is: `(ownerDataverse == null || includedAsInputLevelInOwnerDataverse || isNotInputLevelInOwnerDataverse)`. I added the isNotInputLevelInOwnerDataverse condition in the OR statement, which checks that when ownerDataverse exists, there might not be an input level for that field, and therefore it should be shown by default, as it exists in the metadata block. I also renamed the variables for better understanding. Please let me know if you find it appropriate.

qqmyers: I think what you're saying means your original code was correct -(owner exists and there is no inputlevel for this field) should result in true (metadata block case), and , unless I'm missing something, the new code has the same logic. I agree the new code is clearer, so I'll go ahead and approve the PR.

qqmyers (approved):
Looks good. All comments addressed.

PR #10644, opened 2024-06-21
addDataverse API endpoint extension for input levels and facet list setup

### LINKED ISSUE #10633: Extend create collection API endpoint to include facet list and input level list

Extend the `addDataverse` API endpoint to support sending values for the input level list and facet list, allowing these properties to be included in the `CreateDataverseCommand` during creation, just as the create collection form currently supports in JSF.

Currently, the `addDataverse` endpoint sets these values as null for a new collection: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java#L153

API User

SPA needs and https://github.com/IQSS/dataverse-client-javascript/issues/156

None

Extend the `addDataverse` API endpoint to support sending values for the input level list and facet list

https://github.com/IQSS/dataverse-client-javascript/issues/156

GPortas:

The addDataverse (`/api/dataverses/{identifier}`) API endpoint has been extended to allow adding metadata blocks, input levels and facet ids at creation time, as the Dataverse page in create mode does in JSF.

- Closes #10633 

I had to modify the `CreateDataverseCommand` to add the logic for adding metadata blocks, as well as other logic related to the Dataverse instance and its input levels and facets. I also had to reorder the code for persisting the dataverse collection with all the changes.

The logic I had to add to the command was previously handled in JSF from the Dataverse Page before calling the command, and without these modifications, calling the command from the API with input levels and/or facet ids resulted in an inconsistent state for the created Dataverse. The API was sending facetIds and inputLevels as null, since they were not supported in the endpoint request body, so these issues went unnoticed.

I have implemented the changes in such a way that they do not interfere with the current behavior of JSF, as the command is still consumed in the same way in JSF and I have not made any changes to the DataversePage.class. I have performed several manual tests with logging to ensure that there is no impact on the behavior of JSF, but due to the lack of test coverage in this area, I recommend an analysis of possible implications by the PR reviewer.

IT tests have been implemented, covering cases where metadata blocks, input levels, or facet IDs are set in the dataverse collection. However, I recommend performing manual tests by calling the endpoint and verifying the correct persistence of the dataverse, its metadata blocks, input levels, and facets.

Create JSON file following the new example provided in the docs:  https://github.com/IQSS/dataverse/blob/e65eb9f947beec06ddbfc6f0bfb73f7ecaf60d28/doc/sphinx-guides/source/_static/api/dataverse-complete-optional-params.json

Call the endoint as described in the docs:

`curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X POST "https://demo.dataverse.org/api/dataverses/root" --upload-file dataverse-complete-optional-params.json`

Now you need to verify that all necessary information about the new dataverse collection, its metadata blocks, input levels, and facet IDs has been correctly persisted. 

We can confirm that the information has been correctly persisted if, after entering the dataverse collection page, the input levels and facets sections have the fields set in the JSON.

Input levels:
Screenshot 2024-07-02 at 15 31 08

Facet ids:
Screenshot 2024-07-02 at 15 31 16

As an additional verification, we can check the persisted information in the related database tables.

- `dataverse_metadatablock` table
- `dataversefieldtypeinputlevel` table
- `dataversefacet`table

None

Yes, attached

**Additional documentation**:
None


src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -123,35 +123,47 @@ public Response addRoot(@Context ContainerRequestContext crc, String body) {
...
         try {
+            JsonObject metadataBlocksJson = newDataverseJson.getJsonObject("metadataBlocks");
+            List<DataverseFieldTypeInputLevel> inputLevels = null;
+            List<MetadataBlock> metadataBlocks = null;
+            if (metadataBlocksJson != null) {
+                JsonArray inputLevelsArray = metadataBlocksJson.getJsonArray("inputLevels");
+                inputLevels = inputLevelsArray != null ? parseInputLevels(inputLevelsArray, newDataverse) : null;
+
+                JsonArray metadataBlockNamesArray = metadataBlocksJson.getJsonArray("metadataBlockNames");
+                metadataBlocks = metadataBlockNamesArray != null ? parseNewDataverseMetadataBlocks(metadataBlockNamesArray) : null;
```
sekmiller, line 148: We need to test to make sure that "citation" is included as a metadata block. If it's not in the json we can either add it automatically (probably preferred) or throw an error.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDataverseCommand.java
```diff
@@ -150,30 +160,38 @@ public Dataverse execute(CommandContext ctxt) throws CommandException {
...
+            if (!inputLevelList.isEmpty()) {
+                managedDv.addInputLevelsMetadataBlocksIfNotPresent(inputLevelList);
+            }
             ctxt.fieldTypeInputLevels().deleteFacetsFor(managedDv);
-            for (DataverseFieldTypeInputLevel obj : inputLevelList) {
-                obj.setDataverse(managedDv);
-                ctxt.fieldTypeInputLevels().create(obj);
+            for (DataverseFieldTypeInputLevel inputLevel : inputLevelList) {
+                inputLevel.setDataverse(managedDv);
+                ctxt.fieldTypeInputLevels().create(inputLevel);
```
sekmiller, line 184: We need to validate the input level objects - in the json you can say         "include": false,
        "required": true and the create Dataverse command will add it to the input level table, but you won't be able to create a dataset because the field won't be in the Create Dataset page, but the "required" will be enforced when you try to save it

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -123,35 +123,47 @@ public Response addRoot(@Context ContainerRequestContext crc, String body) {
...
+            newDataverseJson = JsonUtil.getJsonObject(body);
+            newDataverse = jsonParser().parseDataverse(newDataverseJson);
         } catch (JsonParsingException jpe) {
             logger.log(Level.SEVERE, "Json: {0}", body);
             return error(Status.BAD_REQUEST, "Error parsing Json: " + jpe.getMessage());
         } catch (JsonParseException ex) {
             logger.log(Level.SEVERE, "Error parsing dataverse from json: " + ex.getMessage(), ex);
-            return error(Response.Status.BAD_REQUEST,
-                    "Error parsing the POSTed json into a dataverse: " + ex.getMessage());
+            return error(Response.Status.BAD_REQUEST, "Error parsing the POSTed json into a dataverse: " + ex.getMessage());
```
sekmiller, line 136: Do we want to add this to the bundle so it can be translated?

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
GPortas: I just added a test to verify that a citation is always included, even if it is not specified in the JSON.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
GPortas: Added.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CreateDataverseCommand.java
GPortas: Good point. Added. Thanks.

GPortas:
@sekmiller Requested changes addressed.

sekmiller (approved):
Looks Good! Thanks for the updates and welcome back

landreev:
@GPortas Please note that Jenkins tests keep failing. It appears that the 3 failures from last week happened because the version was still 6.2 in the dataverse-parent pom.xml. But I synced the branch with develop earlier today, and the tests failed again. I can't tell why it failed this time right away. 
https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/view/change-requests/job/PR-10644/24/

GPortas:
@landreev I am still looking for the causes.

The unit test action is also failing, the problem is on war file building. In particular, te issue happens when trying to download an artifact: `'Could not find artifact org.primefaces.themes:all-themes:jar:1.0.10.'` You can see it more detailed in the action logs: https://github.com/IQSS/dataverse/actions/runs/10062830362/job/27861057069?pr=10644

This has happened since the develop merge you pushed, it had not happened before. I reran the action several times and encountered the same problem.

In Jenkins, it seems that the job also fails at the same point, when building the war file, take a look at this screenshot:

Screenshot 2024-07-24 at 14 49 19

But war building and unit tests work on my localhost with no problems...

landreev:
@GPortas yes, this is happening because of #10718 (missing dependency; in preexisting build environments it's still working because the jar file is cached under ~/.m2)

landreev:
I'm ready to merge this, but will wait until tomorrow for the Primefaces dependency issue to be fixed; just to see automated tests pass.

PR #9914, opened 2023-09-14
Assert true or false, not equal to true or false

bencomp:
**What this PR does / why we need it**:

**Which issue(s) this PR closes**:

Closes #

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #10678, opened 2024-07-10
Assert true or false, not equal to true or false

bencomp:
**What this PR does / why we need it**: In order to make tests slightly more readable, this PR replaces calls to `assertEquals` with `true` or `false` as the first argument with calls to `assertTrue` and `assertFalse`, respectively. This PR replaces #9914.

Closes no issues.

**Special notes for your reviewer**: I recreated this encouraged by @stevenwinship. It was previously just a draft PR, probably with the idea that other assertions could also be simplified in a similar way. 

**Suggestions on how to test this**: compile and run the tests.

**Additional documentation**: I noticed other constructs that are used in a similar way, e.g. `MatcherAssert.assertThat(target1.equals(target2), Matchers.is(true));` in `DataverseMetadataBlockFacetTest`


qqmyers (approved):
LGTM

ofahimIQSS:
Merging PR - monitoring Tests on Jenkins

PR #10687, opened 2024-07-15
IQSS/10686 - fix dataset thumbnail

### LINKED ISSUE #10686: Dataset thumbnail broken after metadata edit
Due to #5621, actions like editing metadata after a dataset has auto-assigned a thumbnail from one of its image files cause the thumb to break and then be recorded as failed afterwards.
The root cause appears to be the call to getDatasetThumbnail at https://github.com/IQSS/dataverse/blob/5ba74e8a8e75fecec7f4b477abb920453212c53e/src/main/webapp/dataset.xhtml#L89, which, because the ImageThumbConverter.DEFAULT_PREVIEW_SIZE is not available, attempts to getDatasetThumbnail(0) which then fails.

6.3 - confirmed to exist on beta

PR to follow

qqmyers:
**What this PR does / why we need it**: This PR resolves a recently (v6.3) introduced bug that can cause a dataset thumbnail to appear as a broken link and for a datafile to unnecessarily show the default icon rather than a thumb. Further details are in the issue. 

Closes #10686

The fix is to send the real size (400) rather than 0 from the dataset.xhtml by adding the importConstants call to allow ImageThumbConverter.DEFAULT_PREVIEW_SIZE to be used. It appeared necessary to also add the final keyword to the constant in ImageThumbConverter for this to work, so I added final to all of them there. I also added a check to see if size==0 ever occurs and to avoid going through the logic that would set the persistent previewimagefail flag on the file.

To fix files where that flag has been set by this bug, one can call the /clearThumbnailFailureFlag api as described in the release note. I could add a flyway script for this, but the failure flag could have been set for other (real) reasons and doing so would cause all of those files to retry thumb generation on their next call. So rather than do that automatically, I though adding the api call (or letting people call it just for specific files) might be better. I don't have a strong opinion (we didn't have the fail flag not too long ago so retries were happening then.) 

Strangely, this problem appears to be new at QDR (where #5621 came from) although the line of code in question was there already. I don't know what else might have changed - perhaps we are now catching the size==0 error and setting the flag whereas before that was a silent fail (only affecting the logic in the og header) and retry (showing a good thumb in pages).

I was able to trigger this by adding an image file, seeing the thumb created in the dataset, then going to the main page which triggers the logic to select a thumb for the dataset. Once that has happened, going back to the dataset page and making an edit, e.g. changing the metadata, causes the problem - I think the file thumb disappears immediately and going back to the main page/refreshing will cause the dataset thumb to show a broken link. Adding the  PR should stop any new occurrences. To fix one that's broken, you have to call the /clearThumbnailFailureFlag as in the release note.

included
**Additional documentation**:


cmbz:
2024/07/18 - 6.4 proposal request from @poikilotherm

PR #10698, opened 2024-07-18
IQSS/10697 - Improve batch permission indexing

### LINKED ISSUE #10697: Reindex permissions in batches
The [reindexFilesInBatches method ](https://github.com/IQSS/dataverse/blob/4615f4fe372142d842fe3f9572a46f7f3bdf4ff5/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java#L429) actually indexes in one batch and, when called from [indexPermissionsOnSelfAndChildren](https://github.com/IQSS/dataverse/blob/4615f4fe372142d842fe3f9572a46f7f3bdf4ff5/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java#L362) for a permission change on a collection, results in a list of all datafiles in datasets in that collection and all permission docs for them being loaded into memory. Adding some batching would reduce the memory requirements significantly.

Out-of-memory errors at QDR

PR to follow

qqmyers:
**What this PR does / why we need it**: This PR adds checks within the loops over all children of a definition point and all files of those children and now calls the code to actually process files to create the permissions docs for them and then to submit those documents to solr, thus allowing the memory used to cache them to be released. 

Closes #10697

**Special notes for your reviewer**: I have somewhat arbitrarily chosen to trigger creating documents every 100 files and to then send those docs to solr every 20 docs. Those could be configurable or just optimized with some testing, but the key thing is that, prior to this PR, a change to permissions on the root DV loads all files and all docs into memory, so any cap to avoid doing that should be OK. Choosing too small number is hopefully not a big issue any more as we are relying on auto soft/hard commits, so if we make updates more often than 1 sec/30 seconds, solr is just going to cache the submissions on its side.

Note that one of the problems we detected at QDR was that after a permission change and subsequent out-of-memory failure of our test server, when we restarted, we had out-of-date permissions documents. Those are not orphans so they aren't detected by the /api/admin/index/status call, they just result in odd behavior such as a search showing a closed lock on a restricted file when, if you go the page, you can actually download, etc.

This PR doesn't change the fact that all file permission docs have to be updated, which still take a long time (minutes/hours) on a big installation. It does appear to stop the memory use from climbing though.

Create a db with a significant number of files in one or more datasets (thousands?), change the permissions on the root collection, e.g. adding or removing a person from the curator role. Watch the memory use during processing (top?) and verify that it doesn't keep increasing until the permission updates finish (%cpu should drop and there's a message in the log).

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
@@ -204,6 +204,7 @@ public class SolrIndexServiceBean {
                 } else {
                     perms = searchPermissionsService.findDatasetVersionPerms(datasetVersionFileIsAttachedTo);
                 }
+
                 for (FileMetadata fileMetadata : datasetVersionFileIsAttachedTo.getFileMetadatas()) {
                     Long fileId = fileMetadata.getDataFile().getId();
                     String solrIdStart = IndexServiceBean.solrDocIdentifierFile + fileId;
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I made some cosmetic tweaks. I didn't run the code but it seems fine. Approved.

stevenwinship:
solr

Over 5000 files
Red section is adding the role to a new user
Green section is uploading an additional 1000 files

log messages for each file: "reusing cached perms for file 6353"

stevenwinship:
Tested heavily on Perf cluster. No CPU or Memory issues detected.

PR #10706, opened 2024-07-19
IQSS/10705: Refactor permission filter query in indexing

### LINKED ISSUE #10705: Search permission calculation is complex and can be redundant
The recent FeatureFlags.AVOID_EXPENSIVE_SOLR_JOIN was added to the existing complex logic of creating a filter query based on a user's identity and the groups they belong to. At QDR, this logic was further convoluted by our custom logic to allow full-text search on restricted files. In trying to integrate the new flag (which we plan to use in production), I had to do some refactoring (since we need the list of groups for a user in additional places). The result, when backported, is shorter and less convoluted. I also discovered that a loop over the list of dataverses (when not null) was unnecessary as the code didn't make use of its value.

The associated PR should produce the same result as before but will hopefully be easier to maintain, and potentially a bit faster in some cases.

qqmyers:
**What this PR does / why we need it**: See issue. This is a refactor that should be easier to maintain and slightly more efficient.

- Closes #10705

**Special notes for your reviewer**: FWIW: I wouldn't have done this if QDR didn't need to get the string listing the groups for reuse, which meant I had to pull apart the code generating the query itself from just getting that group list.

**Suggestions on how to test this**: Nominally this PR should produce no change/users should still see the same things in search, correct for their permissions. The easiest way to test may be to just set FINE logging on the SearchServiceBean and verify that the queries produced in both the current and new code are the same.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
```diff
@@ -327,24 +332,13 @@ public SolrQueryResponse search(
...
-            }
-        } else {
-            String permissionFilterQuery = this.getPermissionFilterQuery(dataverseRequest, solrQuery, null, onlyDatatRelatedToMe, addFacets);
-            if (permissionFilterQuery != null) {
-                solrQuery.addFilterQuery(permissionFilterQuery);
-            }
+        // -----------------------------------
+        // PERMISSION FILTER QUERY
+        // -----------------------------------
+        String permissionFilterQuery = this.getPermissionFilterQuery(dataverseRequest, solrQuery, onlyDatatRelatedToMe, addFacets);
```
qqmyers, line 338: the dataverse param was never used in getPermissionFilterQuery - that must have been dropped at some point.

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
```diff
@@ -985,226 +979,99 @@ private String getPermissionFilterQuery(DataverseRequest dataverseRequest, SolrQ
...
         }
-
-        //permissionFilterQuery = publicPlusUserPrivateGroup;
-        logger.fine(publicPlusUserPrivateGroup);
-
-        return publicPlusUserPrivateGroup;
-
+        logger.fine("Groups: " + groupString);
+        String permissionQuery = buildPermissionFilterQuery(avoidJoin, groupString);
+        logger.fine("Permission Query: " + permissionQuery);
```
qqmyers, line 1057: This and the query from the existing code line 1203 should be the same in all cases, whether the avoidJoin flag is set or not.

sekmiller (approved):

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
```diff
@@ -1003,226 +997,99 @@ private String getPermissionFilterQuery(DataverseRequest dataverseRequest, SolrQ
...
+        // add joins on all the non-public groups that may exist for the
+        // user:
 
-            // One or more groups *may* also be available for this user. Once again,
-            // do note that Guest users may be part of some groups, such as 
-            // IP groups. 
-            
-            int groupCounter = 0;
+        // Authenticated users and GuestUser may be part of one or more groups; such
+        // as IP Groups.
```
landreev, line 1051: [nitpick] Not asking to revert this comment change, but just want to mention that I don't think the 2nd sentence there was meaningless or redundant. I thought it was slightly counter-intuitive and/or easy to forget that guest users could be parts of some group. Hence emphasizing it.
Maybe that sentence just needed an "also" - "may _also_ be part of some group".

landreev:
Thank you for cleaning it up, it was messy; and unnecessarily duplicated. 

There is one microscopic difference that I'm feeling guilty for reporting, in how the final query is built for the admin user (i.e., when no permission filter queries are required at all). The server.log entry used to be
```
actual filter queries: [dvObjectType:(dataverses OR datasets)]
```
but now it is 
```
actual filter queries: [dvObjectType:(dataverses OR datasets), ]
```
i.e., there's an extra empty query that's being put on the list, and it results in an extra empty `fq=&` in the final combined query. I cannot think of a situation where this would cause a problem... So I'm going to merge it, but figured I would mention it just in case.

PR #10708, opened 2024-07-20
QDR: PID bug fixes

### LINKED ISSUE #10635: DataCite Citations

* When does this issue occur?

When we run script https://guides.dataverse.org/en/latest/_downloads/7e1b7e580244f61d2ac2de279759d154/counter_weekly.sh

None

* What happens?

When we run script to update citations for Datasets, Dataverse uses api end point "updateCitationsForDataset" in src/main/java/edu/harvard/iq/dataverse/api/MakeDataCountApi.java

This method calls Datacite api to take json information about doi (https://api.datacite.org/events?doi="DOI"&source=crossref&page[size]=1000)

When Dataverse has all information, it uses method "parseCitations" in src/main/java/edu/harvard/iq/dataverse/makedatacount/DatasetExternalCitationsServiceBean.java to extract citations information.

This method discriminates between inbound ("cites", "references", "supplements") and outbound ("is-cited-by", "is-referenced-by","is-supplemented-by") relationships. After that, it converts the recived DOI into Dataverse DOI in both cases with:

	String globalId = subjectUri.replace("https://", "").replace("doi.org/", "doi:").toUpperCase().replace("DOI:", "doi:");

In our case, our :Shoulder is "data" and globalId is transformed in "DATA" and Dataverse doesn't find globalID. Because Dataverse transforms https://doi.org/10.34810/data146 into doi:10.34810/DATA146

We don't know, because you transform  the recived DOI with toUpperCase(). Is it necessary?

When we remove the last part ".toUpperCase().replace("DOI:", "doi:");". The script generates citations correctly.

MakeDataCount API

Should Dataverse change the received DOI to uppercase?

In the guide you don't say anything about that and query to put a new :Shoulder has lowercase.

curl -X PUT -d "MyShoulder/" http://localhost:8080/api/admin/settings/:Shoulder

We use 5.11.1 version, but the latest version also has toUpperCase().

No

**An example:**

https://dataverse.csuc.cat/dataset.xhtml?persistentId=doi:10.34810/data146

imagen

**But 0 Citations.**

imagen

An example removing ".toUpperCase().replace("DOI:", "doi:");"
**2 Citations**

imagen

qqmyers:
Multiple small bug fixes/updates related to the PID handling and MDC citation retrieval code:
* avoid deprecated new URL call
* use correct PID setting when calling MDC API for citations
* make DOI comparisons case insensitive
* also insensitive in managed/excluded lists
* simplify parsing of DOIs in citations
* avoid one empty entry in managed/excluded sets when there are no PIDs listed.
* return correct type for DataCiteDOIProvider
* add is-supplement-to relationship (was supplements only)

Closes #10635

**Suggestions on how to test this**: The main tests would be to confirm that MDC citation retrieval works and that using a shoulder in lower/mixed case works, or vice-versa, that typing DOIs with lower case when the shoulder is upper case works (.e.g. /dataset.xhtml?persistentId=doi:10.33564/fk2LCUIJZ and /dataset.xhtml?persistentId=doi:10.33564/FK2LCUIJZ both resolve to the dataset page). 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/AbstractDOIProvider.java
```diff
@@ -26,6 +27,25 @@ public abstract class AbstractDOIProvider extends AbstractPidProvider {
...
+        managedSet = cleanManagedSet;
+        HashSet<String> cleanExcludedSet = new HashSet<String>();
+        for(String entry: excludedSet) {
+            if(entry.startsWith(DOI_PROTOCOL)) {
+                cleanExcludedSet.add(DOI_PROTOCOL + entry.substring(DOI_PROTOCOL.length()).toUpperCase());
+            } else {
+                logger.warning("Non-DOI found in excludedSet of pidProvider id: " + getId() + ": " + entry + ". Entry is being dropped.");
+            }
+        }
+        excludedSet = cleanExcludedSet;
```
stevenwinship, line 48: It might be interesting to merge from develop to see how this affects the CrossRefDOIProvider which also extends the AbstractDOIProvider.

Also, would you consider moving the 2 for loops into a single method call since they are almost identical?

qqmyers: I merged and made a private clean method.

stevenwinship (approved):

ofahimIQSS:
Fix looks good - merging PR. 
[Testing of 10708.docx](https://github.com/user-attachments/files/17548011/Testing.of.10708.docx)

PR #10710, opened 2024-07-20
IQSS/10709 - Improve index/status and /clear-orphan speed

### LINKED ISSUE #10709: Feature Request/Idea: /api/admin/index/status is slow
Running the /api/admin/index/status (and .../index/clear-orphans) calls are slow and have gotten slower in v6.3 as they now look for more types of orphans. They can be MUCH faster. PR to follow.

Admins checking solr status/doing cleanup

using the api call

qqmyers:
**What this PR does / why we need it**: This PR significantly speeds (10-100x from some informal testing) the index/status and clear-orphan calls by using optimized queries rather than loops/object references. This could be the difference in making these calls usable at all on dbs as big as Harvard's.

Closes #10709

**Suggestions on how to test this**: Run /api/admin/index/status on an existing database, the new code should return the same results much faster. Similarly, /index/clear-orphans should do the same as before but faster.

**Is there a release notes update needed for this change?**: Could - it is just a performance improvement in some admin calls

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/search/IndexBatchServiceBean.java
```diff
@@ -50,35 +50,35 @@ public class IndexBatchServiceBean {
...
         logger.info("Beginning indexStatus()");
-        JsonObject contentInDatabaseButStaleInOrMissingFromSolr = getContentInDatabaseButStaleInOrMissingFromSolr().build();
-        JsonObject contentInSolrButNotDatabase = null;
-        JsonObject permissionsInSolrButNotDatabase = null;
         try {
+            JsonObject contentInDatabaseButStaleInOrMissingFromSolr = getContentInDatabaseButStaleInOrMissingFromSolr().build();
+            JsonObject contentInSolrButNotDatabase = null;
+            JsonObject permissionsInSolrButNotDatabase = null;
+
             contentInSolrButNotDatabase = getContentInSolrButNotDatabase().build();
```
sekmiller, line 58: Dumb question, but in 55,56 why not have the "build" in the same line as declaration as in 54?

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -1405,4 +1405,16 @@ public UploadSessionQuotaLimit getUploadSessionQuotaLimit(DvObjectContainer pare
         
         return new UploadSessionQuotaLimit(quota.getAllocation(), currentSize);
     }
+
+    public boolean isInReleasedVersion(Long id) {
+        Query query = em.createQuery("SELECT fm.id FROM FileMetadata fm, DvObject dvo WHERE fm.datasetVersion.id=(SELECT max(dv.id) FROM DatasetVersion dv WHERE dv.dataset.id=dvo.owner.id and dv.versionState=edu.harvard.iq.dataverse.DatasetVersion.VersionState.RELEASED) AND dvo.id=fm.dataFile.id AND fm.dataFile.id=:fid");
```
landreev, line 1410: This does speed things up as advertised. 
But this query relies on the assumption that the max. database id means this is the latest datasetversion. This is not true with the Harvard database (and a few other oldest installations, for historical/migration reasons), and a reason we make an effort to avoid this assumption elsewhere in the code. 
For the immediate purposes of this PR, trying to gauge the worst possible effect of a false response here - we are only going to miss a few orphaned permission documents in Solr, so, nothing fatal, correct? 

`TL;DR`

As for finding the latest version in the db - it has always bothered me that there is no straightforward query for that. I was thinking of adding a boolean `theLatestReleased` to the table - but then we have to be responsible for keeping it true. We should probably consider a flyway script to rearrange the versions order so that we could rely on it going forward. (How much of an effort would that be? - presumably, even on a huge database like Harvard's, the number of affected versions and cascaded objects relying on them would be limited to a manageable number of such that had been migrated from DVN v3 back in the day - ?). 

There are examples elsewhere in the code of attempts to make custom queries for finding the latest released version. The ones in DatasetService and DatasetVersionService all rely on variations of `SELECT id FROM DatasetVersion ...  ORDER BY versionNumber DESC, minorVersionNumber DESC LIMIT 1`. I cannot really think of anything simpler. 

Throughout the MetricsServiceBean we appear to rely on `select ... max(datasetversion.versionnumber + (.1 * datasetversion.minorversionnumber))`. I remember pointing out some years ago that this wasn't going to work for minor version numbers > 10. But it looks like the clause is still there and has since been copy-and-pasted into more methods. (may not be super crucial for the purposes of the Metrics service though - not sure; but we should probably just change it to `.000001` - ?).

landreev: So, the question is, should I just merge it as is? Or should the query be modified along the lines of 
```... WHERE fm.datasetVersion.id=(SELECT dv.id FROM DatasetVersion dv WHERE dv.dataset.id=dvo.owner.id and dv.versionState=edu.harvard.iq.dataverse.DatasetVersion.VersionState.RELEASED ORDER BY dv.versionNumber DESC, dv.minorVersionNumber DESC LIMIT 1)```?

Issue #22, opened 2014-07-09
Licensing & Terms of Use: Add CC0 and Custom Terms of Use to Dataset


eaquigley:
---

Author Name: **Kevin Condon** (@kcondon)
Original Redmine Issue: 2963, https://redmine.hmdc.harvard.edu/issues/2963
Original Date: 2013-05-17

---

This was requested by Merce and Eleni, see RT #158870:

Yes, good point, it would be CC0 and CC-BY, which are the most standard
options offered in other data repositories.

From the technical side, we should plan this for v3.6 and decide where in
the study cataloging these licenses would appear and how the data depositor
would choose a license.

From the legal, policy side, I'll talk to some people in the Berkman
center/Law School to check if choosing one of these licenses is consistent
with the current DVN terms of use.

Merce

sekmiller:
Basic functionality completed

eaquigley:
@sekmiller We need to have a dataset be given the CC0 license as a default. Currently, when a user publishes a dataset without selecting anything, the tab looks like this:
![licenseeterms](https://cloud.githubusercontent.com/assets/5922904/5620601/783a3024-94fb-11e4-8b5d-dca49ee69249.png)

eaquigley:
@kcondon here are the CC-0 and Terms of Use requirements from conversations with the Berkman Center:
- [ ] CC-0 can be used with restricted files
- [ ] CC-0 cannot be used with Terms of Use
- [ ] If a user has inputted Terms of Use, then the CC-0 license should be negated (if already selected)
- [ ] CC-0 only relates to the way the data is used after downloading not access to download

Let me know if you need other requirements.

kcondon:
Not able to save anything/ additional info, CC0, selecting no to CC0 does not open TOU edit box. Passing back.

sekmiller:
Editing should be working now....

kcondon:
OK, did basic functional testing of forms and all works but for TOU additional info, citation requirements throws 500 error. Will open separate ticket.

Passing to Mike for UI review/ cleanup.

kcondon:
Oh, the more comprehensive, consolidated requirements doc is here:
https://docs.google.com/document/d/1Rg0JTLTeF4seiDdoMDaiWR15WZcH5D7xOtvqTuA0mlI/edit

pdurbin:
See also some comments at https://github.com/IQSS/dataverse/issues/805#issuecomment-70290922

mheppler:
Manage
[...]

PR #10712, opened 2024-07-21
IQSS/10711 Cvoc updates

### LINKED ISSUE #10711: Feature Request/Idea: Allow ORCID and ROR to be used together in author field
Using External vocabulary scripts in the citation author field has several slight differences from other examples we've had before:
* While ORCID nominally can have affiliation info, it sounds like the freedom to select affiliation manually is desired, which means the ORCID script needs to manage the name and idType fields while adding the ORCID to the identifier field, and the ROR script has to handle the affiliation field
* Since the name field exists, it makes more sense for the ORCID script to put plain text values there rather than in the identifier field (as our one-field demos have done) - this means that the identifier field is not included in the display by Dataverse (since it has no value) and hence there's no field for the ORCID script to find
* the idType is an internal cvv field which uses a select element rather than an input
* The identifier field is the last of 4 fields, but we want the person name to appear before the affiliation
* When no ORICD exists for a person, we want to show the idType/identifier fields so others can be entered.
* ORCID and ROR have logos which can be included, but need to be styled

Much of this can be handled by changes to the scripts themselves (See [#22](https://github.com/gdcc/dataverse-external-vocab-support/pull/22)), but some aspects require Dataverse changes:
* Dataverse needs to show a child as a managed field if one is configured but only if the term-uri field has a value
* Managed internal CVV fields need to be marked with the data-cvoc-managedfield attribute
* Styling for drop-down lists needs to apply to ORCID/ROR inputs and display of ORCID/ROR images needs to be styled.

Work at QDR to deploy ORCID and ROR

qqmyers:
**What this PR does / why we need it**: Together with [#22](https://github.com/gdcc/dataverse-external-vocab-support/pull/22), this PR is needed to deploy ORCID/ROR scripts in the author field.

Closes #10711

**Special notes for your reviewer**: @DS-INRA @luddaniel - I think this maintains compatibility with your changes, but it would be good to verify these don't break anything for you.

I also just included [an a11y fix for the labels](https://github.com/IQSS/dataverse/pull/10712/commits/b8878b00f3ece6b9b595fa4367e559fe7db2e471) - basically a bug that I noticed in trying to check accessibility for the author/creator field.

**Suggestions on how to test this**: The simplest test is probably to deploy the updated ORCID/ROR scripts and this PR together and assure that they work together on the author field. (This PR is required for things to work and for the db to get the right values, but the scripts are mostly responsible for the exact look&feel.) Turning the scripts off (by deleting the :CVocConf setting) should show author name/idtype/id correctly set for ORCIDs and with ROR identifiers (vs org names) in the affiliation field. Leaving :CVocConf configured but disabling the scripts (e.g. by adding a typo in the scripts or changing the URLs for them in the CVocConf to be incorrect/give 404s) would let you see the DOM structure and confirm, for example, that the idType field is marked with a data-cvoc-managedfield attribute).

**Additional documentation**:


DS-INRA:
Thanks for the mention, we will check if everything is compatible :)

jeromeroucou:
For information, I've run a few tests using the Ontoportal CVOC in addition to ROR and ORCID. I didn't notice any problems editing or displaying keywords. If needed, I can supply the configuration file used, which will need to be adapted to the metadata schemas (TSV file).

pdurbin (approved):

landreev:
Regardless of the comment in the external-vocab PR 22, I think I'm ready to merge this PR. 
But could you please sync the branch with the main develop. 
There's one funny exception stacktrace I am seeing in server.log; I think it should go away once the branch is updated, but I want to confirm.

landreev:
I believe it's all working. Merging.

PR #10713, opened 2024-07-21
Solr 9.8.0 and other lib updates from QDR

qqmyers:
**What this PR does / why we need it**: QDR has updated to Solr 9.8.0 and uses other libraries that are more recent than the community version. This PR shows what QDR has changed.

Closes #

Note that as of Solr 9.8.0, I was seeing the error  `Search Syntax Error: Error from server at http://localhost:8983/solr/collection1: can not sort on a field w/o docValues unless it is indexed=true uninvertible=true and the type supports Uninversion: nameSort]]` with the old definition using `alphaOnlySort`. The change to `string` seems to address the error and seems like a reasonable change to me - alphaOnlySort has an attribute regarding how to sort empty vals which isn't there on string, but I don't think we ever have empty names.

Heads up for Harvard/large sites: The 9.7.0 release notes include that `Apache Lucene upgraded to 9.11.1 introducing tremendous performance improvements when using Java 21 for vector search among other things.` This PR doesn't change the Java version of anything, but it might be worth trying Solr with Java 21 for performance at Harvard/other large sites.

**Suggestions on how to test this**: Generally test indexing, regression, test the solr install docs, test that the Circuit breaker works at Harvard (@landreev might be able to help here)

**Additional documentation**:


pdurbin (dismissed):
This pull request is not complete. When we upgrade Solr we updated docs, Docker files, etc. Please see #10636 for example.

qqmyers:
Thankfully many few places to change now that the files go in /conf/solr! As of now, the only 9.4.1 references are in the old release notes.

docker/compose/demo/compose.yml
```diff
@@ -103,7 +103,7 @@ services:
   solr:
     container_name: "solr"
     hostname: "solr"
-    image: solr:9.4.1
+    image: solr:9.6.1
```
pdurbin, line 106: You're getting Docker stuff. Good. Can you please also bump the version in .env in the root of the repo?

qqmyers: Done FWIW: I missed it since it hadn't been updated from 9.3.0

pdurbin:
Solr is such a moving target. 9.7 is out. Do we care? Stick with 9.6.1? https://solr.apache.org/downloads.html

p.s. There are merge conflicts.

pdurbin:
Again, Solr is such a moving target. 9.7 is out. Do we care? Stick with 9.6.1? https://solr.apache.org/downloads.html

p.s. There are merge conflicts.

qqmyers:
FYI - @donsizemore ran this with solr 9.7.0 to show the tests pass - in https://jenkins.dataverse.org/job/IQSS-Dataverse-Solr_9.7.0/4/console.
Since the solrconfig.xml appears to not be compatible with solr 9.4.1, it looks like we need to update the normal build once this is merged to get the IT tests to work on develop.

pdurbin (commented):
I'm just clearing "request changes" now that we're trying with the latest Solr (9.7).

qqmyers:
I'm guessing the test fails are because we need to build against solr 9.8.0. The solrconfig.xml used here works fine for me to spin up 9.8.0 on a QDR test machine.

donsizemore:
the Jenkins job failed (no permission to update the build status in QDR's GitHub repo) but all tests passed: https://jenkins.dataverse.org/job/IQSS-Dataverse-Solr_9.8.0/66/testReport/

sekmiller:
This is still failing a bunch of tests as of the latest merge from develop

doc/sphinx-guides/source/_static/installation/files/etc/init.d/solr
```diff
@@ -5,7 +5,7 @@
 # chkconfig: 35 92 08
 # description: Starts and stops Apache Solr
 
-SOLR_DIR="/usr/local/solr/solr-9.4.1"
+SOLR_DIR="/usr/local/solr/solr-9.7.0"
```
sekmiller, line 8: Should be 9.8.0 here

sekmiller (changes requested):
There are some failing tests and should probably get the latest from dev

qqmyers:
Changes made. The tests fail because Jenkins is building against solr <9.8.0 . Don made a special run to show the tests pass against the right solr version - https://github.com/IQSS/dataverse/pull/10713#issuecomment-2644152027

sekmiller (approved):

sekmiller:
@ofahimIQSS, as Jim notes above the tests are failing because Jenkins is on the prior version of Solr. When Don updated the version for a special run they passed.

ofahimIQSS:
regression testing passed - merging PR

PR #10727, opened 2024-07-26
Extend the API for dataverse facets retrieval

GPortas:

Closes #

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
@@ -1045,4 +1045,32 @@ public class DataversesIT {
                 .statusCode(BAD_REQUEST.getStatusCode())
                 .body("message", equalTo("Invalid metadata block name: \"" + invalidMetadataBlockName + "\""));
     }
+
+    @Test
+    public void testListFacets() {
+        Response createUserResponse = UtilIT.createRandomUser();
+        String apiToken = UtilIT.getApiTokenFromResponse(createUserResponse);
+        Response createDataverseResponse = UtilIT.createRandomDataverse(apiToken);
+        createDataverseResponse.then().assertThat().statusCode(CREATED.getStatusCode());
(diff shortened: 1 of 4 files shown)


src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -1045,4 +1045,32 @@ public void testAddDataverse() {
                 .statusCode(BAD_REQUEST.getStatusCode())
                 .body("message", equalTo("Invalid metadata block name: \"" + invalidMetadataBlockName + "\""));
     }
+
+    @Test
+    public void testListFacets() {
+        Response createUserResponse = UtilIT.createRandomUser();
+        String apiToken = UtilIT.getApiTokenFromResponse(createUserResponse);
+        Response createDataverseResponse = UtilIT.createRandomDataverse(apiToken);
```
sekmiller, line 1053: Suggestion: instead of using createRandomDataverse and relying on the facets in "Root" to never change, maybe look at the test around lines 975-990 above to see how to set actual facets just in case root ever changes.

sekmiller (changes requested):
Looks good overall - I just have one suggestion on one of the tests. Let me know if you agree. thanks!

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
sekmiller: Should probably get the latest from dev while you're at it.

GPortas: Good point, thanks. I have added a test case for collections with custom facets (not inherited from root) and refactored the code in the UtilIT class a bit.

sekmiller (approved):
Looks Good. Thanks for the updates!

stevenwinship:
@GPortas Could you update and resolve conflicts please? I will re-test once done and merge

GPortas:
Done @stevenwinship thank you.

PR #10729, opened 2024-07-26
Remove InReview lock when dataset is moved to a collection where the author can publish

### LINKED ISSUE #6575: When a dataset is moved from one dataverse to another, the dataset should take on the new dataverse workflow
a dataset was moved from a journal to harvard root after it had been submitted for review.
in root, the dataset workflow  setting did not change from that of the journal-"submit for review" was not removed

see attached screen shot

Screen Shot 2020-01-24 at 12 56 46 PM

*i requested permission from journal to move published dataset back to their journal dataverse as should be the case

sekmiller:
When a dataset is moved to a new Dataverse Collection should take on the same workflow as the new collection. 

Closes #6575 When a dataset is moved to a new Dataverse Collection should take on the same workflow as the new collection. 

The logic is based on the permissions of the default contributor of the new/target collection. If the default contributor has publish dataset permission and the moved dataset is locked because it is in review, that lock is removed. Unit tests also added.

Move datasets to Dataverses with the same and different permissions for the default contributor. This should also work if the default contributor role is custom made for the target collection

kind of an edge case bug fix - probably not necessary to highlight it.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
@@ -7,8 +7,10 @@ package edu.harvard.iq.dataverse.engine.command.impl;
 
 import edu.harvard.iq.dataverse.Dataset;
 import edu.harvard.iq.dataverse.DatasetLinkingDataverse;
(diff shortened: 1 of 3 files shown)


pdurbin (approved):
I didn't test this but the change makes sense. Tests are passing. Approved.

I did give this PR a more descriptive title. Please feel free to edit it further.

PR #10746, opened 2024-08-07
#10403 if a dataset is not published only users with ViewUnpublishedD…

### LINKED ISSUE #10403: Feature Request/Idea: ViewUnpublishedDataset should not include file download
There are two permissions which have undocumented hierarchical relationship: ViewUnpublishedDataset and DownloadFile. The user who has no DownloadFile, but has ViewUnpublishedDataset permission can download files, which is - according to our users and I agree with them - counter intuitive. I expect that the person who does not have right to download files should not be able to download files.

On code level (https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Access.java#L1868):

```Java
private boolean isAccessAuthorized(User requestUser, DataFile df) {
  ...

  if (!published) { // and restricted or embargoed (implied by earlier processing)
    // If the file is not published, they can still download the file, if the user
    // has the permission to view unpublished versions:

    // This line handles all three authenticated session user, token user, and guest cases.
    if (permissionService.requestOn(dvr, df.getOwner()).has(Permission.ViewUnpublishedDataset)) {
      // it's not unthinkable, that a GuestUser could be given
      // the ViewUnpublished permission!
      logger.log(Level.FINE,
        "Session-based auth: user {0} has access rights on the non-restricted, unpublished datafile.",
        dvr.getUser().getIdentifier());
      return true;
    }
  } else { // published and restricted and/or embargoed
    // This line also handles all three authenticated session user, token user, and guest cases.
    if (permissionService.requestOn(dvr, df).has(Permission.DownloadFile)) {
      return true;
    }
  }
  ...
}
```

There are two suggestions from our users:

1. The person who has permission `ViewUnpublishedDataset` can _view_ the dataset and files metadata only without _downloading_ files. The person who has permission `ViewUnpublishedDataset` AND `DownloadFile` can _download_ files from unpublished dataset.
2. The permissions and dependencies among them should be clearly documented. Right now there is nothing about them in the documentation and in the user interface other than their names.

I think that introducing  the first suggestion might break existing permissions, so if it would be introduced 
1. there should be a migration process that would add `DownloadFile` permissions who has `ViewUnpublishedDataset`
2. there should be a flag to turn this feature on, so the administrator of the Dataverse service could decide to follow this policy or not

all

Inconsistency in the current behaviour of the software.

Who can download files of unpublished datasets.

No, but I would like to change who can download files from unpublished datasets.

Yes, see https://github.com/IQSS/dataverse/issues/4389#issuecomment-578191483

pkiraly:
…ataset AND DownloadFile rights can download

There are two permissions which have undocumented hierarchical relationship: ViewUnpublishedDataset and DownloadFile. The user who has no DownloadFile, but has ViewUnpublishedDataset permission can download files, which is - according to our users and I agree with them - counter intuitive. I expect that the person who does not have right to download files should not be able to download files. So this PR adds a second condition: only users can download filed from unpublished datasets who have both ViewUnpublishedDataset and DownloadFile rights.

Closes #10403 

1) Create a user with a role covering ViewUnpublishedDataset (user1), and another one who have role covering ViewUnpublishedDataset and DownloadFile (user2)
2) create a dataset (but not publish it) with a single file
3) be sure that user1 can not download the file, but user2 can

No

No

**Additional documentation**:
No

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -1831,10 +1831,11 @@ public class Access extends AbstractApiBean {
         }
         if (!published) { // and restricted or embargoed (implied by earlier processing)
             // If the file is not published, they can still download the file, if the user
-            // has the permission to view unpublished versions:
+            // has the permission to view unpublished versions and download files:
 
             // This line handles all three authenticated session user, token user, and guest cases.
-            if (permissionService.requestOn(dvr, df.getOwner()).has(Permission.ViewUnpublishedDataset)) {
+            if ( permissionService.requestOn(dvr, df.getOwner()).has(Permission.ViewUnpublishedDataset)
+              && permissionService.requestOn(dvr, df.getOwner()).has(Permission.DownloadFile)) {
                 // it's not unthinkable, that a GuestUser could be given
                 // the ViewUnpublished permission!
                 logger.log(Level.FINE,


qqmyers:
@pkiraly - I haven't tested this, but I assume a change like this would have impacts outside just this one permission check, e.g. file download permissions would need to be added when creating a dataset, and, in existing systems, permissions would have to be added so access to current datasets won't change just by installing the new version, etc. Are you looking into those?

Given that, another question - would it acceptable/easier just to change the label(s) of the permission(s), e.g. ViewAndDownloadFromUnpublishedDataset and DownloadReleasedFile or similar? If there are really use cases where someone needs access to view the dataset but not be able to access the files, then perhaps the permission definitions need to change as you indicate, but if it is just a matter of clarifying what is allowed, perhaps label changes are enough?

pkiraly:
@qqmyers This feature s really requested by our users. I am testing other use cases if it cases any problems elsewhere.

pdurbin:
@pkiraly any news on your further testing? I agree with @qqmyers that the current behavior is so baked into the system that there are probably other places in the code that would need to be updated for consistency.

src/main/java/edu/harvard/iq/dataverse/api/Access.java
stevenwinship, line 1: Consider adding the Permission.DownloadFile permission check to line 682
```
if (draft != null && permissionService.requestOn(req, retrieved).has(Permission.ViewUnpublishedDataset)) {                    
                    String fileIds = getFileIdsAsCommaSeparated(draft.getFileMetadatas());
                    // We don't want downloads from Draft versions to be counted, 
                    // so we are setting the gbrecs (aka "do not write guestbook response") 
                    // variable accordingly:
                    return downloadDatafiles(getRequestUser(crc), fileIds, true, uriInfo, headers, response);
                }
```

pdurbin:
@pkiraly hi! I'm just checking in. Are you still working on this?

pdurbin:
@pkiraly we haven't heard from you so I'm going to close this. Please consider it a soft close. 😄 We can reopen it, if you like. Please see the feedback above, especially the suspicion that there are probably additional places in the code that probably need to be changed.

PR #10632, opened 2024-06-14
DataciteXML changes Plus RelationType field

qqmyers:
**What this PR does / why we need it**: This PR adds a RelationType child field to the related publication parent field and uses it to provide a RelationType in the OpenAire and DataCite XML exports, DataCite XML sent to dataset (and the JSON and OAI_ORE exports which include all fields). It builds upon #10615 and should be reviewed/QA'd after that (or we can create a PR against that branch to more easily see the changes just to add a RelationType.

**Which issue(s) this PR closes**:

- Closes #5889
- Addresses #2778
- Addresses #8108 
- Addresses #5144

Relates to:

- #2917
- #7072
- #8798
- #9330

**Special notes for your reviewer**:

**Suggestions on how to test this**: Nominally the new XMLTemplateTest (and all others) should pass and it should be possible to publish datasets with any/all metadata using a DataCite test account. The log shouldn't contain any issues where DataCite responds with a 422 and indicates that the XML doesn't comply with their 4.5 schema.  There should be lots of additional metadata for related publications, author entries should include ORCID info if provided and affiliations and GrantNumberAgency should have ROR info if a ROR rather than plain text was entered. Typos like having a related publication with id type doi and either no or non-DOI entries for the identifier and url should result in a log message and that particular related Publication not getting included in the XML, but otherwise should not cause a failure to update the XML. Etc. 

FWIW: I have been able to run this on all the QDR production data and have everything update OK (though we have a few typos in the metadata to fix). 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

Yes, it adds "Relation Type" to "Related Publication":

Screenshot 2024-09-17 at 2 40 17 PM
Screenshot 2024-09-17 at 2 40 23 PM

**Is there a release notes update needed for this change?**: included.

**Additional documentation**: As noted in the release note, there's a long doc listing ~all of the intended changes from the previous version - see https://docs.google.com/document/d/1JzDo9UOIy9dVvaHvtIbOI8tFU6bWdfDfuQvWWpC0tkA/edit?usp=sharing.

Changes to the guides can be previewed at https://dataverse-guide--10632.org.readthedocs.build/en/10632/admin/dataverses-datasets.html#send-metadata-to-pid-provider

PR #10615, opened 2024-06-05
Datacite xml improvements

qqmyers:
Note - this is a subset of the changes in #10632 / #10632 adds relatedPublication relationType in addition to updating the XML sent to DataCite as described here. Nominally only #10632 needs to be merged, but this could be merged first to make it easier to see just the relationType changes in #10632.

**What this PR does / why we need it**: This PR updates the XML sent to DOI providers to be richer and more consistent with what is in the OpenAire export. There are differences (most obviously, the code here works for draft datasets and files which are not handled by the OpenAire exporter), most are probably subtle but @jggautier or others may wish to look at/comment on the list of changes I've tried to track in [this doc](https://docs.google.com/document/d/1JzDo9UOIy9dVvaHvtIbOI8tFU6bWdfDfuQvWWpC0tkA/edit?usp=sharing).

This is nominally one of 2 PRs related to DOI metadata - the other will be a change to the citation metadatablock to add a relationshipType to the field(s) where we expect to report the relationships to DataCite/etc.

**Which issue(s) this PR closes**:
Probably closes #5889 when combined with the relationshipType PR.
Closes #

**Special notes for your reviewer**: The PR includes some refactoring of our XML generation code that is common with the DDI export. @poikilotherm's idea of generating a model from the DataCite schema might be a good next step (hopefully easier given the code here to get the content needed). 

**Suggestions on how to test this**: QDR is testing this code now. There is one new test in addition to the existing ones - it could/should be expanded to test a dataset with more of the metadata that is covered by the new code - possibly just reading in one of the ~complete dataset examples we have.  Verification that the result matches the DataCite schema is included.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: only changes the XML sent to the DOI provider which is also exposed as a metadata export. The change makes the OpenAire export somewhat redundant, but as noted above, there are differences which may need to be kept. 

**Is there a release notes update needed for this change?**: yes - will create once the set of changes (in the doc above) is reviewed. Probably should be combined with the citation block change if that can be created in time for the same release.

**Additional documentation**:

PR #10759, opened 2024-08-08
fix: insert correct identifierType into DataCite export XML

vera:

I noticed when exporting the metadata of a dataset with a Permalink PID that the DataCite XML export contained the wrong identifierType:

```
<identifier identifierType="DOI">https://example.com/mypermalink1234</identifier>
```

This PR fixes the bug by using `getProviderType` to insert the correct identifier type.

I also fixed some typos and added a missing return value for `getProviderType`.

not aware of any issue

\-

I tested this locally by creating a dataset with a Permalink PID and then using the "Export Metadata" > "DataCite" button in the UI.

\-

I think not

**Additional documentation**:

\-


pdurbin:
@vera thanks for the pull request!

Just a heads up that there is a significant refactoring by @qqmyers of the DataCite code in the works:

- #10615

qqmyers:
FWIW: The new code does set the identifierType (to "DOI", "Handle", or "URL" (for perma)) which should be allowed values for the DataCite schema. (Maybe the original code was written before exporters/when the output was only sent to DataCite and DOI was always true?) I'm not sure how fast #10615 is moving, so if this gets merged first (so we get the other fixes), I think it's fine.

vera:
Ah, that's great. Sorry, I wasn't aware of that.

Yes, I also checked the DataCite schema. I didn't find a list of allowed values, only that "identifierType" should be a non-empty string, so any string should be fine I think.

qqmyers:
You're right. In this case it wasn't the DataCiteSchema but the [OpenAire Guildelines](https://guidelines.readthedocs.io/en/latest/data/field_identifier.html) that had a list.

johannes-darms:
Can we integrate this into version 6.4 if https://github.com/IQSS/dataverse/pull/10615 is not integrated?

pdurbin:
@vera can you please resolve merge conflicts in this pull request? They are due to us merging this PR:

- #10632

vera:
Hmm, I actually think this PR is not necessary anymore. The main bug this PR fixes (wrong `identifierType` in the DataCite export) is already fixed by #10632 (here: https://github.com/IQSS/dataverse/pull/10632/files#diff-ceea58ab3cfad4c5eb5265ad85516df9af3b75f31b32b764367989d5c59c84a0R225-R237).

I've tested the develop branch with a PermaLink dataset and it's now correct:

`<identifier identifierType="URL">https://csh.nfdi4health.de/resource/health-IP89GF</identifier>`

The rest of the changes in this PR don't need to be merged I think.

pdurbin:
I'm confused why this pull request doesn't have a merge conflict because src/main/resources/edu/harvard/iq/dataverse/pidproviders/doi/datacite_metadata_template.xml was removed in b1e5020d5c44a2c1dd3c8f581482f4db17e1dcd1 🤔 

@vera anyway, I think this PR will need to be adjusted to the new way added here:

- #10632

pdurbin:
> Hmm, I actually think this PR is not necessary anymore.

Oh! @vera please feel free to close this PR if you don't need it. 😄

PR #10760, opened 2024-08-08
Add mimetype for 3D Data readable by the Smithsonian's Voyager app

qqmyers:
**What this PR does / why we need it**: This adds support for *.glb model/gltf-binary 3D models to the mime detection and adds label/facet entries for 3D model / 3D Data

**Which issue(s) this PR closes**:  Needed to support https://github.com/gdcc/dataverse-previewers/issues/57

Closes #

**Special notes for your reviewer**: See https://github.com/gdcc/dataverse-previewers/issues/57

I've also added */* to the @Produces annotation on the file download API call. I'm not sure why it is application/xml right now (perhaps only */* is needed), but just having application/xml caused a 406 error (with type application/json :-) )when the Voyager app sent an erroneous Accept:"text/plain" header (it also failed with 406 if I made their code ask for model/gltf-binary). When the file is actually returned, it has type model/gltf-binary as it should with this PR, but without listing that explicitly or having */*, we return 406 before getting to our code due to the annotation.

**Suggestions on how to test this**: Minimally, upload a file with a glb extension and verify it gets the mimetype and label above and shows up in a 3D Data facet. To test the overall previewer that handles this mimetype, install a local copy of the new Voyager previewer in https://github.com/gdcc/dataverse-previewers/pull/77, register the new previewer using the curl command, try uploading data such as https://dataverse.harvard.edu/file.xhtml?fileId=6931827&version=1.0 (note you have to upload this file after you deploy the branch or you'll have to redetect the mimetype via API (or change in the DB)). It should show a dynamic 3D view like: 
image

**Is there a release notes update needed for this change?**: Maybe? There are a few 'beta' 3D previewers that haven't been added to the guides or release notes. Not sure if we want to do that now or when there's a v1.5 release (could be around the 6.4 release if we agree these are no longer beta).

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -286,7 +286,7 @@ public class Access extends AbstractApiBean {
     @GET
     @AuthRequired
     @Path("datafile/{fileId:.+}")
-    @Produces({"application/xml"})
+    @Produces({"application/xml","*/*"})
     public Response datafile(@Context ContainerRequestContext crc, @PathParam("fileId") String fileId, @QueryParam("gbrecs") boolean gbrecs, @Context UriInfo uriInfo, @Context HttpHeaders headers, @Context HttpServletResponse response) /*throws NotFoundException, ServiceUnavailableException, PermissionDeniedException, AuthorizationRequiredException*/ {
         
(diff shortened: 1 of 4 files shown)


pdurbin (approved):
I didn't test this but the changes are minimal and make sense. Approved.

PR #10767, opened 2024-08-12
Fix dataverses/{identifier}/metadatablocks endpoint to not return fields marked as displayOnCreate=true if there is an input level with include=false

### LINKED ISSUE #10741: Fix ListMetadataBlocksCommand to not return fields marked as displayOnCreate=true if there is an input level with include=false

Create a collection from JSF, in the metadata blocks section, uncheck the option to inherit the metadata blocks from the parent. In citation, deselect the Notes field, which is a field with displayOnCreate=true in the database. 

Submit the new collection. When creating a dataset in the new collection, you will see that Notes does not appear, even though the configuration will remain with displayOnCreate=true in the database. The input level with include=false takes precedence over displayOnCreate=true and therefore is not displayed. 

The API for collection/{id}/metadata blocks, which uses ListMetadataBlocksCommand, returns this field when displayOnCreate=true. 

Since there is an input level with include=false, this field should not be returned from the API.

We need to update the related queries in **MetadataBlockServiceBean**

* When does this issue occur?

We detected it on the SPA, since it uses the mentioned endpoint.

None on JSF

* What happens?

n/a

API users

The API for collection/{id}/metadata blocks, which uses ListMetadataBlocksCommand, should not any field with displayOnCreate=true if there is an input level with include=false.

latest develop

- https://github.com/IQSS/dataverse-frontend/issues/444

GPortas:

The issue caused fields with displayOnCreate=true to be returned for the create dataset form on the SPA when there was an input-level configuration for the collection and field with include=false.

If an input-level configuration with include=false is set, the field should not be returned even if it has displayOnCreate property set to true, as is the case with JSF, where the field is not displayed in the create dataset form.

This PR also refactors the logic of the JsonPrinter class for filtering fields based on possible conditions by using criteria queries instead of post-filtering conditions within the for loop that iterates over the fields of each metadata block.

- Closes #10741

New tests have been added for the problematic scenario

Create a collection from JSF, in the metadata blocks section, uncheck the option to inherit the metadata blocks from the parent. In citation, deselect the Notes field, which is a field with displayOnCreate=true in the database.

Call the following endpoint for the new dataverse, with returnDatasetFieldTypes=true and onlyDisplayedOnCreate=true. This is the same call the SPA uses for the create dataset form.

 curl -H "X-Dataverse-key:<API_TOKEN> "$SERVER_URL/api/dataverses/<DATAVERSE_ID>/metadatablocks?returnDatasetFieldTypes=true&onlyDisplayedOnCreate=true"

The notes text, which has include=false as the input level condition, should not be retrieved.

No

Yes, attached.


src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
```diff
@@ -851,4 +852,73 @@ public String getFieldLanguage(String languages, String localeCode) {
...
+        CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
+        CriteriaQuery<DatasetFieldType> criteriaQuery = criteriaBuilder.createQuery(DatasetFieldType.class);
+
+        Root<MetadataBlock> metadataBlockRoot = criteriaQuery.from(MetadataBlock.class);
+        Root<DatasetFieldType> datasetFieldTypeRoot = criteriaQuery.from(DatasetFieldType.class);
+
+        criteriaQuery.where(criteriaBuilder.and(
+                        criteriaBuilder.equal(metadataBlockRoot.get("id"), metadataBlock.getId()),
+                        datasetFieldTypeRoot.in(metadataBlockRoot.get("datasetFieldTypes")),
+                        criteriaBuilder.isTrue(datasetFieldTypeRoot.get("displayOnCreate"))
```
sekmiller, line 866: Do we also need to say "or 'required' is True"? In my local DB producer name is required and displayoncreate is false - not sure it that's just because my local DB is messed up.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
```diff
@@ -851,4 +852,73 @@ public String getFieldLanguage(String languages, String localeCode) {
...
+
+        Predicate hasNoInputLevelPredicate = criteriaBuilder.not(criteriaBuilder.exists(subquery));
+
+        Predicate displayedOnCreatePredicate = onlyDisplayedOnCreate
+                ? criteriaBuilder.or(
+                criteriaBuilder.isTrue(datasetFieldTypeRoot.get("displayOnCreate")),
+                requiredAsInputLevelPredicate
+        )
+                : criteriaBuilder.conjunction();
+
```
sekmiller, line 910: Looks good. I haven't done much with criteria builder so maybe adding some comments would help.

Thanks

GPortas:
@sekmiller I've updated the code to consider the fields required by the installation (not as a dataverse input level). I’ve added logic to handle the case of conditionally required fields (if a subfield is required but the parent field is not, then do not display it on create).

sekmiller (approved):
Looks good. Thanks for the updates

PR #10775, opened 2024-08-15
Update handling of permalink base-url setting

### LINKED ISSUE #10771: Permalinks in S3 storage identifiers: treatment of slashes
After creating a dataset in a Dataverse configured for Permalink PIDs and S3 storage, I noticed that the dataset's storage identifier looks like this:

`s3://https://example.com/mypermalinks//XYZ...`

(where `https://example.com/mypermalinks/` is the authority I have configured and `XYZ...` is the unique part of the identifier)

In this example, the Permalink PID that is displayed in the Dataverse UI is `https://example.com/mypermalinks/XYZ...`.

Note that the storage identifier contains a double slash preceding the unique part, but the PID does not. Is this a bug in the storage identifier generation?

In my S3 management console as well as using the S3 command line tool `s3cmd`, the double slashes (both the "buggy" slash and the one in `https://`) are interpreted as a kind of "folder" structure that is weird to navigate.

Here is a snippet from the Amazon S3 docs about object key names:

> The console uses the key name prefixes (Development/, Finance/, and Private/) and delimiter ('/') to present a folder structure.
> ...
> Amazon S3 supports buckets and objects, and there is no hierarchy. However, by using prefixes and delimiters in an object key name, the Amazon S3 console and the AWS SDKs can infer hierarchy and introduce the concept of folders.
> The Amazon S3 console implements folder object creation by creating a zero-byte object with the folder prefix and delimiter value as the key.
> ...
> ### Characters that might require special handling
> The following characters in a key name might require additional code handling and likely need to be URL encoded or referenced as HEX. 
> ...
> * ...
> * Forward slash ("/")
> * ...

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

Is this creation of a folder structure in S3 intended? If yes, I think the folder structure should be created in a way that makes more sense (e.g. don't create a weird empty-name folder because of the double slash at the start of a URL (`https://...`)).

In any case, I think it might make sense to URL-encode some or all of the slashes inside Permalink PIDs when generating the storage identifier, as suggested by the Amazon docs (e.g. set a storage identifier like `s3://https%3A%2F%2Fexample.com%2Fmypermalinks%2F%2FXYZ...` (< this would be an option that wouldn't create any nested folder structure inside the S3 bucket)).

What do you think?

vera:

This updates the handling of the configuration setting `dataverse.pid.*.permalink.base-url` to support greater flexibility. Previously, the string "/citation?persistentId=" was automatically appended to the configured base URL. With this update, the base URL will now be used exactly as configured, without any automatic additions.

- Closes #10771 

This issue is about a slightly different topic (treatment of special characters in S3 storage identifiers). However, that issue only occurred because I was using the Permalink configuration in an unintended way to get around the limitation of the automatically inserted "/citation?persistentId=". With this fix, my S3 storage identifiers no longer need to contain special characters.

In case others may be using special characters inside authority/shoulder values for other reasons, we added a warning in the documentation that this may cause issues.

\-

Run `PidUtilTest` which was updated in this PR

\-

Yes, there needs to be a note including upgrade instructions for installations which are already using the `base-url` setting. To keep the same PermaLinks as before, these installations must manually append  "/citation?persistentId=".

(included in this PR)

**Additional documentation**:

Preview docs at https://dataverse-guide--10775.org.readthedocs.build/en/10775/installation/config.html#permalink-specific-settings


qqmyers (approved):
Looks good. Thanks!

johannes-darms:
Can we integrate this fix into the 6.4 release?

pdurbin:
> Can we integrate this fix into the 6.4 release?

Someone (I can't tell who) put it under "6.4 release proposals". See also https://dataverse.zulipchat.com/#narrow/stream/375707-community/topic/Release.206.2E4.20Proposals/near/452163912

pdurbin:
Tests are passing at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10775/1/testReport/ but there was a later failure. I merged the latest from develop to see if we can get a clean run.

vera:
Thank you @pdurbin :) let me know if there's anything I can do!

pdurbin (approved):
I didn't run the code but it makes sense and tests are passing. Approved.

pdurbin:
This seems to work fine. I tested it with the following settings:

        -Ddataverse.pid.providers=perma1
        -Ddataverse.pid.default-provider=perma1
        -Ddataverse.pid.perma1.type=perma
        -Ddataverse.pid.perma1.label=Perma1
        -Ddataverse.pid.perma1.authority=10.5072
        -Ddataverse.pid.perma1.shoulder=FK2/
    #   -Ddataverse.pid.perma1.permalink.base-url=https://perma.dataverse.org

`base-url` was present when I created "test" below and absent when I created "test 2":

Screenshot 2024-09-16 at 4 15 12 PM

Since https://perma.dataverse.org10.5072fk2/XS8UZ2 is all squished together I think I'll add a note to the docs to include a trailing `/`.

PR #10776, opened 2024-08-15
Trailing comma in author name causes error

### LINKED ISSUE #10343: Trailing comma in the author name causes json-ld failure, kills the dataset page
I thought we had an issue for this, but can't find it immediately; but please close if this is a duplicate. 
A super embarrassing/annoying bug: 
Something in the json-ld/schema.org code is expecting a comma in the author string to indicate the "LastName, FirstName" notation; but fails miserably if there's a trailing comma, i.e. "SomeName,". As in
Screen Shot 2024-02-26 at 12 46 24 PM

The dataset page uses the json-ld fragment in the headers and fails catastrophically when the record cannot be exported: 
Screen Shot 2024-02-26 at 12 33 39 PM

Should be trivial to fix, I'm assuming. (in both places, ideally) 
(A quick workaround for the page failure is to place an empty `export_schema.org.cached` in to the dataset directory/s3 folder)

erodde:
Fixes a bug where a trailing comma in an author name leads to an error. Error occured to us in production recently, version 6.2.

- Closes #10343

See unit test or reproduce error by adding author name with trailing comma to dataset. Once dataset is published, error occurs on reload.

I added one just to be sure.


qqmyers (approved):
Makes sense. Thanks!

src/test/java/edu/harvard/iq/dataverse/util/PersonOrOrgUtilTest.java
```diff
@@ -89,6 +89,7 @@ public void testName() {
             verifyIsPerson("kcjim11, kcjim11", "kcjim11", "kcjim11");
             
             verifyIsPerson("Bartholomew 3, James", "James", "Bartholomew 3");
+            verifyIsPerson("Smith, ", null, "Smith");
```
bencomp, line 92: What happens if there is no space at the end? Should you add that as another test case?

erodde: Verifying failed because the compared JsonObject was normalized via an util method. I added the normalization to the comparison in the assertion and added the other test case.

landreev (approved):

poikilotherm:
/push-image to allow for QA testing, thx

poikilotherm (approved):
By applying this fix, I can add a name `Schmidt, ` and it doesn't break JSON-LD anymore when I hit "publish".

grafik

I also used `Schmidt, Test, ` as an example, stays unaffected. API output looks good, too. I verified that `Schmidt, ` leads to an Error 500 page on `gdcc/dataverse:unstable`. `Schmidt, Test, ` was never affected.

PR #10778, opened 2024-08-19
Extension of API `{id}/versions` and `{id}/versions/{versionId}` with an optional ``excludeMetadataBlocks`` parameter

### LINKED ISSUE #10171: Feature Request/Idea: List Versions of a Dataset API reduced response size
Extend the API `api/datasets/%s/versions` so that it only returns version information and omits other properties such as files, metadata blocks and so on. Some API clients are only interested in the existence of versions and not in all details of them, if necessary these can be determined by another request.  Therefore, I would like to extend the API with a flag to get a shorter response. This should be disabled by default to ensure backwards compatibility.

API User

Slow response from `api/datasets/%s/versions` due to the large response body. Most of the information included is not needed as I just want to display a dropdown list with all available versions. 

Extension of the API `api/datasets/%s/versions` 

None

None

johannes-darms:

Extension of API `{id}/versions` and `{id}/versions/{versionId}` with an optional ``excludeMetadataBlocks`` parameter,
that specifies whether the metadataBlocks should be listed in the output. It defaults to ``false``, preserving backward
compatibility. (Note that for a dataset with a large number of versions and/or metadataBlocks having the metadata blocks
included can dramatically increase the volume of the output).

We have slow response from `api/datasets/%s/versions` due to the large response body. Most of the information included (metadatablocks) is not needed as we just want to display a dropdown list with all available versions.

- Closes #10171 

**Suggestions on how to test this**: Call the API once with the flag and once without. 

**Is there a release notes update needed for this change?**: Maybe, it is a new optional property.  

Preview docs at https://dataverse-guide--10778.org.readthedocs.build/en/10778/api/native-api.html#list-versions-of-a-dataset


johannes-darms:
@GPortas we are experiencing some performance issues with our SPA when a user requests information about dataset versions, particularly those with many versions. Are you experiencing similar problems? We believe that reducing the payload by omitting the metadata would solve our problem. As we can load the metadata with another query if needed. What do you think?

GPortas:
> @GPortas we are experiencing some performance issues with our SPA when a user requests information about dataset versions, particularly those with many versions. Are you experiencing similar problems? We believe that reducing the payload by omitting the metadata would solve our problem. As we can load the metadata with another query if needed. What do you think?

I'm not sure if we've experienced issues with the metadata blocks, and if we have, they may have been minor, possibly because we don't tend to add complex metadata block configurations in our test datasets.

It's reasonable to think it will improve performance, as additional queries are omitted. This is somewhat similar to what we did with the files, where we added the optional query parameter called `excludeFiles`. 

This makes me wonder if it might be interesting to create a 'reduced information' endpoint instead of continuing to include parameters for excluding properties in the general endpoint.

pdurbin:
Here are the docs for `excludeFiles`: https://guides.dataverse.org/en/6.4/api/native-api.html#get-version-of-a-dataset

johannes-darms:
> > @GPortas we are experiencing some performance issues with our SPA when a user requests information about dataset versions, particularly those with many versions. Are you experiencing similar problems? We believe that reducing the payload by omitting the metadata would solve our problem. As we can load the metadata with another query if needed. What do you think?
> 
> I'm not sure if we've experienced issues with the metadata blocks, and if we have, they may have been minor, possibly because we don't tend to add complex metadata block configurations in our test datasets.
> 
> It's reasonable to think it will improve performance, as additional queries are omitted. This is somewhat similar to what we did with the files, where we added the optional query parameter called `excludeFiles`.
> 
> This makes me wonder if it might be interesting to create a 'reduced information' endpoint instead of continuing to include parameters for excluding properties in the general endpoint.

Our problem is only partly caused by the large complex metadata block, the other cause is the amount of versions (we have a dataset where an update is published every day, the file changes but the metadata is the same). So the payload of this API becomes huge and by omitting the metadata blocks we can mitigate the problem while still getting the necessary information about versions without introducing paging.     

I'm not a fan of having different endpoints for more or less the same information. It is more code to maintain and more complicated for the user. 

This PR is inspired by the `excludeFiles` feature and the code is quite similar.

pdurbin:
@GPortas at some point we should probably test the SPA against a dataset with lots of versions. We should have datasets like this on the [performance cluster](https://guides.dataverse.org/en/6.4/qa/performance-tests.html).

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -467,11 +473,12 @@ public static JsonObjectBuilder json(DatasetVersion dsv, List<String> anonymized
...
-        );
+        if(includeMetadataBlocks) {
+            bld.add("metadataBlocks", (anonymizedFieldTypeNamesList != null) ?
+                    jsonByBlocks(dsv.getDatasetFields(), anonymizedFieldTypeNamesList)
+                    : jsonByBlocks(dsv.getDatasetFields())
+            );
+        }
         if(returnOwners){
             bld.add("isPartOf", getOwnersFromDvObject(dataset));
         }
```
pdurbin, line 484: I'm just going to add a comment here on the last line of code.

@johannes-darms how do you feel about adding an API test? It might be nice to see this new boolean as an option in UtilIT methods. Plus, it will give confidence that the code is working properly. 😄 I can help with REST Assured assertions, if need be!

doc/release-notes/10171-exlude-metadatablocks.md
```diff
@@ -0,0 +1,4 @@
+Extension of API `{id}/versions` and `{id}/versions/{versionId}` with an optional ``excludeMetadataBlocks`` parameter,
+that specifies whether the metadataBlocks should be listed in the output. It defaults to ``false``, preserving backward
+compatibility. (Note that for a dataset with a large number of versions and/or metadataBlocks having the metadata blocks
+included can dramatically increase the volume of the output).
```
pdurbin, line 4:
```suggestion
included can dramatically increase the volume of the output). See also [the guides](https://dataverse-guide--10778.org.readthedocs.build/en/10778/api/native-api.html#list-versions-of-a-dataset), #10778, and #10171.
```

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1201,6 +1201,8 @@ It returns a list of versions with their metadata, and file list:
 
 The optional ``excludeFiles`` parameter specifies whether the files should be listed in the output. It defaults to ``true``, preserving backward compatibility. (Note that for a dataset with a large number of versions and/or files having the files included can dramatically increase the volume of the output). A separate ``/files`` API can be used for listing the files, or a subset thereof in a given version. 
 
+The optional ``excludeMetadataBlocks`` parameter specifies whether the metadataBlocks should be listed in the output. It defaults to ``false``, preserving backward compatibility. (Note that for a dataset with a large number of versions and/or metadataBlocks having the metadata blocks included can dramatically increase the volume of the output).
```
pdurbin, line 1204:
```suggestion
The optional ``excludeMetadataBlocks`` parameter specifies whether the metadata blocks should be listed in the output. It defaults to ``false``, preserving backward compatibility. (Note that for a dataset with a large number of versions and/or metadata blocks, having the metadata blocks included can dramatically increase the volume of the output).
```

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1225,6 +1227,12 @@ The fully expanded example above (without environment variables) looks like this
 
 The optional ``excludeFiles`` parameter specifies whether the files should be listed in the output (defaults to ``true``). Note that a separate ``/files`` API can be used for listing the files, or a subset thereof in a given version. 
 
+.. code-block:: bash
+
+  curl "https://demo.dataverse.org/api/datasets/24/versions/1.0?excludeMetadataBlocks=false"
+
+The optional ``excludeMetadataBlocks`` parameter specifies whether the metadataBlocks should be listed in the output (defaults to ``false``).
```
pdurbin, line 1234:
```suggestion
The optional ``excludeMetadataBlocks`` parameter specifies whether the metadata blocks should be listed in the output (defaults to ``false``).
```

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -414,15 +414,16 @@ public Response useDefaultCitationDate(@Context ContainerRequestContext crc, @Pa
     @GET
```
pdurbin, line 414: I'm just going to leave this comment on the first line of code...

Can you please merge the latest from develop?

pdurbin (commented):
@johannes-darms overall, looks good. I left some comments. Thanks.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: @johannes-darms me again. We just released 6.5. Can you please merge the latest from develop? Thanks.

johannes-darms: done

johannes-darms:
> @johannes-darms can you please merge the latest from develop? We need this anyway and it will trigger a Jenkins run, which is failing. Also, please consider the suggestions I made in my review. Thanks.

Sorry for the delay. I've merged, adapted the documentation and wrote a simple test. If you need or want more tests I'm happy to write them.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
pdurbin: Tests have been added! Thanks!

sekmiller:
@johannes-darms, sorry to do this to you again, but we've recently released version 6.5, so we need you to get the latest from Dev into this branch. Thanks for your patience.

johannes-darms:
Sadly, the Jenkins build fails and I cannot access the output. Could you forward the information @pdurbin ?

sekmiller:
@johannes-darms Thanks again for your patience.

pdurbin:
@johannes-darms I know this PR has already been merged but to answer your question, the latest API test run passed: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10778/8/testReport/

johannes-darms:
> @johannes-darms I know this PR has already been merged but to answer your question, the latest API test run passed: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10778/8/testReport/

Thanks for sharing but I cannot access the jenkins instance.

pdurbin:
Yes, I know, sorry, that link was a "note to self".

Related:

- #9916

PR #10781, opened 2024-08-19
Improved handling of Globus uploads (experimental async framework)

### LINKED ISSUE #10623: Globus integration: further improvements/refactoring
A placeholder issue for keeping track of next steps in improving Globus support after the conclusion of the initial integration effort. 

At the moment I am aware of a couple of areas where improvements can be made. This is all in the old code that was originally created by the Scholars Portal. While these are not bugs or something that's broken outright, there is a possibility that this refactoring can make the workflow more reliable, especially on a busy prod. server. 

1. After the Globus service receives the confirmation of the completion of the Globus upload, in order to add the file(s) to the dataset it makes an actual `/addFiles` API call to itself, via curl. Once again, this is working in the current implementation and we are not aware of any real problems because of this. But it just would be much cleaner for the service to call the appropriate commands/methods directly. 
2. This one is potentially more important, but may also take more work to implement. As implemented, the asynchronous process of waiting for the Globus upload to complete relies on a service method continuously checking on its status remotely, in a loop (with a sleep interval) for the duration of the process. If nothing else, this means that an upload in progress is not going to survive an application restart before its completion (i.e., the file will end up successfully deposited into the Globus collection, but will not be added to the dataset on the Dataverse side). We need to keep in mind that some of the real life big data uploads will be taking a very long time, days etc., to complete (which is the whole point of Globus etc. etc.). There's also some _anecdotal_ evidence of that looping process having died on the application side while one of the OMAMA data files was being uploaded in prod., even without any restarts involved (possibly on account of something unexpected during one of the intermediate Globus API calls - ?). So, it would make sense to refactor this part as to handle it truly asynchronously. Perhaps by storing the state info about the ongoing uploads in a database table, then have a timer job check on all such known uploads every N seconds. 

### Related
- https://github.com/IQSS/dataverse-pm/issues/268

landreev:

It solves the 2 problems described in the linked issue: 

- An  api call to `/addFiles` via curl has been replaced with calling the corresponding method in AddReplaceFileHelper. In addition to the problems with the curl implementation already described in the issue, the old code was not making any attempt to parse the output of the api; so if `/addFiles` dropped one or more files from the submitted entries, for whatever reason, the Globus service method was still registering that as a complete success. The new code checks the results in a more careful way. 
- A new implementation of the upload transfer polling solves the issue where the Dataverse instance needs to stay up and the same service method needs to loop continuously for the duration of the transfer. The new framework relies on a schedule-run execution of the polling calls and saving the state of the ongoing tasks in the database. This gives an extra advantage of giving an admin a relatively easy way to add the files to the user's dataset after the fact in a scenario where the remote Globus transfer succeeded, but the files failed to be added to the dataset automatically, for whatever reason. However, due to the experimental nature of this new framework, I'm leaving it not enabled by default. Pending gaining more evidence on how it holds up in a prod.-like environment. 

The handling of diagnostics and notifications have generally been improved in the PR. Among other things, there are now explicit failure notifications that are sent to the users when uploads fail completely. The existing implementation only logs such events without notifying the user. 

- Closes #10623

Note that I don't use the EJB timers for the scheduled execution of the task polling. I used the `jakarta.enterprise.concurrent.ManagedScheduledExecutorService` instead which, from what I understand, is a preferred way now.

In order to test this PR, and other Globus-related things going forward, I added a Globus setup to dataverse-internal yesterday (Globus-enabled storage configuration plus the upload/download web app). This volume: 
Screen Shot 2024-09-05 at 4 11 45 PM

Is connected to remote storage at NESE that was set up for us specifically for testing uploads and downloads. This setup is identical to what was previously configured on demo.dataverse.org. I haven't tested it much myself though. 

The instructions on how to upload data via Globus in Datataverse can be found here: https://github.com/IQSS/dataverse.harvard.edu/blob/master/doc/globus/upload.md.

**Additional documentation**:

Preview at https://dataverse-guide--10781.org.readthedocs.build/en/10781/installation/config.html#dataverse-globus-taskmonitoringserver


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
```diff
@@ -102,7 +103,10 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         }
         
         Dataset theDataset = getDataset();        
-        ctxt.permissions().checkUpdateDatasetVersionLock(theDataset, getRequest(), this);
+        //ctxt.permissions().checkUpdateDatasetVersionLock(theDataset, getRequest(), this);
+        // this is an experiment (probably temporary)
```
landreev, line 107: Yes, I meant to reverse (and re-test) this "experiment". The idea was to do a fresh transaction database lookup for any locks that may have been added since the dataset instance was instantiated... I don't think it's needed after all.

src/main/java/edu/harvard/iq/dataverse/globus/TaskMonitoringServiceBean.java
```diff
@@ -0,0 +1,134 @@
...
+    GlobusServiceBean globusService;
+    
+    private static final SimpleDateFormat logFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss");
+    
+    @PostConstruct
+    public void init() {
+        if (systemConfig.isGlobusTaskMonitoringServer()) {
+            logger.info("Starting Globus task monitoring service");
+            int pollingInterval = SystemConfig.getIntLimitFromStringOrDefault(
+                settingsSvc.getValueForKey(SettingsServiceBean.Key.GlobusPollingInterval), 60);
```
qqmyers, line 58: Longer default maybe, even 10 times longer (or more). I can see cases where someone may want to set it to a minute (e.g. testing) but for normal hours-long transfers, seems like wasted effort.

src/main/java/edu/harvard/iq/dataverse/ingest/IngestServiceBean.java
```diff
@@ -345,12 +345,13 @@ public List<DataFile> saveAndAddFilesToDataset(DatasetVersion version,
                         StorageIO<DvObject> dataAccess = DataAccess.getStorageIO(dataFile);
                         //Populate metadata
                         dataAccess.open(DataAccessOption.READ_ACCESS);
-                        
+                        // (this will make a remote call to check if the file exists 
+                        // and obtain its size)
```
qqmyers, line 349: Technically that now happens in the open call above where  there's 
```
            if (isReadAccess) {
                this.setSize(retrieveSizeFromMedia());
```
or similar.

src/main/java/edu/harvard/iq/dataverse/timer/DataverseTimerServiceBean.java
```diff
@@ -120,13 +120,13 @@ public void handleTimeout(jakarta.ejb.Timer timer) {
...
         try {
-            logger.log(Level.INFO,"Handling timeout on " + InetAddress.getLocalHost().getCanonicalHostName());
+            logger.log(Level.FINE,"Handling timeout on " + InetAddress.getLocalHost().getCanonicalHostName());
         } catch (UnknownHostException ex) {
             Logger.getLogger(DataverseTimerServiceBean.class.getName()).log(Level.SEVERE, null, ex);
         }
         
         if (timer.getInfo() instanceof MotherTimerInfo) {
-            logger.info("Behold! I am the Master Timer, king of all timers! I'm here to create all the lesser timers!");
+            logger.fine("Behold! I am the Master Timer, king of all timers! I'm here to create all the lesser timers!");
```
qqmyers, line 129: I'm going to miss these :-(

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
```diff
@@ -82,6 +82,7 @@ public class SystemConfig {
     private String buildNumber = null;
     
     private static final String JVM_TIMER_SERVER_OPTION = "dataverse.timerServer";
+    private static final String JVM_GLOBUS_TASK_MONITORING_OPTION = "dataverse.globus.taskMonitoringServer";
```
qqmyers, line 85: Can this just be a  JvmSettings entry like GLOBUS_CACHE_MAXAGE? (maybe both should be in a Globus namespace - the cache age is in dataverse.files right now).

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -253,11 +253,16 @@ notification.mail.import.filesystem=Dataset {2} ({0}/dataset.xhtml?persistentId=
 notification.mail.globus.upload.completed=Globus transfer to Dataset  <a href="{0}/dataset.xhtml?persistentId={1}" title="{2}">{2}</a> was successful. File(s) have been uploaded and verified. <br><br> {3}<br>
 notification.mail.globus.download.completed=Globus transfer of file(s) from the dataset  <a href="{0}/dataset.xhtml?persistentId={1}" title="{2}">{2}</a> was successful. <br><br> {3}<br>
 notification.mail.globus.upload.completedWithErrors=Globus transfer to Dataset  <a href="{0}/dataset.xhtml?persistentId={1}" title="{2}">{2}</a> is complete with errors. <br><br> {3}<br>
+notification.mail.globus.upload.failedRemotely=Remote data transfer between Globus collections for Dataset <a href="{0}/dataset.xhtml?persistentId={1}" title="{2}">{2}</a> failed, reported via Globus API. <br><br> {3}<br>
```
qqmyers, line 256: between Globus endpoints rather than collections?

qqmyers (changes requested):
Nice! Looks good to me. I made some comments/responded to notes in the code, but other than any open @todos and some minor cleanup, it looks ready for QA to me. Goodbye to exec-ing curl!

I'll put request changes to avoid it switching columns until @landreev  is ready, but probably no need for another review of changes.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
landreev: I thought about that, but, I chose to revert the command to its previous state. The Globus code should continue working at least as well as before. And I am hesitant to change the method in the PermissionServiceBean, for fear of some exotic scenario where something that's working now is going to break.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4009,6 +4009,7 @@ public Response addGlobusFilesToDataset(@Context ContainerRequestContext crc,
         logger.info(" ====  (api addGlobusFilesToDataset) jsonData   ====== " + jsonData);
 
         if (!systemConfig.isHTTPUpload()) {
+            // @todo why isHTTPUpload()? - shouldn't it be checking isGlobusUpload() here?
```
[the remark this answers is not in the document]
landreev: Ok, fixed. Maybe it wasn't really necessary, I guess, since the API methods that need to be called first had proper globus availability checks. But then perhaps there is a use case where an admin would want to disable http and only have globus uploads?

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -634,25 +668,35 @@ private String getGlobusDownloadScript(Dataset dataset, ApiToken apiToken, List<
...
-        boolean fileHandlerSuceeded;
+        boolean fileHandlerSuceeded = false;
         try {
             fileHandler = new FileHandler(logFileName);
             globusLogger.setUseParentHandlers(false);
             fileHandlerSuceeded = true;
         } catch (IOException | SecurityException ex) {
             Logger.getLogger(DatasetServiceBean.class.getName()).log(Level.SEVERE, null, ex);
-            return;
+            //return; // @todo I don't think we need to return here?
```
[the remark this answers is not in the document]
landreev: Yes, the code further down looked like it was written with that in mind, to just proceed using the main logger. Removing the "@todo".

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -676,28 +722,111 @@ public void globusUpload(JsonObject jsonData, ApiToken token, Dataset dataset, S
...
         // GlobusUpload lock
         // Keeping a lock through the add datafiles API call avoids a conflicting edit
-        // and keeps any open dataset page refreshing until the datafile appears
-        if (!(taskStatus.startsWith("FAILED") || taskStatus.startsWith("INACTIVE"))) {
-            datasetSvc.addDatasetLock(dataset,
-                    new DatasetLock(DatasetLock.Reason.EditInProgress, authUser, "Completing Globus Upload"));
+        // and keeps any open dataset page refreshing until the datafile appears.
+        
+        // @todo is it necessary? what is wrong exactly with keeping the Globus 
+        // lock on for the duration of the process?
```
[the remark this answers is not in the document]
landreev: Sure, makes sense. Removing the comment.

src/main/java/edu/harvard/iq/dataverse/timer/DataverseTimerServiceBean.java
landreev: You can still have it - at a low price of one FINE logging setting.

src/main/java/edu/harvard/iq/dataverse/util/SystemConfig.java
landreev: Yeah, the first part is a no-brainer - moved to JvmSettings. As for creating a new globus namespace for both of these settings, I didn't feel as comfortably messing with an existing setting, so I have them both under dataverse.files, if that's ok.

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -1281,5 +1431,59 @@ public void writeGuestbookAndStartTransfer(GuestbookResponse guestbookResponse,
...
+        return em.createNamedQuery("ExternalFileUploadInProgress.findByTaskId").setParameter("taskId", taskId).getResultList();    
+    }
+    
+    public void processCompletedTask(GlobusTaskInProgress globusTask, boolean taskSuccess, String taskStatus, Logger taskLogger) {
+        String ruleId = globusTask.getRuleId();
+        Dataset dataset = globusTask.getDataset();
+        AuthenticatedUser authUser = authSvc.lookupUser(globusTask.getApiToken());
+        if (authUser == null) {
+            // @todo log error message; do nothing
+            // (the fields in GlobusTaskInProgress are not nullable though - ?)
```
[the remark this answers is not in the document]
landreev: That was a great suggestion - idk what I was thinking. (and with this change, there's no need to pass the token to the globusUpload() method at all, so I dropped it)

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -4034,6 +4035,8 @@ public Response addGlobusFilesToDataset(@Context ContainerRequestContext crc,
             return wr.getResponse();
         }
         
+        // @todo check if the dataset is already locked!
```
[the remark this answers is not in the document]
landreev: You meant "of a given type", right? The addDatasetLock() already has a check for an existing lock of the same type (then it will just return the existing lock instead of creating a new one). 
As for implementing this as a hard constraint - are we positive we'll never want multiple locks of the same type - some workflows maybe? 
I addressed this specific situation with a separate lock check. We don't check for locks consistently in our APIs, I'm assuming under the assumption that it will be done when the the relevant commands are executed. But in this specific case, there is a potential for doing a huge amount of work before that UpdateDatasetVersionCommand is called in the end.

cmbz:
@landreev could you please recommend a size for this PR thanks!

landreev:
I resolved the merge conflicts yesterday. Also added some info under "How to Test".

src/main/java/edu/harvard/iq/dataverse/timer/DataverseTimerServiceBean.java
pdurbin: Ha! It's like an old friend at this point. 😄

pdurbin:
@qqmyers good catch. I put in a commit to quiet down that yappy dog: 7b6f81e

I also made a small doc tweak: 2baf62e

Then I tested on internal. With the new flag off I was able to upload file. ✅ 

With the flag on, Globus is telling me "transfer complete" but the file is not in the dataset and the dataset has the "Globus Transfer in Progress" lock. ❌ 

@landreev is going to take a look.

pdurbin:
Turned out to be a doc problem fixed now. Works great! Especially for an experimental feature. 😜 

Also, API tests are passing as of the last code change: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-10781/7/testReport/

Merging.

PR #10794, opened 2024-08-22
IQSS/10793 Improve Handling of Parallel Edit/Publish errors

### LINKED ISSUE #10793: Parallel Edits/Publishing result in unclear errors/abandoned lock
Parallel UI edits to a dataset (i.e. from two browser pages opened to the same dataset, or separate browsers/users, etc.) result in the second one showing an error message noting the edit failed without any useful explanation. The error also returns the user to the main dataset page from the metadata edit pane, causing them to lose any information they had just types.
Similarly, if a user tries to publish a dataset version that has been updated elsewhere since they opened the dataset page, a message saying the publish method failed is shown (with no clear explanation as to why or how to recover). However, in this case the message is only shown briefly as a publish lock is added to the dataset and the error message disappears the first time the page refreshes to see if the lock has been removed. Further, since the Publish  attempt failed, the lock remains until removed by an admin.

As discussed w.r.t. managing this in the SPA, there are a range of improvements that could be made, from simply creating clearer error messages, with some guidance as to how to proceed, to trying to identify what has changed and allowing the user to proceed if they agree the changes are consistent with their own edits/that publishing with those changes is OK.

Since these errors are being seen in the wild now (and are possibly a source of mystery dataset locks), I'll create a PR that will provide better error messages in these two cases and avoid leaving the publish lock on the dataset.

* When does this issue occur?
Edit/Edit or Edit/Publish done in parallel on two copies of the dataset page

6.3/develop

**Screenshots:**

*

qqmyers:
**What this PR does / why we need it**: This PR catches the OptimisticLockExceptions caused by an edit or publish attempt on a dataset that has been changed since the dataset page view was created and provides clearer error messages indicating what happened and how to recover. In addition to the error messages:
* for the Edit Metadata case, this PR causes the display to stay on the metadata edit/terms edit page to allow users to copy their text before it disappears.
* for the Publish case, the Publish lock is removed if an OptimisticLock occurs, so the user is able to refresh the page and try again.

Closes #10793

**Special notes for your reviewer**: I think the double refresh on the terms page was added when the check for file access requests allowed or terms of access required logic was added in #8308 where both the original commandButtons and the new remoteCommand for saving terms do update=":datasetForm,:messagePanel"

**Suggestions on how to test this**: Pretty easy to load two windows for the same dataset, make an edit in one and save and then make an edit to metadata/terms/etc. or try to publish in the other. Since this PR mostly updates the error messages, I hope that other changes on the page work ~as before/hopefully better with the new error messages, but I haven't tried all the buttons.

**Additional documentation**: I only saw one line in the user guide that says you can edit metadata (at the end of the Add New Dataset section) and it seems like this is too much of a corner case to add to that. If there was a whole edit section telling you how to do that, adding a sentence to say you should avoid editing in two windows might make more sense. Or maybe we need a FAQ?


pdurbin (approved):
I didn't run the code but it looks good to me. Catching more exceptions. New messaging about what's going on. Approved.

PR #10797, opened 2024-08-23
Alternate Fix for Curate command failures

qqmyers:
**What this PR does / why we need it**: This PR extends #10786, removing use of other commands within this one to avoid the possibility of partial updates (as was seen prior to both PRs when DataCite was down or when terms (and not metadata) were changed.

Closes #

**Special notes for your reviewer**: Have yet to see the updated IT test run (here or in #10786)

FWIW: In removing the em.merges after adding the initial merge of the dataset, I used the em.contains() test to verify that the initial merge of the dataset also merged the versions, etc.  - not sure I was aware of that call before this, but I think it might be useful in deciding if merges are really needed (as it looks like merging a version gets expensive with many files).

Update a terms field (e.g. terms of use (with no license/custom terms), save, and then, as superuser hit the publish/update-current-version option. It should work now.

Similarly, I think you can misconfigure DataCite to simulate a failure of their service, try any update and publish/update-current-version. The command should succeed and there should be a warning line in the log.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
@@ -1,7 +1,6 @@
 package edu.harvard.iq.dataverse.engine.command.impl;
 
 import edu.harvard.iq.dataverse.authorization.Permission;
-import edu.harvard.iq.dataverse.datavariable.VarGroup;
 import edu.harvard.iq.dataverse.engine.command.CommandContext;
 import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
 import edu.harvard.iq.dataverse.engine.command.RequiredPermissions;
@@ -20,7 +19,6 @@ import edu.harvard.iq.dataverse.FileMetadata;
 import edu.harvard.iq.dataverse.DataFileCategory;
 import edu.harvard.iq.dataverse.DatasetVersionDifference;
 
-import java.util.Collection;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
```diff
@@ -155,45 +164,72 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
-            ExportService instance = ExportService.getInstance();
-            instance.exportAllFormats(getDataset());
-        } catch (ExportException ex) {
-            // Just like with indexing, a failure to export is not a fatal condition.
-            logger.log(Level.WARNING, "Curate Published DatasetVersion: exception while exporting metadata files:{0}", ex.getMessage());
+            ctxt.engine().submit(
+                    new UpdateDvObjectPIDMetadataCommand(savedDataset, getRequest()));
+        } catch (CommandException ex) {
+            // Make this non-fatal? This can be corrected by running the update PID API
+            // later, but who will look in the log?
```
landreev, line 197: If we were willing to make it non-fatal, this command could be moved to `onSuccess()` as well. 
I understand the rationale of "who will look in the log? [and therefore nobody will ever get to update the stale PID metadata]". I'd say this should be a function of how confident you are that with these changes rolling back the transaction will never be a problem again. If there's any trace of doubt, a stale DataCite record is way better than lost metadata. 

The PR looks great overall.

landreev (approved):
Special thanks for adding an extended test.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/CuratePublishedDatasetVersionCommand.java
qqmyers: I guess it's rare enough, and these Curate changes are supposed to be minor enough, that catching any exception and logging is OK/maybe better than having the overall update fail.

I'll update the comment text though - it is non-fatal now, but it could cause the Curate to not succeed if the try/catch is removed.

landreev: Doh, sorry.

pdurbin:
Related:

- #10880

PR #10798, opened 2024-08-26
Extend Direct Upload docs for async stream uploads

JR-1991:

This pull request expands the documentation to provide instructions on handling asynchronous streaming files or chunks to S3. The updated docs emphasize the necessity of the `Content-Length` header and warn that failing to include it will lead to a `501: Not implemented` error. 

I encountered this error in the past but couldn’t identify the cause due to the vague `501` message. I eventually discovered the underlying issue through a StackOverflow discussion. Hence, adding this information to the documentation could help prevent others from experiencing similar frustration.

Extended Direct Upload documentation to clarify required headers for asynchronous streaming of files to S3.

Preview at https://dataverse-guide--10798.org.readthedocs.build/en/10798/developers/s3-direct-upload-api.html

### DIFF
--- a/doc/sphinx-guides/source/developers/s3-direct-upload-api.rst
+++ b/doc/sphinx-guides/source/developers/s3-direct-upload-api.rst
@@ -102,7 +102,24 @@ If the client is unable to complete the multipart upload, it should call the abo
   
     curl -X DELETE "$SERVER_URL/api/datasets/mpload?..."
    
-  
+
+.. _direct-async-upload:
+
+Asynchronous Direct Upload
+--------------------------
+
+When uploading files or chunks asynchronously to S3 via `PUT`, ensure that the `Content-Length` header is set to the size of the file or chunk. If `Content-Length` is not specified, the server will return a `501 Not Implemented` error, causing the upload to fail.
+
+Example `curl` command:
+
+.. code-block:: bash
+
+   curl -X PUT "<supplied url>" \
+        -H "Content-Length: 5242880" \
+        -T <filename>
+
+Replace ``"<supplied url>"`` with your actual pre-signed S3 URL and ``<filename>`` with the file or chunk you are uploading. The``Content-Length`` value should match the size of the file or chunk in bytes. Please note that other required headers mentioned above should also be included in the request. These have been omitted from the example for clarity.
+
 .. _direct-add-to-dataset-api:
 
 Adding the Uploaded File to the Dataset
--- a/doc/sphinx-guides/source/developers/s3-direct-upload-api.rst
+++ b/doc/sphinx-guides/source/developers/s3-direct-upload-api.rst
@@ -73,7 +73,7 @@ In the example responses above, the URLs, which are very long, have been omitted
 
 .. _direct-upload-to-s3:
 
-Upload files to S3
+Upload Files to S3
 ------------------
 
 The client must then use the URL(s) to PUT the file, or if the file is larger than the specified partSize, parts of the file.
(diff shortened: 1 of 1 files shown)


doc/sphinx-guides/source/developers/s3-direct-upload-api.rst
```diff
@@ -102,7 +102,25 @@ If the client is unable to complete the multipart upload, it should call the abo
   
     curl -X DELETE "$SERVER_URL/api/datasets/mpload?..."
    
-  
+
+.. _direct-async-upload:
+
+Asynchronous Direct Upload
```
qqmyers, line 108: I'm confused by this being a section and having a curl example. This isn't really a separate step or kind of upload (as far as S3 is concerned) is it? And the curl command above works? (Or does it fail for some S3 services?)

If I understand, you're seeing this from python - could it be that w/o setting the size in python, when using the async library, you switch to chunked encoding which requires HTTP 1.1 which may not be implemented somewhere (what S3 service are you testing with?)? (I'm not sure if this is what you're seeing - trying to Google quickly. In Java, the DVUploader uses Apache's InputStreamEntity(stream, length) and I've seen an issue where someone tried -1 as a length and the answer reported that -1 switches to chunked mode which requires HTTP 1.1 and they saw a not implemented error)

If it's something like this, I'd suggest adding this as a Note: on the file upload step (which starts at line 74 but doesn't have a header right now for some reason) and either noting that for some servers a size header is required (if the original curl isn't working) or noting that, when writing code, adding the size is required/may be required depending on the S3 service (if it is about HTTP1.1 support) or just noting that if you see a 501 error when developing code to implement this, assuring that a size header is sent is one possible way to resolve it (if we really don't know the underlying cause).  I'd suggest only including a curl example if there are cases where the original curl fails.

JR-1991: Thank you for the quick review, @qqmyers. 

I’ve encountered this issue not only in Python (httpx, aiohttp/aiofiles) but also in Rust (reqwest), and I suspect this might vary between libraries. Hence, the idea was to add it as a note or section in the docs to provide troubleshooting for those who may experience the same issue.

> Which S3 service are you testing with?

I’ve run into this issue while testing `python-dvuploader` and `dataverse-rs` on demo.dataverse.org. It’s also been reported in another installation as detailed in https://github.com/gdcc/python-dvuploader/issues/7. However, I’m not certain about the specific S3 service involved. Interestingly, using the `localstack` service on a local instance didn’t trigger the issue, which can create a misleading sense of security since tests are not failing.

> If it’s something like this, I’d suggest adding this as a Note: on the file upload step […]

I agree, that sounds like a more appropriate solution. I initially added it as a section because I thought developers integrating the Direct Upload API might be the primary audience, and I wanted to ensure it stands out. I would condense the current section into a note without the `curl` example.

JR-1991: Something along these lines? 

```
.. note::
    If you encounter an ``HTTP 501 Not Implemented`` error, 
    ensure the `Content-Length` header is correctly set to 
    the file or chunk size. This issue may arise when streaming 
    files or chunks asynchronously to S3 via ``PUT`` requests, 
    particularly if the library or tool you're using doesn't
    set the ``Content-Length`` header automatically.
```

Ignore the weird newlines. I just tried to fit it into this comment 😁

qqmyers: Sounds great - could you also add the ~missing heading before line 74 as well - just "Upload Files to S3" or similar.  (Weird that I said three steps and step 2 doesn't have a header...)

JR-1991: Sounds good! For the missing section header, should I add `.. _direct-upload-to-s3:` before the "Upload Files to S3" section to ensure it appears in the TOC?

JR-1991: @qqmyers - I have updated the PR

qqmyers: Looks good to me. Thanks.

pdurbin (approved):
Looks great. Merging. Thanks for the PR, @JR-1991 and for the review, @qqmyers!

PR #10762, opened 2024-08-08
Support the full ISO 639-3 list of languages #10578

stevenwinship:
**What this PR does / why we need it**: Some codes are still not managed. In the cases encountered, frm (Medieval French) and fro (Old French).

**Which issue(s) this PR closes**: Figure out whether, or how to support the extended ISO 639-3 list of languages #8578

Closes #8578 
Closes #7757

**Special notes for your reviewer**:

**Suggestions on how to test this**: start with preexisting datasets created by older version of dataverse
Follow Update the Citation metadata block from release notes. Make sure the languages are the same and that some new languages can be added

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**: Yes. to be included in this PR

**Additional documentation**: The ISO data was downloaded from 
https://iso639-3.sil.org/code_tables/download_tables#Complete%20Code%20Tables:~:text=iso%2D639%2D3_Code_Tables_20240415.zip

PR #10801, opened 2024-08-26
a proposed refinement (simplification) of some of the "main" language names in the new list

landreev:
See my note in #10762 for details.


qqmyers (approved):
These, and the logic on how to simplify, look fine to me. If any concerns arise, i.e. someone has a reason the main label needs to be different, include more variants, I think they could be handled in new issues.

PR #10802, opened 2024-08-26
Fix addDataverse expected request body structure

### LINKED ISSUE #10800: Fix addDataverse expected request body structure

The docs specify that the request JSON for a new dataverse supports an optional "metadataBlocks" object, with the following supported sub-objects:

- metadataBlockNames
- inputLevels
- facetIds

But the API expects facetIds to come from outside the "metadataBlocks" object. 

We should update the API to expect facetIds inside the "metadataBlocks" object. 

* When does this issue occur?

Using addDataverse API

API users

Y

GPortas:

The docs specify that the request JSON for a new dataverse supports an optional "metadataBlocks" object, with the following supported sub-objects:

- metadataBlockNames
- inputLevels
- facetIds

But the API expects facetIds to come from outside the "metadataBlocks" object. 

I am fixing the API to expect facetIds inside the "metadataBlocks" object. 

- Closes #10800 

Taking a look at the updated IT tests and seeing they pass should be enough, but we can additionally test the endpoint applying custom facetIds as specified in `dataverse-complete-optional-params.json`.

None.

Yes. Attached.


sekmiller (approved):

doc/release-notes/10800-add-dataverse-request-json-fix.md
```diff
@@ -0,0 +1 @@
+Fixed the "addDataverse" API endpoint (/dataverses/{id} POST) expected request JSON structure to parse facetIds as described in the docs.
```
pdurbin, line 1: @GPortas heads up that I didn't add this to the 6.4 release notes because it pertains to a feature in #10644 that hasn't shipped yet (both will be in 6.4). From the perspective of the release notes, we don't need to mention bugs that never made it into a release.

Issue #3616, opened 2017-02-01
`image_url` from Search API results no longer yields a downloadable image


pdurbin:
The Search API has a documented feature wherein a field called `image_url` appears in search results but this no longer works as of some performance optimizations that where made in the 4.2 timeframe regarding thumbnails. I discovered this while poking around in the code for #3559 and @landreev encouraged me to open an issue.

An example is `"image_url":"https://demo.dataverse.org/api/access/dsCardImage/2"` from http://guides.dataverse.org/en/4.6/api/search.html . Instead of an image you always get `204 No Content`.

No one has complained that this `image_url` feature doesn't work any more. The only place where I'm aware that it's used is the proof of concept Android app I I knocked together a while ago: https://github.com/IQSS/dataverse-android . Here's a screenshots of how they where used:

![mkwrzkj](https://cloud.githubusercontent.com/assets/21006/22518333/381a1c5a-e87a-11e6-88e7-28d460c964fd.png)

pdurbin:
I'll open this back up when I feel like there's demand for it. Closing.

Issue #9788, opened 2023-08-17
"Dataverse" responses in serach and native API


johannes-darms:
### What steps does it take to reproduce the issue?/ What happens?

1. Obtain the information of a single dataverse via `curl "https://demo.dataverse.org/api/dataverses/root" | jq`  Response: [1]
2. Search for root dataverse via `curl "https://demo.dataverse.org/api/search?type=dataverse&q=root" | jq`  Response: [2]
3. Compare the responses [1] and [2]. I except the search response to be a subset of the details provided by the native API to obtain a single dataverse. However, "published_at" is only available via search. "alias" is called "identifier" within the search response, which is a bit confusing especially since there is a property "id" with the numerical database id of the dataverse.

[1]
```
{
  "status": "OK",
  "data": {
    "id": 2067773,
    "alias": "DVC_2023",
    "name": "Dataverse Community Meeting, 2023",
    "affiliation": "Harvard University",
    "dataverseContacts": [
      {
        "displayOrder": 0,
        "contactEmail": "sbarbosa@hmdc.harvard.edu"
      }
    ],
    "permissionRoot": true,
    "description": "This collection supports the Dataverse Community Meeting, 2023 training website.",
    "dataverseType": "TEACHING_COURSES",
    "ownerId": 1,
    "creationDate": "2023-06-05T07:40:29Z",
    "theme": {
      "id": 716,
      "logo": "NE_Revolution_Logos_Seal_Primary_RGB.jpg",
      "tagline": "I love football!!",
      "linkUrl": "https://www.boston.com/",
      "linkColor": "000d08",
      "textColor": "888888",
      "backgroundColor": "f7b100",
      "logoBackgroundColor": "08e7f7"
    }
  }
}
```
[2]
```
{
  "status": "OK",
  "data": {
    "q": "root",
    "total_count": 3,
    "start": 0,
    "spelling_alternatives": {},
    "items": [
      {
        "name": "Logo Bug Root",
        "type": "dataverse",
        "url": "https://demo.dataverse.org/dataverse/logobugroot",
        "identifier": "logobugroot",
        "published_at": "2021-11-23T11:30:06Z"
      },
      {
        "name": "Tania Schlatter Dataverse in Root",
        "type": "dataverse",
        "url": "https://demo.dataverse.org/dataverse/TSD_Root",
        "identifier": "TSD_Root",
        "published_at": "2021-11-16T11:37:10Z"
      },
      {
        "name": "Oscar Moreno Dataverse",
[...]

Issue #9713, opened 2023-07-18
Feature Request/Idea: Clarify/improve creation and publication dates of collections in results of Search API and "View a Dataverse Collection" endpoint


jggautier:
**Overview of the Feature Request**

**What kind of user is the feature intended for?**
(Example users roles: API User, Curator, Depositor, Guest, Superuser, Sysadmin)
API User

**What inspired the request?**
My use of the Search API and Native API to get information about Dataverse collections

**What existing behavior do you want changed?**
Could we review and possibly improve the dates in the results of the two API endpoints?

When items in Search API results are collections (when "type" is "dataverse"), it looks like one date is always included and it's always called "published_at". When the collection is published, that date is the collection's publication date. When the collection is unpublished, that date seems to be the collection's creation date.

This collection is unpublished and what the Search API returns includes a "published_at" date:
<kbd><img width="488" alt="Screenshot 2023-07-18 at 11 40 23 AM" src="https://github.com/IQSS/dataverse/assets/18374574/8d6be8bc-c901-4d0d-95e2-9c02034a7ad5"></kbd>

This makes it harder to determine if and when collections are published and what each collection's creation and publication dates are.

When using the Native APIs ["View a Dataverse Collection" endpoint](https://guides.dataverse.org/en/5.13/api/native-api.html#view-a-dataverse-collection), it looks like one date is always called "creationDate" and it's the date when the collection was created.

This collection is published and what the "View a Dataverse Collection" endpoint returns includes only a createDate:
<kbd><img width="404" alt="Screenshot 2023-07-18 at 12 03 46 PM" src="https://github.com/IQSS/dataverse/assets/18374574/7ee6601f-d2ba-453a-9bbf-4adcffec1267"></kbd>

This makes it harder to determine if collections are published and impossible to know when they were published.

**Any brand new behavior do you want to add to Dataverse?**
Maybe for the Native APIs "View a Dataverse Collection" endpoint, when the collection is published, a publicationDate key could be added. Its presence or absence could be used to figure out if the collection is published and its publication date. Additionally, maybe a key could be added that more explicitly shows the publication state of the collection, something like "isPublished" or publicationState.

Maybe for the Search API, a created_at key could be added that always shows when the collection was created. And the published_at key could be used only to show when published collections were published.

**Any open or closed issues related to this feature request?**
None that I could find.

PR #10811, opened 2024-08-29
Search API results payload extension

### LINKED ISSUE #10810: Extend Search API object type payloads to include necessary information for collection item cards

Currently, the search API does not return all the attributes needed for datasets, files, and collections required to display the collection item cards. The goal is to extend the payload to include the missing properties so that js-dataverse can provide the necessary information to the SPA.

API User

- https://github.com/IQSS/dataverse-client-javascript/pull/181

None

Extend Search API object type payloads to include necessary information for collection item cards

- https://github.com/IQSS/dataverse-client-javascript/pull/181

Yes

GPortas:

Currently, the search API does not return all the attributes needed for files and collections, required to display the file and collection item cards.

I have extended the Search API result payloads to include:

- For the Dataverse entity: "affiliation", "parentDataverseName", "parentDataverseIdentifier" and "image_url" (optional).
- For the DataFile entity: "releaseOrCreateDate" and "image_url" (optional).
- For the Dataset entity: "image_url" (optional).

The schema.xml file for Solr has been updated to include a new field called dvParentAlias for the parentDataverseIdentifier. This field is now indexed in the IndexServiceBean.

The `image_url` field was already included in the SolrSearchResult JSON, but it wasn’t returned by the API because it was appended only after the Solr query was processed in the SearchIncludeFragment of JSF. Now, the field is set in SearchServiceBean, ensuring it is always returned by the API when an image is available.

However, in the `search.rst` documentation, the field was incorrectly listed as being returned by the API, even though it never was since it was only set in JSF.

I have reused the same URL generation logic from JSF. But currently, URL generation differs between datasets and dataverses&files. For datasets, an absolute URL is generated, while for files and dataverses, a Base64 URL is generated. In a future iteration, it would be beneficial to standardize the URL format so that the same type of URL is returned for all entities. I have mentioned this in the docs.

- Closes #10810 

For the next released version, a Solr reindex will be necessary to apply the new schema.xml version. (Mentioned in the release notes)

Create dataverses, datasets and files and test the search endpoint to verify that new params are correctly retrieved.

`curl "http://localhost:8080/api/search?q=*&sort=date&order=desc" -H "X-Dataverse-Key: <YOUR_API_KEY>"
`

No

Yes

**Additional documentation**:
No

Preview docs at https://dataverse-guide--10811.org.readthedocs.build/en/10811/api/search.html

**Related issues**:

- #3616


johannes-darms:
Does this PR also resolved/includes the requested changes of https://github.com/IQSS/dataverse/issues/9788?

GPortas:
> Does this PR also resolved/includes the requested changes of #9788?

@johannes-darms No. This PR adds new fields to the Search API result, without changing existing ones. The scopes are different. 

I think what is proposed in https://github.com/IQSS/dataverse/issues/9788 requires further discussion, in particular for maintaining backward compatibility in the endpoint, since existing fields would be redefined / renamed.

johannes-darms:
> > Does this PR also resolved/includes the requested changes of #9788?
> 
> @johannes-darms No. This PR adds new fields to the Search API result, without changing existing ones. The scopes are different.

Partly the issue list difference between list/search and get, but also the missing items needed to recreate the dataverse/collection cards based on the search response like logo, affiliation and description. Some of these are already implemented in this PR. Could we add the rest in this PR as well? 

> I think what is proposed in #9788 requires further discussion, in particular for maintaining backward compatibility in the endpoint, since existing fields would be redefined / renamed.

Only suggestion two of the problem changes a property, the rest are just additions. Phil has already proposed a solution that would work. I think this is a great PR to fix the problems listed in #9788 and #9713.

GPortas:
> > > Does this PR also resolved/includes the requested changes of #9788?
> > 
> > 
> > @johannes-darms No. This PR adds new fields to the Search API result, without changing existing ones. The scopes are different.
> 
> Partly the issue list difference between list/search and get, but also the missing items needed to recreate the dataverse/collection cards based on the search response like logo, affiliation and description. Some of these are already implemented in this PR. Could we add the rest in this PR as well?
> 
> > I think what is proposed in #9788 requires further discussion, in particular for maintaining backward compatibility in the endpoint, since existing fields would be redefined / renamed.
> 
> Only suggestion two of the problem changes a property, the rest are just additions. Phil has already proposed a solution that would work. I think this is a great PR to fix the problems listed in #9788 and #9713.

@johannes-darms 

How about making it incremental and allowing this PR to go into review as it is now, since it's ready to close #10810 and there's SPA work dependent on this. 

Once it's accepted and merged, we can work on a new spike issue to add the fields you're requesting. Does that sound good to you?

johannes-darms:
There are only two properties missing, a link to the logo and information about the "linked" status. I think the effort to add these in this PR is rather small. I'm not a fan of changing APIs in a short period of time, especially considering that these changes require index adjustment and re-indexing. So I'd rather have one PR than two, but I'm not the one in charge ;). So my option does not matter that much... 

However, to recreate the current visualisation, two properties are missing from the API response. I've attached a screenshot and marked the missing properties with a blue circle. The [related fronted code](https://github.com/IQSS/dataverse-client-javascript/pull/181/files#diff-37b77d53e3159ca38f22429fc9c1dbb77f78830d3b28e62c7dbf91c6ba99891dR12) already has the needed `imageUrl` property, so why don't we add it to the API in this PR? The other property about the `linked`[ status is already index ](https://github.com/IQSS/dataverse/blob/develop/src/main/webapp/search-include-fragment.xhtml#L508), just not exposed.
Screenshot 2024-09-03 at 14 11 45

GPortas:
@johannes-darms 

I am working on adding your requested properties. `imageUrl` was straightforward to expose, as you describe.

I see that the linked icon depends on the `isInTree` boolean flag property from the Solr result. Actually this property is not indexed; instead, it is post-calculated in the UI logic and appended to the `SolrSearchResult`, specifically in `SearchIncludeFragment`, as can be seen here: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java#L1475. 

If we want to expose the flag as it is used by the JSF UI, we need to move the calculation logic out of `SearchIncludeFragment`, potentially to `SearchServiceBean` or `IndexServiceBean`.

What is already indexed and therefore straightforward to expose is `subtreePaths`, but for now I'm not sure if this information is enough to calculate the linked status, as in my tests I can't get the complete subtree starting from the root dataverse, which I believe is what we would need to determine the linked status. This field only returns the first dataverse below root. For instance, if I have this hierarchy: `root > dv1 > dv2 > dv3`, subtreePaths for entity dv2 returns `"/<dv1_id>"`, instead of `"/<dv1_id>/<dv2_id>" `which is what I expect (the entire subtree).

I'm still looking into this, but I think determining the linked status may require exploring a solution that would go beyond the scope of this PR / issue.

johannes-darms:
> I am working on adding your requested properties. `imageUrl` was straightforward to expose, as you describe.

Many thanks! 

> I see that the linked icon depends on the `isInTree` boolean flag property from the Solr result. Actually this property is not indexed; instead, it is post-calculated in the UI logic and appended to the `SolrSearchResult`, specifically in `SearchIncludeFragment`, as can be seen here: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/search/SearchIncludeFragment.java#L1475. [...]
> I'm still looking into this, but I think determining the linked status may require exploring a solution that would go beyond the scope of this PR / issue.

Bummer, I totally agree with you that this refactoring should be a seperated PR.

doc/release-notes/10810-search-api-payload-extensions.md
```diff
@@ -0,0 +1,52 @@
...
+```javascript
+"items": [
+    {
+        ...
+        "image_url": "http://localhost:8080/api/datasets/2/logo"
+        ...
+(etc, etc)
+```
+
+The image_url field was already part of the SolrSearchResult JSON, but it wasn’t returned by the API because it was appended only after the Solr query was executed in the SearchIncludeFragment of JSF. Now, the field is set in SearchServiceBean, ensuring it is always returned by the API when an image is available.
```
pdurbin, line 50:
```suggestion
The image_url field was already part of the SolrSearchResult JSON (and incorrectly appeared in Search API documentation), but it wasn’t returned by the API because it was appended only after the Solr query was executed in the SearchIncludeFragment of JSF. Now, the field is set in SearchServiceBean, ensuring it is always returned by the API when an image is available.
```

doc/sphinx-guides/source/api/search.rst
```diff
@@ -91,16 +103,26 @@ https://demo.dataverse.org/api/search?q=trees
...
+                    "published_at":"2016-05-10T12:57:27Z",
+                    "publicationStatuses": [
+                        "Published"
+                    ],
+                    "affiliation": "Dataverse.org",
+                    "parentDataverseName": "Root",
+                    "parentDataverseIdentifier": "root"
                 },
                 {  
                     "name":"Darwin's Finches",
```
pdurbin, line 128: Datasets can have `identifier_of_dataverse` but not `parentDataverseIdentifier`, right?

doc/sphinx-guides/source/api/search.rst
```diff
@@ -91,16 +103,26 @@ https://demo.dataverse.org/api/search?q=trees
                     "dataset_name": "Dataset One",
                     "dataset_id": "32",
                     "dataset_persistent_id": "doi:10.5072/FK2/XTT5BV",
-                    "dataset_citation":"Spruce, Sabrina, 2016, \"Spruce Goose\", http://dx.doi.org/10.5072/FK2/XTT5BV, Root Dataverse, V1"
+                    "dataset_citation":"Spruce, Sabrina, 2016, \"Spruce Goose\", http://dx.doi.org/10.5072/FK2/XTT5BV, Root Dataverse, V1",
+                    "publicationStatuses": [
+                        "Published"
+                    ],
+                    "releaseOrCreateDate": "2016-05-10T12:53:39Z"
```
pdurbin, line 110: Files do not get `parentDataverseIdentifier`, right?

pdurbin (commented):
Overall, this looks great. I'm asking some questions in this review.

doc/sphinx-guides/source/api/search.rst
```diff
@@ -151,6 +173,8 @@ https://demo.dataverse.org/api/search?q=trees
         }
     }
 
+Note that the image_url field, if exists, will be returned as a regular URL for Datasets, while for Files and Dataverses, it will be returned as a Base64 URL.
```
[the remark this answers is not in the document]
GPortas: Just for keeping track in the PR. I repost the same info I sent to you on Slack.

For the SPA, using an absolute URL is more suitable than Base64, as it allows the browser to automatically cache the resources. However, I believe there might be performance-related reasons for using Base64. For reference, see: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/search/SolrSearchResult.java#L42.

doc/sphinx-guides/source/api/search.rst
GPortas: Yes. parentDataverseIdentifier is only for Dataverses.

doc/sphinx-guides/source/api/search.rst
GPortas: Right! Just for Dataverses

doc/sphinx-guides/source/api/search.rst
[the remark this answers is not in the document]
qqmyers: Browser caching should work for the current UI as well, so I'm doubtful that there's a big performance gain embedding base64.

pdurbin (approved):
I didn't run the code but it seems reasonable and @GPortas addressed my concern about `image_url` by explaining in the guides that it is expected to change in #10831. Tests are passing. Approved!

PR #10813, opened 2024-08-29
IQSS/10812 - Add details to error messages

### LINKED ISSUE #10812: Feature Request: The current error handler hides detailed info that is sent from the underlying class where the error occurs.
In trying to debug a use of the Semantic API, I realized that the detailed error created in the class handling the call (in this case: `Can't append to a single-value field that already has a value: https://dataverse.org/schema/core#dataAccessPlace`) was being replaced by a generic `Bad Request. The API request cannot be completed with the parameters supplied. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.`  in our WebApplicationExceptionHandler class.

Incorporate the problem-specific error message in the response to the user.

Yes

qqmyers:
**What this PR does / why we need it**: This PR incorporates any messages added to a BadRequestException when it is thrown into the final message the API user sees. If there is no message, the error response is generic as before.

Closes #10812

**Special notes for your reviewer**: Feel free to improve the message

**Suggestions on how to test this**: Make a call such as
`curl -H 'X-Dataverse-key:e6affd89-03f5-4b61-8ed6-d9af41228df2' -X PUT -d '{"https://dataverse.org/schema/core#fileTermsOfAccess":{"https://dataverse.org/schema/core#dataAccessPlace":"OTR"}}' -H "Content-type:application/ld+json" -H "Accept:application/ld+json"  http://localhost:8080/api/datasets/10345/metadata` twice on a dataset (once to add the value, and the second time to trigger an error ("Can't append to a single-value field that already has a value: https://dataverse.org/schema/core#dataAccessPlace") - and verify that the test above is part of the API 400 response you get. (FWIW: If you want the call to succeed, add ?replace=true to the call).

**Is there a release notes update needed for this change?**: - It's a change to the "message" included in the error responses for some API calls (and we already have messages that change due to i18n), so I don't think it's needed, but I'd be happy to have a "Error messages for some API calls have more detail" as a release note and/or to have it mentioned in a change log.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/errorhandlers/WebApplicationExceptionHandler.java
@@ -18,6 +18,8 @@ import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.apache.commons.lang3.StringUtils;
+
 /**
  * Catches all types of web application exceptions like NotFoundException, etc etc and handles them properly.
  */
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
I didn't test this but it seems like a reasonable fix. Approved.

PR #10816, opened 2024-08-29
makes auto-select feature for dataset thumbnail optional 

### LINKED ISSUE #10815: Make the auto-select dataset thumbnail feature optional
There is a bug in 6.3 with designated dataset thumbnails that breaks collection pages. I'm not describing it in detail here since there will be a dedicated issue for it. Turning the auto-select feature off does not fully solve this problem, but it will mitigate it significantly, because most newly published datasets will not have the designated thumbnail to cause a problem.

landreev:
(copy-and-paste from linked issue:)
There is a bug in 6.3 with designated dataset thumbnails that breaks collection pages. I'm not describing it in detail here since there will be a dedicated issue for it. Turning the auto-select feature off does not fully solve this problem, but it will mitigate it significantly, because most newly published datasets will not have the designated thumbnail to cause a problem.

Closes #10815

**Additional documentation**:


landreev:
I believe I have a real fix, so I'll close this.

PR #10818, opened 2024-08-29
IQSS/10814 Improve dataset version differencing

### LINKED ISSUE #10814: Feature Request: Improve Dataset version differencing performance
**Overview of the Feature Request** The DatasetVersionDifference class uses some double loops over FileMetadata/DataFiles from the old and new versions which are known to be inefficient (the code has a todo). Making this faster would improve the performance of editing a dataset with many files as well as displaying the versions table on the dataset page.

PR to follow

qqmyers:
**What this PR does / why we need it**: The change in this PR makes differencing during the UpdateDatasetVersionCommand for a dataset with 10K files go from taking ~ 8 seconds to 17 ms. Presumably it improves the Version table on the dataset as well though I haven't tested that at scale.

The PR also includes some cleanup of the terms comparison - basically folding the possible null TermsOfUseAndAccess on the original or new version cases into one case by creating an empty TermsOfUseAndAccess for whichever one is missing, and adding a method to reduce repeated code.

Closes #10814

**Special notes for your reviewer**: The test currently tests against the original algorithm and against a hardcoded expected value in the test. There's no real reason to test against the original algorithm except to validate that the hardcoded expected values in the test are correct, so I've marked the test code doing the check against the original as @Deprecated. 

The current tests don't do anything to test scaling (they basically try adding one file in each category (changed metadata, changed variable-level metadata, added file, deleted file, replaced file) and see if they differences are calculated as expected/as before. If someone wants to add a loop with ~1K files, we could probably start seeing the difference.

**Suggestions on how to test this**: QA should probably both confirm that the differencing appears to work for the Unit test cases (add a version or two and change metadata, add a file, delete a file, replace a file between versions) and make sure the difference makes sense. Checking the changed variable-level metadata probably requires testing with a tab file and perhaps editing with the DataCurator (Borealis tool for editing variable metadata) or editing in the db. 

Scalability can be estimated by finding/creating a published dataset with several K files and checking the time to load the version table. (Maybe there's an API call?) I think all we need is to verify this is _much_ faster/ worth the effort. (I added logging statements to print the start/stop times before/after the new/old algorithms and then ran the code on a dataset with 10K files. I've left the timing statements for the new algorithm in for fine logging, but the original code doesn't have any.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionDifference.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionDifference.java
@@ -8,6 +8,8 @@ import edu.harvard.iq.dataverse.util.StringUtil;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.logging.Logger;
 
@@ -15,6 +17,7 @@ import org.apache.commons.lang3.StringUtils;
 import edu.harvard.iq.dataverse.util.BundleUtil;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetVersionDifference.java
```diff
@@ -1829,4 +1615,12 @@ private static boolean fieldsAreDifferent(DatasetField originalField, DatasetFie
         }
         return false;
     }
+
+    List<FileMetadata> getgetChangedVariableMetadata() {
```
sekmiller, line 1619: typo here "getget"?

sekmiller (approved):
OK, this looks good. thanks for the typo fix.

ofahimIQSS:
Testing passed, merging PR

[Testing of 10818.docx](https://github.com/user-attachments/files/17818500/Testing.of.10818.docx)

PR #10820, opened 2024-09-02
A Fix for the Dataset Thumbnail Bug on Publish, 10819

### LINKED ISSUE #10819: Publishing breaks designated Dataset thumbnail, messes up Collection page
Appears to be new in 6.3. 
To reproduce: 
- Create a dataset 
- Upload 1 image file that's below the thumbnail generation cutoff
- Confirm that you are seeing the thumbnail on the dataset page
- Go to the collection page, confirm that the thumbnail is showing there as well, for this new draft dataset (i.e., this image file has become the designated thumbnail for the dataset)

Now publish the dataset.
In the process, this image file will be marked as though the thumbnail generation failed (`dvobject.previewimagefail = true`) in the db; making the default icon appear on the dataset page instead of the thumbnails. However, the DataFile will stay marked as the designated thumbnail for the dataset (via `dataset.thumbnailfile_id`), and that in turn will mess up the rendering of the collection page, like this: 
Screen Shot 2024-08-27 at 3 44 04 PM
(Note: it may be somewhat intermittent - I've seen the publishing succeed a couple of times without triggering this issue; just keep adding files and publishing until it happens)

If you go to `Edit Dataset` -> `Thumbnails`, the thumbnail will still show there perfectly intact. Un-assigning the file there and choosing the default icon instead fixes the collection page. Alternatively, setting `previewimagefail = false` in the db (via `/api/admin/clearThumbnailFailureFlag/{id}` or directly) fixes the proper thumbnail display on both pages.

landreev:

The bug is described in the linked issue with a screenshot provided. 

Closes #10819

I understand that the issue was a combined side effect of the PRs #10509 and #5621 (a recent merge, despite the low number). A jsf rendering rule was effectively requesting a 0-sized thumbnail for a dataset. This failed internally and the new failure-detecting code would catch that and mark the datafile with `previewimagefail = true`. (Interestingly, the test actually returned true, thus resulting in a non-working thumbnail url in the `og:image` tag).  Because of a different code path for drafts, this would only happen once the dataset was published. In the end, the dataset was ending up with `thumbnailfile_id` pointing to a datafile marked as a `previewimagefail`. The resulting failure was/looked much worse on the collection page. 

The above took longer to explain than it did to fix. But please note that the one line fix in the PR does not perform the same exact test as the jsf line it replaced. I assumed that it was close enough. But if the same exact test is needed, it would just need to be implemented as a DatasetPage bean method. 

**Additional documentation**:


src/main/webapp/dataset.xhtml
```diff
@@ -86,7 +86,7 @@
                     <meta property="og:title" content="#{DatasetPage.title}" />
                     <meta property="og:type" content="article" />
                     <meta property="og:url" content="#{DatasetPage.dataverseSiteUrl}/dataset.xhtml?persistentId=#{dataset.globalId}" />
-                    <meta property="og:image" content="#{DatasetPage.dataset.getDatasetThumbnail(ImageThumbConverter.DEFAULT_PREVIEW_SIZE) == null ? DatasetPage.dataverseSiteUrl.concat(resource['images/dataverse-icon-1200.png']): DatasetPage.dataverseSiteUrl.concat('/api/datasets/:persistentId/thumbnail?persistentId=').concat(DatasetPage.dataset.getGlobalId().asString())}" />
+                    <meta property="og:image" content="#{empty DatasetPage.thumbnailString ? DatasetPage.dataverseSiteUrl.concat(resource['images/dataverse-icon-1200.png']): DatasetPage.dataverseSiteUrl.concat('/api/datasets/:persistentId/thumbnail?persistentId=').concat(DatasetPage.dataset.getGlobalId().asString())}" />
```
qqmyers, line 89: Note: Part of #10687 was a different fix for this line - looks like it was asking for a size 0 preview. This fix here is fine - neither are actually asking for the same size that the api call returns (ImageThumbConverter.DEFAULT_CARDIMAGE_SIZE), so using a cached value makes sense.

qqmyers (approved):
Looks fine. I didn't test but the fix makes sense.

src/main/webapp/dataset.xhtml
landreev: Oh wow, I completely missed all that - that you already had an issue and a PR for the bug in question. :(

landreev: But yes, my logic was that if we have a 140px thumbnail available, generating the tiny one should not be a problem either. 
I will add a test for `if (size == 0) { ... return null; }` to DatasetUtil.java though.

qqmyers: That's in the other PR I think

landreev:
@qqmyers 

> Looks fine. I didn't test but the fix makes sense.

I should add the fix instructions to the release note too. Rather than telling them to use the `clearThumbnailFailureFlag`, either globally or on the files affected, I was thinking of suggesting a db query: 

```
UPDATE dvobject SET previewimagefail=FALSE,previewimageavailable=TRUE WHERE id IN (SELECT f.id FROM dvobject f, dataset d WHERE d.thumbnailfile_id = f.id AND f.previewimagefail IS TRUE);
```
What do you think?

qqmyers:
> @qqmyers
> 
> > Looks fine. I didn't test but the fix makes sense.
> 
> I should add the fix instructions to the release note too. Rather than telling them to use the `clearThumbnailFailureFlag`, either globally or on the files affected, I was thinking of suggesting a db query:
> 
> ```
> UPDATE dvobject SET previewimagefail=FALSE,previewimageavailable=TRUE WHERE id IN (SELECT f.id FROM dvobject f, dataset d WHERE d.thumbnailfile_id = f.id AND f.previewimagefail IS TRUE);
> ```
> 
> What do you think?

A query makes sense, but I guess I'd include the clear flag option  as well for those who aren't comfortable with db changes.

landreev:
@qqmyers Recommending the API solution, before the db query makes perfect sense, will do.
I'm just a little hesitant to suggest the "clear all" API - as that may have a potential to cause more trouble than its worth (?), if there are any real failure cases in the database.

qqmyers:
Other than being a little more work for the instance - Dataverse will try to generate ones w/o a fail one time before resetting the flag - I don't think it will cause issues. (The flag is just for performance to avoid retrying on every access - having to try one time isn't much).

src/main/webapp/dataset.xhtml
landreev: That's what I meant, that I was going to copy-and-paste these lines from #10687; done. 
Same for the release note, added the recommendations to use the Clear apis, decided not to mention the direct database query.

PR #10825, opened 2024-09-04
IQSS/10824 - Add logic to suppress query tool display for non-public files.

### LINKED ISSUE #10824: Query Tools show for restricted files on file page
Query tools are currently on supposed to work for publicly readable data files. On the dataset page, query tool buttons are suppressed for restricted/embargoed/not-yet-released files. On the file page, this doesn't happen.
* What happens?
A configured query tool shows on the file page but fails with a 403.

6.3+

TBD

qqmyers:
**What this PR does / why we need it**: Stops query tools from displaying on the file page when the file is not public. (Query tools are currently designed only to work with public files).

Closes #10824

**Special notes for your reviewer**: FWIW: I made the fix such that a theoretical query tool that reads only an aux file would still display. That's different from the DatasetPage check which doesn't address this case. 

**Suggestions on how to test this**: Look for the Ask-the-data query tool on a file that is of the right type to use it but is not yet released, is restricted, is embargoed, etc. and verify that it doesn't show in the list of tools on the file page.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/FilePage.java
```diff
@@ -305,6 +305,16 @@ private void loadExternalTools() {
             exploreTools = exploreTools.stream().filter(tool ->tool.accessesAuxFiles()).collect(Collectors.toList());
             queryTools = queryTools.stream().filter(tool ->tool.accessesAuxFiles()).collect(Collectors.toList());
             toolsWithPreviews = toolsWithPreviews.stream().filter(tool ->tool.accessesAuxFiles()).collect(Collectors.toList());
+        } else {
+            // Don't list queryTools for non-public files
+            // Note - this logic is not the same as isPubAliclyDownloadable which appears to be true for a draft-only file
```
pdurbin, line 310:
```suggestion
            // Note - this logic is not the same as isPubliclyDownloadable which appears to be true for a draft-only file
```

pdurbin (approved):
I haven't tested it, but this change makes sense to me.

sekmiller:
@qqmyers when you have a chance can you please get the latest from dev? thanks!

PR #10841, opened 2024-09-12
IQSS/10840 - DatasetTypesIT test fix

### LINKED ISSUE #10840: DatasetTypesIT can fail
One test selects a random name from a UUID which can be all numeric, which is not allowed for dataset type names, causing the test to fail. PR to follow.

qqmyers:
Simple fix for a random test failure - when the generated alphanumeric name is randomly all numeric, which is not allowed.

Closes #10840 

Simple fix - we've had this class of issue before.

**Suggestions on how to test this**: Just verify the test passes - probably not worth trying to see the problem on the current branch - https://jenkins.dataverse.org/blue/organizations/jenkins/IQSS-Dataverse-Develop-PR/detail/PR-10632/16/tests/ shows one occurence.

**Additional documentation**:


pdurbin (commented):
I see @sekmiller is already looking at this but consider this an "approve" from me. 😄

sekmiller (approved):
Does what it purports to do

PR #10855, opened 2024-09-17
standardize image url

### LINKED ISSUE #10831: Standardize the image_url field of the Search API so that it uses regular URLs instead of base64 for all result types

Currently, image_url returns base64 URLs for files and dataverses, while it returns a regular URL for datasets. The goal is to standardize all URLs to use the same format. We have chosen to use regular URLs instead of base64 for all cases.

API User

https://github.com/IQSS/dataverse/pull/10811#discussion_r1750411773

Use regular URLs in image_url fields for all result types of the Search API.

None

https://github.com/IQSS/dataverse/pull/10811

Yes

stevenwinship:
**What this PR does / why we need it**: Currently, image_url returns base64 URLs for files and dataverses, while it returns a regular URL for datasets. The goal is to standardize all URLs to use the same format. We have chosen to use regular URLs instead of base64 for all cases.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/10831

Closes #10831 

**Suggestions on how to test this**: SearchIT has a test for datafiles. Collection test can be done manually by adding a logo and calling a search. verify that the image_url is there and is reachable.

**Is there a release notes update needed for this change?**: Yes, since this changes the body of the search response

**Additional documentation**:

### DIFF
--- a/doc/release-notes/10831-standardize-image-url-of-search-api.md
+++ b/doc/release-notes/10831-standardize-image-url-of-search-api.md
@@ -1,0 +1,28 @@
+Search API (/api/search) response will now include new image_url format for the Datafile and Dataverse logo.
+Note to release not writer: this supersedes the release note 10810-search-api-payload-extensions.md
+
+For Dataverse:
+
+- "image_url" (optional)
+
+```javascript
+"items": [
+    {
+        "name": "Darwin's Finches",
+        ...
+        "image_url":"/api/access/dvCardImage/{identifier}"
+(etc, etc)
+```
+
+For DataFile:
+
+- "image_url" (optional)
+
+```javascript
+"items": [
+    {
+        "name": "test.txt",
+        ...
+        "image_url":"/api/access/datafile/{identifier}?imageThumb=true"
+(etc, etc)
+```
--- a/doc/release-notes/10831-standardize-image-url-of-search-api.md
+++ b/doc/release-notes/10831-standardize-image-url-of-search-api.md
@@ -1,5 +1,5 @@
 Search API (/api/search) response will now include new image_url format for the Datafile and Dataverse logo.
-Note to release not writer: this supersedes the release note 10810-search-api-payload-extensions.md
+Note to release note writer: this supersedes the release note 10810-search-api-payload-extensions.md
 
 For Dataverse:
 
--- a/doc/sphinx-guides/source/api/changelog.rst
+++ b/doc/sphinx-guides/source/api/changelog.rst
@@ -7,6 +7,11 @@ This API changelog is experimental and we would love feedback on its usefulness.
     :local:
     :depth: 1
 
+v6.4
+----
+
+- ** /api/search?q=**: Json values for image_url in DataFiles and Collections have changed from Base64URL ("data:image/png;base64,...) to "/api/access/datafile/{identifier}?imageThumb=true" and "/api/access/dvCardImage/{identifier}" respectively. This was done to match the image_url of Dataset.
+
 v6.3
 ----
 
--- a/doc/sphinx-guides/source/api/changelog.rst
+++ b/doc/sphinx-guides/source/api/changelog.rst
@@ -7,11 +7,6 @@ This API changelog is experimental and we would love feedback on its usefulness.
     :local:
     :depth: 1
 
-v6.4
-----
-
-- ** /api/search?q=**: Json values for image_url in DataFiles and Collections have changed from Base64URL ("data:image/png;base64,...) to "/api/access/datafile/{identifier}?imageThumb=true" and "/api/access/dvCardImage/{identifier}" respectively. This was done to match the image_url of Dataset.
-
 v6.3
 ----
 
--- a/src/main/java/edu/harvard/iq/dataverse/ThumbnailServiceWrapper.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ThumbnailServiceWrapper.java
@@ -48,6 +48,18 @@ public class ThumbnailServiceWrapper implements java.io.Serializable  {
     private Map<Long, DvObject> dvobjectViewMap = new HashMap<>();
     private Map<Long, Boolean> hasThumbMap = new HashMap<>();
 
+    public String getFileCardImageAsUrl(SolrSearchResult result) {
+        if (result.isHarvested()) {
+            return null;
+        }
+
+        if (result.getEntity() == null) {
+            return null;
+        }
+
+        return SystemConfig.getDataverseSiteUrlStatic() + "/api/access/datafile/" + result.getEntity().getId() + "?imageThumb=true";
+    }
+
     // it's the responsibility of the user - to make sure the search result
     // passed to this method is of the Datafile type!
     public String getFileCardImageAsBase64Url(SolrSearchResult result) {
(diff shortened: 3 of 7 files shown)


doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -7,6 +7,11 @@ This API changelog is experimental and we would love feedback on its usefulness.
     :local:
     :depth: 1
 
+v6.4
+----
+
+- ** /api/search?q=**: Json values for image_url in DataFiles and Collections have changed from Base64URL ("data:image/png;base64,...) to "/api/access/datafile/{identifier}?imageThumb=true" and "/api/access/dvCardImage/{identifier}" respectively. This was done to match the image_url of Dataset.
+
 v6.3
 ----
```
pdurbin, line 17: In #10811 the following line was added:

"Note that the image_url field, if exists, will be returned as a regular URL for Datasets, while for Files and Dataverses, it will be returned as a Base64 URL. We plan to standardize this behavior so that the field always returns a regular URL. (See: https://github.com/IQSS/dataverse/issues/10831)"

Can we remove it?

qqmyers (approved):
Changes look good overall. The PR is currently minimalistic (~per discussion) so I think the issue owner (@GPortas ) should verify that it meets the need and either request changes or add new issues as required. Things I see that are relevant:

- Both the new datafile and dataverse/collection links work when an image exists. They were existing API calls that have just been added to the search results in this PR. 
- When an image doesn't exist, the file call returns 404 and the collection returns 204(No Content).  Should they be changed to be the same? Should an image_url not be sent when  there isn't one?

- There is currently no API call to add a Dataverse image, which, as @stevenwinship  pointed out, means there's no automated test of the dataverse/collection download. Manual testing shows that it works, so probably not a big issues. That said - when is an upload API call going to be needed? Should that be added to this PR or be a new issue?

- I'm guessing the file API call, which uses a query param, won't be cached by the browser. Is that something that should be fixed now/later (i.e. by providing a URL w/o a query param)? Is it even an issue given that in many cases, that call will end up redirecting to a signed S3 download URL for the thumbnail (does the browser cache answers from redirects?).

Given that I think the PR works as is, I'm going to approve it, but if anyone thinks more scope should be added before it moves forward, just note the change and move it back to in progress.

pdurbin:
I'm putting it back to "in progress" for this: https://github.com/IQSS/dataverse/pull/10855#discussion_r1765292251

doc/sphinx-guides/source/api/changelog.rst
stevenwinship: removed. also modified the example json

doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -7,6 +7,11 @@ This API changelog is experimental and we would love feedback on its usefulness.
     :local:
     :depth: 1
 
+v6.4
+----
+
+- ** /api/search?q=**: Json values for image_url in DataFiles and Collections have changed from Base64URL ("data:image/png;base64,...) to "/api/access/datafile/{identifier}?imageThumb=true" and "/api/access/dvCardImage/{identifier}" respectively. This was done to match the image_url of Dataset.
```
pdurbin, line 13: Do we still need this release note if we merge this before we release 6.4?

stevenwinship: removed for now. can be re-added if this doesn't make 6.4

pdurbin:
As discussed [in Slack](https://iqss.slack.com/archives/C03R1E7T4KA/p1726681465036819), I'm giving this the 6.4 milestone (thanks for the 👍 @cmbz) because it fixes something (`image_url`) in #10811 which has already been merged and will be part of 6.4. Basically, since we're shipping an update to `image_url`, we should ship the fixed version.

doc/release-notes/10831-standardize-image-url-of-search-api.md
```diff
@@ -0,0 +1,28 @@
+Search API (/api/search) response will now include new image_url format for the Datafile and Dataverse logo.
+Note to release note writer: this supersedes the release note 10810-search-api-payload-extensions.md
+
+For Dataverse:
+
+- "image_url" (optional)
```
GPortas, line 6: If the image_url is always returned, even though the URL returns a 404 for files and a 204 for collections, then it is not an optional field. Anyway, check my review comment, where I suggest not adding the fields to the payload in these cases.

GPortas (commented):
1) `image_url` field when no image is available.

I find it a bit confusing and not very practical for the API consumer that, when no image exists, a URL is still returned in the `image_url` field, which results in 404 or 204 codes. 

I think it’s more appropriate to prevent this scenario directly from the backend, rather than requiring all consumers to handle the logic of checking whether images actually exist.

After updating this, we should also mention the optionality of the field in the documentation.

2) API call to add a Dataverse image

This API call will indeed be necessary for the SPA when we implement the functionality to add images to a collection. 

Normally, when I extend the API, it often happens that I have to implement write endpoints just for the setup of the API tests for the read endpoints, as is the case here. 

The good thing is that these endpoints will already be implemented by the time the related UI feature need to be implemented in the SPA.

If it’s not too big to implement now (Given the upcoming `6.4` release), I would implement the endpoint in this PR. If not, we can keep it as is and implement it later along with the pending automated tests.

3) Caching and query params & redirects

I would probably address this point in a future iteration, as I see it more as an optimization than a critical aspect for the SPA and other consumers to operate. 

I also don't believe that switching from returning URLs with query parameters to using a different structure with path parameters is a backwards compatibility issue to worry about.

doc/release-notes/10831-standardize-image-url-of-search-api.md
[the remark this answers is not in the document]
g-saracca: @GPortas what if when no image is available the endpoint just returns `image_url: null` ?
In that way we will know that `image_url` could be either a `string`(image exists) or `null`.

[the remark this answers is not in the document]
stevenwinship: removed image_url if none exists

src/main/java/edu/harvard/iq/dataverse/ThumbnailServiceWrapper.java
```diff
@@ -49,15 +49,14 @@ public class ThumbnailServiceWrapper implements java.io.Serializable  {
     private Map<Long, Boolean> hasThumbMap = new HashMap<>();
 
     public String getFileCardImageAsUrl(SolrSearchResult result) {
-        if (result.isHarvested()) {
-            return null;
-        }
 
-        if (result.getEntity() == null) {
-            return null;
+        if (!result.isHarvested() && result.getEntity() != null && (!((DataFile)result.getEntity()).isRestricted()
```
qqmyers, line 53: This should probably use the logic in https://github.com/IQSS/dataverse/blob/9a1e49453731ffc57f4dc13fdea88e6728957ff2/src/main/java/edu/harvard/iq/dataverse/api/Access.java#L1759 or https://github.com/IQSS/dataverse/blob/9a1e49453731ffc57f4dc13fdea88e6728957ff2/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L5831-L5836 to address embargo and retention

stevenwinship: @qqmyers Wouldn't the Access call using  the image_url go through the isAccessAuthorized() method anyways? So if we supply the url the user then needs to make the call to get the image which would fail is they don't have access. I don't want to duplicate the code so I would need to make it public and be able to call it from the wrapper to prevent sending the url.

qqmyers: Yes, and it would make sense to change the code to public or move the relevant parts to a util class to avoid duplicating code.

stevenwinship: @qqmyers I went with the second option of copying the code from DatasetPage

GPortas (approved):
Works good. Merging.

PR #10865, opened 2024-09-19
IQSS/10379 Fix Metrics API Bugs

### LINKED ISSUE #10379: Metrics caching doesn't handle dataLocation parameter
The dataset related calls in https://github.com/IQSS/dataverse/blob/1dd5fa4717cef2667fa19491d5d2ba778480a472/src/main/java/edu/harvard/iq/dataverse/api/Metrics.java#L193-L313 support a dataLocation parameter to get info for local, remote (harvested), or all datasets. However, the caching doesn't handle the parameter. Thus, whichever call is made first is cached and served as the response regardless of whether the parameter is different the next time. A partial workaround is to clear the cache before making a call. 

A fix would involve adding the param value (after sanitizing) to the metricName being cached so that values for all three options are cached separately.

* When does this issue occur?
Metrics API

develop/6.1+

qqmyers:
**What this PR does / why we need it**: This PR addresses a bug reported in #10379 where a cached value did not include the choice of whether to report local, remote (harvested), or all datasets for the /datasets and /datasets/monthly endpoints, resulting in incorrect values when different values were used and a cached value was available. The PR uses the existing mechanism to cache the results with the local/remote/all dataLocation choices to avoid this.

The PR also addresses the issue raised in https://github.com/IQSS/dataverse/pull/10710#discussion_r1766904473 where the old queries to find the latest relevant version were broken if/when the minor version was greater than 9. The new queries rely on ordering the results by both the major and minor version numbers and using DISTINCT ON to pick the first entry (the one with the highest major version, and among those, the highest minor version.

Closes #10379

**Special notes for your reviewer**: I did some checking of the results from the old/new queries in sql: e.g. in simplest form, comparing 
` select DISTINCT ON (datasetversion.dataset_id) datasetversion.id from datasetversion join dataset on dataset.id = datasetversion.dataset_id where versionstate='RELEASED' order by datasetversion.dataset_id, datasetversion.versionnumber desc, datasetversion.minorversionnumber desc;`

with  
`select datasetversion.dataset_id || ':' || max(datasetversion.versionnumber || datasetversion.minorversionnumber) from datasetversion join dataset on dataset.id = datasetversion.dataset_id where versionstate='RELEASED' group by dataset_id;`

and they appear to give the same results. In going through the queries for different metrics, I realized that the string value being returned was never used, so I've simplified the selects to just return just the dataset id (when all we're doing is count(*)) or the datasetversion id (when we're matching against the version to, for example, find filemetadatas in that version).

Compare the results of any/all of the metrics APIs with the changed code. A potential shortcut that covers most endpoints would be to install/run the https://github.com/gdcc/dv-metrics app and view it's webpage and then compare the graphs produced before and after the PR. Whether testing manually or via the app, you need to `truncate metric;` to remove the cached values or call the[ /clearMetricsCache api call](https://guides.dataverse.org/en/latest/api/native-api.html#metrics).

**Additional documentation**:


landreev (approved):

ofahimIQSS:
Tested in internal environment. Fix looks good. 

[Testing of 10865.docx](https://github.com/user-attachments/files/17512635/Testing.of.10865.docx)

PR #10870, opened 2024-09-23
Fix NPE using CVOC

### LINKED ISSUE #10869: 6.3 : Indexing issue while using CVOC setting
Since Dataverse v6.3 (related to #10505 coded by myself), if you are using [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) with these specific conditions : 
- a child datasetfield as `term-uri-field` parameter 
- **AND** an empty json object for the `managed-fields` parameter

Then you will experience no indexation on update or create of dataset because of a `NullPointerException` handled by the code avoiding red alert message but catched silently causing no document to be add to solr.
Freshly created dataset will not appear and updated dataset will still appear but with unupdated metadata.

If you are in this context, I see no workaround but to patch your fork or deploy a fixed/superior version.

@qqmyers @pdurbin Could you help me for the communication and if it's suitable maybe the packaging of a new 6.3 version like https://groups.google.com/g/dataverse-community/c/evn5C-pyrS8/m/JrH9vp47DwAJ ?
At least add it to 6.4 milestone.
Thanks in advance and sorry for this.

luddaniel:

Fixes a possible `NullPointerException` that prevents the proper indexing of a dataset.

Closes #10869

Play around [:CVocConf](https://guides.dataverse.org/en/6.3/installation/config.html#cvocconf) and having multiple configuration activated like SKOMOS + ORCID

No

I don't think

### DIFF
--- a/doc/release-notes/10869-fix-npe-using-cvoc.md
+++ b/doc/release-notes/10869-fix-npe-using-cvoc.md
@@ -1,0 +1,1 @@
+This release fixes a bug in the external controlled vocabulary mechanism (introduced in v6.3) that could cause indexing to fail when a script is configured for one child field and no other child fields were managed.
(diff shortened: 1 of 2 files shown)


qqmyers (approved):
Makes sense. - Please add a one line release note, e.g.  - This release fixes a bug in the external controlled vocabulary mechanism (introduced in v6.3) that could cause indexing to fail when a script is configured for one child field and no other child fields were managed. 

FWIW: I didn't see this error using ORCID and ROR for the author field (requiring #10712) because the NPE is avoided due to the ORCID script registering to manage the other fields. - not sure if that provides a way to work-around in 6.3 or not, i.e. by adding a fake registration for a second (empty, missing?) script that would include the managed fields.

Issue #8813, opened 2022-06-23
More useful facet displays for dates


adam3smith:
**What kind of user is the feature intended for?**
Data re-user

**What inspired the request?**
Improving data findability on QDR; User feedback

**What existing behavior do you want changed?**
Display of dates in facets: Currently, dates are simply displayed as individually, filterable years in the facet. That's un-intuitive, imprecise, and diverges from standard behavior in other search interfaces (that's why, I assume, date is rarely used as a facet in existing Dataverse)

**Any brand new behavior do you want to add to Dataverse?**
Some version of a date-range selector. Here are some implementations from other catalogs:
[Princeton library discovery layer ](https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=testing)(cool visual + number implementation):
![image](https://user-images.githubusercontent.com/128986/175392379-80341e63-ae10-48fb-8c80-54c7ba9b466c.png)

[Summons library discovery layer](https://syracuse.summon.serialssolutions.com/#!/search?ho=t&include.ft.matches=f&fvf=ContentType,Book%20%2F%20eBook,f%7CContentType,Dissertation,f%7CContentType,Book%20Chapter,f%7CContentType,Government%20Document,f&l=en&q=testing) -- has exact date input and common intervals
![image](https://user-images.githubusercontent.com/128986/175392635-c6c919f8-9754-4d26-9476-1de4b0117667.png)

[UK Data](https://beta.ukdataservice.ac.uk/datacatalogue/studies/?Search=testing#!?Search=testing&Page=1&Rows=10&Sort=1&DateFrom=440&DateTo=2022 - barebones implementation but wanted to include a data repository)
![image](https://user-images.githubusercontent.com/128986/175392797-91a385e9-6023-47ee-a3d7-29eae85c2017.png)

**Any related open or closed issues to this feature request?**
No

pdurbin:
Related:

- https://github.com/IQSS/dataverse-frontend/issues/278

pdurbin:
Please see this new PR:

- https://github.com/IQSS/dataverse/pull/10887

Issue #370, opened 2014-07-09
Index: Index according to data type to facilitate range facets and queries.


eaquigley:
---

Author Name: **Kevin Condon** (@kcondon)
Original Redmine Issue: 3785, https://redmine.hmdc.harvard.edu/issues/3785
Original Date: 2014-03-27
Original Assignee: Philip Durbin

---

This ticket is a placeholder for indexing fields that contain something other than strings, eg. dates and floats, to support range queries for either facets or searches, depending on what is finally decided.

Example date: production date
Example float: Coverage Central Wavelength(m)

---

Related issue(s): #70, #485
Redmine related issue(s): [3478](https://redmine.hmdc.harvard.edu/issues/3478), [3900](https://redmine.hmdc.harvard.edu/issues/3900)

---

kcondon:
Closing as a duplicate and merging with #70.

pdurbin:
Please see this new PR:

- https://github.com/IQSS/dataverse/pull/10887

Issue #401, opened 2014-07-09
Add Range as a fieldType


eaquigley:
---

Author Name: **Gustavo Durand** (@scolapasta)
Original Redmine Issue: 3816, https://redmine.hmdc.harvard.edu/issues/3816
Original Date: 2014-04-04
Original Assignee: Steve Kraffmiller

---

None

---

Related issue(s): #368, #485
Redmine related issue(s): [3783](https://redmine.hmdc.harvard.edu/issues/3783), [3900](https://redmine.hmdc.harvard.edu/issues/3900)

---

scolapasta:
This is for fields like Timer period (start) and time oeriod (end) to define them as one field type instead of 2.

pdurbin:
Related: 

- #5370

poikilotherm:
At https://dataverse.zulipchat.com/#narrow/stream/379673-dev/topic/metadata.20validators.20per.20field I have outlined an idea how to make this happen using custom validators, so we could keep field types as data types. (A range is not a data type, but a constraint)

pdurbin:
We have at least partial support for this thanks to this PR @vera made:

- #10887

PR #10887, opened 2024-09-27
feat: index numerical and date fields in Solr with appropriate types + more targeted search result highlighting

vera:

Currently, all fields regardless of type are indexed in Solr as English text (`text_en`). With this PR, numerical and date fields are indexed in Solr with appropriate types:

| Field type defined in TSV  | Field type indexed in Solr |
| ------------- | ------------- |
| `int`  | `plong`  |
| `float` | `pdouble` |
| `date`  | `date_range` (`solr.DateRangeField`)  |

I chose to index dates as `DateRangeField` because they can be used to represent dates to any precision, e.g. a day YYYY-MM-DD, a month YYYY-MM or a year YYYY. See: [Date Formatting and Date Math :: Apache Solr Reference Guide](https://solr.apache.org/guide/solr/latest/indexing-guide/date-formatting-math.html)

This matches the [allowed formats in a date field as defined by Dataverse](https://guides.dataverse.org/en/latest/admin/metadatacustomization.html#fieldtype-definitions).

This means that range queries are now possible on numerical and date fields, e.g. `exampleIntegerField:[25 TO 50]` or `exampleDateField:[2000-11-01 TO 2014-12-01]`.

This PR implements ranged queries as discussed in #370 (issue was already closed)

This issue is related to #8813 and https://github.com/IQSS/dataverse-frontend/issues/278 (the range queries that are now possible lay the groundwork for a nicer search facet UI)

For testing, I've created a sample TSV containing all relevant fields [here](https://gist.github.com/vera/56d27968671fee80fae4b9fab95b1cad).

## Regression Testing

- On a server running "develop", publish some datasets that use the affected fields mentioned in the release note (coverage.Depth, etc.). Be sure dates, integers and floats are all used by populating various different fields.
- Deploy this branch
- Test, test, test for any regressions

## Feature testing

1. Load sample TSV and update + reload Solr schema [as described in docs](https://guides.dataverse.org/en/latest/admin/metadatacustomization.html)
2. In the UI:
   1. Activate metadata block
   2. Activate facets for all three fields
   3. Create dataset with values in all three fields
3. Run test range queries via the search bar, e.g. `exampleIntegerField:[25 TO 50]` or `exampleDateField:[2000-11-01 TO 2014-12-01]`
4. Check that facets are working correctly

Facets still look the same as before. There is only a small change in the highlighting of search results, see my comment below

Yes, there should be an info text describing the new feature + instructions for how to activate the feature:

* the Solr schema.xml needs to be updated
* all datasets need to be reindexed

**Additional documentation**:

/


vera:
Additionally, I've set `hl.requireFieldMatch` to `true`:

> If false, all query terms will be highlighted for each field to be highlighted (hl.fl) no matter what fields the parsed query refer to. If set to true, only query terms aligning with the field being highlighted will in turn be highlighted.

https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html

Two reasons:

1. Querying solr with a date range query with activated highlighting using the default (unified) highlighter without `requireFieldMatch` triggers a 500 error in Solr (see [my post on the Solr mailing list](https://lists.apache.org/thread/j18r2j4pwxo9nfc20jzfwlcsgs5t9tcm). My guess is that Solr is attempting to highlight the matched date range within fields in a nonsensical way which triggers the error)
2. I think this improves the highlighting of search results, previously a match of my search term is highlighted anywhere even if I limited my query to a specific field, e.g. here "replication" is also highlighted in the title even though I limited my search specifically to the description:

image

With this change, the highlighting is limited to specific fields if the query is:

image
image
image

johannes-darms:
@qqmyers Would it make sense to include this feature in the next release? It's a rahter small adaption that improve the serach experience.

pdurbin:
@johannes-darms it's a very cool feature that adds a lot of value, something I've wanted for years.

Let's see what @cmbz and @scolapasta think.

qqmyers:
One question - is it possible to enter or have legacy values that don't fit the new types that would break indexing?

cmbz:
2024/10/15: Added to sprint ready after conversation with @pdurbin

vera:
@qqmyers good question. When I tried to enter to invalid data (non-integer in an integer field, non-float in a float field, or non-date in a date field), I got the following errors via the UI:

image

...and via the API:

```{"status":"ERROR","message":"Validation Failed: Example integer field  is not a valid integer. (Invalid value:edu.harvard.iq.dataverse.DatasetFieldValueValue[ id=null ]), Example floating point field  is not a valid number. (Invalid value:edu.harvard.iq.dataverse.DatasetFieldValueValue[ id=null ]), Example date field  is not a valid date. \"yyyy\" is a supported format. (Invalid value:edu.harvard.iq.dataverse.DatasetFieldValueValue[ id=null ]).java.util.stream.ReferencePipeline$3@6e9605fb"}```

[The validator code seems to be quite old](https://github.com/IQSS/dataverse/blob/d039a108f992f170d5323a26a6d6f7dafb14029b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java#L26), so I don't know if there could be any installations with legacy invalid values entered before it was added.

However, looking at the validator code, I found that date fields allow some formats which are not documented: `YYYY` followed by `AD` or `BC`, a "Bracket format" (not familiar with this), and datetime formats (`yyyy-MM-dd'T'HH:mm:ss`, `yyyy-MM-dd'T'HH:mm:ss.SSS` and `yyyy-MM-dd HH:mm:ss`).

When I add a dataset using one of those formats and try to index it, the indexing fails. The Dataverse log shows an error like `dev_solr> org.apache.solr.common.SolrException: ERROR: [doc=dataset_2_draft] Error adding field 'exampleDateField'='2024-09-01 12:34:56' msg=Couldn't parse date because: Improperly formatted datetime: 2024-09-01 12:34:56` and the dataset is missing from the UI.

So, yes, there may be some installations with date field values using the above formats which would cause invisible datasets due to indexing errors. I am not sure how we should deal with this. Are those date formats intended to be officially fully supported?

If no, it might be OK to offer a workaround in the upgrade instructions like "Please ensure that all date fields containing legacy dates in formats other than `YYYY`, `YYYY-MM`, `YYYY-MM-DD`, or `YYYY-MM-DDThh:mm:ssZ` are not updated to the new `date_range` type in the Solr schema. Otherwise, datasets with these legacy dates will fail to index and disappear from your Dataverse page.".

If yes, this feature becomes a bit more complicated, because Solr does not support those date formats and we would need to work around that somehow.

**Comment by vera** [comment]
@pdurbin I've just added an API test for the range queries as you suggested.

**Comment by qqmyers** [comment]
I don't know what is required but I wouldn't be surprised if BC dates are something people want and want to have indexed. Perhaps @jggautier would know more about what legacy values exist and what's required. 
W.r.t. validation - I'd also make sure that the API calls don't allow bad values - I know there was newer validation code just added there. W.r.t. the code, I'd definitely suggest doing checks for int, float, date values when indexing or somehow assure that a bad legacy value can't break the overall submission and we just drop that field rather than have the dataset not index. (I didn't see such a check but I might have missed it).

**Comment by vera** [comment]
Solr does support BC dates, however in a different format than `YYYYBC`:

> `-0009` – The year 10 BC. A 0 in the year position is 0 AD, and is also considered 1 BC.
https://solr.apache.org/guide/solr/latest/indexing-guide/date-formatting-math.html

Is the code doing checks for API-submitted datasets different from the code doing the UI checks? I assumed it was both the same code I linked above, since the error messages are the same.

> I'd definitely suggest doing checks for int, float, date values when indexing or somehow assure that a bad legacy value can't break the overall submission and we just drop that field rather than have the dataset not index.

Yes, that would be nice.

**Comment by jggautier** [comment]
Hi all. I haven't been following this issue closely enough to contribute and won't have the time to catch up. But I agree with Jim and encourage folks to look into how these fields are being used now. My dataset at https://doi.org/10.7910/DVN/2SA6SN might be helpful for seeing who's using the fields in different ways. And the list of contacts in [our spreadsheet of Dataverse installations](https://docs.google.com/spreadsheets/d/1bfsw7gnHlHerLXuk7YprUT68liHfcaMxs1rFciA-mEo) might help for contacting particular installations to learn more.

**Comment by pdurbin** [comment]
Here's the related issue about BC dates:

- #10843

**Comment by vera** [comment]
I've just pushed a commit that implements the suggestion above (if encountering a bad legacy value in an int/float/date field, just drop that field, but index the rest of the dataset).

So, if a dataset contains a bad legacy value in an int/float/date field, this means that queries on that field will not yield that dataset, since the field hasn't been indexed. But for any other query, the dataset will still be found. (I've also added a test showing this)

I think this is a small limitation. And we could add BC support relatively easily in a future PR.

**Comment by pdurbin** [comment]
@vera I poked your PR today. I'm very excited about this feature. I started looking into how existing fields will be affected. Please take a look at this PR of suggestions:

- https://github.com/vera/dataverse/pull/1

Also, it's expected that range queries do not produce highlights, right? I'm getting hits but no highlights:

Screenshot 2024-11-15 at 5 08 14 PM
Screenshot 2024-11-15 at 5 08 09 PM

**Comment by vera** [comment]
Thank you! Yes, I think that's expected. I also haven't seen any highlighting for range queries, and didn't find any indication that Solr supports it in the documentation. I've just sent a mail to the Solr user mailing list to make sure. I'll take a look at your suggestions now. :)

**Comment by vera** [comment]
@pdurbin Suggestions looked good, I have merged them!

**Review by pdurbin — APPROVED** [review]
API tests are passing. I'm approving this. @vera please note that I added "regression testing" to "how to test this". I played around with an empty database (a new installation) but I didn't test with existing data.

One last thing I'll mention is that while working on this I created a pull request at #11024. I think it would be get the doc changes into the guides. I used them to update the Solr schema.xml file.

**Comment by ofahimIQSS** [comment]
I came across an issue while testing this in my local. 
After I load sample TSV and update + reload Solr schema [as described in docs](https://guides.dataverse.org/en/latest/admin/metadatacustomization.html).
1. Create a dataverse using the new Solr Metadata fields
2. Create a dataset within the same collection, ensuring solr metadata fields are populated as follows:
 Integer: 22 Float: 19 Date: 2024
3. Save and publish the dataset
4. Go back to search and try to find the dataset

Issue: Dataset does not display on the UI after publishishing

https://iqss.slack.com/files/U07NM5RQ0NB/F083GSANQKU/screen_recording_2024-12-03_at_1.29.58___pm.mov
[server.log.txt](https://github.com/user-attachments/files/17997514/server.log.txt)

**Comment by pdurbin** [comment]
@ofahimIQSS can you please provide more of server.log?

Also, let's add @vera to the PR to let her know you're having some trouble. Maybe she can help

**Comment by ofahimIQSS** [comment]
[serverlog.txt](https://github.com/user-attachments/files/17999259/serverlog.txt)
 attaching extended server.log file

**Comment by pdurbin** [comment]
@ofahimIQSS ah you're still re-testing. Yes, the error in your log...

`dev_solr> org.apache.solr.common.SolrException: ERROR: [doc=dataset_65] unknown field 'exampleDateField'\`

... means that you need to update your schema.xml file.

I'm not sure if this helps, but I added this one-liner...

`curl http://localhost:8080/api/admin/index/solr/schema | docker run -i --rm -v ./docker-dev-volumes/solr/data:/var/solr gdcc/configbaker:unstable update-fields.sh /var/solr/data/collection1/conf/schema.xml`

... to a PR I'm working on at #11024.

**Comment by ofahimIQSS** [comment]
Hi Vera - I’ve tried to validate this ticket but could not see the datasets after I publish them on my local. Steps to reproduce:

1. Build PR in local environment
2. Load the solr tsv file
3. Update Solr schema curl http://localhost:8080/api/admin/index/solr/schema | docker run -i --rm -v ./docker-dev-volumes/solr/data:/var/solr gdcc/configbaker:unstable update-fields.sh /var/solr/data/collection1/conf/schema.xml 
4. Clear all data from Solr and start Async Reindex based on https://guides.dataverse.org/en/6.4/admin/solr-search-index.html
5. Create a Collection with Solr Field Types Test Metadata
6. Create a dataset within the collection and publish
Issue: After publishing dataset, dataset is not appearing on the UI. Server.log file can be found below. 
[server.log.txt](https://github.com/user-attachments/files/18011696/server.log.txt)

**Comment by ofahimIQSS** [comment]
Overall, PR looks good. One observation I had was with the "Example Date Field". I can enter in a 2 digit value and save it as a date but the values specified for that field are (YYYY-MM-DD, YYYY-MM, or YYYY). 

image

**Comment by vera** [comment]
@ofahimIQSS thanks for testing! Yes, I am seeing the same behaviour. The document indexed in Solr for that metadata looks like this:

image

While `exampleFloatField` and `exampleIntegerField` are indexed, `exampleDateField` is not because "11" is an invalid date value (according to Solr).

I think this is a pre-existing bug in the [date field validation code](https://github.com/IQSS/dataverse/blob/d039a108f992f170d5323a26a6d6f7dafb14029b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValueValidator.java#L99) not caused by this PR. We might want to open an issue for that.

**Comment by pdurbin** [comment]
Hmm, this makes me wonder about two and three digit years from ancient history.

For example, the philosopher [Epictetus](https://en.wikipedia.org/wiki/Epictetus) was born in 50 and died in 135. It sounds like Dataverse doesn't want to index these as real dates. Do we have to pad them as 0050 and 0135? (I haven't tried this.)

This issue about BCE dates is related:

- #10843

To quote @vera from that issue, "Since the Solr search index underlying Dataverse supports BC dates (using the ISO 8601 format: 1 BC = +0000, 2 BC = -0001, and so on)."

**Comment by vera** [comment]
Yes, padding with zeroes seems to be the way intended by ISO8601/Solr. I just tried inputting "0011" instead of "11" and it indexes fine.

File: doc/release-notes/10887-solr-field-types.md [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+Additionally, search result highlighting is now more accurate, ensuring that only fields relevant to the query are highlighted in search results. If the query is specifically limited to certain fields, the highlighting is now limited to those fields as well.
+
+## Upgrade Instructions
+
+7\. Update Solr schema.xml file. Start with the standard v6.5 schema.xml, then, if your installation uses any custom or experimental metadata blocks, update it to include the extra fields (step 7a).
+
+Stop Solr (usually `service solr stop`, depending on Solr installation/OS, see the [Installation Guide](https://guides.dataverse.org/en/6.5/installation/prerequisites.html#solr-init-script)).
+
+```shell
+service solr stop
```
ofahimIQSS, line 46:
```suggestion
sudo service solr stop
```

doc/release-notes/10887-solr-field-types.md
```diff
@@ -0,0 +1,82 @@
...
+This change enables range queries when searching from both the UI and the API, such as `dateOfDeposit:[2000-01-01 TO 2014-12-31]` or `targetSampleActualSize:[25 TO 50]`.
+
+Dataverse administrators must update their Solr schema.xml (manually or by rerunning `update-fields.sh`) and reindex all datasets.
+
+Additionally, search result highlighting is now more accurate, ensuring that only fields relevant to the query are highlighted in search results. If the query is specifically limited to certain fields, the highlighting is now limited to those fields as well.
+
+## Upgrade Instructions
+
+7\. Update Solr schema.xml file. Start with the standard v6.5 schema.xml, then, if your installation uses any custom or experimental metadata blocks, update it to include the extra fields (step 7a).
+
```
ofahimIQSS, line 42:
```suggestion
Run the commands below as a non-root user.
```

doc/release-notes/10887-solr-field-types.md
```diff
@@ -0,0 +1,82 @@
...
+```shell
+service solr stop
+```
+
+Replace schema.xml
+
+```shell
+wget https://raw.githubusercontent.com/IQSS/dataverse/v6.5/conf/solr/schema.xml
+cp schema.xml /usr/local/solr/solr-9.4.1/server/solr/collection1/conf
```
ofahimIQSS, line 53:
```suggestion
sudo cp schema.xml /usr/local/solr/solr-9.4.1/server/solr/collection1/conf
```

doc/release-notes/10887-solr-field-types.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+Start Solr (but if you use any custom metadata blocks, perform the next step, 7a first).
+
+```shell
+service solr start
```
ofahimIQSS, line 59:
```suggestion
sudo service solr start
```

doc/release-notes/10887-solr-field-types.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+7a\. For installations with custom or experimental metadata blocks:
+
+Before starting Solr, update the schema to include all the extra metadata fields that your installation uses. We do this by collecting the output of the Dataverse schema API and feeding it to the `update-fields.sh` script that we supply, as in the example below (modify the command lines as needed to reflect the names of the directories, if different):
+
+```shell
+	wget https://raw.githubusercontent.com/IQSS/dataverse/v6.5/conf/solr/update-fields.sh
+	chmod +x update-fields.sh
+	curl "http://localhost:8080/api/admin/index/solr/schema" | ./update-fields.sh /usr/local/solr/solr-9.4.1/server/solr/collection1/conf/schema.xml
```
ofahimIQSS, line 69:
```suggestion
	curl "http://localhost:8080/api/admin/index/solr/schema" | sudo ./update-fields.sh /usr/local/solr/solr-9.4.1/server/solr/collection1/conf/schema.xml
```

ofahimIQSS:
Wanted to drop another observation (edge case) I found during testing related to display. When entering very long values in the meta data fields, the numbers go past the border instead of wrapping around. 
image
To test, populate the Solr Field Types Test Metadata as follows:
Integer: add as many 0's followed by a 2. ie. 0000...00002
Floating Point: 2.01000290329039 - copy and paste everything after the decimal to make it a long
Date: 2024-11-11 (or any valid date)

pdurbin:
@vera I'm afraid we had to bump the milestone to 6.6 but I think this feature will be very popular! Thanks again! ❤️

ofahimIQSS:
Testing Passed - Merging PR 

https://github.com/user-attachments/assets/37f50c13-3ab4-4cfc-b5e2-6efda9d66d73

pdurbin:
@vera and @ofahimIQSS can you please take a look at this follow-up PR I just made? Thanks.

- #11140

pdurbin:
Related:

- #401

PR #10901, opened 2024-10-01
Deaccessioned dataset file edit fix.

### LINKED ISSUE #9351: Editing a deaccessioned dataset: the file edit options should allow the "delete" function to work
From a deaccessioned version, any file-related action (deleting, restricting files, adding tags, etc), results in error banners and confusing success banners to appear and doesn't cause a draft version to be created

Visit a deaccessioned dataset to edit for republishing. login, and select file, delete files

select files and select the delete option, or any other option

* When does this issue occur?

When you attempt to revive/edit a deaccessioned dataset for republishing

Dataset landing page

* What happens?
* 
Error message that files can't be deleted/etc

curators, superusers

expected file actions to work and create new draft

latest release 5.13

none

**Screenshots:**

*

jp-tosca:
It was reported on [#9351](https://github.com/IQSS/dataverse/issues/9351) that trying to update any of the files for a deaccessioned dataset would result in an error message.

My test showed that this would only happen when the only previous version available is deaccessioned, for example if you create a dataset and have version 1, then make version 2 and deaccession version 2 it will not show this error. 

While reviewing this issue we found that even if the UI would fail this was possible on the API due a parameter sent that would cause to bypass a check, this caused that even the new test would pass with the old code. 

You can test this by: 

1. Deploy develop branch
2. Create a dataset with file
3. Publish the dataset
4. Deaccession the dataset
5. Try to delete a file and you will see an error
6. Get the id of the file (lmk if you need help with this step)
7. Try to delete the file with http://localhost:8080/api/files/ and you will succeed (you shouldn't)
8. Undeploy and make the changes to the file Files.java on your develop branch, deploy and repeat steps 1-7, you should get an error message on both.
9. This is where the fix comes in, once that the fix from this branch from UpdateDatasetVersionCommand are deployed the functionality should be restored. 

Closes #9351

Yes

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Dataset.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataset.java
@@ -333,13 +333,18 @@ public class Dataset extends DvObjectContainer {
         return getVersions().get(0);
     }
 
-    public DatasetVersion getLatestVersionForCopy() {
+    public DatasetVersion getLatestVersionForCopy(boolean includeDeaccessioned) {
         for (DatasetVersion testDsv : getVersions()) {
-            if (testDsv.isReleased() || testDsv.isArchived() || testDsv.isDeaccessioned()) {
+            if (testDsv.isReleased() || testDsv.isArchived() 
+                || (testDsv.isDeaccessioned() && includeDeaccessioned)) {
                 return testDsv;
             }
         }
         return getVersions().get(0);
+    }
+
+    public DatasetVersion getLatestVersionForCopy(){
+        return getLatestVersionForCopy(false);
     }
 
     public List<DatasetVersion> getVersions() {
--- a/src/main/java/edu/harvard/iq/dataverse/Dataset.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataset.java
@@ -335,7 +335,7 @@ public class Dataset extends DvObjectContainer {
 
     public DatasetVersion getLatestVersionForCopy() {
         for (DatasetVersion testDsv : getVersions()) {
-            if (testDsv.isReleased() || testDsv.isArchived()) {
+            if (testDsv.isReleased() || testDsv.isArchived() || testDsv.isDeaccessioned()) {
                 return testDsv;
             }
         }
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
@@ -115,7 +115,7 @@ public class UpdateDatasetVersionCommand extends AbstractDatasetCommand<Dataset>
          */
         if(persistedVersion==null) {
             Long id = getDataset().getLatestVersion().getId();
-            persistedVersion = ctxt.datasetVersion().find(id!=null ? id: getDataset().getLatestVersionForCopy().getId());
+            persistedVersion = ctxt.datasetVersion().find(id!=null ? id : getDataset().getLatestVersionForCopy().getId());
         }
         
         //Will throw an IllegalCommandException if a system metadatablock is changed and the appropriate key is not supplied.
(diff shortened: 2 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/Dataset.java
```diff
@@ -335,7 +335,7 @@ public DatasetVersion getLatestVersion() {
 
     public DatasetVersion getLatestVersionForCopy() {
         for (DatasetVersion testDsv : getVersions()) {
-            if (testDsv.isReleased() || testDsv.isArchived()) {
+            if (testDsv.isReleased() || testDsv.isArchived() || testDsv.isDeaccessioned()) {
```
qqmyers, line 338: This looks like it will fix the npe problem, but this method is used in many places and I'm not sure that all of them should include a deaccessioned version. For example - https://github.com/IQSS/dataverse/blob/4c1e069082b740849e3ad0ade41a926407ec52f5/src/main/java/edu/harvard/iq/dataverse/export/ExportService.java#L152 would export info about a deaccessioned version. We might need a separate getLatestPublicVersion method for such cases unless there's some other way to fix.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
```diff
@@ -115,7 +115,7 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
          */
         if(persistedVersion==null) {
             Long id = getDataset().getLatestVersion().getId();
-            persistedVersion = ctxt.datasetVersion().find(id!=null ? id: getDataset().getLatestVersionForCopy().getId());
+            persistedVersion = ctxt.datasetVersion().find(id!=null ? id : getDataset().getLatestVersionForCopy(true).getId());
```
qqmyers, line 118: Adding the true parameter so that only this call is affected resolved my earlier comment. Thinking further on this particular use  - we're getting the persistedVersion so we can see if there are changes to a system metadatablock (line 122) in which case you need a key to make changes. It's an interesting question whether that comparison should be versus the last public/published version if there is one or against a later deaccessioned one (if it exists).  Probably rare enough that we should just ignore my comment and accept the fix as is.

jp-tosca:
Changes should be good to review @qqmyers, the copy still needs to be created on the api but not sent as a parameter, after making these changes both UI and the API should fail before the changes and the change should resolve this issue for both.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -667,6 +667,60 @@ public void testCreatePublishDestroyDataset() {
         deleteDatasetResponse.prettyPrint();
         assertEquals(200, deleteDatasetResponse.getStatusCode());
 
+        // Start of deaccession test.
```
qqmyers: Wow - the testCreatePublishDestroyDataset method is doing quite a bit more than testing destroy! +1 for splitting it up some day.

qqmyers (approved):
This looks fine (I made minor comment changes). 
For posterity - this does change the currently inconsistent behavior between the UI and API for file deletion - the latter was triggering edit-draft logging while the former did not. It sounds like this isn't used much now, and, in #10890, I'm proposing further changes to this - basically making edit-draft logging an optional feature and then working to enable it for any/all operation that update the dataset version if enabled (versus it only happening for methods that send in a clone datasetversion now). So - in talking with @jp-tosca, we thought making the api consistent with the UI, which also allows an IT test for this bug, is a reasonable approach in this PR.

ofahimIQSS:
I tested this ticket a little bit differently. I pushed the ticket to internal to test the fix. Fix looks good and no error message was displayed after deleting a file from a deaccessioned dataset. I was able to reproduce the defect in demo (without the fix) so I could see what the issue was. I attached screenshots of the testing below.  
[10901 Testing.docx](https://github.com/user-attachments/files/17248300/10901.Testing.docx)

PR #10906, opened 2024-10-03
Optimize permission lookups for a user

### LINKED ISSUE #6467: Optimize permission lookups for a user
When we recently added the ability to modify the "Host Dataverse" upon creating a dataverse / dataset, we ran into the "what dataverses does User x have Permission y on" challenge (in this case add dataverse/ dataset permission).

Because of how our permission system<sup>*</sup> works with groups, there is currently no straightforward way to start from the user and get the dataverses for which that user has a certain permission, without having to get all dataverses and check each one (though we mitigated some for this by filtering the list with an autocomplete and only checking a subset of all dataverses).

<sup>*</sup>if we ever allowed permissions to cascade from one dataverse to its child, that would also present a challenge

@landreev and I discussed at the time that it would be nice to be able to start from the user direction, so this issue is to track the investigation / solution for how to make that work.

related #784

stevenwinship:
**What this PR does / why we need it**: The need to query "what dataverses does User x have Permission y on" 

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/6467

- Closes #6467

**Special notes for your reviewer**: I'm not crazy about the api being called /allowedcollections/. Any suggestions for a better name would be appreciated.

**Suggestions on how to test this**: See IT test. Also test a group within a group for nested group permissions. I have tested it manually and it was working. For Shib testing I tested the sql manually. This still needs to be tested in a running environment.

**Additional documentation**:


doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
```diff
@@ -0,0 +1,9 @@
+The following API have been added:
```
pdurbin, line 1: Once we've finalized the name of the API endpoint, please add it to the API Guide.

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
```diff
@@ -0,0 +1,9 @@
...
+
+/api/users/{identifier}/allowedcollections/{permission}
```
pdurbin, line 3: I can live with this name. I would probably camel case it as `allowedCollections`.

How would we extend this API to datasets or files, if we needed to? Like below?

/api/users/{identifier}/allowedDatasets/{permission}
/api/users/{identifier}/allowedFiles/{permission}

Just playing around below, maybe instead we could have...

/api/users/{identifier}/permissions/collections/{permission}
/api/users/{identifier}/permissions/datasets/{permission}
/api/users/{identifier}/permissions/files/{permission}

? You might want to get some other opinions.

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
```diff
@@ -0,0 +1,9 @@
...
+
+This API lists the dataverses/collections that the user has access to via the permission passed.
+By passing "any" as the permission the list will return all dataverse/collections that the user can access regardless of which permission is used.
+This API can be executed only by administrators.
```
pdurbin, line 7:
```suggestion
This API can be executed only by superusers.
```

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
```diff
@@ -0,0 +1,9 @@
...
+Valid Permissions are: AddDataverse, AddDataset, ViewUnpublishedDataverse, ViewUnpublishedDataset, DownloadFile, EditDataverse, EditDataset, ManageDataversePermissions,
+ManageDatasetPermissions, ManageFilePermissions, PublishDataverse, PublishDataset, DeleteDataverse, DeleteDatasetDraft, and "any" as a wildcard option.
```
pdurbin, line 9: It's fine to list these here but it might be nice to have an API to list all these permissions.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -100,6 +100,41 @@ public class PermissionServiceBean {
     @Inject
     DatasetVersionFilesServiceBean datasetVersionFilesServiceBean;
 
+    private static final String LIST_ALL_DATAVERSES_USER_HAS_PERMISSION = "WITH grouplist AS (\n" +
```
pdurbin, line 103: To avoid all the "\n" and for more maintainability we could use the fancy new text blocks: https://docs.oracle.com/en/java/javase/15/text-blocks/index.html

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
```diff
@@ -0,0 +1,9 @@
+The following API have been added:
+
+/api/users/{identifier}/allowedcollections/{permission}
+
+This API lists the dataverses/collections that the user has access to via the permission passed.
```
pdurbin, line 5: What about cases where the user is granted access at runtime based on their membership in Shibboleth groups or IP groups?

We have this related comment in the code:

```
/**
 * We don't expect this to support Shibboleth groups because even though
 * a Shibboleth user can have an API token the transient
 * shibIdentityProvider String on AuthenticatedUser is only set when a
 * SAML assertion is made at runtime via the browser.
 */
```

On a related note, that comment is above the call to this method on ServiceDocumentManagerImpl

`public List<Dataverse> getDataversesUserHasPermissionOn(AuthenticatedUser user, Permission permission)`

Should this method be replaced by the new one in this PR:

`public List<Dataverse> findPermittedCollections(AuthenticatedUser user, int permissionBit)`

?

src/main/java/edu/harvard/iq/dataverse/api/Users.java
```diff
@@ -260,5 +261,26 @@ public Response getTracesElement(@Context ContainerRequestContext crc, @Context
             return ex.getResponse();
         }
     }
-
+    @GET
+    @AuthRequired
+    @Path("{identifier}/allowedcollections/{permission}")
+    @Produces("text/csv, application/json")
```
pdurbin, line 267: Are we purposefully returning CSV as an option? Do we have a use case? All the tests seem to be for JSON.

pdurbin (commented):
Some initial feedback. My eyes glazed over at the SQL but I'll request a review from @scolapasta for that part.

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
stevenwinship: I wouldn't have bother listing them in the release note but I wanted to point out the "any" permission for a wildcard. But since this wasn't asked for maybe we just leave it out of the release notes?

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
stevenwinship: I'll fix that

src/main/java/edu/harvard/iq/dataverse/api/Users.java
stevenwinship: just json. I'll remove it

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
stevenwinship: I like the top APIs better. I think the allowedCollections etc. is more descriptive of what is being returned.

@qqmyers @scolapasta Any comment :)

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
pdurbin: Maybe, but do you think it's potentially useful to list these permissions via API? Maybe not for this PR but in the future.

The permissions should probably be listed in the guides for now. That way, the release note could link to them.

stevenwinship: An API would be nice. There is already a way to list the permissions in a role of the UI so adding a simple list of all permissions should be a quick add.

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
stevenwinship: I'll swap it out and test it

stevenwinship: @pdurbin I added IP Group support but unfortunately ServiceDocumentManagerImpl doesn't have access to the request and therefore the ip address of the caller. Unless you know of a way to get it?

pdurbin: I don't see an obvious way. However, the SWORD API has never supported IP Groups and to my knowledge nobody has asked for it, so I think it's ok.

The main use case for IP Groups is read-only access, such as walking into a library and having access to data because you are on the library's IP range.

src/main/java/edu/harvard/iq/dataverse/api/datadeposit/SWORDv2ServiceDocumentServlet.java
```diff
@@ -29,6 +31,13 @@ public void init() throws ServletException {
 
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        String ipAddress = req.getHeader("X-FORWARDED-FOR");
```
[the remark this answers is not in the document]
stevenwinship: Good Idea. Done.

doc/release-notes/6467-optimize-permission-lookups-for-a-user.md
scolapasta: I'm fine with either set of names.

scolapasta (commented):
Hi, finally am managing to review - looks good overall, a few questions, easier to ask as a general comment imo:

1. Does the sql query deal with nested groups (I think so, with the "With" but wanted to confrim?
2. Does the last union (ip groups) need the exists / permissions bit clause?
3. Do we have any sense on performace of this query yet?

stevenwinship:
> Hi, finally am managing to review - looks good overall, a few questions, easier to ask as a general comment imo:
> 
> 1. Does the sql query deal with nested groups (I think so, with the "With" but wanted to confrim?
> 2. Does the last union (ip groups) need the exists / permissions bit clause?
> 3. Do we have any sense on performace of this query yet?

1. Yes it handles groups within groups. I did test that
2. Sorry it's a bit confusing. The permission bit part is added in the"AND @IPRANGESQL" by calling findPermittedCollections
3. There hasn't been any formal performance testing but local tests as well as testing on the perf server were pretty good.

scolapasta:
> > Hi, finally am managing to review - looks good overall, a few questions, easier to ask as a general comment imo:
> > 
> > 1. Does the sql query deal with nested groups (I think so, with the "With" but wanted to confrim?
> > 2. Does the last union (ip groups) need the exists / permissions bit clause?
> > 3. Do we have any sense on performace of this query yet?
> 
> 1. Yes it handles groups within groups. I did test that
> 2. Sorry it's a bit confusing. The permission bit part is added in the"AND @IPRANGESQL" by calling findPermittedCollections
> 3. There hasn't been any formal performance testing but local tests as well as testing on the perf server were pretty good.

OK on 1 and 3. For 2, maybe I'm still missing it - for the query part added in lines 143-152 for the ip groups, I don't see any reference to permisison bits).

stevenwinship:
> OK on 1 and 3. For 2, maybe I'm still missing it - for the query part added in lines 143-152 for the ip groups, I don't see any reference to permisison bits).

Ok. You are correct. there is no permission bits in IP groups. This is a bit confusing but here goes

in regular groups:
"_roleAlias": "contributor", 
where contributor has permissions where each bit denotes a permission (for file download bit 5  is set)
sql: (dataverserole.permissionbits & 16 !=0)

in IP Groups the role is the specific permission so there is no need to check the bit
sql: look for 'fileDownloader' in WHERE roleassignment.assigneeidentifier IN (SELECT CONCAT('&ip/', persistedglobalgroup.persistedgroupalias) as assignee...) 
{
    "status": "OK",
    "data": {
        "assignee": "&ip/ipGroupebc7aa00",
        "roleId": 2,
        **"_roleAlias": "fileDownloader",**
        "definitionPointId": 4
    }
}

scolapasta:
> in IP Groups the role is the specific permission so there is no need to check the bit

I don't think that's accurate (or at least now how it's designed. Sure, IP groups are mostly meant to allow read access (and usually file download), BUT it could be for the role "viewer". 

And technically, if you're logged in and from a certain ip address, then contributor role should work too. (as examples)

scolapasta (approved):

ofahimIQSS:
This PR has conflicts which need to be resolved.

ofahimIQSS:
API Testing done - no issues found. 
[OptimizePermAPItests.txt](https://github.com/user-attachments/files/18697311/OptimizePermAPItests.txt)

ofahimIQSS:
Testing Passed - Merging PR

PR #10926, opened 2024-10-15
Fix relationType display value bug

qqmyers:
**What this PR does / why we need it**: The [recent changes](https://github.com/IQSS/dataverse/pull/10632) to include more information in the XML sent to DataCite, including the new relationType child field for related publications, is sending the display value of that field rather than the real CVV value. DataCite is rejecting values that do not match it's schema. In English, any of the values that are multiple words, e.g. `IsSupplementTo`, having the display value `Is Supplement To`, cause DOI updates to fail and therefor for dataset publication to fail. (In other languages, presumably all values would fail.) Dataset that do not have a Related Publication entry, or have a default null value for the relationType, are unaffected (nulls are turned into an IsSupplementTo value in code which does not rely on the display value).

This PR make a simple change to include the CVV value itself rather than display value. 

- Closes #

**Special notes for your reviewer**:The change is made in the DatasetVersion method which is used to create jsonld and json for other purposed - the other uses right now do not use the relationType value at all, but, if/when they do, it seems like the real CVV value is what should be sent.

**Suggestions on how to test this**: Configure Dataverse to use DataCite, add a publication with the Is Supplement To relationType and make sure publication succeeds, that there is no error in the log, that DataCite Fabrica shows the information as expected.

**Is there a release notes update needed for this change?**: This is probably patch/announce worthy.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
@@ -1390,7 +1390,10 @@ public class DatasetVersion implements Serializable {
                                 relatedPublication.setIdNumber(subField.getDisplayValue());
                                 break;
                             case DatasetFieldConstant.publicationRelationType:
-                                relatedPublication.setRelationType(subField.getDisplayValue());
+                                List<String> values = subField.getValues_nondisplay();
+                                if (!values.isEmpty()) {
+                                    relatedPublication.setRelationType(values.get(0)); //only one value allowed
+                                }
                                 break;
                         }
                     }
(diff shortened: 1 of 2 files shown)


landreev (approved):
Nice test.

ofahimIQSS:
Hi @qqmyers - when I tested this PR in Internal, I could not see the "Relation Type" field on the UI of edit dataset metadata page so I couldn't test this. Please see screenshot below:

image

Steps: 
1. Go to https://dataverse-internal.iq.harvard.edu/
2. Create Dataset
3. Scroll to Related Publication Section

Expected Result: Relation Type field should be displayed under Related Publication section.

qqmyers:
I'm guessing that step 6. Update metadata blocks of the [Upgrade instructions for v6.4](https://github.com/IQSS/dataverse/releases/tag/v6.4) wasn't done on that machine. Once the new citation metadata block is installed, the field should be there. (Could then be issues with indexing if step 7 update solr schema wasn't done.)

ofahimIQSS:
Testing Complete - fix looks good. Merging PR. 
[Testing of 10926.docx](https://github.com/user-attachments/files/17637825/Testing.of.10926.docx)

PR #10938, opened 2024-10-18
adding description info to the fileDsc seciton in DDI CodeBook. #5051

### LINKED ISSUE #5051: File description metadata of ingested files are not in the DDI exported metadata
I've seen this omission/bug since at least Dataverse version 4.9 and verified it in Dataverse version 6.1:

In the exported metadata of datasets with ingested files, for files that Dataverse ingests, the file description text (which depositors enter for uploaded files) are not in the exported DDI.xml. For example, for the dataset at https://doi.org/10.7910/DVN/1ZPAKL, the ingested file's file description, "Data from vignette survey experiment conducted in Denmark in June 2023", is not in [the exported DDI](https://dataverse.harvard.edu/api/datasets/export?exporter=ddi&persistentId=doi:10.7910/DVN/1ZPAKL).

I think next steps would be to figure out if:
- this is a bug, so the file descriptions of ingested files should be in the DDI, or
- this is just an omission, and we have to figure out where in the DDI xml the file description should be

landreev:

(the PR is a draft for now)

Apparently, users have been asking for this since 2018 - for tabular files that have the Description field populated, this label was never exported in the DDI (non-ingested files always had their descriptions exported, in the corresponding `<otherMat>` sections). 
There is no obvious field under `<fileDscr>` in the DDI Codebook schema for it - probably the reason we chose not to export it back in the day (?) - but putting it into another dedicated free text `<note>` field seems like a reasonable solution. 

- Closes #5051

**Additional documentation**:


src/main/java/propertyFiles/Bundle.properties
```diff
@@ -1934,7 +1934,7 @@ file.downloadBtn.format.all=All File Formats + Information
 file.downloadBtn.format.tab=Tab-Delimited
 file.downloadBtn.format.original={0} (Original File Format)
 file.downloadBtn.format.rdata=RData
-file.downloadBtn.format.var=Variable Metadata
+file.downloadBtn.format.var=DDI Codebook v2
 file.downloadBtn.format.citation=Data File Citation
 file.download.filetype.unknown=Original File Format
 file.more.information.link=Link to more file information for
```
pdurbin, line 1940: I'm putting this here at the end... code coverage is ever so slightly down. At our last monthly meeting we talked about trying to move it up. Maybe we should add tests.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -1934,7 +1934,7 @@ file.downloadBtn.format.all=All File Formats + Information
 file.downloadBtn.format.tab=Tab-Delimited
 file.downloadBtn.format.original={0} (Original File Format)
 file.downloadBtn.format.rdata=RData
-file.downloadBtn.format.var=Variable Metadata
+file.downloadBtn.format.var=DDI Codebook v2
```
pdurbin, line 1937: I'm happy to see this cleanup, this renaming to "DDI Codebook v2". Can we mention it in a release note (along with the feature in general)?

pdurbin (commented):
Overall, looks fine. I'm trusting that "notes" is the right place to put the file descriptions. I did leave a couple other comments.

src/main/java/propertyFiles/Bundle.properties
landreev: I started adding something to the DDI export test; but then stopped myself, seeing how the issue was a "3", and I was also told to focus on something urgent-ish r/n.

src/main/java/propertyFiles/Bundle.properties
landreev: > I'm happy to see this cleanup, this renaming to "DDI Codebook v2". Can we mention it in a release note (along with the feature in general)?

I'm generally very pro-documentation, and usually an eager writer of release notes... But this feels like a case where I'm inclined to push back. There is no "feature" really - this would honestly be like adding a release note on having fixed a typo in the guide.  This PR is addressing a minor inconsistency, something we potentially know about a file that we weren't exporting; but it's worth noting that nobody has ever asked for it - it was just something that Julian noticed (6 years ago, lol). Same for editing the UI labels - I just clarified that the DDI format we produce is in fact DDI version such and such. Doesn't seem like release note material - ?

It also feels like it I went over the initially allocated "size 3" just writing this paragraph. :)

pdurbin: Fair enough. It'll be a nice surprise for people who are really paying attention. 😄

src/main/java/propertyFiles/Bundle.properties
pdurbin: Ok. Just doing my due diligence. Resolving. 😄

stevenwinship (approved):

ofahimIQSS:
Hey @landreev, on trying to test this I had an observation. 

When I upload a CSV file, I am able to see the correct DDI output (see below):
image

When I tested uploading a Stata file, I noticed that the description appears under <fileDscr>
image

Here is the Stata Test File (I compressed it so I could add it here, please unzip it first): 
[ATHLET2.DTA.zip](https://github.com/user-attachments/files/17624033/ATHLET2.DTA.zip)

landreev:
@ofahimIQSS 
Sorry for the delay, I pretty much forgot about this PR. 
The fact that the 2 files are differently formatted in the DDI output is simply because only the Stata file did get successfully ingested as a tabular file. I.e., Dataverse was able to successfully parse the file and extract the extra metadata from it that describes the individual variable columns and calculate summary statistics (see the guides for more information about ingest). The `<fileDscr>` section in the DDI is only used for files that we were able to successfully ingest. 

We cannot guarantee to be able to successfully ingest any file in a _potentially_ "ingestable" file format. This is especially true with CSV. We are going to try to parse and ingest any CSV file uploaded by the user (as long as it is below the ingestable size limit, if defined), but it may or may not succeed. Some of the more common reasons why we may fail to ingest a CSV file: ingest will stop if the first line is not a comma-separated list of what looks like the names of the individual variables;  ingest will fail unless every row contains the same number of comma-separated fields. 

The `<fileDscr>` entry for the ingested Stata file in the screenshot looks correct. (the whole point of the small change in this PR was to add an extra `<notes>` field with the file description). 
There are 2 `<notes>` entries because they are used for 2 different things - one for the UNF (the data signature we calculate for tab. files), the other one is the new note for the file description. `<notes>` is a field in the DDI that can be used for any arbitrary text. We use it to encode any information for which we do not have a specifically reserved, dedicated field in the DDI.

ofahimIQSS:
Thanks for the clarification @landreev - Merging PR
[Testing of 10938.docx](https://github.com/user-attachments/files/17737947/Testing.of.10938.docx)

PR #10958, opened 2024-10-23
Fixes updateAttributes endpoint

GPortas:
- Closes: https://github.com/IQSS/dataverse-security/issues/98

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
@@ -1,0 +1,110 @@
+package edu.harvard.iq.dataverse.engine.command.impl;
+
+import edu.harvard.iq.dataverse.Dataverse;
+import edu.harvard.iq.dataverse.authorization.Permission;
+import edu.harvard.iq.dataverse.engine.command.AbstractCommand;
+import edu.harvard.iq.dataverse.engine.command.CommandContext;
+import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
+import edu.harvard.iq.dataverse.engine.command.RequiredPermissions;
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+import edu.harvard.iq.dataverse.engine.command.exception.IllegalCommandException;
+import edu.harvard.iq.dataverse.engine.command.exception.PermissionException;
+import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
+
+import java.util.Collections;
+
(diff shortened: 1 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
```diff
@@ -0,0 +1,110 @@
...
+     * @throws PermissionException if the user doesn't have permission to modify this attribute.
+     */
+    private void setBooleanAttributeForFilePIDs(CommandContext ctxt) throws CommandException {
+        if (!getRequest().getUser().isSuperuser()) {
+            throw new PermissionException("You must be a superuser to change this setting",
+                    this, Collections.singleton(Permission.EditDataset), dataverse);
+        }
+        if (!ctxt.settings().isTrueForKey(SettingsServiceBean.Key.AllowEnablingFilePIDsPerCollection, false)) {
+            throw new PermissionException("Changing File PID policy per collection is not enabled on this server",
+                    this, Collections.singleton(Permission.EditDataset), dataverse);
```
sekmiller, line 101: Minor point, but if we get here we have a super user so wouldn't it be an illegal command exception - not a permission exception?

GPortas: I initially thought the same, but I chose to use the PermissionException since that was the original exception thrown before moving this logic into the command. It's not a significant backward compatibility issue, but that was the reason behind my decision. However, I can switch to the IllegalCommandException if you think it fits better.

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -837,30 +836,70 @@ public void testAttributesApi() throws Exception {
...
         Response deleteCollectionResponse = UtilIT.deleteDataverse(newCollectionAlias, apiToken);
-        deleteCollectionResponse.prettyPrint();
         assertEquals(OK.getStatusCode(), deleteCollectionResponse.getStatusCode());
+
+        // Cannot update root collection from a regular user:
+
+        changeAttributeResp = UtilIT.setCollectionAttribute("root", "name", newCollectionName, apiToken);
+        changeAttributeResp.then().assertThat()
+                .statusCode(UNAUTHORIZED.getStatusCode());
+
```
sekmiller, line 897: You're not asserting anything that is particular to Root here, correct? Any random user without Edit Dataverse permissions would get the same Unauthorized on any Dataverse

sekmiller (approved):
Looks good. Thanks for responding to my questions.

PR #10967, opened 2024-10-25
Update to #10705 fixing a blank filter query

qqmyers:
**What this PR does / why we need it**: Per the comment in #10705, the original PR could add a (hopefully harmless) blank filter query to the request. This PR fixes the logic - the method generating the queries returned "" when none were needed but the check to include them was for the string to != null. The PR changes to check for not null, empty, or blank.

- Closes #

**Suggestions on how to test this**: Verify that an empty query is not included.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
@@ -1047,7 +1047,7 @@ public class SearchServiceBean {
         // add joins on all the non-public groups that may exist for the
         // user:
 
-        // Authenticated users - and GuestUser - may be part of one or more groups; such
+        // Authenticated users, *and the GuestUser*, may be part of one or more groups; such
         // as IP Groups.
         groups = groupService.collectAncestors(groupService.groupsFor(dataverseRequest));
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
```diff
@@ -1047,7 +1047,7 @@ private String getPermissionFilterQuery(DataverseRequest dataverseRequest, SolrQ
         // add joins on all the non-public groups that may exist for the
         // user:
 
-        // Authenticated users and GuestUser may be part of one or more groups; such
+        // Authenticated users - and GuestUser - may be part of one or more groups; such
```
pdurbin, line 1050:
```suggestion
        // Authenticated users, and GuestUser, may be part of one or more groups; such
```

Or maybe "(and GuestUser)"? The dashes throw me off a bit.

pdurbin (commented):
The code looks fine, though I didn't test it. Added a small suggestion.

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
qqmyers: Should have mentioned - this was in response to https://github.com/IQSS/dataverse/pull/10706#discussion_r1815453266 indicating that it's useful to emphasize that the GuestUser can also be in groups. I tried a switch to using * instead.

landreev:
I hope it's ok for me to just approve and merge it quickly; since this is a straightforward followup to the #10705 that I QAed and merged yesterday. @pdurbin unless you want to click "approve" as well.

landreev (approved):

pdurbin (approved):
I didn't test this, but it looks good.

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
pdurbin: Looks great now. Very clear. Thanks. Approved.

PR #10971, opened 2024-10-25
Api auth required openapi

### LINKED ISSUE #10970: Feature Request:
There was a conversation on [zullip](https://dataverse.zulipchat.com/#narrow/channel/446770-docs/topic/API.20docs.20from.20OpenAPI/near/477786826) at for the need of documenting the authentication requirements on the OpenAPI specification.

jp-tosca:
There was a conversation on [zullip](https://dataverse.zulipchat.com/#narrow/channel/446770-docs/topic/API.20docs.20from.20OpenAPI/near/477786826) at for the need of documenting the authentication requirements on the OpenAPI specification. 

- Closes #10970


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1269,6 +1270,7 @@ private String getWebappImageResource(String imageName) {
     })
     @Tag(name = "saveAuxiliaryFileWithVersion", 
          description = "Save Auxiliary File With Version")
+    @SecurityRequirement(name = "api_token")
```
pdurbin, line 1273: Why is it "api_token" and not "API token" or "API_Token" or whatever else? Is it arbitrary?

jp-tosca: Seems that we also have to define the "security schema" and we name that security schema, we also set a type. The security schema could be named something else but in this case api_key is the name

pdurbin:
Soft close for now. We might want this in the future.

PR #10278, opened 2024-01-29
WIP: PoC OIDC Signup (create user account) via API call

johannes-darms:
**What this PR does / why we need it**:
A PoC that shows how a user account can be created via an API call. This requires the configuration of an OIDC Authentication provider.    

**Which issue(s) this PR closes**:

Closes #

**Special notes for your reviewer**:
The client in keycloak must be adapted to enabled and enforce the consent 
Screenshot 2024-01-29 at 10 26 57

**Suggestions on how to test this**:

https://github.com/IQSS/dataverse/assets/8245181/5583669c-07f3-4306-b120-591c2726a3b6

Once the user is created. The consent is stored in keycloak.
Screenshot 2024-01-29 at 10 29 30

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #10972, opened 2024-10-28
Handle unregistered users in BearerTokenAuthMechanism and implement user registration mechanism

### LINKED ISSUE #10959: Handle unregistered users in BearerTokenAuthMechanism

Based on the design implemented in the OIDC PoC for the SPA, we need to manage the case in BearerTokenAuthMechanism where there is no registered user account in Dataverse, even though the token is valid in the identity provider. 

Different ways to handle this have been discussed, from creating the account transparently to the user to returning some type of response to the API caller indicating that it is necessary to accept the terms of use before registering. This issue therefore involves an initial phase of analysis before implementing the final solution.

In the PoC implemented, we are simply creating the user from the JWT claims if the user is not registered: https://github.com/IQSS/dataverse/pull/10910

SPA user / API user

- https://github.com/IQSS/dataverse-frontend/pull/516

- BearerTokenAuthMechanism when user is authenticated but no signed up

- BearerTokenAuthMechanism account creation

- https://github.com/IQSS/dataverse-frontend/pull/516
- https://github.com/IQSS/dataverse/pull/10910

Yes

GPortas:

This PR includes a mechanism for registering new users in Dataverse through a Bearer Token. 

**BearerTokenAuthMechanism**

The Bearer Token authentication functionality (available through feature flag) has been extended to properly handle cases where BearerTokenAuthMechanism successfully validates the token but cannot identify any Dataverse user because there is no account associated with the token. See the current development version of  BearerTokenAuthMechanism. (https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/auth/BearerTokenAuthMechanism.java#L58).

The BearerTokenAuthMechanism has been extended to return a meaningful error in the scenario described above. Specifically, the calling user would receive a 403 error, indicating that the token is validated but the user is not registered in Dataverse.

Additionally, the entire BearerTokenAuthMechanism class has been refactored, moving authentication and token verification logic to AuthService, since this logic is also utilized by the new user registration functionality included in this PR, and it also results in a simpler version of BearerTokenAuthMechanism, more testable and with more limited responsability.

**User registration endpoint**

The new endpoint takes the bearer token and a UserDTO object with the required properties to create a user account in Dataverse. It only works if the bearer token feature flag is enabled. A RegisterOidcUserCommand handles the validation of the bearer token and user data, and the account creation process. It can return various errors, including a new InvalidFieldsCommandException, which lists any validation errors in the UserDTO. This exception is now handled by AbstractApiBean like other command exceptions and can be used for future purposes.

- Closes #10959 

- See https://github.com/IQSS/dataverse/pull/10972#issuecomment-2469982188 to understand why the OIDC UserInfo endpoint is not used in this design

- The IT for the new registration endpoint is disabled because it depends on the containerized environment, which is not available in Jenkins. Additionally, the feature flag would not be active, so it would fail for this reason as well. Changes have been made to the Keycloak realm JSON to suit the needs of this IT, particularly the modification of roles in the test realm to allow an admin user to create a random test user in the IT setup.

Visual inspection of the API integration tests. Overall, this PR is extensively tested with unit and integration tests, but it is in the ITs where we can see how the endpoint calls work and the different responses that can be obtained.

We can also run this branch locally and reproduce the same calls made in the ITs using curl.

No

Yes, attached.

**Additional documentation**:

None


johannes-darms:
@GPortas Thanks for the PR!

There is an [UserInfo Endpoint ](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) to obtain information (so called `claims`) about an OIDC user and in many cases properties are already embeeded into the passed Token and just need to be extracted. Further, it is possible to request consent before transmission of personal data to a thrid party (https://openid.net/specs/openid-connect-core-1_0.html#Consent). Hence, the OIDC specificaiton provides all the needed parts to create user accounts. 

I'd rather have an implementation that's inline with the specification than a custom JSON that's sent to an API, because someone needs to get the information for the AuthorisationServer, transform it into the custom JSON and initiate the registration request. This extra code is unnecessary IMHO. There is already a working [WIP] (https://github.com/IQSS/dataverse/pull/10278) for this.

GPortas:
> @GPortas Thanks for the PR!
> 
> There is an [UserInfo Endpoint ](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) to obtain information (so called `claims`) about an OIDC user and in many cases properties are already embeeded into the passed Token and just need to be extracted. Further, it is possible to request consent before transmission of personal data to a thrid party (https://openid.net/specs/openid-connect-core-1_0.html#Consent). Hence, the OIDC specificaiton provides all the needed parts to create user accounts.
> 
> I'd rather have an implementation that's inline with the specification than a custom JSON that's sent to an API, because someone needs to get the information for the AuthorisationServer, transform it into the custom JSON and initiate the registration request. This extra code is unnecessary IMHO. There is already a working [WIP] (#10278) for this.

Hi @johannes-darms,

Thank you for your message. Since certain claims in the OIDC UserInfo endpoint—such as email—are optional and can vary by identity provider, I’ve decided not to rely on this endpoint for retrieving user data. Consistent access to the user’s email is necessary for registration in Dataverse, so I’ve implemented a custom endpoint to ensure we receive all required information reliably during user registration.

In many cases, such as when using Keycloak, user information can be obtained directly from the JWT access token on the client side (e.g., email, first name, last name), as the necessary claims are included—eliminating the need for additional calls. If the token does not include these claims, we can then request the information from the identity provider’s UserInfo endpoint (if supported), or manually complete any missing data.

johannes-darms:
The set of [standard claims]( https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)list `email` so its IHMO save to assume this property is provided. Do you have an example where this property is not provided?

I completely understand your point and the reasoning behind it. However, with this approach we lose a bit of the interoperability that the use of standards and their protocols gives us. As an alternative, we could tell people that their authentication server must support these claims in order to be used with Dataverse?

GPortas:
@johannes-darms 

While the email claim is part of the OIDC standard, not all identity providers include it by default. Some IdPs (e.g., Keycloak, Auth0) allow administrators to configure which claims are returned so it is up to the IdP’s configuration, user consent, and requested scopes whether or not the email claim is made available.

Forcing authentication servers to support certain claims could generate pushback, as decentralized authentication would then depend on the specific needs of one application (Dataverse).

qqmyers:
FWIW: Both Google and ORCID allow people to create accounts where there is no email available (Google doesn't require it, ORCID allows you to not share it publicly/with apps). Google further allows accounts w/o a last name. At QDR, we use the UserInfo endpoint, but we still need a form to get email/last name if they don't exist before we can create an account. (QDR code isn't directly relevant as we do single sign on with Drupal and the form is on that side rather than Dataverse.) It looked like it may be possible to have Keycloak ask the user for missing information, which could have avoided custom code on our end, but we already had the Drupal form and have some other ~demographic questions we ask, so it was easier to do it there.

src/main/java/edu/harvard/iq/dataverse/api/Users.java
```diff
@@ -266,4 +270,24 @@ public Response getTracesElement(@Context ContainerRequestContext crc, @Context
...
+    public Response registerOidcUser(String body) {
+        if (!FeatureFlags.API_BEARER_AUTH.enabled()) {
+            return error(Response.Status.INTERNAL_SERVER_ERROR, BundleUtil.getStringFromBundle("users.api.errors.bearerAuthFeatureFlagDisabled"));
+        }
+        Optional<String> bearerToken = extractBearerTokenFromHeaderParam(httpRequest.getHeader(HttpHeaders.AUTHORIZATION));
+        if (bearerToken.isEmpty()) {
+            return error(Response.Status.BAD_REQUEST, BundleUtil.getStringFromBundle("users.api.errors.bearerTokenRequired"));
+        }
+        try {
+            JsonObject userJson = JsonUtil.getJsonObject(body);
```
qqmyers, line 284: Can the json be optional/only used if needed? It seems odd (possibly dangerous?) to let someone use a token with one name/email and register an account for a different name/email. Further on in the code, the token is being used to call the UserInfo endpoint, so everything the provider shares is available.

GPortas: @qqmyers Are you suggesting that we send optional "preferred" fields, such as preferredEmailAddress and preferredUsername, and only use them when the IdP is unable to provide us with this information?

I’m concerned that a user might believe they registered with a specific email and username, while different values were actually saved. However, this issue could be avoided if the SPA retrieves these values from the IdP or from JWT token claims (if available) and displays them in the new user registration form as uneditable fields.

qqmyers: I was thinking more that sending a name/email that is inconsistent with the OIDC account would fail, and that you wouldn't need to send JSON at all if the OIDC account includes all needed fields / only send the fields missing from the OIDC account.

johannes-darms:
As I said, I understand your reasons and they make sense. However, it is a very pessimistic and certainly necessary fallback implementation to assume that the authentication server will not fulfil the requirements and the user will have to provide most, if not all, of the information. But in many cases the authentication server provides all the information, or at least can be customised to provide the information needed (In keycloack I can configure the claims per client), and in these cases a dedicated json is not required at all and account generation could work without user interaction or forms, and this advantage is unfortunately not fully exploited with this implementation.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
```diff
@@ -0,0 +1,128 @@
...
+                throw new IllegalCommandException(BundleUtil.getStringFromBundle("registerOidcUserCommand.errors.userAlreadyRegisteredWithToken"), this);
+            }
+
+            UserInfo userClaimsInfo = oidcUserInfo.getUserClaimsInfo();
+
+            // Update the UserDTO object with available OIDC user claims; keep existing values if claims are absent
+            userDTO.setUsername(getValueOrDefault(userClaimsInfo.getPreferredUsername(), userDTO.getUsername()));
+            userDTO.setFirstName(getValueOrDefault(userClaimsInfo.getGivenName(), userDTO.getFirstName()));
+            userDTO.setLastName(getValueOrDefault(userClaimsInfo.getFamilyName(), userDTO.getLastName()));
+            userDTO.setEmailAddress(getValueOrDefault(userClaimsInfo.getEmailAddress(), userDTO.getEmailAddress()));
```
johannes-darms, line 48: Could we also try to use the[ ` email_verified`]( https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim)?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
```diff
@@ -0,0 +1,128 @@
...
+            Map<String, String> fieldErrors = validateUserFields(ctxt);
+            if (!fieldErrors.isEmpty()) {
+                throw new InvalidFieldsCommandException(
+                        BundleUtil.getStringFromBundle("registerOidcUserCommand.errors.invalidFields"),
+                        this,
+                        fieldErrors
+                );
+            }
+
+            ctxt.authentication().createAuthenticatedUser(userRecordIdentifier, userDTO.getUsername(), userDisplayInfo, true);
```
johannes-darms, line 67: Would its make sense to use the `updated_at` claim to somehow know when we need to update our copy of user properties?

GPortas: Your suggestion is interesting, though it introduces some complexity in my opinion. 

We'd need to determine the behavior if new claims are either available or unavailable for use within Dataverse. For example, if a username changes in the IdP but is already taken by another user in Dataverse, we’d face a collision right?. 

Additionally, where would be the ideal place to check for these updates? Should we verify on each request using the token, or at regular intervals? Anyway I see this more as a future iteration than something to figure out and implement now.

johannes-darms: Yes, this is definitely a future requirement. During login/API requests, an out-of-bound request to the userinfo endpoint is required if the `update_at` property within the JWS is newer, or similarly an alternative process probality time based if the `update_at` is not available. 

And yes, we have the same problems as with account creation, and we won't solve them until we drop unique constraints. This has some implications. For most properties it does not matter, but once we miss updates on the email property from the OIDC server we cannot reach the user. This is a problem we already have ....

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
johannes-darms: This is a feature that is currently only available for Shibboleth and not that visable. 

https://github.com/IQSS/dataverse/blob/f838e76baa055243a80331be57543060ae25e6a9/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java#L635

doc/sphinx-guides/source/api/auth.rst
```diff
@@ -81,6 +81,29 @@ To test if bearer tokens are working, you can try something like the following (
...
+Beyond the ``api-bearer-auth`` feature flag, there is another flag called ``api-bearer-auth-json-claims`` that can be enabled to allow sending missing user claims in the registration JSON. This is useful when the identity provider does not supply the necessary claims. However, this flag will only be considered if the ``api-bearer-auth`` feature flag is enabled. If the latter is not enabled, the ``api-bearer-auth-json-claims`` flag will be ignored.
+
+With the ``api-bearer-auth`` feature flag enabled, you can include the following properties in the request JSON:
+
+- ``username``
+- ``firstName``
+- ``lastName``
+- ``emailAddress``
+
+Note that even if they are included in the JSON, if it is possible to retrieve the corresponding claims from the identity provider, these values will be ignored and the ones from the identity provider will be used instead.
```
qqmyers, line 103: If I understood, in the comments, you suggested that ignoring the values could be confusing. Would it be better to just fail if you send info that is already in the token? (+1 from me)

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3343,6 +3343,12 @@ please find all known feature flags below. Any of these flags can be activated u
     * - api-session-auth
       - Enables API authentication via session cookie (JSESSIONID). **Caution: Enabling this feature flag exposes the installation to CSRF risks!** We expect this feature flag to be temporary (only used by frontend developers, see `#9063 <https://github.com/IQSS/dataverse/issues/9063>`_) and for the feature to be removed in the future.
       - ``Off``
+    * - api-bearer-auth
```
qqmyers, line 3346: FWIW: I originally thought that there would be a separate api-bearer-registration flag rather than overloading this flag. Is it always true that if people allow bearer tokens that also want people to be able to create new accounts this way? I'm not sure, but unless someone thinks this use case is real, I'm OK with it as is with one flag.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -978,4 +976,73 @@ public ApiToken getValidApiTokenForUser(User user) {
...
+            if (providers.isEmpty()) {
+                logger.log(Level.WARNING, "Bearer token detected, no OIDC provider configured");
+                throw new AuthorizationException(BundleUtil.getStringFromBundle("authenticationServiceBean.errors.bearerTokenDetectedNoOIDCProviderConfigured"));
+            }
+
+            // Attempt to validate the token with each configured OIDC provider.
+            for (OIDCAuthProvider provider : providers) {
+                try {
+                    // Retrieve both user identifier and user info
+                    Optional<UserRecordIdentifier> userRecordIdentifier = provider.getUserIdentifier(accessToken);
```
qqmyers, line 1016: getUserIdentifier calls getUserInfo at (https://github.com/IQSS/dataverse/blob/bbf29cc42bf6b897ffc7903a1f5bc87052d4071f/src/main/java/edu/harvard/iq/dataverse/authorization/providers/oauth2/oidc/OIDCAuthProvider.java#L336) - can the code be refactored to avoid calling it twice? Maybe calling getUserInfo first and then just getUserRecord ?

FWIW: I was concerned that I didn't see the access token being validated anywhere, but since the callback to the userinfo has to succeed for either the getUserIdentifier or getUserInfo to succeed, we know that the token is valid.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
```diff
@@ -0,0 +1,129 @@
...
+    private void updateUserDTO(UserInfo userClaimsInfo, boolean provideMissingClaimsEnabled) {
+        if (provideMissingClaimsEnabled) {
+            // Update with available OIDC claims, keep existing values if claims are absent
+            userDTO.setUsername(getValueOrDefault(userClaimsInfo.getPreferredUsername(), userDTO.getUsername()));
+            userDTO.setFirstName(getValueOrDefault(userClaimsInfo.getGivenName(), userDTO.getFirstName()));
+            userDTO.setLastName(getValueOrDefault(userClaimsInfo.getFamilyName(), userDTO.getLastName()));
+            userDTO.setEmailAddress(getValueOrDefault(userClaimsInfo.getEmailAddress(), userDTO.getEmailAddress()));
+        } else {
+            // Always use the claims provided by the OIDC provider, regardless of whether they are null or not
+            userDTO.setUsername(userClaimsInfo.getPreferredUsername());
```
qqmyers, line 81: Per the comment earlier - should we fail here if you send JSON with these values/with conflicting values?

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -3062,3 +3062,31 @@ openapi.exception.invalid.format=Invalid format {0}, currently supported formats
...
+users.api.errors.bearerAuthFeatureFlagDisabled=This endpoint is only available when bearer authentication feature flag is enabled.
+users.api.errors.bearerTokenRequired=Bearer token required.
+users.api.errors.jsonParseToUserDTO=Error parsing the POSTed User json: {0}
+users.api.userRegistered=User registered.
+
+#RegisterOidcUserCommand.java
+registerOidcUserCommand.errors.userAlreadyRegisteredWithToken=User is already registered with this token.
+registerOidcUserCommand.errors.invalidFields=The provided fields are invalid for registering a new user.
+registerOidcUserCommand.errors.userShouldAcceptTerms=Terms should be accepted.
+registerOidcUserCommand.errors.provideMissingClaimsEnabled.emailAddressFieldRequired=It is required to include an emailAddress field in the request JSON for registering the user.
```
qqmyers, line 3075: This results in a lot of duplicate text. Can you parameterize in some way to avoid it. E.g. either have a message that 'the following fields must be provided in the (token from the identity provider, in the request JSON)' or 'the field (name) must be provided...' ?
Also - I don't think users can/should contact the identity provider - that should probably be the standard contact your Dataverse admin/support team/whatever we usually say.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -978,4 +976,73 @@ public ApiToken getValidApiTokenForUser(User user) {
...
+                    Optional<UserInfo> userInfo = provider.getUserInfo(accessToken);
+
+                    // If either is present, return the result
+                    if (userRecordIdentifier.isPresent() || userInfo.isPresent()) {
+                        logger.log(Level.FINE, "Bearer token detected, provider {0} confirmed validity and provided user info", provider.getId());
+                        return new OIDCUserInfo(userRecordIdentifier.get(), userInfo.get());
```
qqmyers, line 1022: If you return the OAuth2UserRecord from getUserRecord as suggested above, do you have all the info you need? Or could you add the missing fields to it? Versus creating a new aggregator class?

qqmyers (changes requested):
Looks good overall - lots of changes, lots of tests! I made a few suggestions for changes and this probably needs a release note.

src/test/java/edu/harvard/iq/dataverse/api/UsersIT.java
```diff
@@ -515,6 +517,161 @@ public void testDeleteAuthenticatedUser() {
...
+        // Should return error when the provided User JSON is valid but the provided Bearer token is invalid
+        registerOidcUserResponse = UtilIT.registerOidcUser(
+                "{"
+                        + "\"termsAccepted\":true"
+                        + "}",
+                "Bearer testBearerToken"
+        );
+        registerOidcUserResponse.then().assertThat()
+                .statusCode(UNAUTHORIZED.getStatusCode())
+                .body("message", equalTo("Unauthorized bearer token."));
```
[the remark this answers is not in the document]
GPortas: In this case, the error messages are meaningful for the test assertions, as depending on the point where the flow fails, one message or another will be returned, even though the HTTP error code is the same (401) (See the different AuthorizationExceptions that are thrown in https://github.com/IQSS/dataverse/blob/73c407997b0a9972aa1e1989aef7bc4056f2128b/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java#L1001). 

As far as I know, Dataverse does not implement custom response error codes, so I had to rely on the response error messages for this test. I understand the concern about string comparisons and am open to revisiting this approach if alternative mechanisms (such as custom response error codes) become available in the future.

doc/sphinx-guides/source/installation/config.rst
GPortas: There is a separate flag. I'm not sure if perhaps you hadn't noticed it. There are two flags: `api-bearer-auth `and `api-bearer-auth-provide-missing-claims` (the one used in registration when the provider's claims are insufficient). I added both to the docs, as `api-bearer-auth` was missing from the Feature Flags section.

doc/sphinx-guides/source/api/auth.rst
```diff
@@ -81,6 +81,29 @@ To test if bearer tokens are working, you can try something like the following (
...
 
+It may happen that when you try to authenticate a user for the first time with a bearer token, it does not have an associated user account in Dataverse. In this case, it is necessary to register the user using the following endpoint:
+
+.. code-block:: bash
+
+  curl  -H "Authorization: Bearer $TOKEN" -X POST http://localhost:8080/api/users/register --data '{"termsAccepted":true}'
+
+It is essential to send a JSON that includes the property ``termsAccepted`` set to true, which indicates that you accept the terms of service of Dataverse. Otherwise, you will not be able to create an account.
+
+In this JSON, we can also include the fields ``position`` or ``affiliation``, in the same way as when we register a user through the Dataverse UI. These fields are optional, and if not provided, they will be persisted as empty in Dataverse.
```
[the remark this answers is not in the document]
GPortas: They are not available, at least by default.

These are not standard OIDC claims, as these properties are specific to the Dataverse domain. They cannot be found in the PersonClaims class of the Nimbus SDK neither. 

In my opinion, these properties should not be linked to the `api-bearer-auth-json-claims` flag, as they represent Dataverse-specific properties that should be provided during sign-up (if desired, since they are optional).

doc/sphinx-guides/source/installation/config.rst
qqmyers: Right. I meant 3 - allow authorizing by token, allow reg by token, allow reg by token with user supplied info. At QDR (a fork) we create accounts via a different mechanism tied to our single sign on mechanism and will eventually want api access via token but not registration via token. As I said above, I don't know that there's a real non-fork use case where other people will want this, so I think it is OK to stay with just the 2 flags you have. We can make a change in the future if needed.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
GPortas: I have refactored the code to rely solely on OAuth2UserRecord. Good catch, thank you.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
```diff
@@ -0,0 +1,166 @@
...
+
+            ctxt.authentication().createAuthenticatedUser(userRecordIdentifier, userDTO.getUsername(), userDisplayInfo, true);
+
+        } catch (AuthorizationException ex) {
+            throw new PermissionException(ex.getMessage(), this, null, null, true);
+        }
+    }
+
+    private void updateUserDTO(OAuth2UserRecord oAuth2UserRecord, boolean provideMissingClaimsEnabled) throws InvalidFieldsCommandException {
+        if (provideMissingClaimsEnabled) {
```
qqmyers, line 71: So the check that claims and json match is only done when the provideMissing is enabled? COuld/should it be checked for the other case as well? I.e. if you're not allowed to provide missing info but do should you also get a warning (at least if it is different?)? (Or have I missed that that is happening somewhere?)

qqmyers (changes requested):
Thanks for all the updates. I'll pass it back with a  couple questions.

doc/sphinx-guides/source/api/auth.rst
```diff
@@ -81,6 +81,29 @@ To test if bearer tokens are working, you can try something like the following (
   
   curl -H "Authorization: Bearer $TOKEN" http://localhost:8080/api/users/:me
 
+It may happen that when you try to authenticate a user for the first time with a bearer token, it does not have an associated user account in Dataverse. In this case, it is necessary to register the user using the following endpoint:
+
+.. code-block:: bash
+
+  curl  -H "Authorization: Bearer $TOKEN" -X POST http://localhost:8080/api/users/register --data '{"termsAccepted":true}'
+
+It is essential to send a JSON that includes the property ``termsAccepted`` set to true, which indicates that you accept the terms of service of Dataverse. Otherwise, you will not be able to create an account.
```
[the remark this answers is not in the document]
johannes-darms: However, would it make sense to add a property per OIDC provider that automatically sets this value to true? I'm thinking of internal institutional installations where its acceptance is implicit or already covered during account creation.

doc/sphinx-guides/source/api/auth.rst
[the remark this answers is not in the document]
GPortas: I agree to leave it as is for now. In the future, we can follow the same approach as with the other fields—send them only if the IdP doesn’t provide them. However, since these aren’t standard claims, we’ll likely need to define an expected name in Dataverse for these fields, which might vary by provider.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
```diff
@@ -0,0 +1,166 @@
...
+            throw new InvalidFieldsCommandException(
+                    BundleUtil.getStringFromBundle("registerOidcUserCommand.errors.invalidFields"),
+                    this,
+                    fieldErrors
+            );
+        }
+    }
+
+    private String getValueOrDefault(String oidcValue, String dtoValue) {
+        return (oidcValue == null || oidcValue.isEmpty()) ? dtoValue : oidcValue;
```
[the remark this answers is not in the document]
GPortas: Please review my latest changes, where I added these checks and also verify that the string is not a blank string with spaces, which I believe should be considered invalid as well.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
GPortas: I've decided to throw an error instead of a warning to simplify the process and prevent the user from being registered with unexpected values. We stop the flow if we detect that the user sends claims in the JSON request and the feature flag is disabled. Let me know what you think.

GPortas:
Thank you for your detailed review (This is a big PR!). We’ve implemented several improvements. Please take a look at the latest changes and let me know if you have any additional suggestions. @qqmyers

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
qqmyers: Sounds good - now regardless of the flag if you send values that won't be used, you get an error.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/RegisterOIDCUserCommand.java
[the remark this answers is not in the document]
qqmyers: FWIW: We often use StringUtils.isBlank() to do the null or empty or blank check.

doc/sphinx-guides/source/api/auth.rst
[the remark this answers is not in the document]
qqmyers: I agree. I think some config option like that for terms, affiliation, and role would make sense, but I'd definitely suggest a new PR rather holding this up and making it even bigger.

qqmyers (approved):
Looks good. As discussed in other comments, I think/agree getting more field types from the claims makes sense and will probably be needed at some point, but this code solves the immediate issue and gives code that can be built on.

doc/sphinx-guides/source/installation/config.rst
johannes-darms: There is, we want account generation via token but don't allow the user to supply information.

johannes-darms:
> Looks good. As discussed in other comments, I think/agree getting more field types from the claims makes sense and will probably be needed at some point, but this code solves the immediate issue and gives code that can be built on.

I totally agree. This PR improves OIDC compatibility quite a bit and builds a solid foundation for further improvements (getting more information from an authorisation server, possibly via configurable `claims`/`scopes` per provider). 

The only conceptual problem I see with the approach is that the user is not asked for consent before the information flow. This is part of the [specification ]( https://openid.net/specs/openid-connect-core-1_0.html#Consent). The required processes and attributes are part of the standard and could be "just" used instead of reinventing the wheel with custom json properties. Terms and conditions, policies and other information that the authorisation server can display can be provided during client registration (https://openid.net/specs/openid-connect-registration-1_0.html#toc) (see Keyclack OID Client -> Advanced>Fine grained OIDC configuration). 

IHMO: As stated above, if the required properties (scopes/claims) are not provided by the OIDC server (can be confiured per client) or released by the user for use by a dataverse, the OIDC server configuration is simply not compatible or the user is not willing to participate. Implementing UI/APIs around it to obtain information via another channel does work, but requires additional code, circumvents the standard and somehow questions the use of using a protocol with the sole purpose of obtaining identity information.

doc/sphinx-guides/source/installation/config.rst
qqmyers: I think we're talking about different things. I was asking whether there's a case where you'd want tokens to work but not allow new accounts to be created with them. To address your point though: I think the api-bearer-auth-provide-missing-claims flag in the PR is already trying to address what you want, except, as you point out, the flag doesn't cover the terms in json. I think it could make sense to treat the terms accepted json the same as name/email and make the api-bearer-auth-provide-missing-claims flag cover all three, or perhaps better to have another flag (api-bearer-auth-terms-accepted-via-OIDC?) - with documentation that you must then make sure your client is configured to present the terms (and/or name/email) during the initial consent.

While I know we want to get this PR out, perhaps it makes sense to add an api-bearer-auth-terms-accepted-via-OIDC flag now and add the ~same logic we have for name/email for the terms (you can't submit the json if not allowed)? If I understand, that would allow a pure Keycloak config (except for role/affiliation which are not required fields).

qqmyers:
I think we do still ask the user for consent, it is just that the full terms are not presented there. In some sense, I see this discussion about whether it is better to have Keycloak ask for missing information, or whether the Dataverse client can do it. I'm not sure why one is better than the other (though both are possible). In some sense the argument that you should use Keycloak for this can just be applied further - if you need Keycloak to add info not supplied by, for example, ORCID, you should just require ORCID to add the fields you need or not use it. My guess is that the use case where a Dataverse instance would like to use an OIDC service that is not under their control enough for it to be configured to supply all the fields we'd like is real and having the API/SPA able to handle that case w/o requiring Keycloak seems reasonable. That said, we should definitely support a fully configured Keycloak in the middle too.

I think that means, minimally to make the terms acceptance in json be configurable by flag now, and, in the future, add support for the optional affiliation/role fields to come from claims?

johannes-darms:
>I think we do still ask the user for consent, it is just that the full terms are not presented there.

It does not ask if Dataverse is allowed to copy personal information (name, email,...) from an OIDC server to Dataverse. Nor, as said, are the ToS displayed. The OIDC protocol has concepts to solve this, but they aren't used. I'm happy with this implementation, as it's certainly an improvement. However, it can be improved in this respect by implementing the standard and the protocol. Yes, not all OIDC authentication servers support all features, support all requirements, and always provide meaningful responses. Solution for these cases have to be worked out and implemented later, and some of the solutions in the PR already solve them, but deviations from the standard should be avoided where possible.

>My guess is that the use case where a Dataverse instance would like to use an OIDC service that is not under their control enough for it to be configured to supply all the fields we'd like is real and having the API/SPA able to handle that case w/o requiring Keycloak seems reasonable.

Yes, it makes perfect sense. However, you need to make sure that the OIDC providers that support these features can be used without additional code, properties and configuration. This is already partly done with the feature flags.

>I think that means, minimally to make the terms acceptance in json be configurable by flag now, and, in the future, add support for the optional affiliation/role fields to come from claims?

Yes, of course this is the way to go. However, it lacks the behaviour that should happen when terms acceptance is disabled in JSON. This is basically an implementation of the protocol (i.e. creating an additional authentication request with a `promt=consent` option (see [spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest))).

g-saracca:
@ofahimIQSS I think we should wait for this one, while doing the SPA integration and forcing to send `username` when the OIDC token has the preferred_username present the endpoint should fail and is not failing.
`If properties are provided in the JSON, but corresponding claims already exist in the identity provider, an error will be thrown, outlining the conflicting properties.`

GPortas:
@johannes-darms @qqmyers 

Ok, I see two distinct topics here. First, we have user consent, which is configured independently from the Terms of Service (ToS) in the client settings. It controls whether a user is prompted to approve data sharing and permissions with a specific client during login.

On the other hand, the 'Fine Grain OpenID Connect Configuration > Terms of Service' option allows you to attach the Relying Party's (in this case, Dataverse) ToS URL in the OIDC metadata.

Just to clarify, it is not valid to assume that accepting the consent prompted by 'Login Settings: Consent Required' automatically means the user has accepted the Terms of Service (ToS). While the two concepts may be related, they serve distinct purposes and have different legal and technical implications. I'm not entirely sure if we're on the same page here, so I wanted to mention this to avoid misunderstandings. Sometimes, with so many concepts involved, it's difficult to clarify everything through GitHub messages. I’d also be happy to discuss it on Zoom if that’s easier.

To address user consent acceptance, we could create a specific flag for Keycloak providers that checks whether the consent has been accepted in the provider before accessing the user’s information. This could be done using the Keycloak API.

On the other hand, Keycloak doesn’t have a built-in feature to automatically handle ToS acceptance as part of the standard consent flow for OIDC. The Terms of Service URL configured under Fine Grain OpenID Connect Configuration is part of the OIDC discovery document metadata and is not shown in the Keycloak UI. It is meant to be referenced by OIDC clients, and if you want users to accept the ToS explicitly, you would need to handle that separately.

One potential workaround could be to include the ToS link in the client consent text, which is configurable. However, I believe this might be too much of a workaround and could lead to misunderstandings. Another option could be to create a custom flag (e.g., tosAccepted=true) stored as a user attribute. But the question is: Isn’t it easier to manage the ToS from Dataverse, which is under our control, than to wait for Keycloak IdPs to apply ad-hoc configurations for ToS acceptance?

qqmyers:
+1 for a flag. It is easy for someone to configure Keycloak as in 
image
. Whether that's better than having an ad-hoc screen somewhere in a Dataverse app, or hoping that some tool using the api actually shows the user instead of hardcoding a yes in the json seems like something to leave up to individual installations. Leaving the flag false (allowing a yes in Json) seems like a good default as you can make sure the terms are shown in the SPA and that default and the SPA will work for places that don't use Keycloak/don't have control over the OIDC screen.

johannes-darms:
> @johannes-darms @qqmyers
> 
> Ok, I see two distinct topics here. First, we have user consent, which is configured independently from the Terms of Service (ToS) in the client settings. It controls whether a user is prompted to approve data sharing and permissions with a specific client during login.

Those topics are depending on the jursitication and legal team involved different, overlapping or the same concepts. But in most cases its easier to ask once for permission than many times. Thus, the combination of both* into one conceptional technical step. Sorry, that I wasn't clear about this. 
*(For the shake of completeness there is a third document, the privacy policy, which must also be shown and accepted by the user/) 

> On the other hand, the 'Fine Grain OpenID Connect Configuration > Terms of Service' option allows you to attach the Relying Party's (in this case, Dataverse) ToS URL in the OIDC metadata.

Yes, this is part of the protocol definition of [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest):
```
policy_uri
    OPTIONAL. URL that the Relying Party Client provides to the End-User to read about how the profile data will be used. The value of this field MUST point to a valid web page. The OpenID Provider SHOULD display this URL to the End-User if it is given. If desired, representation of this Claim in different languages and scripts is represented as described in [Section 2.1](https://openid.net/specs/openid-connect-registration-1_0.html#LanguagesAndScripts). 
tos_uri
    OPTIONAL. URL that the Relying Party Client provides to the End-User to read about the Relying Party's terms of service. The value of this field MUST point to a valid web page. The OpenID Provider SHOULD display this URL to the End-User if it is given. If desired, representation of this Claim in different languages and scripts is represented as described in [Section 2.1](https://openid.net/specs/openid-connect-registration-1_0.html#LanguagesAndScripts).
````

> Just to clarify, it is not valid to assume that accepting the consent prompted by 'Login Settings: Consent Required' automatically means the user has accepted the Terms of Service (ToS). While the two concepts may be related, they serve distinct purposes and have different legal and technical implications. I'm not entirely sure if we're on the same page here, so I wanted to mention this to avoid misunderstandings. Sometimes, with so many concepts involved, it's difficult to clarify everything through GitHub messages. I’d also be happy to discuss it on Zoom if that’s easier.

I think we are, but we can still chat in zulip later..

> To address user consent acceptance, we could create a specific feature flag in Dataverse for Keycloak providers that checks whether the consent has been accepted in the provider before accessing the user’s information. This could be done using the Keycloak API.

First, if you do implement it, please do not make it keylock-specific. Second, you do not need to store this information in the resource server (dataverse). If the permission is given, you get the data, otherwise you cannot get the information because the permission server does not give out user data.  The protocol even indicates why no information is transmitted. See [here](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest):
```
consent
    The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client. If it cannot obtain consent, it MUST return an error, typically consent_required. 
```

> On the other hand, Keycloak doesn’t have a built-in feature to automatically handle ToS acceptance as part of the standard consent flow for OIDC. The Terms of Service URL configured under Fine Grain OpenID Connect Configuration is part of the OIDC discovery document metadata and is not shown in the Keycloak UI. It is meant to be referenced by OIDC clients, and if you want users to accept the ToS explicitly, you would need to handle that separately.

Yes, this is a fundamental issue with OIDC. The `tos_uri` and `policy_uri` properties are optional during client registration. Unfortunately, the protocol only defines  ` The OpenID Provider SHOULD display this URL to the End-User if it is given.`, so we can't really control if and how this information is displayed and handled by the authentication server. There are many different solutions, but they are all vendor specific and all out of our control (i.e. the entity running and configuring an OpenID provider/authentication server can change the behaviour, but we as the resource server (developers/operators) cannot).

What can we do about it? We as resource servers (developers/operators) want a (legally binding) property that tells us that the user has read, understood and accepted the required terms/policies, right? This means that every user has to set this flag. When we create and control a user interface, we can ensure that the required information is available, displayed to a user, scrolled through and actively accepted by a click. Once we no longer control the visualisation (i.e. a user-facing API sets the property), we cannot infer that the user has seen, interacted with or accepted anything. All we have is a property claiming that this happened. Whether this is sufficient is a question for a legal team and will vary from jurisdiction to jurisdiction. (For our installation it is not!) The ODIC protocol with `promtpt=consent` ensures that at least a screen is shown and a consent button is pressed before any user identifiable information is transferred before an account can be created. Whether this is sufficient (to also count as accepting the ToS) also depends on what the screen looks like, what information is displayed, and what the legal team thinks about it, as said above, we as resource servers (developers/operators) cannot control this. 

So how do we solve this problem? Let the user configure what they need (i.e. what their legal team requires) and set reasonable defaults. 

First, allow `promtpt=consent` to be used during account creation (possibly configurable via a flag). This will ensure that at least every user is happy with the transfer of their personally identifiable information. Second, implicitly set the ToS as accepted during account creation if `promtpt=consent` is used (configurable via a flag). This allows the problem to be delegated to the ODIC service and avoids the problem within dataverse. 
If `promtpt=consent` or implicit ToS acceptance is not enabled via feature flags, requires an explicit property to set ToS acceptance.

qqmyers (approved):
OK - looks good. I suggested some rewording in the docs for the API call - feel free to accept or edit further - I was basically trying to make the default path come first.

ofahimIQSS:
Merging PR tests are passing

vera:
I was just about to open an issue about this when I found this PR, which already solves the problem :+1: I just tested this with a bunch of APIs and it's all working well.  Thanks @GPortas for this big improvement!

GPortas:
I'm glad it's useful for you! @vera

PR #10983, opened 2024-10-30
DANS-1691 - possible fix for NPE when using IndirectList

qqmyers:
**What this PR does / why we need it**: DANS is seeing occasional 500 errors from a pre-publication workflow on a dataset with 12K files. The error 

- Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
+++ b/src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
@@ -218,8 +218,13 @@ public class FileMetadata implements Serializable {
              */
             if (!(fileCategories instanceof ArrayList)) {
                 List<DataFileCategory> newDFCs = new ArrayList<DataFileCategory>();
-                for (DataFileCategory fdc : fileCategories) {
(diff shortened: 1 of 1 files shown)


pdurbin:
I saw a very similar error here:

- #11293

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDatasetVersionCommand.java
```diff
@@ -280,13 +281,28 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
                 AuthenticatedUser au = (AuthenticatedUser) getUser();
                 ctxt.datasetVersion().writeEditVersionLog(dvd, au);
             }
+            if ( theDataset != null ) {
```
qqmyers, line 284: FWIW: I think this was unrelated/ related to other optimizations. If this PR is used for fixing the indirect list issue, this should be looked at /probably be removed/moved to it's own PR.

PR #11003, opened 2024-11-06
Fix lookups of Harvested datasets with lower-case versions of persistent identifiers in the database

qqmyers:
**What this PR does / why we need it**: The recent changes to support lower case shoulders in #10708 accidentally broke retrieval of harvested datasets with lowercase DOIs (which is newly possible with DataCite as the Harvesting source). This fix  stores harvested DOIs in their original case (useful if the harvesting source is not case-insensitive even though it should be) but changes to using case-insensitive DB queries so case-insensitive retrieval works.

- Closes #

**Suggestions on how to test this**: Verify that the dataset page works for DOIs when the db identifier for it is in lower/mixed case in the database. This only happens naturally when Harvesting from a source that provides a lower/mixed case DOI. To test, editing the db entry for a normal/internal DOI would work. Regression testing - datasets with normal DOIs, PermaLinks, Handles as the identifier/alternative identifier should continue to work and be case insensitive (i.e. one can change the dataset page URL to use lowercase letters and it should still work.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DvObject.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DvObject.java
@@ -27,9 +27,9 @@ import jakarta.persistence.*;
     @NamedQuery(name = "DvObject.ownedObjectsById",
 			query="SELECT COUNT(obj) FROM DvObject obj WHERE obj.owner.id=:id"),
     @NamedQuery(name = "DvObject.findByGlobalId",
-            query = "SELECT o FROM DvObject o WHERE UPPER(o.identifier)=:identifier and o.authority=:authority and o.protocol=:protocol and o.dtype=:dtype"),
+            query = "SELECT o FROM DvObject o WHERE UPPER(o.identifier)=UPPER(:identifier) and o.authority=:authority and o.protocol=:protocol and o.dtype=:dtype"),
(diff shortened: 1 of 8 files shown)


src/main/java/edu/harvard/iq/dataverse/DvObject.java
```diff
@@ -27,17 +27,17 @@
     @NamedQuery(name = "DvObject.ownedObjectsById",
 			query="SELECT COUNT(obj) FROM DvObject obj WHERE obj.owner.id=:id"),
     @NamedQuery(name = "DvObject.findByGlobalId",
-            query = "SELECT o FROM DvObject o WHERE o.identifier=:identifier and o.authority=:authority and o.protocol=:protocol and o.dtype=:dtype"),
+            query = "SELECT o FROM DvObject o WHERE UPPER(o.identifier)=:identifier and o.authority=:authority and o.protocol=:protocol and o.dtype=:dtype"),
```
stevenwinship, line 30: Use UPPER(o.identifier)=UPPER(:identifier). 
or
WHERE o.identifier ILIKE :identifier

stevenwinship (approved):

landreev:
@qqmyers @stevenwinship thank you for the extra effort with the index, etc.

PR #11011, opened 2024-11-08
10909 Support for OAI-PMH harvesting from DataCite 

### LINKED ISSUE #10909: Add support for OAI-harvesting from DataCite
DataCite maintains an OAI server (https://oai.datacite.org/oai) serving records for every DOI they have registered. There is a lot of interest in being able to harvest from them (since these are all registered DOIs, they will be redirecting to the original archival location of the actual studies/datasets etc.)

There is a couple of issues that must be addressed before our OAI client implementation is able to do that. 

1. The oai_dc import code in Dataverse expects the metadata fragment to be self-contained, and, most importantly have the main persistent identifier (the DOI in this case) to be present in the `<dc:identifier>` field. DataCite however does not include the main DOI in the oai_dc - since they are using these DOIs as the OAI identifiers as well, they assume that it is enough to include them in the OAI record header, in the `<identifier>` field, like this:  
```
<record>
<header>
      <identifier>doi:10.7910/dvn/tjclkp</identifier>
      <datestamp>2023-01-03T21:08:00Z</datestamp>
      <setSpec>HARVARDU</setSpec>
      <setSpec>GDCC.HARVARD-DV</setSpec>
</header>
<metadata>
      <oai_dc:dc xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/oai_dc/ http://www.openarchives.org/OAI/2.0/oai_dc.xsd">
         <dc:title>Open Source at Harvard</dc:title>
         <dc:creator>Durbin, Philip</dc:creator>
         <dc:publisher>Harvard Dataverse</dc:publisher>
         <dc:date>2017</dc:date>
         <dc:date>Issued: 2017</dc:date>
         <dc:description>The tabular file contains information ...</dc:description>
         <dc:contributor>Durbin, Philip</dc:contributor>
         <dc:type>Dataset</dc:type>
     </oai_dc:dc>
</metadata>
</record>
```
  Without the `<dc:identifier>`, our code in its current form is failing to import the record above. 
All that needs to be done, we need to add some logic to use the identifier from the OAI header in situations like this. (We actually used to do that in one of the previous iterations of the harvester). 

2. DataCite OAI implementation offers a very promising feature of accepting arbitrary search queries as the OAI set names (https://support.datacite.org/docs/datacite-oai-pmh#arbitrary-queries). This would make it possible to harvest individual records by the DOIs (something we've been asked for specifically) or any possible subsets of their offerings. 
Example: 
```
echo "doi%3A10.7910/DVN/TJCLKP" | base64 
ZG9pJTNBMTAuNzkxMC9EVk4vVEpDTEtQCg==
```
Now you can harvest this "set" made up of one dataset above, as in
https://oai.datacite.org/oai?verb=ListRecords&metadataPrefix=oai_dc&set=~ZG9pJTNBMTAuNzkxMC9EVk4vVEpDTEtQCg==
_Unfortunately_ for whatever reason, the above notation only works in ListRecords, but not in ListIdentifiers, that Dataverse actually uses. From talking to Datacite, they may be able to fix it eventually - but not in an instant, "oh yeah, we just had this one line commented out" way. 
We should go ahead and implement support for harvesting using `ListRecords` (it should be faster, if nothing else; we handle it via `ListIdentifiers` then `GetRecord`, one record at a time, for various historical reasons - but it may come handy in other situations, to have both modes supported (and configurable, per client maybe?)

Clearly, we don't want to touch the current, JSF-based harvesting clients UI. But making the changes above, in the import and harvesting back end code, and then making it possible to set up or configure a client via the `/api/harvest/clients` API to take advantage of these improvements should be both useful and sufficient.

landreev:

The underlying goal was to be able to harvest metadata directly from DataCite. Which makes it possible for a Dataverse instance to harvest datasets from institutions and schools who don't maintain their own OAI servers, as long as they register their DOIs with DataCite. 

2 major features needed to be added to accommodate this, as described in the linked issue (although one has since been added as a standalone PR #11049 and is already in 6.5). On top of that a few other fixes and improvements have been added. (for example, it is now possible to schedule harvests via the API - this has been a GUI-only feature until now)

Note that everything in this PR is already in prod. use at IQSS via a deployment of a custom experimental patch of v6.5. This had to be done in the context of ongoing collaborations to accommodate the relevant deadlines. The 2 prod. collections involved are: 

https://dataverse.harvard.edu/dataverse/bertarelli (note that in this instance the harvested content is included in their subcollections alongside "real", locally-deposited datasets)
https://dataverse.harvard.edu/dataverse/designsafe

- Closes #10909

I'm about to mark this PR "ready for review". This is true as far as the underlying Dataverse code is concerned however, at the moment the branch is built with a local copy of the customized xoai jars. This is temporary, pending the needed changes being incorporated into a gdcc-released version which is something that needs to happen before this PR is merged. 

See the release note and the API guide. 

The following is an example harvesting client configuration that will harvest a set made from a single dataset, Gary King's doi:10.7910/DVN/9L6A8X: 

```
{
    "useOaiIdentifiersAsPids": true,
    "useListRecords": true,
    "allowHarvestingMissingCVV": false,
    "set": "~ZG9pOjEwLjc5MTAvRFZOLzlMNkE4WAo=",
    "nickName": "harvest9L6A8X",
    "dataverseAlias": "INSERTYOURCOLLECTION",
    "type": "oai",
    "style": "default",
    "harvestUrl": "https://oai.datacite.org/oai",
    "archiveUrl": "https://oai.datacite.org",
    "archiveDescription": "The metadata for this Dataset was harvested from DataCite. Clicking the dataset link will take you directly to the original archival location, as registered with DataCite.",
    "metadataFormat": "oai_dc"
}
```

The magic behind the set name in the configuration above, that allows to harvest just this specific dataset: 
The native DataCite API query: 
```https://api.datacite.org/dois?query=doi:10.7910/DVN/9L6A8X```
Encoding the query definition in base64: 
```
echo "doi:10.7910/DVN/9L6A8X" | base64
ZG9pOjEwLjc5MTAvRFZOLzlMNkE4WAo=
```

**Additional documentation**:

### CODE REVIEW

**Comment by landreev** [comment]
The last Jenkins failure was from my new test added as part of this PR. It was the result of a conflict with something recently merged into develop (the last Jenkins run was triggered by syncing the branch with develop). Fixing now.

**Review by qqmyers — APPROVED** [review]
Looks OK to me. I ignored all of the local library changes though, and did not test. Should we add "Waiting" to this so it doesn't go through QA with the local stuff?

File: doc/release-notes/10909-datacite-oai-harvesting.md [hunk 1]
```diff
@@ -0,0 +1,10 @@
+### OAI Harvesting from DataCite
+
+DataCite maintains an OAI server (https://oai.datacite.org/oai) that serves records for every DOI they have registered. There's been a lot of interest in the community in being able to harvest from them. This way, it will be possible to harvest metadata from institution X even if the institution X does not maintain an OAI server of their own, if they happen to register their DOIs with DataCite. One extra element of this harvesting model that makes it especially powerful and flexible is the DataCite's concept of a "dynamic OAI set": a harvester is not limited to harvesting the pre-defined set of ALL the records registered by the Institution X, but can instead harvest virtually any arbitrary subset thereof; any query that the DataCite search API understands can be used as an OAI set (!).
+
+A few technical issues had to be resolved in the process of adding this functionality so, as of this release it is being offered as somewhat experimental. Its beta version is nevertheless already in use at IQSS with seemingly satisfactory results.
```
qqmyers, line 5: Aside re: our "experimental" label. There are groups who avoid these features assuming they might go away/who are surprised to hear that Harvard is using them and is unlikely to drop them completely. We might want to think about removing the 'experimental' language after one version, or changing to 'cutting edge' or something.

landreev: Agree. Also, the label may be unnecessary in this case. Not that I'm 100% positive it's not going to fail for anyone; but I don't think it can cause any real damage either.

landreev:
> Looks OK to me. I ignored all of the local library changes though, and did not test. Should we add "Waiting" to this so it doesn't go through QA with the local stuff?

Good question. The xoai-side changes are simple; even if Oliver makes me re-implement them from scratch, as I expect... Ok, let me think about it, but I'll probably add a "waiting" label, but then communicate to Omer that it could make sense to get a head start on testing, if he has spare cycles.

doc/release-notes/10909-datacite-oai-harvesting.md
landreev: I got rid of the "experimental" part and otherwise slightly refined the docs.

landreev:
I removed the "waiting" label. If a new version of xoai is released before our 6.6, I'll make another pr.

ofahimIQSS:
When I pushed this PR to internal, the app would not start up and I got stuck on the Payara page. 

#Caused by: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 6.5.0.7

image

landreev:
The flyway script in this branch has a different number now  - `V6.5.0.8.sql`.
What you are seeing must be the result of testing different branches that have different `V6.5.0.7.sql` in them.  I.e., the script named `V6.5.0.7.sql` that is now in develop (merged from #11217 yesterday) is different from the script under the same name from another branch (@stevenwinship's, possibly?) that's already been tested on dataverse-internal, hence the mismatch. 

The only way to deal with a situation like this is to manually remove the offending entries from the `flyway_schema_history` database table.

landreev:
P.S. The same inherent problem is now in place for this PR and #11277: both have `V6.5.0.8.sql` in them. Whichever is merged first, the script will need to be renamed `*.9.sql` in the other. The `flyway_schema_history` table will need to be manually edited between deploying the 2 branches on the same server.

ofahimIQSS:
Testing Passed, no issues found. Merging. 

image

image

image

PR #10997, opened 2024-11-01
github actions from v3 to v4

stevenwinship:
**What this PR does / why we need it**: Starting November 30, 2024, GitHub Actions customers will no longer be able to use v3 of [actions/upload-artifact](https://github.com/actions/upload-artifact) or [actions/download-artifact](https://github.com/actions/download-artifact).

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/10993

- Closes #10993

**Special notes for your reviewer**: I chose to update all the V3 to V4 actions in the yml files. Only the upload and download were required but I felt it's better to not mix/load multiple library versions if possible.
Due to a bug in download-artifact@V4 the version was bumped to 4.1.7 to include the fix
As mentioned in an issue that there was a bug in download V4 that was fixed in 4.1.7 https://github.com/actions/upload-artifact/issues/478
Update: v4 now gets the latest so download-artifact can now use v4 to get the fix in 4.1.7 (latest is v4.1.8 as of this PR)

**Suggestions on how to test this**: I've made temporary changes to various files to force most of the actions to run. Obviously any release specific actions were not run. I think the best way to test it is to merge it and monitor the build system. 

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #9251, opened 2023-01-03
add check to look for updates to Github actions being used

carlsonp:
**What this PR does / why we need it**:

This adds a file that will automatically check daily the various Github Actions used to see if there are updates to the actions themselves.

Will be no-op on docs and codebase.

**Which issue(s) this PR closes**: None

**Special notes for your reviewer**:

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #11050, opened 2024-11-25
Update actions/checkout 2->3+

qqmyers:
**What this PR does / why we need it**: I'm seeing warnings that actions/checkout 2 uses the deprecated node 12 and an update to at least v3 using node 16 is needed. There is a v4 (node 20), and, in looking through the code, I see several other v3 actions that have v4 versions.

This PR updates all of these to v4.

- Closes #

**Suggestions on how to test this**: Guessing this is like the earlier #10997 PR in that we're just looking to make sure things don't break. If they do, we can either address the issue or roll-back the relevant change. (The shift from checkout v2 is one we should keep so we stop using node 12, but everything else is ~optional.) 

The changes are across many tasks - building the guides, review dog, pushing containers, pushing the spi, etc. 

**Additional documentation**:


pdurbin:
This PR from @carlsonp might help us know when to bump versions in the GitHub Actions we use:

- #9251

bencomp (approved):

pdurbin (approved):
Seems fine. Approved. Sadly, we have no way to test this so let's revert if anything breaks.

Also, as @qqmyers notes, we should close the open PRs (after we merge this and confirm nothing is broken?) created by dependabot that have the same fixes (#11100 etc.) at some point.

ofahimIQSS:
Performed short round of regression testing in internal - no issues uncovered. 

https://github.com/user-attachments/assets/a00c9f49-4708-43eb-b905-25f7dd7abb2f

PR #11052, opened 2024-11-26
11044 edit facet bug

### LINKED ISSUE #11044: Multiple listing for facets after editing the list
I was testing other things when I went to a collection's Edit/General Info form and added an additional facet, and then added another one. Looking at the list of existing facets in the form, I began to see doubles (and at QDR, even quadruples). 

image

This appears to happen on beta.dataverse.org (JSF UI) but not demo.dataverse.org, so I'm guessing it is a problem introduced post v6.4. (My QDR test machine has code merged from develop just a couple days ago.

At QDR, I verified that duplicates are in the database, e.g. multiple entries in dataversefacet for the same datasetfieldtype_id and dataverse_id: 
image

I'm not sure what is happening with the display order.

* When does this issue occur?

* What happens?

develop branch post 6.4

**Screenshots:**

*

sekmiller:
Fixes an issue where the facets would be added multiple times on Edit Dataverse

- Closes #11044 Multiple listing of Facets after Dataverse Edit

The issue was that while the existing facets were deleted in the DB, they were left as part of the Dataverse being edited so when the dataverse is saved the facets were re-created. Similar situation for field input levels.

**Suggestions on how to test this**: Edit Dataverse via api and UI and verify that the facets and dataset field input levels are being saved correctly with no doubling of entries. (see original ticket for example of failure state.)

**Additional documentation**: None - fixing documented behavior


qqmyers:
Haven't tried to review - would it possible to also add a db constraint to avoid more than one entry per datasetfieldtype_id + dataverse_id combo?

sekmiller:
@qqmyers I thought about adding a DB constraint, but I was worried that if it was violated it would fail badly - or at least in a way that would be difficult for the end user to understand/fix.

src/main/java/edu/harvard/iq/dataverse/DataverseFacetServiceBean.java
```diff
@@ -42,23 +42,24 @@ public void delete(DataverseFacet dataverseFacet) {
...
-        
+
         dataverseFacet.setDisplayOrder(displayOrder);
         dataverseFacet.setDatasetFieldType(fieldType);
         dataverseFacet.setDataverse(ownerDv);
-        
-        ownerDv.getDataverseFacets().add(dataverseFacet);
+
         em.persist(dataverseFacet);
+        ownerDv.getDataverseFacets().add(dataverseFacet);
```
qqmyers, line 61: Just curious - did this have to move after the persist?

qqmyers (approved):
Looks good - I haven't tested but the changes seem reasonable given the issue. @sekmiller - there are two failing tests though. Perhaps the deleteFile one is just timing but the other looks like it's facet-related. I'll approve, conditional on it being able to pass the tests before QA completes.

src/main/java/edu/harvard/iq/dataverse/DataverseFacetServiceBean.java
sekmiller: maybe not. I put it back and it still seems to work

ofahimIQSS:
Passed testing - Merging PR 
[Testing of 11052.docx](https://github.com/user-attachments/files/18010231/Testing.of.11052.docx)

PR #8744, opened 2022-05-26
File Recognition - Add support for files without extensions

adaybujeda:
**What this PR does / why we need it**:
Use the full filename to determine the file type when the file do not have an extension.

**Which issue(s) this PR closes**:

- Closes #8740 

**Special notes for your reviewer**:
Supported files in this PR are `Makefile` and `Snakemake`

**Suggestions on how to test this**:
Upload a file without extension as the ones suggested above.
Files with extension work as before.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
No

**Is there a release notes update needed for this change?**:
No

**Additional documentation**:
No

PR #11054, opened 2024-11-26
IQSS/10108 Stata mimetype refinement for direct upload

### LINKED ISSUE #10108: Not ingested file into Dataverse
I've provided a link to a binary file that was uploaded to Dataverse on December 1, 2021. You can access it here:  (https://entrepot.recherche.data.gouv.fr/file.xhtml?persistentId=doi:10.15454/YUFRUW/ICGF0B&version=2.0)
After downloading the file in its original .dat format and attempting to reupload it to Dataverse, the file was not successfully ingested. Please refer to the screenshot below for more details.

Capture2

qqmyers:
**What this PR does / why we need it**: As noted in the issue and recently in https://groups.google.com/g/dataverse-community/c/TBx0TTins2k, our ~extensible mimetype detector mechanism assumes starting from a local temp file which isn't useful with direct uploads. The most visible problem we have with this is that all Stata *.dta files get an application/x-stata type to start and our ingest sends all of those to the Stata 13 ingestor, which then fails for Stata 14/15 files. With normal uploads, the local detector reads the first few bytes of the file and assigns a more specific type, e.g. application/x-stata-14 (or 15, etc) which gets routed to the correct version of the Stata ingestion code.

Since the determination of Stata version only relies on reading the first ~42 bytes of the file, and only needs those in a buffer/doesn't require a File to start from, this PR adds code to retrieve the required bytes and run the Stata version check on direct upload(S3) and presumably remote files/Globus files on S3 (cases where the storageidentifier is provided during upload and where getInputStream works.)

The PR also has some notes about the potential to clean things up further and implement other mimetype detectors that only require a subset of bytes in a more extensible framework. I don't have any plan to implement this but incremental steps to add other detectors to the direct upload path are probably possible if there are other cases where we're seeing problems.

- Closes #10108

**Special notes for your reviewer**: I also added a note that I think there's a no-op section now where we check files by extension when the extension is null. If someone knows more about the history there, perhaps we can restore whatever was supposed to happen there, or we could delete it.

The code also makes a slight logging change. After the code was updated to allow getting Mimetypes using the full filename, we've been getting info-level suggestions to perhaps add that specific filename to MimeTypeDetectionByFileName.properties which is pretty noisy. I changed the code to keep info-level if the extension is checked and isn't in MimeTypeDetectionByFileExtension.properties but dropped it to fine for the filename check.

**Suggestions on how to test this**: Find a Stata file with version 14 or 15 and upload it via direct upload/using a direct upload store, verify that it gets a mimetype including the version and is ingested (assuming ingest size limits, etc. allow). (Some Stata files have a *.dta extension - searching for that might help in finding an example.)

**Additional documentation**: slightly updated the docs to indicate the Stata check is now done in direct upload.


src/test/java/edu/harvard/iq/dataverse/api/S3AccessIT.java
```diff
@@ -377,6 +382,236 @@ public void testDirectUpload() {
...
+            expectedException = ex;
+        }
+        assertNotNull(expectedException);
+        // 404 because the file has been sucessfully deleted
+        assertEquals(404, expectedException.getStatusCode());
+
+    }
+
+    @Test
+    public void testDirectUploadDetectStataFile() {
```
pdurbin, line 411: @qqmyers please take a look at this test I added (mostly copied from the one above).

qqmyers: This test is OK, and it passes, so I'm fine with including it. A better and shorter one would be to either submit the mimetype as application/octet-stream or even application/x-stata so that the new detection works during upload rather than requiring it to be triggered in a redetect call. I'm not sure that testing download helps here either - that's just part of the cut/paste I assume.

src/test/java/edu/harvard/iq/dataverse/api/S3AccessIT.java
```diff
@@ -377,6 +382,236 @@ public void testDirectUpload() {
...
+//        //upload a tabular file via native, check storage id prefix for driverId
+//        String pathToFile = "scripts/search/data/tabular/1char";
+//        Response addFileResponse = UtilIT.uploadFileViaNative(datasetId.toString(), pathToFile, apiToken);
+//        addFileResponse.prettyPrint();
+//        addFileResponse.then().assertThat()
+//                .statusCode(200)
+//                .body("data.files[0].dataFile.storageIdentifier", startsWith(driverId + "://"));
+//
+//        String fileId = JsonPath.from(addFileResponse.body().asString()).getString("data.files[0].dataFile.id");
+        long size = 1000000000l;
```
qqmyers, line 479: I think this only works (here and the test above) because the localstack must be configured with this as the part size? If the part size were smaller, e.g. the min of 5 MB, using this would return URLs for multiple parts whereas the small test files would only be 1 part still.

src/test/java/edu/harvard/iq/dataverse/api/S3AccessIT.java
pdurbin: In 614dc7e44c I'm passing application/octet-stream but file detection on initial upload doesn't seem to be working.

In that commit you'll see the comment "Should the contentType be application/x-stata-14 already?"

Later, after the redetect call, the content type is application/x-stata-14.

Am I testing it wrong?

qqmyers: FileUtil line 433 is checking to make sure the base type is application/x-stata to start, which would happen if you send the .dta extension on the file (or if you sent application/x-stata to start). FWIW: This is mainly for efficiency/ to avoid checking all files for stata14 and to just test those that were getting missed/left as x-stata before.

pdurbin: Thanks for talking this out today. I updated the tests to have the file end in .dta and file detection is working, like you said.

src/test/java/edu/harvard/iq/dataverse/api/S3AccessIT.java
pdurbin: Buh. I'm not sure.

src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java
```diff
@@ -495,6 +532,7 @@ public static String determineFileType(File f, String fileName) throws IOExcepti
                 logger.fine("mime type recognized by extension: "+fileType);
             }
         } else {
+            //ToDo - if the extension is null, how can this call do anything
```
pdurbin, line 535:
```suggestion
```

I'm not sure why this todo is here so I'm suggesting we delete it. In this PR...

- https://github.com/IQSS/dataverse/pull/8744

... we added support for detecting files like Makefile or Dockerfile. Having no extension is legit. Maybe I'm just misunderstanding the todo. 🤷

ofahimIQSS:
continuous-integration/jenkins/pr-merge is failing on this ticket.

qqmyers:
Guessing it's a timing issue - not waiting for ingest to finish. I've added a sleepForLock to fix it.

ofahimIQSS:
Was able to reproduce issue in internal and saw that stata 14 file wasn't getting ingested properly. 

Fix looks good: 
image

Going to merge PR after continuous-integration/jenkins/pr-merge completes

PR #11061, opened 2024-12-03
sql script/flyway to fix file access requests

### LINKED ISSUE #10714: Access requests missing on the File Permission Page since upgrading from 6.0 to 6.2
We have received tickets from users that they can’t find the list of access requests anymore. 
They know that their requests were there because they can find them in the notifications. 

- create a request on a restricted file in Dataverse 6.0
The database record looks like this: 
```
dvndb=> SELECT * FROM fileaccessrequests;
 datafile_id | authenticated_user_id |      creation_time      
-------------+-----------------------+-------------------------
           8 |                     6 | 2024-07-15 14:53:54.757
(1 row)
```
- Upgrade to 6.2 (possibly to 6.1 will also work)
The record will then look like this
```
dvndb=> SELECT * FROM fileaccessrequests;
 datafile_id | authenticated_user_id |      creation_time      | request_state | id | guestbookresponse_id 
-------------+-----------------------+-------------------------+---------------+----+----------------------
           8 |                     6 | 2024-07-15 14:53:54.757 |               |  1 |                     
(1 row)
```
The `request_state` is `NULL` instead of `CREATED`, which results in not being shown in the list of requests. 
The flyway script (V6.0.0.1__9599-guestbook-at-request.sql) has added the ‘request_state’ column, but it then has no state (NULL). 

The following SQL should fix this
```
UPDATE fileaccessrequests SET request_state='CREATED' WHERE request_state IS NULL;
```

However, when users have re-issued the requests because they were not handled, the above won't work: 
```
ERROR:  duplicate key value violates unique constraint "created_requests"
``` 
To detect those cases use the following query: 
```
SELECT * 
FROM fileaccessrequests 
WHERE (datafile_id, authenticated_user_id) IN (
  SELECT datafile_id, authenticated_user_id
  FROM fileaccessrequests
  GROUP BY datafile_id, authenticated_user_id
  HAVING COUNT(*) > 1
) ORDER BY datafile_id, authenticated_user_id;
```
And then as a solution remove that old requests using this query: 
```
DELETE FROM fileaccessrequests 
WHERE id IN (
  SELECT id 
  FROM fileaccessrequests 
  WHERE (datafile_id, authenticated_user_id) IN (
    SELECT datafile_id, authenticated_user_id
    FROM fileaccessrequests
    GROUP BY datafile_id, authenticated_user_id
    HAVING COUNT(*) > 1
  ) AND request_state IS NULL
);
```

stevenwinship:
**What this PR does / why we need it**: To prevent duplicate keys when user requests access to a file

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/10714

- Closes #10714

**Suggestions on how to test this**: Delete docker-dev-volumes directory.
Check out develop branch first, build and deploy in docker.
Add dataset with multiple files (restricted) and publish the dataset.
From another user request access multiple times for multiple files.
Manually delete the "Created" from the request_state **UPDATE fileaccessrequests SET request_state=NULL;**
Checkout this PR, build and deploy.
You should see any duplicate requests deleted from the database (oldest will remain for each file/user). Once updated the database will have only one request for each file/user and the UI will show "Access Requested"

**Additional documentation**:


src/main/resources/db/migration/V6.5.0.1.sql
```diff
@@ -0,0 +1,10 @@
+-- Fixes File Access Requests when upgrading from Dataverse 6.0
+-- See: https://github.com/IQSS/dataverse/issues/10714
+DELETE FROM fileaccessrequests
+WHERE creation_time <> (SELECT MIN(creation_time)
+                        FROM fileaccessrequests far2
+                        WHERE far2.datafile_id  = fileaccessrequests.datafile_id
+                          AND far2.authenticated_user_id  = fileaccessrequests.authenticated_user_id
+                          AND far2.request_state is NULL);
+
+UPDATE fileaccessrequests SET request_state='CREATED' WHERE request_state is NULL;
```
sekmiller, line 10: Did you consider putting a DB constraint on as Jim suggested?

sekmiller (commented):
The update script looks good. Just wondering if you considered putting on a DB constraint as Jim suggested.

src/main/resources/db/migration/V6.5.0.1.sql
stevenwinship: @qqmyers Could you add your suggestion here please?

sekmiller (approved):

ofahimIQSS:
Before PR in local
image

After PR in local
image

Merging PR no issues found

PR #11066, opened 2024-12-05
remove duplicate DatasetFieldTypes

### LINKED ISSUE #10472: Review/modify: jsonPrinter for datasetFieldType

As written the jsonPrinter for datasetFieldType adds subfields whenever !fld.getChildDatasetFieldTypes().isEmpty() calling the printer recursively. The jsonPrinter for metadatabocks calls the jsonPrinter for datasetfieldtype for all so the subfields appear twice - once as a subfield and once as a stand alone field. It's a little confusing to see these fields twice. A possible solution would be to filter out those subfields where isChild is true - That would work for our current dataset field types but would fail if we ever implement subfields of subfields. We also have to be careful not to disrupt the current users of the getMetadataBlock api endpoint.
The SPA team has noticed the issue and is working around it for CreateDataset see https://github.com/IQSS/dataverse/pull/10418

(Example users roles: API User)

Work on the PR noted above

Possibly changing the getMetadataBlock api to only show each datasetFieldType once

stevenwinship:
**What this PR does / why we need it**: Prevent duplicate DatasetFieldTypes from being returned via getMetadataBlock api endpoint

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/10472

- Closes #10472

**Additional documentation**:


sekmiller (approved):
Looks good. I'm a little worried about the hard-coded dataset field type counts in the integration tests, but I guess that's the way it already was.

ofahimIQSS:
There seems to be some conflicts with the PR. Could you please resolve? Thanks!

ofahimIQSS:
tested in internal - no longer seeing dupes in datasetFieldTypes

PR #11072, opened 2024-12-06
IQSS/11069 fix index status query

### LINKED ISSUE #11069: Error When Accessing /api/admin/index/status Endpoint
Description
While testing in the internal environment, the following issue was encountered:

When running the command:

bash
Copy code
curl http://localhost:8080/api/admin/index/status
as documented in [Solr Search Index Documentation](https://preview.guides.gdcc.io/en/develop/admin/solr-search-index.html#id3), an error is logged in the server logs.

Steps to Reproduce
1. Run the application on the internal environment.
2. Execute the following command:
curl http://localhost:8080/api/admin/index/status
3. Observe the error in the server logs.

Expected Behavior
The endpoint should return the Solr indexing status without logging an error in the server logs.

Actual Behavior
An error is logged in the server.log. See the attached log excerpt for details.

Possible Cause
This may be a database-related issue. Further investigation is needed to confirm.

Environment
Commit: Using Develop branch [82dad99]
Environment: Internal
API Endpoint: /api/admin/index/status
Documentation: [Solr Search Index Documentation](https://preview.guides.gdcc.io/en/develop/admin/solr-search-index.html#id3)
Observed Behavior: Error logged in server.log
Attachments
Error Log:
[Issue server.log.txt](https://github.com/user-attachments/files/18042281/Issue.server.log.txt)

qqmyers:
**What this PR does / why we need it**: This PR fixes the query used in the admin/index/status API call. A last minute change to the query in #10710 added during review used a feature only allowed in native queries. The change in this PR adapts the query to use native syntax to fix the /index/status call.

It also fixes two other issues that can break the status call and indexing in general for single value fields used with a cvoc script. In both cases, an 'else if' is needed rather than a 'if' in the code. In the status case this resulted in a null pointer exception when orphan permission docs exist. In the latter, two code sections were adding the same field to the solr document. With a single/non-multi field (e.g. depositor) this resulted in multiple values in a single-value solr field.

- Closes #11069

**Special notes for your reviewer**: The PR also drops some per-file logging to finest to simplify debugging.

**Suggestions on how to test this**: Try the api/admin/index/status call, verify that it completes and provides a report in the server.log. For the orphan permission error - there shouldn't be a null pointer reported in server.log when there are orphan permission docs. If your repository doesn't have any, I'm not sure it's worth trying to generate them. For the cvoc error, I think the external cvoc config in use on demo, which should have the depositor field associated with the ORCID script, should show the problem. (The problem doesn't appear in v6.4 for some reason afaik, but some change in 6.5 has surfaced the issue.)  In addition to seeing the dataset indexed when a depositor entry exists, you can also look for any logs in the logs/process-failures subdirectory. There should not be any related to the dataset being tested.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
@@ -1407,8 +1407,7 @@ public class DataFileServiceBean implements java.io.Serializable {
     }
 
     public boolean isInReleasedVersion(Long id) {
-        Query query = em.createQuery("SELECT fm.id FROM FileMetadata fm, DvObject dvo WHERE fm.datasetVersion.id=(SELECT dv.id FROM DatasetVersion dv WHERE dv.dataset.id=dvo.owner.id and dv.versionState=edu.harvard.iq.dataverse.DatasetVersion.VersionState.RELEASED ORDER BY dv.versionNumber DESC, dv.minorVersionNumber DESC LIMIT 1) AND dvo.id=fm.dataFile.id AND fm.dataFile.id=:fid");
-        query.setParameter("fid", id);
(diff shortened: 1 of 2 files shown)


ofahimIQSS:
Adding server.log file 
[server.log --- 1.txt](https://github.com/user-attachments/files/18064682/server.log.---.1.txt)

landreev (approved):
All makes sense.

ofahimIQSS:
Fix looks good, no issues encountered during testing. 
[11069 server.log](https://github.com/user-attachments/files/18083895/11069.server.log)

PR #11080, opened 2024-12-10
#11076 Allow User with View Unpublished Dataset Perms to access Preview URL without becoming Preview URL User

### LINKED ISSUE #11047: Disabling private URL doesn't work if you switch types w/o refreshing the page.
 Go to the Private URL dialog, create and then disable one type of URL (General/Anonymized). Then return (without refreshing the overall page) and create a private URL of the other type. Then try to disable it. The [confirmation dialog](https://github.com/IQSS/dataverse/blob/b8e47582d3308bf3886e1a2592f97bbb44bc9dcd/src/main/webapp/dataset.xhtml#L1291-L1302) will show the wrong type and will not disable the URL. Refreshing the overall page fixes the problem, so it looks ~like the dialog is just not getting refreshed when a URL is created.

* When does this issue occur?
Probably mostly during testing since people will probably only create one type and won't be immediately disabling them.

dataset page

* What happens?

develop/after #10961 

**Screenshots:**

*

### LINKED ISSUE #11076: Disable URL Not Working After Accessing Preview URL
Issue Summary: The "Disable URL" feature for the Preview URL does not work when the same user accesses the generated Preview URL in a different browser tab.

Steps to Reproduce:

URL: https://beehive.dataverse.org/
Username: dataverseadmin

1. Log in to the Dataverse instance:
2. Create an Unpublished Dataset and save it.
3. Edit the dataset and navigate to Preview URL.
4. Click on Create General Preview URL.
5. Copy the generated Preview URL and paste/load it in a different browser tab.
6. Return to the Preview URL screen and click Disable URL.
7. Expected Behavior: The Preview URL should be disabled and become inaccessible immediately.

Actual Behavior: The "Disable URL" feature does not work after the Preview URL is accessed in another browser tab.

Impact: This issue allows continued access to the Preview URL even after it is supposed to be disabled, potentially leading to unintended exposure of dataset content.

Environment Details:

Build: Development branch, v. 6.4 build 1672-82dad99 

Server.log file can be found here: 
[server.log --- preview URL.txt](https://github.com/user-attachments/files/18066329/server.log.---.preview.URL.txt)

sekmiller:
Going back to 2020 there was a known issue that as a logged in user with permission on the draft version of a dataset, if you accessed a preview url while logged in you would be logged out.

- Closes #11076 Disable Preview URL not working
- Closes #11047 Disabling Preview URL fails if you switch types

Added a test to the Private URL init method

Access Preview URL while logged in with View Unpublished Version Permissions. You should remain logged in and see the full draft version.

If you access a Preview URL while logged in as someone without View Unpublished, you will be logged out (be changed to a Preview URL user)

If not logged in, Preview URL access should be unchanged.

No

Edge case - probably not worth noting

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/privateurl/PrivateUrlPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/privateurl/PrivateUrlPage.java
@@ -1,6 +1,10 @@
 package edu.harvard.iq.dataverse.privateurl;
 
+import edu.harvard.iq.dataverse.Dataset;
+import edu.harvard.iq.dataverse.DatasetServiceBean;
+import edu.harvard.iq.dataverse.DataverseRequestServiceBean;
 import edu.harvard.iq.dataverse.DataverseSession;
+import edu.harvard.iq.dataverse.PermissionsWrapper;
 import edu.harvard.iq.dataverse.authorization.users.PrivateUrlUser;
 import java.io.Serializable;
 import java.util.logging.Logger;
@@ -20,8 +24,14 @@ public class PrivateUrlPage implements Serializable {
 
     @EJB
     PrivateUrlServiceBean privateUrlService;
+    @EJB
+    DatasetServiceBean datasetServiceBean;
     @Inject
     DataverseSession session;
+    @Inject
+    PermissionsWrapper permissionsWrapper;
+    @Inject
+    DataverseRequestServiceBean dvRequestService;
 
     /**
      * The unique string used to look up a PrivateUrlUser and the associated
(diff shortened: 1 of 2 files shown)


qqmyers:
The deleteFile test is flaky ([as reported to the core team last week](https://iqss.slack.com/archives/C03R1E7T4KA/p1733341812488609)) -  depending on the order of returned files which appears to now be random. I restarted the build for now and will look into fixing the test.

pdurbin (approved):
@ofahimIQSS @sekmiller and I have been discussing.

From my perspective, the bug is not completely fixed in that some users will still be logged out (the ones who don't have permission to the unpublished dataset, which will be most users), but at least the person creating the Preview URL (has permission already, that is) will no longer be logged out. It's definitely an improvement. And Omer likes this better than what's in develop. Approved.

ofahimIQSS:
An observation I had during testing which is an edge case - 

https://github.com/user-attachments/assets/eb764b44-2350-457f-8e16-e271d16e3ba3

qqmyers:
I think your edge case is #11047

sekmiller:
This should also fix #11047 - the refresh of the confirmation popup needed some attention.

ofahimIQSS:
> This should also fix #11047 - the refresh of the confirmation popup needed some attention.

Fix looks good - No issues to report back. Merging PR and proceeding with regression testing v6.5

PR #11081, opened 2024-12-10
Fix testDeleteFile

qqmyers:
**What this PR does / why we need it**: We've recently seen testDeleteFile fail to find a file (cc0.png) because the test looks for the data.files[0] element when the file is actually in data.files[1]. I've verified that between Jenkins runs where the test fails and then succeeds, that the order of files in the array is simply reversed. The code already had secondary checks for the files based on a path-based check for a file with the correct name, so I've simply deleted the checks that are hardcoded to the array position.

- Closes #

**Special notes for your reviewer**: We have MANY checks for files at data.files[0] in the test code. I don't know how many of them are checks when there are multiple files, and we (AFAIK) haven't seen other test failures due to this. That said, we may want to enforce a consistent order in the returned files to avoid more tests breaking which might make the API easier to use as well (if files are always in id order for example). With some checking in recent code, I didn't see where a change to order (making it random) might have occurred.

**Suggestions on how to test this**: Verify the tests pass. Watch over time to see that there are no testDeleteFile failures in other PRs after this is merged.

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
@@ -2274,9 +2274,6 @@ public class FilesIT {
         // Check file 2 still in v1.0
         Response v1 = UtilIT.getDatasetVersion(datasetPid, "1.0", apiToken);
         v1.prettyPrint();
-        v1.then().assertThat()
-                .body("data.files[0].dataFile.filename", equalTo("cc0.png"))
-                .statusCode(OK.getStatusCode());
         
         Map<String, Object> v1files1 = with(v1.body().asString()).param("fileToFind", "cc0.png")
                 .getJsonObject("data.files.find { files -> files.label == fileToFind }");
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
This PR should fix the failure we saw at https://jenkins.dataverse.org/job/IQSS-dataverse-develop/1942/console , namely:

`FilesIT.testDeleteFile:2278 1 expectation failed.", "JSON path data.files[0].dataFile.filename doesn't match.", "Expected: cc0.png", "  Actual: orcid_16x16.png"`

... after merging #11080. We're discussing it in Slack: https://iqss.slack.com/archives/C03R1E7T4KA/p1733929269215319

In REST Assured we should be able to avoid using `[0]` and instead do a find or findall (I forget) but I'm in favor of merging this PR as-is. That is, I'm ok with removing these assertions in the interest of getting the test suite passing more often.

Hopefully we'll remember to add them back in a less brittle way. 😬

pdurbin (approved):
As discussed in standup, we're keeping the status checks.

Also, I was confused by now I understand Jim's point that the code below is already checking for filenames in a cleaner way, without using `[0]`, for example:


        Map<String, Object> v1files1 = with(v1.body().asString()).param("fileToFind", "cc0.png")
                .getJsonObject("data.files.find { files -> files.label == fileToFind }");
        assertEquals("cc0.png", v1files1.get("label"));

So yes, it's great that we're getting rid of the brittle, old-style checks. Approved!

ofahimIQSS:
Merging PR - will monitor jenkins

Issue #11083, opened 2024-12-11
MyData and non-harvested datasets in harvested dataverses


plecor:
The MyData API [implicitely assumes that all datasets in a dataverse linked to a harvesting client are harvested datasets](https://github.com/IQSS/dataverse/blob/6e3a25050fb36951215808f0187c347797af60be/src/main/java/edu/harvard/iq/dataverse/mydata/MyDataFinder.java#L120). This produces incomplete results.

**Steps to reproduce**
 - create a dataverse and publish it
 - make this dataverse an harvest client and harvest some datasets
 - create a user with 'Dataset creator' rights in this dataverse
 - create a dataset with this user in this dataverse
 - go to 'My Data'
 - select only the role 'Dataset Creator'

Tested on 5.13, 6.0 and develop (6.4)
 
**Results**
The request to load the data fails with a NullPointerException and results are never shown:
> api/v1/mydata/retrieve?selected_page=1&dvobject_types=Dataverse&dvobject_types=Dataset&published_states=Published&published_states=Unpublished&published_states=Draft&published_states=In%20Review&published_states=Deaccessioned&role_ids=5&&mydata_search_term=
> java.lang.NullPointerException: Cannot invoke "java.util.List.isEmpty()" because "results" is null
> at edu.harvard.iq.dataverse.mydata.MyDataFinder.runStep2DirectAssignments(MyDataFinder.java:500)
> at edu.harvard.iq.dataverse.mydata.MyDataFinder.runFindDataSteps(MyDataFinder.java:177)
> at edu.harvard.iq.dataverse.mydata.DataRetrieverAPI.retrieveMyDataAsJsonString(DataRetrieverAPI.java:347)
> ...

The underlying cause of the NPE is that all checks regarding the absence of results are made before filtering against harvested dataverses.
If the user has other datasets in another dataverse, the request doesn't fail but the dataset in the harvesting dataverse is not listed.

**Are you thinking about creating a pull request for this issue?**
I can look into a PR to fix the NPE. 
Adapting MyData to list the missing datasets feels like a bigger undertaking.

pdurbin:
@plecor if you can fix the NPE, please do! Thanks for creating this issue!

jggautier:
Do y'all think we could also mention this bug in [the page on the guides](https://guides.dataverse.org/en/6.4/api/native-api.html#mydata)?

[...]

PR #11086, opened 2024-12-11
 Prevent NPE in MyData when all dataverses are harvested dataverses

plecor:
As discussed in #11083, MyData was checking for empty results and then filtering against harvested dataverses, producing a NullPointerException later on when all the results are harvested dataverses.

This moves the empty results check after filtering against harvested dataverses, resulting in a 'Sorry, no results were found." API response instead of the previous server error.

This improves #11083  but does not close it (non-harvested datasets in harvested dataverses are still filtered).

See the issue for steps to reproduce.

No interface change

This could be mentioned (see snippet) but has no major impact.

**Additional documentation**:

I updated the IT test with some extra use cases. I'm not actually testing the harvested dataverse case as I am not sure if the expected result is truly the same as the current result.

Bug mentioned at https://dataverse-guide--11086.org.readthedocs.build/en/11086/api/native-api.html#mydata


pdurbin:
@plecor thanks for the PR! ❤️ We'll talk about it soon in a bug triage meeting. I went ahead and threw a small size (3) on it.

testDeleteFile is failing but please don't worry about that. It's unrelated and we plan to fix it in #11081.

sekmiller:
Hi @plecor, Thanks for this PR. We have recently released version 6.5 of the Dataverse Software so please update your branch so that it will be ready for QA. Thanks again. (I am in the process of reviewing your submission, but it is likely that we won't be able to merge it until early in the coming year.)

sekmiller (approved):
Thanks for the update! Looks good. passing to QA.

PR #11024, opened 2024-11-15
Format Solr schema.xml using update-fields.sh on Linux

pdurbin:
Preview docs at https://dataverse-guide--11024.org.readthedocs.build/en/11024/developers/tips.html#updating-the-solr-schema-developers

**What this PR does / why we need it**:

We need a consistent way to format our Solr schema.xml file. If we have some people using Mac and others using Linux, the `git diff` output will be harder to read due to sorting differences.

**Which issue(s) this PR closes**:

- Closes #11023

**Special notes for your reviewer**:

No need to review this yet. It's a draft. I'm waiting for this PR to be merged first:

- #10887

**Suggestions on how to test this**:

Use the schema.xml file in this PR. It should work the same.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No.

**Is there a release notes update needed for this change?**:

No, this is dev territory.

**Additional documentation**:

None.

PR #11096, opened 2024-12-12
IQSS/11095- Account for multivalue needed by cvoc scripts

### LINKED ISSUE #11095: External CVoc - indexing can fail for single valued fields
**What steps does it take to reproduce the issue?** Configuring a CVoc script on a field that is not multivalued will fail if the script is configured to add more than one value to the index, i.e. it adds both the identifier and a name, or includes translations. An example would be using the [depositorOrcid](https://github.com/gdcc/dataverse-external-vocab-support/blob/main/examples/config/depositorOrcid.json) example :CVocConf setting.

* When does this issue occur? When an identifier (vs a free text entry) is entered.

dataset page

* What happens?
indexing fails

all

develop/6.5+

**Screenshots:**

*

Yes

qqmyers:
**What this PR does / why we need it**: This PR changes the /api/admin/index/solr/schema output to correctly indicate that a field should be multivalued when it is the term-uri field of an external controlled vocabulary config. (Note the format of the output/etc. is not changed, just the content - same as if a metadatablock were editing to make a field multivalued.)

- Closes #11095

**Suggestions on how to test this**:  Configure the ORCID script on the depositor field as described in the issue, test that indexing doesn't work now, works after the update and updating the solr schema/restarting solr.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Index.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Index.java
@@ -475,7 +475,7 @@ public class Index extends AbstractApiBean {
                  */
                 logger.info("email type detected (" + nameSearchable + ") See also https://github.com/IQSS/dataverse/issues/759");
             }
-            String multivalued = Boolean.toString(datasetFieldType.getSolrField().isAllowedToBeMultivalued()|| cvocTermUriMap.containsKey(datasetFieldType.getId()));
+            String multivalued = Boolean.toString(datasetFieldType.getSolrField().isAllowedToBeMultivalued() || cvocTermUriMap.containsKey(datasetFieldType.getId()));
             // <field name="datasetId" type="text_general" multiValued="false" stored="true" indexed="true"/>
             sb.append("    <field name=\"" + nameSearchable + "\" type=\"" + type + "\" multiValued=\"" + multivalued + "\" stored=\"true\" indexed=\"true\"/>\n");
         }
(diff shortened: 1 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Index.java
```diff
@@ -474,7 +475,7 @@ public String getSolrSchema() {
                  */
                 logger.info("email type detected (" + nameSearchable + ") See also https://github.com/IQSS/dataverse/issues/759");
             }
-            String multivalued = datasetField.getSolrField().isAllowedToBeMultivalued().toString();
+            String multivalued = Boolean.toString(datasetFieldType.getSolrField().isAllowedToBeMultivalued()|| cvocTermUriMap.containsKey(datasetFieldType.getId()));
```
pdurbin, line 478:
```suggestion
            String multivalued = Boolean.toString(datasetFieldType.getSolrField().isAllowedToBeMultivalued() || cvocTermUriMap.containsKey(datasetFieldType.getId()));
```

pdurbin (commented):
I haven't run the code but I assume this is a no-op if a developer were to recreate the schema.xml file using the process I'm suggesting developers follow in this pull request (please take a look):

- #11024

That is, it's a no-op because we don't ship any sort of cvoc config out of the box. It's a feature that installations opt-in to.

I'm almost ready to send this to QA (test are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11096/1/testReport/ ) but I did leave a few tiny suggestions.

qqmyers:
Yes it's a no-op, but if Dataverse ever does ship with some external cvoc scripts registered, the right thing to do would be to run update-fields.sh on an instance with those scripts configured - so consistent with #11024.

pdurbin (approved):

PR #11099, opened 2024-12-16
API Guide: update auth section on bearer tokens

qqmyers:
This was a suggested edit for #10972 that I'm guessing wasn't seen before QA/merge. 

Preview at https://dataverse-guide--11099.org.readthedocs.build/en/11099/api/auth.html#bearer-tokens

### DIFF
--- a/doc/sphinx-guides/source/api/auth.rst
+++ b/doc/sphinx-guides/source/api/auth.rst
@@ -87,22 +87,20 @@ To register a new user who has authenticated via an OIDC provider, the following
 
   curl  -H "Authorization: Bearer $TOKEN" -X POST http://localhost:8080/api/users/register --data '{"termsAccepted":true}'
 
-If the feature flag ``api-bearer-auth-handle-tos-acceptance-in-idp``` is disabled, it is essential to send a JSON that includes the property ``termsAccepted``` set to true, indicating that you accept the Terms of Use of the installation. Otherwise, you will not be able to create an account. However, if the feature flag is enabled, Terms of Service acceptance is handled by the identity provider, and it is no longer necessary to include the ``termsAccepted``` parameter in the JSON.
-
-In this JSON, we can also include the fields ``position`` or ``affiliation``, in the same way as when we register a user through the Dataverse UI. These fields are optional, and if not provided, they will be persisted as empty in Dataverse.
-
-There is another flag called ``api-bearer-auth-provide-missing-claims`` that can be enabled to allow sending missing user claims in the registration JSON. This is useful when the identity provider does not supply the necessary claims. However, this flag will only be considered if the ``api-bearer-auth`` feature flag is enabled. If the latter is not enabled, the ``api-bearer-auth-provide-missing-claims`` flag will be ignored.
-
-With the ``api-bearer-auth-provide-missing-claims`` feature flag enabled, you can include the following properties in the request JSON:
+By default, the Bearer token is expected to include the following claims that will be used to create the user account:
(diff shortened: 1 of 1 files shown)


pdurbin (commented):
I'm adding my two cents but stopping short of approving this because I think @GPortas or someone closer to this work should affirm it's accurate.

doc/sphinx-guides/source/api/auth.rst
```diff
@@ -87,22 +87,20 @@ To register a new user who has authenticated via an OIDC provider, the following
...
 
 - ``username``
 - ``firstName``
 - ``lastName``
 - ``emailAddress``
 
-If properties are provided in the JSON, but corresponding claims already exist in the identity provider, an error will be thrown, outlining the conflicting properties.
+The one parameter required by default is ``termsAccepted`` which must be set to true, indicating that the user has seen and accepted the Terms of Use of the installation. 
+
+If the feature flag ``api-bearer-auth-handle-tos-acceptance-in-idp`` is enabled (along with the ``api-bearer-auth`` feature flag), Dataverse assumes that the Terms of Service acceptance was handled by the identity provider, e.g. in the OIDC ``consent`` dialog, and the ``termsAccepted`` parameter is not needed.
```
johannes-darms, line 99: @pdurbin: Could we do this? 
"If the feature flag ``api-bearer-auth-handle-tos-acceptance-in-idp`` is enabled (along with the ``api-bearer-auth`` feature flag), Dataverse assumes that the Terms of Service acceptance is handled by the Identity Provider and therefore the ``termsAccepted'' parameter is not needed."

The current process is not true authentication. The system receives a token that may not be intended for it, and uses it to create an account without prompting. The real authentication takes place somewhere else, we assume the SPA, where hopefully the consent screen is used. With this formulation, the behaviour is less ambiguous, since Dataverse itself has no means of using a consent screen.

pdurbin: @qqmyers what do you think?

qqmyers: This is a documentation PR trying to clarify what has already been implemented/merged in [#10972](https://github.com/IQSS/dataverse/pull/10972). 

If there's something to change other than tweaking language, it should probably be a new issue (which should probably be in the dataverse-security repo). FWIW: I could be wrong, but I don't believe that "The system receives a token that may not be intended for it, and uses it to create an account without prompting" is true - the received token is used to call the userinfo API of registered providers only, and only if one of them returns a valid userinfo response is an account created.

pdurbin:
> I'm adding my two cents but stopping short of approving this because I think @GPortas or someone closer to this work should affirm it's accurate.

At standup @GPortas said he'd take a look. Thanks!

GPortas (approved):
Looks good to me. The new language tweaks make the behavior much clearer.

doc/sphinx-guides/source/api/auth.rst
johannes-darms: > FWIW: I could be wrong, but I don't believe that "The system receives a token that may not be intended for it, and uses it to create an account without prompting" is true - the received token is used to call the userinfo API of registered providers only, and only if one of them returns a valid userinfo response is an account created.

This scenario only applies if a third party has a valid token (which stays valid during the user info API calls) and send it without user consent to a dataverse instance. This is to be honest a malicious scenario, which most likely will not occur.

ofahimIQSS:
proof read, no issues found. merging pr

PR #11114, opened 2024-12-19
IQSS/11113 fix file perm doc creation

### LINKED ISSUE #11113: Indexing creates unused permissions docs
When a file is only in one of the indexed versions (draft, last released), Dataverse currently creates a permissions doc in solr for the file for both versions. This results in an orphaned permission doc that can be detected via /api/admin/index/status and removed via /api/admin/index/clear-orphans but will otherwise remain in solr (reindexing only removes permissions docs that should be there).

6.5/develop - probably a long-standing bug

**Are you thinking about creating a pull request for this issue?** PR tbd

qqmyers:
**What this PR does / why we need it**: This PR adds a check in the indexing code to only create a file permission doc in solr only for the datasetversions the file is in.

- Closes #11113

**Special notes for your reviewer**: As the note in the code suggests, there are cleaner ways to do this but since this is a relatively rare case (indexing when there is a currently draft version where a file has been added/dropped) I'm not sure refactoring to get more efficiency is worth the effort. (Not to say that adding/removing files is rare, but having two versions to index only occurs while there is a draft and presumably a draft like this doesn't often stay around forever, so I'm guessing this affects only a small number of files per index run. With the bug, the orphan perm docs do accumulate over time though, so fixing the bug somehow is worthwhile.)

**Suggestions on how to test this**: Create a dataset(s), where the draft version has files not in the original published version or vice versa. Then run /api/admin/index/status and you should see orphaned perm docs (docs in solr but not in the db). After the PR, this should not happen. (If you use the same solr, run /api/admin/index/clear-orphans and either edit the dataset to cause reindexing or just reindex it via the api.)

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11113-avoid-orphan-perm-docs.md
+++ b/doc/release-notes/11113-avoid-orphan-perm-docs.md
@@ -1,0 +1,5 @@
+This release fixes a bug that caused Dataverse to generate unnecessary solr documents for files when a file is added/deleted from a draft dataset. These documents could accumulate and potentially impact performance.
+
+Assuming the upgrade to solr 9.7.0 also occurs in this release, there's nothing else needed for this PR. (Starting with a new solr insures the solr db is empty and that a reindex is already required.)
+
+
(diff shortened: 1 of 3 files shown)


sekmiller (approved):

ofahimIQSS:
image

Just wanted to confirm, these datafiles should be displayed (permissionsInIndexButNotDatabase)

qqmyers:
After the PR, calling the status API should not show new permission docs in index but not db, but if you have not run api/admin/index/clear-orphans there could be ones from before the PR. Normal reindexing would not get rid of them.

ofahimIQSS:
Merging PR - no issues observed

PR #11116, opened 2024-12-19
Correct /api/roles API to respond with 403 Forbidden (not 401 Unauthorized) when auth is good but no permission

### LINKED ISSUE #10340: Incorrect API error response when forbidden
**What steps does it take to reproduce the issue?** Discovered in code review of #10299:
https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L604 and probably https://github.com/IQSS/dataverse/blob/82585f91e2a4367b284350c2b8923387a507685e/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L797 result in 401/Unauthorized error codes rather than (the correct) 403/Forbidden.

* When does this issue occur?
~Any API call resulting in a PermissionException in the underlying command when the user does not have sufficient permission.

v6.1

pdurbin:

The [Show Role](https://dataverse-guide--11116.org.readthedocs.build/en/11116/api/native-api.html#show-role) API endpoint was returning 401 Unauthorized when a permission check failed. This has been corrected to return 403 Forbidden instead. That is, the API token is known to be good (401 otherwise) but the user lacks permission (403 is now sent). See also the [API Changelog](https://dataverse-guide--11116.org.readthedocs.build/en/11116/api/changelog.html).

- Closes #10340

Only the one API endpoint was affected by this change:

Screenshot 2024-12-20 at 2 34 14 PM

I added a test for it.

Also, the "Show Role" code suggests that aliases are supported but they don't seem to work. So when I updated the docs I made sure to use an example with an ID.

Try the "Show Role" API and check if the status code is 403 or 401. See the updated docs for global roles vs roles that are created in a collection. It's a bit confusing, to be honest. 😬 

No.

Yes, included.

**Additional documentation**:

I updated the Show Role docs while I was in there: https://dataverse-guide--11116.org.readthedocs.build/en/11116/api/native-api.html#show-role

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -424,13 +424,13 @@ Creates a new role under Dataverse collection ``id``. Needs a json file with the
   export SERVER_URL=https://demo.dataverse.org
   export ID=root
 
-  curl -H "X-Dataverse-key:$API_TOKEN" -X POST "$SERVER_URL/api/dataverses/$ID/roles" --upload-file roles.json
+  curl -H "X-Dataverse-key:$API_TOKEN" -H "Content-type:application/json" -X POST "$SERVER_URL/api/dataverses/$ID/roles" --upload-file roles.json
 
 The fully expanded example above (without environment variables) looks like this:
 
 .. code-block:: bash
 
-  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X POST -H "Content-type:application/json" "https://demo.dataverse.org/api/dataverses/root/roles" --upload-file roles.json
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -H "Content-type:application/json" -X POST "https://demo.dataverse.org/api/dataverses/root/roles" --upload-file roles.json
 
 For ``roles.json`` see :ref:`json-representation-of-a-role`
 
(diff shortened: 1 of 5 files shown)


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4563,12 +4563,22 @@ Create Role
...
 
+.. _show-role:
+
 Show Role
 ~~~~~~~~~
 
-Shows the role with ``id``::
+You must have ``ManageDataversePermissions`` to be able to show a role that was created using :ref:`create-role-in-collection`. Global roles (:ref:`create-global-role`) only be shown with a superuser API token.
+
+A curl example using an ``ID``:
```
sekmiller, line 4573: Should we say that you may also use a role's alias?

pdurbin: @sekmiller it didn't work for me. This is what I put in the description of this PR:

"Also, the "Show Role" code suggests that aliases are supported but they don't seem to work. So when I updated the docs I made sure to use an example with an ID."

pdurbin: @sekmiller yes, as discussed in person. Good catch. Please see the commit message in d4f9189280 where I added this and made a few more improvements while I was in there.

ofahimIQSS:
image
tested in internal

ofahimIQSS:
Dont know if its related to this ticket but here is an observation I had. I am unable to edit the name of a newly created role. Says,  The role was not able to be saved. Role with this alias already exists. –


https://github.com/user-attachments/assets/fc5f7ce4-479e-4326-bfaa-f672a9925bea

ofahimIQSS:
Reproduced issue in Demo without PR:
image

Tested in Internal with PR
image
image

Tests are passing - no issues observed during testing.

PR #11119, opened 2024-12-20
Fix JAR build issues

GPortas:

- Closes #

**Additional documentation**:

### DIFF
--- a/src/main/resources/META-INF/services/edu.harvard.iq.dataverse.pidproviders.PidProviderFactory
+++ b/src/main/resources/META-INF/services/edu.harvard.iq.dataverse.pidproviders.PidProviderFactory
@@ -1,0 +1,6 @@
+edu.harvard.iq.dataverse.pidproviders.doi.crossref.CrossRefDOIProviderFactory
+edu.harvard.iq.dataverse.pidproviders.doi.datacite.DataCiteProviderFactory
+edu.harvard.iq.dataverse.pidproviders.doi.ezid.EZIdProviderFactory
+edu.harvard.iq.dataverse.pidproviders.doi.fake.FakeProviderFactory
+edu.harvard.iq.dataverse.pidproviders.handle.HandleProviderFactory
+edu.harvard.iq.dataverse.pidproviders.perma.PermaLinkProviderFactory
--- a/src/main/resources/META-INF/services/io.gdcc.spi.export.Exporter
+++ b/src/main/resources/META-INF/services/io.gdcc.spi.export.Exporter
@@ -1,0 +1,10 @@
+edu.harvard.iq.dataverse.export.DCTermsExporter
+edu.harvard.iq.dataverse.export.DDIExporter
+edu.harvard.iq.dataverse.export.DataCiteExporter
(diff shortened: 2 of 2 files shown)


qqmyers:
Hmm - the [@AutoService(PidProviderFactory.class) annotation](https://github.com/IQSS/dataverse/blob/637c21684d45d8d176cd13bec1481877311959df/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/datacite/DataCiteProviderFactory.java#L10) (and similar for exporters) is supposed to do that. Do you have some problem with the [autoservice dependency](https://github.com/IQSS/dataverse/blob/637c21684d45d8d176cd13bec1481877311959df/pom.xml#L528-L534)?

pdurbin:
@GPortas this looks like progress toward a solution! I'm going to also add a link to https://stackoverflow.com/questions/66000185/spring-factories-in-meta-inf-folder-not-added-to-target-folder-after-packaging/66000292#66000292 which you mentioned [in Slack](https://iqss.slack.com/archives/C010LA04BCG/p1734368107392029).

I'm sure @qqmyers is right that something is amiss with `@AutoService` not working properly. I checked and we first added `@AutoService` in 2016 in https://github.com/IQSS/dataverse/commit/e560a34e89b12a08b0e936e0cc8bd429f7a8c7c5 as part of #3308 and we talk about it in our docs: https://guides.dataverse.org/en/6.5/developers/metadataexport.html

We haven't bumped the version since then, since 2016. We're still on 1.0-rc2 even though https://github.com/google/auto/blob/main/service/README.md and https://github.com/google/auto/releases show never versions. The latest seems to be 1.11.0. Maybe it's worth trying a new version? To me, a version bump would be better than having to maintain a static list of files, as you've done (understandably!) in this PR.

It's so strange that `@AutoService` has been working for all of us all this time and then it suddenly stopped working for you! Good job tracking this down!

qqmyers:
FWIW: #10713 updates autoservice to 1.1.1 which we've been using at QDR.

GPortas:
> FWIW: #10713 updates autoservice to 1.1.1 which we've been using at QDR.

Yes, in one of my tests, I updated the dependency to version 1.1.1, but it made no difference.

> Hmm - the [@AutoService(PidProviderFactory.class) annotation](https://github.com/IQSS/dataverse/blob/637c21684d45d8d176cd13bec1481877311959df/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/datacite/DataCiteProviderFactory.java#L10) (and similar for exporters) is supposed to do that. Do you have some problem with the [autoservice dependency](https://github.com/IQSS/dataverse/blob/637c21684d45d8d176cd13bec1481877311959df/pom.xml#L528-L534)?

I think for some reason I don't know yet it isn't working for me...

pdurbin:
> Yes, in one of my tests, I updated the dependency to version 1.1.1, but it made no difference.

Once we're on the latest it would be legit to open an issue at https://github.com/google/auto/issues

davidverdu:
I had the same issue in my local environment and this patch fixed the error. Thank you for posting this PR.

PR #11068, opened 2024-12-06
IQSS/8431 Optional Version Note

qqmyers:
**What this PR does / why we need it**: This PR adds a version note to be able to record the reason a version was created/how it differs from the prior version. The functionality is optional (turned on via a feature flag.

Per tech hour discussion, in addition, an attempt has been made to clear some technical debt - where the reason for deacccessioning was stored in a column called versionnote, and the forwarding url for the deaccessioned version was stored in the archivenote field, while there was also a deaccessionlink column that appears to have the same purpose but is only visible in the API/JSON output for a dataset version. 

**Which issue(s) this PR closes**:

- Closes #8431
- Closes #11310

**Special notes for your reviewer**: AFAIK, the deaccessionlink column was only used in v3 while the archivenote field was used in v4/in the deaccession UI. It looks like at one point (~v4.13) there was an attempt to validate the archiveNote to require a URL but there were legacy values that are not URLs so that was disabled. In cleaning things up, I've added steps in the flyway script to combine the deaccessionlink and archivenote fields (presumably only one was ever filled out for any given datasetversion) and to address the non-url legacy values in the UI by not trying to turn them into a link as is done now (presumably causing some error/broken link).

**Suggestions on how to test this**: Regression test that deaccessioning still works, existing deaccession reason and forwarding URLs appear for existing/older deaccessioned datasets (and new ones). Then turn on the feature via the enable-version-note feature flag. Verify that a version Note can be added when viewing a draft version via the version table on the dataset page, as well as during publication. Verify that the API can be used to add a versionNote to draft (for those who can edit the dataset) and for any version for superusers, e.g. 
`curl -X PUT -d 'No substantive changes were made in the project. Depositor replaced files with new versions correcting typos' http://localhost:8080/api/datasets/<id>/versions/2.0/versionNote -H "X-Dataverse-Key:$KEY"`

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
<kbd>image</kbd>

<kbd>image</kbd>

<kbd>393402921-a802f2cd-457a-4cf3-b6d3-4334a5033bcb</kbd>

(images from QDR)
**Is there a release notes update needed for this change?**: included

**Additional documentation**: user/api docs tbd

PR #11125, opened 2024-12-30
Globus support: download optimizations

### LINKED ISSUE #11057: Globus integration: improve handling of downloads
Having had another collection of NESE-stored/Globus-accessible datasets published at IQSS, with the combination of a large volume of data, large numbers of files per dataset AND the apparent popularity of the data with actual users, some limitations of the Globus download framework have become apparent. 

(starting a list of issues that need to be addressed below; work in progress)

1. The same thing that was addressed in #10781 (merged in Sept.), is clearly a problem with downloads as well. Even when everything is working as it should, the reliance on continuous looping for the duration of the remote transfer in the current implementation is bound to cause problems. So, the same async. framework where the state of an ongoing upload is saved in the database needs to be implemented for downloads as well. 

2. When problems arise (such as network problems; or the issues with the Globus service on the data storage end, as the case may have been last week), there are assorted problems with how Dataverse handles such. A simplest of examples, when Globus Service gets back `"status": "ACTIVE", … "nice_status": "CONNECT_FAILED"` when checking on an ongoing download task, it assumes that it has failed beyond recovery and proceeds to remove the permission. (All it means is that there was a failure to connect to the remote Globus service, but "ACTIVE" means just that - the remote client will keep trying to reconnect; which is bound to keep failing with the permission permanently removed, even if the service becomes available).

landreev:

The PR improves the reliability of the Globus *download* framework, primarily by extending the new task-monitoring system first implemented for *uploads* in #10781. There are a few other improvements. For example, it fixes a somewhat exotic bug where Globus downloads weren't counted, but only for multi-file downloads and only when a guestbook popup was enabled. 
The improvements are implemented in response to/based on the experience with the NESE Data Storage integration with Dataverse at IQSS. Most of these are already in prod. use there, deployed as an experimental beta build. 

- Closes #11057

The single most important improvement is that, similarly to what was implemented in #10781 for uploads, the ongoing download tasks can now be monitored asynchronously, with the state saved in the database. This makes the management of the temporary access rules more robust, guaranteeing that Dataverse will register the completion of each task, even if there was a server restart in between. 

Assorted other fixes and improvements were added. For example, it is now possible for a user to have simultaneous downloads runnig on files from the same dataset (when the task state is saved in the database, upon completion of a task it is easy to check if any other active tasks are using the same access rule and, if so, avoid deleting it). I fixed something misguided I did in #10781 when I first implemented saving the task state in the database: I misunderstood how the client tokens worked; and tried to save the token for each task, thinking that it needed to be reused throughout the life of the task. In reality, the same token can be used for multiple tasks on the same endpoint; also, for a long-running task that saved token has a good chance of expiring - and I didn't have a provision for that. Now the monitoring service simply caches the access tokens for each endpoint that it manages, and refreshes them as needed. General error handling, logging and state checking has been improved. 

dataverse-internal has active Globus configuration tied to a fully-functioning remote storage volume at NESE, identical to our prod. volume there. There are existing datasets with multiple large-ish (100MB+) Globus-stored files (for example: https://dataverse-internal.iq.harvard.edu/dataset.xhtml?persistentId=doi:10.70122/FK2/QZQPQE); more data can be uploaded for testing as needed. QA would involve verifying that such files can be downloaded; with an emphasis on repeating downloads from the same dataset, and perhaps parallel downloads of different files from the same dataset at the same time. For the end user, everything should work as described in the [Globus download instruction](https://github.com/IQSS/dataverse.harvard.edu/blob/master/doc/globus/download-quickstart.md) (written for the prod. users, but fully applies to the test configuration on internal). 

One extra thing  not directly related to the Globus tech fixed in this PR: there was a glaring inefficiency in the dataset page code in how it was looking up various external tools and previewers for each file on the page, introduced 6 releases ago (!) but then forgotten about. It was re-discovered during this effort and I committed the fixes since they were trivial. So, please check on the previewers and such, before and after, and confirm that they are still shown on the page properly. 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -751,20 +828,56 @@ public void globusUpload(JsonObject jsonData, Dataset dataset, String httpReques
...
+        while (taskState == null && retries < retriesLimit) {
+            try {
+                taskState = getTask(endpoint.getClientToken(), taskIdentifier, globusLogger);
+            } catch (ExpiredTokenException ete) {
+                // We have just obtained this token seconds ago - this shouldn't 
+                // really happen - ? 
+            }
+            retries++;
+            try {
+                Thread.sleep(3000);
```
stevenwinship, line 856: you should only sleep if the taskState == null and there are no more retries. This will delay the loop process an extra 3 seconds every time for no reason.

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -751,20 +828,56 @@ public void globusUpload(JsonObject jsonData, Dataset dataset, String httpReques
...
             Long datasetId = rulesCache.getIfPresent(ruleId);
             if (datasetId != null) {
-                // Will not delete rule
+                // This will only "invalidate" the local cache entry, will not 
+                // delete or invalidate the actual Globus rule
                 rulesCache.invalidate(ruleId);
             }
+        } else {
+            // we'll proceed anyway, under the assumption that we will make 
+            // another attempt to look it up later
```
stevenwinship, line 880: I'm not crazy about code blocks with no actual code but ok. Maybe add a logger.fine saying the rule was not found?

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -1181,73 +1301,110 @@ public void globusDownload(String jsonData, Dataset dataset, User authUser) thro
...
+            } catch (ExpiredTokenException ete) {
+                // We have just obtained this token seconds ago - this shouldn't 
+                // really happen (?)
+                endpoint = getGlobusEndpoint(dataset);
+            }
+            
+            retries++;
+            try {
+                Thread.sleep(3000);
+            } catch (InterruptedException ie) {
```
stevenwinship, line 1344: Same as above. This is an extra 3 second delay after the taskState is found.

src/main/webapp/filesFragment.xhtml
```diff
@@ -544,7 +544,12 @@
                     
                     <!-- TO-DO FIX RSYNC LOGIC !fileMetadata.dataFile.filePackage or
                                                     fileMetadata.dataFile.filePackage and systemConfig.HTTPDownload -->
-                     <ui:fragment rendered="#{DatasetPage.isShowQueryButton(fileMetadata.dataFile.id) and fileDownloadHelper.canDownloadFile(fileMetadata)}">
+                    <ui:fragment rendered="#{fileMetadata.dataFile.tabularData
+                                             and fileMetadata.dataFile.released
+                                             and not fileMetadata.dataFile.restricted
+                                             and not dataFileServiceBean.isActivelyEmbargoed(fileMetadata)
+                                             and DatasetPage.isShowQueryButton(fileMetadata.dataFile.id)
```
qqmyers, line 551: isShowQueryButton handles all of these (and retention period) except for tabularData, which is at least partially handled by registering for specific mime types. (Is the concern that it shows when ingests fails? (Does the Ask the Data' tool fail if ingest fails or can it do something with just the file?).
Whatever the correct logic - perhaps any changes can just go in isShowQueryButton() ?

src/main/java/edu/harvard/iq/dataverse/globus/TaskMonitoringServiceBean.java
```diff
@@ -53,45 +56,160 @@ public void init() {
...
+        });
+        
+        tasks.forEach(t -> {
+
+            GlobusTaskState retrieved = null; 
+            int attempts = 2; 
+            // we will make an extra attempt to refresh the token and try again
+            // in the event of an exception indicating the token is stale 
+            
+            String globusClientToken = getClientTokenForStorageDriver(t.getDataset(), false);
```
qqmyers, line 158: Not a big deal, but this section is common to up and download.

stevenwinship (changes requested):

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -209,6 +233,24 @@ public JsonObject requestAccessiblePaths(String principal, Dataset dataset, int
...
+        if (requestPermStatus == 409) {
+            // This is a special case - a 409 *may* mean that the rule already
+            // exists for this endnote and for this user (if, for example, 
+            // Dataverse failed to remove it after the last upload has completed). 
+            // That should be ok with us (but let's confirm that is indeed the 
+            // case; alternatively it may mean that permissions cannot be issued 
+            // for some other reason):
+            String ruleId = getRuleId(endpoint, principal, "rw");
+            if (ruleId != null) {
+                requestPermStatus = 201;
```
qqmyers, line 246: Should this get logged? Something went wrong with a prior removal (which doesn't have to stop this request but shouldn't be ignored as it's a bug/problem if perms aren't going away)

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
```diff
@@ -639,7 +691,27 @@ public int setPermissionForDownload(Dataset dataset, String principal) {
...
+            // Dataverse failed to delete it after the last download by this 
+            // user, or if there is another download from the same user 
+            // currently in progress. The latter is now an option when the 
+            // "asynchronous mode" is enabled for task monitoring (since all the 
+            // ongoing tasks are recorded in the database, it is possible to check 
+            // whether it is safe to delete the rule on completion of a task, vs.
+            // if other tasks are still using it). If that's the case, we'll 
+            // confirm that the rule does exist and assume that it's ok to 
+            // proceed with the download. 
+            if (FeatureFlags.GLOBUS_USE_EXPERIMENTAL_ASYNC_FRAMEWORK.enabled()) {
```
qqmyers, line 707: So unlike the upload case, this is still failing if Dataverse failed to delete it after the last download? It's only allowing multiple parallel downloads. (So no reason to log here as this is valid versus evidence of some problem).

qqmyers (commented):
Looks good - I suggested a couple minor changes but I think the overall logic in the updates make sense.

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
landreev: Correct, will fix.

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
landreev: With this change it is NOT failing if the rule is already there (whether because Dataverse failed to delete it after a past download; or because it was created for a download that's still running) and the new monitoring mode is in use.

It is these extra lines above that actually make it possible to have multiple downloads from the same dataset. Prior to this pr, Dataverse would try to request an access rule, get back a 409, and pass it to the app; which in turn would tell the user that "download could not be started". Now the code intercepts the 409, verifies that the "r" permission for this user/this folder does exist, then returns 201, and the app proceeds to initiate the download. On the completion of the task, the monitoring service code checks if there are still any other tasks in the database-stored queue using the same rule, and only deletes it if there are none.

qqmyers: Got it. So a) why the check for the feature flag for download and not upload? and b) can you distinguish between parallel downloads and a prior failure (would the ruleId be in the cache for multiple downloads but not for a failed delete?)? As with the upload case, I'm worried that problems that leave a permission in place will get ignored because they now don't stop a new request and there's no logging.

src/main/webapp/filesFragment.xhtml
landreev: Funny thing you noticed this... I put together these logic modes in response to something you pointed out to me a few releases back - namely that the rendering logic for this button introduced an ungodly number of datafileservice.find() calls (it's a case of  x7 jsf renderings); the whole thing was added to a release at the last minute, AND after I had gone to some trouble reducing the number of database lookups in the page. The above did not completely fix it, but at least reduced the lookups, by narrowing when the method is called to tabular, released, ... etc. files (the tool is only supposed to run on public tab. files, etc. - the logic is inside the method; unfortunately, it's only applied there after the db. lookups). So, I added these bandaid lines to the custom changes I deploy in prod. at iqss... and forgot about it/never made a real pr. These changes made it back into this branch kind of by accident; but when I noticed I figured this would be a good opportunity to check it in. 

Ideally, that method should not be called on the db id of the filemetadata in the first place; the page has access to actual entities. The main reason I never bothered with a cleaner fix was that it felt wrong to invest any work into jsf; since it was allegedly going awy any sec. 

Was a longer explanation than the issue is worth, for sure...

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
landreev: a) For downloads: if the old-style continuous task monitoring is used, the code is falling back to the old behavior; under the assumption that there may be another active task using the rule, and then it would be impossible to prevent the one that finishes first from deleting the rule and killing the other task(s). 
I'm not doing that for uploads, under the assumption that it is impossible to start more than one - since the dataset gets locked. So, if we found an existing  rule, it must be from a finished or failed task. 
I agree that this needs to be logged; and I should probably add a warning to the guide somewhere to keep an eye for these warnings. 
b) I haven't been able to come up with a way to distinguish between a stale vs. still in use rule, without the database record. I don't think we can rely on the cache for this. We do remove the rule from the cache when we start transfers; that can be modified somehow, but then we are back to relying on the server staying up (we now have real-life cases of downloads, and occasional uploads that take days); plus the issue of multiple servers at IQSS and possibly other places. 

BTW, do you know/remember why the rules Dataverse creates have no expiration dates? - making them expire/disappear eventually on their own would help, I think. 

c) I'm realizing we need a dedicated admin api call for clearing up stale access rules. I've been doing that outside of Dataverse, but it would be trivial to add. Again, I'll only be able to tell for sure if it's stale or not with the db tracking on. But then this is something I'll likely feel comfortable dropping the "experimental" label from and making the default behavior in 6.7.

src/main/webapp/filesFragment.xhtml
landreev: (if I have a sec, I'll go ahead and rewrite the `DatasetPage.isShowQueryButton()` method to be less of a hog, now that I had to remember the whole thing in vivid detail)

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
qqmyers: re: expiration - I don't know why we don't use them - perhaps no one saw that option? Or it's hard to predict how long you need? Seems like some conservative passive expiration would make sense though.

src/main/webapp/filesFragment.xhtml
qqmyers: Hmm - is that because we use the file id and then look it up in that method? Could we make it use the Datafile as a param and avoid the lookups and keep it all in the java code?

landreev: That's what I was trying to say last night, yes; that's how it should've been implemented in the first place. 
The conversation in question happened 1.5 yrs ago, right after 5.14, apparently: 
Screen Shot 2025-02-26 at 9 00 08 AM
Screen Shot 2025-02-26 at 9 02 46 AM

src/main/java/edu/harvard/iq/dataverse/globus/GlobusServiceBean.java
landreev: We actually want to have a short sleep there before making that first call; so I rearranged the order/got rid of some extra sleep lines outside of that block.
The redundant sleep lines there we a bit messy. But actual extra  sleep time doesn't hurt anything there since the methods are executed in the background, via the `@Asynchronous` annotation.

src/main/webapp/filesFragment.xhtml
landreev: There was also this in the same thread: 

Screen Shot 2025-02-26 at 10 48 13 AM

... might as well, at this point.

landreev: I added a sentence to "how to test" explaining that QA should include checking on the previewers and tools before and after.

stevenwinship (approved):

ofahimIQSS:
Getting internal server error when I push PR to internal:

qqmyers:
Looks like the error is not related to the PR - a consequence of testing #11068 first which drops a datasetversion column.

ofahimIQSS:
Currently there is an issue holding this up for testing:
https://github.com/IQSS/dataverse/pull/11282#issuecomment-2704078292

ofahimIQSS:
@landreev How would I get past this?:

https://github.com/user-attachments/assets/0abf4437-7d2f-46db-aca2-84bf3e46780c

ofahimIQSS:
Testing Evidences:

image
image

Parallel Upload successful with large files:
image
image

ofahimIQSS:
Maybe not related to the PR but here is an observation I made:
When there are too many ineligible files selected on a Globus download, the message box doesn't allow user to scroll down. 

https://github.com/user-attachments/assets/5429ba1d-7461-49a7-8f12-7ef6da31528b

ofahimIQSS:
> Maybe not related to the PR but here is an observation I made: When there are too many ineligible files selected on a Globus download, the message box doesn't allow user to scroll down and hit continue button.

A separate PR has been opened to track this. Merging PR.

landreev:
Thank you for merging the PR and for the detailed report! 
Sorry for the delay, I just got around to looking into this. 
(remember - I am, technically, off today 🥲) 
Special thanks to @qqmyers for the assistance with testing. 

Opening the new issue was the right thing to do, that's what I would have recommended myself. I'll add a comment there.

PR #11134, opened 2025-01-07
IQSS/11133-search-fix

### LINKED ISSUE #11133: Search Service: an error condition under permission query optimization mode was introduced in 6.5
During the cleanup and refactoring of the permission query building code in SearchServiceBean, under #10706, one very specific use case/workaround got overwritten. 
When the permission query optimization mode is enabled (`avoid-expensive-solr-join=true`) AND the user has more than 1 active group, the combined query must be built a certain way that's different from what we've been using otherwise. 
Unfortunately, this was not caught before/during the 6.5 release - because the combination of the above conditions was needed for the issue to manifest itself. 

I have however enabled the optimization on demo, where all the users (including the guest/anonymous user, apparently) are members of multiple groups. So, going forward an issue like this will be caught during the deployment there. 

Will be easy to rectify. Jim and I talked about this yesterday, he already has a fix that appears to be working at QDR.

qqmyers:
**What this PR does / why we need it**: Uses the form of the join query we need that works. (There was a reversion on #10706.) As noted in the issue, this only affects the case when the AVOID_EXPENSIVE_SOLR_JOIN is true and the user is in multiple groups and is not a superuser.

- Closes #11133

**Suggestions on how to test this**:  Create a user in multiple groups that is not a super user, set `dataverse.feature.avoid-expensive-solr-join` to true, restart payara, and verify that this user can see public data and their own datasets. Regression test that other users aren't affected.  Note that `dataverse.feature.add-publicobject-solr-field` must also be set true (and payara restarted) and reindexing done if that wasn't true before. (This flag makes indexing add the publicObject_b=true flag for public objects which is what allows us to avoid the expensive join.)

[edit from @landreev:] Strictly speaking, this PR contains a small change in how search queries are formed when NOT in the "avoid solr join" mode as well; so, yet another regression test would be to also verify that searches are working properly **without** this optimization enabled. It is extremely unlikely that there are any such problems though. 
demo.dataverse.org now has this solr optimization mode enabled, mirroring the production. This means that, going forward, any issue like this will get detected when the release candidate is deployed on demo as part of of the release process. 

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11133-search-fix.md
+++ b/doc/release-notes/11133-search-fix.md
@@ -1,0 +1,3 @@
+### Search fix when using AVOID_EXPENSIVE_SOLR_JOIN=true
+
+Dataverse v6.5 introduced a bug which causes search to fail for non-superusers in multiple groups when the AVOID_EXPENSIVE_SOLR_JOIN feature flag is set to true. This releases fixes the bug.
--- a/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SearchServiceBean.java
@@ -335,9 +335,13 @@ public class SearchServiceBean {
         // -----------------------------------
         // PERMISSION FILTER QUERY
         // -----------------------------------
-        String permissionFilterQuery = this.getPermissionFilterQuery(dataverseRequest, solrQuery, onlyDatatRelatedToMe, addFacets);
-        if (!StringUtils.isBlank(permissionFilterQuery)) {
-            solrQuery.addFilterQuery(permissionFilterQuery);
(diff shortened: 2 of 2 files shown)


landreev (approved):
Thank you for making a speedy PR!
Looks straightforward and appears to be working

ofahimIQSS:
I'm not seeing majority of collections and datasets for a normal user. Only seeing 2 datasets from Jan 6, 2025

set flag to true:

image
Logged into Internal using normaluser1/admin1
image

Here is what a super user sees:
image

ofahimIQSS:
additionally, when i created a collection and a dataset - only the dataset shows up in mydata. Collection is missing. 
image

ofahimIQSS:
PR Looks good - able to see all data for a normal user. Regression tested using other users. Merging. 
image

PR #11136, opened 2025-01-08
fix: correctly set "multiValued" property in Solr schema for deeply nested fields

### LINKED ISSUE #9200: Feature Request/Idea: UI support for nested compound fields

Currently, Dataverse doesn't fully support nested compound fields (i.e. compound fields within compound fields). More precisely, the UI cannot display them, although the API and the internal data model already appear to support them. In some constellations, the SOLR schema is also not updated correctly.

Any user using the UI to view/edit datasets (curator, depositor, guest, superuser...) and API users

We would like to use custom metadata schemas that contain nested compound fields.

DataCite v>=4.1 also contains fields like this, e.g. see "7. Contributor" (0-n) with its children "7.4 nameIdentifier" (0-n) and "7.5 affiliation" (0-n) in https://schema.datacite.org/meta/kernel-4.4/doc/DataCite-MetadataKernel_v4.4.pdf

I have tested the following three basic scenarios of nested compound fields (load custom TSV, upload dataset via API, export dataset via API):

1. compound field nested within compound field
2. like 1. but outer field has `allowmultiple TRUE`
3. like 1. but both fields have `allowmultiple TRUE`

The TSV files and JSON datasets are here: [nested_compound_test.tar.gz](https://github.com/IQSS/dataverse/files/10124812/nested_compound_test.tar.gz)

Although the UI cannot display the nested fields, in scenarios 1 and 3 the export actually matched the initial upload, thus it seems the API and the internal data model already support this feature. In scenario 2 I got an error message at export (`Exception thrown from bean: org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at http://solr:8983/solr/collection1: ERROR: [doc=dataset_9_draft] multiple values encountered for non multiValued field`). After manually editing the SOLR schema file to set `multiValued="true"`, the export succeeded here as well.

The SOLR schema should be updated correctly when a TSV with nested compound fields is uploaded.

The UI should be able to display nested compound fields when creating, editing and viewing a dataset.

Currently, it looks like this (empty because inner children aren't shown):

Screenshot 2022-11-30 at 17-15-55 Add New Dataset - Root

https://github.com/IQSS/dataverse/issues/377 is related. When allowing multiples of a compound child, in some cases the compound child itself should also be a compound, e.g.:

"Author" is a compound field. To allow an author to have multiple identifiers, the children "Identifier Scheme" and "Identifier" should be grouped in a repeatable nested compound field.

(See also "7. Contributor" (0-n), "7.4 nameIdentifier" (0-n), "7.4.a nameIdentifierScheme", "7.4.b schemeURI" in https://schema.datacite.org/meta/kernel-4.4/doc/DataCite-MetadataKernel_v4.4.pdf (present since DataCite v4.1 as property 7.5))

cc @johannes-darms

vera:

This PR fixes a bug in the `/api/admin/index/solr/schema` endpoint reported in #9200. The `multiValued` property was not set correctly for some deeply nested metadata fields, causing indexing errors:

`org.apache.solr.common.SolrException: ERROR: [doc=dataset_1479] multiple values encountered for non multiValued field`

This is because the code only checked for whether the direct parent of a metadata field is multi-valued to see whether a child metadata field must be declared as multi-valued in the Solr schema. However, all ancestor fields (parents, grandparents, etc.) need to be checked. This PR updates the code to handle deeper metadata nesting properly.

This is related to #9200, but doesn't close the issue, because it also describes UI issues related to deeply nested fields, but it's a fix of the Solr schema bug described there.

/

1. Upload TSV metadata block file e.g. [nested_compound_test2.csv](https://github.com/user-attachments/files/18348950/nested_compound_test2.csv) (had to rename to CSV to upload here)
2. Activate metadata block
3. Use `/api/admin/index/solr/schema` to update Solr's `schema.xml`
4. Reload Solr

(see https://guides.dataverse.org/en/latest/admin/metadatacustomization.html#updating-the-solr-schema)

5. Upload dataset as JSON e.g. [nested_compound_test2.json](https://github.com/user-attachments/files/18348987/nested_compound_test2.json)
6. Confirm the dataset indexes without errors

/

/

**Additional documentation**:

/

### DIFF
--- a/doc/release-notes/11136-solr-nested.md
+++ b/doc/release-notes/11136-solr-nested.md
@@ -1,0 +1,1 @@
+Deeply nested metadata fields are not supported but the code used to generate the Solr schema has been adjusted to support them. See #11136.
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldType.java
@@ -541,17 +541,26 @@ public class DatasetFieldType implements Serializable, Comparable<DatasetFieldTy
                 solrType = SolrField.SolrType.FLOAT;
             }
 
-            Boolean parentAllowsMultiplesBoolean = false;
-            if (isHasParent()) {
-                if (getParentDatasetFieldType() != null) {
-                    DatasetFieldType parent = getParentDatasetFieldType();
-                    parentAllowsMultiplesBoolean = parent.isAllowMultiples();
+            Boolean anyParentAllowsMultiplesBoolean = false;
+            DatasetFieldType currentDatasetFieldType = this;
+            // Traverse up through all parents of dataset field type
+            // If any one of them allows multiples, this child's Solr field must be multi-valued
+            while (currentDatasetFieldType.isHasParent()) {
+                if (currentDatasetFieldType.getParentDatasetFieldType() != null) {
+                    DatasetFieldType parent = currentDatasetFieldType.getParentDatasetFieldType();
+                    if (parent.isAllowMultiples()) {
+                        anyParentAllowsMultiplesBoolean = true;
+                        break; // no need to keep traversing
+                    }
+                    currentDatasetFieldType = parent;
+                } else {
+                    break;
                 }
             }
             
             boolean makeSolrFieldMultivalued;
             // http://stackoverflow.com/questions/5800762/what-is-the-use-of-multivalued-field-type-in-solr
-            if (allowMultiples || parentAllowsMultiplesBoolean || isControlledVocabulary()) {
+            if (allowMultiples || anyParentAllowsMultiplesBoolean || isControlledVocabulary()) {
                 makeSolrFieldMultivalued = true;
             } else {
                 makeSolrFieldMultivalued = false;


pdurbin:
Before switching to this branch, I'm just adding a some detail of the errors one sees on the develop branch when trying to create a dataset using the JSON and TSV above:

`dev_dataverse> request: http://solr:8983/solr/collection1`
`dev_dataverse> Remote error message: ERROR: [doc=dataset_3_draft] multiple values encountered for non multiValued field identifier_type2: [ORCID, ORCID]`
`...`
`dev_solr> 2025-01-31 16:43:46.984 ERROR (qtp1331270134-27) [c: s: r: x:collection1 t:null-27] o.a.s.h.RequestHandlerBase Client exception => org.apache.solr.common.SolrException: ERROR: [doc=dataset_3_draft] multiple values encountered for non multiValued field identifier_type2: [ORCID, ORCID]`

Also, I don't think this will surprise anyone, but the JSF UI does not support deeply nested fields. It's completely out of scope to fix this in JSF. It looks like this when you click "create dataset":

Screenshot 2025-01-31 at 11 34 42 AM

ofahimIQSS:
PR looking great. I followed the testing steps and did further regression testing. No issues found. Release snippet looks good as well. 

image

Server log attached 
[server.log --- Testing PR 11136.txt](https://github.com/user-attachments/files/18663288/server.log.---.Testing.PR.11136.txt)

PR #11138, opened 2025-01-08
Search API show_type_counts=true now includes all object types

### LINKED ISSUE #11127: Search API -  Counts per object type behavior
Recently a new query param has been added to the search api in order to return the count per object type(`&show_type_counts=true`).
To fully replicate the JSF version, we should change two different behaviors of how and when this object is returned.

**_First behaviour:_**
We need the 3 object types to always be present in the object regardless of the object types sent by the URL.
For example for this search 👇
https://beta.dataverse.org/api/v1/search?q=*&subtree=root&per_page=10&type=dataverse&type=dataset&show_type_counts=true

We have this response 👇, but we need that even if we have not sent the type=file, the count on them is returned anyway and that applies to all object types.

Screen Shot 2025-01-02 at 10 08 56

**_Second behaviour:_**
When doing a search that finds no results, the object “total_count_per_object_type” is not returned at all.
For example for this search 👇
https://beta.dataverse.org/api/v1/search?q=ioioioioio&subtree=root&per_page=10&type=dataverse&type=dataset&show_type_counts=true

We have this response 👇, I think it might be better to keep sending the object with each property set to 0 even if no results were found, to respect what the API says to return when sending `&show_type_counts=true`

Screen Shot 2025-01-02 at 10 19 25

sekmiller:
**What this PR does / why we need it**: Extends the Search API so that if show_type_counts is true then each dv object type will be returned including when the count is 0 - or even if there are no actual results in the search return.

- Closes #11127 Search API - Counts per object type behavior

**Special notes for your reviewer**: Pretty straight forward. Added tests to Search IT. Did see some weirdness with my local solr install that seem to be unrelated to this, and I was able to resolve them.

**Suggestions on how to test this**: Good examples in the ticket. You'll need to exercise the Search API in a way that the results include only some of the object types or return no records at all

**Is there a release notes update needed for this change?**: Yes, included

**Additional documentation**: the Search API documentation does not show the previous behavior of including the show-type-counts param. Should it be added or is this more like an SPA feature that we don't expect regular api users to use?


src/main/java/edu/harvard/iq/dataverse/api/Search.java
```diff
@@ -211,13 +212,50 @@ public Response search(
...
+                        if (filesMissing) {
+                            objectTypeCounts.add(SearchConstants.UI_FILES, 0);
+                        }
                     }
+
                 }
+                if (showTypeCounts && solrQueryResponse.getTypeFacetCategories().isEmpty()) {
+                    objectTypeCounts.add(SearchConstants.UI_DATAVERSES, 0);
+                    objectTypeCounts.add(SearchConstants.UI_DATASETS, 0);
+                    objectTypeCounts.add(SearchConstants.UI_FILES, 0);
```
g-saracca, line 256: @sekmiller I have one question, the properties are sent with the UI namings, shouldn't we use the ones from the Search API instead of the WEB UI ones?

https://github.com/IQSS/dataverse/blob/5073daa79241fb6602a6f039df37e570b71fc59b/src/main/java/edu/harvard/iq/dataverse/search/SearchConstants.java#L17

sekmiller: I was trying to match what we get here: "total_count_per_object_type":{"Dataverses":105,"Datasets":71,"Files":0}}
via objectTypeCounts.add(facetLabel.getName(), facetLabel.getCount());

facetLabel.getName() returns the capitalized version

g-saracca: That's fine, it was just a suggestion, if it works well that way let's leave it as it is. 
We can map it to another properties in js-dataverse or spa repo layers.

sekmiller: OK. Great. Honestly I was a little unsure of where to get it from, but I didn't want to add another constant anywhere so I used what was available.

g-saracca:
I tested this while extending the Get Collection Items use case in js-dataverse with this PR preview image and it works like a charm. 👏🏼

doc/release-notes/11127-update-search-api-to-show-dvobject-type-counts.md
```diff
@@ -0,0 +1,3 @@
+### show-type-counts Behavior changed in Search API
+
+In the Search API if you set show-type-counts=true the response will include all object types (Dataverses, Datasets, and Files) even if the search result for any given type is 0.
```
pdurbin, line 3:
```suggestion
In the Search API if you set show_type_counts=true the response will include all object types (Dataverses, Datasets, and Files) even if the search result for any given type is 0.

See also the [guides](https://preview.guides.gdcc.io/en/develop/api/search.html#parameters), #11127 and #11138.
```

src/main/java/edu/harvard/iq/dataverse/api/Search.java
```diff
@@ -34,6 +34,7 @@
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Response;
+import java.math.BigDecimal;
```
pdurbin, line 37:
```suggestion
```

pdurbin (commented):
Looks good. Just a couple very minor things.

sekmiller:
Thanks for the suggestions, Phil.

pdurbin (approved):
I didn't test this but looks good.

ofahimIQSS:
Ran tests in internal, no issues found - merging PR 
image
image

PR #11144, opened 2025-01-09
Update Dataverse API removes metadatablocks if optional params are omitted

### LINKED ISSUE #11130: Update Dataverse API removes metadatablocks if optional params are omitted
According to https://guides.dataverse.org/en/latest/api/native-api.html#update-a-dataverse-collection, 
`Prepare a JSON file containing the fields for the properties you want to update. You do not need to include all the properties, only the ones you want to update.` However, the changes in #11026 added consequences for not adding some optional fields: 
```
When it comes to omitting these fields in the JSON:

Omitting facetIds or metadataBlockNames causes the Dataverse collection to inherit the corresponding configuration from its parent.

Omitting inputLevels removes any existing custom input levels in the Dataverse collection.

Omitting the entire metadataBlocks object in the request JSON would exclude the three sub-objects, resulting in the application of the two changes described above.

``` 

In combination, this means that trying to use the API to edit the dataverseType, for example, by just sending the properties to update, as of 6.5, now removes the metadatablocks associated with the dataverse/collection and sets it to not be the metadata root.

For child collections, this presumably causes them to start inheriting from the root. For the root collection, the case where this bug was seen, it causes Internal Server Errors when creating datasets, and puts the root collection in a state where its metadatablocks can't be edited in the UI.

Use the API above to make a change to dataverseType, see that the subcollection now inherits metadatablocks/the root collection is broken if you changed it, see that creating a dataset causes an error.

6.5/develop

For those with access, see discussion at https://iqss.slack.com/archives/C03R1E7T4KA/p1735848411767949

**Screenshots:**

*

stevenwinship:
**What this PR does / why we need it**: Better control over updating dataverse metadatablocks - update/delete-inherit

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11130

- Closes #11130

**Special notes for your reviewer**: See discussion in parent issue

**Suggestions on how to test this**: See DataversesIT. Test different combinations of Json with the three blocks and two flags.

**Additional documentation**: native-api

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractWriteDataverseCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractWriteDataverseCommand.java
@@ -19,15 +19,13 @@ abstract class AbstractWriteDataverseCommand extends AbstractCommand<Dataverse> 
     private final List<DataverseFieldTypeInputLevel> inputLevels;
     private final List<DatasetFieldType> facets;
     protected final List<MetadataBlock> metadataBlocks;
-    private final boolean resetRelationsOnNullValues;
 
     public AbstractWriteDataverseCommand(Dataverse dataverse,
                                          Dataverse affectedDataverse,
                                          DataverseRequest request,
                                          List<DatasetFieldType> facets,
                                          List<DataverseFieldTypeInputLevel> inputLevels,
-                                         List<MetadataBlock> metadataBlocks,
-                                         boolean resetRelationsOnNullValues) {
+                                         List<MetadataBlock> metadataBlocks) {
         super(request, affectedDataverse);
         this.dataverse = dataverse;
         if (facets != null) {
@@ -45,7 +43,6 @@ abstract class AbstractWriteDataverseCommand extends AbstractCommand<Dataverse> 
         } else {
             this.metadataBlocks = null;
         }
-        this.resetRelationsOnNullValues = resetRelationsOnNullValues;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractWriteDataverseCommand.java
```diff
@@ -59,46 +56,61 @@ public Dataverse execute(CommandContext ctxt) throws CommandException {
...
+    */
     private void processInputLevels(CommandContext ctxt) {
         if (inputLevels != null) {
-            if (!inputLevels.isEmpty()) {
+            if (inputLevels.isEmpty()) {
+                ctxt.fieldTypeInputLevels().deleteFacetsFor(dataverse);
+            } else {
                 dataverse.addInputLevelsMetadataBlocksIfNotPresent(inputLevels);
+                ctxt.fieldTypeInputLevels().deleteFacetsFor(dataverse);
+                inputLevels.forEach(inputLevel -> {
```
sekmiller, line 109: Do we have to worry about incompatible input levels where they might put in a dataset field type that doesn't belong to the installed metadata blocks of the given dataverse"

stevenwinship: I'm not sure I understand the question. The new code does exactly what the old code did for inputLevel list that isn't empty. My goal was to remove the resetRelationsOnNullValues flag and implement a way to delete data or ignore the missing data based on the list being empty or null.

sekmiller (approved):
This looks good. thanks for adding the tests.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractWriteDataverseCommand.java
sekmiller: Sorry, what I was wondering about is what if you add an input level for a dataset field that isn't already in a metadata block that is part of the dataverse you're updating - the answer is in line 107 - if you add an input level for a dataset field that is not already included in the dataverse then  dataverse.addInputLevelsMetadataBlocksIfNotPresent(inputLevels) - will add that metadata block to the dataverse.

PR #11150, opened 2025-01-13
Allow collections to be moved in the superuser dashboard

### LINKED ISSUE #10304: Add "Move Dataverse collection" to the Move Data part of the Dashboard
An [API endpoint](https://guides.dataverse.org/en/5.14/admin/dataverses-datasets.html#move-a-dataverse-collection) exist for the often asked task of moving datasets, however, contrary to "move dataset" it is absent of the UI Dashboard.

N.B. an issue will be created and linked for SPA, however as this is a great quality of life improvement and only a small addition as the API already exists, I think it could be added to the current UI too.

Superuser

Frequent need for non tech admins to move collections

Add a "Move Collection" button to the dashboard :
image
Open to discussion whether we should rather or not have move dataset and dataverse in the same move data page and not two buttons in the dashboard, but I think user would be more error prone if they have both in the same page.

Leading to a similar page than the current Move Data page (see under) but with labels for Dataverse colections :
image
Modified labels for discussion & copy-pasting purposes : 
- (info message) Move Dataverse collection – Manage and curate your installation by moving a dataverse collection from one host dataverse collection to another. See also Managing Datasets and Dataverses in the Admin Guide.
- (left section title) Dataverse to Move
- (left section object label) Dataverse
- (left section watermark) Enter Identifier of the Dataverse to move

SPA issue :
- https://github.com/IQSS/dataverse-frontend/issues/309

**Will you be able to contribute a Pull Request for this issue?**
Yes, we plan to contribute the code for this and will try to have it ready for review/integration in 6.5

stevenferey:

- Closes #10304 

**Additional documentation**:


src/main/java/propertyFiles/Bundle.properties
```diff
@@ -542,6 +542,9 @@ dashboard.card.harvestingserver.sets={0, choice, 0#Sets|1#Set|2#Sets}
 dashboard.card.harvestingserver.btn.manage=Manage Server
 dashboard.card.metadataexport.header=Metadata Export
 dashboard.card.metadataexport.message=Dataset metadata export is only available through the {0} API. Learn more in the {0} {1}API Guide{2}.
+dashboard.card.move.data=Data
+dashboard.card.move.dataset.manage=Move Dataset
+dashboard.card.move.dataverse.manage=Move Collection
```
sekmiller, line 547: Can we say "Move Dataverse" or "Move Dataverse Collection" here?

stevenferey: yes, it may be better to put "Move Dataverse" in order to keep a consistency of display ?

Capture d’écran du 2025-01-10 11-28-48

sekmiller:
@stevenferey can you make that bundle change for "Move Dataverse"? Also, the "force move" via the UI, might be a nice to have, but not necessary since it's available via the api.

ofahimIQSS:
Testing Passed - merging pr

image
image
image

Issue #11129, opened 2025-01-02
 Feature Request: Contact(sending feedback to contacts) API endpoint 


ChengShi-1:
**Overview of the Feature Request**
In SPA, users are allowed to send feedback to contacts of a collection, dataset or file.
Input would include **subject**, **message**, **from** (email autocomplete if login, or filled out by guest users).

In Native API, it has a similar API endpoint sending feedback to contact, but it is under admin path, and it returns contacts' email that could be sensitive. [Send Feedback To Contact(s)](https://guides.dataverse.org/en/6.4/api/native-api.html#id271)

**What kind of user is the feature intended for?**
API User and all SPA users

**What inspired the request?**
SPA Q4.9
https://github.com/IQSS/dataverse-client-javascript/pull/232

stevenwinship:
@ChengShi-1  @pdurbin @qqmyers 
Please comment on the following warning about email spam from the current FeedbackApi:

/**
     * This method mimics the contact form and sends an email to the contacts of the
     * specified Collection/Dataset/DataFile, optionally ccing the support email
     * address, or to the support email address when there is no target object.
     *
     * !!!!! This should not be moved outside the /admin path unless/until some form
     * of captcha or other spam-prevention mechanism is added. As is, it allows an
     * unauthenticated user (with access to the /admin api path) to send email from
     * anyone to any contacts in Dataverse. (It also does not do much to validate
     * user input (e.g. to strip potentially malicious html, etc.)!!!!
     **/

pdurbin:
The comment was added by @qqmyers in #9186 and looks accurate to me.

I added the submitFeedback method itself in #4571 along with API tests that call into it. The method was only meant for testing, for making assertions about the behavior of the app.

pdurbin:
Closed by merging this PR:

- #11162

PR #11162, opened 2025-01-16
add new feedback api

stevenwinship:
**What this PR does / why we need it**: In SPA, users are allowed to send feedback to contacts of a collection, dataset or file.
Input would include subject, message, from (email autocomplete if login, or filled out by guest users).

In Native API, it has a similar API endpoint sending feedback to contact, but it is under admin path, and it returns contacts' email that could be sensitive. [Send Feedback To Contact(s)](https://guides.dataverse.org/en/6.4/api/native-api.html#id271)

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/11129

- Closes # 11129

**Additional documentation**: config and native-api additions


src/main/java/edu/harvard/iq/dataverse/api/SendFeedbackAPI.java
```diff
@@ -0,0 +1,110 @@
...
+     * !!!!! This should not be moved outside the /admin path unless/until some form
+     * of captcha or other spam-prevention mechanism is added. As is, it allows an
+     * unauthenticated user (with access to the /admin api path) to send email from
+     * anyone to any contacts in Dataverse. It also does not do much to validate
+     * user input (e.g. to strip potentially malicious html, etc.)!!!!
+     **/
+    @POST
+    @AuthRequired
+    @Consumes("application/json")
+    public Response submitFeedback(@Context ContainerRequestContext crc, JsonObject jsonObject) {
```
sekmiller, line 48: I think it might be helpful to get the json as a string and then parse it within the try so that you can give the user better feedback. I tried it with invalid json and I get a 500 error and an empty response.

src/main/java/edu/harvard/iq/dataverse/api/SendFeedbackAPI.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            JsonNumber jsonNumber = jsonObject.getJsonNumber("targetId");
+            DvObject feedbackTarget = null;
+            if (jsonNumber != null) {
+                feedbackTarget =  dvObjSvc.findDvObject(jsonNumber.longValue());
```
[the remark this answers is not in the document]
stevenwinship: Probably not hard but it would require a documentation change that makes the api under /admin different from this new one. Would it be more confusing to have the json files differ between the 2 APIs?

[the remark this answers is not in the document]
stevenwinship: I'll make the change

src/main/java/edu/harvard/iq/dataverse/api/SendFeedbackAPI.java
stevenwinship: good idea

stevenwinship: I made the change to pass in a string and validate it

src/main/java/edu/harvard/iq/dataverse/api/SendFeedbackAPI.java
```diff
@@ -0,0 +1,125 @@
...
+                if (feedbackTarget == null) {
+                    feedbackTarget = dvObjSvc.findByGlobalId(idtf, DvObject.DType.Dataset);
+                }
+                if (feedbackTarget == null) {
+                    feedbackTarget = dvObjSvc.findByGlobalId(idtf, DvObject.DType.DataFile);
+                }
+            }
+
+            // feedbackTarget and idtf are both null this is a support feedback and is ok
+            if (feedbackTarget == null && idtf != null) {
```
sekmiller, line 68: I think we need to say && (idtf != null || jsonNumber != null) because now if the jsonNumber doesn't return a dvObject it will go to general support, which we probably don't want.

sekmiller (changes requested):
Just one more small logic change (when there's no feedback target) and I think we're good to go.

src/main/java/edu/harvard/iq/dataverse/api/SendFeedbackAPI.java
stevenwinship: Good catch. the idtf used to have a string version of the jsonNumber but now it doesn't

// idtf will hold the "targetId" or the "identifier". If neither is set then this is a general feedback to support
String idtf = jsonNumber != null ? jsonNumber.toString() : jsonObject.containsKey("identifier") ? jsonObject.getString("identifier") : null;

sekmiller (approved):
OK. Looks good. Thanks for the updates.

ofahimIQSS:
This PR Has conflicts - please resolve

ofahimIQSS:
Testing Passed -merging PR

PR #11170, opened 2025-01-20
Rezip shape files with upper case extension

### LINKED ISSUE #11169: Shape files with upper upper case extensions are not re-zipped

Upload a zip with a set of shape files with one or more upper case extensions.

* When does this issue occur?

* What happens?

Only the shape files with lower case extensions are re-zipped. If a mandatory zip file had an upper case extensions, none are zipped

all

Shape files with upper case extension should be in the re-zipped file.

**Screenshots:**

*

Yes

jo-pol:

- Closes #11169

Execute the changed test without the changing the ShapeFileHandler. The test will fail. With the changed ShapeFileHandler  the test succeeds again.
See also the issue.

No

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/ShapefileHandler.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/ShapefileHandler.java
@@ -8,6 +8,7 @@ import java.nio.file.StandardCopyOption;
 import java.util.Date;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import java.util.zip.ZipFile;
 import java.util.HashMap;
 import java.util.*;
@@ -561,7 +562,7 @@ public class ShapefileHandler{
         if (ext_name == null){
             return false;
         }
-        return SHAPEFILE_ALL_EXTENSIONS.contains(ext_name);
+        return SHAPEFILE_ALL_EXTENSIONS.contains(ext_name.toLowerCase());
     }
     /*
         Does a list of file extensions match those required for a shapefile set?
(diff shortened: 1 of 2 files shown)


pdurbin (approved):
I didn't run the code but it makes sense and [tests are passing](https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11170/1/testReport/).

There's no release note snippet but I think that's ok. It's a fairly minor fix. Approved.

qqmyers:
I'm not completely sure, but it looks like there are 3 shape files with .SHP (capitals) extensions in your zip. If this PR is working, I think that it would cause the 3 sets of 5 files each for those *.SHP files to be recognized as compound shape files and rezipped as 3 zip files - so a 15-12 = 12 file difference between the two. You can confirm further by looking to see if there are 3 zip files: gis_osm_places_a_07_1.zip, gis_osm_pofw_07_1.zip, and gis_osm_pofw_a_07_1.zip after the PR whereas in your list of 41, you'll find individual gis_osm_pofw_a_07_1.* files instead of zips.

ofahimIQSS:
> I'm not completely sure, but it looks like there are 3 shape files with .SHP (capitals) extensions in your zip. 

Correct, I edited 3 of the lowercase .shp to .SHP

>If this PR is working, I think that it would cause the 3 sets of 5 files each for those _.SHP files to be recognized as compound shape files and rezipped as 3 zip files - so a 15-12 = 12 file difference between the two. You can confirm further by looking to see if there are 3 zip files: gis_osm_places_a_07_1.zip, gis_osm_pofw_07_1.zip, and gis_osm_pofw_a_07_1.zip after the PR whereas in your list of 41, you'll find individual gis_osm_pofw_a_07_1._ files instead of zips.

The math makes sense now - in that case it looks like the PR is working as expected. 

Merging PR

PR #11192, opened 2025-01-28
Dataset Migration API documentation improvements

mjlassila:
This PR is related to topic [Metadata format for Dataset Migration API in 6.5](https://dataverse.zulipchat.com/#narrow/channel/378866-troubleshooting/topic/Metadata.20format.20for.20Dataset.20Migration.20API.20in.206.2E5) in the chat.

I added a mention to the docs that the OAI-ORE format used in the export is not directly compatible with the Dataset Migration API and made the example file more comprehensive by including few additional fields from Geospatial  and Social Science and Humanities Metadata blocks.

Preview at https://dataverse-guide--11192.org.readthedocs.build/en/11192/developers/dataset-migration-api.html


qqmyers:
FWIW: The geospatial and social blocks are given local context URIs because those bocks do not have a blockURI defined. See the docs or the citation block for how that can be done - https://github.com/IQSS/dataverse/blob/7283dbd982f05a757c91ad03452c69e1f0f6f4d2/scripts/api/data/metadatablocks/citation.tsv#L1-L2.

Without that, a block is considered to be locally defined with terms different (w.r.t. URI) from those at other instances. That's possibly useful for experimental blocks, or modified ones, but leads to the problem with the migration API (and I assume semantic  API as well) unless the receiving Dataverse instance has blocks that set the blockURI to be the same as the auto-generated one in the source Dataverse.

doc/sphinx-guides/source/_static/api/dataset-migrate.jsonld
```diff
@@ -35,5 +55,9 @@
   "title": "http://purl.org/dc/terms/title",
   "citation": "https://dataverse.org/schema/citation/",
   "dvcore": "https://dataverse.org/schema/core#",
-  "schema": "http://schema.org/"
-}}
+  "schema": "http://schema.org/",
+  "geospatial": "dataverse.siteUrl/schema/geospatial#",
+  "socialscience": "dataverse.siteUrl/schema/socialscience#",
```
pdurbin, line 60:
```suggestion
  "geospatial": "https://dataverse.org/schema/geospatial#",
  "socialscience": "https://dataverse.org/schema/socialscience#",  
```

This comment is mostly for @qqmyers who left a note at https://github.com/IQSS/dataverse/pull/11192#issuecomment-2618762630 but really for anyone who's following this conversation.

Instead of dataverse.siteUrl should we use https://dataverse.org for the geospatial and socialscience blocks? https://dataverse.org is the blockURI used by the citation block. This would allow us to remove the step for replacing the dataverse.siteUrl described further down on the page.

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -5,7 +5,7 @@ The Dataverse software includes several ways to add Datasets originally created
 
 This experimental migration API offers an additional option with some potential advantages:
 
-* metadata can be specified using the json-ld format used in the OAI-ORE metadata export
+* metadata can be specified using the json-ld format used in the OAI-ORE metadata export (please note that the json-ld generated by OAI-ORE metadata export is not directly compatible with the Migration API, check example file below for reference)
```
pdurbin, line 8: Can we say a little more about how the generated OAI-ORE format is not compatible? I'm a bit confused and not sure what I should be taking away from a look at the example file. 🤔

doc/sphinx-guides/source/developers/dataset-migration-api.rst
```diff
@@ -31,7 +31,7 @@ To import a dataset with an existing persistent identifier (PID), the provided j
   
   curl -H X-Dataverse-key:$API_TOKEN -X POST $SERVER_URL/api/dataverses/$DATAVERSE_ID/datasets/:startmigration --upload-file dataset-migrate.jsonld
 
-An example jsonld file is available at :download:`dataset-migrate.jsonld <../_static/api/dataset-migrate.jsonld>` . Note that you would need to replace the PID in the sample file with one supported in your Dataverse instance.
+An example jsonld file is available at :download:`dataset-migrate.jsonld <../_static/api/dataset-migrate.jsonld>` . Note that you would need to replace the PID in the sample file with one supported in your Dataverse instance. You also need to replace the dataverse.siteUrl in jsonld @context with your current Dataverse site URL.
```
pdurbin, line 35: I'm just putting this comment at the bottom.

@mjlassila from our [conversation](https://dataverse.zulipchat.com/#narrow/channel/378866-troubleshooting/topic/Metadata.20format.20for.20Dataset.20Migration.20API.20in.206.2E5/near/494778081) I was sort of hoping you'd include your BaseX and XQuery script at https://gist.github.com/mjlassila/ecdbd11447ccdf87995db20bfc5e686c 😄 . It it worth writing up and including? (I don't even know how to run it but I'm happy to learn!)

pdurbin (commented):
Some initial thoughts. Thanks for the PR! ❤️

doc/sphinx-guides/source/_static/api/dataset-migrate.jsonld
qqmyers: Because the social/geo blocks don't define a block URI (for which https://dataverse.org/schema/socialscience and https://dataverse.org/schema/geospatial are good/standard choices), the ORE exporter uses the local URL, e.g. https://pas.rd.tuni.fi/schema/geospatial. If I wanted to import that json-ld into QDR, I'd have to change that URI to https://data.qdr.syr.edu/schema/geospatial. (dataverse.siteUrl is just a variable in the lines above that has to be replaced.) If the blocks are updated, the ORE will use https://dataverse.org/schema/geospatial and there wouldn't be any need to translate it to migrate into a new Dataverse instance. Given that such changes would only be for v6.6+, the docs probably should still change to explain that URIs from <= v6.5 sites need to be edited for these blocks/any blocks without a blockURI.

pdurbin: Oh, I understand better now what you meant about locally defined. The URLs have to match.

In that case, my suggestion doesn't make sense. We should stick with dataverse.siteURL like @mjlassila originally wrote. I'll resolve this.

mjlassila:
> FWIW: The geospatial and social blocks are given local context URIs because those blocks do not have a blockURI defined. See the docs or the citation block for how that can be done [...]
> Without that, a block is considered to be locally defined with terms different (w.r.t. URI) from those at other instances. 

@qqmyers : Thanks for clarification! If out of the box geospatial and social blocks do not have blockURI defined, should I keep the example file as it is or modify it somehow? If I leave these out from the example file @context, citation metadata gets imported via Dataset Migration API but geospatial and social blocks will drop out silently.

qqmyers:
@mjlassila - that's hard to answer clearly. I just created #11195 as an issue to get those blocks changed. Once a change is merged there, we probably want the docs to talk about the current situation and probably still note that migrating from earlier versions has the problem you're documenting. To not hold up your PR, I'd suggest that you just go ahead and document how it is now (add additional explanation if you want) and we'll treat updating the docs as part of completing #11195.

mjlassila:
> To not hold up your PR, I'd suggest that you just go ahead and document how it is now (add additional explanation if you want) and we'll treat updating the docs as part of completing #11195.

@qqmyers  - I added a short explanation of why defining a local URI for community metadata blocks is currently necessary.

pdurbin (approved):
This is a nice improvement to the docs. Approved!

I did go ahead and try the updated JSON file and it created a dataset for me.

I did have to make this change...

```
< "@id": "doi:10.33564/FK27U7YBV",
---
> "@id": "doi:10.5072/FK2/FK27U7YBV",
```

... otherwise I get "Bad Request. The API request cannot be completed with the parameters supplied. Details: Cannot recreate a dataset that has a PID that doesn't match the server's settings".

This is pre-existing. I'm just noting it for record. 😄 

Here's a screenshot of the dataset:

Screenshot 2025-01-30 at 1 47 53 PM

ofahimIQSS:
docs/readthedocs.org:dataverse-guide is failing on this PR

ofahimIQSS:
Proofread - looks good. Merging PR

PR #11203, opened 2025-01-31
Request Access fixes

qqmyers:
**What this PR does / why we need it**: As reported in [dataverse.harvard.edu issue 247](https://github.com/IQSS/dataverse.harvard.edu/issues/247#issuecomment-2627541100), when terms of use/terms of access cause the guestbook-terms-popup-fragment to appear and when guestbook at request is on, the dialog shows the basic guestbook questions when it shouldn't.

In investigating this, I also realized that the terms of access, if they exist, aren't show when a request for access was made from the per-file menus in the files table (versus selecting and using the button at the top).

This PR fixes both of these issues and makes text changes related to [another comment in dataverse.harvard.edu issue 247](https://github.com/IQSS/dataverse.harvard.edu/issues/247#issuecomment-2627507682) that try to clarify that the choice in the dataverse edit UI to select guestbooks at request or download is only a choice about the guestbook mode and only affects datasets where a guestbook is configured.

- Closes #

**Suggestions on how to test this**: The bug fixes in this PR are only intended to affect the following two cases:
* guestbooks at request, no guestbook, terms of use and/or terms of access requiring terms to show: the popup should not ask for name, email, etc. (see links to images in issue 247 above).
* terms of access exist, and request for access or download made from the per-file dropdown menus in the file table: terms of access should be shown /the dialog should be the same as if you select the same file and use the request access/download button at the top of the table.
Given the complexity of the code, its possible that the changes here affect other situations so regression testing to verify that the UI is unchanged when a guestbook exists, when terms of use/access aren't set, guestbook at download is configured, etc. should be done.

For the text changes, just verify that the dataverse edit/general info tab has an entry now called "Guestbook Mode" instead of "Guestbook Entry Option" and the popup help includes the phrase "(when configured for a dataset)".

**Additional documentation**:


sekmiller (approved):

ofahimIQSS:
Merging Pr - no issues found during testing

PR #11205, opened 2025-02-03
clarify which PID settings are required

pdurbin:
**What this PR does / why we need it**:

People are confused. The way it's worded, it sounds like `dataverse.spi.pidproviders.directory` is required. Please see https://dataverse.zulipchat.com/#narrow/channel/378866-troubleshooting/topic/DOI.20config.20change.2C.20server.20down/near/497448462

PR #11209, opened 2025-02-03
Fix Flaky DatasetsIT test

qqmyers:
**What this PR does / why we need it**: seen in #11205 - The call to edit metadata failed with a 403 due to the finalize publication lock not having been released. This PR adds a sleep for lock to avoid that.

- Closes #

**Suggestions on how to test this**: Just merge? Should avoid future failures from this part of the code, but hard to test otherwise.

**Additional documentation**:


ofahimIQSS:
merging and observing

PR #11128, opened 2025-01-02
Upgrade to Payara 6 2025.2

qqmyers:
**What this PR does / why we need it**: Updates to Payara 6.2025.2, fixing a few places where JSF now appears to send a null rather than an empty string, causing some new NPEs.

**Which issue(s) this PR closes**:

- Closes #11126

**Special notes for your reviewer**: FWIW: It looks like just copying the domain across to the new version works so we may only want to show 'Option 1' form the release notes. I originally did 'Option2' - just copying the parts of domain.xml and other files to the new domain which has some advantages (easier to diff against the latest payara domain.xml, gets new cacerts file, makes it possible to track payara domain.xml changes in github (against /conf/payara/domain.xml), etc.) but it's definitely more work/more confusing.

With further testing at QDR, I've noticed a steady/overwhelming stream of errors of the form  `JSF1103: The metadata facet must be a direct child of the view in viewId /dataverse.xhtml]]`.  I did not notice any change in UI behavior though. Digging in a bit, it appears that any page that uses `<ui:composition template="/dataverse_template.xhtml">` was including the f:metadata farther down in the tree. I've made changes to move them to being a direct child of the view which appears to get rid of the log messages, again w/o any impact on functionality that I've seen. I haven't been able to confirm, but my guess is that while the error is now getting flagged, the f:metadata was still be interpreted correctly.

**Suggestions on how to test this**: Check new installs, docker, upgrade instructions - basically regression testing and testing instructions, there shouldn't be any functional changes. Could confirm that the current v6.5 war does not work with Payara 6.2025.2, but I think that's well enough known. 

Testing should also involve grepping for `JSF1103` in the logs to assure that there are no more instances (e.g. .xhtml files I missed.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**: included

**Additional documentation**: references to 6.2024.6 changed to 6.2025.2 in docs

https://dataverse-guide--11128.org.readthedocs.build/en/11128/developers/tips.html#jsf1103-errors

PR #11210, opened 2025-02-03
allow SameSite cookie attribute to be configured

pdurbin:

- Closes https://github.com/IQSS/dataverse-security/issues/27

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/container/base-image.rst
+++ b/doc/sphinx-guides/source/container/base-image.rst
@@ -281,6 +281,18 @@ provides. These are mostly based on environment variables (very common with cont
       - See :ref:`:ApplicationServerSettings` ``http.request-timeout-seconds``.
 
         *Note:* can also be set using any other `MicroProfile Config Sources`_ available via ``dataverse.http.timeout``.
+    * - ``DATAVERSE_COOKIE_SAME_SITE_VALUE``
+      - ``Lax``
+      - String
+      - Can be set to ``Strict``, ``Lax``, or ``None``. See :ref:`samesite-cookie-attribute`, :ref:`:ApplicationServerSettings` :ref:`http.cookie-same-site-value` and `Payara docs <https://docs.payara.fish/community/docs/6.2024.6/Technical%20Documentation/Payara%20Server%20Documentation/General%20Administration/Administering%20HTTP%20Connectivity.html>`_.
+
+        *Note:* can also be set using any other `MicroProfile Config Sources`_ available via ``dataverse.cookie-same-site-value``.
+    * - ``DATAVERSE_COOKIE_SAME_SITE_ENABLED``
+      - ``1``
+      - Bool, ``0|1`` or ``false|true``
+      - See :ref:`samesite-cookie-attribute` and :ref:`:ApplicationServerSettings` :ref:`http.cookie-same-site-enabled`.
+
+        *Note:* can also be set using any other `MicroProfile Config Sources`_ available via ``dataverse.cookie-same-site-value``.
     * - ``PAYARA_ADMIN_PASSWORD``
       - ``admin``
       - String
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -151,6 +151,15 @@ Password complexity rules for "builtin" accounts can be adjusted with a variety 
 - :ref:`:PVGoodStrength`
 - :ref:`:PVCustomPasswordResetAlertMessage`
 
+.. _samesite-cookie-attribute:
+
+SameSite Cookie Attribute
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The SameSite cookie attribute is defined in upcoming revision to `RFC 6265 <https://datatracker.ietf.org/doc/html/rfc6265>`_ (HTTP State Management Mechanism) called `6265bis <https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19>`_ ("bis" meaning "repeated"). The possible values are "None", "Lax", and "Strict". "Strict" is intended to help prevent Cross-Site Request Forgery (CSRF) attacks, as described in the RFC proposal and an OWASP `cheetsheet <https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#samesite-cookie-attribute>`_. "Lax" is a middle ground between "Strict" and "None".
+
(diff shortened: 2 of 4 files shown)


doc/release-notes/ds27-samesite-attr.md
```diff
@@ -0,0 +1,18 @@
...
+
+New Dataverse installations are now set to "Lax" out of the box by the installer (in the case of a "classic" installation) or through an updated base image (in the case of a Docker installation). See upgrade instructions below.
+
+
+See also [the guides](https://dataverse-guide--11210.org.readthedocs.build/en/11210/installation/config.html#samesite-cookie-attribute), https://github.com/IQSS/dataverse-security/issues/27 and #11210.
+
+## Upgrade instructions
+
+To bring your Dataverse installation in line with the new "Lax" (as opposed to "None") value described in [the guides](https://dataverse-guide--11210.org.readthedocs.build/en/11210/installation/config.html#samesite-cookie-attribute), we recommend running the following commands:
```
qqmyers, line 12: This sounds like doing nothing means an installation would be using SameSite:None, but I think by default current installation are not sending anything, which is interpreted as 'Lax' by modern browsers. I.e. installations aren't less safe if they do nothing. Maybe 'To explicitly specify the Lax setting or to switch to Strict, you can use the following commands...' ?  Perhaps we should also say 'The value 'None' should never be used with Dataverse.' ?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3385,6 +3394,32 @@ To facilitate large file upload and download, the Dataverse Software installer b
...
 and restart Payara to apply your change.
 
+.. _http.cookie-same-site-value:
+
+http.cookie-same-site-value
+++++++++++++++++++++++++++++
+
+See :ref:`samesite-cookie-attribute` for context.
+
+The Dataverse installer configures the Payara **server-config.network-config.protocols.protocol.http-listener-1.http.cookie-same-site-value** setting to "Lax". From `Payara's documentation <https://docs.payara.fish/community/docs/6.2024.6/Technical%20Documentation/Payara%20Server%20Documentation/General%20Administration/Administering%20HTTP%20Connectivity.html>`_, the other possible values are "Strict" or "None". To change this to "Strict", for example, you could run the following command...
```
qqmyers, line 3404: The value for this appears to do nothing if the http.cookie-same-site-enabled isn't set to true.

Conversely, setting http.cookie-same-site-enabled=true without setting this one ends up sending None as the default which is worse than not having a SameSite setting at all. I think the docs should warn people about these issues.

qqmyers (changes requested):
The bin/asadmin work fine on an existing installation. I did suggest some doc changes to clarify that installations are not less secure now that our recommendation of going to Lax, and to stress that using None on purpose, or accidentally increases risks.

doc/release-notes/ds27-samesite-attr.md
qqmyers: FWIW: Not sure how you got a None for the Jessionid in the picture - I see no setting at all at QDR and on a test server. Could you have had the http.cookie-same-site-enabled=true already?

qqmyers:
Another note - I think Strict breaks remote login - at least ORCID which is what I tested. Lax works as expected.

doc/release-notes/ds27-samesite-attr.md
pdurbin: As discussed [in Slack](https://iqss.slack.com/archives/C010LA04BCG/p1739205363451439), Firefox was leading me astray. I'm now testing with curl and I 100% agree with @qqmyers that SameSite is absent by default.

I'm updated the release notes and docs to reflect this new understanding. I also recommend against using "None".

@qqmyers please take another look. Thanks!

doc/sphinx-guides/source/container/base-image.rst
```diff
@@ -281,6 +281,18 @@ provides. These are mostly based on environment variables (very common with cont
       - See :ref:`:ApplicationServerSettings` ``http.request-timeout-seconds``.
 
         *Note:* can also be set using any other `MicroProfile Config Sources`_ available via ``dataverse.http.timeout``.
+    * - ``DATAVERSE_COOKIE_SAME_SITE_VALUE``
```
pdurbin, line 284: When I try to change DATAVERSE_COOKIE_SAME_SITE_VALUE to "Strict" in the compose file, it doesn't seem to have an effect. I still see "Lax" when I check the cookie with curl.

I opened a thread in Zulip about this: https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/DATAVERSE_HTTP_TIMEOUT.20not.20having.20an.20effect/near/499094581

I'm not sure why it isn't working. 🤷 I'm hoping @poikilotherm will save me! ❤️

qqmyers: Did you also set DATAVERSE_COOKIE_SAME_SITE_ENABLED?

qqmyers (approved):
Looks good. Odd that  SameSite is repeated in curl. FWIW: if I look at the first call to Dataverse in the Chrome or Edge console, the Set-Cookie entry in the response header only shows one copy (even selecting 'raw' for the response headers). Perhaps they are de-duplicating even for raw?

doc/sphinx-guides/source/container/base-image.rst
pdurbin: Well, it's already set.

pdurbin: Through the Dockerfile, I mean.

pdurbin: Hold on, I need to add something to the Dockerfile.

pdurbin: Bah. It doesn't work. As discussed at standup, I switched Docker to be hard coded to Lax and updated the docs to say you have to rebuild the base image if you want Strict or whatever. See d62ac2f8df

ofahimIQSS:
Hello! Can you please update this PR with the latest from Dev. Thanks!

pdurbin:
Sure, done!

ofahimIQSS:
Adding Testing Evidences:

image
image

ofahimIQSS:
> Looks good. Odd that SameSite is repeated in curl. FWIW: if I look at the first call to Dataverse in the Chrome or Edge console, the Set-Cookie entry in the response header only shows one copy (even selecting 'raw' for the response headers). Perhaps they are de-duplicating even for raw?

What Jim mentioned, any way we can fix that or shall we proceed as is. @pdurbin

ofahimIQSS:
merging PR

ofahimIQSS:
Integration Tests are failing - please advise: 
image

pdurbin:
@ofahimIQSS I merged the latest from develop to pick up this fix:

- #11319

Now all tests are passing.

ofahimIQSS:
Merging PR

pdurbin:
> Note: we suspect that the repeated SameSite attribute is a bug in Payara. We should retest after we've upgraded Payara in https://github.com/IQSS/dataverse/pull/11128 before reporting it upstream.

I just reported it upstream:

- https://github.com/payara/Payara/issues/7254

PR #11222, opened 2025-02-06
ORCID: Support authenticated ORCIDs in account profile

qqmyers:
**What this PR does / why we need it**: This PR adds one area of functionality supported through a grant to GDCC from the [ORCID Global Participation Fund](https://info.orcid.org/global-participation-fund-announces-fourth-round-of-awardees/).

Specifically, it allows Dataverse users to add a link to their ORCID profile which then allows Dataverse to use automatically add their ORCID to their author metadata when they create a dataset.

This functionality leverages Dataverse's existing support for login via ORCID, but can be turned on independently of it. If ORCID login is enabled, the user's ORCID will automatically be added to their profile. If the user has logged in via some other mechanism, they are able to click a button to initiate a similar authentication process in which the user must login to their ORCID account and approve the connection. 

This PR should be functional as is, but it is possible/likely that further changes could be made based on user feedback.

**Which issue(s) this PR closes**:

- Closes #7284

**Special notes for your reviewer**: 

**Suggestions on how to test this**: Use/create an ORCID OAuth2 provider as was previously possible and verify that logging in with that provider adds the ORCID to the user account (Account Info page), that when logging in with a builtin/other account, you can click the 'Add Authenticated ORCID' button to add one, and use the Edit Account/Remove ORCID menu item to remove it, that while the ORCID is in the account from whatever route, that creating a dataset results in the user with their ORCID being added as an author of the dataset.

To verify that this new functionality can be configured w/o allowing ORCID login, reinstall the ORCID provider with its "enabled" flag set false (or toggle the enabled field in the db, e.g. `update authenticationproviderrow set enabled='f' where id='orcid';` for a standard config.)

Also verify that if no ORCID provider exists, the ability to add an ORCID is not displayed.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
Account Info showing option to add ORCID
image
ORCID login
image
ORCID Approval screen
image
ORCID in Account Info pane showing remove orcid menu item

image
New Dataset showing ORCID of author used when auto-filling.
image

**Is there a release notes update needed for this change?**: included

**Additional documentation**:

Preview docs at https://dataverse-guide--11222.org.readthedocs.build/en/11222/installation/orcid.html

PR #11225, opened 2025-02-06
Flaky XmlMetadataTemplateTest fix

qqmyers:
**What this PR does / why we need it**: Some obsolete code in the XmlMetadataTemplate class that tried to find the PidProvider for a 10.5072/FK2* identifier was causing intermittent failures - depending on whether other tests had added/remove a DOI provider for that authority/shoulder. This was seen in #10567 and #11222 recently. This PR removes the obsolete code which is doing the PidProvider lookup.

- Closes #

**Special notes for your reviewer**: FWIW: This code's from #10632 where the pidprovider was originally used as a way to call a method in the DatasetFieldServiceBean (since XMLMetadataTemplate isn't a bean). That was replaced by 
https://github.com/IQSS/dataverse/blob/3aea1482dcc2f97741ba79b2f4b93b069b425871/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/XmlMetadataTemplate.java#L651 , but it appears the initial lookup of the PID provider wasn't removed.

**Suggestions on how to test this**: Basic regression testing w.r.t. creating/publishing a DataCite DOI dataset? Nominally there should be no effect except for no future test fails.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/XmlMetadataTemplate.java
+++ b/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/XmlMetadataTemplate.java
@@ -69,8 +69,6 @@ public class XmlMetadataTemplate {
     public static final String XML_SCHEMA_VERSION = "4.5";
 
     private DoiMetadata doiMetadata;
-    //QDR - used to get ROR name from ExternalVocabularyValue via pidProvider.get
-    private PidProvider pidProvider = null;
 
     public XmlMetadataTemplate() {
     }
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
Tests are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11225/2/testReport/

I didn't run the code but it makes sense. We're deleting some QDR-specific cruft, it looks like. Approved.

ofahimIQSS:
tests are passing but regression testing underway - wlll provide further updates shortly

PR #11230, opened 2025-02-07
Delete Files From Dataset

qqmyers:
This PR adds a deleteFiles method to the native API for datasets, using PUT as the HTTP verb to match the existing delete metadata calls. 

- Closes #

Note for posterity: To be able to use DELETE with a payload, Payara has to have an option set:
```
asadmin set
configs.config.default-config.network-config.protocols.protocol.http-listener-1.http.allow-payload-for-undefined-http-methods=true

```
The consensus in slack was to not use this approach, but if we decide to in the future, we'd need to include instructions to make this change in Payara.

There's an IT test that creates a dataset with files, try to delete some, publishes the dataset and then tries to delete others. It also tests using a bad fileId or datasetId or using a user w/o permissions. Those or other cases could be manually tested as well. The documentation (in the native api section) gives a curl example.

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -3547,9 +3547,6 @@ To remove all links to blocks, send an empty array.
 Delete Files from a Dataset
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Delete Files from a Dataset
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
 Delete files from a dataset. This API call allows you to delete multiple files from a dataset in a single operation.
 
 .. code-block:: bash
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -3544,6 +3544,42 @@ To update the blocks that are linked, send an array with those blocks.
 
 To remove all links to blocks, send an empty array.
 
+Delete Files from a Dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Delete Files from a Dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Delete files from a dataset. This API call allows you to delete multiple files from a dataset in a single operation.
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+  export PERSISTENT_IDENTIFIER=doi:10.5072/FK2ABCDEF
+
+  curl -H "X-Dataverse-key:$API_TOKEN" -X PUT "$SERVER_URL/api/datasets/:persistentId/deleteFiles?persistentId=$PERSISTENT_IDENTIFIER" \
+  -H "Content-Type: application/json" \
+  -d '{"fileIds": [1, 2, 3]}'
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X PUT "https://demo.dataverse.org/api/datasets/:persistentId/deleteFiles?persistentId=doi:10.5072/FK2ABCDEF" \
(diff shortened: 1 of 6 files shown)


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -3544,6 +3544,42 @@ To update the blocks that are linked, send an array with those blocks.
 
 To remove all links to blocks, send an empty array.
 
+Delete Files from a Dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Delete Files from a Dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
stevenwinship, line 3551: Remove duplicate header

stevenwinship (changes requested):

src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/XmlMetadataTemplate.java
```diff
@@ -98,13 +96,6 @@ private void generateXML(DvObject dvObject, OutputStream outputStream) throws XM
...
-        //QDR - used to get ROR name from ExternalVocabularyValue via pidProvider.get
-        GlobalId pid = null;
-        pid = dvObject.getGlobalId();
-        if ((pid == null) && (dvObject instanceof DataFile df)) {
-                pid = df.getOwner().getGlobalId();
-            }
-        pidProvider = PidUtil.getPidProvider(pid.getProviderId());
         XMLStreamWriter xmlw = XMLOutputFactory.newInstance().createXMLStreamWriter(outputStream);
         xmlw.writeStartElement("resource");
         boolean deaccessioned=false;
```
stevenwinship, line 101: This change is already in a PR that is in QA. Once it's merged please remove this and merge from develop

stevenwinship (approved):

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -5368,4 +5370,67 @@ public Response updateDatasetTypeLinksWithMetadataBlocks(@Context ContainerReque
...
+        }
+        return response(req -> {
+            Dataset dataset = findDatasetOrDie(id);
+            // Convert JsonArray to List<Long>
+            List<Long> fileIdList = new ArrayList<>();
+            for (JsonValue value : fileIds) {
+                fileIdList.add(((JsonNumber) value).longValue());
+            }
+            // Find the files to be deleted
+            List<FileMetadata> filesToDelete = dataset.getOrCreateEditVersion().getFileMetadatas().stream()
```
janvanmansum, line 5392: @qqmyers : maybe require all the file ids listed in the request to be in the dataset, instead of just ignoring the ones not found?

qqmyers: They aren't just ignored. Line 5400 below will return a 400 if all the files aren't in the dataset.

janvanmansum: OK, but still, it could fail faster and return a list of missing IDs.

janvanmansum: What I propose is compiling the list of file IDs in the request that are not in the dataset and returning the 400 status at line 5395 with in the body a message like "Files not found in dataset: [..ids..]". If I understand the code correctly, the failure will currently be thrown out of the `commandEngine.submit()` call.

qqmyers: It is currently thrown out in the second of the next two if clauses. The first checks to see if there were no files that match, and the second checks to see if there are fewer files to delete than were in the original list. (Basically unless all files are in the latest dataset version, you get a 400 error, but you'll get different messages depending on whether there are none or some.)

In general, I'm hesitant to do extra work to help an API caller correct their input - any script using the call should be getting the right info to send, and the list of file ids in the dataset is available through other calls. That said, if there's a use case where it is hard for a script to figure out what to send, we could return more details. Perhaps even send the ones that were found instead since that's really the useful list?

ofahimIQSS:
I see continuous integration keeps failing for this PR.

ofahimIQSS:
Testing Passed in internal - no issues found. Merging PR

PR #11118, opened 2024-12-19
add ROR as an Author Identifier Type

pdurbin:
**What this PR does / why we need it**:

This PR adds ROR as an Author Identifier Type.

**Which issue(s) this PR closes**:

- Closes #11075
- Closes #11149

**Special notes for your reviewer**:

As of 1fe95d9 a second regex for ROR was added. Now we have "ROR" to match just the identifier and "ROR_FULL_URL" to match the full ROR URL. The latter is used when exporting the "Datacite" format.

I have a question. How can a ROR uniquely identify an author? If I use the ROR for Harvard ( https://ror.org/03vek6s52 ), how can someone know which of the thousands of Harvard affiliates I am? 🤔 

Screenshot 2024-12-19 at 3 32 22 PM

~~Also, is it ok that ROR has been added to the bottom of the list? That's what we've always done before. Here's how it looks:~~

Screenshot 2024-12-19 at 3 31 46 PM

Nevermind. This was fixed in ddb6bc9. ROR is now after ORCID:

Screenshot 2024-12-20 at 11 50 35 AM

**Suggestions on how to test this**:

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

Yes, see screenshots above.

**Is there a release notes update needed for this change?**:

Yes, included.

**Additional documentation**:

Preview at https://dataverse-guide--11118.org.readthedocs.build/en/11118/user/dataset-management.html#adding-a-new-dataset

PR #11246, opened 2025-02-12
IQSS/11242 ExternalIdentifier fix

### LINKED ISSUE #11242: Dataverse External Author Vocabulary error when sending ORCID values to DataCite
1. Activate CVocConf (https://github.com/gdcc/dataverse-external-vocab-support/blob/main/examples/authorIDandAffilationUsingORCIDandROR.md)
2. Create Dataset
3. Introduce any author using ORCID author functionality 
4. Publish Dataset

- While publishing Dataset.

- You can see it on this page: $SERVER/api/datasets/export?exporter=Datacite&persistentId=doi:xxxx/xxxx, after Dataset publication.

**What happens?**

- Datacite does not recive ORCID values and Datacite Metadata export has not ORCID fields.

- The problem is that the file people.js (https://github.com/gdcc/dataverse-external-vocab-support/blob/main/scripts/people.js) fills the hidden field Author -> Identifier with the pattern https://orcid.org/xxxx-xxxx-xxxx-xxxx and then Dataverse, in ExternalIdentifier.java, does not consider that possibility and the getIdentifierAsUrl method of the DatasetAuthor class return null.

- All users

- Datacite recived ORCID fields and Datacite Metadata export was correct.

- 6.4

- No

**Screenshots:**

* ORCID author functionality
Image

* Hidden Identifier Field
Image

* Database Registry
Image

* Datacite Metadata export
Image

* Datacite Test Registry
Image

- Our organization fixed the issue by changing src/main/java/edu/harvard/iq/dataverse/ExternalIdentifier.java so that Author -> Identifier with the format: https://orcid.org/xxxx-xxxx-xxxx-xxxx is accepted. We are not sure if would be better to change (https://github.com/gdcc/dataverse-external-vocab-support/blob/main/scripts/people.js) instead.

* Correct Datacite Metadata export
Image

* Correct Datacite Test Registry
Image

* Code Change Possible Solution
Image

qqmyers:
**What this PR does / why we need it**: This PR fixes a bug in the ExternalIdentifier class that caused ORCIDs in URL form to not be recognized, breaking part of the external vocab functionality w.r.t. ORCID.

- Closes #11242

**Special notes for your reviewer**: As the fix is generic, I reverted changes in the ROR recognition to avoid having two separate versions. Also added tests for formatting or ORCID and ROR.
Also - I noted that it looks like some tests in DatasetFieldValueValidatorTest duplicate those in ExternalIdentifier. If a reviewer confirms that, I'd be happy to remove them (and the isValidAuthorIdentifier() method that is only used in those tests).

**Suggestions on how to test this**: Verify that publishing a dataset with an ORCID in numeric form or URL form as an author ID, the ORCID appears in the DataCite XML. Verify that either form of ORCID/ROR show up as valid links in the page display (without external vocab scripts turned on).

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/ExternalIdentifier.java
```diff
@@ -4,7 +4,7 @@
 import java.util.regex.Pattern;
 
 public enum ExternalIdentifier {
-    ORCID("ORCID", "https://orcid.org/%s", "^\\d{4}-\\d{4}-\\d{4}-(\\d{4}|\\d{3}X)$"),
+    ORCID("ORCID", "https://orcid.org/%s", "^(https:\\/\\/orcid.org\\/)?\\d{4}-\\d{4}-\\d{4}-(\\d{4}|\\d{3}X)$"),
```
pdurbin, line 7: As of https://github.com/IQSS/dataverse/pull/11118/commits/b9d0146b07c9c88e0b5b8759fe2cff10df19029d we say the following:

`When entering author identifiers, select the type from the dropdown (e.g. "ORCID") and under "Identifier" enter just the unique identifier (e.g. "0000-0002-1825-0097") rather than the full URL (e.g. "https://orcid.org/0000-0002-1825-0097").`

Should this be re-written?

jggautier: Hey all. @pdurbin pinged me about this.

The short form, or the non-URL form, is preferable for ORCIDs that users type in or paste in the Author Identifier field, because that's what makes the link clickable, right? So the same will hold true for RORs that users type in or paste in the Author Identifier field, right?

qqmyers: The earlier fix changed ROR so the short form would be a link. This PR makes it so that either form of ORCID or ROR should result in a valid link.

jggautier: Ah okay. Also, when you wrote "happy to change it if we want to allow a choice for manual entry", I thought we can already allow a choice for manual entry of identifiers. That's how it's set up on Demo Dataverse and planned for Harvard Dataverse.

Or did you mean _when_ installations configure an external vocab script so that users can enter the identifier in the Identifier field that appears when they enter their own name instead of selecting one from the suggestions?

About the text in [b9d0146](https://github.com/IQSS/dataverse/commit/b9d0146b07c9c88e0b5b8759fe2cff10df19029d), it's written to apply to all identifiers, so I'm testing on Demo Dataverse to review how the other identifiers that users enter in the Author Identifier field are displayed, especially when the identifier has a URL form.

qqmyers: I meant that we could change the text Phil highlighted to tell people they can enter either form. As I mentioned, there's no validation so they could just enter a random string. That won't get recognized as a real ORCID/ROR so won't go to DataCite, etc. So the guides are, at some level, just telling people what we recommend. We can keep that to recommending the short form, or we can tell people that either is fine and they can enter it how they like.

With the ext. vocab script, the URL form is used. That's more robust for other fields, like depositor, where there is no separate field to indicate the identifier type. I think it is possible to just type a text string for the name and then manually select ORCID and enter an ORCID (same as for any other identifier type), but I don't know that there's ever a reason we need to tell people about that. (And if we need to shut that option off, we could. E.g. the script could remove ORCID as an option when it shows the ID type field.)

jggautier: I see a difference between how one identifier, VIAF, is displayed as a link depending on if the URL form is used or not. In the Metadata tab of the dataset page of https://doi.org/10.7910/DVN/GCDIHK in Harvard Dataverse, the VIAF ID is displayed as a link when the non-URL form is entered in the Author Identifier field, and displayed as text when the URL form is entered. Here's a screenshot of what I mean from a test dataset I made on Harvard Dataverse:

<kbd>Screenshot 2025-02-12 at 2 04 19 PM</kbd>

But on Demo Dataverse, VIAF IDs are displayed as text regardless of what the user enters in the Identifier field. 

So I'm testing on Harvard Dataverse as well so see what happens for the other identifiers in that installation.

qqmyers: The other identifier types do not currently recognize the URL forms. I can expand the scope here to allow that. If you want this, let me know. That still leaves the question of whether we then tell users that either form of an identifier is OK in the docs, but at least all identifiers would work the same way.

For the problem on demo - it's a bug/limitation that when a field is managed by a cvoc script, Dataverse's internal reformatting of the value isn't applied, so for non-ORCID ids in the author identifier field, you'll see what is typed. That could be fixed a number of ways, probably most easily in the script itself. If it's a priority, I'd suggest an issue in the external vocab repo.

qqmyers: I've updated the code so that all identifiers will handle both URL and short form (when no scripts are running).
I expanded the tests so that all identifier types are actually checked and verified that both the URL and short forms get format()ed to the URL form (which is specifically relevant to this issue)
I also fixed the ISNI matcher which didn't support ISNIs ending in X instead of a digit.
I've left the text alone.
I'll look for the issue in the ext-vocab repo - guessing we'll need to have similar regex detection to decide when to turn the values into links.

With that, I think this is ready to continue forward?

jggautier: Ah! So "The exception is the ISNI identifier type; it's never displayed as a link" isn't true. It is displayed as a link for ISNIs that end in digits, but not for ISNIs that end in X. Thanks for looking into that and for fixing it!

I'm not sure what you mean by "all identifiers will handle both URL and short form". Does that mean that for an identifier that has a URL form and a short form, when a user enters either form into the Author Identifier field, it's displayed as a link on the dataset page and it's included in DataCite exports and sent to DataCite (for repositories that register DOIs)?

jggautier: I created that GitHub issue in the ext-vocab repo. It's at https://github.com/gdcc/dataverse-external-vocab-support/issues/33. There are some details I couldn't be sure of because:
- I'm testing on Demo Dataverse and the change won't happen on Demo Dataverse's Author fields until v6.6 is released and applied to Demo Dataverse, like how RORs are displayed on the dataset page
- the Author fields on Demo Dataverse are behaving in ways I didn't expect, like how the dataset page displays an ORCID logo even when the user enters the ORCID by using the Author Identifier Type and Author Identifier fields
- I'm not sure about what changes were made so that the dataset page displays IDs as links when users enter IDs in Author fields where the cvoc script isn't applied. This is what I'm trying to clarify in my last comment.

I can help update that GitHub issue as I understand more.

qqmyers: All the identifier types have both a short and URI form (FWIW DAI was not handled before - I've added it). Entering either as the author identifier will cause it to display as a link in the metadata display. The DataCite XML generator adds the identifier, in URI form, regardless of the identifier type, so yes they are all included regardless of the short or URI input form.

pdurbin: Something should be written somewhere about this confusing situation.

What if you run Dataverse for several years with out the cvoc scripts? Your database has lots of "just identifiers" values in it?

Then, you turn on cvoc scripts and run for another couple years. Now your database has a mix of (older) "just identifiers" and newer "full URL" values?

Do I have this right? Is this what we want? Shouldn't the database be consistent?

Apologies if I'm misunderstanding.

jggautier: For installation managers and curators who expect at least some users to have to enter or paste identifiers into the Author Identifier field, I think they would be happy to know about this change. I imagine that they'll no longer have to tell their users to enter the identifiers in a particular way, since there will be better support for both ways that we expect users to enter identifiers when they have to enter them.

So maybe a release note could help those installation managers and curators know this? If we agree, I'd be happy to write one or help write one.

> What if you run Dataverse for several years with out the cvoc scripts? Your database has lots of "just identifiers" values in it?
> 
> Then, you turn on cvoc scripts and run for another couple years. Now your database has a mix of (older) "just identifiers" and newer "full URL" values?

Yeah, I agree. And this has been the situation with Dataverse installations for a while, even those whose managers haven't used the cvoc scripts. Their databases have combinations of "just identifiers" (or the "short forms"), "full URL" values, and the wrong Author Identifier Type chosen or no Author Identifier Type chosen.

> Do I have this right? Is this what we want? Shouldn't the database be consistent?

From my understanding, the inconsistency in the database has complicated our ability to ensure that when most of the identifiers are displayed on the dataset page, the identifiers are links that users can click on to learn more about the person or organization. And its complicated our ability to include identifiers in metadata exports in order to improve discovery and re-use of those identifiers in Dataverse repositories and in other systems, like DataCite.

From @qqmyers's work in this PR, I've gotten the impression that this complication, having a mix of "just identifiers" and "full URL" values in the database, has been manageable, at least for the Author Identifier field, even if the inconsistency isn't ideal.

In other cases, like licenses and terms of use metadata, I think we've decided that inconsistencies in the database aren't manageable, so we've added field validation, "multiple license support", and included ways for repositories to make existing values in their databases more consistent, like the "Flyway scripts".

I think of field validation and ways to change database values in bulk as ways to address inconsistencies in other parts of the database.

The cvoc scripts, and the ability to associate ORCIDs with more user accounts and prefill Author Identifier fields with those ORCIDs, also address this by making it less likely that users will need to enter anything in the Author Identifier field, and giving installation managers more control over what gets in the database. Those things are new, and won't work all of the time, like in cases where users need and are allowed to use other types of identifiers.

Maybe before considering other ways of ensuring that the identifiers are displayed as links and help make deposits more findable within and outside of Dataverse repositories, it might be helpful to learn how well what's been done so far is working.

src/test/java/edu/harvard/iq/dataverse/DatasetFieldValueValidatorTest.java
```diff
@@ -111,6 +111,9 @@ public void testIsValid() {
         assertFalse(result);
     }
 
+    /* The next 7 tests below appear to duplicate tests in the ExternalIdentifierTest class.
+     * The ones here use the isValidAuthorIdentifier method which is only used in testing (and probabkly could be static).
+     */
```
sekmiller, line 116: Yes, these tests duplicate those in the ExternalIdentifierTest and can probably be removed. The isValidAuthorIdentifier method is actually used in the DatasetAuthor class (around line 100).

qqmyers: I think DatasetAuthor uses isValidIdentifier() without 'Author' . I'll go ahead and remove the tests.

jggautier:
@qqmyers I saw that in the release notes you wrote that "the shorter form is still recommended when doing manual entry." 

But on Tuesday I wrote that I thought you're changes meant that installation managers and curators "will no longer have to tell their users to enter the identifiers in a particular way, since there will be better support for both ways that we expect users to enter identifiers when they have to enter them."

But I'm not sure why the shorter form is still recommended when doing manual entry. Could you write more about that here?

qqmyers:
Just depends on what you want to see:
image

or
image

I assumed the former is still preferable, but feel free to edit the instructions as you see fit.

sekmiller (approved):

jggautier:
Thanks @qqmyers!

> FWIW: ORCID also says the short form should be used "When human readability is more important than technical correctness. ".

I think they mean that when the page can't display the full URL, it should display the short form, like `0000-0002-2771-9344`, instead of displaying part of the full URL, like `https://orcid.org/0000-000`.

> One of the criticisms of the short form is that "omitting the URL scheme and domain can lead to confusion elsewhere" but we display the type of the identifier and the logo, and we include the long form in machine-readable formats. All of them appear to be concerned that the short forms, when displayed as text, hide the fact that these are links, but our UI is HTML and shows them as links and "Copy URL" works to get the URL form.

I agree. I'm glad we're getting more specific about the why.

I've also assumed that displaying the full URL makes it easy for others to copy and paste it to other places, and that's why a lot of depositors have been pasting full URL forms of identifiers into several identifiers fields. That's been my experience when I've had to copy and paste ORCIDs and DOIs. Although we could also say that because on Dataverse pages the text are links, people can just as easily copy the text links' URLs to paste them in other places.

> Also FWIW: The ORCID/ROR scripts use the inline form, and, for ORCIDs, it's also true that all of our entries are unauthenticated and should probably be displayed in that format instead.

This is interesting! But I think outside of the scope of this PR, right?

> We also have legacy records that are in the short form (as Phil pointed out), which means changing the recommendation to long form will result in differences between old/new records.

Thanks yeah. There are already differences in the records in Dataverse repositories, which I wrote is unfortunate but manageable.

I think that the only reason why we've recommended the short form was because Dataverse depended on that form to display IDs and include them in metadata exports. That's changed in this PR, so I think we should recommend the URL form to support what many people are already doing and because the dataset page has the room to display the full URL forms.

jggautier:
@ofahimIQSS, in my last comment I'm suggesting that we change what we write in this PR's release notes. I don't think we've agreed so I'm waiting to hear back from @qqmyers and possibly @sekmiller and @pdurbin before this is merged.

ofahimIQSS:
I tried testing this PR this morning in internal but am getting internal server error on creation/accessing datasets/collections. 
image

ofahimIQSS:
Had an observation with this PR. ROR url is not displayed on datacite after it is added on a second publish. 

Steps to reproduce:
1. Create a new dataset with all required fields, include just the ORCID. 
2. Publish the Dataset
3. Check the datacite XML file. The ORCID is displayed
4. Now edit the metadata, add a ROR ID
5. Publish, check datacity XML File
The ROR ID is missing from the XML. 

image

Test Data: https://dataverse-internal.iq.harvard.edu/api/datasets/export?exporter=Datacite&persistentId=doi%3A10.70122/FK2/G1HUDB

qqmyers:
Step 5 hasn't been done/didn't work - the dataset is still draft.

PR #11247, opened 2025-02-12
11188 API to get list of dataset versions with summary of differences

### LINKED ISSUE #11188: Dataset Versions API: add version differences
For the SPA, we need to update the List Dataset Versions API with the differences between each version, for displaying in the Dataset Versions tab.
For example, 

Image

There is already an endpoint for getting the differences between any two versions -  https://guides.dataverse.org/en/latest/api/native-api.html#compare-versions-of-a-dataset.  The same logic can be re-used to return the differences between versions.

API user

SPA: Dataset Versions Tab

Add version difference information to List Dataset Versions API

sekmiller:
**What this PR does / why we need it**: For the SPA we need an api that returns a list of Dataset Versions for the Versions Tab. The versions tab also shows a summary of differences between consecutive versions when available so that is returned as part of the response.

- Closes #11188 Dataset Versions API add version differences

**Special notes for your reviewer**: As per request of Ellen, the Dataset Field type names and Metadata block names are in display name format. The DatasetsIT#testCompareDatasetVersionsAPI is failing - not sure why it shouldn't have been changed in my branch, but the order of two metadata field is reversed.

**Suggestions on how to test this**: Try the api for a dataset with multiple versions and see that the response gives the info required to load into the versions tab of the dataset page. If there's a draft version use tokens with and without view unpublished permissions.

**Additional documentation**: added to the native api doc


src/main/webapp/dataset-versions.xhtml
```diff
@@ -77,28 +77,28 @@
...
-                        <h:outputText rendered="#{(blockNote[1]) > 0 and (blockNote[2] + blockNote[3]) > 0}" value=", " />
-                        <h:outputText rendered="#{(blockNote[2]) > 0}" value="#{blockNote[2]} #{bundle['file.dataFilesTab.versions.removed']}" />
-                        <h:outputText rendered="#{(blockNote[2]) > 0 and (blockNote[3]) > 0}" value=", " />
-                        <h:outputText rendered="#{(blockNote[3]) > 0}" value="#{blockNote[3]} #{bundle['file.dataFilesTab.versions.changed']}" />
+                        <h:outputText rendered="#{blockNote.added > 0}" value="#{blockNote.added} #{bundle['file.dataFilesTab.versions.added']}" />
+                        <h:outputText rendered="#{(blockNote.added) > 0 and (blockNote.changed + blockNote.deleted) > 0}" value=", " />
+                        <h:outputText rendered="#{(blockNote.deleted) > 0}" value="#{blockNote.deleted} #{bundle['file.dataFilesTab.versions.removed']}" />
+                        <h:outputText rendered="#{(blockNote.changed) > 0 and (blockNote.changed) > 0}" value=", " />
```
stevenwinship, line 100: should this be 
(blockNote.**deleted**) > 0 and (blockNote.changed)

stevenwinship (changes requested):

src/main/webapp/dataset-versions.xhtml
sekmiller: Yes. nice catch. thanks.

stevenwinship (approved):

g-saracca:
Endpoint working correctly, already using it and testing it in [this issue](https://github.com/IQSS/dataverse-client-javascript/issues/260) I am working on.

PR #11259, opened 2025-02-13
Added new option in getDataverse endpoint for optionally retrieving child count

GPortas:

Adds a new query param, returnChildCount, to getDataverse endpoint for optionally retrieving the child count, which represents the number of dataverses, datasets, or files within the dataverse.

- Closes #11259

If you want to obtain the child count of a Dataverse, you must set ``returnChildCount`` query parameter to ``true``. Example:
 
`curl "https://demo.dataverse.org/api/dataverses/root?returnChildCount=true"
`

No

Yes, attached.

**Additional documentation**:

N/A

### DIFF
--- a/doc/release-notes/11255-get-dataverse-api-return-child-count.md
+++ b/doc/release-notes/11255-get-dataverse-api-return-child-count.md
@@ -1,0 +1,1 @@
+Added a new query param, ``returnChildCount``, to getDataverse endpoint(``/api/dataverses/<ID>``) for optionally retrieving the child count, which represents the number of dataverses, datasets, or files within the dataverse.
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -177,6 +177,15 @@ Usage example:
 
   curl "https://demo.dataverse.org/api/dataverses/root?returnOwners=true"
 
+If you want to include the child count of the Dataverse, which represents the number of dataverses, datasets, or files within the dataverse, you must set ``returnChildCount`` query parameter to ``true``.
+
+Usage example:
+
+.. code-block:: bash
+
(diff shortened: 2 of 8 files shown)


GPortas:
> @GPortas do we need to account for published/unpublished based on the user making the call? Also do we need to go more than one layer in (get the objects owned by the collections owned by the given collection)?

For the purpose of this extension, we do not need to differentiate between published and unpublished states based on the user. FWIW, this endpoint uses the same database query as the JSF logic we are replicating in the SPA to determine whether a Dataverse has data (see [Dataverse.java#L49](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/Dataverse.java#L49)).

The key difference is that, for the API, instead of simply returning whether the Dataverse has data or is empty, I found it more useful for other use cases to return the child count, so the SPA only needs to check if the value is greater than 0.

Additionally, we only need to return the direct child count, not multiple layers.

Would it be helpful to clarify these points further in the field naming or documentation to avoid confusion?

ofahimIQSS:
no issues found with this PR - merging

pdurbin:
> Adds a new query param, returnChildCount, to getDataverse endpoint for optionally retrieving the child count, which represents the number of dataverses, datasets, or files within the dataverse.

@GPortas @sekmiller @ofahimIQSS are these direct children only?

Or does it count children, grandchildren, great-grandchildren, etc?

I'm asking because we might we might want to clarify the guides and release note.

sekmiller:
@pdurbin it is only direct children

pdurbin:
@sekmiller @GPortas ok, thanks. I made a PR to clarify the docs:

- #11286

The idea is avoid disappointment. API users might otherwise assume `returnChildCount=true` is a way to get all children, grandchildren, great grandchildren, etc.

PR #11266, opened 2025-02-14
index after file ingest

### LINKED ISSUE #11182: Search API not returning variables and observations for tabular data
The search API doesn't return the variables and observations of newly ingested files.  It seems to need a subsequent update to the file, such as publishing the dataset, or editing the file metadata, for the search API to return the ingested version of the file.

This is a problem for the SPA because it is using the Search API for getting all the file information it displays on the Collection Page.

Create a Dataset and add a tabular file to the Dataset.  Go to the JSF Dataverse page, and see that the file has been ingested and that it is showing tabular data information, for example:

Image

Retrieve the file information from the Search API:

http://localhost:8000/api/v1/search?q=*&subtree=ingested-file-collection&type=file

The returned file is not the ingested version, it doesn't contain variables or observations:

```
{
  "status": "OK",
  "data": {
    "q": "*",
    "total_count": 1,
    "start": 0,
    "spelling_alternatives": {},
    "items": [
      {
        "name": "pums_1000.csv",
        "type": "file",
        "url": "http://localhost:8080/api/access/datafile/24",
        "file_id": "24",
        "description": "",
        "file_type": "Comma Separated Values",
        "file_content_type": "text/comma-separated-values",
        "size_in_bytes": 16969,
        "md5": "874aef476ffb8a1e98d68fcf5f780990",
        "checksum": {
          "type": "MD5",
          "value": "874aef476ffb8a1e98d68fcf5f780990"
        },
        "dataset_name": "Ingested File Dataset",
        "dataset_id": "23",
        "dataset_persistent_id": "doi:10.5072/FK2/SDRRRM",
        "dataset_citation": "Admin, Dataverse, 2025, \"Ingested File Dataset\", https://doi.org/10.5072/FK2/SDRRRM, Root, DRAFT VERSION",
        "restricted": false,
        "canDownloadFile": true,
        "publicationStatuses": [
          "Unpublished",
          "Draft"
        ],
        "releaseOrCreateDate": "2025-01-24T13:42:42Z"
      }
    ],
    "count_in_response": 1
  }
}
```

* When does this issue occur?
Search API

NA

* What happens?
doesn't return  ingested version of the file

API users

Should return Tabular file type, variables and observations.

The unstable Dataverse Docker image

pdurbin:

Files that are successfully ingested are not reindexed at tabular files. This means they don't have fields such as "variables", "observations", etc.

- Closes #11182

I made the simplest possible fix for this first attempt. Probably we should reduce the amount of indexing if we can. For example, if several newly ingested files come from the same dataset, we should probably only reindex the dataset at the end? I'm not sure how long ingesting will take, though. 🤔 

- Upload a file that will successfully ingest
- Check it with the Search API? Do you see fields like "variables" and "observations"?

No.

I didn't bother.

**Additional documentation**:

None.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessageBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessageBean.java
@@ -23,6 +23,7 @@ package edu.harvard.iq.dataverse.ingest;
 import edu.harvard.iq.dataverse.*;
 import edu.harvard.iq.dataverse.authorization.AuthenticationServiceBean;
 import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
+import edu.harvard.iq.dataverse.search.IndexServiceBean;
 import edu.harvard.iq.dataverse.util.BundleUtil;
 
 import java.sql.Timestamp;
@@ -60,6 +61,7 @@ public class IngestMessageBean implements MessageListener {
     @EJB IngestServiceBean ingestService;
     @EJB UserNotificationServiceBean userNotificationService;
     @EJB AuthenticationServiceBean authenticationServiceBean;
+    @EJB IndexServiceBean indexService;
 
    
     public IngestMessageBean() {
@@ -111,6 +113,7 @@ public class IngestMessageBean implements MessageListener {
                         // and "mixed success and failure" emails. Now we never list successfully
                         // ingested files so this line is commented out.
                         // sbIngestedFiles.append(String.format("<li>%s</li>", datafile.getCurrentName()));
+                        indexService.indexDataset(datafile.getOwner(), true);
                     } else {
                         logger.warning("Error occurred during ingest job for file id " + datafile_id + "!");
                         sbIngestedFiles.append(String.format("<li>%s</li>", datafile.getCurrentName()));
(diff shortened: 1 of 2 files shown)


src/main/java/edu/harvard/iq/dataverse/ingest/IngestMessageBean.java
```diff
@@ -111,6 +113,7 @@ public void onMessage(Message message) {
                         // and "mixed success and failure" emails. Now we never list successfully
                         // ingested files so this line is commented out.
                         // sbIngestedFiles.append(String.format("<li>%s</li>", datafile.getCurrentName()));
+                        indexService.indexDataset(datafile.getOwner(), true);
```
qqmyers, line 116: I'd suggest using asyncIndexDataset - that will put the indexing in a queue and if multiple ingests complete at nearly the same time indexing will only get done once (fewer times anyway).

pdurbin: Sure, done in b6f484c.

qqmyers (approved):
Looks good - just a release note?

src/test/java/edu/harvard/iq/dataverse/api/SearchIT.java
```diff
@@ -1838,4 +1842,53 @@ public void testShowTypeCounts() throws InterruptedException  {
...
+        uploadFile.prettyPrint();
+        uploadFile.then().assertThat()
+                .statusCode(OK.getStatusCode())
+                .body("data.files[0].label", equalTo("data.csv"));
+
+        assertTrue(UtilIT.sleepForLock(datasetId.longValue(), "Ingest", apiToken, UtilIT.MAXIMUM_INGEST_LOCK_DURATION), "Failed test if Ingest Lock exceeds max duration " + pathToDataFile);
+
+        Long fileId = JsonPath.from(uploadFile.body().asString()).getLong("data.files[0].dataFile.id");
+
+        Response search = UtilIT.search("entityId:" + fileId, apiToken);
```
qqmyers, line 1885: If this fails, may need a sleep for indexing as well.

ofahimIQSS:
Tests are now passing, mering pr

PR #11267, opened 2025-02-14
return more perms from /api/access/datafile/{id}/userPermissions

### LINKED ISSUE #11226: User Permissions on a File - Extend API information
Extend the [Get User Permissions on a File](https://guides.dataverse.org/en/latest/api/dataaccess.html#get-user-permissions-on-a-file) api endpoint to also return this information:
- canEditFileMetadata
- canRestrictFile
- canReplaceFile
- canDeleteFile

API User

SPA Q1 2025 Roadmap

None

More information about user permissions on a file.

pdurbin:

API users (especially the frontend team, which is writing the [new UI](https://github.com/IQSS/dataverse-frontend)) need more granularity when checking file permissions.

- Closes #11226

Instead of adding these four new permissions, the API user could simply check the existing "canEditOwnerDataset" instead. I'm curious where these permissions come from? JSF? They aren't in our main [Permission class](https://github.com/IQSS/dataverse/blob/v6.5/src/main/java/edu/harvard/iq/dataverse/authorization/Permission.java).

I cleaned up existing assertions and added a new "no perms" user and additional assertions.

Create some users. Give them various permissions. Check the `/api/access/datafile/{id}/userPermissions` API. Does it return the right results?

No.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -1711,6 +1711,10 @@ public class Access extends AbstractApiBean {
         jsonObjectBuilder.add("canDownloadFile", permissionService.userOn(requestUser, dataFile).has(Permission.DownloadFile));
         jsonObjectBuilder.add("canManageFilePermissions", permissionService.userOn(requestUser, dataFile).has(Permission.ManageFilePermissions));
         jsonObjectBuilder.add("canEditOwnerDataset", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
+        jsonObjectBuilder.add("canEditFileMetadata", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
+        jsonObjectBuilder.add("canRestrictFile", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
+        jsonObjectBuilder.add("canReplaceFile", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
+        jsonObjectBuilder.add("canDeleteFile", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
         return ok(jsonObjectBuilder);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Access.java
```diff
@@ -1711,6 +1711,10 @@ public Response getUserPermissionsOnFile(@Context ContainerRequestContext crc, @
         jsonObjectBuilder.add("canDownloadFile", permissionService.userOn(requestUser, dataFile).has(Permission.DownloadFile));
         jsonObjectBuilder.add("canManageFilePermissions", permissionService.userOn(requestUser, dataFile).has(Permission.ManageFilePermissions));
         jsonObjectBuilder.add("canEditOwnerDataset", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
+        jsonObjectBuilder.add("canEditFileMetadata", permissionService.userOn(requestUser, dataFile.getOwner()).has(Permission.EditDataset));
```
qqmyers, line 1714: So the code returns the answer to Permission.EditDataset 5 times in a row?

g-saracca: Hi @pdurbin and @qqmyers, if you think that with the canEditOwnerDataset property we are already covered to handle the actions on a file I think this PR is not really going to be needed.

g-saracca: FYI I went ahead and use the `canEditOwnerDataset` permission to allow or not edit file actions in the UI like **_edit file metadata, restrict, replace and delete file._**
I think is enough with that, sorry for wasting your time with this, I thought they were different permissions.

g-saracca:
@pdurbin FYI I went ahead and use the canEditOwnerDataset permission to allow or not edit file actions in the UI like edit file metadata, restrict, replace and delete file.
I think is enough with that, sorry for wasting your time with this, I thought they were different permissions.
If you want we can close this PR and the related issue.

pdurbin:
@g-saracca no worries. It took very little time to make the PR. Please feel free to close the PR and issue if you don't need them.

g-saracca:
Thanks @pdurbin, closing then. We already have what we need for the Delete File functionality in the SPA.

PR #11269, opened 2025-02-17
IQSS/11264 - Add pseudo-exponential backoff to dataset page refresh

### LINKED ISSUE #11264: Javascripts loading in a loop with datasets in “in review” status

1. create a new dataset with only Contributor permissions
2. Submit for review the dataset

ex: https://demo.dataverse.org/dataset.xhtml?persistentId=doi%3A10.70122%2FFK2%2FGHTWC0&version=DRAFT

* When does this issue occur?

This issue occurs with "in review" datasets on Dataverse installation with external vocabularies enabled.

Everything returns to normal once the dataset is published or returned to the [author]([url](url))

Dataset view

* What happens?

All users

tested with v6.2 and v6.5

**Screenshots:**

Image

qqmyers:
**What this PR does / why we need it**: The dataset page currently refreshes every 10 seconds looking for locks to be removed, which is not very useful for long-running locks, e.g. in review, or ingest, publish, etc. with large/many files. This PR adds a pseudo-exponential backoff - in the front-end JavaScript - in the rate of refresh to reduce the load. The constants chose make the updates slightly faster for the first two refreshes and then quickly backs off to 5 minute refreshes (after ~ 90 seconds, the new algorithm will have had fewer total refreshes and will only be refreshing ~once per minute)

- Closes #11264

**Special notes for your reviewer**: This is a minor improvement, but hopefully reduces server load for in review datasets in particular, as well as long running Globus transfers, pub with many files, etc.

**Suggestions on how to test this**: Put a dataset in review and check the refresh rate in the browser console - verify it is slowing down.

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11264-slower refresh for long running locks.md
+++ b/doc/release-notes/11264-slower refresh for long running locks.md
@@ -1,0 +1,1 @@
+When a dataset has a long running lock, including when it is 'in review', Dataverse will now slow the page refresh rate over time.
--- a/src/main/webapp/dataset.xhtml
+++ b/src/main/webapp/dataset.xhtml
@@ -858,6 +858,11 @@
                                 $(this).ready(function () {
                                     refreshIfStillLocked();
                                 });
+                            
+                                var initialInterval = 5000; // 5 seconds
+                                var maxInterval = 300000; // 5 minutes
+                                var currentInterval = initialInterval;
+                                var backoffFactor = 1.2; // Exponential factor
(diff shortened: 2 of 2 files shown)


sekmiller (approved):
This is fine. We could in the future think about making the max interval settable - or try something with varying intervals based on type of lock

ofahimIQSS:
PR tests passing. No issues found during regression.

PR #11273, opened 2025-02-19
Edit Metadata API endpoint - support empty values

### LINKED ISSUE #11243: Edit Metadata API endpoint - not allowing empty values
Try to update a dataset metadata sending a payload with empty values like:
```
[
...rest of required fields,
{
    "value": "",
    "typeClass": "primitive",
    "multiple": false,
    "typeName": "subtitle"
}]
```
And you will receive an error like:
```
{
    "status": "ERROR",
    "message": "Error parsing dataset update: Empty value for field: Subtitle"
}
```

* What happens?
API endpoint is returning an error, but on updates it should allow empty values on not required fields as users may want to remove a value from a field and leave it empty.

API users

Not to receive an error and update the metadata correctly.

unstable

After this gets fixed we should continue with frontend issue https://github.com/IQSS/dataverse-frontend/issues/598.

**Screenshots:**

Image

Image

Image

GPortas:

Extended the Edit Metadata API endpoint:

- This endpoint now allows removing fields (by sending empty values), as long as they are not required by the dataset.
- New ``sourceInternalVersionNumber`` optional query parameter, which prevents inconsistencies by managing updates that may occur from other users while a dataset is being edited.

A significant part of the changes involve refactoring the previous logic and addressing missing test coverage (Coveralls reports a +0.1% increase in coverage with these changes). Since I needed to modify the metadata validation logic and its update process, I first had to ensure that the underlying logic was both testable and extensible.

- Closes #11243

I have identified some areas for improvement in the field validation. For example, as mentioned in my original message on Slack:

_It looks like the API currently allows editing multiple fields in a dataset's metadata by setting empty values (e.g. ["", ""]) (https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1188), since the only check performed is that the list is not empty ([]). However, in the case of single fields, it ensures that the string is not empty (https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1191).
Is this intended behavior? Or should the API also check that the values in the list are not empty? This will change in the context of https://github.com/IQSS/dataverse/issues/11243, but I’d like to clarify the current expected behavior._

Finally, I have decided not to implement any changes to the validation logic aside from those necessary for field removal, in order to avoid overextending the scope of this PR and further delaying it. I am willing to create separate issues to apply the enhancements in the future if necessary.

Follow the updated docs for Edit Metadata API endpoint.

No

Yes, attached.

**Additional documentation**:

None


src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -461,44 +455,227 @@ public void testAddUpdateDatasetViaNativeAPI() {
...
+                "          },\n" +
+                "          \"authorAffiliation\": {\n" +
+                "            \"typeName\": \"authorIdentifierScheme\",\n" +
+                "            \"value\": \"\"\n" +
+                "          }\n" +
+                "        }\n" +
+                "      ]\n" +
+                "    }\n" +
+                "  ]\n" +
+                "}";
```
stevenwinship, line 546: Same comment as above

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -461,44 +455,227 @@ public void testAddUpdateDatasetViaNativeAPI() {
...
+                "            \"multiple\": false,\n" +
+                "            \"typeClass\": \"primitive\",\n" +
+                "            \"value\": \"DistributorAffiliation1\"\n" +
+                "          }\n" +
+                "          \n" +
+                "        }\n" +
+                "      ]\n" +
+                "    }\n" +
+                "  ]\n" +
+                "}\n";
```
stevenwinship, line 582: Same

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -461,44 +455,227 @@ public void testAddUpdateDatasetViaNativeAPI() {
...
+                "            \"multiple\": false,\n" +
+                "            \"typeClass\": \"primitive\",\n" +
+                "            \"value\": \"\"\n" +
+                "          }\n" +
+                "          \n" +
+                "        }\n" +
+                "      ]\n" +
+                "    }\n" +
+                "  ]\n" +
+                "}\n";
```
stevenwinship, line 617: Same

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -461,44 +455,227 @@ public void testAddUpdateDatasetViaNativeAPI() {
...
+        jsonString = "{\n" +
+                "  \"fields\": [\n" +
+                "    {\n" +
+                "      \"typeName\": \"alternativeTitle\",\n" +
+                "      \"multiple\": true,\n" +
+                "      \"typeClass\": \"primitive\",\n" +
+                "      \"value\": [\"Alternative1\",\"Alternative2\"]\n" +
+                "    }\n" +
+                "  ]\n" +
+                "}\n";
```
stevenwinship, line 637: Same

stevenwinship (approved):

ofahimIQSS:
tests passing in local - merging PR

PR #11281, opened 2025-02-21
External Search Refactor

qqmyers:
**What this PR does / why we need it**: This PR implements an experimental capability to have several configurable search services. It's a minimalistic implementation of the general design that's been discussed, but, with the example GetExternalSearchServiceBean and PostExternalSearchServiceBean classes it is intended to support developing/demonstrating a separate tool, and specifically the AI dataset search coming from Harvard. 

Which engine is used is controlled by a Mprofile/jvm option: -Ddataverse.search.default-service=<name>

where name is 'getExternalSearch', 'postExternalSearch', etc. as defined in the classes.

(The *ExternalSearchServiceBeans send all of the search parameters from Dataverse to a third party web service (via GET or POST) and expect to get back an ordered JsonArray of matching objects/datasets with "DOI"s and "Distance"s. These classes then send a query to Solr via Dataverse's internal SolrSearchServiceBean that will generate the output needed for the display. (Solr doesn't preserve the order in the query. The beans reorder the results based on Distance (smallest first).)

Both of the *ExternalSearchServiceBean classes also rely on a new/temporary `:<Get/Post>ExternalSearchUrl` parameter (easier to change than a JvmSetting) and an `:<Get/Post>ExternalSearchName` setting.

The GetExternalSearchServiceBean has been 'tested' with :GetExternalSearchUrl pointed at a static page returning a list of valid DOIs and Distances - can see in the logs that all the search params are sent as query params, can verify that final the search results contain only those entities, that the order is the same as what's sent, etc. The POST version is being tested on the QA machine with @siacus's AI search via POST.

There are other example beans: goldenOldies that only returns entries with entityIds <1000, and oddlyEnough that only returns entityIds that are odd. Neither of these is practical, but they allow some testing w/o any external service and, for oddlyEnough, show how one can deal partially deal with paging, changing the number of total results, etc. that come from solr.

This PR also restores the query_entities parameter in the search api. It remains defaulted to true (it was hardcoded true previously), but setting query_entities=false will now avoid the post-search lookup of additional metadata in postgres.

- Closes #

The ability to find/enable SearchService implementations packaged in separate jars works, but they can't be built separately from the main Dataverse code (since they use core classes like Dataverse, SolrSearchServiceBean, SettingsService, etc.). 

They way maven is comfigured now, the war doesn't include any new search services. To build the separate jars, you can call, for example,
`mvn clean package -DskipTests=true -Pexternal-search-get -Pexternal-search-post -Ptrivial-search-examples`

I'm not sure how/where the separate jars should be distributed (additional resources for the release?) Hopefully someone with more maven skills than me can separate jar builds more cleanly (even if we don't immediately drop references to internal Dataverse classes).

The SolrSearchServiceBean is @Stateless - as it has been. In early dev, I switched it to @Singleton since @Stateless is injected as a proxy (which implements SearchService but is not a SolrSearchServiceBean instance.) I think the current code is handling that OK.

**Suggestions on how to test this**: The main test would be to try this on the QA machine which is configured it to use Stefano's AI search. That can either be tested by API, soon via the SPA, or by setting the default to the postExternalSearch service. However, the latter, since it excludes collections and files, will mess up the JSF UI in ways that might be confusing. (If the SPA is deployed there, the UI should allow you to switch between installed engines.) 

Aside from validating that that works, some regression/performance testing should be done to verify the design doesn't break or slow current solr searches (shouldn't break, could slow).

**Additional documentation**: the api change is documented in the api guide, there's a new search-services developer page with details on building/configuring.


doc/release-notes/11281-ExternalSearch.md
```diff
@@ -0,0 +1,11 @@
...
+
+Dataverse now has an experimental capability to dynamically add and configure new search engines.
+The current Dataverse user interface can be configured to use a specified search engine instead of the built-in solr search.
+The search API now supports an optional &searchEngine query parameter that allows using any configured search engine.
+An additional /api/search/engines endpoint allows discovery of the engines installed.
+
+In addition to two trivial example engines designed for testing, Dataverse ships with two search engine classes that support calling an enternally-hosted search service (via HTTP GET or POST).
+These classes rely on the internal solr search to perform access-control and to format the final results, simplifying development of such an external engine.
+
+Details about the new functionality are described in
```
sekmiller: Do we want to put in a link here to the search-services documentation?

doc/sphinx-guides/source/developers/search-services.rst
```diff
@@ -0,0 +1,129 @@
...
+       
+       SolrQueryResponse search(DataverseRequest dataverseRequest, List<Dataverse> dataverses, String query,
+               List<String> filterQueries, String sortField, String sortOrder, int paginationStart,
+               boolean onlyDatatRelatedToMe, int numResultsPerPage, boolean retrieveEntities, String geoPoint,
+               String geoRadius, boolean addFacets, boolean addHighlights) throws SearchException;
+
+       default void setSolrSearchService(SearchService solrSearchService);
+   }
+
+The interface allows you to provide a service name and dsiplay name, and to respond to the same search parameters that are normally sent to the solr search engine.
```
sekmiller: typo: display

src/main/java/edu/harvard/iq/dataverse/DataversePage.java
```diff
@@ -89,8 +88,6 @@ public enum LinkMode {
     @EJB
     EjbDataverseEngine commandEngine;
     @EJB
-    SearchServiceBean searchService;
-    @EJB
     DatasetFieldServiceBean datasetFieldService;
```
sekmiller, line 91: Does the DataversePage need a SearchServiceFactory named "searchService"? OK, maybe not.

sekmiller:
Taking myself off here. I didn't find anything to object to, but I'd feel better if someone else took a look before passing to QA.

src/main/java/edu/harvard/iq/dataverse/DataversePage.java
qqmyers: Yeah - it's in the SearchIncludeFragment class.

doc/release-notes/11281-ExternalSearch.md
qqmyers: done

src/main/java/edu/harvard/iq/dataverse/search/GetExternalSearchServiceBean.java
```diff
@@ -0,0 +1,114 @@
...
+                .request(MediaType.APPLICATION_JSON).get();
+
+        if (response.getStatus() != 200) {
+            throw new SearchException("External search service returned status " + response.getStatus(), null);
+        }
+        try {
+            // Parse response and process results
+            String responseString = response.readEntity(String.class);
+            logger.finest("External search returned: " + responseString);
+            return postProcessResponse(responseString, dataverseRequest, retrieveEntities, addFacets, addHighlights);
```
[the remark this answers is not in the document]
qqmyers: fixed w/o a second try block

doc/release-notes/11281-ExternalSearch.md
```diff
@@ -0,0 +1,11 @@
+### Configurable Search Services
+
+Dataverse now has an experimental capability to dynamically add and configure new search engines.
+The current Dataverse user interface can be configured to use a specified search engine instead of the built-in solr search.
+The search API now supports an optional &searchEngine query parameter that allows using any configured search engine.
+An additional /api/search/engines endpoint allows discovery of the engines installed.
+
+In addition to two trivial example engines designed for testing, Dataverse ships with two search engine classes that support calling an enternally-hosted search service (via HTTP GET or POST).
```
pdurbin: Here's what I see out of the box at http://localhost:8080/api/search/services

```
{
  "status": "OK",
  "data": [
    {
      "name": "goldenOldies",
      "displayName": "Demo: Search Over First 1K Entries"
    },
    {
      "name": "solr",
      "displayName": "Dataverse Standard Search"
    },
    {
      "name": "oddlyEnough",
      "displayName": "Demo: Odd Results Only"
    }
  ]
}
```

I would prefer to only see only the "solr" service out of the box. I don't think the two toy examples should be activated by default.

doc/sphinx-guides/source/api/search.rst
```diff
@@ -29,16 +29,19 @@ type             string   Can be either "dataverse", "dataset", or "file". Multi
 subtree          string   The identifier of the Dataverse collection to which the search should be narrowed. The subtree of this Dataverse collection and all its children will be searched.  Multiple "subtree" parameters can be used to include multiple Dataverse collections. For example, https://demo.dataverse.org/api/search?q=data&subtree=birds&subtree=cats .
```
pdurbin, line 29: I made a PR to suggest some doc changes:

- https://github.com/GlobalDataverseCommunityConsortium/dataverse/pull/40

pdurbin (commented):
I'm still looking at code but here's some high level and doc feedback for now.

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceFactory.java
```diff
@@ -0,0 +1,110 @@
...
+            throw new IllegalArgumentException("Unknown search service: " + name);
+        }
+        service.setSolrSearchService(solrSearchService);
+        return service;
+    }
+
+    public SearchService getDefaultSearchService() {
+        String defaultService = JvmSettings.DEFAULT_SEARCH_SERVICE.lookupOptional().orElse("solr");
+        logger.log(Level.INFO, "Using default search service: {0}", defaultService);
+        return getSearchService(JvmSettings.DEFAULT_SEARCH_SERVICE.lookupOptional().orElse("solr"));
```
pdurbin: Can this be refactored to call `JvmSettings.DEFAULT_SEARCH_SERVICE.lookupOptional().orElse("solr")` only once?

Also, should `solr` be a static variable define somewhere? Maybe call it `fallbackSearchService` or something?

src/main/java/edu/harvard/iq/dataverse/search/AbstractExternalSearchServiceBean.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public void setSettingsService(SettingsServiceBean settingsService) {
+        this.settingsService = settingsService;
+    }
+
+    /**
+     * Creates a SolrQueryResponse object from the external search service response.
+     * The external service is expected to return a JSON object with the following
+     * structure: { "results": [ { "DOI": "doi:10.3886/ICPSR09083.v1", "Distance":
+     * 0.30227208137512207 },... ] }
```
pdurbin: Are only DOIs supported? What about Permalinks and Handles, etc.?

src/main/java/edu/harvard/iq/dataverse/api/Search.java
```diff
@@ -267,6 +282,24 @@ public Response search(
         }
     }
 
+    
+    @GET
+    @Path("/services")
+    public Response getSearchEngines() {
```
pdurbin, line 290: We get a dump of search engines/services like this:

```
{
  "status": "OK",
  "data": [
    {
      "name": "goldenOldies",
      "displayName": "Demo: Search Over First 1K Entries"
    },
    {
      "name": "solr",
      "displayName": "Dataverse Standard Search"
    },
    {
      "name": "oddlyEnough",
      "displayName": "Demo: Odd Results Only"
    }
  ]
}
```

It would be nice to also be able to see which one is the default.

src/main/java/edu/harvard/iq/dataverse/search/SearchService.java
```diff
@@ -0,0 +1,98 @@
...
+
+    String getServiceName();
+    String getDisplayName();
+    
+    /**
+     * Import note: "onlyDatatRelatedToMe" relies on filterQueries for providing
+     * access to Private Data for the correct user
+     *
+     * In other words "onlyDatatRelatedToMe", negates other filter Queries
+     * related to permissions
```
pdurbin:
```suggestion
     * Import note: "onlyDataRelatedToMe" relies on filterQueries for providing
     * access to Private Data for the correct user
     *
     * In other words "onlyDataRelatedToMe", negates other filter Queries
     * related to permissions

```

src/main/java/edu/harvard/iq/dataverse/search/SearchServiceFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        loadSearchServices();
+    }
+
+    private void loadSearchServices() {
+        // Load built-in services
+        Set<Bean<?>> beans = beanManager.getBeans(SearchService.class);
+        for (Bean<?> bean : beans) {
+            SearchService service = (SearchService) beanManager.getReference(bean, SearchService.class,
+                    beanManager.createCreationalContext(bean));
+            if ("solr".equals(service.getServiceName())) {
```
pdurbin: Elsewhere I suggested introducing a constant called something like `String fallbackSearchService = "solr"` which could be reused rather than hard coding "solr".

src/main/java/edu/harvard/iq/dataverse/settings/SettingsServiceBean.java
```diff
@@ -679,7 +679,15 @@ Whether Harvesting (OAI) service is enabled
...
-        ContactFeedbackMessageSizeLimit
+        ContactFeedbackMessageSizeLimit,
+        //Experimental setting to allow connecting to a GET external search service expecting a GET request with query parameter mirroring the search API query parameters (without search_service) 
+        GetExternalSearchUrl,
+        //Experimental setting to provide a display name for the GET external search service
+        GetExternalSearchName,
+        //Experimental setting to allow connecting to a POST external search service expecting a POST request with a JSON payload mirroring the search API query parameters 
+        PostExternalSearchUrl,
+        //Experimental setting to provide a display name for the POST external search service
+        PostExternalSearchName,
```
pdurbin, line 690: These database settings should be documented on config page I like I did for the JVM options in this PR:

- https://github.com/GlobalDataverseCommunityConsortium/dataverse/pull/40

That is, we can just list them, call them experimental, and link to the other page.

They should be in the release note too.

doc/release-notes/11281-ExternalSearch.md
qqmyers: The pom.xml (lines 870-875) excludes all of these services from the war (confirmed with `unzip -l dataverse-6.6.war`), so the only way they should appear is if you have a jar deployed and the directory configured.

pdurbin: I didn't do anything special to configure anything. I just ran the normal `mvn -Pct clean package docker:run` from https://guides.dataverse.org/en/6.6/developers/dev-environment.html#quickstart

src/main/java/edu/harvard/iq/dataverse/search/AbstractExternalSearchServiceBean.java
qqmyers: Added support for PID, keeping DOI as an alternate since that's what the currently deployed AI service uses

doc/release-notes/11281-ExternalSearch.md
qqmyers: Hmm - the docker build isn't using the war build where the exclusions are currently? Not sure I know how to fix the docker build.

pdurbin: @poikilotherm I'm not sure how busy you'll be tomorrow (I have the day off) but if you have time to take a look, please do! ❤️

pdurbin: If you do just `mvn -Pct clean package` (without the `docker:run`) you'll find a jar file at target/dataverse-6.6.jar. This is what Docker operates on.

If you peek inside, you'll find the "oldies" and "oddly" classes:

```
% jar -tf target/dataverse-6.6.jar | grep -i oldies
edu/harvard/iq/dataverse/search/GoldenOldiesSearchServiceBean.class
% jar -tf target/dataverse-6.6.jar | grep -i oddly 
edu/harvard/iq/dataverse/search/OddlyEnoughSearchServiceBean.class
```

Is this the problem? Should those classes be excluded from the jar file used by Docker?

pdurbin: I started a thread on Zulip about this as well: https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/search.20services.3A.20goldenOldies.2C.20oddlyEnough/near/525912106

pdurbin: I just found this:

"packagingExcludes applies to war archive. It has no impact to "exploded" war." -- https://stackoverflow.com/questions/36644308/packagingexcludes-not-working-in-maven-war-plugin/36646805#36646805

pdurbin: I'm looking at https://stackoverflow.com/questions/13229956/how-to-exclude-files-from-being-copied-by-maven-into-the-exploded-war and trying to get `warSourceExcludes` to work but no luck so far. (This is under the `maven-war-plugin` config under the "ct" profile.)

pdurbin: @qqmyers as discussed in standup, maybe you can add a flag of some sort to prevent the example search services ("oldies" and "oddly") from loading? I guess a feature flag could be used. 🤷

pdurbin: @qqmyers Oliver is typing on mobile some ideas here: https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/search.20services.3A.20goldenOldies.2C.20oddlyEnough/near/525912106

poikilotherm: Some remarks.

1. Containers use the exploded WAR, not the JAR.
2. We can always exclude stuff from being copied via assembly.xml in src/main/docker or even use dockerignore files. (Assembly may be cleaner and potentially reuse the list)
3. All if these exclusions become much easier when putting them in a subpackage. Globbing is way more fun!
4. Shouldn't these go in a separate Maven module? (Eventually?)
5. I love and appreciate the idea of using a feature  flag to make enabling the external search an informed admin choice.

pdurbin: I played around with this:

```
% git diff
diff --git a/pom.xml b/pom.xml
index 87fbb5f066..d9bfebf85b 100644
--- a/pom.xml
+++ b/pom.xml
@@ -867,6 +867,7 @@
                             <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
                         </manifest>
                     </archive>
+                    <!-- Keep these excludes in sync with the ones in src/main/docker/assembly.xml. -->
                     <packagingExcludes>
                         WEB-INF/classes/edu/harvard/iq/dataverse/search/GetExternalSearchServiceBean.class,
                         WEB-INF/classes/edu/harvard/iq/dataverse/search/PostExternalSearchServiceBean.class,
diff --git a/src/main/docker/assembly.xml b/src/main/docker/assembly.xml
index 62cd910ef9..819e480666 100644
--- a/src/main/docker/assembly.xml
+++ b/src/main/docker/assembly.xml
@@ -7,6 +7,11 @@
             <outputDirectory>app</outputDirectory>
             <excludes>
                 <exclude>WEB-INF/lib/**/*</exclude>
+                <!-- Keep these excludes in sync with packagingExcludes in the main pom.xml. -->
+                <exclude>WEB-INF/classes/edu/harvard/iq/dataverse/search/GetExternalSearchServiceBean.class</exclude>
+                <exclude>WEB-INF/classes/edu/harvard/iq/dataverse/search/PostExternalSearchServiceBean.class</exclude>
+                <exclude>WEB-INF/classes/edu/harvard/iq/dataverse/search/GoldenOldiesSearchServiceBean.class</exclude>
+                <exclude>WEB-INF/classes/edu/harvard/iq/dataverse/search/OddlyEnoughSearchServiceBean.class</exclude>
             </excludes>
         </fileSet>
```

Then I ran `mvn -Pct clean package docker:run` as usual.

`GoldenOldiesSearchServiceBean.class` (for example) is now absent from the `target/docker/gdcc/dataverse/unstable/build/maven/app/WEB-INF/classes/edu/harvard/iq/dataverse/search` directory.

However, "goldenOldies" still active in the app. I can see it at http://localhost:8080/api/search/services like this:

```
{
  "status": "OK",
  "data": {
    "services": [
      {
        "name": "goldenOldies",
        "displayName": "Demo: Search Over First 1K Entries"
      },
      {
        "name": "solr",
        "displayName": "Dataverse Standard Search"
      },
      {
        "name": "oddlyEnough",
        "displayName": "Demo: Odd Results Only"
      }
    ],
    "defaultService": "solr"
  }
}
```

src/main/java/edu/harvard/iq/dataverse/api/Search.java
```diff
@@ -268,6 +284,28 @@ public Response search(
...
+        Map<String, SearchService> availableEngines = searchServiceFactory.getAvailableServices();
+        String defaultServiceName = JvmSettings.DEFAULT_SEARCH_SERVICE.lookupOptional().orElse("solr");
```
pdurbin: Instead of hard-coding "solr" here, should we use the new INTERNAL_SOLR_SERVICE_NAME variable? We'd have to make it public.

src/main/java/edu/harvard/iq/dataverse/search/SolrSearchServiceBean.java
```diff
@@ -82,70 +82,16 @@ public class SearchServiceBean {
...
-     * @param numResultsPerPage
-     * @return
-     * @throws SearchException
-     */
-    public SolrQueryResponse search(DataverseRequest dataverseRequest, List<Dataverse> dataverses, String query, List<String> filterQueries, String sortField, String sortOrder, int paginationStart, boolean onlyDatatRelatedToMe, int numResultsPerPage) throws SearchException {
-        return search(dataverseRequest, dataverses, query, filterQueries, sortField, sortOrder, paginationStart, onlyDatatRelatedToMe, numResultsPerPage, true, null, null);
+    
+    @Override
+    public String getServiceName() {
+        return "solr";
```
pdurbin: Should we use the new SearchServiceFactory.INTERNAL_SOLR_SERVICE_NAME variable instead of hard-coding "solr" here? We'd have to make it public.

pdurbin (commented):
I found a couple more hard-coded "solr" instances.

Also, I played around with trying to keep the "oldies" and "oddly" stuff out of the production Docker app. See https://github.com/IQSS/dataverse/pull/11281#discussion_r2164017028

doc/release-notes/11281-ExternalSearch.md
qqmyers: I've made some changes like the one above in assembly.xml that appear to drop the test classes from the various intermediate exploded war and even the docker-build.tar file, but I continued to see the test classes via the /search/services api. Since the intent was to not have any such classes in the Dataverse release itself, I've just gone ahead and skipped loading any internal search services except for the solr service itself. This is probably a reasonable permanent fix unless/until we want to distribute more than one search service out-of-the-box and is independent of any future refactoring to actually remove the test classes from the docker build as is done with the war / to separate building the search service jars, etc.

I think that means this can go forward into ready for QA?

pdurbin: Yes, I'm only seeing the "solr" service now, out of the box:

```
{
  "status": "OK",
  "data": {
    "services": [
      {
        "name": "solr",
        "displayName": "Dataverse Standard Search"
      }
    ],
    "defaultService": "solr"
  }
}
```

As you're saying the addition to assembly.xml in d49256ca30 isn't actually having any effect but I guess we can leave it in there. I added comments to the two places where the lists are maintained in ec348a743f.

Also in 965ee18b95 I fixed the indenting around that last code change you made.

poikilotherm: I ran this locally but *with* Jim's fix. Works! Do y'all want me to try again when patching out the fix?

Just did it: I can reproduce that the services are shown in "api/search/services". Looking into the running container I see that the class files are present. So no wonder why they are loaded... :wink:

poikilotherm: I know why this is happening during testing with the usual container dev env. https://github.com/poikilotherm/dataverse/blob/8ce370c0aa37e422a58ae464da0671c46d3e82f6/docker-compose-dev.yml#L77-L77 :facepalm:

poikilotherm: I can verify that removing the mount means it does not appear in the running app. So anyone starting Dataverse with just the container and not mounting the class files doesn't see them. Please decide whether we want the fix or not. We can't exclude them in the WAR plugin as already mentioned very easily. Maybe the mid term goal of moving them to a proper place is best and a workaround is fine for now. The workaround should probably have a TODO attached, so we remove it at some point.

ofahimIQSS:
Testing passed - merging

PR #11286, opened 2025-02-26
clarify direct children only when using returnChildCount=true

pdurbin:

Just a doc update to clarify that the following PR is for direct children only:

- #11259


sekmiller (approved):

ofahimIQSS:
proofread - merging PR, no issues found.

PR #11163, opened 2025-01-16
CSL Citation support

qqmyers:
**What this PR does / why we need it**: This PR adds support for viewing/copying dataset citations in any of the journal/society citation formats described in the [Citation Style Language](https://citationstyles.org/) like Zotero, Mendeley, and others. (Strictly just the 1000K+ "independent" ones as coded, but that could be changed.)

**Which issue(s) this PR closes**:
It's related to #10512 in that it provides a copy-to-clipboard option for retrieving a citation, but it does not add a copy button for the main (custom) citation displayed on the dataset page.

It's also relevant w.r.t. #9994. It doesn't adopt the design discussed there of making the EndNote/RIS/BibTeX formats into exports (which are currently only available for the latest version).

- Closes #

**Special notes for your reviewer**: The PR uses a combination of Java and Javascript libraries that have overlapping functionality. As coded, the Java library is only used to create the ~native CSLJson format which is then sent to the front end where Javascript reads that and generates any of the available formats. I'm hoping this makes more sense for the SPA than doing everything in the back-end Java. The PR includes an `/api/pids/<pid>/csl` api call that will return the CSLJson required by the front-end Javascript.

The code also leverages DataCite's API to get their version of CSL for DataCite DOIs for the latest published version of the dataset. This should assure that Dataverse's output would match that from those generated externally from the DOI. For drafts and older versions, and for other PID types, the Java library noted above is used to generate the CSLJson required. This is a compromise but gives us the opportunity to see if/when DataCite's mapping to CSLJson differs from our own.

The PR includes an API call as well, which is hopefully useful for the SPA - as a way to get the EndNote/RIS/BibTeX formats (and the Internal one without the JSON wrapper if that's helpful) as well as to get the CSLJson which could then be transformed by the JavaScript library used here or others to generate all the other styled formats. The one thing that this PR does not do that is supported by the current download mechanism is assigning a filename to the download. That shouldn't be hard to do if it's useful, but would take a little refactoring.

**Suggestions on how to test this**: Create a Dataset with lots of metadata and verify that the popup allows you to generate an copy any format you select. 

Set the dataverse.csl.common-styles JVM option to a comma-separated list of formats other than "chicago-author-date, ieee" and verify that the entries for common styles in the dialog match.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
image
image
image

**Is there a release notes update needed for this change?**: added

**Additional documentation**: UI, API, and setting documented.

PR #11313, opened 2025-03-06
DANS/CSL - Fix create dataset bug

qqmyers:
**What this PR does / why we need it**: Late changes to #11163 led to a failure when trying to create a new dataset in the JSF UI as reported in slack (see https://iqss.slack.com/archives/C04MJ76A34Z/p1741295920525239 ). In brief, code to get the JSON needed to generate different citation styles was moved to where it would run in the dataset page create mode. That failed because the new DatasetType is null at this point. A fix is to put the relevant code in a ui:fragment that is only rendered when editMode is not CREATE. That's what this PR does.

- Closes #

**Special notes for your reviewer**: Using https://github.com/IQSS/dataverse/pull/11313/files?w=1 you can see the only change is the open/close for the ui:fragment.

**Suggestions on how to test this**: The only change is in the dataset.xhtml, so verifying the you can create a dataset and the new citation styles still work for an existing dataset should do it.

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dataset.xhtml
+++ b/src/main/webapp/dataset.xhtml
@@ -2094,55 +2094,57 @@
                             </div>
                         </p:fragment>
                     </p:dialog>
-                    <script src="https://cdn.jsdelivr.net/npm/citation-js"
-                        type="text/javascript"></script>
-                    <script>
-                        const Cite = require('citation-js');
-                        let cite = new Cite(#{version.getCitation("CSL", true, false)});
-                    </script>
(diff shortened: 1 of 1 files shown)


pdurbin (approved):
The test run at https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11313/1/console failed with "Connection to ec2-44-202-163-64.compute-1.amazonaws.com closed by remote host" but this happens intermittently sometimes. I kicked off another run but as this PR only changes xhtml it shouldn't affect the API tests.

I didn't test this PR but the change seems reasonable. I'm sending this to QA.

PR #11319, opened 2025-03-07
MailTest Fix

### LINKED ISSUE #11318: Integration Test Failure After Payara Upgrade (Stable / JDK 17) - MailSessionProducerIT
Description:

After merging the Payara upgrade PR, I noticed that the Maven Tests for Stable / JDK 17 Integration Tests are failing. The test failure is related to the setup of a fake email server in a Docker container, which verifies if emails are being sent.

Comment from PR:

Don left a comment on the PR: [link to PR comment](https://github.com/IQSS/dataverse/pull/11128#issuecomment-2704717942).
Jim mentioned that the failure might be due to running the branch against an earlier Payara version, causing a version mismatch between the Payara upgrade and the test environment.
Jim also suggested that since the test checks email sending functionality, we should verify if emails are actually being sent in deployment. If emails are sent successfully, the failure might be attributed to a flaky test.

Steps to Reproduce:
Merge the Payara upgrade PR.
Run the Maven Tests for Stable / JDK 17 Integration Tests.
Observe the failing tests related to the fake email server setup in the Docker container.

Expected Result:
The integration tests should pass, verifying that emails are sent correctly via the fake email server.

Actual Result:
The integration tests are failing, possibly due to a mismatch in Payara versions or a flaky test.

Additional Information:
Jim's suggestion: Confirm whether emails are being sent in the deployment, as it might help confirm whether the test is flaky or if there's an actual issue.
The test might also fail in Payara version 6.2025.2, which could indicate a compatibility issue.

Action Required:
Verify if emails are being sent in the deployment.
Investigate if the test failure is caused by a flaky test.
Decide how to handle the flaky test going forward, especially considering the potential mismatch with Payara versions.

qqmyers:
**What this PR does / why we need it**: Somewhere in the update it looks like the test certificate is now being checked to make sure the common name matches the host. In the test, that is "localhost". The original cert had "example.org". This PR creates a new cert/key pair that has localhost as the common name and uses a longer key (based off an earlier guess that short keys might also cause failures in the next 10 years of the cert's life.

- Closes #11318 

**Suggestions on how to test this**: Verify the test passes

**Additional documentation**:


pdurbin (approved):
The tests in maven_unit_test.yml are passing now. Nice comment too. Thanks, @qqmyers! Approved! 🚀

poikilotherm:
@qqmyers thanks for looking into this! You sparked my interest on a late evening to look into this in my spare time. 

I was very curious why the RSA key length would be the crucial thing to fix here, so I tried to find changes in Eclipse Angus Mail to corroborate that.

Turns out when I added the certificate back in the day I made a rookie mistake by using the wrong `CN` :roll_eyes: So I went ahead and fixed that, also adding a OpenSSL config file to make the creation process less manual. I hope you don't mind me being a bit pushy with forcing the changes into your branch.

ofahimIQSS:
Tests Passed - Merging PR.

PR #11326, opened 2025-03-10
DANS CSL - disable dialog in any edit mode

qqmyers:
**What this PR does / why we need it**: In #11313 a change was made to not initialize the CSL related dialog/JavaScripts when creating a dataset which caused a failure. With that fix, it appears that the dialog still gets loaded in edit modes when it isn't needed. While this isn't fatal w.r.t. generating the UI, the scripts do throw a warning in the console:
`Uncaught SyntaxError: Failed to execute 'appendChild' on 'Node': Identifier 'Cite' has already been declared`
This PR just broadens the render condition for this part of the code to match that used for the top part of the dataset page (where the citation displays), i.e. #{empty DatasetPage.editMode}

- Closes #

**Special notes for your reviewer**: Simple fix - nice to have in 6.6 since the CSL functionality is new. Shouldn't change performance.

**Suggestions on how to test this**: Look in browser console with CSL configured and see that the error above disappears when editing metadata. Verify CSL styles can still be generated in the dialog.

**Additional documentation**:


ofahimIQSS:
Testing Passed - Merging PR

PR #11329, opened 2025-03-11
Dashboard: fix npe from breadcrumbs

### LINKED ISSUE #11328: Internal Server Error on Dashboard Links
Issue Summary
Clicking on any link on the Dashboard page results in an Internal Server Error.

Steps to Reproduce

[RegTest - Server.log.txt](https://github.com/user-attachments/files/19185979/RegTest.-.Server.log.txt)
https://github.com/user-attachments/assets/8662d0cd-63cd-4cee-b804-fddbcf8b9f97

Expected Behavior
The selected page should load successfully without errors.

Actual Behavior
Clicking on any link within the Dashboard results in an Internal Server Error, preventing access to related sections.

qqmyers:
**What this PR does / why we need it**: Fixes the dashboard problem

- Closes #11328 

**Special notes for your reviewer**: As far as I can tell, the dataverseId parameter does nothing and the dashboard never uses anything but the root dataverse id (in fact the code hardcodes the dataverseId view param to the root dataverse id regardless of what is sent in the URL). Seems like this could all be removed, but to avoid new issues for 6.6, I've just replaced the null value with param.dataverseId if sent and removed the dataverseId param completely if it isn't.

**Suggestions on how to test this**: Try all the buttons on the dashboard. 

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dashboard-movedataset.xhtml
+++ b/src/main/webapp/dashboard-movedataset.xhtml
@@ -15,7 +15,7 @@
             <f:viewAction action="#{dataverseSession.updateLocaleInViewRoot}"/>
             <f:viewAction action="#{DashboardMoveDatasetPage.init}"/>
             <f:viewAction action="#{dataverseHeaderFragment.initBreadcrumbs(dataverseServiceBean.findRootDataverse())}"/>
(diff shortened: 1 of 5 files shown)


pdurbin:
This fix is required due to the Payara version bump we made:

- #11128

stevenwinship (approved):

ofahimIQSS:
Tests are passing - verified all Dashboard links are now working. Merging PR.
image

PR #11347, opened 2025-03-14
return all type totals regardless of types being returned

### LINKED ISSUE #11280: Search API - Count per object
Navigate to https://beta.dataverse.org/spa/collections and open browser dev tools, select the Network tab and filter requests by the word 'search'.

* What happens?
The Search API count per object property is returning as 0 the types that are not sent through the endpoint as `type`query parameters.
For example for this API call:
```
https://beta.dataverse.org/api/v1/search?q=*&show_facets=true&sort=date&order=desc&show_type_counts=true&subtree=root&per_page=10&type=dataverse&type=dataset`
```
the `total_count_per_object_types` is  returned as:
```
{
    "Datasets": 60,
    "Dataverses": 40,
    "Files": 0
}
```

The Files count should not be returned as 0, but as the correct value of existing files even if the query param `type=file` is not sent, this is to replicate JSF counting behaviour in the collection page.

All SPA Users and API Users

The API to return the counting on all types even if the `type` query parameter for that type was not sent.

unstable

stevenwinship:
**What this PR does / why we need it**: All totals counts should be returned regardless of the "type"(s) being requested in the filter

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11280

- Closes #11280

**Suggestions on how to test this**: make various search api calls. See SearchIT tests

**Additional documentation**:


src/test/java/edu/harvard/iq/dataverse/api/SearchIT.java
```diff
@@ -1891,4 +1920,47 @@ public void testTabularFiles() throws IOException {
...
+
+        Integer datasetId = UtilIT.getDatasetIdFromResponse(createDataset);
+        String datasetPid = UtilIT.getDatasetPersistentIdFromResponse(createDataset);
+
+        Path pathToDataFile = Paths.get(java.nio.file.Files.createTempDirectory(null) + File.separator + "data.csv");
+        String contentOfCsv = ""
+                + "name,pounds,species,treats\n"
+                + "Midnight,15,dog,milkbones\n"
+                + "Tiger,17,cat,cat grass\n"
+                + "Panther,21,cat,cat nip\n";
```
sekmiller, line 1952: cute

sekmiller (approved):

ofahimIQSS:
tests passing, merging PR

PR #11362, opened 2025-03-21
Change to duplicate role assignment error message

### LINKED ISSUE #11191: Collection Level "Assign role" API error message should specify: User already has this role, and not include "dataset" (which is incorrect)
When using this [Native AP](https://guides.dataverse.org/en/6.5/api/native-api.html#assign-a-new-role-on-a-dataverse-collection)

If the user already has the role, the error message reads "{"status":"ERROR","message":"User already has this role for this **dataset**"}% instead of **"collection"**

* When does this issue occur?
When a user already has the role type in question

* What happens?

superuser, curators, anyone with permission to run this script on a collection level

correct error message: "User already has this role" remove "datasets" and no need to mention dataverse or collection

most recent

NO

**Screenshots:**

stevenwinship:
**What this PR does / why we need it**: Error message is specific to user and dataverse even though it is also used for groups and datasets. Message needs to be made more generic.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11191

- Closes #11191 

**Suggestions on how to test this**: See IT tests

**Additional documentation**:


pdurbin (approved):
I didn't run the code myself but the fix is straighforward and tests are passing: https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11362/1/testReport/

Approved.

PR #11378, opened 2025-03-27
EditDDI API empty frequency bug

### LINKED ISSUE #11377: EditDDI API empty frequency bug
In EditDDI API, ``curl -H "X-Dataverse-key:$API_TOKEN" -X PUT "$SERVER_URL/api/edit/$ID" --upload-file $FILE``
if in xml file variable has a category with declared frequency (weighted or unweighted) but frequency number is missing, the api will work and return 200 status, but export DDI function would be broken, and DDI xml would not be produced because frequency would be null instead of a real number. For example:
``
    <catgry>
      <catValu>1</catValu>
      <labl level="category">Under 18</labl>
      <catStat type="freq">2</catStat>
      <catStat wgtd="wgtd" type="freq"/>
    </catgry>
    <catgry>
``
Here frequency is defined but does not have value ``<catStat wgtd="wgtd" type="freq"/>``

* When does this issue occur?
In this case Data Explorer also would not work since Dataverse would not be able to produce DDI xml file

If frequency is declared but missing the edit api should return not acceptance error (406) of bad xml input file

Yes

lubitchv:
In EditDDI api when frequency is declared in input xml file but missing the api returns 200 code, but this breaks DDI exporter since frequency is declared but missing (null).
For that case in this PR EditDDI api returns 400 code of bad request of bad xml file. Integration test for that case was updated.

- Closes #11377 

Upload spss or STATA file with categories. Download Variable Metadata xml for that file. Edit xml by removing frequency number (empty frequency). Use ``curl -H "X-Dataverse-key:$API_TOKEN" -X PUT "$SERVER_URL/api/edit/$ID" --upload-file $FILE`` to update the var metedata, where $ID is file id and $File is updated xml. One should get 400 error.


src/main/java/edu/harvard/iq/dataverse/api/EditDDI.java
```diff
@@ -110,12 +110,16 @@ public Response edit(@Context ContainerRequestContext crc, InputStream body, @Pa
         Map<Long, VariableMetadata> mapVarToVarMet = new HashMap<Long, VariableMetadata>();
         Map<Long,VarGroup> varGroupMap = new HashMap<Long, VarGroup>();
         try {
-            readXML(body, mapVarToVarMet,varGroupMap);
+            if (!readXML(body, mapVarToVarMet, varGroupMap)) {
+                return error(Response.Status.BAD_REQUEST, "bad xml file");
```
qqmyers, line 114: to help with debugging, could the error text for this be different from when there's an XML exception? Is this error "xml missing fields" or "xml isn't valid DDI" or ?

lubitchv: @qqmyers I updated as you suggested. Now it catches the original exception and returns 400 error with a message  frequency or weighted frequency being declared but missing.

lubitchv:
I updated the branch from IQSS develop, no other changes. But there seems to be a problem with the test RemoteOverlayAccessIOTest.testRemoteOverlayFiles The problem is with https://data.qdr.syr.edu/resources/images/dataverse_progect_logo.svg Replacing https://data.qdr.syr.edu/resources with https://demo.dataverse.org/resources seems to solve the problem. The change was made here https://github.com/IQSS/dataverse/commit/3674a6f86c9df19ae0aac57ec4e3ebcee513fd01

ofahimIQSS:
looks like continuous integration is failing still

lubitchv:
@ofahimIQSS Which test is failing? Is it me or does it fail for everyone?

qqmyers:
There was a problem in the develop branch - if you merge with the latest the test should pass.

ofahimIQSS:
looks good, merging

PR #11388, opened 2025-04-02
add if workflows is IQSS running

### LINKED ISSUE #11386: Workflows deploy Beta WAR file scp error
When you fork the dataverse repo and then perform an update from the original repo in "develop," it always results in an error in the GitHub workflow. A solution would be to add ```if: ${{ github.repository_owner == 'IQSS' }}``` in the "deploy_beta_testing.yml" file in line 47. This would solve the problem. Perhaps there are other ways to do it, but this would work for all forks. Only those who have a beta version would face an issue, as it would no longer be executed.

* What happens?
workflow fail in github

BenediktMeierUIT:
Only affects forked repos.

- Closes #11386

However, it may be the case that if you have your own beta environments, they may no longer work. Therefore, something else would need to be done with a variable or something like that.

When you create a fork and then make an update in "develop," no error message appears anymore.

### DIFF
--- a/.github/workflows/deploy_beta_testing.yml
+++ b/.github/workflows/deploy_beta_testing.yml
@@ -43,6 +43,7 @@ jobs:
 
   deploy-to-payara:
     needs: build
+    if: ${{ github.repository_owner == 'IQSS' }}
     runs-on: ubuntu-latest
 
     steps:


qqmyers (approved):
Let's try it - easy to make sure beta still deploys at IQSS

ofahimIQSS:
Copy Labels from issue to pull request is failing on this.

ofahimIQSS:
As discussed during stand up, we will create a separate PR to disable running the "Copy Labels" job for external PRs going forward. 

When I push the PR to internal, I'm getting a 500 error page. Might need to update this PR with latest from dev branch. 

image

qqmyers:
FWIW: This PR doesn't have any code changes to test. It just tries to stop the deploy to beta action when run somewhere other than IQSS. The only thing you could try to confirm is that the deploy to beta works when you merge.

ofahimIQSS:
Thanks @qqmyers ! Merging and observing.

PR #11406, opened 2025-04-08
500->409 for concurrent updates

jo-pol:

With concurrent API calls to `Datasets.createAssignment` you could get a response code 500 "transaction aborted" with details only in the server logs. This PR changes the response into: code 409 (conflict) "User already has this role for this dataset".

- Closes #

Run a script with at least two threads. Each thread executes a single loop that adds an deletes the same role-assignment.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -66,6 +66,7 @@ import jakarta.ws.rs.*;
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.core.*;
 import jakarta.ws.rs.core.Response.Status;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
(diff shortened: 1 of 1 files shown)


sekmiller:
Hi @jo-pol, can you take a look at/resolve the merge conflicts on this branch?  Thanks!

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -66,6 +66,7 @@
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.core.*;
 import jakarta.ws.rs.core.Response.Status;
+import org.apache.commons.lang.exception.ExceptionUtils;
```
stevenwinship, line 69:
```suggestion
import org.apache.commons.lang3.exception.ExceptionUtils;
```

stevenwinship:
@jo-pol It looks like you may need to merge again.

stevenwinship (approved):

stevenwinship:
2 tests in DatasetsIT are still failing. Please fix these and re-submit for review

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2086,6 +2087,10 @@ public Response createAssignment(@Context ContainerRequestContext crc, RoleAssig
             return ok(
                     json(execCommand(new AssignRoleCommand(assignee, theRole, dataset, createDataverseRequest(getRequestUser(crc)), privateUrlToken))));
         } catch (WrappedResponse ex) {
+            if (ExceptionUtils.getRootCause(ex).getMessage().contains("duplicate key")) {
```
qqmyers: @jo-pol - looks like one of the exceptions getting to here has a null message, so the getMessage().contains fails. If you check that getMessage()!=null first, it should fix the test.

pdurbin:
This PR has merge conflicts.

stevenwinship:
I tried to fix the merge conflicts but I don't have permission to push the fix to this PR

ofahimIQSS:
looks good - merging

PR #11407, opened 2025-04-08
Prevent needless optimistic lock exception for role assignment removal

jo-pol:

It prevents optimistic lock exception for removal of role-assignments. There is no conflict possible with other changes to a dataset.

- Closes #

Tabular ingest must be switched on

    curl -X PUT -d 2000000000 http://localhost:8080/api/admin/settings/:TabularIngestSizeLimit

Run a script with an API call that uploads a  [zip](https://github.com/user-attachments/files/19648533/50.zip) with many csv files and subsequently loops (before the ingest lock is released) over API calls that add and delete of a role-assignment on the same dataset.

**Additional documentation**:

See also https://github.com/IQSS/dataverse/pull/11406 and the DANS issue https://drivenbydata.atlassian.net/browse/DD-1818


qqmyers (changes requested):
@jo-pol - this PR currently does nothing - just spacing changes. Should the change be to `@ManyToOne` - without the merge cascade?

ofahimIQSS:
looks good from my end - merging PR

PR #11411, opened 2025-04-08
11391 display on create with template

### LINKED ISSUE #11391: Fields with displayOnCreate setting not visible if collection has a default template
1. Add one metadata block with at least one block to a Dataverse collection.
2. Set `displayOnCreate` to `true` for a metadata field at input level using the API, e.g., `geographicUnit`.
3. Confirm that the field appears when creating a dataset in the collection (as expected).
4. Now add a dataset template to the collection and set it as default.
5. Try to create a new dataset again and the field is no longer shown during dataset creation.

It occurs after setting a default template for a Dataverse collection, even if the template does not override any metadata field settings. Or by modifying the `displayOnCreate` at input level property of a collection that already has a default template.

**Which page(s) does it occur on?**
During dataset creation, the affected fields are no longer visible.

**What happens?**
The metadata fields configured with `"displayOnCreate": true` at input level stop being displayed on the dataset creation form if the collection has a default template.

All users creating datasets in the collection.

Metadata fields with `"displayOnCreate": true` at input level should always appear on the creation form, regardless of whether a template is set or not.

v6.6

New feature introduced in https://github.com/IQSS/dataverse/pull/11224, https://github.com/IQSS/dataverse/pull/11312 and https://github.com/IQSS/dataverse/pull/11343

sekmiller:
**What this PR does / why we need it**: Makes fields set "display on Create" at the Dataverse level visible on dataset create in the presence of a default template for that dataset (especially if the template does not have default values for that Metadata block)

- Closes #11391 

**Special notes for your reviewer**: Not sure why the "isDisplayOnCreate" in the Metadata block doesn't work. I added a transient boolean that is updated in the DatasetVersionUI when the metadata blocks are added.

Set dataset field types outside of the citation block to "displayOnCreate" for a given Dataverse - 
```
curl -X PUT -H "X-Dataverse-key:27987dd2-8ae1-4b6d-b3f8-78a2ccfcad4d" -H "Content-Type:application/json" "http://localhost:8080/api/dataverses/temp-doc/inputLevels" -d '[{"datasetFieldTypeName":"geographicCoverage", "required":false, "include":true,  "displayOnCreate":true}, {"datasetFieldTypeName":"country", "required":false, "include":true, "displayOnCreate":true}]'
``` 

Add a template to the dataset, but don't set default values in the metadata block where you set "displayOnCreate". Make that template the default.

Add a dataset - the dataset fields set as display on create should be available for edit.

Also, you can try it with no default template but select one after "add dataset" - as above, the dataset fields set as display on create should be available for edit.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
@@ -458,6 +458,10 @@ public class DatasetVersionUI implements Serializable {
                         mdb.setEmpty(false);
                         datasetFieldsForView.add(dsf);
                     }
+                    //Seeting Local Display on Create on mdb when there are any set at dataverse level 
+                    if ( dsf.getDatasetFieldType().getLocalDisplayOnCreate() != null ){
+                       mdb.setLocalDisplayOnCreate(dsf.getDatasetFieldType().getLocalDisplayOnCreate());
+                    }
                 }
             }
 
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
@@ -458,7 +458,7 @@ public class DatasetVersionUI implements Serializable {
                         mdb.setEmpty(false);
                         datasetFieldsForView.add(dsf);
                     }
-                    //Seeting Local Display on Create on mdb when there are any set at dataverse level 
+                    //Setting Local Display on Create on mdb when there are any set at dataverse level 
                     if ( dsf.getDatasetFieldType().getLocalDisplayOnCreate() != null ){
                        mdb.setLocalDisplayOnCreate(dsf.getDatasetFieldType().getLocalDisplayOnCreate());
                     }
--- a/src/main/java/edu/harvard/iq/dataverse/MetadataBlock.java
+++ b/src/main/java/edu/harvard/iq/dataverse/MetadataBlock.java
@@ -101,13 +101,29 @@ public class MetadataBlock implements Serializable, Comparable {
     }
     
     public boolean isDisplayOnCreate() {
+        // relying on "should" doesn't seem to work in context of a template 
+        // adding a transient that is updated in the DatasetVersionUI to fix
         for (DatasetFieldType dsfType : datasetFieldTypes) {
             boolean shouldDisplayOnCreate = dsfType.shouldDisplayOnCreate();
             if (shouldDisplayOnCreate) {
                 return true;
             }
         }
+        if (getLocalDisplayOnCreate() != null){
+            return getLocalDisplayOnCreate();
+        }
         return false;
+    }
+    
+    @Transient
+    private Boolean localDisplayOnCreate;
+
+    public Boolean getLocalDisplayOnCreate() {
+        return localDisplayOnCreate;
+    }
+
+    public void setLocalDisplayOnCreate(Boolean localDisplayOnCreate) {
+        this.localDisplayOnCreate = localDisplayOnCreate;
     }
 
     public String getDisplayName() {
--- a/src/main/java/edu/harvard/iq/dataverse/ManageTemplatesPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/ManageTemplatesPage.java
@@ -121,7 +121,7 @@ public class ManageTemplatesPage implements java.io.Serializable {
         Template newOne = templateIn.cloneNewTemplate(templateIn);
         String name = BundleUtil.getStringFromBundle("page.copy") +" " + templateIn.getName();
         newOne.setName(name);
-        newOne.setUsageCount(new Long(0));
+        newOne.setUsageCount(Long.valueOf(0));
         newOne.setCreateTime(new Timestamp(new Date().getTime()));
         newOne.setDataverse(dataverse);
 


src/main/java/edu/harvard/iq/dataverse/MetadataBlock.java
```diff
@@ -101,14 +101,30 @@ public void setDatasetFieldTypes(List<DatasetFieldType> datasetFieldTypes) {
     }
     
     public boolean isDisplayOnCreate() {
+        // relying on "should" doesn't seem to work in context of a template
```
qqmyers, line 104: should relies on a transient datsetfieldtype localdisplayoncreate that must not be getting getting set in the DatasetVersionUI init method. If it were, it think it should work. Is it just because of the order of these 2 lines: https://github.com/IQSS/dataverse/blob/039b3b7266ad76b56b550d0eb3316c78f84d7125/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L2135-L2136 - the updateDatasetFieldInputLevels sets localdisplayOncreate, right after the dvUI is inited. (Same ordering of these two calls in the DatasetPage.edit method.)

src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
```diff
@@ -458,6 +458,10 @@ public void setMetadataValueBlocks(DatasetVersion datasetVersion) {
                         mdb.setEmpty(false);
                         datasetFieldsForView.add(dsf);
                     }
+                    //Setting Local Display on Create on mdb when there are any set at dataverse level 
+                    if ( dsf.getDatasetFieldType().getLocalDisplayOnCreate() != null ){
+                       mdb.setLocalDisplayOnCreate(dsf.getDatasetFieldType().getLocalDisplayOnCreate());
```
qqmyers, line 463: If there's one with displayoncreate true and a later one with false, this sets the block displayonCreate back to false? 

Also not seeing how the fieldtype localdisplayoncreate can be set at this point (versus always being null).

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
```diff
@@ -464,11 +466,50 @@ public boolean isRequired() {
...
+                        if (dftilTest.isRequired() && !child.isRequired()) {
+                            hasRequiredChildrenDV = true;
+                        }
+                    }
+                }
+            }
+        }
+        return hasRequiredChildrenDV;
+    }
+
```
qqmyers, line 512: Aside from lines 518/519, it looks like this method and the next one are the same. Is there a way to consolidate?

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
```diff
@@ -464,11 +466,50 @@ public boolean isRequired() {
             {
                 required = false;
             }
+
```
qqmyers, line 469: are lines 465-468 needed with the new code below? Seems like if a compound field is required for some reason - required is true at line ~462 but would be set false if there are no required children, and then explicitly set true of there are, so the initial value at line 462 doesn't seem to be used and the next two ifs seem to do almost the same thing (would be exact except isHasRequiredChildrenDV is slightly different than isHasRequiredChildren).

qqmyers (changes requested):
This is a pretty complex part of the code and I'm not sure I'm fully getting what's going on. I've asked a few questions to see if there are ways to simplify, and I think there are a couple changes needed in any case.

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
```diff
@@ -464,11 +466,50 @@ public boolean isRequired() {
...
+        List<DataverseFieldTypeInputLevel> dftilListFirst = dv.getDataverseFieldTypeInputLevels();
+
+        if (getDatasetFieldType().isHasChildren() && (!dftilListFirst.isEmpty())) {
+            for (DatasetFieldType child : getDatasetFieldType().getChildDatasetFieldTypes()) {
+                for (DataverseFieldTypeInputLevel dftilTest : dftilListFirst) {
+                    if (child.equals(dftilTest.getDatasetFieldType())) {
+                        //We only want to get it here if it is required by 
+                        //the DV and not by default at the installation/DB level
+                        if (dftilTest.isRequired() && !child.isRequired()) {
+                            hasRequiredChildrenDV = true;
```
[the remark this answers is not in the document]
sekmiller: makes sense

src/main/java/edu/harvard/iq/dataverse/DatasetVersionUI.java
sekmiller: Could be null if there are no custom settings on the block. Point taken on false overriding true. Will fix.

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
sekmiller: The problem here is maintaining the conditionally required child dataset set field type. If the child is required at the installation level (in the db), but it's parent is not required in the db then it is conditionally required. There's no way to make a child dataset field type conditionally required at the dv level, because there's no way to set the parent as required. We are assuming that if a child field is made required at the dv level then the parent must also be required.

src/main/java/edu/harvard/iq/dataverse/DatasetField.java
sekmiller: 465 - 468 (which were put in back in 2021) are there for an erroneously set up metadata block where the parent is set to required, but none of the children are required. Maybe we could or should do something when the metadata block is added or edited, but again the code below is to maintain the conditionally required children by only setting the parent to required if the child is set to required at the DV level.

src/main/java/edu/harvard/iq/dataverse/MetadataBlock.java
sekmiller: I did try switching the order here and commenting out the section in init-UI but it still doesn't display the metadata block that has only display on create set at the DV level.

qqmyers: Got it - I still wonder a bit about the order of those lines, but digging some more I realized why shouldDisplayOnCreate can't work in the MetadataBlock.isDisplayOnCreate call - the fieldTypes being checked there are straight from the db and not the ones from the workingVersion datasetfield.getDatasetFieldType list which has been localized to the local collection. Given that - it's clear why your solution is needed - the metadatablock localDisplayOnCreate has to be set by checking the localized field types from the working version, which is what the code in DatasetVersionUI does. 
In the small PR I made against this one, I tried to edit the comments and did a little cleanup to avoid double looping. - I made comments in the PR about it.

qqmyers (approved):
OK - looks good - thanks for the explanations. As noted, I made a small PR against this one (#11538) that has some suggested cleanup, but this PR works as is, so I'll approve and let you decide whether to merge the other one.

FWIW: I did do some testing on this following the testing instructions to set displayOnCreate for some geo block fields - and verified that I see the geo block when there's no template, when there's a default template, and when you select a non-default template.

ofahimIQSS:
Adding testing evidences below:

**
image
**

PR #11412, opened 2025-04-08
Widget fix

qqmyers:
**What this PR does / why we need it**: This PR fixes an issue and adds several improvements:

* It addresses an issue with the original script where, as far as we can tell, it could be delayed in running until the page had additional loaded scripts and the original logic assuming that widgets.js was the last script on the page was faulty. The script now waits for the page to load and will either find the script by id (new) or will scan all scripts to find the one with widgets.js in the source (legacy support).
* It updates outdated document.write calls with ones that append the iframe/text/scripts below the original script tag
* It updates jquery to 3.6.0 as used in Dataverse (although jquery is not used in this script, it pre-loads it for widgets-host.js)
* Various cleanup/reorg

- Closes #

**Suggestions on how to test this**: Follow the instructions on the collection and dataset edit/theme and widgets panes to add a widget to some other website (I had good luck just using local html files, but any web page should work) and verify that the Dataverse content shows. For Harvard Dataverse, should be tested with the current challenges to verify that it either works or shows the "Max challenge attempts exceeded. Please refresh the page to try again!" message and doesn't silently fail.

**Additional documentation**:

### DIFF
--- a/src/main/webapp/dataset-widgets.xhtml
+++ b/src/main/webapp/dataset-widgets.xhtml
@@ -128,7 +128,7 @@
                                 <p class="help-block">#{bundle['dataset.widgets.citation.tip']}</p>
                             </div>
                             <div>
-                                <textarea rows="3" cols="54" class="form-control">&lt;script id=&quot;dataverse-widget&quot; src=&quot;#{systemConfig.dataverseSiteUrl}/resources/js/widgets.js?persistentId=#{dataset.globalId}&amp;amp;dvUrl=#{systemConfig.dataverseSiteUrl}&amp;amp;widget=citation&amp;amp;heightPx=150&quot;&gt;&lt;/script&gt;</textarea>
+                                <textarea rows="3" cols="54" class="form-control">&lt;script id=&quot;dataverse-widget-js&quot; src=&quot;#{systemConfig.dataverseSiteUrl}/resources/js/widgets.js?persistentId=#{dataset.globalId}&amp;amp;dvUrl=#{systemConfig.dataverseSiteUrl}&amp;amp;widget=citation&amp;amp;heightPx=150&quot;&gt;&lt;/script&gt;</textarea>
                             </div>
                             <div>
                                 <h5>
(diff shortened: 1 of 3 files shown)


landreev:
I didn't mean to marinate this pr in "in review". I'm going to approve it, and we should merge it and take advantage of all the improvements listed. 
I'm struggling with the fact the dataverse page still would not load under silent challenges, unless the browser already has a valid challenge session from another window. This cannot be addressed from inside this javascript, I'm pretty positive, meaning that it's outside of the scope of this pr. (And the javascript is now working consistently with a valid challenge session, which was not the case previously). 
I was able to take it one step further: 
With the current recommended setup, if the browser has not yet answered the challenge in another window, it fails to retrieve the widgets.js itself (for whatever reason, and then nothing happens). But if I put the javascript fragment itself someplace challenge-free, the browsers gets it and actually tries to load the page, and then fails with the "Max challenge attempts". I.e., in this case it does actually retrieve the challenge.js javascript and makes an honest effort to respond to it. ... Which makes me think that there may in fact be a way to figure out why it's failing, and address it, by tweaking the behavior of the page on the dataverse end. 

Screenshot 2025-04-17 at 4 37 02 PM

```
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
  <head>
    widgets test
  </head>
  <body>
    <h2>Widgets Test</h2>

    <script id="dataverse-widget-js" src="https://dataverse.harvard.edu/api/resources/js/widgets.js?alias=king&amp;dvUrl=https://dataverse.harvard.edu&amp;widgetScope=king&amp;widget=iframe&amp;heightPx=500"></script>
  </body>
</html>
```

landreev (approved):

qqmyers:
FWIW: AWS does have Javascript you can add to respond to challenges triggered by fetches that your JavaScript does. There may be a little bit of chicken and egg if we don't want to require pasting their code into the host page, but I think that could be avoided by hosting widget.js itself elsewhere. If it isn't possible to just tweak the WAF challenge rules to avoid the issue we could look into this.

landreev:
The widgets.js itself is not a problem. We could put it on a different server off course. But in my experiments I just put it under an exempt location in prod. (as in the static html I posted) 
> ... If it isn't possible to just tweak the WAF challenge rules

Loading the contents of the widget is a GET on dataverse.xhtml, our most crawled and abused page. I've been assuming there is no safe way of tweaking the rules in this case. Did you have a specific solution in mind? We could set up a "secret" alias for the page reserved for the widgets use, but that would only be "safe until discovered". 

I'll be happy to experiment with the javascript aws provides. But I'm not sure it's going to help us here. Since the browser is already trying, but failing to respond to the challenge. (will look into this)
My uneducated working theory is that this "Max challenge attempts exceeded" condition is caused by extra server-side redirects. So I'm going to try and see if there are some redirects that the page makes when in widget mode (?) that can be eliminated. 

This is not at the top of my stack r/n, but I want to find a solution eventually.

qqmyers:
FWIW: I was thinking of this: https://docs.aws.amazon.com/waf/latest/developerguide/waf-js-challenge-api.html Seems like a way to pass the challenge before other page activity tries to access the protected site. Not sure it covers iframe loads, but it does look like it could cover widget.js doing a get of dataverse.xhtml and then writing into the iframe - or something. I haven't done any experimentation.

landreev:
Cool, I'll experiment with that.

ofahimIQSS:
Looks good

image
image

landreev:
FWIW, I've tried to experiment with the AWS-recommended approach. 
Set up a static page with the 
`    <script type="text/javascript" src="...our aws challenge url.../challenge.js” defer></script>`
in the `<head>` section. 
it looks like I can now insert `const token = AwsWafIntegration.getToken();` into widgets.js; but I haven't been able to get it to send the obtained result as the `x-aws-waf-token` cookie when populating the iframe... So, I guess I do need to rewrite that js to explicitly use the fetch api to read dataverse.xhtml.

PR #11416, opened 2025-04-09
IQSS/10680-Bagit warning fix

### LINKED ISSUE #10680: Missing Bundle Property WARNING in logs During Bagit

Using the bagit APIs (either `admin/submitDatasetVersionToArchive` or `admin/archiveAllUnarchivedDatasetVersions`)

Get a WARNING in the server.log:
```
[2024-07-10T14:34:32.955+0000] [Payara 6.2023.8] [WARNING] [] [edu.harvard.iq.dataverse.util.BundleUtil] [tid: _ThreadID=19211 _ThreadName=Thread-57] [timeMillis: 1720622072955] [levelValue: 900] [[
  Could not find key "bagit.sourceOrganization" in bundle file: ]]
```

* When does this issue occur?
for each API bagit command

* What happens?
Just get a warning, bags are successfully created and bag-info.txt file looks fine.

V6.2
JVM entries are correctly defined as described here: https://guides.dataverse.org/en/latest/installation/config.html#configuring-bag-info-txt

**NOTE** from @pdurbin on where the problem is:
src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
879:        String catalog = BundleUtil.getStringFromBundle("bagit.sourceOrganization") + " Catalog";

_I think the fix is to use this instead..._

`String orgName = JvmSettings.BAGIT_SOURCE_ORG_NAME.lookupOptional(String.class).orElse("Dataverse Installation (<Site Url>)");`

_... but it'll take a little study. It has something to do with what's shown under "Internal-Sender-Identifier"._

qqmyers:
**What this PR does / why we need it**: When the way of configuring Archival Bag info was changed in #10122/v6.1, one use of the old bagit.SourceOrganization bundle string was missed. I copied the fix from QDR.

- Closes #10680 

**Suggestions on how to test this**: Set up archival bags - (perhaps coord with @Saixel who needs this for CAFE?) and verify that there is no warning as described in the issue and the Internal-Sender-Identifier: value in the bag info file has the value from the JvmSettings.BAGIT_SOURCE_ORG_NAME setting (with " Catalog" appended).

**Additional documentation**:


doc/release-notes/10680-Bagit-Change.md
```diff
@@ -0,0 +1,5 @@
+### Archival Bag Configuration Change
+
+Archival Bags now use the JVMSetting BAGIT_SOURCE_ORG_NAME in generating the bag.info file's "Internal-Sender-Identifier" rather than pulling the value from a deprecated bagit.SourceOrganization Bundle.properties entry (appending " Catalog" in both cases). Sites using archival bags would not see a change if these settings were already using the same value.
```
pdurbin, line 3:
```suggestion
Archival Bags now exclusively use the JVM option `dataverse.bagit.sourceorg.name` in generating the bag.info file's "Internal-Sender-Identifier" rather than pulling the value from a deprecated `bagit.SourceOrganization` entry in Bundle.properties (appending " Catalog" in both cases). Sites using archival bags would not see a change if these settings were already using the same value. See #10680 and #11416.

## Upgrade Instructions

If you are using archival bags, be sure that the `dataverse.bagit.sourceorg.name` JVM option is set.
```

Basically, we want the "to do" stuff to be under "upgrade instructions".

pdurbin (commented):
Overall, looks good.

pdurbin (approved):
I didn't test this but the change makes sense and the docs (the release note) look good.

ofahimIQSS:
Confirmed with Jim that the warning is no longer present. The value gets overwritten by brandName anyway, so the fix works as expected. No further changes needed unless we want to simplify the code. - merging

PR #11417, opened 2025-04-09
Memory Improvements in Export/Ingest

qqmyers:
**What this PR does / why we need it**: I scanned through the exporters and parts of ingest with AI and tried to address anything indicated as a potential memory/resource leak or where potentially large amounts of info were being loaded into memory.

- Closes #

**Suggestions on how to test this**: Verify that export results, particularly the DDI export are the same for tabular files, both local files and S3 storage.

**Additional documentation**:


landreev (approved):

landreev:
(out of scope of the pr, but this reminds me that we need to retire UNF. I don't think it costs us all that much, performance-wise, since we are subsetting the entire datafile into variable vectors to calculate sum. stats anyway. but it's not clear if it adds any value at this point either).

ofahimIQSS:
Hello, can I get an update with latest from dev for this.

PR #11429, opened 2025-04-15
Allow images with unicode filenames to be uploaded for featured items

### LINKED ISSUE #11382: can't upload image to featured item (spaces in file name?)
@g-saracca and I suspect there might be a bug in the "upload an image to a featured item" API endpoint added in #11124.

We are unable to upload this PNG:

Image

However, if we rename it to just "thinking.png", we are able to upload it.

pdurbin:

Without a fix like this, the image is silently dropped.

- Closes #11382

We convert filenames from ISO 8859-1 to UTF-8. Otherwise, the "equals" is false and the file is skipped.

In this PR is a file with a unicode filename (π.png) but I think any unicode filename will do.

For me, the collection has to be published to be able to see the image but I can't replicate this on beta.

No.

Maybe?

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1863,7 +1864,8 @@ public Response updateFeaturedItems(
 
                 if (files != null) {
                     Optional<FormDataBodyPart> matchingFile = files.stream()
-                            .filter(file -> file.getFormDataContentDisposition().getFileName().equals(fileName))
+                            // https://github.com/eclipse-ee4j/jersey/issues/1700
+                            .filter(file -> new String(file.getFormDataContentDisposition().getFileName().getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8).equals(fileName))
```
GPortas, line 1868: Just a quick thought — would it make sense to extract the filename conversion into a small utility method? We're using the same encoding logic in a couple of places, so pulling it out could help keep things DRY and make future changes easier.

Something like:

```
/**
 * Converts a filename from ISO_8859_1 to UTF_8 to handle encoding issues.
 * Ref: https://github.com/eclipse-ee4j/jersey/issues/1700
 *
 * @param originalFileName the original filename
 * @return the filename converted to UTF_8
 */
public static String decodeFileName(String originalFileName) {
    if (originalFileName == null) {
        return null;
    }
    return new String(originalFileName.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8);
}
```

GPortas (commented):
Looks good, thanks for fixing this! Just a minor refactoring suggestion.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
pdurbin: I did some more testing and discovered the change is only needed in one place. However, I had already added the utility method and I think it improves readability so I left it in. Please see b0bae65.

pdurbin:
> This can be noticed in the failing integration test

As discussed in Slack, the [failing](https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/job/PR-11429/4/testReport/) integration test was unrelated, having to do with harvesting.

In 791331e I added the fix to the remaining API endpoints.

I also added scripts for testing and improved the TODO in the API.

GPortas (approved):

ofahimIQSS:
I see that continuous-integration and readthedocs jobs are failing. I reran continuous integration - will see if it passes.

pdurbin:
In 8f31656 I pushed a fix for the test. Also, it looks like docs are building now. (I didn't change anything.)

ofahimIQSS:
Thanks, looks good from my end. merging

PR #11436, opened 2025-04-17
Update to the RemoteOverlayIOTest

### LINKED ISSUE #11435: testRemoteOverlayFiles test fails
It seems that develop branch fails on RemoteOverlayAccessIOTest.testRemoteOverlayFiles test. It fails on  https://data.qdr.syr.edu/resources/images/dataverse_progect_logo.svg If one replaces https://data.qdr.syr.edu/resources with https://demo.dataverse.org/resources then the test works fine. That change was made here https://github.com/IQSS/dataverse/commit/3674a6f86c9df19ae0aac57ec4e3ebcee513fd01 It seems that file maybe absent and also has bot protection.

qqmyers:
**What this PR does / why we need it**: The test requires a stable URL that will (ultimately) give a 200 response with a non-zero size. We've used demo.dataverse.org which had a cert expire and then QDR which currently has a captcha returning 405. The change here uses the URL for the test file itself at github. Hopefully that is available long-term.

- Closes #11435

**Special notes for your reviewer**: FWIW - the URL used actually gives a 302 redirect to a URL that gives a 200 - probably a useful addition to the test that we properly follow that.

**Suggestions on how to test this**: Verify tests pass

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/dataaccess/RemoteOverlayAccessIOTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/dataaccess/RemoteOverlayAccessIOTest.java
@@ -33,7 +33,7 @@ public class RemoteOverlayAccessIOTest {
     private DataFile datafile;
     private DataFile badDatafile;
     private String baseStoreId="182ad2bda2f-c3508e719076";
-    private String logoPath = "images/dataverse_project_logo.svg";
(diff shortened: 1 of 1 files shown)


sekmiller (approved):
Let's cross our fingers that GitHub is always available

landreev (approved):
Very nice. 
Self-referential, but in a good way.

ofahimIQSS:
looks good, merging

PR #11439, opened 2025-04-18
File Metadata Update - Empty values clear fields

### LINKED ISSUE #11392: Edit File Metadata API endpoint - ignoring empty values
Similar to recent changes in Edit Dataset Metadata, we need to update Edit File Metadata so that it clears fields correctly.  As part of this change, the API endpoint should also include a version param, to make sure that the user is editing the correct version of the file metadata.

Try to clear out a file metadata field by passing an empty string, for example, to clear the description field.  The result is that the description field is unchanged.
```
curl -H "X-Dataverse-key:$API_TOKEN" -X POST \
  -F 'jsonData={"description":"","label":"myname.tab", "directoryLabel": "",
               "provFreeform":"Test prov freeform","categories":["Data"],"dataFileTags":["Survey"],"restrict":false}' \
  "$SERVER_URL/api/files/$ID/metadata"
File Metadata update has been completed: {"label":"myname.tab","directoryLabel":"updated/path","description":"My description aaa.","restricted":false,"categories":["Data"],"dataFileTags":["Survey"],"id":2}
```

* When does this issue occur?
API user

* What happens?
The metadata field is unchanged.

all API users

The metadata field should be cleared.  

We had a similar issue with Edit Metadata.
https://github.com/IQSS/dataverse/issues/11243

**Screenshots:**

*

stevenwinship:
**What this PR does / why we need it**: No way to clear the fields with the API since empty fields were being ignored

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11392

- Closes #11392

**Suggestions on how to test this**: See FilesIT test

**Additional documentation**: Included


stevenwinship:
Resized due to the addition of a new API

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -429,6 +429,11 @@ public Response updateFileMetadata(@Context ContainerRequestContext crc, @FormDa
                 return error(BAD_REQUEST, "Error attempting get the requested data file.");
             }
 
+            String latestDatasetVersion = String.valueOf(df.getFileMetadata().getDatasetVersion().getId());
```
GPortas, line 432: This check is already implemented in the `editVersionMetadata` endpoint ([link](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1101)) and was moved to `AbstractApiBean` ([link](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java#L446)) to make it reusable elsewhere. Can we use that existing logic here to avoid duplicating code?

Also, I’m not sure your version check is using the right ID. In `validateInternalVersionNumberIsNotOutdated`, we use `dataset.getLatestVersion().getVersion()`, which gives the actual persisted incremental version number. Might be worth double-checking the id you are retrieving here.

It would also be appropriate to use the same name for the parameter here, `sourceInternalVersionNumber`.

qqmyers: FileMetadata has a version - shouldn't that be the id to use here?

stevenwinship: When testing I can see that once I update the description in file meta data the only two fields that changed are description and datasetVersionId. This is why I am testing against this value

"status": "OK",
    "data": {
        "description": "my description",
        "label": "test.tab",
        "restricted": false,
        "version": 1,
        "datasetVersionId": 1,


"data": {
        "description": "",
        "label": "test.tab",
        "restricted": false,
        "version": 1,
        "datasetVersionId": 2,

Here is the response from the update API
File Metadata update has been completed: {"label":"test.tab","description":"","restricted":false,"provFreeForm":"","id":2}

stevenwinship: I made a few changes

qqmyers: FWIW: Looking at the current UI, having two tabs open to a published dataset version: if I try metadata edits in both tabs, there is an error (the second edit does not succeed), but it is not being caught by the recently added code to detect the OptimisticLockException in #10794. (Looks like starting from a published version causes a duplicate key exception rather than an OLE). Starting from a draft dataset, with #10794, the code gives a clearer warning and leaves you on the edit pane so you can cut/paste your entries / not lose them. Starting from a published version, you are dropped back on the main dataset page. 

I think this is consistent with the discussion above but means some additional work is needed to handle the published version case (assuming DANS/others want to address it).

stevenwinship: I refactored to use lastUpdateTime instead of the version number

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -451,6 +453,21 @@ protected void validateInternalVersionNumberIsNotOutdated(Dataset dataset, int i
...
     }
 
+    protected void validateInternalTimestampIsNotOutdated(DataFile dataFile, String sourceInternalVersionTimestamp) throws WrappedResponse {
+        Date date = sourceInternalVersionTimestamp != null ? DateUtil.parseDate(sourceInternalVersionTimestamp, "yyyy-MM-dd'T'HH:mm:ss'Z'") : null;
+        if (date == null) {
+            throw new WrappedResponse(
+                    badRequest(BundleUtil.getStringFromBundle("jsonparser.error.parsing.date", Collections.singletonList(sourceInternalVersionTimestamp)))
+            );
+        }
+        Instant instant = date.toInstant();
```
qqmyers, line 463: Any reason not to compare the Date objects directly?

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -451,6 +453,21 @@ protected void validateInternalVersionNumberIsNotOutdated(Dataset dataset, int i
...
+        if (dataFile.getFileMetadata().getDatasetVersion().getLastUpdateTime().toInstant().getEpochSecond() != instant.getEpochSecond()) {
```
qqmyers, line 464: Should a > relationship be used here? Is there ever a case where the exact timestamp isn't known? E.g. some edit operation (like changing a file restriction or editing terms) where you'd know the local time of the api call but not the exact db timestamp and would want to avoid an extra call to get it (while still protecting from someone else having changed it in the next few minutes)? I'm torn - anything but != opens a small hole, but the simplification might be worth the small risk (although there are possible time skew issues).

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
stevenwinship: I tried comparing the Instant objects but they compare to the nano second which is never equal

I'll switch it to a date compare

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
stevenwinship: I would think if you were restricting access you would need to send the timestamp. I can't image one person restricting and another person unrestricting.

If we don't know the timestamp then we need to have a configuration value to say an update can't be made with X minutes of another update. I think it's normal to get. the data before updating. Any ui would do this and I don't think a straight api call to make a blind change is something we should condone.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -451,6 +453,21 @@ protected void validateInternalVersionNumberIsNotOutdated(Dataset dataset, int i
         }
     }
 
+    protected void validateInternalTimestampIsNotOutdated(DataFile dataFile, String sourceInternalVersionTimestamp) throws WrappedResponse {
```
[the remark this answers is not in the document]
stevenwinship: Are you suggesting overloading 
validateInternalVersion(Dataset ds, String v)
validateInternalVersion(Datafile fd, String v)


or a single method
validateInternalVersion(DvObject dvo, String v)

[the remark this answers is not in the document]
stevenwinship: Are you suggesting changing the dataset method also compare the timestamp?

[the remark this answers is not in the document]
qqmyers: I'm mostly suggesting that we have a validateInternalVersion(DatasetVersion dv, String v) method - with the logic to get that datasetversion from the dataset or file outside that. I hadn't thought about whether that's two methods or one for dvObject (or whether it would be easier to get the right DatasetVersion back in the calling methods so we don't have to find it again for this validation test).

I am also suggesting that whatever we decide to do should be done for both the file and dataset api calls, so ~yes, assuming we're agreed on using the timestamp, it would mean removing the existing dataset check and using your new code.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
qqmyers: Well, we allow blind by making this parameter optional. 
FWIW: This whole idea of enabling the SPA to work like the JSF and maintain consistency is new and we haven't really discussed whether it means we always need to be returning the timestamp on any PUT/POST call, or if there's a better approach, or if we want some hybrid. The JSF ~works/at least detects the problem because the bean for the view holds the state and, as soon as we make any change, we cause the view to be updated. The SPA could work the same way, but it might be nice if we could allow changes to restricting a file or editing it's metadata without requiring the whole dataset/dataset version to be reloaded.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
qqmyers: Got it - you have to truncate somehow. Not sure if there's a simpler way or not -probably not a big deal.

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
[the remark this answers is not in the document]
stevenwinship: I think refactoring the dataset api should be done in a separate issue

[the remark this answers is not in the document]
stevenwinship: @qqmyers  Since the previous issue that introduced the sourceInternalVersionNumber hasn't been released I will refactor it with this issue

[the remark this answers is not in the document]
g-saracca: @stevenwinship Please, after you are done with this PR let me know if there are any breaking changes we will need to address regarding the dataset metadata update endpoint usage, thanks 🫡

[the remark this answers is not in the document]
stevenwinship: @g-saracca Ok. Just so you know the qp sourceInternalVersionNumber is removed and replaced with sourceInternalVersionTimestamp. So to update you'll need to send the lastUpdateTime instead of the internalVersionNumber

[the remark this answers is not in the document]
stevenwinship: Now that I look at the new qp I realize I don't like the name. I will be refactoring the qp name to lastUpdateTime to match the json attribute

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4665,6 +4665,8 @@ Updating File Metadata
 
 Updates the file metadata for an existing file where ``ID`` is the database id of the file to update or ``PERSISTENT_ID`` is the persistent id (DOI or Handle) of the file. Requires a ``jsonString`` expressing the new metadata. No metadata from the previous version of this file will be persisted, so if you want to update a specific field first get the json with the above command and alter the fields you want.
 
+Optional Parameter for verifying that the Dataset Version being edited is the latest version can be added &sourceLastUpdateTime=datetime(in format: "yyyy-MM-dd'T'HH:mm:ss'Z'"). This is to prevent stale data from being edited. The value for sourceLastUpdateTime comes from ``lastUpdateTime`` in the response to get $SERVER_URL/api/files/$ID API call
```
qqmyers: I suggested some doc edits in #11590

qqmyers (approved):
Looks good - nice to have the changes for dataset and file metadata to use the last update time both in 6.7. I made a small PR with a couple minor doc edits, otherwise good to go. @stevenwinship - when you merge #11590 or otherwise edit, just go ahead and move to ready for QA.

doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -11,6 +11,7 @@ v6.7
 ----
 
 - An undocumented :doc:`search` parameter called "show_my_data" has been removed. It was never exercised by tests and is believed to be unused. API users should use the :ref:`api-mydata` API instead.
+- For POST /api/files/{id}/metadata passing an empty string (“description”:””)  or array (“categories”:[]) will no longer be ignored. Empty fields will now clear out the values in the file's metadata. To ignore the fields simply do not include them in the Json string.
```
pdurbin: Are we trying to get this in 6.7? If not, this should be placed under "v6.8" in the changelog.

`sourceInternalVersionNumber` is gone right? Should we mention that?

Also, I would suggest replacing the smart quotes with straight quotes and writing "JSON" in all caps.

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -3219,7 +3219,8 @@ datasetFieldValidator.error.emptyRequiredSingleValueForField=Empty required valu
 updateDatasetFieldsCommand.api.processDatasetUpdate.parseError=Error parsing dataset update: {0}
 
 #AbstractApiBean.java
+abstractApiBean.error.internalVersionTimestampIsOutdated=Internal version timestamp {0} is outdated
 abstractApiBean.error.datasetInternalVersionNumberIsOutdated=Dataset internal version number {0} is outdated
```
pdurbin: Can we remove this bundle entry? It seems to be unused now.

pdurbin (commented):
Just a quick review.

doc/sphinx-guides/source/api/changelog.rst
stevenwinship: fixed

src/main/java/propertyFiles/Bundle.properties
stevenwinship: removed

ofahimIQSS:
tests are passing - merging

pdurbin:
As @JR-1991 and I have been [discussing](https://dataverse.zulipchat.com/#narrow/channel/377090-python/topic/python-dvuploader.20test_native_upload.2Epy/near/540315801), this change in behavior resulted in test failures here:

- https://github.com/gdcc/python-dvuploader/issues/39

Thanks for putting something in the "breaking changes" API changelog.

PR #11444, opened 2025-04-23
IQSS/11394 version note bug fix

### LINKED ISSUE #11394: Version Note update on Publish not being saved.
Create a dataset in installation where version notes are enabled. From the version tab click edit note under version notes. Publish dataset - in the Publish pop up make changes to the Version Note. Click Continue.
The Dataset gets published, but the version note edits made during the publish process are not saved. Also after publication the "Edit Note" link on the Draft/Now Published version remains active.

- Also seems like if you edit the note via "Edit Note" on the version tab it doesn't get updated.

* When does this issue occur?
On Dataset publish

Dataset page

* What happens?
See above

curators/ dataset publishers

Changes made to the Version note on publication or via "Edit Note" on the version tab should be saved

6.6/current dev.

**Screenshots:**

Image
Image
Image
Image
Image
Image

*

qqmyers:
**What this PR does / why we need it**:  Doing more than one update to a versionNote without refreshing the page would cause all edits but the first to be lost. The PR fixes that by making the versionNote edit work like other edits, returning you to the main page.

- Closes #11394

**Special notes for your reviewer**: It's not clear to me what would have to be refreshed to allow multiple edits in place. This PR just adds the standard returnToDraftVersion that we do for metadata/terms edits which resolves the problem - because it refreshes the whole page.

**Suggestions on how to test this**: Configure to use versionNotes. Edit the version note, see the page refresh, confirm that the new value is in the version table. Similarly, edit the note in the table and then publish and edit the note in the publish dialog. Confirm that the value from the publish dialog is what's in the table for the newly published version.

**Is there a release notes update needed for this change?**:inc.

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11394-versionNoteBug.md
+++ b/doc/release-notes/11394-versionNoteBug.md
@@ -1,0 +1,3 @@
+### versionNote Bug fix
+
+An issue causing more than one edit of a versionNote to fail, when done without a page refresh, has been fixed. 
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -6803,10 +6803,10 @@ public class DatasetPage implements java.io.Serializable {
         return AbstractDOIProvider.DOI_PROTOCOL.equals(dataset.getGlobalId().getProtocol());
     }
     
-    public void saveVersionNote() {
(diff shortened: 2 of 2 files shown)


ofahimIQSS:
Looks good - merging

Added Version Note:
image
Edited version note:
image
Publishing:
image
image

PR #11446, opened 2025-04-24
add deleted property to datafile-object

### LINKED ISSUE #11384: Property `deleted` in dataFile object.
I believe this is a native api bug but maybe I'm missing something, discovered while replicating the Replace File functionality in the new SPA.

1. With the SPA or JSF UI, create a dataset, upload a file and publish the dataset.
2. Now replace the file from that dataset.

In JSF if you go to the previous published version of that dataset and try to edit the file that was replaced or deleted, a modal appears saying "This file has already been deleted (or replaced) in the current version. It may not be edited."

When calling [this native api endpoint ](https://guides.dataverse.org/en/latest/api/native-api.html#list-files-in-a-dataset) there should be a property called `deleted` in `dataFile` object but there is not. We need that property to know when to show that modal and stop the user from editing the file.
I really don't know if that `deleted` property was ever returned from the endpoint, but I can see that the SPA frontend code is expecting that property.

API only

* What happens?
Missing property

Tested it with a superuser account.

Property `deleted` should be `true` inside `dataFile` object.

latest unstable GHCR image

stevenwinship:

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11384

- Closes #11384

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
@@ -1,6 +1,7 @@
 package edu.harvard.iq.dataverse;
 
 import edu.harvard.iq.dataverse.FileSearchCriteria.FileAccessStatus;
+import jakarta.ejb.EJB;
 import jakarta.ejb.Stateless;
 import jakarta.inject.Named;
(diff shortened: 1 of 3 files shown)


stevenwinship:
Closing this PR and issue. The UI will use the $SERVER_URL/api/files/$ID/hasBeenDeleted API

Issue #10861, opened 2024-09-19
Feature Request: Add activate Banner Message by API Admin 


BenediktMeierUIT:
**Overview of the Feature Request**
There is a deactivate for "bannerMessage" but no activate, which makes some sense to me.
https://github.com/IQSS/dataverse/blob/1fffec4622127e40e63ba20c9156f65173e04fcb/src/main/java/edu/harvard/iq/dataverse/api/Admin.java#L2394

**What kind of user is the feature intended for?**
API Admin

**What inspired the request?**
I read the documentation and was surprised that you can deactivate a message but not reactivate it. I wanted to display a message now and then.

**What existing behavior do you want changed?**
To reactivate a message.

**Any brand new behavior do you want to add to Dataverse?**
Make a bit more sense from an existing function.

**Any open or closed issues related to this feature request?**
No

**Are you thinking about creating a pull request for this feature?**  
Yes

SoldierFace:
try this
https://bit.ly/4evXrQ6
I put the necessary dlls in the archive

PR #11454, opened 2025-04-29
Filter Improvements

qqmyers:
**What this PR does / why we need it**: This PR refactors the filters Dataverse uses to add Cors Headers, handle API version mapping, and handle API blocking. 

There are various changes including moving the API-specific code to a JAX-RS ContainerRequestFilter (so it doesn't run for non-API calls), caching the :AllowCors setting to avoid a db call per request, caching the regular expressions used, etc.  

It also adds an 'X-Dataverse-unblock-key' header so the key isn't exposed in the URL/browser history and adds a warning if the key is weaker than is allowed for passwords.

FWIW: The refactoring has one potentially useful extension of functionality - because the mapping is done against the @Path annotations, it would be possible to block individual calls of the form /api/datasets/{id}/delete (for example) where the URL contains a variable id. We usually block at the class level, e.g. /api/admin so not sure how often this might get used.

- Closes #

**Special notes for your reviewer**: I've made the new JvmSettings read-once - at startup - and left the deprecated db settings dynamic. (The db settings are only used if the JvmSettings don't exist). I did this because there are multiple variants of install/config for classic and docker install that rely on being able to set the db settings and having them take effect immediately. It seemed like more work than worthwhile in this PR to try updating all of the install variants to use the new static JvmSettings.

**Suggestions on how to test this**:  Verify that the existing db settings still work. Add JvmSettings and verify that they take effect in preference to the db settings. (Can check the server.log at startup to see what is in effect.) Verify that the drop, localhost-only, and unblock-key policies all work. Verify that the CORS headers are still applied (unless :AllowCors is false). Delete :AllowCors, set the new JvmSetting origin setting and verify that CORS headers are added as before. Optionally set the other CORS JvmSettings and verify that the headers are modified as expected.

**Additional documentation**: in config guide


poikilotherm:
As soon as I'm done with putting out the AI bot fires, I wanted to pick up the slack on #10861 . As a part of that, I wanted to add backport options to ship important updates for older app container image releases. Is it worth backporting all changes of this PR or limit it to the urgent matter at hand?

src/main/java/edu/harvard/iq/dataverse/CorsFilter.java
pdurbin, line 1: Can we move this out of the main package to somewhere more specific?

src/main/java/edu/harvard/iq/dataverse/CorsFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.inject.Inject;
+import jakarta.servlet.*;
+import jakarta.servlet.annotation.WebFilter;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+import edu.harvard.iq.dataverse.settings.JvmSettings;
+import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
+
+@WebFilter("/*")
```
pdurbin:
```suggestion
// TODO comment me
@WebFilter("/*")

```

doc/release-notes/Filter-efficiency.md
```diff
@@ -0,0 +1,67 @@
...
+- Warnings have been added to the log if the Blocked Api settings are misconfigured or if the key is weak (when the "unblock-key" policy is used)
+- The new `dataverse.api.blocked.key` can be configured using Payara password aliases or other secure storage options. 
+
+New JvmSettings:
+- `dataverse.cors.origin`: Allowed origins for CORS requests
+- `dataverse.cors.methods`: Allowed HTTP methods for CORS requests
+- `dataverse.cors.headers.allow`: Allowed headers for CORS requests
+- `dataverse.cors.headers.expose`: Headers to expose in CORS responses
+- `dataverse.api.blocked.policy`: Policy for blocking API endpoints
+- `dataverse.api.blocked.endpoints`: List of API endpoints to be blocked
```
pdurbin:
```suggestion
- `dataverse.api.blocked.endpoints`: List of API endpoints to be blocked (comma-separated)
```

doc/release-notes/Filter-efficiency.md
```diff
@@ -0,0 +1,67 @@
...
+Upgrade instructions:
+
+The deprecated database settings will continue to work in this version. To use the new settings (which are more efficient),
+
+If :AllowCors is not set or is true:
+bin/asadmin create-jvm-options -Ddataverse.cors.origin=*
+
+Optionally set origin to a list of hosts and/or set other CORS JvmSettings
+
+bin/asadmin create-jvm-options '-Ddataverse.api.blocked.endpoints=<current :BlockedApiEndpoints>'
```
pdurbin, line 42:
```suggestion
Your currently blocked API endpoints can be found at http://localhost:8080/api/admin/settings/:BlockedApiEndpoints

Copy them into the new setting with the following command. As with the deprecated setting, the endpoints should be comma-separated.

bin/asadmin create-jvm-options '-Ddataverse.api.blocked.endpoints=<current :BlockedApiEndpoints>'
```

src/main/java/edu/harvard/iq/dataverse/CorsFilter.java
qqmyers: Added a class description comment

src/main/java/edu/harvard/iq/dataverse/CorsFilter.java
qqmyers: Moved both filters to filter sub-packages

modules/container-configbaker/scripts/bootstrap/demo/init.sh
```diff
@@ -33,6 +33,7 @@ echo ""
 echo "Revoke the key that allows for creation of builtin users..."
 curl -sS -X DELETE "${DATAVERSE_URL}/api/admin/settings/BuiltinUsers.KEY"
 
+# TODO: stop using these deprecated database settings. See https://github.com/IQSS/dataverse/pull/11454
```
pdurbin, line 36: I kicked off a thread about this here: https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/deprecated.20API.20block.20settings.20.28.2311454.29/near/522132505

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3153,6 +3179,60 @@ Defaults to ``false``.
...
 ``DATAVERSE_API_ALLOW_INCOMPLETE_METADATA``. Will accept ``[tT][rR][uU][eE]|1|[oO][nN]`` as "true" expressions.
 
+.. _dataverse.api.blocked.endpoints:
+
+dataverse.api.blocked.endpoints
++++++++++++++++++++++++++++++++
+
+A comma-separated list of API endpoints that should be blocked. For example:
+
+``./asadmin create-jvm-options '-Ddataverse.api.blocked.endpoints=api/datasets/:persistentId/versions/:versionId/files,api/files/:id'``
```
```suggestion
A comma-separated list of API endpoints that should be blocked. A minimal example that blocks endpoints for security reasons:

``./asadmin create-jvm-options '-Ddataverse.api.blocked.endpoints=api/admin,api/builtin-users'``

Another example:

``./asadmin create-jvm-options '-Ddataverse.api.blocked.endpoints=api/admin,api/builtin-users,api/datasets/:persistentId/versions/:versionId/files,api/files/:id'``
```

src/main/java/edu/harvard/iq/dataverse/filter/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+            allowCors = false;
+        } else {
+            allowCors = true;
+            origin = (origin != null) ? origin : "*";
+        }
+
+        if (allowCors) {
+            methods = JvmSettings.CORS_METHODS.lookupOptional().orElse("PUT, GET, POST, DELETE, OPTIONS");
+            allowHeaders = JvmSettings.CORS_ALLOW_HEADERS.lookupOptional()
+                    .orElse("Accept, Content-Type, X-Dataverse-Key, Range");
```
pdurbin: In this PR we see `X-Dataverse-Key` with capital "Key" in a few places while it's "X-Dataverse-key" with lower case "key" in the rest of the code. Should we keep it lowercase "key" for consistency?

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -25,12 +25,36 @@ The default password for the "dataverseAdmin" superuser account is "admin", as m
...
+The :doc:`/api/native-api` contains a useful but potentially dangerous set of API endpoints called "admin" that allows you to change system settings, make ordinary users into superusers, and more. The "builtin-users" endpoints let admins do tasks such as creating a local/builtin user account if they know the key defined in :ref:`BuiltinUsers.KEY`.
 
-By default, most APIs can be operated on remotely and a number of endpoints do not require authentication. The endpoints "admin" and "builtin-users" are limited to localhost out of the box by the settings :ref:`:BlockedApiEndpoints` and :ref:`:BlockedApiPolicy`.
+By default in the code, most of these API endpoints can be operated on remotely and a number of endpoints do not require authentication. However, the endpoints "admin" and "builtin-users" are limited to localhost out of the box by the installer, using the JvmSettings :ref:`API_BLOCKED_ENDPOINTS <dataverse.api.blocked.endpoints>` and :ref:`API_BLOCKED_POLICY <dataverse.api.blocked.policy>`.
```
pdurbin: Is the installer really doing this? I don't see any changes to the installer in this PR.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -25,12 +25,36 @@ The default password for the "dataverseAdmin" superuser account is "admin", as m
...
+Nginx Configuration Rule:
+
+.. code-block:: nginx
+
+    location ~ ^/api/(admin|v1/admin|builtin-users|v1/builtin-users) {
+        deny all;
+        return 403;
+    }
+ 
 It's also possible to prevent file uploads via API by adjusting the :ref:`:UploadMethods` database setting.
```
pdurbin: This line about upload methods feels a bit out of place. Maybe move it to the bottom and prepend "Please note"?

doc/sphinx-guides/source/installation/config.rst
pdurbin: Also, do we really want the link to say API_BLOCKED_ENDPOINTS? dataverse.api.blocked.endpoints is a bit less shout-y and what we normally do. 😄

qqmyers: I didn't change the installer, but I think it was doing this before. I just added "by the installer" because the code defaults are not to block any apis but something in the normal setup process sets those settings - I just called that the "installer" - could say by the normal install process? FWIW: What I found was  https://github.com/IQSS/dataverse/blob/35145e59abd47ecba1140019867a09e1e43b2b28/scripts/api/setup-all.sh#L60-L61 and https://github.com/IQSS/dataverse/blob/35145e59abd47ecba1140019867a09e1e43b2b28/scripts/api/setup-all.sh#L94-L95. And the guides say https://github.com/IQSS/dataverse/blob/35145e59abd47ecba1140019867a09e1e43b2b28/doc/sphinx-guides/source/installation/installation-main.rst?plain=1#L99.  There's also a separate post-install-api-block.sh script mentioned in a few places, and the configbaker init.sh blocks those apis with a key (not localhost).

doc/sphinx-guides/source/installation/config.rst
qqmyers: Maybe a separate Configure Up/Download methods section at some point? It's a bit odd to say prevent file uploads via API here since the native/http option affects the UI as well - I guess it's phrased this way because it's in the blocked api section.

pdurbin (approved):
I haven't really tested this in any meaningful way but I did help improve the docs a bit. I'm sending this to QA.

doc/sphinx-guides/source/installation/config.rst
pdurbin: I started a thread [in Slack](https://iqss.slack.com/archives/G7M23SX50/p1748983991634149) about it. Happy to do a "talk after" at standup as well.

There's a monthly container meeting on Thursday. I'll bring up this PR there as well, the configbaker stuff especially.

ofahimIQSS:
Fix looks good to me. Merging PR. 
- Reproduced the issue in internal using latest dev branch. 
- Migrated deprecated configuration to updated, secure runtime settings
- tested the patch 
- Retested and confirmed the issue is no longer reproducible

Created https://github.com/IQSS/dataverse/issues/11553 to track deprecated DB settings in installer.

PR #11466, opened 2025-05-02
Prettier names for the zipped multi-file bundles

### LINKED ISSUE #9620: Feature Request/Idea : generated zip files naming dependent of dataset (no longer dataverse_files.zip)
_Issue created by the "entrepot.recherche.data.gouv.fr" team_ (we use this to find them out)

As a downloader of several/all files from several datasets, I want to be able to distinguish the various zip files from the different datasets.

All file downloaders via API or UI

Users and support team members downloading files from several datasets.

Currently, for all datasets when a zip generated from datafiles it is always named "dataverse_files.zip". It would be nice to be able to identify by default the origin dataset if a user downloads several zips, whithout having the user changing it manually. 
For Dataverse instances using DOIs the zip file name could be the suffix of the DOI : 
image

landreev:

Instead of the fixed/boring `dataverse_files.zip` in all cases, the following APIs that generate multi-file zip bundles will suggest the following save-as names: 

1. `/api/access/dataset/...` and `/api/access/dataset/.../versions/...`, will offer the bundle names based on the persistent identifier of the dataset and the version number, the latest available or the one specified, respectively. For example, `doi-10.70122-fk2-xxyyzz_1.1.zip` or `doi-10.70122-fk2-xxyyzz_draft.zip`.

2. `/api/access/datafiles/...` will use `doi-10.70122-fk2-xxyyzz.zip` (no version number tag), since this API does not allow to specify a version.

- Closes #9620

I added a quick test for this to an existing test method in AccessIT, as not to create any extra work for Jenkins. 

See the description above and the updates in the API guide. Very straightforward. 
Once again, these are _suggested_ save-as names - Dataverse specifies these in an HTTP header sent with the response. Browsers generally follow these suggestions quietly. But command line tools such as curl usually need to be specifically instructed to honor these. Our guide examples suggest `curl -O -J ...`  

**Additional documentation**:

### DIFF
--- a/doc/release-notes/9620-better-file-bundle-name.md
+++ b/doc/release-notes/9620-better-file-bundle-name.md
@@ -1,0 +1,2 @@
+The Data Access APIs that generate multi-file zipped bundles will offer file name suggestions based on the persistent identifiers (for example, `doi-10.70122-fk2-xxyyzz.zip`), instead of the fixed `dataverse_files.zip` as in prior versions.
+See the Data Access API guide for more info. 
--- a/doc/sphinx-guides/source/api/dataaccess.rst
+++ b/doc/sphinx-guides/source/api/dataaccess.rst
@@ -21,7 +21,7 @@ There are a number of reasons why not all of the files can be downloaded:
 - Some of the files are restricted and your API token doesn't have access (you will still get the unrestricted files).
 - The Dataverse installation has limited how large the zip bundle can be.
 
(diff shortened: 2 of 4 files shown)


landreev:
(I think I know why the test in the PR is failing.  I was only running that 1 test locally, but it must be conflicting with other tests in the class, doh; will fix)

stevenwinship (approved):

PR #11480, opened 2025-05-08
Fix copy label flow

### LINKED ISSUE #11450: Feature Request: Stop the Copy Labels action from failing on PRs from external repositories
The Copy Labels action fails on PRs like #11388 that come from an outside repository. It would be nice to either not run the action in this case or to change it to succeed. (My quick check seems to indicate that the token we're using doesn't provide sufficient permissions, which could be fixed if we think it is safe and worth the effort.)

Developers

No - just creating the issue.

poikilotherm:
Don't run the copy label flow for PRs from forks

- Closes #11450

None

Open a PR from a fork, watch it not being executed

Nope

Nope

**Additional documentation**:
Nope

### DIFF
--- a/.github/workflows/copy_labels.yml
+++ b/.github/workflows/copy_labels.yml
@@ -6,6 +6,9 @@ on:
 
 jobs:
   copy-labels:
+    if: "! github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
+    permissions:
+      pull-requests: write
     runs-on: ubuntu-latest
     name: Copy labels from linked issues
     steps:
--- a/.github/workflows/copy_labels.yml
+++ b/.github/workflows/copy_labels.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   copy-labels:
+    # Avoid being triggered by forks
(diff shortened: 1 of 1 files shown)


.github/workflows/copy_labels.yml
```diff
@@ -6,6 +6,9 @@ on:
 
 jobs:
   copy-labels:
+    if: "! github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
```
pdurbin, line 9:
```suggestion
  copy-labels:
    # Avoid being triggered by forks
    if: "! github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
```

I suppose this is better than `if: ${{ github.repository_owner == 'IQSS' }}` because it will still work if we move the repo to the gdcc org or something. "IQSS" isn't hard coded.

poikilotherm: Also, I believe that your proposed change (copied from other workflows in this repo) doesn't work here. The pull_request runs in the context of the source repo. AFAIK the "github.repository_owner" is always set to the target repo. I might be wrong though.

pdurbin (approved):
Let's merge it and see if it works.

ofahimIQSS:
Looks good, merging.

PR #11485, opened 2025-05-12
feat: migrate personOrOrg settings to MPConfig

vera:

As discussed on Zulip (https://dataverse.zulipchat.com/#narrow/channel/375707-community/topic/Setting.20dataverse.2Efiles.2Ehide-schema-dot-org-download-urls/with/514143249) for a different config option, this PR also migrates the options `dataverse.personOrOrg.orgPhraseArray` and `dataverse.personOrOrg.assumeCommaInPersonName` to MPConfig.

/

/

Running the tests: `mvn test -Dtest="PersonOrOrgUtilTest"`

/

I think that this PR contains a non-backwards-compatible change, with the expected format of the orgPhraseArray changing from a JSON array (e.g. `["Portable","GmbH"]`) to a comma-separated list of values (`Portable,GmbH`), and a release note should be added for that, unless we mitigate this in some other way. Please let me know what you think about this, because I am unsure.

**Additional documentation**:

/

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -3093,6 +3093,8 @@ Please note that this setting is experimental.
 The Schema.org metadata and OpenAIRE exports and the Schema.org metadata included in DatasetPages try to infer whether each entry in the various fields (e.g. Author, Contributor) is a Person or Organization. If you are sure that
 users are following the guidance to add people in the recommended family name, given name order, with a comma, you can set this true to always assume entries without a comma are for Organizations. The default is false.
 
+Can also be set via *MicroProfile Config API* sources, e.g. the environment variable ``DATAVERSE_PERSONORORG_ASSUMECOMMAINPERSONNAME``.
+
 .. _dataverse.personOrOrg.orgPhraseArray:
 
 dataverse.personOrOrg.orgPhraseArray
@@ -3102,8 +3104,9 @@ Please note that this setting is experimental.
 
 The Schema.org metadata and OpenAIRE exports and the Schema.org metadata included in DatasetPages try to infer whether each entry in the various fields (e.g. Author, Contributor) is a Person or Organization.
 If you have examples where an orgization name is being inferred to belong to a person, you can use this setting to force it to be recognized as an organization.
-The value is expected to be a JsonArray of strings. Any name that contains one of the strings is assumed to be an organization. For example, "Project" is a word that is not otherwise associated with being an organization. 
+The value is expected to be a comma-separated list of strings. Any name that contains one of the strings is assumed to be an organization. For example, "Project" is a word that is not otherwise associated with being an organization.
 
+Can also be set via *MicroProfile Config API* sources, e.g. the environment variable ``DATAVERSE_PERSONORORG_ORGPHRASEARRAY``.
 
 .. _dataverse.api.signature-secret:
 
--- a/src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
+++ b/src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
@@ -263,6 +263,11 @@ public enum JvmSettings {
     //CSL CITATION SETTINGS
     SCOPE_CSL(PREFIX, "csl"),
     CSL_COMMON_STYLES(SCOPE_CSL, "common-styles"),
+
+    // PersonOrOrgUtil SETTINGS
+    SCOPE_PERSONORORG(PREFIX, "personOrOrg"),
+    ASSUME_COMMA_IN_PERSON_NAME(SCOPE_PERSONORORG, "assumeCommaInPersonName"),
+    ORG_PHRASE_ARRAY(SCOPE_PERSONORORG, "orgPhraseArray"),
     ;
 
     private static final String SCOPE_SEPARATOR = ".";
--- a/src/main/java/edu/harvard/iq/dataverse/util/PersonOrOrgUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/PersonOrOrgUtil.java
@@ -1,15 +1,11 @@
 package edu.harvard.iq.dataverse.util;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.logging.Logger;
 
-import jakarta.json.JsonArray;
+import edu.harvard.iq.dataverse.settings.JvmSettings;
 import jakarta.json.JsonObject;
 import jakarta.json.JsonObjectBuilder;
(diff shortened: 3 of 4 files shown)


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3093,6 +3093,8 @@ Please note that this setting is experimental.
 The Schema.org metadata and OpenAIRE exports and the Schema.org metadata included in DatasetPages try to infer whether each entry in the various fields (e.g. Author, Contributor) is a Person or Organization. If you are sure that
 users are following the guidance to add people in the recommended family name, given name order, with a comma, you can set this true to always assume entries without a comma are for Organizations. The default is false.
 
+Can also be set via *MicroProfile Config API* sources, e.g. the environment variable ``DATAVERSE_PERSONORORG_ASSUMECOMMAINPERSONNAME``.
```
pdurbin, line 3096: Hmm, `ASSUMECOMMAINPERSONNAME` is a bit hard to read. I understand you're keeping it the same as the old setting: `dataverse.personOrOrg.assumeCommaInPersonName`.

What sort of options do we have? Can we add underscores like `ASSUME_COMMA_IN_PERSON_NAME` and still have backward compatibility? (I'm not sure if there's backward compatibility with the `ASSUMECOMMAINPERSONNAME` anyway.)

Also, can we please have a release note snippet?

pdurbin (commented):
Thanks for the pull request! I left some comments.

doc/sphinx-guides/source/installation/config.rst
vera: I've now changed the names of the two settings to kebab case, which is what they should be called according to the naming conventions (https://guides.dataverse.org/en/latest/developers/configuration.html#adding-a-jvm-setting).

This means they can now be set using `DATAVERSE_PERSON_OR_ORG_ASSUME_COMMA_IN_PERSON_NAME` and `DATAVERSE_PERSON_OR_ORG_ORG_PHRASE_ARRAY`. (Or `asadmin create-jvm-options '-Ddataverse.person-or-org.assume-comma-in-person-name=true` and `asadmin create-jvm-options '-Ddataverse.person-or-org.org-phrase-array=Org,GmbH'`). I've tested to confirm all of these options work.

For backwards compatibility, adding an alias so that the old name `asadmin create-jvm-options '-Ddataverse.personOrOrg.assumeCommaInPersonName=true'` can still be used was easy.

However, for the `orgPhraseArray` setting, the expected value format has changed as well. Previously, a list like `["Org","GmbH"]` was expected, but now it's a list like `Org,GmbH`. Afaik it wouldn't be as easy to still support parsing the old value format. So I haven't added an alias for that option.

Is it OK to drop support for the old `orgPhraseArray` setting and require admins to migrate option name + value format? Considering it's documented as an experimental setting.

Also, I've just added a release note snippet that documents all this as well.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -263,6 +263,11 @@ public enum JvmSettings {
     //CSL CITATION SETTINGS
     SCOPE_CSL(PREFIX, "csl"),
     CSL_COMMON_STYLES(SCOPE_CSL, "common-styles"),
+
+    // PersonOrOrgUtil SETTINGS
+    SCOPE_PERSONORORG(PREFIX, "person-or-org", "dataverse.personOrOrg"),
```
qqmyers: Mostly curious - is the alias here needed?

src/main/java/edu/harvard/iq/dataverse/util/PersonOrOrgUtil.java
```diff
@@ -73,6 +61,7 @@ public static JsonObject getPersonOrOrganization(String name, boolean organizati
 
         boolean isOrganization = !isPerson && Organizations.getInstance().isOrganization(name);
         if (!isOrganization) {
+            String[] orgPhrases = JvmSettings.ORG_PHRASE_ARRAY.lookupOptional(String[].class).orElse(new String[]{});
```
qqmyers: It appears that the shift from a static block to looking these up as needed is just to be able to use the @JvmSetting annotation in tests? Putting the code here means these lookups will be done ~1M times for a /reExportAll command for Harvard Dataverse. Could/should we just call updated versions of the static methods in the tests?

qqmyers (approved):
Overall, switching to MPConfig this way looks fine. I do wonder whether removing the static initialization is worth it.

ofahimIQSS:
it looks like readthedocs job failed

pdurbin:
Here's the failure:

Warning, treated as error:
/Users/PDurbin/github/iqss/dataverse/doc/sphinx-guides/source/admin/metadataexport.rst:68:undefined label: 'dataverse.personororg.assumecommainpersonname'

pdurbin:
I fixed the Sphinx errors in 9a5a102.

pdurbin:
@landreev Jim and I would like you to take a look, please. This might result in a lot of calls on export.

pdurbin:
@vera can you please resolve merge conflicts?

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
vera: Hm, I am not sure but since I don't see any other scopes with aliases in that file, I guess not. It could be removed.

vera:
> @vera can you please resolve merge conflicts?

Yes, done!

src/main/java/edu/harvard/iq/dataverse/util/PersonOrOrgUtil.java
pdurbin: We discussed this PR today and we're still concerted about these ~1 million calls.

@landreev is on vacation for a week or so but we'd like him to take a look when he gets back.

landreev: I agree that we should avoid making these lookups in normal operations, if at all possible. Sorry I missed this.

qqmyers: @vera - can you revert the change here that and go back to having a static initialization and setX methods for use in testing? I'll move this to In Progress until you're finished.

vera: @qqmyers done :)

qqmyers (approved):
Looks good - thanks for reverting the dynamic lookup.

ofahimIQSS:
tests passed - merging PR

PR #11486, opened 2025-05-12
Harvesting client improvements: configurable delay between GetRecord calls; a fix for a problem with long-running DataCite harvests

### LINKED ISSUE #11473: Add a setting for configuring a rate limit in Harvesting Client (a limit on outgoing calls)
Since individual OAI records are generally very small and fast to both download and import, it is very easy for a harvesting client to generate a very large number of calls to the remote OAI server in a short period of time. 
This option will allow harvesting from a server that enforces hard rate limits on incoming calls. 
Should be very easy to implement.

landreev:

This is based on a patch that I made a while ago for another Dataverse instance. But it has come handy here at HDV and it may be of benefit to other instances out there. 
The changes are quite straightforward. 

From the accompanying release note: 

A setting has been added for configuring sleep intervals between OAI calls for specific harvesting clients. Making it possible to harvest uninterrupted from servers enforcing rate limit policies. See the configuration guide for details. Additionally, this release fixes a problem with harvesting from DataCite OAI-PMH where initial, long-running harvests were failing on sets with large numbers of records.

- Closes #11473

Create a harvesting client and harvest something. For example, the `controlTestSet` from demo. Confirm that it's working _without_ this new, optional setting. 
Create the setting as described in the guide. If you configure an interval for the client you're testing, there will be an .info message in the log that the specified number of milliseconds will be used as the delay. If you specify it for a different client, confirm that the client you're testing is unaffected and the delay utilized is 0; i.e. no delay. 

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11473-harvesting-client-improvements.md
+++ b/doc/release-notes/11473-harvesting-client-improvements.md
@@ -1,1 +1,5 @@
 A setting has been added for configuring sleep intervals between OAI calls for specific harvesting clients. Making it possible to harvest uninterrupted from servers enforcing rate limit policies. See the configuration guide for details. Additionally, this release fixes a problem with harvesting from DataCite OAI-PMH where initial, long-running harvests were failing on sets with large numbers of records.
+
+## New Database Settings
+
+- :HarvestingClientCallRateLimit


landreev:
Un-drafting this thing.

landreev:
(will sync w/ develop shortly)

landreev:
Synced the branch with develop just now. There were no merge conflicts to resolve however, contrary to what GitHub was saying. 🤔

src/main/java/edu/harvard/iq/dataverse/harvest/client/HarvesterServiceBean.java
```diff
@@ -291,6 +303,14 @@ private void harvestOAIviaListIdentifiers(OaiHandler oaiHandler, DataverseReques
 
             MutableBoolean getRecordErrorOccurred = new MutableBoolean(false);
 
+            if (sleepBetweenCalls > 0) {
```
stevenwinship: Looks like this will sleep before the first record processes which doesn't seem necessary.
Also in the method below. Maybe also make this code block a separate method call so it's not duplicated :)
Maybe change to   
 if (!first && sleepBetweenCalls > 0) {
             first = false;

stevenwinship: I just noticed in the second method that sleeps the sleep comes after the processRecord (also not necessary after the last record processed.

stevenwinship: Maybe it doesn't matter if it's 1 or 2 seconds. I'll leave it up to you

landreev: Not having the code duplicated is a valid suggestion, will do. 

But I don't think it matters much, whether to sleep before or after the GetRecord call is made. In the GetRecord case specifically, that "first" call is actually made as the first in the series of N calls after one ListIdentifiers call has already been made... 
But, most importantly, as you said, nobody will ever be setting these intervals to more than ~ 1 sec. or a fraction thereof; and the task is being performed in the background, asynchronously anyway.

landreev: "in the _series_ of" - autocorrect doing its thing

landreev: dac02028082d1502944954de25af83bda97471f8

landreev:
@stevenwinship Thanks for adding the test - looks good.

doc/release-notes/11473-harvesting-client-improvements.md
```diff
@@ -0,0 +1 @@
+A setting has been added for configuring sleep intervals between OAI calls for specific harvesting clients. Making it possible to harvest uninterrupted from servers enforcing rate limit policies. See the configuration guide for details. Additionally, this release fixes a problem with harvesting from DataCite OAI-PMH where initial, long-running harvests were failing on sets with large numbers of records.
```
pdurbin, line 1:
```suggestion
A setting has been added for configuring sleep intervals between OAI calls for specific harvesting clients. Making it possible to harvest uninterrupted from servers enforcing rate limit policies. See the configuration guide for details. Additionally, this release fixes a problem with harvesting from DataCite OAI-PMH where initial, long-running harvests were failing on sets with large numbers of records.

## New Database Settings

- :HarvestingClientCallRateLimit
```

landreev: thanks!

landreev:
I just killed and deleted the latest Jenkins built, since it was triggered by a change in the release note. 
The last Jenkins job that ran did in fact succeed.

sekmiller:
If you add the setting via the api with a bad json file you still get a success message "Setting :HarvestingClientCallRateLimit added", but when you run the harvest you get the "Disabling all harvesting client delay intervals completely until fixed!" . in the log. Is there any way that we can note this on setting create or put the error in the UI on Harvest instead of just in the log?

landreev:
Hmm. I based this implementation on, or copied it from, how Oliver implemented a similar map setting for tabular ingest size limits. So yes, it is possible to create it with junk values that are only validated when used; similarly to `:TabularIngestSizeLimit`. 

I couldn't think of a completely trivial/cheap way to quickly add validation on create. So I'd like to petition to let it slip. On the rationale that this one is an even more exotic feature than the tab. ingest limits; and something only sysadmins - advanced users, by definition - will have to tinker with; very rarely, if ever. 

Do we _ever_ validate any other settings on create though?

> If you add the setting via the api with a bad json file you still get a success message "Setting :HarvestingClientCallRateLimit added", but when you run the harvest you get the "Disabling all harvesting client delay intervals completely until fixed!" . in the log. Is there any way that we can note this on setting create or put the error in the UI on Harvest instead of just in the log?

PR #11491, opened 2025-05-15
Feat/10134

### LINKED ISSUE #10134: Feature Request/Idea: Linking unpublished datasets

We would like to be able to link the unpublished dataset to other dataverses. Of course the dataset should only become visible in these dataverses once it is actually published.

API user, depositor

We would like to allow users who are creating a dataset to specify which dataverses they would like the dataset to appear in while they are creating the dataset. This makes the process easier because users don't have to go back and do the linking later once the dataset is published.

Currently, linking a unpublished dataset fails:

```
$ curl -H "X-Dataverse-key: $API_TOKEN" -X PUT http://localhost:8080/api/datasets/$DATASET_ID/link/$DATAVERSE_ALIAS
{"status":"ERROR","message":"Can't link a dataset that has not been published or is not harvested"}
```

https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/LinkDatasetCommand.java#L44-L46

none

did not find any

cc @johannes-darms

vera:

See #10134:

> We would like to be able to link the unpublished dataset to other dataverses. Of course the dataset should only become visible in these dataverses once it is actually published.

- Closes #10134

/

The following tests involve linking of datasets:

`mvn test -Dtest="MoveIT#testMoveLinkedDataset"`
`mvn test -Dtest="DatasetsIT#testCreateDeleteDatasetLink"`
`mvn test -Dtest="DatasetsIT#testLinkingDatasets"`
`mvn test -Dtest="LinkIT"`

Yes, this includes a minor UI change. The "Link" and "Unlink" buttons are now shown on the dataset page even if the dataset is unpublished.

I've added a short release note.

**Additional documentation**:

/


stevenwinship (approved):

ofahimIQSS:
tests are passing, merging.

PR #11492, opened 2025-05-15
feat: allow non-superusers to list the links of a dataset via API

vera:

This PR changes the [API for listing the dataverses a dataset has been linked to](https://guides.dataverse.org/en/latest/admin/dataverses-datasets.html#list-collections-that-are-linked-from-a-dataset) (`api/datasets/$linked-dataset-id/links`) to be accessible for non-superusers.

The UI already shows this list to any authenticated user as part of the "Link Dataset" popup (see screenshot), so I think this is a reasonable change.

image

I've added a test that checks that listing linked dataverses is now possible for non-superusers. The test also confirms that unpublished dataverses are not "leaked" through this list (a linked dataverse is only part of the list if it's either published or the requesting user has permission to view that unpublished dataverse).

/

/

I've added a test for this change: `mvn test -Dtest="LinkIT#testListLinks"`

/

I will add a short release note.

**Additional documentation**:

/


stevenwinship (approved):

ofahimIQSS:
image

looking good, merging.

PR #11495, opened 2025-05-15
Fix DDI Exception

qqmyers:
**What this PR does / why we need it**: The recently merged #11417 caused an exception in the logs: 
  javax.xml.stream.XMLStreamException: Trying to write END_DOCUMENT when document has no root (ie. trying to output empty document).
        at com.ctc.wstx.sw.BaseStreamWriter.throwOutputError(BaseStreamWriter.java:1596)
        at com.ctc.wstx.sw.BaseStreamWriter.reportNwfStructure(BaseStreamWriter.java:1625)
        at com.ctc.wstx.sw.BaseStreamWriter._finishDocument(BaseStreamWriter.java:1451)
        at com.ctc.wstx.sw.BaseStreamWriter.close(BaseStreamWriter.java:247)
        at edu.harvard.iq.dataverse.export.DDIExporter.exportDataset(DDIExporter.java:63)

This was not noticed during review/QA. Investigating, it appears that the reason for the exception is that the XMLStreamWriter was never used (and thus closing it raised the empty doc exception). This PR fixes the problem by removing the unused writer.

- Closes #

**Special notes for your reviewer**: Perhaps this means that there was indeed a resource leak in this class that removing the writer will now fix.

**Suggestions on how to test this**: re-export a dataset or publish a dataset and verify there's no exception in the logs.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/export/DDIExporter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/export/DDIExporter.java
@@ -47,25 +47,11 @@ public class DDIExporter implements XMLExporter {
 
     @Override
     public void exportDataset(ExportDataProvider dataProvider, OutputStream outputStream) throws ExportException {
-        XMLStreamWriter xmlw = null;
-        //XMLStreamWriter is not auto-closable - can't use try-with-resources here
         try {
-            xmlw = XMLOutputFactory.newInstance().createXMLStreamWriter(outputStream);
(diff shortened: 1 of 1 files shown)


sekmiller (approved):

ofahimIQSS:
after i deployed PR to internal, getting search engine error on home page. Can we update this PR with latest?

image

PR #11496, opened 2025-05-15
Reduce useless warnings about missing vocab values

qqmyers:
**What this PR does / why we need it**: To support i18n, we have a warning when looking up CVV values if an entry can't be found in a .properties file for the relevant block/language. However, when the request is for the default language, the value in the block tsv file is used as a default and the fact that there's no property file for the default lang, or its out of date, etc. doesn't matter as much since the user sees the same thing regardless.

This PR just tweaks the logging to only warn for the non-default-lang case and switches to fine logging if someone wants to see the default case.

Hopefully a useful reduction in log volume, particularly with custom metadata blocks/non-i18n installations.

Thought there was one but I don't see it.

- Closes #

**Suggestions on how to test this**: Try on a machine with the CAFE block - verify the "Error finding ..." warnings are gone.

**Additional documentation**:


ofahimIQSS:
good fix, not seeing the errors. Merging.

PR #11500, opened 2025-05-19
Parametrized datasource in Keycloak SPI

GPortas:

- Partially closes https://github.com/IQSS/dataverse-frontend/issues/679 

**Additional documentation**:


qqmyers (approved):
I'm not sure any of us can do justice in a review. From what I know of Keycloak, this makes sense and, as noted, it is already working to support beta and qa. I noticed the build failed so I've restarted a run. I think if that passes and QA can briefly test it (create built in users on beta and QA via jsf/api and then verify that the user can login via the SPA (showing that keycloak is checking the two databases on the two different machines), it's good to go.

ofahimIQSS:
looks good, thanks! Going to merge.

PR #11520, opened 2025-05-23
Feature to limit file uploads to datasets

### LINKED ISSUE #11275: Feature Request: (internal request) Add quota-like limit on the number of files in a dataset
Opening the issue per recent discussion with/by request from @sbarbosadataverse. 
The feature has been long overdue and will be most useful to have. 

Short version: thousands+ of files in a Dataset causes all sorts of issues. Most users would be better off with the data repackaged in fewer large bundles. However, we do not make any attempt to stop users from uploading unlimited numbers of files; resulting in problems for them and for us. So, we need to start doing this asap. 

More detailed justification/rationale: Dataverse is not particularly good at handling datasets with large numbers (more than some hundreds, really) of files. It's not likely that the SPA will fully fix, or even dramatically address this issue. (Datasets with thousands or tens of thousands - as the case may be with a few monstrous cases at IQSS - are inherently hard for humans to page through and otherwise unmanageable; and there are inherent performance problems with such datasets in the API as well). In reality, it is exceptionally rare that any depositor has an actual need to have that many separate files (a hard requirement for the individual files to have DOIs would be one such scenario). Most people would be 100% better off with their data repackaged as fewer, larger bundles. But, since we don't make any effort to stop them, users just upload however many files they have without understanding the consequences. 

In part this could be addressed by better education/documentation/warnings. But ultimately, we need a mechanism for enforcing a hard stop when a pre-set limit is reached. In other words, it should be implemented along the same lines as the support for storage quotas by size, that we now have: when the limit is reached, the dataset becomes read-only, until/unless some files are deleted. Just like with size quotas, it should be possible to set this limit for the instance as a whole, with an option to override it by configuring specific limits for sub-collections or even individual datasets, if needed. 

The feature is important enough that I believe investing into adding it to the "old", jsf-based UI is entirely justified. I also believe that the amount of work needed in the UI would be very manageable, since the functionality can be built around/added to the already existing size-based quota limits.

stevenwinship:
**What this PR does / why we need it**: Need a way to prevent thousands of file from being uploaded due to the issues that causes.

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/11275

- Closes #11275

**Suggestions on how to test this**: See IT tests

**Additional documentation**: native-api and config


landreev:
Just a quick status update: 
I did get to look into, build and play with the PR today. Will be adding comments/asking questions tomorrow.

landreev:
Very happy to see this implemented. 
I may have found a small bug: if the limit is set to 20, and I have a dataset with 19 files, it appears that I can upload 2 more files. I cannot add any files to a dataset that's already at 20. (but please re-test/confirm)

src/main/java/edu/harvard/iq/dataverse/DvObject.java
```diff
@@ -140,6 +140,9 @@ public String visit(DataFile df) {
     
     @Column(insertable = false, updatable = false) private String dtype;
 
+    @Column( nullable = true )
+    private Integer datasetFileCountLimit;
```
landreev: It is super important to have this config setting implemented for both collections AND datasets. In real life, I'm guessing it's going to be a somewhat common case where a specific dataset will need to be given a higher limit, because of some respectable reason. It appears to be working consistently when defined on either level. 

However - and this may be penny-pinching, admittedly - I'm wondering if we want this column to be in the DvObject table; seeing how most DvObjects are files. Please at least consider making it a DvObjectContainer-only element. (see dvObjectContainer.storageDriver for an example; it ends up being an extra column in the Dataverse and Dataset tables each).

stevenwinship: I'm confused about the penny-pinching comment. Columns in the database that are null take up no space and therefore add no pennies to all the DvObject Datafiles.

qqmyers: It does have to get loaded into the objects though. Further - it's just odd that a file has a datasetFileCountLimit, which wouldn't be the case if it's on DvObjectContainer.

stevenwinship: Ok. I'll move it

landreev:
The most important feedback I have is for the UI. I am pretty positive the curation will want the limits to be communicated to the user clearly upfront. When we have quotas enabled, the user is told, when in upload mode on the editdatafiles.xhtml page, how much they are still allowed to add. Something along the same lines - you already have N files in the dataset, the limit is M. 

It is our default approach at this point to avoid investing dev. work into the jsf ui. Because "the spa is coming any second now" etc. But this is important; plus, it sounds like authors/depositors will be using the old ui for some foreseeable future.

src/main/java/edu/harvard/iq/dataverse/DvObject.java
landreev: Yeah, phrasing it in terms of objects makes more sense. I just wanted to emphasize that it was possible, even though there is no dedicated table in the db for DvObjectContainer.

stevenwinship: I moved it

landreev:
[edit: I agree that this is not necessary. The limits are included in the main dataset api output; no need for a dedicated api then.]

I have another simple, but important request: 
Could you please add a GET api, something like `/api/datasets/<id>/effectiveFileNumberLimit` (feel free to pick a prettier name, obviously). 
Once this is merged, we'll need to pass this to the SPA team, to have them mirror the functionality there asap. And they will need an API for knowing the limit that's in effect for the user's dataset.

landreev:
I'm still not seeing these numbers shown in the UI on `editdatafiles.xhtml?mode=UPLOAD`:

> The most important feedback I have is for the UI. I am pretty positive the curation will want the limits to be communicated to the user clearly upfront. When we have quotas enabled, the user is told, when in upload mode on the editdatafiles.xhtml page, how much they are still allowed to add. Something along the same lines - you already have N files in the dataset, the limit is M.
> 
> It is our default approach at this point to avoid investing dev. work into the jsf ui. Because "the spa is coming any second now" etc. But this is important; plus, it sounds like authors/depositors will be using the old ui for some foreseeable future.

landreev:
I am still seeing the behavior, both on create, and when adding files to an existing dataset, where I'm allowed to upload any number of files; then, when I click save, the files past the limit are skipped silently, without any feedback for the user. 
If, for some reason, it requires too much work, to be able to drop the extra files while still on the upload page, I could understand the argument that as long as this old JSF UI is functional, we can live with it not being pretty or user-friendly; since it's being retired... etc. 
But, I don't understand why this should be a problem. Because if I go to `editdatafiles.xhtml?mode=UPLOAD` when the dataset is already at the limit, I get very clear feedback when I'm trying to upload more files, and I'm not allowed to do so: 

Screenshot 2025-06-08 at 8 00 13 PM

It should be easy to do the same thing if I still have N files to add before reaching the limit - accept the N files, then reject the remaining ones with the message above. Unless I'm missing something, that is (but the CreateNewDataFiles command is called after each individual file upload is completed).

stevenwinship:
image

stevenwinship:
image

landreev (approved):
Looks good. 
Let's try and get it into 6.7.

ofahimIQSS:
Adding testing evidence: 

`[centos@dataverse-internal bin]$ ./asadmin create-jvm-options "-Ddataverse.files.default-dataset-file-count-limit=3"
Created 1 option(s)
Command create-jvm-options executed successfully.`

image

ofahimIQSS:
I found a way to by pass the file upload limit. - you can currently test this in internal which has max file upload limit to 3. 
1. Create a dataset (using normal user)
2. upload 5 files on dataset creation page
3. on dataset homepage, under files, should only show 3. 
4. Delete 1 or 2 files out of the 3. 
5. Select Upload file
6. Upload a folder - upload a folder with 20 files

Afterwards, I can see all the files from that folder which exceeds the file limit of 3 and I can proceed with publishing the dataset. 

Video Below: 

https://github.com/user-attachments/assets/e03f73b7-2c09-441e-89fb-e469698cfff4

ofahimIQSS:
Looks like the system handles dropbox alright. 

https://github.com/user-attachments/assets/c1bd7f80-84d8-49be-af70-febb6019d5d0

stevenwinship:
Taking a look at this. Thanks for finding this

ofahimIQSS:
great fix, merging!

qqmyers:
Does this make sense: https://github.com/IQSS/dataverse/blob/1c31486ec5c2669ed402b1006978463a4e022acd/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L1995-L1996 - if you can edit the dataset, you can change the limit yourself?

PR #11551, opened 2025-06-04
Added a control to publishing a Dataset and cleaning dataversesubjects table

### LINKED ISSUE #11536: Inconsistency between subject metrics for dataverses and datasets
### Observation

When a dataset is published without a subject, the [/api/info/metrics/dataverses/bySubject](https://demo.dataverse.org/api/info/metrics/dataverses/bySubject) API returns results including `"N/A"`.
However, the [/api/info/metrics/datasets/bySubject](https://demo.dataverse.org/api/info/metrics/datasets/bySubject) API does not include results with `"N/A"`.

This behavior can be reproduced on the demo environment: https://demo.dataverse.org/

### Raised questions

- Should an entry be avoided in the `dataversesubjects` table when a dataset is published without a subject?

- Is it expected for an API response to include the value `"N/A"` even though this value is not part of the `subject` controlled vocabulary?

stevenferey:
Added a control to publishing a Dataset and cleaning the dataverses sujects table having a controlledvocabularyvalue_id corresponding to N/A

- Closes #11536 

**Additional documentation**:


sekmiller:
@stevenferey  There are merge conflicts and, unfortunately this will not make it into version 6.7 of dataverse, so please update the sql script name. Thanks!

sekmiller (approved):
Looks good. The sql script is written for a post-6.7 merge.

ofahimIQSS:
Hello! Can you bump the version in pom.xml to 6.7 please? Thanks!

pdurbin:
> Hello! Can you bump the version in pom.xml to 6.7 please? Thanks!

Done. I just merged the latest from develop.

src/main/resources/db/migration/V6.7.0.1.sql
pdurbin, line 1: This should have been called V6.7.1.2.sql

pdurbin: @stevenferey we had to revert this in PR #11759 because the Flyway script had the wrong name. Can you please create a fresh PR? Thanks!

stevenferey: @pdurbin, I just created PR #11765 .
I hope this works. Thanks.

PR #10598, opened 2024-05-30
8941 adding file count in solr (v2)

luddaniel:
This PR is following the closed PR #9823 with a branch up to date and more details.

**What this PR does / why we need it**:

This PR handles 2 subjects as one go : 
- adding file count of a dataset version directly in solr Document (useful for API Search uses, ex: searching for datasets without file `http://localhost:8080/api/search?q=fileCount:0`)
- moved the work of calculating the file count of a dataset version in indexing part instead of result json building, meaning : speed up the response when using big item number per page (ex : `&per_page=1000`)

_two birds one stone_

**Which issue(s) this PR closes**:

- Closes #8941

**Special notes for your reviewer**:

As the original PR was not merged after 9 month in open state, some competing developments were made on the same lines of code. I just finished the rework after merging develop a0028c3 and noticed that @GPortas was related to those developments. @GPortas can you please be part of reviewers and approve this PR ?

**Suggestions on how to test this**:

Checking Search API results of dataset version with files and no file.

**Is there a release notes update needed for this change?**:

May be not required, tell me.

**Demo**:

https://github.com/IQSS/dataverse/assets/83018819/056801a7-3e37-4e33-9c4e-e4891d3b4f17

Screenshot of API Search result of the same dataset after publishing

Screenshot from 2024-05-30 15-14-31

PR #11555, opened 2025-06-05
feat: datasetCount for dataverses

### LINKED ISSUE #10190: Feature Request/Idea: Filter dataverses by number of datasets

Using the search API `/api/search`, I would like to filter dataverses by number of datasets.

API user

I'd like to find only dataverses that have content, e.g. filter by number of datasets > 0.

Number of datasets in a dataverse should be added to the solr index so it can be used as a filter in the search API.

no

no

vera:

This PR adds a `datasetCount` for each collection to the search index. The count includes all datasets (published, linked or harvested).

This allows users to filter collections using the datasetCount (e.g., `datasetCount:[1000 TO *]`). Also, the value is returned in Dataverse search results via the Search API.

As recommended by @pdurbin I've used the `fileCount` feature PRs as inspiration (https://github.com/IQSS/dataverse/pull/6623 + https://github.com/IQSS/dataverse/pull/10598).

- Closes #10190

Regarding the points commented by @cmbz in the issue (https://github.com/IQSS/dataverse/issues/10190#issuecomment-2835506352):

> * Some subtlety here: what gets counted? (e.g., linked datasets, harvested datasets)

For simplicity, I've decided to count all datasets that would be shown when visiting the collection in the UI. This means that the count includes published, linked and harvested datasets in the collection or in any subcollections.

If there is interest in further subdividing the dataset count, more subcounts could be added in the future (e.g. `publishedDatasetCount`, `linkedDatasetCount`, `harvestedDatasetCount`).

> * Considerations: Does this add overhead to reindexing process?

I've had to add new dataverse indexing calls in the following locations:

* After a dataset is published, the owning dataverse and any linking dataverses are reindexed. (added to this PR after #11491)
* After dataset destruction, previously only the owning dataverse was reindexed. Now, any linking dataverse is also reindexed.
* After linking a dataset to a dataverse, the dataverse is reindexed.
* After a harvesting of datasets is complete, the target dataverse is reindexed.
* After a harvest client is deleted, the target dataverse is reindexed.
* In general, if the reindexed dataverse is not the root dataverse, all of its parents up until the root are also reindexed.

I believe this overhead is not too large, but let me know what you think.

`mvn test -Dtest="SearchIT#testDataverseDatasetCounts"`

/

I wrote a short release note.

**Additional documentation**:

/

### DIFF
--- a/modules/dataverse-parent/pom.xml
+++ b/modules/dataverse-parent/pom.xml
@@ -132,7 +132,7 @@
  
     <properties>
         <!-- This is a special Maven property name, do not change! -->
-        <revision>6.7</revision>
+        <revision>6.7.1</revision>
     
         <target.java.version>17</target.java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>


src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -165,20 +165,22 @@ private List<Dataset> findByOwnerId(Long ownerId, boolean onlyPublished) {
...
                     .setParameter("ownerId", ownerId)
                     .getResultList();
         } else {
             List<Dataset> results = em.createNamedQuery("Dataset.findByOwnerId")
                     .setParameter("ownerId", ownerId).getResultList();
             for (Dataset ds : results) {
-                if (ds.isReleased() && !ds.isDeaccessioned()) {
+                if (onlyPublished && ds.isReleased() && !ds.isDeaccessioned()) {
+                    retList.add(ds.getId());
+                } else if (includeHarvested && ds.isHarvested()) {
```
qqmyers: If onlyPublished=true and includeHarvested=true, this doesn't include harvested because of the else?

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
```diff
@@ -315,6 +315,15 @@ public Future<String> indexDataverse(Dataverse dataverse, boolean processPaths)
         }
         
         solrInputDocument.addField(SearchFields.SUBTREE, dataversePaths);
+
+        if (dataverse.isReleased()) {
+            // Get all datasets published in this dataverse, including harvested datasets
+            int numberOfPublishedDatasets = dataverseService.findAllDataverseDatasetChildren(dataverse.getId(), true, true).size();
```
qqmyers: It might be worth testing this versus a custom query at scale (or, alternatively, to just test this versus the current indexing on a large database). I _think_ this should just result in a proxy list where calling size() won't try to instantiate a real list of dataset objects, but I'm not sure. If it does do extra work compared to just a select count(*) query for the right datasets, it could slow or increase memory requirements significantly.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
vera: I'm fairly sure the code works with onlyPublished=true and includeHarvested=true, because that's how it's called when the count is calculated (https://github.com/IQSS/dataverse/pull/11555/files/6d302bafb42bec487a89ae593ec64a29bc142c61#diff-1a01af3c6f5f9893743e27eb5b74bcf87abadf4aea927690bd9f8d179926349dR321), and the count does include harvested datasets: https://github.com/IQSS/dataverse/pull/11555/files/6d302bafb42bec487a89ae593ec64a29bc142c61

I think I found that `ds.isReleased()` is not true for harvested datasets, because they don't have a publicationDate.

qqmyers: OK - makes sense - could you add a comment about that?

vera: I've refactored the code a bit and added some comments, I think it makes more sense now: https://github.com/IQSS/dataverse/pull/11555/commits/7eee55eefe04b5191c7cfa0cf652791608ef56c8

ofahimIQSS:
to be discussed during next weeks triage.

src/main/java/edu/harvard/iq/dataverse/search/IndexServiceBean.java
vera: I have been testing this with our test system which includes a larger dataverse (around 30k datasets) and I think that you are correct and that I am seeing some significant slowing of API responses which require reindexing of that large dataverse. I will investigate this further and update this PR as soon as I am able.

vera: I've just pushed a commit using a NamedQuery to calculate the datasetCount. With our 30k dataset dataverse, reindexing the dataverse is now speedy again.

stevenwinship:
Please resolve the conflicts.

vera:
@stevenwinship the conflicts are now resolved.

stevenwinship (approved):

PR #11477, opened 2025-05-08
Change CI workflows to release tagged container images for Dataverse

poikilotherm:
**What this PR does / why we need it**:
People demand container images of Dataverse with tags of released versions of Dataverse.
This PR will meet their demands.

Open TODOs before this is ready for review:
- [x] Add a README for src/backports
- [x] Re-enable the "run at IQSS only" guards again once review is done
- [x] Add label in app image with reference to base image tag (not unless we can figure out fabric8io/docker-maven-plugin#1865)
- [x] Update docs on supported images, also add note about scheduled maintenance
- [x] Update devguide/makerelease page about container images
- [x] ~~Preview images for base image~~ -> new issue, new dev cycle.
- [x] Add release note snippet
- [x] Configbaker: add maintenance script, for now using rolling tags only (very low change rate here, not many deps installed)
- [x] Configbaker: pin down the versions of software installed in the image to avoid nasty surprises
- [x] Configbaker: add backports
- [x] Add Renovate Bot config to update the pinned dependencies in containers every Sunday 

**Which issue(s) this PR closes**:

- Closes #10618

**Special notes for your reviewer**:
*It needs to be decided if not running the Maven Unit Test workflow on **push** to master and not on new tags.*
Otherwise, this is only touching container stuff, low risk for anyone else.

**Suggestions on how to test this**:
1. Take a look at the action runs over at https://github.com/gdcc/wip-dataverse-base-image/actions/workflows/container_maintenance.yml. You can also dispatch a workflow run from the dropdown, just make sure to set dry-run mode.
2. You can also execute the CI scripts locally, just use `DRY_RUN=1` to avoid building anything or `DAMP_RUN=1` to see Maven build Docker images but not push them to the Hub. Make sure to call the scripts from the repo's root via `./.github/workflows/scripts/containers/maintain-YYY.sh <branch names you want included>`

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
None.

**Is there a release notes update needed for this change?**:
:battery: included

**Additional documentation**:
Preview docs at https://dataverse-guide--11477.org.readthedocs.build/en/11477/container/index.html

PR #11566, opened 2025-06-09
Don't push images with buildx we haven't build

poikilotherm:

Fixing #11477 last step of pushing images to the Hub. This could not be tested before without messing up Docker Hub images.

- Closes #11477 

Sorry for not catching this earlier - I though skipping the build was enough and would also skip the pushing. Apparently it doesn't.

Merge, then run the maintenance workflow manually, including the force option.


ofahimIQSS:
Looks good, merging!

PR #11589, opened 2025-06-23
Add datasetType to JSON serialization of DatasetVersion

### LINKED ISSUE #11573: Feature Request: add datasetType to `JSONPrinter` of `DatasetVersion`
With the recent support for dataset types, it would be good to learn about the type of a dataset from any version of the dataset retrieved as JSON from the API.

This saves another roundtrip, retrieving the Dataset JSON from the datasetId on its version. Much easier to use for client libs this way!

CC @JR-1991

poikilotherm:

Add `datasetType` to JSON serialization of DatasetVersion, saving an API call roundtrip to retrieve the type within client libs such as easyDataverse and others.

- Closes #11573 

This is very minimalistic. There might be some debate necessary how we shall deal with stuff, as in theory the type of a dataset is not version specific but static like the identifier... The current JSON API is potentially not very well designed to handle all of this.

1. Deploy Dataverse with the change
2. Retrieve a dataset version
3. Watch for the dataset type, it should be "dataset" (as by default, no other types are present)

Nope.

I don't think so.

**Additional documentation**:
None.


poikilotherm:
I added a unit test, merged the latest from develop and did add some docs.

Here's how the docs look like rendered (red box marks the changes):
image

ofahimIQSS:
Looks good - merging!

poikilotherm:
Thanks for merging as always @ofahimIQSS !

PR #11596, opened 2025-06-26
#11546 fix compare separator logic

### LINKED ISSUE #11546: Publication fails with file PIDs enabled for files some files
As first reported in #11494 QA testing, publication can fail for some datasets. The full use case appears to involve instances where file PIDs are enabled, there are files that were created and still in draft before #7334 / v6.4 (and thus did not have a global id yet) but were created before #10790 / v6.6.

The root cause is related to https://github.com/IQSS/dataverse/blob/9e57e73296fd9bdd35cf0ab1ecb550b956a37340/src/main/java/edu/harvard/iq/dataverse/pidproviders/AbstractPidProvider.java#L209-L217 which fails if the separator is non-null and not the same as that of the PidProvider being used, and https://github.com/IQSS/dataverse/blob/7a3964b1aff229a9c73cb56e0cc1d4a254023eae/src/main/resources/db/migration/V6.5.0.4.sql#L1 which set all the existing null separators to the non-null empty string.

In retrospect, that should probably have been limited entries using protocol 'perma'.

* When does this issue occur?
Attempts to publish datasets meeting the criteria above.

* What happens?
*Publication fails and the log shows: `The separator of the DvObject () does not match the configured separator (/)`

v6.6+

**What would fix this?**
A work-around would be to to change the db so that the separator is null. I think something like 
`update dvobject set separator=null where separator='' and identifier is null;`
should work but it's possible that there are cases related to permalinks where that may be too broad.

This could potentially be added as a fix in v6.7 without much work.

A more thorough fix would be to change the code in the AbstractPidProvider. Minimally, the separator checkcould/should only be done if the authority or identifier were non-null. More broadly, this code may not be needed as indicated by the comment before the method - it was used before multiple pid support was added (with the separator part added in #10790 following the example of the earlier code), or it should at least be refactored to only allow a new PID to be based on the existing values when authority and protocol are not null.

My (possibly wrong) understanding of the purpose of the original code was to handle cases where one PID account could mint identifiers with more than one authority and you'd want something like a dependent file to use the authority of the parent dataset if that were possible. With multiple pid providers, I don't think this is necessary any more - one account could be configured with two pid providers, each with a different authority. Whether there's really any case where the code currently sends in a dvobject that has a non-null protocol and authority (but doesn't already have an identifier/PID itself) is TBD - I didn't see any looking quickly.

**Screenshots:**

*

At some point - I see the issue at QDR at least. My guess is doing a little research to confirm the work-around sql change and getting that (or whatever the final version is) into 6.7 would be worthwhile. I might have time for that but would be happy to hand it off to someone else. If anyone wants to look into the more thorough fix, I would be even happier to hand it off.

sekmiller:
**What this PR does / why we need it**: Publishing was failing when separator was empty and didn't match the default separator

- Closes #11546

**Suggestions on how to test this**: create files with file PIDs disabled then enable file PIDs (curl -X PUT -d 'true' http://localhost:8080/api/admin/settings/:FilePIDsEnabled) you should be able to publish and get file PIDs assigned.

**Additional documentation**:


sekmiller:
The Flyway script would have been a bandaid and would not help in the case where files are created while file PIDs are not enabled, and an attempt was made to publish after file PIDs were enabled

ofahimIQSS:
fix looks good. Merging.

Issue #5482, opened 2019-01-23
Error editing dataverse permissions with Postgres 9.3


qqmyers:
As of v4.10.1, QDR has been seeing internal server errors when trying to go to the Edit/Permissions page for Dataverses. I tracked this down to the query at https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java#L245 which was updated with #4883. On our instance, the RoleAssignment objects returned from this call have getDefinitionPoint()==null which causes an NPE at line 248. When I tried using the RoleAssignment.listByAssigneeIdentifier_DefinitionPointId query instead, it worked and the RoleAssignment returned has the correct definition point, so from that, and from inspecting the DB, I don't see any problem with our DB contents. 
I'm not able to reproduce this on demo.dataverse.org. Given that, and the fact that the problem is directly with a db query, I suspect that the problem is an incompatibility with our Postgres version (still at 9.3, planning to upgrade) but I don't know this for sure. I did verify that the updated 4.2.2 driver noted in the 4.9.1 release notes does not seem to help.
We'll see if this is resolved after we upgrade. If so, it means v4.10.1 is not compatible with Postgres 9.3 (and perhaps 9.4, 9.5) and the install prerequisites should be updated. If that's not it, there may be some other aspect of our configuration that is an issue. If this affects anyone else, I'd be interested to hear about it.

pameyer:
@qqmyers On a system with postgres 9.3.25 that I just finished upgrading to data verse 4.10.1 , I don't see any internal server errors on the Edit/Permissions (for dataverses) page.  So there may be another factor in play than the postgres version.

cmbz:
To focus on the most important features and bugs,     we are closing issues created before 2020 (version 5.0)     that are not new feature requests with the label 'Type: Feature'.    

If you created this issue and you feel the team     should revisit this decision, please reopen the issue and leave a comment.

PR #11599, opened 2025-06-30
optimize check for existing role assignment

jo-pol:

Optimize new role assignments.  
We  have over 94K datasets on https://archaeology.datastations.nl/dataverse/root/?q=pan 
Assigning a role to another dataset can take over a minute.

- Closes #

* Create the same role assignment twice. Before and after the fix, the system should react the second time with an invalid request.
* Set a small timeout for API calls (on our VM some 30ms) or measure the spent time in some other way.
* Create a script that loops over creating a new dataset and an API call to create a a new role assignment.
* With an increasing number of datasets the number of timeouts will increase.
* apply the fix
* Run the script again. Assignments will succeed again in time.

no

no

**Additional documentation**:


qqmyers (approved):
This looks good to me. FWIW: I looked at this code w.r.t. #5482 and QDR switched to using the more efficient query that was in the code before #4883 (same query as in this PR)  back in 2019. (QDR switched due to errors we saw, not efficiency, but the efficiency argument makes sense.)

ofahimIQSS:
Hello - can you update PR with latest from dev branch please!

ofahimIQSS:
testing passed - merging PR

PR #11610, opened 2025-07-07
make limit update a superuser only function

stevenwinship:
What this PR does / why we need it: Need a way to prevent thousands of file from being uploaded due to the issues that causes.

Which issue(s) this PR closes: https://github.com/IQSS/dataverse/issues/11275

Closes https://github.com/IQSS/dataverse/issues/11275

Suggestions on how to test this: See IT tests

Does this PR introduce a user interface change? If mockups are available, please link/include them here:

Is there a release notes update needed for this change?: Included

Additional documentation: native-api and config

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
@@ -135,7 +135,6 @@ public class Dataverses extends AbstractApiBean {
     @Path("{identifier}")
     public Response addDataverse(@Context ContainerRequestContext crc, String body, @PathParam("identifier") String parentIdtf) {
         Dataverse newDataverse;
-
         try {
             newDataverse = parseAndValidateAddDataverseRequestBody(body);
         } catch (JsonParsingException jpe) {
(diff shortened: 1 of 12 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -135,8 +135,13 @@ public Response addRoot(@Context ContainerRequestContext crc, String body) {
     @Path("{identifier}")
     public Response addDataverse(@Context ContainerRequestContext crc, String body, @PathParam("identifier") String parentIdtf) {
         Dataverse newDataverse;
+        AuthenticatedUser u;
         try {
-            newDataverse = parseAndValidateAddDataverseRequestBody(body);
+            u = getRequestAuthenticatedUserOrDie(crc);
+            newDataverse = parseAndValidateAddDataverseRequestBody(body, u.isSuperuser());
```
qqmyers: Would it be easier/simpler to just check the resulting newDataverse to see if the limit was set rather than pushing isSuperUser down to the parser?

Also - would it make sense to fail if a non-superuser tries to set the limit rather than silently ignoring that param?

stevenwinship: Changed

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -2385,6 +2385,7 @@ file.message.replaceSuccess=The file has been replaced.
 
 # File Add/Replace operation messages
 file.add.count_exceeds_limit=Number of files can not exceed the maximum number of files allowed for this Dataset ({0}).
+file.dataset.error.set.file.count.limit=Only Superuser can set the Dataset File Count Limit.
```
pdurbin, line 2388:
```suggestion
file.dataset.error.set.file.count.limit=Only a superuser can set the dataset file count limit.
```

ofahimIQSS:
looks good from my end, merging

Issue #10128, opened 2023-11-17
MPCONFIG settings are not easily testable in API tests


pdurbin:
In our API tests (REST Assured), I know how to set database settings at the start of a test and unset them at the end. This allows me to exercise functionality, increase code coverage, etc.

MPCONFIG settings are used increasingly but from what I can tell, it's not easy to test them from our API tests.

A case in point is this PR I'm working on:

- #10122

I created some new MPCONFIG settings and I'd like to make sure they're working. It's easy enough to manually test but how do I vary MPCONFIG settings in API tests?

See also the [related thread](https://dataverse.zulipchat.com/#narrow/stream/379673-dev/topic/MPCONFIG.20in.20API.20tests/near/402729549) in Zulip.

Issue #8013, opened 2021-07-21
request access: when a user requests access to a file and is rejected, that action is saved and should be available via API


sbarbosadataverse:
At this time, rejected users are not saved in any way on the file level permissions page:
1. This removes and doesn't allow the dataset owner to track their actual numbers of data requests on their restricted file/dataset
2. This doesn't provide the user with a list of previous requests they can track and keep a record of rejected users
3. This has been requested in support tickets into Harvard Dataverse repeatedly.

sbarbosadataverse:
@TaniaSchlatter I created this ticket for the request access-rejected user scenario

qqmyers:
FWIW: The database is tracking granted and rejected requests these days, but there's no UI/API way to retrieve them yet. It would be relatively easy to add a parameter to the https://guides.dataverse.org/en/latest/api/dataaccess.html#list-file-access-requests call, e.g. ?includeHistory=true (mirroring the flag name from #11268 for curationStatus) to at least make the info accessible. It should also be possible to add the info to the current UI or SPA, but some guidance would be needed as to where to display it (showing the history for all files in the dataset page (permissions tab) could be a long list, could show it only on the file page? or?).

pdurbin:
Right, as discussed [in Slack](https://iqss.slack.com/archives/C03R1E7T4KA/p1740421436528709?thread_ts=1740417754.612809&cid=C03R1E7T4KA) as of the following PR there are three possible RequestState values that can be saved in the database: CREATED, GRANTED, REJECTED.

- #9599

Yes, it would be great to expose this via API.

cmbz:
2025/03/03
- Would require a backend db change
- Would require UI development

cmbz:
2025/03/03
- Moved to dv backlog column @scolapasta

scolapasta:
Ah, I hadn't realized we were now storing rejected. So yes, if that's the case, no DB change; still needs UI design (especially to handle when large number of rejects exist).

cmbz:
2025-04-09
Next steps for whomever picks up issue
- Confirm APIs don't exist
- Implement support in the API
- Open an SPA issue for the UI part (design and implementation)
Sizing at 30 initially

pdurbin:
Based on @cmbz's comment above I retitled this issue.

[...]

PR #11612, opened 2025-07-07
RoleAccess auditing/history functionality

qqmyers:
**What this PR does / why we need it**: Tracks the history of all RoleAssignment changes on a dataset or file, making it easy to see when roles were assigned and by whom, whether a user has had permission on an item in the past, etc. The history of role assignments is available in the (JSF) UI and via API, e.g. 
image

The API supports JSON and CSV outputs. 

All three cases are handled: assignments on a collection, assignments on a dataset, assignments to one or more files in a dataset. 

- Closes #

* At least through the UI, it looks like you can only assign the filedownloader role to files.  The history mechanism should work if there are other roles possible.
* To simplify the display, granting access to multiple files will group all files where access is granted/revoked at the same time by the same user into one line in the display. Same time currently means within the same minute. 
* Logging and displaying the history are controlled by a feature flag, the functionality is off by default.
* The download buttons in the UI use the API, requesting the CSV version via an Accept "text/csv" header. This required a) using a signedUrl to avoid requiring the experimental api-session-auth flag (this could be a general mechanism to allow the JSF UI to call the api as needed), and B) using Javascript to be able to send the header and write a file (versus just having a link to the raw URL - the browser would call that and show the json output on the screen instead of downloading csv).
* I created some IT tests (as well as local tests) - I don't know how to run them as part of the Jenkins build as they'd require the flag to be true (which I assume we don't want as the default).

**Suggestions on how to test this**: Turn on the flag, assign/revoke various roles, check that the history makes sense. The displays are ordered by revocation date (unrevoked first), then assignment date. Note that either the assign or revoke dates and assigner/revoker fields can be null - the revoked ones because the assignments are active, the assign one when the assignment was made before RA history functionality was turned on.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: see image above.

**Additional documentation**: API and UI docs included, feature flag documented.


doc/release-notes/11612-RAHistory.md
```diff
@@ -0,0 +1,17 @@
+# Role Assignment History Tracking
+
+Dataverse can now track the history of role assignments, allowing administrators to see who assigned or revoked roles, when these actions occurred, and which roles were involved. This feature helps with auditing and understanding permission changes over time.
+
+## Key components of this feature:
+
+- **Feature Flag**: The functionality can be enabled/disabled via the `ROLE_ASSIGNMENT_HISTORY` feature flag
```
pdurbin: Please indicate whether the feature is on or off by default. From below it sounds like it's off by default. Is there a reason why we don't want it on?

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1440,6 +1440,78 @@ The fully expanded example above (without environment variables) looks like this
...
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -H "Accept: application/json" "https://demo.dataverse.org/api/datasets/dvAlias/assignments/history"
+
+The response is a JSON array of role assignment history entries with the following structure for each entry:
+
+.. code-block:: json
+
+  {
+    "definedOn": "1",
```
pdurbin, line 1529: Elsewhere we have both `definitionPointId` and `definitionPointName`:

```
src/test/java/edu/harvard/iq/dataverse/api/DeactivateUsersIT.java
83:                .body("data.traces.roleAssignments.items[0].definitionPointName", equalTo(dataverseAlias))
84:                .body("data.traces.roleAssignments.items[0].definitionPointId", equalTo(dataverseId))
```

Should we consider showing both instead of just the id?

At the very least, should we consider switching to `definitionPointId`, for consistency? It appears above and in the following JsonPrinter method:

    public static JsonObjectBuilder json(RoleAssignment ra) {
        return jsonObjectBuilder()
                .add("id", ra.getId())
                .add("assignee", ra.getAssigneeIdentifier())
                .add("roleId", ra.getRole().getId())
                .add("roleName", ra.getRole().getName())
                .add("_roleAlias", ra.getRole().getAlias())
                .add("privateUrlToken", ra.getPrivateUrlToken())
                .add("definitionPointId", ra.getDefinitionPoint().getId());
    }

Obviously, internal consistency across the JSON and CSV formats for this feature is desired, as well, so we should update both if we make any updates.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4006,6 +4078,135 @@ Upon success, the API will return a JSON response with a success message and the
...
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -H "Accept: text/csv" "https://demo.dataverse.org/api/datasets/3/assignments/history"
+
+The CSV response has column headers mirroring the json entries. They are internationalized (when internationalization is configured).
+
+Note: This feature requires the "role-assignment-history" feature flag to be enabled (see :ref:`feature-flags`).
+
+Dataset Files Role Assignment History
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Get the history of role assignments for the files in a dataset. This API call returns a list of role assignments and revocations for all files in the specified dataset.
```
pdurbin, line 4266: It doesn't show file requests, right? That's what this issue is about:

- #8013

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -4058,6 +4130,135 @@ Upon success, the API will return a JSON response with a success message and the
...
+
+  curl -H "X-Dataverse-key:$API_TOKEN" -H "Accept: text/csv" "$SERVER_URL/api/datasets/$ID/assignments/history"
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -H "Accept: text/csv" "https://demo.dataverse.org/api/datasets/3/assignments/history"
+
+The CSV response has column headers mirroring the json entries. They are internationalized (when internationalization is configured).
```
pdurbin:
```suggestion
The CSV response has column headers mirroring the JSON entries. They are internationalized (when internationalization is configured).
```

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1440,6 +1440,78 @@ The fully expanded example above (without environment variables) looks like this
...
+
+  curl -H "X-Dataverse-key:$API_TOKEN" -H "Accept: text/csv" "$SERVER_URL/api/dataverses/$ID/assignments/history"
+
+The fully expanded example above (without environment variables) looks like this:
+
+.. code-block:: bash
+
+  curl -H "X-Dataverse-key:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -H "Accept: text/csv" "https://demo.dataverse.org/api/dataverses/3/assignments/history"
+
+The CSV response has column headers mirroring the json entries. They are internationalized (when internationalization is configured).
```
pdurbin:
```suggestion
The CSV response has column headers mirroring the JSON entries. They are internationalized (when internationalization is configured).
```

doc/sphinx-guides/source/user/dataverse-management.rst
```diff
@@ -119,14 +119,16 @@ Clicking on Permissions will bring you to this page:
...
-When you access a Dataverse collection's permissions page, you will see three sections:
+When you access a Dataverse collection's permissions page, you will see three or four sections:
 
 **Permissions:** Here you can decide the requirements that determine which types of users can add datasets and sub Dataverse collections to your Dataverse collection, and what permissions they'll be granted when they do so.
 
 **Users/Groups:** Here you can assign roles to specific users or groups, determining which actions they are permitted to take on your Dataverse collection. You can also reference a list of all users who have roles assigned to them for your Dataverse collection and remove their roles if you please.
 
 **Roles:** Here you can reference a full list of roles that can be assigned to users of your Dataverse collection. Each role lists the permissions that it offers.
 
+**Role Assignment History** If enabled, you'll be able to see the history of when roles have been assigned and revoked and by whom.
```
pdurbin, line 130: Should we link to the feature flag here?

src/main/java/edu/harvard/iq/dataverse/DataverseRoleServiceBean.java
```diff
@@ -329,4 +372,164 @@ For a given permission and dataverse Id get all of the roles (built-in or owned
...
+    }
+
+    public static class RoleAssignmentHistoryConsolidatedEntry {
+        private String roleName;
+        private String assigneeIdentifier;
+        private String assignedBy;
+        private Date assignedAt;
+        private String revokedBy;
+        private Date revokedAt;
+        private List<Long> definitionPointIds;  // New field
```
pdurbin:
```suggestion
        private List<Long> definitionPointIds;
```

It won't be new forever. 😄

src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
```diff
@@ -567,7 +584,45 @@ public void setRenderFileMessages(boolean renderFileMessages) {
         this.renderFileMessages = renderFileMessages;
     }
 
+    public String getsignedUrlForRAHistoryCsv() {
```
pdurbin: getSignedUrlForRAHistoryCsv is preferred, right?

src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
```diff
@@ -567,7 +584,45 @@ public void setRenderFileMessages(boolean renderFileMessages) {
...
+        String apiPath = "/api/v1/datasets/" + dataset.getId() + "/files/permissions/history";
```
pdurbin, line 589: I wonder how many of these hard-coded `/v1`'s we have in the codebase. This should probably be a constant so we can find them all some day when we need to.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AssignRoleCommand.java
```diff
@@ -61,21 +64,23 @@ public AssignRoleCommand(PrivateUrlUser privateUrlUser, DataverseRole memberRole
...
+            throw new IllegalCommandException("User " + user.getUserIdentifier() + " is deactivated and cannot be given a role.", this);
         }
-        // TODO make sure the role is defined on the dataverse.
-        RoleAssignment roleAssignment = new RoleAssignment(role, grantee, defPoint, privateUrlToken, anonymizedAccess);
-        return ctxt.roles().save(roleAssignment);
     }
+    if(isExistingRole(ctxt)){
+        throw new IllegalCommandException(BundleUtil.getStringFromBundle("datasets.api.grant.role.assignee.has.role.error"), this);
+    }
+    // TODO make sure the role is defined on the dataverse.
```
pdurbin, line 78: I'm confused about this TODO. Are we worried about custom roles? And I thought roles were system-wide rather than being defined on a collection. (I realize this comment is preexisting and that it was simply moved around in this PR.)

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -2577,7 +2577,7 @@ static Response grantRoleOnDataset(String definitionPoint, String role, String r
                 .post("api/datasets/:persistentId/assignments?key=" + apiToken + "&persistentId=" + definitionPoint);
     }
 
-    static Response revokeRole(String definitionPoint, long doomed, String apiToken) {
+    static Response revokeRoleOnDataverse(String definitionPoint, long doomed, String apiToken) {
```
pdurbin, line 2599: Nice cleanup 👍

src/test/java/edu/harvard/iq/dataverse/api/RoleAssignmentHistoryIT.java
```diff
@@ -0,0 +1,396 @@
...
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import edu.harvard.iq.dataverse.authorization.DataverseRole;
+
+/**
+ * Integration tests for the Role Assignment History API endpoints. 
+ * 
+ * Note: These tests require the role-assignment-history FeatureFlag to be true and are not run in normal builds
```
pdurbin, line 24: Yeah, it's a bummer we can't turn on a feature flag during a test. I opened an issue about this:

- #10128

Obviously, if we ever have this feature turned on by default, we make it so these new tests run in Jenkins by adding it to tests/integration-tests.txt.

doc/release-notes/11612-RAHistory.md
qqmyers: Yes, it's off by default - I chose that primarily to keep the current functionality as the default - seems like most of our flags are in that direction.

doc/sphinx-guides/source/api/native-api.rst
qqmyers: Not sure - 'both' only exist for dataverses (id, alias) whereas datasets/files could use (id, PID (if available)), and the file-level call can actually return a comma-separated list of file ids in the definedOn field.

doc/sphinx-guides/source/api/native-api.rst
qqmyers: Right - this is just when assignments have been made or revoked. Whether a request is granted/denied is related but separate (denied would not have any RA history entries).

src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
qqmyers: fixed - for both this and the ManagePermissionsPage. (Interesting that the call to in in the JSF appears to be case insensitive for that character).

src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
qqmyers: Not sure - I added a comment that these are there specifically because the /v1 (or /v2 someday) has to be included for the signature to validate since the /api/datasets/... call is expanded to have the version in it by the servlet(?) in front of Dataverse.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AssignRoleCommand.java
qqmyers: I don't understand it either - perhaps custom is right. It is an 11 year old comment on a command that does https://github.com/IQSS/dataverse/blob/808d68fbf33364f35450b44fb55ab1676092caa7/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AssignRoleCommand.java#L28

src/main/java/edu/harvard/iq/dataverse/ManageFilePermissionsPage.java
pdurbin: Ah, I didn't know that. Thanks for adding a comment!

doc/release-notes/11612-RAHistory.md
pdurbin: Sure, but you could have not even hidden it behind a feature flag. Is there anything you're worried about? Why not just give this feature to everyone?

doc/sphinx-guides/source/api/native-api.rst
pdurbin: Ok. Maybe we can revisit this when we build this feature in the SPA. Or maybe we won't have to. 🤷

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AssignRoleCommand.java
pdurbin: Bleh. Ok, as I say to my dog, "leave it!" We can circle back another day. 😄

ofahimIQSS:
need an update here to resolve conflicts

ofahimIQSS:
looks good to me - merging

PR #11619, opened 2025-07-08
XML Parsing Updates

qqmyers:
**What this PR does / why we need it**: This PR centralizes creation of the various XML parsers used in Dataverse in a new util class and turns on recommended  options.

- Closes #

**Special notes for your reviewer**: FWIW: For the XMLInputFactory class creation, the XMLInputFactory.IS_COALESCING flag was not consistently true in all uses in the code, but it I saw no code that would require it to be false and I'm guessing that the places it wasn't used, we weren't expecting CData sections.

I also caught/added a final close for some uses of the XMLStreamReader. 

**Suggestions on how to test this**: Nominally this should not change parsing of our expected XML inputs. Aside from unit tests, we should assure that the changes don't break Harvesting (e.g. from DataCite) and ORCID linking/logging.

**Additional documentation**:


qqmyers:
After triggering two flaky test fails - see #11620 for one, the other is a ~timing deadlock trying to delete the dataset while the publishing onSuccess had yet to finish - this has passed all tests and should be ready for review.

ofahimIQSS:
cleared from my end - will do further regression testing on harvesting.

PR #11620, opened 2025-07-08
Fix Cache test failure

qqmyers:
**What this PR does / why we need it**: I just saw the following error on #11619, which doesn't affect the cache:
`CacheFactoryBeanTest.testAuthenticatedUserGettingRateLimited:171 expected: <120> but was: <122>`
This PR should fix that by avoiding parallel execution with other tests that can affect the cache.

- Closes #

**Special notes for your reviewer**: Looking at the test, I see I had earlier added an @ResourceLock("cache") annotation on the test that just failed, but I failed to add the same annotation to the other tests that use the cache. It looks like all three tests need it to avoid the potential test fail.

**Suggestions on how to test this**: Verify that the CacheFactoryBeanTests succeed.

**Additional documentation**:


pdurbin (approved):
Seems like a reasonable fix and tests are passing.

ofahimIQSS:
Tests passed - merging

PR #11622, opened 2025-07-09
OIDC-Brokered Shibboleth users API auth

GPortas:

- Closes #

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
```diff
@@ -995,10 +996,23 @@ public AuthenticatedUser lookupUserByOIDCBearerToken(String bearerToken) throws
...
+        AuthenticatedUser authenticatedUser;
+        if (FeatureFlags.API_BEARER_AUTH_USE_SHIB_USER_ON_ID_MATCH.enabled() && oAuth2UserRecord.hasShibAttributes()) {
+            logger.log(Level.FINE, "OAuth2UserRecord has Shibboleth attributes");
+            String userPersistentId = ShibUtil.createUserPersistentIdentifier(oAuth2UserRecord.getShibIdp(), oAuth2UserRecord.getShibUniquePersistentIdentifier());
+            authenticatedUser = lookupUser(ShibAuthenticationProvider.PROVIDER_ID, userPersistentId);
+            if (authenticatedUser != null) {
+                logger.log(Level.FINE, "Shibboleth user found for the given bearer token");
+                return authenticatedUser;
+            }
+        } else if (FeatureFlags.API_BEARER_AUTH_USE_BUILTIN_USER_ON_ID_MATCH.enabled()) {
```
qqmyers, line 1008: FWIW (not related to the PR's purpose): I wonder if this could/should have the equivalent of  the hasShibAttributes check, i.e if a given login method in Keycloak is intended to match against builtin-users, a hardcoded 'builtin' attribute should be sent. That might, for example, allow username/password login to match a builtin account but stop someone using ORCID or Google etc from matching a builtin account (and perhaps we add recognizing existing ORCID/Google accounts at some point if there are enough to worry about).

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3726,6 +3726,9 @@ please find all known feature flags below. Any of these flags can be activated u
...
       - ``Off``
+    * - api-bearer-auth-use-shib-user-on-id-match
+      - Allows the use of a Shibboleth user account when an identity match is found during API bearer authentication. This feature enables automatic association of an incoming IdP identity with an existing Shibboleth user account, bypassing the need for additional user registration steps. This feature only works when the feature flag ``api-bearer-auth`` is also enabled. **Caution: Enabling this feature flag exposes the installation to potential user impersonation issues depending on the specifics of the IdP configured (For example, if it is configured such that an attacker can create a new account in the IdP, or configured social login account, matching a Dataverse built-in account).**
```
qqmyers: Cut/paste error - I don't thing the caution fits, as minimally this doesn't involve matching the built-in account. It seems to me that the Shib case is stricter in that the idp would truly have to be malicious and lie about the shib provider name to be a threat (or not send expected shib attributes). (see other comment about whether the builtin user case can be made stricter). Otherwise, with a properly configured IdP, a malicious user can't create an account to take over someone else's - if I'm thinking correctly about it. If that's right, the warning might be reduced - Enabling this flag could result in impersonation risks if (and only if) used with a misconfigured IdP.

doc/release-notes/11605-existing-shib-external-users-auth.md
```diff
@@ -0,0 +1,2 @@
+Implemented a new feature flag ``dataverse.feature.api-bearer-auth-use-shib-user-on-id-match``, which allows the use of
+a Shibboleth user account when an identity match is found during API bearer authentication.
```
qqmyers: Just wordsmithing - feel free to change further: the flag 'supports use of the new Dataverse client in instances that have historically allowed login via Shibboleth. Specifically, with this flag, when an OIDC bridge is configured to allow OIDC login with validation by the bridged Shibboleth providers, users with existing Shibboleth-based accounts in Dataverse can login to those existing accounts and thereby maintain access to their existing content and retaining their roles. (For security reasons, Dataverse's current support for direct login via Shibboleth cannot be supported in browser-based clients.)'

qqmyers (approved):
Looks good to me. I made a couple suggestions for documentation changes, and a side comment about a potential improvement for the builtin account matching, but nothing to hold up QA.

doc/sphinx-guides/source/installation/config.rst
GPortas: Makes sense. I just updated it.

src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java
GPortas: It makes a lot of sense. Since I’m going to work on https://github.com/IQSS/dataverse/issues/11645, I can take the change you’re suggesting into consideration, thanks!

ofahimIQSS:
Looks good, I was able to login using my harvard credentials. 
image

PR #11629, opened 2025-07-11
Fix View Styled Citations/ Remove extra doi: from URLs

qqmyers:
**What this PR does / why we need it**: The current implementation of CSL Styles citations adds an extra `doi:` in the DOI URL shown, e.g. 
image

This PR fixes that: 
image

- Closes #

**Suggestions on how to test this**: Look at a CSL citation in the UI before/after, verify that the unnecessary/incorrect `doi:` is removed. Could check with PermaLinks as well.

**Additional documentation**:


ofahimIQSS:
Fix looks good - merging

PR #11630, opened 2025-07-11
Performance: Adding a db statement cache

qqmyers:
**What this PR does / why we need it**: This PR adds a configurable LRU statement cache to the DataSourceProducer class. The size/type are mpconfig changeable - large installations may want a value > 50 (100-200?)

- Closes #

**Special notes for your reviewer**: This is one of several things I found when looking for performance improvements/recommended production payara settings. It's hard to tell without a large test instance whether this in particular is helping, but I've definitely seen an aggregate improvement in concert with the other recommendations for changes in domain.xml (see #11631).

**Suggestions on how to test this**: Perf test with and without

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -930,7 +930,7 @@ Connection & Statement Leaks
      - If enabled, leaked connection will be reclaimed by the pool after connection leak timeout occurs.
      - ``false``
    * - dataverse.db.statement-leak-timeout-in-seconds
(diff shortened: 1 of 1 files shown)


poikilotherm:
Are there metrics we can activate to monitor this?

qqmyers:
Reading more, it looks like the postgres jdbc driver should be doing this already so a second cache won't make sense. Closing...

poikilotherm:
FWIW: we could look into JPA level caching of queries. One of the reasons for the JSF SettingsWrapper was to avoid unnecessary queries to the DB.

PR #11635, opened 2025-07-14
XOAI jars updated with the relevant XMLInputFactory commit from the upstream xoai repo

landreev:

Updates these jar dependencies to address the recently found XML parsing issues 

- Closes #

I've tested harvesting a few random sets in XML (oai_dc or ddi) formats. Wouldn't hurt to double-check and confirm that it's still working. 

**Additional documentation**:


landreev:
Harvested a few things from various sources; seems to be working, un-drafting.

landreev:
Yes, it must have been on account of the cached .m2 versions. Giving the local jars new versions fixed the build - maven updated .m2 from local_lib... etc. (the actual tests are still running; keeping an eye on the Jenkins run)
[edit: tests passed]

ofahimIQSS:
tested https://data.progedo.fr/oai, looks good 1712 files completed transfer without issues. I'll go through a few more sets during regression.

PR #11658, opened 2025-07-18
Improved Featured item efficiency during publication

qqmyers:
**What this PR does / why we need it**: The original code streamed through all files in the latest dataset version to create a list of all file ids, and did so once for each featured item, as a way to see if a datafile used in a FeaturedItem was in the latest version. This PR just checks that directly via df.isInDatasetVersion().

- Closes #

**Special notes for your reviewer**: Just ran across this while hunting for an optimistic lock exception.

**Suggestions on how to test this**: Just regression test that publishing a dataset with FeaturedItems for some of its files deletes the items at publication time if the files are restricted in or deleted from the latest version.

**Additional documentation**:


ofahimIQSS:
regression tested in internal - no issues found. Merging PR

Issue #11648, opened 2025-07-17
Notifications API: remove hard-coded dataset urls from message text


ekraffmiller:
In the Notifications API, the message text sometimes contains a hard-coded reference to the JSF Dataset page. This needs to be more dynamic so that it can be used by the SPA, or other clients that don't use the JSF frontend.

For example, here are some notifications currently defined in the bundle.properties:
```
notification.ingestCompleted=Dataset <a href="/dataset.xhtml?persistentId={0}" title="{1}">{1}</a> has one or more tabular files that completed the <a href="{2}/{3}/user/dataset-management.html#tabular-data-files" title="Tabular Data Files - Dataverse User Guide" target="_blank" rel="noopener">tabular ingest process</a> and are available in archival formats.
notification.ingestCompletedWithErrors=Dataset <a href="/dataset.xhtml?persistentId={0}" title="{1}">{1}</a> has one or more tabular files that are available but are not supported for <a href="{2}/{3}/user/dataset-management.html#tabular-data-files" title="Tabular Data Files- Dataverse User Guide" target="_blank" rel="noopener">tabular ingest</a>.
```

qqmyers:
FWIW: For something like this, I think redefining the API to only give the parameters, e.g. the message type, and the {0}, {1}, {2} parameters ("pid":"doi:10.5072/FK2ABCDEF", "title":"My Best Dataset", "guides_url": ...) and having the SPA (and other clients) manage the text, would be better than trying to add flexibility to generate multiple string variants on the back end. (#11644 is ~similar in that we have another API where we're trying to generate the right message for the UI via an API (in this case needing to know the language) - rather than adding a locale parameter, the API could just send 0 as a result and allow the UI to craft the right text.)

pdurbin:
Sure, let's move text to the SPA but email notifications are sent out by the backend so we can't move them.

I spent an hour or so futzing with code. Here's a draft PR, just for conversation:

- #11665

PR #11665, opened 2025-07-21
introduce "UI" setting for JSF vs SPA

pdurbin:

This is just a draft PR to help us start thinking about how to approach this issue:

- #11648

The first use case is to avoid returning a JSF/xhtml URL to the SPA.

- Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/MailServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/MailServiceBean.java
@@ -281,7 +281,7 @@ public class MailServiceBean implements java.io.Serializable {
         if (emailAddress != null){
            Object objectOfNotification =  getObjectOfNotification(notification);
            if (objectOfNotification != null){
-               String messageText = getMessageTextBasedOnNotification(notification, objectOfNotification, comment, requestor);
+               String messageText = getMessageTextBasedOnNotification(notification, objectOfNotification, comment, requestor, null);
                String subjectText = MailUtil.getSubjectTextBasedOnNotification(notification, objectOfNotification);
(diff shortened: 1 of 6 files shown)


ekraffmiller:
Another thing to consider for this PR: the notification text that is currently being returned is for emails, not for the Notifications tab in the Account Page.  For example, for a new role, the API returns
{
        "id": 45,
        "type": "ASSIGNROLE",
        "subjectText": "Root: You have been assigned a role",
        "messageText": "Hello, \nYou are now Admin for the dataverse \"Scientific Research\" (view at http://localhost:8080/dataverse/test-publish-collection-1753266873375 ).",
        "sentTimestamp": "2025-07-23T10:19:09Z"
      },

But for the Notifications tab, we see a different message:

Screenshot 2025-07-23 at 9 09 05 AM

So to support this, as Jim suggested, the dynamic values to insert into the translation text can be passed as additional parameters.

qqmyers:
FWIW: To handle mail, I think we'd need a global (truly global or per-user) setting for preferred UI rather than a param on specific api calls - otherwise taking the same action from different UIs will result in different messages/one user might get emails pointing to different UIs if their colleagues make changes from different UIs.

It would also be nice to get the URL generation out of the properties files, i.e. by keeping the templates for dataverse/dataset/datafile page generation for different UIs in the db (which might mean new UIs could register their page links some day).

pdurbin:
Closing. This was just an experiment. I got farther in another experiment at https://github.com/IQSS/dataverse/tree/11648-notification-patterns

As mentioned at https://github.com/IQSS/dataverse/issues/11673#issuecomment-3132862895 @GPortas is taking a swing at this. We'll compare notes afterward.

Issue #9063, opened 2022-10-14
New API auth mechanism for SPA frontend requests to APIs (developers only for now)


pdurbin:
## Overview of the Feature Request

In order to use a Single Page Application (SPA) architecture, the SPA (written in React, Vue, Angular, Web Components, etc.) needs to be able to authenticate against Dataverse APIs.

## What kind of user is the feature intended for?

Frontend developers using React or similar.

## What inspired the request?

https://github.com/GPortas/dataverse-react-poc by @GPortas relies on a fork of Dataverse at https://github.com/GPortas/dataverse/tree/session_api_auth that allows a JSESSIONID session cookie to be used to auth against the Dataverse APIs.

To use the words from the README:

"In particular, this PoC focuses on testing the following points:

- New API auth mechanism using JSESSIONID cookie for new front-end requests to the Native API

It is necessary to locally deploy Dataverse with this branch: https://github.com/GPortas/dataverse/tree/session_api_auth

That branch has the JSESSIONID cookie Native API auth implemented, necessary for this PoC."

## Any related code?

If we were to accept the changes as-is, they can be previewed here:

https://github.com/IQSS/dataverse/compare/develop...GPortas:dataverse:session_api_auth

## Any related open or closed issues?

- IQSS/dataverse#7827
- IQSS/dataverse#8987
- IQSS/dataverse-pm#14
- IQSS/dataverse#9051

qqmyers:
This can open any GET calls that have side effects to a CSRF issue - we should assure we don't have any or add other protections before we open the api to session cookies.

pdurbin:
Today in an auth meeting I offered to create an issue but @GPortas just reminded me I already created this one! 😄 

I just added "developers only for now" to emphasize that the new auth mechanism will be off by default and hidden behind a feature flag. Only developers will turn this on for now. Production installations should not turn it on due to CSRF (mentioned above) and possibly other security concerns. What we're trying to do is unblock development of a new React frontend.

pdurbin:
@GPortas regarding the feature flag...

I pointed you toward this example which uses a database setting: https://guides.dataverse.org/en/5.12.1/installation/config.html#allowapitokenlookupviaapi

However, these days we're trying to use MPCONFIG (MicroProfile Config API) instead. @poikilotherm wrote extensive docs here: https://guides.dataverse.org/en/5.12.1/developers/configuration.html

[...]

PR #11681, opened 2025-07-23
11525 retrieve collections a user can create datasets in

### LINKED ISSUE #11519: "My Data" API Uses JSF Session Instead of SPA Auth
**Summary:**
When accessing the "My Data" page via the SPA interface, the API sometimes returns data for the previously authenticated JSF user, not the currently logged-in SPA user. This leads to incorrect data being shown in the "My Data" page, including draft datasets that do not belong to the currently authenticated user.

**Steps to Reproduce:**
1. Log in via JSF UI as User A.
2. Create or confirm presence of draft datasets.
3. Without clearing cookies or ending the JSF session, log in via SPA as User B.
4. Navigate to the "My Data" page in the SPA.
5. Observe the datasets listed.

**Expected Result:**
The page should only display datasets relevant to User B.

Actual Result:
The page displays datasets (e.g., drafts) that belong to User A, due to the API falling back on the JSF session instead of the current SPA context.

### LINKED ISSUE #11525: API endpoint to retrieve the collections a user can create datasets on

This API endpoint is required for the dataset creation page in the SPA, although it may be useful for other use cases as well.

Image

API User

SPA needs

None

New API endpoint to retrieve the collections a user can create datasets on

Possibly https://github.com/IQSS/dataverse/pull/10906

Yes

stevenwinship:
**What this PR does / why we need it**: This API endpoint is required for the dataset creation page in the SPA, for selecting the host collection. It will list any Collections/Dataverses the user can add a Dataset to.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11525 https://github.com/IQSS/dataverse/issues/11519

- Closes #11525
- Closes #11519

**Suggestions on how to test this**: Also see https://github.com/IQSS/dataverse/issues/11519 for instruction on how to reproduce the bug that is also fixed in this PR

**Additional documentation**: included


g-saracca:
@stevenwinship the JSON response will look something like this?
`[ { id: 3, alias: 'foo-collection', name: 'Foo Collection' }, { ... } ]`

stevenwinship:
The JSON Response will look like this:

`{
    "count": 1,
    "items": [
        {
            "id": 1,
            "name": "Root",
            "alias": "root"
        }
    ]
}



{
    "count": 17,
    "items": [
        {
            "id": 1,
            "name": "Root",
            "alias": "root"
        },
        {
            "id": 2,
            "name": "dv8627b6a1",
            "alias": "dv8627b6a1"
        },
`

stevenwinship:
I added 'count' and 'items' to make it easier to add pagination if needed at some point

g-saracca:
That's great, thanks!

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -7840,3 +7840,18 @@ Parameters:
 
 ``per_page`` Number of results returned per page.
 
+MyData Collection List
+----------------------
+
+The MyData Collection List API is used to get a list of the collections an authenticated user can create a Dataset in.
+Param identifier={userName} is used by a superuser to get the collections for a specific user.
+
+A curl example listing collections
```
GPortas:
```suggestion
A curl example listing collections:
```

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -3,21 +3,16 @@
...
 import edu.harvard.iq.dataverse.api.auth.AuthRequired;
+import edu.harvard.iq.dataverse.authorization.Permission;
+import edu.harvard.iq.dataverse.engine.command.impl.GetUserPermittedCollectionsCommand;
 import edu.harvard.iq.dataverse.search.SolrQueryResponse;
 import edu.harvard.iq.dataverse.search.SolrSearchResult;
 import edu.harvard.iq.dataverse.api.AbstractApiBean;
 import edu.harvard.iq.dataverse.authorization.AuthenticationServiceBean;
 import edu.harvard.iq.dataverse.authorization.DataverseRole;
 import edu.harvard.iq.dataverse.authorization.DataverseRolePermissionHelper;
-//import edu.harvard.iq.dataverse.authorization.MyDataQueryHelperServiceBean;
```
GPortas, line 20: Why is this commented? If it is not used, why not removing it?

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -455,7 +446,42 @@ public String retrieveMyDataAsJsonString(
...
         return jsonData.build().toString();
     }
-   
+
+    @GET
+    @AuthRequired
+    @Path(retrieveDataPartialAPIPath + "/collectionList")
+    @Produces("application/json")
+    public String retrieveMyCollectionList(@Context ContainerRequestContext crc, @QueryParam("userIdentifier") String userIdentifier) {
+        if ((session.getUser() != null) && (session.getUser().isAuthenticated())) {
```
GPortas: Session is not suitable for API consumers other than JSF (for example, SPA). In fact, that’s why we created https://github.com/IQSS/dataverse/issues/11519
 — to prevent the API from depending on session cookies.

GPortas: In general this method duplicates logic from `retrieveMyDataAsJsonString`. Please refactor to extract the shared logic and eliminate the duplication.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -455,7 +446,42 @@ public String retrieveMyDataAsJsonString(
...
+                return this.getJSONErrorString(
+                        BundleUtil.getStringFromBundle("dataretrieverAPI.user.not.found", Arrays.asList(userIdentifier)),
+                        null);
+            }
+        }
+        try {
+            JsonObjectBuilder jsonObjBuilder = execCommand(new GetUserPermittedCollectionsCommand(createDataverseRequest(getRequestUser(crc)), searchUser, Permission.AddDataset.name(), true));
+            return jsonObjBuilder.build().toString();
+        } catch (WrappedResponse ex) {
+            return null;
```
GPortas: You will be able to `return wr.getResponse();` if you change the method return type to `Response`. This way we can return the current error message and not lose this error information along the way.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -455,7 +446,42 @@ public String retrieveMyDataAsJsonString(
                                 
         return jsonData.build().toString();
     }
-   
+
+    @GET
+    @AuthRequired
+    @Path(retrieveDataPartialAPIPath + "/collectionList")
+    @Produces("application/json")
+    public String retrieveMyCollectionList(@Context ContainerRequestContext crc, @QueryParam("userIdentifier") String userIdentifier) {
```
GPortas: Please return a Response object, consistent with other API endpoints. This seems to be copy-pasted from `retrieveMyCollectionList`, but that endpoint should also be updated to follow the standard pattern.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
```diff
@@ -46,14 +52,19 @@ public JsonObjectBuilder execute(CommandContext ctxt) throws CommandException {
...
+
         if (collections != null) {
             JsonObjectBuilder job = Json.createObjectBuilder();
-            JsonArrayBuilder jab = Json.createArrayBuilder();
-            for (Dataverse dv : collections) {
-                jab.add(json(dv));
-            }
             job.add("count", collections.size());
-            job.add("items", jab);
+            if (minimalOutput != null && minimalOutput) {
```
GPortas: Use primitive `boolean` instead of `Boolean`. Simpler, and you avoid the null check.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
```diff
@@ -18,6 +17,7 @@
 import java.util.logging.Logger;
 
 import static edu.harvard.iq.dataverse.util.json.JsonPrinter.json;
+import static edu.harvard.iq.dataverse.util.json.JsonPrinter.jsonArray;
 
 @RequiredPermissions({})
 public class GetUserPermittedCollectionsCommand extends AbstractCommand<JsonObjectBuilder> {
```
GPortas: This command should return `Dataverse` objects, not a JsonObjectBuilder. It is the responsibility of `JsonPrinter` to create the `JsonObjectBuilder`. I know this was merged prior to this PR, but please refactor it to follow the proper layering and avoid mixing responsibilities.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
```diff
@@ -46,14 +52,19 @@ public JsonObjectBuilder execute(CommandContext ctxt) throws CommandException {
...
             job.add("count", collections.size());
-            job.add("items", jab);
+            if (minimalOutput != null && minimalOutput) {
+                job.add("items", jsonArray(collections));
+            } else {
+                JsonArrayBuilder jab = Json.createArrayBuilder();
+                for (Dataverse dv : collections) {
+                    jab.add(json(dv));
+                }
+                job.add("items", jab);
```
GPortas: As I mentioned above, the composition of this JSON should be moved to the JsonPrinter class, receiving as a parameter the Dataverse objects returned by this command.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -455,7 +446,42 @@ public String retrieveMyDataAsJsonString(
...
+            searchUser = getUserFromIdentifier(userIdentifier);
+            if (searchUser == null) {
+                return this.getJSONErrorString(
+                        BundleUtil.getStringFromBundle("dataretrieverAPI.user.not.found", Arrays.asList(userIdentifier)),
+                        null);
+            }
+        }
+        try {
+            JsonObjectBuilder jsonObjBuilder = execCommand(new GetUserPermittedCollectionsCommand(createDataverseRequest(getRequestUser(crc)), searchUser, Permission.AddDataset.name(), true));
+            return jsonObjBuilder.build().toString();
```
GPortas: You’ll be able to use the `ok` method from `AbstractApiBean` once you update the method’s response type to `Response`.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
GPortas: As we do in other commands, a class-level explanatory command would be very helpful. At first glance, it’s a bit unclear what the permitted collections are. If these depend on the passed permission name, I would specify it.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -7840,3 +7840,18 @@ Parameters:
 
 ``per_page`` Number of results returned per page.
 
+MyData Collection List
+----------------------
+
+The MyData Collection List API is used to get a list of the collections an authenticated user can create a Dataset in.
+Param identifier={userName} is used by a superuser to get the collections for a specific user.
```
GPortas: The parameter name is `userIdentifier`, not `identifier`.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -7840,3 +7840,18 @@ Parameters:
...
+
+A curl example listing collections
+
+.. code-block:: bash
+
+  export API_TOKEN=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+  export SERVER_URL=https://demo.dataverse.org
+
+  curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/mydata/retrieve/collectionList?identifier="
```
GPortas: `?identifier=` 

The example is missing a value for the query parameter. It would also be useful to include an example call both with and without the query parameter.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -323,6 +323,19 @@ public static JsonObjectBuilder json(Dataverse dv, Boolean hideEmail, Boolean re
         return bld;
     }
 
+    // For UI drop down list. Only needing display name and identifier
```
GPortas: This isn’t really for the UI; it’s intended for any API caller. I would remove this comment and rename the method to clearly indicate that it produces minimal output. Alternatively, you might find a better approach once the responsibility for creating the JsonObjectBuilder is fully moved from the command to this class.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -323,6 +323,19 @@ public static JsonObjectBuilder json(Dataverse dv, Boolean hideEmail, Boolean re
...
+    public static JsonArrayBuilder jsonArray(List<Dataverse> collections) {
+        JsonArrayBuilder jsonArrayBuilder = Json.createArrayBuilder();
+        for (Dataverse dataverse : collections) {
```
GPortas: `for (Dataverse dataverse : collections)
`
Two different concepts here—what should we use, collection or dataverse?

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
```diff
@@ -104,6 +106,120 @@ public void testRetrieveMyDataAsJsonString() {
         assertEquals(200, deleteUserResponse.getStatusCode());
     }
 
+    // Test getting a list of collections that the user can add datasets to
+    @Test
+    public void testRetrieveMyDataCollections() throws InterruptedException {
```
GPortas, line 112: In none of the test cases do we check that the id, name, or alias parameters are returned correctly.

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
```diff
@@ -104,6 +106,120 @@ public void testRetrieveMyDataAsJsonString() {
...
+        items = retrieveMyCollectionListResponse.getBody().jsonPath().getList("items");
+        assertEquals(rootCount + user1DataverseCount + user2DataverseCount, items.size());
+
+        // Superuser gets the list of Dataverses/Collections for bad username
+        retrieveMyCollectionListResponse = UtilIT.retrieveMyCollectionList(superUserApiToken, "badUserName");
+        retrieveMyCollectionListResponse.prettyPrint();
+        retrieveMyCollectionListResponse.then().assertThat()
+                .body("success", equalTo(false))
+                .body("error_message", startsWith("No user found for:"))
+                .statusCode(OK.getStatusCode());
```
GPortas: An important reason to return a `Response` in the endpoint instead of a string is that we will always get the correct error code. We can’t return OK if an error has occurred. This is unhelpful for any client.

doc/release-notes/11525-retrieve-collections-a-user-can-create-datasets-in.md
```diff
@@ -0,0 +1,5 @@
+### New API to retrieve a list of collections that an authenticated user can create a dataset in
+
+The API GET /api/mydata/retrieve/collectionList will return all the dataverse objects that the user select from in the SPA UI
```
GPortas: I think what you mention in the title above should be enough; I would omit this UI-specific detail.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
GPortas: See my other comment in DataRetrieverApiIT.java

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -100,6 +100,10 @@ public class PermissionServiceBean {
     @Inject
     DatasetVersionFilesServiceBean datasetVersionFilesServiceBean;
 
+    private static final String LIST_ALL_DATAVERSES_SUPERUSER_HAS_PERMISSION = """
+        SELECT * FROM DATAVERSE
```
GPortas: I’m a bit concerned that this could become a performance bottleneck, especially when the number of dataverse collections is large. We can make it more efficient by selecting only the fields we actually need when requesting a minimalOutput (as in the endpoint you implemented).

For example, we could create a reduced version of the query for the minimalOutput case, such as:

`SELECT id, name, alias FROM DATAVERSE
`

This should lower I/O, and combined with proper indexing, we could achieve better performance. I’m not sure if it makes sense to change this in the current PR, but at the very least I’d flag it as something Omer should test during QA, particularly with a scenario involving many dataverse collections.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
GPortas: It would be great to include unit tests for this, especially as you refactor the class. Adding them would align with our approach in other commands and ensure better maintainability.

GPortas (changes requested):
My feedback mostly focuses on design and architecture. Let’s stick to proper layering and the Single Responsibility Principle, and ensure API responses are consistent with clear error codes. Each PR is also a good opportunity to reduce technical debt and gradually refactor toward current best practices.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
GPortas: Updating `retrieveMyCollectionList ` may cause backward compatibility issues, so let’s consider keeping the existing endpoints that return String and creating new versions returning Response.

Or, if we think it’s more appropriate to stop supporting the String version, we could simply update it to Response. This is a good discussion topic.

Perhaps updating the existing endpoints is outside the scope of this PR, but at a minimum, we should ensure that the same behavior is not introduced in the new endpoint.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
qqmyers: sounds like #11301 which we reverted due to performance.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: @GPortas @qqmyers 
Doing this will also change the /mydata/retrieve to return a Response instead of a string. I hope that's ok since the json will change. I'll list this as a break to backward compatibility

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
stevenwinship: Modified

doc/release-notes/11525-retrieve-collections-a-user-can-create-datasets-in.md
stevenwinship: Modified

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
stevenwinship: Modified

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
stevenwinship: added a test

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: refactored

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: refactored

stevenwinship: removed session

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: removed bean

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: I refactored it to return a response

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
stevenwinship: Done

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
stevenwinship: unit test added for JsonPrinter.jsonArray

doc/sphinx-guides/source/api/native-api.rst
stevenwinship: Fixed

doc/sphinx-guides/source/api/native-api.rst
stevenwinship: fixed

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: I wasn't able to return a Response for the /retrieve endpoint in this PR due to the rework needed in mydata.js
This can be done in a later issue/pr since it's beyond the scope of this ticket

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
```diff
@@ -3,29 +3,24 @@
...
-import jakarta.json.JsonObjectBuilder;
 
 import java.util.List;
 import java.util.logging.Logger;
 
-import static edu.harvard.iq.dataverse.util.json.JsonPrinter.json;
-
 @RequiredPermissions({})
-public class GetUserPermittedCollectionsCommand extends AbstractCommand<JsonObjectBuilder> {
+public class GetUserPermittedCollectionsCommand extends AbstractCommand<List<Dataverse>> {
```
GPortas, line 36: Still not addressed / replied: https://github.com/IQSS/dataverse/pull/11681#discussion_r2291527903

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -344,8 +344,9 @@ public static JsonObjectBuilder json(Dataverse dv, Boolean hideEmail, Boolean re
         return bld;
     }
 
-    // For UI drop down list. Only needing display name and identifier
-    public static JsonArrayBuilder jsonArray(List<Dataverse> collections) {
+    public static JsonObjectBuilder jsonArray(List<Dataverse> collections) {
+        JsonObjectBuilder job = Json.createObjectBuilder();
+        job.add("count", collections.size());
         JsonArrayBuilder jsonArrayBuilder = Json.createArrayBuilder();
         for (Dataverse dataverse : collections) {
```
GPortas: Still not addressed / replied: https://github.com/IQSS/dataverse/pull/11681/files/3ba39fe1f5ed3641f499fd6ed74c78edc3fae17d..baf0b007e0610a3da5d7242d5ec3597dad7a440d#r2291568279

src/test/java/edu/harvard/iq/dataverse/util/json/JsonPrinterTest.java
```diff
@@ -485,6 +482,29 @@ public void testDatasetWithNondefaultType() {
...
+            Dataverse dv = new Dataverse();
+            dv.setAlias("alias" + i);
+            dv.setName("Alias" + i);
+            dv.setId(i);
+            collections.add(dv);
+        }
+        JsonObjectBuilder job = JsonPrinter.jsonArray(collections);
+        JsonObject result = job.build();
+        assertNotNull(result);
+        System.out.println(result);
```
GPortas: Printing adds noise in test logs. Consider removing or replacing with assertions. If you need to debug, use a logger at debug level instead.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -94,165 +98,53 @@ public class DataRetrieverAPI extends AbstractApiBean {
...
-        
+
         return jsonData.build().toString();
-        
+
     }
 
+    private void verifyAuth (ContainerRequestContext crc, String userIdentifier) throws WrappedResponse {
+        // Handle calls from JSF where the User is in the session
+        boolean requiresApiKey = getRequestApiKey() != null;
```
GPortas: The use of requiresApiKey and the related if check below is unnecessary at this point. The API key, detected by   ([ApiKeyAuthMechanism](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/auth/ApiKeyAuthMechanism.java?utm_source=chatgpt.com)
) is only one of the possible API authentication mechanisms to identify a user, and it is handled within the [CompoundAuthMechanism](https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/auth/CompoundAuthMechanism.java?utm_source=chatgpt.com)
 when filtering an incoming API request before any endpoint is invoked.

By the time getRequestAuthenticatedUserOrDie() is called, the authenticated user should already be resolved—regardless of whether authentication came via API key or any other mechanism handled by CompoundAuthMechanism.

In the case of a session cookie, the situation is different, since the parts of the API that authenticate via session were implemented before the API auth filter with the compound mechanism, and are independent of it.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -94,165 +98,53 @@ public class DataRetrieverAPI extends AbstractApiBean {
...
+        User requestUser = getRequestUser(crc);
+        if (!requiresApiKey && session != null && session.getUser() != null && requestUser instanceof GuestUser) {
```
GPortas: Why do you add `requestUser instanceof GuestUser` ?

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -272,38 +164,16 @@ public String retrieveMyDataAsJsonString(
...
-                authUser = searchUser;
-                OTHER_USER = true;
-            } else {
-                return this.getJSONErrorString(
-                        BundleUtil.getStringFromBundle("dataretrieverAPI.user.not.found", Arrays.asList(userIdentifier)),
-                        null);
-            }
+        try {
+            verifyAuth(crc, userIdentifier);
+            OTHER_USER = !authUser.equals(searchUser);
```
GPortas: Since we have the opportunity, let’s rename this variable from OTHER_USER to otherUser—it’s not a constant, so uppercase doesn’t make sense here.

GPortas (commented):
Thanks for addressing the requested changes.

Almost ready. Just a couple of comments.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: It was there before cthe check for an api key. I'll remove it

stevenwinship: Done

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: Done

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: I'm a little confused by this explanation. This refactor must also fix the bug where SPA running in a second tab causes an existing JSF session to be used. Maybe the best solution is to create 2 separate endpoints (1 for JSF and 1 for SPA and all other api calls).

qqmyers: My understanding (correct me if I'm wrong) is that the Compound auth filter would check for an OIDC token (and api keys) and, if the https://github.com/IQSS/dataverse/blob/6275ea99c0bf59457fc76166536393af0620f147/src/main/java/edu/harvard/iq/dataverse/settings/FeatureFlags.java#L32 flag is true, just using the @AuthRequired annotation on the api call would also pick up the session. However, since we want this api call to also work when that flag isn't on, we need an extra check of the session here (as in edu/harvard/iq/dataverse/api/auth/SessionCookieAuthMechanism) - only if the user is guest after the filter. (and nominally only if the feature flag is false).

stevenwinship: Should we be using this flag?

- A feature flag called "api-session-auth" has been added temporarily to aid in the development of the new frontend (#9063) but will be removed once bearer tokens (#9229) have been implemented. There is a security risk (CSRF) in enabling this flag! Do not use it in production! For more information, see https://guides.dataverse.org/en/5.14/installation/config.html#feature-flags

qqmyers: In production, and this PR: no

stevenwinship: It looks like we clear the session user on logout. I think we need to clear the session user when logging in as well.
I have user1 logged into JSF Dataverse in 1 tab then open a second tab (SPA) and log in as user2. The log statements I added show the session.getUser as still being user1. 

Even logging out of user2 in SPA and logging back in doesn't clear the session. I still see user1

I will add code to check the flag if the user is a guest

qqmyers: We ~do this in DataverseSession - https://github.com/IQSS/dataverse/blob/6275ea99c0bf59457fc76166536393af0620f147/src/main/java/edu/harvard/iq/dataverse/DataverseSession.java#L149-L156 but that may be getting skipped in the SPA/OIDC login/logout.

qqmyers: Checking the flag isn't very relevant - no one should be trying to call these apis with only a session except from the current UI, so when it's a guest user, it should almost always be because we need to get the session user. If the flag is true and an authenticated session exists, the filter would work and we wouldn't have a guest user here.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: Had to add this back. This check determines if the user is a guest  and can fall back on the session.

src/test/java/edu/harvard/iq/dataverse/util/json/JsonPrinterTest.java
```diff
@@ -491,16 +493,127 @@ public void testJsonArrayDataverseCollections() {
...
         JsonArray items = result.getJsonArray("items");
         JsonObject item6 = items.getJsonObject(6);
-        System.out.println(item6);
         assertEquals(6, item6.getInt("id"));
         assertEquals("Alias6", item6.getString("name"));
         assertEquals("alias6", item6.getString("alias"));
     }
 
+    @Test
+    public void testJsonTermsOfUseAndAccess() {
```
GPortas, line 505: I'm not sure why these additions, which don't correspond to this PR, are being added here. It's strange.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
stevenwinship: changed to dataverses

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
stevenwinship: added the comment

// No specific permission is needed to execute this command. To filter the collection a list of permissions are passed in

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
stevenwinship: addressed

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
stevenwinship: addressed

src/test/java/edu/harvard/iq/dataverse/util/json/JsonPrinterTest.java
stevenwinship: removed

src/test/java/edu/harvard/iq/dataverse/util/json/JsonPrinterTest.java
stevenwinship: I didn't add them. I think it was a glitch in the merge from develop. Those lines were already there

stevenwinship: image

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
```diff
@@ -3,29 +3,24 @@
...
 
 import java.util.List;
 import java.util.logging.Logger;
 
-import static edu.harvard.iq.dataverse.util.json.JsonPrinter.json;
-
-@RequiredPermissions({})
-public class GetUserPermittedCollectionsCommand extends AbstractCommand<JsonObjectBuilder> {
+@RequiredPermissions({}) // No specific permission is needed to execute this command. To filter the collection a list of permissions are passed in
+public class GetUserPermittedCollectionsCommand extends AbstractCommand<List<Dataverse>> {
```
GPortas: This is the kind of explanatory comment I was referring to @stevenwinship.

```suggestion
/**
 * Command that retrieves all {@link Dataverse} collections for which a given
 * {@link AuthenticatedUser} has the specified permission.
 * <p>
 * The permission is provided as a string corresponding to one of the names
 * in the {@link Permission} enumeration (e.g. {@code Permission.AddDataset.name()}).
 * If the special value {@code "any"} is passed, all collections for which
 * the user has at least one permission are returned.
 * </p>
 *
 * <p>
 * Example:
 * <pre>
 * new GetUserPermittedCollectionsCommand(request, user, Permission.AddDataset.name());
 * </pre>
 * will return the list of collections where the user can add datasets.
 * </p>
 */
@RequiredPermissions({})
public class GetUserPermittedCollectionsCommand extends AbstractCommand<List<Dataverse>> {
```

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
GPortas: Unit tests for `GetUserPermittedCollectionsCommand` are still missing @stevenwinship.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetUserPermittedCollectionsCommand.java
GPortas: I don’t want this PR to be blocked any longer, but unit tests for this class should be added at some point.

ofahimIQSS:
fix looks good, merging

PR #11268, opened 2025-02-15
IQSS/9247 Curation Status Updates

qqmyers:
**What this PR does / why we need it**: This PR extends the External/Curation Status Label mechanism:
- adding tracking of who creates the status label and when, 
- keeping a history of past statuses
- updating the CSV report to include the creation time and assigner of a status
- updating the getCurationStatus api call to return a JSON object for the status with label, assigner, and create time
- adding an includeHistory query param for these API calls to allow seeing prior statuses
- adding a facet to allow filtering by curation status (for users able to set them)
- adding the creation time to solr as a pdate to support search by time period, e.g. current status set prior to a give date
- standardizing the language around 'curation status' vs 'external status'
- adding a 'curation-status' class to displayed labels to allow styling

**Which issue(s) this PR closes**:

- Closes #9247

**Special notes for your reviewer**:

**Suggestions on how to test this**: Verify that curation statuses can be set/removed when [the overall curation status mechanism is enabled](https://guides.dataverse.org/en/latest/admin/dataverses-datasets.html#configure-a-dataverse-collection-to-allow-use-of-a-given-curation-label-set). Verify that the status can be removed. Verify that status is removed at publication. Verify that the /api/datasets/{id}//curationStatus and /api/datasets/{id}/listCurationStates api calls return correct results, with and without the ?includeHistory=true flag. (Most of this is tested in IT tests as well.)

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**: New facet:
image

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #11685, opened 2025-07-25
Re-fix Curation Status bug

### LINKED ISSUE #11684: curation labels cannot be updated when creationTime is null and the label was set before DV 6.7
If a dataset has curation labels which were set before DV release 6.7, the labels cannot be updated neither API call or GUI.

* When does this issue occur?
If you want to update the curation label via API call or GUI.

On every page the curation label is displayed

* What happens?

All users

The curation label should change.

v6.7

**Screenshots:**

*

qqmyers:
**What this PR does / why we need it**: The bug reported in #11684 was originally seen at QDR in May and fixed in [this commit](https://github.com/IQSS/dataverse/pull/11268/commits/6ebf7ca0ab0ab62c8fef0777064131f347ac6687), which was part of the initial #11268 PR. However, it appears that other changes in v6.7 have made the issue reappear - I can confirm we are seeing it again at QDR. 

The original fix was to add NULLS LAST to the ORDER DESC annotation on the [DatasetVersions curationStatuses list](https://github.com/IQSS/dataverse/blob/b3e1c78013e9125ae5af701d8205f55a6885ccec/src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java#L220-L221). For some reason (we did upgrade to [a newer version of the postgres driver](https://github.com/IQSS/dataverse/blob/b3e1c78013e9125ae5af701d8205f55a6885ccec/modules/dataverse-parent/pom.xml#L153)), NULLS LAST is no longer putting entries with null createtimes last. This PR adds an explicit sort (using a reverse sort to get DESC order, which I think is still true from the annotation) along with a nullsLast sort which is what appears to no longer happen in the initial retrieval of the list.

- Closes #11684

FWIW: I think this is a better approach than[ assigning a 'fake' createtime as suggested in the issue](https://github.com/IQSS/dataverse/issues/11684#issuecomment-3116523822).

**Suggestions on how to test this**: If you can't find a dataset with a pre #11268 curation status (which would have a null createtime in the db), you should be able to add several curationstatus entries and then edit one of them to have a null createtime in the db. Prior to this fix, the entry with null will show as the current status when the page is refreshed. After, it will not.

**Is there a release notes update needed for this change?**: added.

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11685-CurationStatus_fix.md
+++ b/doc/release-notes/11685-CurationStatus_fix.md
@@ -1,0 +1,6 @@
+The updates to support keeping the history of curation status labels added in #11268
+will incorrectly show curation statuses added prior to v6.7 as the current one, regardless of 
+whether newer statuses exist. This PR corrects the problem.
+
+(As a work-around for 6.7 admins can add createtime dates (must be prior to when 6.7 was installed) to the curationstatus table
+ for entries that have null createtimes. The code fix in this version properly handles null dates as indicating older/pre v6.7 curationstatuses.)


stevenwinship (approved):
I verified the order shows correct now

ofahimIQSS:
test done - merging

Issue #11683, opened 2025-07-24
/api/notifications/all fails for user with large notification volume (e.g., dataverseAdmin)


ofahimIQSS:
# Bug Report: `/api/notifications/all` fails for user with large notification volume

## Environment

- **Branch:** `develop` (latest as of July 2025)
- **User:** `dataverseAdmin` (has thousands of notifications)
- **Tested during:** PR testing for [#11664](https://github.com/IQSS/dataverse/pull/11664)
- **API Call:**
  ```bash
  curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/notifications/all"

**Bug Description**
The /api/notifications/all endpoint fails when invoked by a user who has a very large volume of notifications. This issue was discovered while testing [PR #11664](https://github.com/IQSS/dataverse/pull/11664), but it exists in the develop branch independently of that PR.

**Steps to Reproduce**
Log in as a superuser or admin account with thousands of notifications (e.g., dataverseAdmin)

curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/notifications/all"
Observe that the API call fails (e.g., no response or internal server error)

**Expected Behavior**
The API call should return the user's notifications, or at least respond with a subset (paginated or filtered), without crashing or timing out.

**Actual Behavior**
The request fails entirely

**

[11664 server.log](https://github.com/user-attachments/files/21413173/11664.server.log)

**
Relevant stack trace is available in the server.log from internal testing.

**Notes**
This is not a regression introduced by PR #11664

**Suggested Fix**
Add pagination support to /api/notifications/all
Add filtering options (e.g., ?type=..., ?since=...)
Enforce maximum limits and document default behavior for large volumes

pdurbin:
In the stackstrace, @ofahimIQSS and I were wondering what "far" is:

`Caused by: java.lang.NullPointerException: Cannot invoke "edu.harvard.iq.dataverse.FileAccessRequest.getGuestbookResponse()" because "far" is null at edu.harvard.iq.dataverse.MailServiceBean.getMessageTextBasedOnNotification(MailServiceBean.java:457)`

It turns out it's this part of the code:

```
case REQUESTFILEACCESS:
    //Notification to those who can grant file access requests on a dataset when a user makes a request
    DataFile datafile = (DataFile) targetObject;

    pattern = BundleUtil.getStringFromBundle("notification.email.requestFileAccess");
[...]

PR #11704, opened 2025-07-31
11703 return is default property get dataset templates

### LINKED ISSUE #11703: Return isDefault property in Get Dataset Templates endpoint
For the SPA, we need the isDefault property.

g-saracca:
We need an extra field in the Get Dataset Templates response.

- Closes #11703 

I've extended an existing test.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -1569,6 +1569,7 @@ public class JsonPrinter {
         return jsonObjectBuilder()
                 .add("id", template.getId())
                 .add("name", template.getName())
+                .add("isDefault", template.isIsDefaultForDataverse())
                 .add("usageCount", template.getUsageCount())
                 .add("createTime", template.getCreateTime().toString())
                 .add("createDate", template.getCreateDate())
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
@@ -2335,6 +2335,7 @@ public class DataversesIT {
         );
         createTemplateResponse.then().assertThat().statusCode(OK.getStatusCode())
                 .body("data.name", equalTo("Dataverse template"))
+                .body("data.isDefault", equalTo(true))
                 .body("data.usageCount", equalTo(0))
                 .body("data.termsOfUseAndAccess.license.name", equalTo("CC0 1.0"))
                 .body("data.datasetFields.citation.fields.size()", equalTo(1))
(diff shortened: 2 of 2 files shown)


g-saracca:
Moving to On Hold, I found other issues with the API that are beyond my limited knowledge of Java.
- API throws 500 when adding a custom license/dataset terms from the JSF UI, 
`jakarta.ejb.EJBException: Cannot invoke “[edu.harvard.iq](http://edu.harvard.iq/).dataverse.license.License.getId()” because “license” is null`
- API is returning templates from parent dataverses, even if the JSF "Include Templates from Root" checkbox is unchecked.
- With my changes above the `isDefault` property is being returned ok when creating a template from the API but it is always being returned as false even if setting it as default from the JSF UI.

g-saracca:
I have tested the fixes, everything works as expected.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/ListDataverseTemplatesCommand.java
```diff
@@ -26,14 +26,22 @@ public ListDataverseTemplatesCommand(DataverseRequest request, Dataverse dataver
 
     @Override
     public List<Template> execute(CommandContext ctxt) throws CommandException {
-        List<Template> templates = new ArrayList<>();
+        List<Template> availableTemplates = new ArrayList<>(dataverse.getTemplates());
 
-        if (dataverse.getOwner() != null) {
-            templates.addAll(dataverse.getParentTemplates());
+        if (!dataverse.isTemplateRoot() && dataverse.getOwner() != null) {
+            availableTemplates.addAll(dataverse.getParentTemplates());
```
sekmiller, line 32: Do we need to make this recursive for multiple generations of owners? While dv.getOwner() != null...

sekmiller:
Under DataversesIT the test TestCreateAndGetTemplates is failing. It looks like the issue could be that the CreateTemplate api under Dataverses is putting "isDefault" into the DTO, but the conversion of the DTO into the actual template that gets created doesn't do anything with isDefault - and remember you actually have to update the Dataverse object by setting the default template to this newly created template.  That's not something that happens in the jsf ui where you create a template and then press the "Make Default" button to make it the default. Let me know if you have any questions.

g-saracca:
Assigning @GPortas since I opened the PR, but he continued with the work.

GPortas:
@sekmiller 

Thanks. I fixed the CreateTemplateCommand to persist the default template in the Dataverse entity.

I also added a minor fix to the notifications API (not related to the original PR scope).

ofahimIQSS:
Hello! I tested this PR with the dataverseadmin account on internal environment. 
curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/notifications/all?inAppNotificationFormat=true"

 --- This worked 

curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/notifications/all" --- when running this, i got empty flower brackets {}

See log files. The Testing file shows the curl commands I ran along with the second file which is the server logs. 
[Notifications API Testing.log](https://github.com/user-attachments/files/22141933/Notifications.API.Testing.log)
[NotificationsAPI-server.log](https://github.com/user-attachments/files/22141934/NotificationsAPI-server.log)

ofahimIQSS:
https://github.com/IQSS/dataverse/issues/11683 --- existing Notifications API issue

PR #11700, opened 2025-07-30
Revert #11301 and bump version to 6.7.1

pdurbin:
**What this PR does / why we need it**:

After [deploying](https://github.com/IQSS/dataverse.harvard.edu/issues/345) 6.7 to Harvard Dataverse, we found a serious performance problem:

- #11698

The problem is being caused by this PR:

- #11301

We [decided](https://docs.google.com/document/d/13ci_BIgdejyW9d1ogg-xAqGQicyIRbklheYvXqDzuX8/edit?usp=sharing) to create a 6.7.1 hotfix release and revert the problematic PR above.

**Which issue(s) this PR closes**:

- Closes #11698

**Special notes for your reviewer**:

We haven't done a hotfix release in a while. Things to note:

- Yes, we are intentionally merging to master.
- I had to revert #11301 manually. I think I got it right, but please review it carefully.
- This PR contains the normal "bump version and change base image" commits for any release.
- I already bumped the version in Jenkins, so I'm hoping API tests will pass. Please confirm this.
- I'm making a PR to update the "making releases" doc based on this experience: FIXME

**Suggestions on how to test this**:

On a server with a lot of collections (such as the QA server) try bringing up the "create dataset" and "edit dataset" screens on 6.7. This should be quite slow (for me on Harvard Dataverse it timed out after about 2.5 minutes). Compare with this PR.

Also, on 6.7 the "host dataverse" field will sometimes be absent. "Do not show the _host dataverse_ field when there is nothing to choose." As of this PR the "host dataverse" field will always be present. See the release note snippet in #11301 (there's also language about not erasing your input).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

Yes. See above.

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #11705, opened 2025-07-31
when cutting a release, base.image.version needs to be "revision" #11700

pdurbin:
I made a mistake when creating the hotfix branch/PR for 6.7.1:

- #11700

I toggled the base.image.version properly in the parent pom.xml when I should have left it alone. As the comment says just above these lines:

`When cutting a release by changing the revision property above, set this to "${revision}"`

So that's what this PR does. As with the PR that wasn't quite right, this is targeting the master branch.

All this base.image.version stuff comes into play when we publish the release and GitHub Action jobs get kicked off to push images to Docker Hub. We want those images to be made properly, hence this PR.


ofahimIQSS:
Merging

PR #11706, opened 2025-07-31
6.7.1 merge to develop

pdurbin:

Now that we've create a hotfix release, we need the fix itself (reverting a bad PR) and the version bump (to 6.7.1) in the "develop" branch.


sekmiller (approved):
Looks same as master update

PR #11707, opened 2025-07-31
11692 search api bug

### LINKED ISSUE #11692: API to search files within a dataset version not returning expected results

* When does this issue occur?

When I use the API endpoint `GET https://demo.dataverse.org/api/datasets/:persistentId/versions/:latest-published/files?limit=10&offset=0&searchText=100MB&persistentId=doi:10.70122/FK2/VMUDWY` to search the files within a dataset, I expect to find all files containing the string `100MB` in the label for the dataset `https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/VMUDWY`. However, this is not happening.

I'm using the endpoint described above to get the paginated files on a dataset version with the `searchText` query string to search within the dataset version.

(1) the Dataverse guides https://guides.dataverse.org/en/6.6/api/native-api.html#list-files-in-a-dataset on this endpoint says "Filtering by search text is also optionally supported. The search will be applied to the labels and descriptions of the dataset files, to return the files that contain the text searched in one of such fields."

(2) However, when I use the endpoint I experience a different result:
- Search is not done on the description on this endpoint.
- Search in the API is done only on the label field of the file, and it has problems with uppercases. When I search for `compose`  in lowercase and the labels are `compose.yml` and `docker-compose.yml`, it finds the results.
- However, when I search for `Captura` and the label is `Captura de pantalla` it does not find the result.
- If I search `aptura` it does find the file `Captura de pantalla`.

A possible cause for this behavior could be on these lines on the source code: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java#L297 . Dataverse seems to lowercase the search input and do a `LIKE %captura%` on the query.

(3) The misleading detail is that on the Dataverse UI, the search behavior on the dataset detail view is different
- It does search on the description field. So if I search by a word on the description like `awesome` it finds the file.
- It search words from the beginning of the words bypassing lowercase or uppercase. So, If I search `captura` on the UI it finds the file `Captura de pantalla` but it doesn't find the file if I search `aptura`.

This happens to all users using the API endpoint `/api/datasets/:persistentId/versions/:latest-published/files` with the `searchText` parameter.

- Find results even when uppercases are involved. Searching `100MB` or `Captura` should return results
- In order to meet the documentation, searching also on the description would be desirable. Or documentation should be updated.

- 6.7 on demo.dataverse.org

**Screenshots:**

sekmiller:
**What this PR does / why we need it**: Fixes an issue with the get files versions api where the search text was not being compared properly - the target field was not being set to lower case as the search text was. Also adds a search on "description" which is noted in the doc, but not actually implemented in the code.

- Closes #11692

This doesn't address the inconsistency with the solr search ("Advanced Search") which does not return a result if the search term does not include the first letter(s) of the word in the title or description

Try the get files by version api and include a search term:
curl "http://localhost:8080/api/datasets/:persistentId/versions/3.0/files?persistentId=doi:10.5072/FK2/056AZX&searchText=kinsale"

No
probably not. (list under "other bug fixes")

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetVersionFilesServiceBean.java
@@ -294,7 +294,8 @@ public class DatasetVersionFilesServiceBean implements Serializable {
         String searchText = searchCriteria.getSearchText();
         if (searchText != null && !searchText.isEmpty()) {
             searchText = searchText.trim().toLowerCase();
-            predicates.add(criteriaBuilder.like(fileMetadataRoot.get("label"), "%" + searchText + "%"));
+            predicates.add(criteriaBuilder.or(criteriaBuilder.like(criteriaBuilder.lower(fileMetadataRoot.get("label")), "%" + searchText + "%"),
+                    criteriaBuilder.like(criteriaBuilder.lower(fileMetadataRoot.get("description")), "%" + searchText + "%")));
         }
         return criteriaBuilder.and(predicates.toArray(new Predicate[]{}));
     }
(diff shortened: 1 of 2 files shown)


stevenwinship (approved):
Looks good. I tested the change and it looks fine.

ofahimIQSS:
good to go, merging
image

Issue #8374, opened 2022-01-27
as a data provider, I'd like to look up a dataset's available upload mechanisms


donsizemore:
Currently, the wonderful DV-Uploader has no way to determine whether the storage driver housing a given dataset is capable of direct uploads, and so defaults to normal uploads unless instructed to use direct.

Soon, Dataverse will add HttpOverlay store- and Globus support, so this is a placeholder issue for an API endpoint which might allow such a look up externally.

In the particular case of direct-to-S3, this could allow DV-Uploader and other clients to default to direct unless told not to, which would save Dataverse application nodes some serious wear and tear caused by users who may not know the difference.

PR #11716, opened 2025-08-05
feature to extend the get storage driver API

### LINKED ISSUE #11695: Feature Request: API GET Dataset storageDriver extension
Extend the GET endpoint to see current dataset storageDriver (see [docs](https://guides.dataverse.org/en/latest/admin/dataverses-datasets.html#configure-a-dataset-to-store-all-new-files-in-a-specific-file-store), "The current driver can be seen using:"). 
To return not only the name of the storage volumen but the type also.

API User

SPA development.

Extend the response information from this:
```
{
  "status": "OK",
  "data": {
    "message": "s3"
  }
}
```
to this:
```
{
  "status": "OK",
  "data": {
    "name": "s3",
    "type": "not-sure-what-types-there-are"
  }
}
```

stevenwinship:
**What this PR does / why we need it**: SPA needs more information in the Json response to get the storage driver info

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11695

- Closes #11695

**Special notes for your reviewer**: backward incompatible Json response.

**Additional documentation**: changelog updated

### DIFF
--- a/doc/release-notes/11695-change-api-get-storage-driver.md
+++ b/doc/release-notes/11695-change-api-get-storage-driver.md
@@ -1,0 +1,10 @@
+## Get Dataset/Dataverse Storage Driver API
+
+### Changed Json response - breaking change!
+
+The API for getting the Storage Driver info has been changed/extended.
+/api/datasets/{identifier}/storageDriver
+/api/admin/dataverse/{dataverse-alias}/storageDriver
+changed "message" to "name" and added "type" and "label"
+
+See also [the guides](https://dataverse-guide--11664.org.readthedocs.build/en/11664/api/native-api.html#configure-a-dataset-to-store-all-new-files-in-a-specific-file-store), #11695, and #11664.
--- a/doc/sphinx-guides/source/api/changelog.rst
+++ b/doc/sphinx-guides/source/api/changelog.rst
@@ -13,6 +13,7 @@ v6.8
 - For POST /api/files/{id}/metadata passing an empty string ("description":"")  or array ("categories":[]) will no longer be ignored. Empty fields will now clear out the values in the file's metadata. To ignore the fields simply do not include them in the JSON string.
 - For PUT /api/datasets/{id}/editMetadata the query parameter "sourceInternalVersionNumber" has been removed and replaced with "sourceLastUpdateTime" to verify that the data being edited hasn't been modified and isn't stale.
 - For GET /api/dataverses/$dataverse-alias/links the Json response has changed breaking the backward compatibility of the API.
+- For GET /api/admin/dataverse/{dataverse-alias}/storageDriver and /api/datasets/{identifier}/storageDriver the driver name is no longer returned in data.message. This value is now returned in data.name.
 
 v6.7
 ----
--- a/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
@@ -20,6 +20,7 @@ import edu.harvard.iq.dataverse.api.auth.AuthRequired;
 import edu.harvard.iq.dataverse.settings.JvmSettings;
 import edu.harvard.iq.dataverse.util.StringUtil;
 import edu.harvard.iq.dataverse.util.cache.CacheFactoryBean;
+import edu.harvard.iq.dataverse.util.json.JsonPrinter;
 import edu.harvard.iq.dataverse.util.json.NullSafeJsonBuilder;
 import edu.harvard.iq.dataverse.validation.EMailValidator;
 import edu.harvard.iq.dataverse.EjbDataverseEngine;
@@ -2196,7 +2197,7 @@ public class Admin extends AbstractApiBean {
             return wr.getResponse();
         }
         //Note that this returns what's set directly on this dataverse. If null/DataAccess.UNDEFINED_STORAGE_DRIVER_IDENTIFIER, the user would have to recurse the chain of parents to find the effective storageDriver
-        return ok(dataverse.getStorageDriverId());
+        return ok(JsonPrinter.jsonStorageDriver(dataverse.getStorageDriverId()));
     }
     
     @PUT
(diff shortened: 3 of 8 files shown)


g-saracca:
@stevenwinship 
I was testing this with the SPA dockerized environment and both endpoints `/api/datasets/{identifier}/storageDriver` and `/api/admin/dataverse/{dataverse-alias}/storageDriver` are returning:
```
{
    "status": "OK",
    "data": {
        "name": "undefined",
        "type": "Undefined"
    }
}
```
These are the [JVM_ARGS](https://github.com/IQSS/dataverse-frontend/blob/6d08b5eaf14747742075cd5b9b98bcaa49734b24/dev-env/docker-compose-dev.yml#L57) we set for the dockerized environment.

landreev:
I'm glad to see that this hasn't been merged yet. We clearly need this API to also say whether direct upload is enabled on this storage volume; as in something like 
```
{
    "status": "OK",
    "data": {
        "name": "storageVolume",
        "type": "s3",
        "directUpload": false
    }
}
```
The SPA supports direct uploads _only_. In the current implementation, the hard-coded assumption is that any S3 volume has direct upload; this is true in the IQSS prod. environment, but is not universally the case. Case in point - the setup on demo.

donsizemore:
@qqmyers refresh my addled memory here; I couldn't find the issue that I opened about allowing external tools such as dv-uploader to remotely determine whether direct-upload is enabled but this sounds extremely useful.

qqmyers:
possibly #8374 . Maybe less important but probably worth sending whether directdownload is enabled too. I could also see `dataverse.files.<id>.upload-out-of-band, dataverse.files.<id>.ingestsizelimit, and dataverse.files.<id>.url-expiration-minutes`
being relevant for setting up a client. (The other params are either not useful or are sent elsewhere (i.e. part size is sent when you try to upload so you can break things up correctly).

landreev:
> Maybe less important but probably worth sending whether directdownload is enabled too.

Yep, crossed my mind too. And, possibly, the max file size limit as well? (which is also defined per storage volume)
Also, the overall good news is that once we make the output of this API multi-valued, if we ever need to add more information to it further down the road, it's going to be trivial and not going to break anything.

> I couldn't find the issue that I opened about allowing external tools such as dv-uploader to remotely determine whether direct-upload is enabled but this sounds extremely useful.

If that was #8374, we can add it to `closes` statement here (provided this will satisfy everything requested there). 
But, as of now, DvUploader makes that determination by simply requesting the direct upload urls where it will get "Direct upload not supported ..." when that's the case. On the jsf pages (for the purposes of, for example, offering the "Upload Folder" option) this is done via Java methods in the beans. Having a simple API call will of course be the clean solution.

stevenwinship:
@g-saracca 
RE: I was testing this with the SPA dockerized environment and both endpoints /api/datasets/{identifier}/storageDriver and /api/admin/dataverse/{dataverse-alias}/storageDriver are returning: 

For the local docker deployment the storageDriverId is undefined for "root"

Are you asking for the endpoint to return the default in this case?

qqmyers:
These methods let you get any config param for any driver: https://github.com/IQSS/dataverse/blob/de2b48fd095b7c4a91a3f7e8f817868e71679a72/src/main/java/edu/harvard/iq/dataverse/dataaccess/StorageIO.java#L657-L687

landreev:
Yes, this should be perfectly sufficient, as the final definition of "done": 
```
{
    "status": "OK",
    "data": {
        "name": "storageVolume",
        "type": "s3",
        "directUpload": false,
        "directDownload": true
    }
}
```

g-saracca:
Hey @stevenwinship, I was taking a look at this and I think I found an issue.
Get Dataset Storage Driver `/api/datasets/:persistentId/storageDriver` works fine, for the SPA docker dev config returning:
```
  "data": {
    "name": "s3",
    "type": "s3",
    "label": "s3",
    "directUpload": true,
    "directDownload": true
  }
```
But the Get Dataverse Storage Driver, e.g. `/api/admin/dataverse/root/storageDriver` initially is returning:
```
  "data": {
    "name": "undefined",
    "type": "Undefined"
  }
```
If I then go to the JSF Edit Dataverse page and I select Storage "s3" (default selected is "s3 (Default)" in the UI) then I get:
```

  "data": {
    "name": "s3",
    "type": "s3",
    "label": "s3"
  }
```
No `directUpload` or `directDownload` properties.

stevenwinship:
@g-saracca 
Are you adding the ?getEffective=true to the admin request? The endpoint returns the values defined for the specific dataverse, and not the default. "root" has nothing assigned so it will return "undefined" unless you add the qp ?getEffective=true

qqmyers:
FWIW: The code uses getEffective<X> (store, pidprovider, etc) to indicate when you want the one that is active for a given collection/dataset rather than looking for a direct assignment on that object. Maybe getEffective=true would be clearer?

stevenwinship:
@qqmyers I'll make that change

g-saracca:
> @g-saracca Are you adding the ?defaultToOwner=true to the admin request? The endpoint returns the values defined for the specific dataverse, and not the default. "root" has nothing assigned so it will return "undefined" unless you add the qp ?defaultToOwner=true

No, that should be the problem, thanks. 
I guess probably almost every time we are going to be using the endpoint with that query param in true.

stevenwinship:
@g-saracca changed defaultToOwner to getEffective

g-saracca:
All good now using `?getEffective=true`.
```
{
  "status": "OK",
  "data": {
    "name": "s3",
    "type": "s3",
    "label": "s3"
  }
}
```

g-saracca:
@stevenwinship I'm thinking now that could be a good idea to extend the endpoint that lists all available drivers in the installation(`/api/admin/dataverse/storageDrivers`) to also return `name`, `type` and `label`, but I can create a separate issue for that.

g-saracca:
Is there a reason why this is not approved yet?

landreev (approved):
This appears to have fallen through the cracks; I for one assumed (again) that it was merged already, and I take responsibility for that. 
Approving now...

ofahimIQSS:
continuous integrations been failing on this one

ofahimIQSS:
tests now passing - looks good from my end - merging.

PR #11721, opened 2025-08-06
Fix astrophysics RedshiftValue compound field declaration

GPortas:

- Closes #

**Additional documentation**:


GPortas:
@pdurbin To test it, we can reload the updated TSV in the QA instance and verify on the SPA that we can update the fields of the astrophysics metadata block of a dataset without encountering any issues during saving.

ofahimIQSS:
fix is good, merging

PR #11734, opened 2025-08-08
"My Data" API Uses JSF Session Instead of SPA Auth

### LINKED ISSUE #11519: "My Data" API Uses JSF Session Instead of SPA Auth

stevenwinship:

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11519

- Closes #11519

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
+++ b/src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
@@ -11,6 +11,7 @@ import edu.harvard.iq.dataverse.DvObject;
 import edu.harvard.iq.dataverse.DvObjectServiceBean;
 import edu.harvard.iq.dataverse.RoleAssigneeServiceBean;
 import edu.harvard.iq.dataverse.api.auth.AuthRequired;
+import edu.harvard.iq.dataverse.authorization.users.User;
 import edu.harvard.iq.dataverse.search.SolrQueryResponse;
 import edu.harvard.iq.dataverse.search.SolrSearchResult;
 import edu.harvard.iq.dataverse.api.AbstractApiBean;
@@ -142,7 +143,7 @@ public class DataRetrieverAPI extends AbstractApiBean {
     
     
     public boolean isSuperuser(){
-        return (session.getUser() != null) && session.getUser().isSuperuser();
+        return (authUser != null) && authUser.isSuperuser();
     }
     
     private AuthenticatedUser getUserFromIdentifier(String userIdentifier){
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -279,23 +279,26 @@ public String retrieveMyDataAsJsonString(
...
+            authUser = getRequestAuthenticatedUserOrDie(crc);
+        } catch (WrappedResponse e) {
+        }
+        if (authUser == null) {
+            if ((session.getUser() != null) && (session.getUser().isAuthenticated())) {
+                authUser = (AuthenticatedUser) session.getUser();
             }
         }
+        if (authUser == null) {
+            return this.getJSONErrorString(
```
GPortas, line 293: As I mentioned in https://github.com/IQSS/dataverse/pull/11681#discussion_r2291478624, we need to refactor these endpoints so they return a Response instead of String. This is essential to provide clients with manageable responses.

I’m not sure if this is something we should do in this PR, but at least, we should create an action item for this.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -279,23 +279,26 @@ public String retrieveMyDataAsJsonString(
...
-                authUser = getRequestAuthenticatedUserOrDie(crc);
-            } catch (WrappedResponse e) {
-                return this.getJSONErrorString(
-                    BundleUtil.getStringFromBundle("dataretrieverAPI.authentication.required"),
-                    BundleUtil.getStringFromBundle("dataretrieverAPI.authentication.required.opt")
-                );
+        authUser = null;
+        try {
+            authUser = getRequestAuthenticatedUserOrDie(crc);
+        } catch (WrappedResponse e) {
```
GPortas, line 285: Once we change the return type to `Response`, we can simply use `return wr.getResponse();` instead of leaving the catch block empty.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
```diff
@@ -279,23 +279,26 @@ public String retrieveMyDataAsJsonString(
...
+            return this.getJSONErrorString(
+                    BundleUtil.getStringFromBundle("dataretrieverAPI.authentication.required"),
+                    BundleUtil.getStringFromBundle("dataretrieverAPI.authentication.required.opt"));
+        }
 
         // For superusers, the searchUser may differ from the authUser
         AuthenticatedUser searchUser = null;
         // If the user is a superuser, see if a userIdentifier has been specified and use that instead
-        if ((authUser.isSuperuser()) && (userIdentifier != null) && (!userIdentifier.isEmpty())) {
+        if (isSuperuser() && (userIdentifier != null) && (!userIdentifier.isEmpty())) {
```
GPortas, line 301: Since authUser will always be non-null at this point, you can keep the previous `authUser.isSuperuser()` check instead of calling` isSuperuser() `.

And you can delete the  ` isSuperuser` method, since it is only used from here.

doc/release-notes/11519-my-data-api-bug-using-session-auth-for-spa.md
```diff
@@ -0,0 +1,5 @@
+## BUG Fix
+
+Call to API /api/mydata/retrieve was using JSF session data to get the user (if logged in on another browser tab). This caused the other user's data to be returned.
```
GPortas, line 3:
```suggestion
The API endpoint ``/api/mydata/retrieve`` was previously using JSF session data to identify the user. This could result in another user's data being returned if they were logged in via JSF and making requests from a different client within the same domain, such as the SPA. This issue has been resolved: the API now prioritizes Bearer Token authentication, using session cookie authentication only as a fallback for JSF requests.
```

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
GPortas: If you send an invalid bearer token along with a valid session cookie, looking at this code, wouldn’t it fail silently, falling back to using the session cookie? Shouldn’t the flow be stopped as soon as an invalid bearer token is detected?

GPortas (commented):
Some comments to be addressed.

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: @GPortas 
It was refactored in https://github.com/IQSS/dataverse/pull/11681 
I think we should close this PR when that one is merged since it also fixes this issue

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: The new validation handles this
see https://github.com/IQSS/dataverse/pull/11716

stevenwinship:
On HOLD pending https://github.com/IQSS/dataverse/pull/11716

src/main/java/edu/harvard/iq/dataverse/mydata/DataRetrieverAPI.java
stevenwinship: This was removed in https://github.com/IQSS/dataverse/pull/11716

stevenwinship:
Closed with merge of https://github.com/IQSS/dataverse/pull/11716

PR #11735, opened 2025-08-11
reload citation.tsv when upgrading, a follow up to #11632

pdurbin:

#11632 was merged without any upgrade instructions. This will make new installations and existing installations out of sync with each other.

- Closes: none

I don't believe we need to touch the Solr schema.xml or do any reindexing or reexporting.

Reload the block on test servers if we haven't already.

No.

Yes.

**Additional documentation**:

None.


stevenwinship (approved):

ofahimIQSS:
looks good, merging

PR #11741, opened 2025-08-13
11710 find dataverses for linking

### LINKED ISSUE #11710: Feature Request: API Get/Find Dataverses For Dataset and Dataverse Linking

For the SPA Link Dataset feature we need a new API endpoint that returns the available dataverses for linking for the calling user.
After discussing with @sekmiller,  I believe that to replicate the JSF behavior, the new endpoint should receive the query(collection name or alias -- that will be entered by the user) and the dataset identifier that want's to link.
Also instead of filtering from a query, an endpoint that just returns all the dataverses available for linking could work and we could filter in the frontend side.
Sek mentioned this [filterDataversesForLinking](https://github.com/IQSS/dataverse/blob/b446904c01d229722ee53dc3a8a8f52fbbc962b0/src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java#L494) method.

**_Update:_** @ekraffmiller also mentioned that we need similiar info for the dataverse linking. I've updated the title of the issue. See [comment](https://github.com/IQSS/dataverse/issues/11710#issuecomment-3145681862).

The API response could have below example format and return empty array `[]` if nothing found.
```
   [
      {
        id: 1,
        alias: 'dataverse-alias',
        name: 'Dataverse Name'
      },
      {
        id: 4,
        alias: 'foo',
        name: 'Foo'
      }
   ]
```

This is what we want to replicate 👇.

Image

API Users

SPA Q3 Roadmap

N/A

New API endpoint.

This [open PR](https://github.com/IQSS/dataverse/pull/11691) is adding new specific permissions to Link Dataset/Dataverse.

sekmiller:
**What this PR does / why we need it**: Adds a new api endpoint which may be used to retrieve a list of Dataverse Collections to which a given Dataset or Dataverse Collection may be linked. The end point also takes in a search term which currently must be the start of the collections' names.

- Closes #11710 

One piece of new functionality is adding a search feature to the query on permissions - so that you can pass a search term along with a user and permission to further filter the collections returned. 
The also fixes an unrelated test for retrieving a Dataverse Collection's Templates. In an existing database if there are templates created in Root then those templates will be inherited by any collection that is a child of Root that also shares Root's metadata blocks.Since the test asserts that only the newly created template should be available to the collection the inherited template causes a failure. I updated the test so that the child dataverse will have different metadata blocks than Root so that the only template available is the one created in the child as assumed by the test.
Also changed the name of findDvoByIdAndFeaturedItemTypeOrDie since I used it here and it really doesn't depend on the dvObject being a featured item.
Also added a new json object builder for dataverses which only returns the id, alias and name of the dataverse, this was needed to return the response suggested in the ticket.

**Suggestions on how to test this**: Try the new api endpoints and verify that the expected collections are returned.
-The end point also takes in an optional search term which is tested against the collections' names.
- Try it with a blank search term and a large database - near prod if possible to test for performance.
-The user calling this API must have Link Dataset or Link Dataverse permission on the Dataverse Collections returned and given the object type.
-there is also a queryParam for "alreadyLinked" which returns a list of dataverses to which the queried dvObject (either dataset or dataverse) is linked, to be used for unlinking. (Only those collections that the requester has linking permissions will be returned.
- also for performance - test with a superuser who will have perms on all collections.
-If the Collection has already been linked to the given Dataset or Collection, it will not be returned.

no

-included
**Additional documentation**:
docs added under native api

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -40,6 +40,7 @@ import edu.harvard.iq.dataverse.util.MailUtil;
 import edu.harvard.iq.dataverse.workflow.Workflow;
 import edu.harvard.iq.dataverse.workflow.step.WorkflowStepData;
 
+import java.io.IOException;
 import java.util.*;
 import jakarta.json.Json;
 import jakarta.json.JsonArrayBuilder;
@@ -1663,6 +1664,23 @@ public class JsonPrinter {
         }
 
         return jsonArrayBuilder;
+    }
+
+    public static JsonObjectBuilder jsonStorageDriver(String storageDriverId, Dataset dataset) {
+        JsonObjectBuilder jsonObjectBuilder = new NullSafeJsonBuilder();
+        jsonObjectBuilder.add("name", storageDriverId);
+        jsonObjectBuilder.add("type", DataAccess.getDriverType(storageDriverId));
+        jsonObjectBuilder.add("label", DataAccess.getStorageDriverLabelFor(storageDriverId));
+        if (dataset != null) {
+            jsonObjectBuilder.add("directUpload", DataAccess.uploadToDatasetAllowed(dataset, storageDriverId));
+            try {
+                jsonObjectBuilder.add("directDownload", DataAccess.getStorageIO(dataset).downloadRedirectEnabled());
+            } catch (IOException ex) {
+                logger.fine("Failed to get Storage IO for dataset " + ex.getMessage());
+            }
+        }
+
+        return jsonObjectBuilder;
     }
 
     public static JsonArrayBuilder json(List<UserNotification> notifications, AuthenticatedUser authenticatedUser, boolean inAppNotificationFormat) {
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
@@ -1,5 +1,6 @@
 package edu.harvard.iq.dataverse.api;
 
+import edu.harvard.iq.dataverse.dataaccess.DataAccess;
 import edu.harvard.iq.dataverse.util.json.JsonParseException;
 import edu.harvard.iq.dataverse.util.json.JsonParser;
 import edu.harvard.iq.dataverse.util.json.JsonUtil;
@@ -2701,6 +2702,26 @@ public class DataversesIT {
         // Templates retrieval should fail if the user lacks dataverse edit permissions
         getTemplateResponse = UtilIT.getTemplates(dataverseAlias, secondApiToken);
         getTemplateResponse.then().assertThat().statusCode(UNAUTHORIZED.getStatusCode());
+    }
+
+    @Test
+    public void testGetStorageDriver() {
+        Response updatedStorageDriver = UtilIT.getStorageDriver("root", getSuperuserToken(), Boolean.TRUE);
+        updatedStorageDriver.prettyPrint();
+        updatedStorageDriver.then().assertThat()
+                .body("data.name", CoreMatchers.notNullValue())
+                .body("data.type", CoreMatchers.notNullValue())
+                .body("data.label", CoreMatchers.notNullValue())
+                .body("data.directUpload", CoreMatchers.nullValue())
+                .body("data.directDownload", CoreMatchers.nullValue())
+                .statusCode(200);
+
+        // Root without default is undefined
+        updatedStorageDriver = UtilIT.getStorageDriver("root", getSuperuserToken(), null);
+        updatedStorageDriver.prettyPrint();
+        updatedStorageDriver.then().assertThat()
+                .body("data.name", CoreMatchers.equalTo(DataAccess.UNDEFINED_STORAGE_DRIVER_IDENTIFIER))
+                .statusCode(200);
     }
 
     private String getSuperuserToken() {
--- a/doc/release-notes/11695-change-api-get-storage-driver.md
+++ b/doc/release-notes/11695-change-api-get-storage-driver.md
@@ -1,0 +1,12 @@
+## Get Dataset/Dataverse Storage Driver API
+
+### Changed Json response - breaking change!
+
+The API for getting the Storage Driver info has been changed/extended.
+/api/datasets/{identifier}/storageDriver
+/api/admin/dataverse/{dataverse-alias}/storageDriver
+changed "message" to "name" and added "type" and "label"
+
+Also added query param for /api/admin/dataverse/{dataverse-alias}/storageDriver?getEffective=true to recurse the chain of parents to find the effective storageDriver
+
+See also [the guides](https://dataverse-guide--11664.org.readthedocs.build/en/11664/api/native-api.html#configure-a-dataset-to-store-all-new-files-in-a-specific-file-store), #11695, and #11664.
--- a/doc/sphinx-guides/source/admin/dataverses-datasets.rst
+++ b/doc/sphinx-guides/source/admin/dataverses-datasets.rst
@@ -60,6 +60,10 @@ The current driver can be seen using::
 
     curl -H "X-Dataverse-key: $API_TOKEN" http://$SERVER/api/admin/dataverse/$dataverse-alias/storageDriver
 
+Or to recurse the chain of parents to find the effective storageDriver::
+
+    curl -H "X-Dataverse-key: $API_TOKEN" http://$SERVER/api/admin/dataverse/$dataverse-alias/storageDriver?getEffective=true
+
 (Note that for ``dataverse.files.store1.label=MyLabel``, ``store1`` will be returned.)
 
 and can be reset to the default store with::
--- a/doc/sphinx-guides/source/api/changelog.rst
+++ b/doc/sphinx-guides/source/api/changelog.rst
@@ -13,6 +13,7 @@ v6.8
 - For POST /api/files/{id}/metadata passing an empty string ("description":"")  or array ("categories":[]) will no longer be ignored. Empty fields will now clear out the values in the file's metadata. To ignore the fields simply do not include them in the JSON string.
 - For PUT /api/datasets/{id}/editMetadata the query parameter "sourceInternalVersionNumber" has been removed and replaced with "sourceLastUpdateTime" to verify that the data being edited hasn't been modified and isn't stale.
 - For GET /api/dataverses/$dataverse-alias/links the Json response has changed breaking the backward compatibility of the API.
+- For GET /api/admin/dataverse/{dataverse-alias}/storageDriver and /api/datasets/{identifier}/storageDriver the driver name is no longer returned in data.message. This value is now returned in data.name.
 - For PUT /api/dataverses/$dataverse-alias/inputLevels custom input levels that had been previously set will no longer be deleted. To delete input levels send an empty list (deletes all), then send the new/modified list.
 - For GET /api/externalTools and /api/externalTools/{id} the responses are now formatted as JSON (previously the toolParameters and allowedApiCalls were a JSON object and array (respectively) that were serialized as JSON strings) and any configured "requirements" are included.
 
--- a/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Admin.java
@@ -20,6 +20,7 @@ import edu.harvard.iq.dataverse.api.auth.AuthRequired;
 import edu.harvard.iq.dataverse.settings.JvmSettings;
 import edu.harvard.iq.dataverse.util.StringUtil;
 import edu.harvard.iq.dataverse.util.cache.CacheFactoryBean;
+import edu.harvard.iq.dataverse.util.json.JsonPrinter;
 import edu.harvard.iq.dataverse.util.json.NullSafeJsonBuilder;
 import edu.harvard.iq.dataverse.validation.EMailValidator;
 import edu.harvard.iq.dataverse.EjbDataverseEngine;
@@ -2180,7 +2181,8 @@ public class Admin extends AbstractApiBean {
     @GET
     @AuthRequired
     @Path("/dataverse/{alias}/storageDriver")
-    public Response getStorageDriver(@Context ContainerRequestContext crc, @PathParam("alias") String alias) throws WrappedResponse {
+    public Response getStorageDriver(@Context ContainerRequestContext crc, @PathParam("alias") String alias,
+                                     @QueryParam("getEffective") Boolean getEffective) throws WrappedResponse {
         Dataverse dataverse = dataverseSvc.findByAlias(alias);
         if (dataverse == null) {
             return error(Response.Status.NOT_FOUND, "Could not find dataverse based on alias supplied: " + alias + ".");
@@ -2193,8 +2195,12 @@ public class Admin extends AbstractApiBean {
         } catch (WrappedResponse wr) {
             return wr.getResponse();
         }
-        //Note that this returns what's set directly on this dataverse. If null/DataAccess.UNDEFINED_STORAGE_DRIVER_IDENTIFIER, the user would have to recurse the chain of parents to find the effective storageDriver
-        return ok(dataverse.getStorageDriverId());
+
+        if (getEffective != null && getEffective) {
+            return ok(JsonPrinter.jsonStorageDriver(dataverse.getEffectiveStorageDriverId(), null));
+        } else {
+            return ok(JsonPrinter.jsonStorageDriver(dataverse.getStorageDriverId(), null));
+        }
     }
     
     @PUT
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -3690,7 +3690,7 @@ public class Datasets extends AbstractApiBean {
             return error(Response.Status.NOT_FOUND, "No such dataset");
         }
 
-        return response(req -> ok(dataset.getEffectiveStorageDriverId()), getRequestUser(crc));
+        return ok(JsonPrinter.jsonStorageDriver(dataset.getEffectiveStorageDriverId(), dataset));
     }
 
     @PUT
--- a/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
@@ -6119,6 +6119,60 @@ createDataset = UtilIT.createRandomDatasetViaNativeApi(dataverse1Alias, apiToken
     }
 
     @Test
+    public void testSetGetDatasetStorageDriver() {
+        Response createUser = UtilIT.createRandomUser();
+        createUser.then().assertThat().statusCode(OK.getStatusCode());
+        String apiToken = UtilIT.getApiTokenFromResponse(createUser);
+        String username = UtilIT.getUsernameFromResponse(createUser);
+        Response makeSuperUser = UtilIT.makeSuperUser(username);
+        assertEquals(200, makeSuperUser.getStatusCode());
+        Response createDataverse = UtilIT.createRandomDataverse(apiToken);
+        String dataverseAlias = UtilIT.getAliasFromResponse(createDataverse);
+        Response createDataset = UtilIT.createRandomDatasetViaNativeApi(dataverseAlias, apiToken);
+        Integer datasetId = UtilIT.getDatasetIdFromResponse(createDataset);
+
+        Response storageDrivers = UtilIT.listStorageDrivers(apiToken);
+        storageDrivers.prettyPrint();
+        JsonObject data = JsonUtil.getJsonObject(storageDrivers.getBody().asString());
+        String first = data.getJsonObject("data").keySet().iterator().next();
+        String name = data.getJsonObject("data").getString(first);
+
+        Response setDriver = UtilIT.setDatasetStorageDriver(datasetId, first, apiToken);
+        setDriver.prettyPrint();
+        assertEquals(200, setDriver.getStatusCode());
+        Response getDriver = UtilIT.getDatasetStorageDriver(datasetId, apiToken);
+        getDriver.prettyPrint();
+        assertEquals(200, getDriver.getStatusCode());
+        getDriver.then().assertThat()
+                .body("data.name", CoreMatchers.equalTo(name))
+                .body("data.type", CoreMatchers.notNullValue())
+                .body("data.label", CoreMatchers.notNullValue())
+                .body("data.directUpload", CoreMatchers.notNullValue())
+                .body("data.directDownload", CoreMatchers.notNullValue())
+                .statusCode(OK.getStatusCode());
+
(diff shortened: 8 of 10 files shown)


g-saracca:
Hi @sekmiller, thanks for this, I'm not going to review because of my limited Java knowledge but I have 2 comments:
1. Do we need different permissions to check if user can link dataverse and dataset now that they have been updated in vera's PR? 
If so we will need to extend two more endpoints:
- `datasets/userPermissions/[datasetID]` to return a new boolean property named `canLinkDataset`.
- `dataverses/[dataverseIDorAlias]/userPermissions` to return a new boolean property named `canLinkDataverse`.

2. Related to the test fix for retrieving a Dataverse Collection's Templates, I found out that the API is returning templates from parent dataverses, even if the JSF "Include Templates from Root" checkbox is unchecked, do you see the same behavior?

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -732,6 +732,50 @@ Note: you must have "Add Dataset" permission in the given collection to invoke t
 
 .. _featured-collections:
 
+List Dataverse Collections to which a given Dataset or Dataverse Collection may be linked
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The user must provide a search term to limit the list of Dataverse Collections returned. The search term will be compared to the name of the Dataverse Collections and must include the beginning of the Dataverse Collections' names.
```
g-saracca: @sekmiller could we do something different here for the search term?

Instead of adding the search term as a path parameter could we pass it as a query string?
e.g.: "$SERVER_URL/api/dataverses/$ID/$OBJECT_TYPE/linkingDataverses?q=${SEARCH_TERM}"

This way we can have spaces or special characters in the search term without the need of encoding.
The search term could be also optional (omit it to get all results)

sekmiller: @g-saracca , yes that makes sense. The existing method does not allow for a blank search term, but I can work with Leonid to change it.

sekmiller: The reason that we don't allow a blank search term is because we have to iterate through all of the dataverses returned by the query first to see if they've already been linked and second to see if the requestor has permission to link to that dataverse. In my testing with my local db that only has about 5000 dv objects it paused for a good long time when the search term is empty

g-saracca: I understand, but what about adding pagination? lets say a hardcoded `per_page` for now of 10 items. This will improve by a lot the performance. In this way it doesn't matter if you send or not a search query, you will get at most 10 items.

doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -732,6 +732,50 @@ Note: you must have "Add Dataset" permission in the given collection to invoke t
 
 .. _featured-collections:
 
+List Dataverse Collections to which a given Dataset or Dataverse Collection may be linked
```
pdurbin:
```suggestion
List Dataverse Collections to Which a Given Dataset or Dataverse Collection May Be Linked
```

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -101,7 +101,7 @@ WITH grouplist AS (
                   WHERE explicitgroup_authenticateduser.containedauthenticatedusers_id = @USERID
                 )
                         
-                SELECT * FROM DATAVERSE WHERE id IN (
+                SELECT * FROM DATAVERSE WHERE id  IN (
```
GPortas: Revert space

```suggestion
                SELECT * FROM DATAVERSE WHERE id IN (
```

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -157,7 +157,7 @@ AND EXISTS (SELECT id FROM dataverserole WHERE dataverserole.id = roleassignment
                        AND @IPRANGESQL
                      )
                   )
-                )
+                )
```
GPortas, line 164: Revert space
```suggestion
                )
```

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -921,7 +921,8 @@ private boolean hasUnrestrictedReleasedFiles(DatasetVersion targetDatasetVersion
     public List<Dataverse> findPermittedCollections(DataverseRequest request, AuthenticatedUser user, Permission permission) {
         return findPermittedCollections(request, user, 1 << permission.ordinal());
     }
-
+
```
GPortas: Revert extra line

```suggestion
```

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -960,6 +961,7 @@ public List<Dataverse> findPermittedCollections(DataverseRequest request, Authen
         return null;
     }
 
+
```
GPortas: Revert extra line

```suggestion
```

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -505,35 +505,91 @@ public List<Dataverse> filterByAliasQuery(String filterQuery) {
         return ret;
     }
     
-    public List<Dataverse> filterDataversesForLinking(String query, DataverseRequest req, Dataset dataset) {
+    public List<Dataverse> filterDataversesForLinking(String query, DataverseRequest req, DvObject dvo) {
```
GPortas, line 508: It looks like this method is only called from the JSF-related DatasetPage class, and furthermore, only dataset dv objects are ever sent. So, if I'm not mistaken, your changes would never be used. What's the reason for this update, if it's related to JSF and also never gets executed?

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -601,7 +601,7 @@ protected DvObject findDvo(@NotNull final String id) throws WrappedResponse {
      * @throws WrappedResponse
      */
     @NotNull
-    protected DvObject findDvoByIdAndFeaturedItemTypeOrDie(@NotNull final String dvIdtf, String type) throws WrappedResponse {
+    protected DvObject findDvoByIdAndTypeOrDie(@NotNull final String dvIdtf, String type, Boolean testForReleased) throws WrappedResponse {
```
GPortas: `testForReleased` could be a primitive, simply `boolean`.

Also the name `testForReleased` isn't very clear. If what it does is enable the validation for featured items, why did you name it that way?

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1771,6 +1773,58 @@ public Response linkDataverse(@Context ContainerRequestContext crc, @PathParam("
...
+            if (!mergedWithSearch.isEmpty()) {
+                for (Dataverse dv : mergedWithSearch) {
+                    dvBuilder.add(json(dv, true));
+                }
+            }
+            return ok(dvBuilder);
+        } catch (WrappedResponse wr) {
+            return wr.getResponse();
+        } catch (Exception e) {
+            return error(Status.BAD_REQUEST, e.getLocalizedMessage());
```
GPortas: This is too generic. I think if we properly separate the origin of this exception, we can provide a more specific error. A localized exception message isn't user-friendly.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1771,6 +1773,58 @@ public Response linkDataverse(@Context ContainerRequestContext crc, @PathParam("
...
         }
     }
+    
+    @GET
+    @AuthRequired
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("{identifier}/{type}/linkingDataverses")
+    public Response getLinkingDataverseList(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf, @QueryParam("searchTerm") String searchTerm, @PathParam("type") String type) {
+        //first determine what you are linking based on identifier and type
+
```
GPortas, line 1781: This endpoint is handling too much business logic. Let's encapsulate it into a dedicated command, such as `GetLinkingDataverseListCommand` to separate the business logic from the API layer. Once that's done, please add unit tests for the new command to cover the different scenarios. This refactoring will make the endpoint cleaner and more robust, aligning with our standard practice of using commands.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1771,6 +1773,58 @@ public Response linkDataverse(@Context ContainerRequestContext crc, @PathParam("
...
+            DataverseRequest dvReq = new DataverseRequest(requestUser, (IpAddress) null);
+            List<Dataverse> dataversesForLinking;
+            dataversesForLinking = permissionService.findPermittedCollections(dvReq, requestUser, Permission.LinkDataset);
+
+            List<Dataverse> mergedWithSearch = new ArrayList<>();
+            dataversesForLinking = dataverseService.removeUnlinkableDataverses(dataversesForLinking, dvObject);
+            
+            //Only do search lookup if search term is there. Otherwise just include the collections based on perms
+            if (!searchTerm.isEmpty()) {
+                dataversesForLinkingSearch = dataverseService.filterDataversesByNamePattern(searchTerm);
```
GPortas: I notice a potential performance concern here (N+1 query issue). Fetching all permitted collections and then performing a separate lookup for the search term is a performance bottleneck, especially as the number of collections grows.

A more efficient and scalable approach is to push the filtering down to the database level. Instead of executing separate queries and merging the results, we should integrate the searchTerm as an optional WHERE clause directly into the findPermittedCollections query. This allows the database to handle the filtering, returning only the exact result set we need in a single pass.

Furthermore, consider if the removeUnlinkableDataverses logic can also be integrated into the main query. The goal is to perform as much of the filtering as possible within a single, optimized database query, rather than relying on expensive in-memory processing and loops.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1771,6 +1773,58 @@ public Response linkDataverse(@Context ContainerRequestContext crc, @PathParam("
...
+                //search term empty then add all based on perms
+                mergedWithSearch.addAll(dataversesForLinking);
+            }
+            
+            JsonArrayBuilder dvBuilder = Json.createArrayBuilder();
+            if (!mergedWithSearch.isEmpty()) {
+                for (Dataverse dv : mergedWithSearch) {
+                    dvBuilder.add(json(dv, true));
+                }
+            }
```
GPortas, line 1797: JsonPrinter should be fully responsible of building the response, given the entire list.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -290,6 +290,17 @@ public static JsonObjectBuilder json(Workflow wf){
 
         return bld;
     }
+    
+    public static JsonObjectBuilder json(Dataverse dv, Boolean mini) {
```
GPortas: You can use primitive here.

```suggestion
    public static JsonObjectBuilder json(Dataverse dv, boolean minimal) {
```

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -677,6 +677,153 @@ public void testImportDDI() throws IOException, InterruptedException {
...
+        UtilIT.setSuperuserStatus(username, Boolean.FALSE);
+        
+        //should get an empty list because dataset is already linked
+        getLinkableDataverses = UtilIT.getLinkableDataverses("dataset", datasetPersistentId, apiToken, dataverseAliasForLinking);
+        getLinkableDataverses.prettyPrint();
+                getLinkableDataverses.then().assertThat()
+                .statusCode(OK.getStatusCode())
+                .body("data.size()", equalTo(0));
+                
+    //set user api back to super user for cleanup
```
GPortas:
```suggestion
        //set user api back to super user for cleanup
```

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -2285,17 +2432,39 @@ public void testUpdateInputLevelDisplayOnCreateOverride() {
...
 
+        String newName = "New Test Dataverse Name";
+        String newAffiliation = "New Test Dataverse Affiliation";
+        String newDataverseType = Dataverse.DataverseType.TEACHING_COURSES.toString();
+        String[] newContactEmails = new String[]{"new_email@dataverse.com"};
+        String[] newInputLevelNames = new String[]{"geographicCoverage"};
+        String[] newFacetIds = new String[]{"contributorName"};
+        String[] newMetadataBlockNames = new String[]{"citation", "geospatial", "biomedical"};
+
+        // Assert that the error is returned for having both MetadataBlockNames and inheritMetadataBlocksFromParent
```
GPortas: What error are you asserting here? Maybe I am not understanding the goal here...

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -2285,17 +2432,39 @@ public void testUpdateInputLevelDisplayOnCreateOverride() {
...
+    public void testCreateAndGetTemplates() throws JsonParseException  {
         Response createUserResponse = UtilIT.createRandomUser();
         String apiToken = UtilIT.getApiTokenFromResponse(createUserResponse);
 
         Response createSecondUserResponse = UtilIT.createRandomUser();
         String secondApiToken = UtilIT.getApiTokenFromResponse(createSecondUserResponse);
-
+        
+        /*
+        We need to make this a non-inherited metadatablocks so the get template will only get templates from current dv
```
GPortas, line 2604: Probably I need more context to better understand this.

GPortas (changes requested):
Some requested changes and comments.

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
sekmiller: Yes. In my first pass at this for the api, I modified this method to take in either a dataset or dataverse since that was what the api would pass - then I applied the permissions. After discussing with German, I realized that getting the collections the user had permissions on first would be more efficient. I left the method as I had modified it since it was doing what it did before for datasets - and if we wanted the dataverse version to do the same for jsf it would be ready.

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
sekmiller: This is to fix the issue I've mentioned with creating and getting templates - in the original test the newly created collection has the same metadata blocks as root so that when you get the templates for the child collection the template from root gets picked up and the test fails for count (expecting 1; found 2).  I'm changing the metadata blocks in the child collection so that the template from root can't be inherited and the test will only find 1 as expected.

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
sekmiller: see my comment above.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
sekmiller: Not sure what you're going for here - I have the builder with the minimal boolean so that the dv list only has Name, Alias and Id since that's all we need to display the list for linking.

GPortas: This is actually not a big change. What I mean is that you can move all the composition logic from JsonArrayBuilder into a new method in JsonPrinter, which would then call the method with the minimal boolean parameter. This ensures that all JSON-building logic remains properly encapsulated within JsonPrinter, as intended by design.

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
GPortas: I see. However, I don’t understand the comment `// Assert that the error is returned for having both MetadataBlockNames and inheritMetadataBlocksFromParent` in the context you mentioned. Which error are you asserting now? Should this comment perhaps be updated?

g-saracca:
I'm starting with the JS-Dataverse use case with the GHCR image from this PR, will add any error I encounter here:

- If calling the api without any auth method and I'm receiving below error instead of a Unauthorized related message.
```
{
  "status": "ERROR",
  "message": "class edu.harvard.iq.dataverse.authorization.users.GuestUser cannot be cast to class edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser (edu.harvard.iq.dataverse.authorization.users.GuestUser and edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser are in unnamed module of loader org.glassfish.web.loader.WebappClassLoader @5776dbf3)"
}
```

- The endpoint is including root collection in the response, e.g `/api/v1/dataverses/coll1/dataverse/linkingDataverses`, being `coll1` a sub-collection of the root-collection, should include it?

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
sekmiller: Got it. I think I just copied the comment along with the code I was really after. There isn't an error asserted - I changed the copied code so that there wouldn't be an error - I just wanted the new dataverse to have different metadata blocks than Root. - I updated the comment to reflect that.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
sekmiller: Updated the printer - added the "minimal" logic to the larger Dataverse printer. Re-ran all of DataversesIT (no failures) to hopefully assure us that I didn't break anything.

sekmiller:
> I'm starting with the JS-Dataverse use case with the GHCR image from this PR, will add any error I encounter here:
> 
> * If calling the api without any auth method and I'm receiving below error instead of a Unauthorized related message.
> 
> ```
> {
>   "status": "ERROR",
>   "message": "class edu.harvard.iq.dataverse.authorization.users.GuestUser cannot be cast to class edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser (edu.harvard.iq.dataverse.authorization.users.GuestUser and edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser are in unnamed module of loader org.glassfish.web.loader.WebappClassLoader @5776dbf3)"
> }
> ```
> 
>  * The endpoint is including root collection in the response, e.g `/api/v1/dataverses/coll1/dataverse/linkingDataverses`, being `coll1` a sub-collection of the root-collection, should include it?

I'll look at the cast exception. I don't think root should be included. I'll check the logic there. Thanks

sekmiller:
@g-saracca My latest check in has fixes for Guest User, bad Identifier, and removes parent collections from the linkable list. (I haven't put in an integration test for the parent thing yet, but at least all of the existing tests pass).
Let me know if you see/find anything else.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
sekmiller: In my first pass at this I tried adding the search term to the permittedCollections query, but I got an error saying that I was allowing possibly malicious user input. I can take another crack at it. Not sure about adding the unlinkable Dataverses here - I'll have to think about that.

g-saracca:
Everything working good now 👏🏼

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
sekmiller: I named it that way because originally the method was only for finding Featured Items. In the case of using it for Featured Items there is an additional test for whether the object is released as is required to be featured.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
sekmiller: Added a command

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLinkingDataverseListCommand.java
```diff
@@ -0,0 +1,66 @@
...
+import edu.harvard.iq.dataverse.engine.command.exception.IllegalCommandException;
+import edu.harvard.iq.dataverse.util.BundleUtil;
+import java.util.List;
+
+/**
+ *
+ * @author stephenkraffmiller
+ */
+@RequiredPermissions({})
+public class GetLinkingDataverseListCommand extends AbstractCommand<List<Dataverse>> {
```
GPortas, line 26: Can you add unit tests for this command? As we do for other commands.

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -291,67 +291,67 @@ public static JsonObjectBuilder json(Workflow wf){
...
-        }
-        if (dv.getFilePIDsEnabled() != null) {
-            bld.add("filePIDsEnabled", dv.getFilePIDsEnabled());
-        }
-        bld.add("effectiveRequiresFilesToPublishDataset", dv.getEffectiveRequiresFilesToPublishDataset());
-        bld.add("isReleased", dv.isReleased());
+                .add("name", dv.getName());
+        //minimal refers to only returning the id alias and name for 
+        //use in selecting collections available for linking
+        if (!minimal) {
```
GPortas: You can use a guard clause to handle the minimal case and exit early. Probably more readable.
```
    if (minimal) {
        return bld;
    }

    // The code below adds the not minimal logic and fields...
```

sekmiller: Good idea.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetLinkingDataverseListCommand.java
sekmiller: I can add more tests, but all of the command's functions are tested in DataversesIT

GPortas (approved):
Thanks for addressing the changes. It still requires properly testing the functionality during QA.

g-saracca:
@sekmiller, today after updating the GHCR image on my dockerized environment, I started receiving below error when calling the endpoint:

Command edu.harvard.iq.dataverse.engine.command.impl.GetLinkingDataverseListCommand@4598a330 failed: Exception thrown from bean: jakarta.ejb.EJBTransactionRolledbackException: Exception thrown from bean: java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because "allWithPerms" is null

ofahimIQSS:
[11710server.log](https://github.com/user-attachments/files/22909407/11710server.log)
looks good, merging

Issue #10684, opened 2024-07-14
Feature Request: Enabling configuration of Dataverse using simple file(s)


poikilotherm:
**Overview of the Feature Request**
Let's enable using a [TOML](https://toml.io/en/v1.0.0) file to configure JVM options for starters. Ideally extend to configuring "DB" options and REST API DTO-based options, too, providing a unified approach to how one configures a Dataverse instance. Basically, enable sth like `/etc/dataverse/config.toml`in the spirit of many UNIX/Linux services.

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
Many not-so-experienced Dataverse admins have a hard time setting up their JVM options right.
With our now ever growing list of options (recently the PID providers were added), it's easy to end up with a mess of options.
It's friggin' complicated!

DB options these days cannot be provisioned from the same place as JVM options. Having to very different ways of configuring things as well as adding API endpoints for model based config approaches (auth, licenses, ...) is not making it easier for newbies and non-superhero-admins to follow.

Even though our JVM options start to be scoped and hierarchical, in reality the configuration requires a flat structure as system properties in domain.xml, long env var names, etc. There is one exception: the Dir Config Source allows to create folders with files. But it seems hardly used in classic installations and is clearly geared towards container usage.

Enabling configuration file(s) allows to provision a Dataverse instance from configuration management system in a much easier way. There is loads of tooling around to manage TOML files in idempotent ways, while editing domain.xml is a lot harder. Humans tend to like TOML more than YAML and even more than XML. (Let alone the fact that domain.xml is a VERY large and complex file.) Serving one or more of these files from a K8s ConfigMap, maybe even generated by a K8s Operator is simple and makes people have more "control" over their deployments.

**What existing behavior do you want changed?**
I want to be able to configure JVM options using a TOML file. Here's an example.

Instead of configuring all these options:
```
DATAVERSE_PID_PROVIDERS: "zb-test"
DATAVERSE_PID_DEFAULT_PROVIDER: "zb-test"
DATAVERSE_PID_ZB_TEST_TYPE: "datacite"
DATAVERSE_PID_ZB_TEST_LABEL: "DataCite Test Fabrica"
DATAVERSE_PID_ZB_TEST_AUTHORITY: "10.0346"
DATAVERSE_PID_ZB_TEST_SHOULDER: "JUELICH-DATA-BETA/"
[...]

PR #11745, opened 2025-08-14
11744 CORS: echo request Origin + add Vary

### LINKED ISSUE #11744: CORS: Multi-origin ACAO invalid — echo single Origin and set Vary: Origin
- Set JVM `dataverse.cors.origin` to multiple origins (comma/space separated), e.g. `https://libis.github.io, https://gdcc.github.io`.
- From a browser (or curl with an `Origin` header), call any Dataverse API endpoint (preflight `OPTIONS` and the actual request) with `Origin` set to one of the allowed sites.
- Observe response headers and browser console/network.

- On cross-origin API calls when `dataverse.cors.origin` contains multiple origins.
- Also surfaces when CORS-related config values include surrounding quotes.
- Wildcard `*` combined with credentialed requests (cookies/Authorization) is rejected by browsers.

- Any API endpoint accessed cross-origin, e.g. `/api/files/*`, `/api/access/*`, `/api/externalTools`.

**What happens?**
- `Access-Control-Allow-Origin` is sent as a literal comma-separated list (e.g., `https://a.example, https://b.example`) which browsers reject ("header contains multiple values").
- CORS header values sometimes include quotes (e.g., `"Accept, Content-Type,…"`), causing preflight failures.
- `Vary: Origin` is not set when echoing an origin, allowing caches to serve the wrong ACAO across different origins.

- Any browser-based clients calling the API from allowed third‑party origins (e.g., previewers/marketplace).

- Server echoes the single request `Origin` only when it matches the allowed list.
- `Vary: Origin` is present when echoing a specific origin.
- Configured CSV values are sanitized (no quotes, normalized lists).
- Wildcard `*` is used only when explicitly configured and with awareness that credentialed requests will fail per browser rules.

- 6.7.1 (Payara 6)

- #11365 (Allow `/api/externalTools` to be CORS enabled) — related CORS context but different scope (endpoint access vs. header correctness).

**Screenshots:**
- Browser console typically shows: “has been blocked by CORS policy: The ‘Access-Control-Allow-Origin’ header contains multiple values”.
- Network tab shows ACAO with the comma-separated list or quoted values.

- Yes. Plan:
  - Echo the single request `Origin` when present in `dataverse.cors.origin` (supports multiple origins).
  - Add `Vary: Origin` when echoing.
  - Sanitize methods/allow/expose header lists (strip quotes, normalize).
  - Remove reliance on deprecated `:AllowCors` DB fallback.
  - Add unit tests for wildcard, multi-origin echo, disallowed origin, and sanitization.

ErykKul:

- Fix CORS to correctly handle multiple browser origins without proxy changes.
- Echo the request Origin when it’s in the allowlist and add Vary: Origin so caches behave.
- Sanitize CSV lists for methods/headers to avoid quoted/invalid values breaking preflight.
- Standardize origin parsing to comma-separated; remain whitespace-tolerant.
- Rely on JVM options/MicroProfile only for dataverse.cors.* config (deprecated DB fallback removed).

- Closes #11744

- Wildcard "*" ACAO remains invalid with credentials; docs warn accordingly. Echo logic only applies when a concrete allowlist is configured.
- Config keys used (JVM options/MicroProfile):
  - dataverse.cors.origin (comma-separated list or "*")
  - dataverse.cors.methods
  - dataverse.cors.headers.allow
  - dataverse.cors.headers.expose
- Added Vary: Origin only when echoing a specific Origin. With "*", Vary is not added.
- Unit tests cover wildcard, single/multiple origins, Vary behavior, and header sanitization.
- An unrelated full-suite error (CacheFactoryBeanTest.init, cgroups) was observed locally; targeted CORS tests pass.

1) Configure comma-separated origins via JVM options/MicroProfile, e.g.:
   - dataverse.cors.origin = https://app1.example, https://app2.example
   - dataverse.cors.methods = GET, POST, OPTIONS
   - dataverse.cors.headers.allow = Authorization, Content-Type
   - dataverse.cors.headers.expose = Content-Disposition

2) From a client, send requests with different Origin headers and verify:
   - When Origin matches allowlist: Access-Control-Allow-Origin echoes that origin and response includes Vary: Origin.
   - When Origin doesn’t match: no ACAO header is returned.
   - With wildcard configured (dataverse.cors.origin = *): ACAO is "*"; note that credentialed requests will still fail per browser rules.

3) Preflight check: Send OPTIONS with Origin + Access-Control-Request-Method. Verify Access-Control-Allow-Methods/Headers reflect sanitized config values.

4) Automated tests: run CorsFilterTest (JUnit 5) to validate the above behaviors.

- No UI changes.

- Yes. Included: `doc/release-notes/11744-cors-echo-origin-vary.md`.

**Additional documentation**:

- If helpful, I can add a short section to the installation/admin docs describing CORS configuration via JVM options/MicroProfile and the comma-separated origin convention.

## Verification (Manual Tests)
Replace DV_URL with your base URL and ORIGIN with one you configured.

Preflight (OPTIONS):
```bash
curl -i -X OPTIONS \
  -H "Origin: https://libis.github.io" \
  -H "Access-Control-Request-Method: GET" \
  "${DV_URL}/api/info/version"
```
Expect:
* `HTTP/1.1 200` (or appropriate success)
* `Access-Control-Allow-Origin: https://libis.github.io`
* `Vary: Origin`

Actual request:
```bash
curl -i \
  -H "Origin: https://libis.github.io" \
  "${DV_URL}/api/info/version"
```
Expect same echoed ACAO and `Vary: Origin`.

Multi-origin sanity check (second origin):
```bash
curl -i -H "Origin: https://gdcc.github.io" "${DV_URL}/api/info/version" | grep -i "Access-Control-Allow-Origin"
```
Should echo `https://gdcc.github.io`.

Wildcard mode check (after setting `*`):
```bash
curl -i -H "Origin: https://whatever.example" "${DV_URL}/api/info/version" | grep -i "Access-Control-Allow-Origin"
```
Should show `*` (and no `Vary: Origin`).


src/main/java/edu/harvard/iq/dataverse/filter/CorsFilter.java
```diff
@@ -1,69 +1,130 @@
...
+            }
+
+            response.setHeader("Access-Control-Allow-Methods", methods);
+            response.setHeader("Access-Control-Allow-Headers", allowHeaders);
+            response.setHeader("Access-Control-Expose-Headers", exposeHeaders);
         }
         chain.doFilter(servletRequest, servletResponse);
     }
+
+    /** Remove surrounding quotes and collapse internal quotes. */
```
qqmyers: A minor point, but can these be simpler? I.e. just by splitting on "\\s*,\\s*" or, if we really need to avoid " chars here (and not in other settings?) with "\\s*\\"*,\\"*\\s*"? Same for the next method. (If not, I'm OK with methods but maybe they could go in SettingsWrapper or a utility for re-use?) Also note that #11811 adds a COMMA_BETWEEN_OPTIONAL_WHITE_SPACE constant for splitting in SettingsWrapper.

kmk142789 (approved):

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /** Strip outer quotes, remove remaining quotes, trim. */
+    public static String sanitize(String raw) {
+        if (raw == null)
+            return null;
+        String v = raw.trim();
+        if (v.length() >= 2 && v.startsWith("\"") && v.endsWith("\"")) {
+            v = v.substring(1, v.length() - 1);
+        }
+        return v.replace("\"", "").trim();
```
qqmyers: Any reason to trim or remove surrounding quotes (prior 4 lines) givrn this line? It should do all of it by itself I think.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /** Split into an ordered immutable list of sanitized non-empty tokens. */
+    public static List<String> split(String rawCsv) {
+        String sanitized = sanitize(rawCsv);
+        if (sanitized == null || sanitized.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return Arrays.stream(COMMA_WS.split(sanitized))
+                .map(String::trim)
```
qqmyers: Isn't surrounding whitespace removed by the split?

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+                .filter(s -> !s.isEmpty())
+                .map(CsvUtil::sanitize)
```
qqmyers: Similar - is this needed? sanitize just removes all quotes and trims whitespace, but the COMMA_WS pattern assures there is no preceding/trailing whitespace.

doc/release-notes/11744-cors-echo-origin-vary.md
```diff
@@ -0,0 +1,33 @@
+# 11744: CORS handling improvements
+
+Modernizes CORS so browser integrations (previewers, external tools, JS clients) work correctly with multiple origins and proper caching.
+
+## Highlights
+* Echoes the request origin (`Access-Control-Allow-Origin`) when it matches `dataverse.cors.origin`.
+* Adds `Vary: Origin` for per-origin responses (not for wildcard).
+* Supports comma‑separated origin list; any `*` in the list = wildcard mode.
+* CORS now only enabled when `dataverse.cors.origin` is set (deprecated `:AllowCors` no longer enables it).
+* Sanitizes CORS CSV settings (`dataverse.cors.methods`, `dataverse.cors.headers.allow`, `dataverse.cors.headers.expose`).
```
qqmyers: Sanitizes probably won't make sense to users. Is this "Allows use of spaces and/or double quote characters for readability in all comma-separate-value settings"?

doc/sphinx-guides/source/api/external-tools.rst
```diff
@@ -11,6 +11,9 @@ Introduction
 
 External tools are additional applications the user can access or open from your Dataverse installation to preview, explore, and manipulate data files and datasets. The term "external" is used to indicate that the tool is not part of the main Dataverse Software.
 
+.. note::
+  Browser-based preview or explore tools that make XHR/fetch calls back to the Dataverse API must have CORS explicitly enabled on the Dataverse installation via :ref:`dataverse.cors.origin <dataverse.cors.origin>`. The legacy ``:AllowCors`` database setting is deprecated and no longer enables CORS by itself. Be sure the origins hosting your tool (or ``*`` when appropriate) are included in ``dataverse.cors.origin``; otherwise requests from your tool will be blocked by the browser even if the tool itself loads correctly.
```
qqmyers: simplify: "...that use the Dataverse API..." ? - since you already say browser-based.

Also - the :AllowCors is more than deprecated right? It's been removed? I think in that case it shouldn't be discussed in the Guides any more (it remains in past versions).

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -57,6 +57,15 @@ Allow CORS for S3 Buckets
 **IMPORTANT:** One additional step that is required to enable direct uploads via a Dataverse installation and for direct download to work with previewers and direct upload to work with dvwebloader (:ref:`folder-upload`) is to allow cross site (CORS) requests on your S3 store.
 The example below shows how to enable CORS rules (to support upload and download) on a bucket using the AWS CLI command line tool. Note that you may want to limit the AllowedOrigins and/or AllowedHeaders further.  https://github.com/gdcc/dataverse-previewers/wiki/Using-Previewers-with-download-redirects-from-S3 has some additional information about doing this.
 
+Dataverse itself will only emit the necessary ``Access-Control-*`` headers to browsers when CORS has been explicitly enabled via the JVM/MicroProfile setting :ref:`dataverse.cors.origin <dataverse.cors.origin>`. The legacy database setting ``:AllowCors`` no longer turns CORS on. You must both:
```
qqmyers: same here - drop AllowCors

doc/sphinx-guides/source/developers/big-data-support.rst
```diff
@@ -57,6 +57,15 @@ Allow CORS for S3 Buckets
...
+
+* Configure an appropriate ``dataverse.cors.origin`` value (single origin, comma-separated list, or ``*``) on the Dataverse application server; and
+* Configure a matching/compatible CORS policy on each S3 bucket (and any CDN/proxy in front of it) that will be used for direct upload or for redirect (download-redirect) operations consumed by previewers.
+
+If you specify multiple origins in ``dataverse.cors.origin`` Dataverse will echo back the requesting origin (when it matches) and will include ``Vary: Origin`` so that shared caches do not serve one origin's response to another. If you configure ``*`` Dataverse will respond with ``Access-Control-Allow-Origin: *`` (note that browsers will not allow credentialed requests with a wildcard).
+
+Make sure the bucket CORS configuration ``AllowedOrigins`` is at least as permissive as the origins you configure in ``dataverse.cors.origin``. If the bucket allows ``*`` but the Dataverse application only allows a subset, the browser will still enforce the more restrictive application response.
+
```
qqmyers, line 68: Thanks - lots of good detail.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3678,17 +3678,22 @@ The following settings control Cross-Origin Resource Sharing (CORS) for your Dat
...
 
 ``./asadmin create-jvm-options '-Ddataverse.cors.origin=https://example.com,https://subdomain.example.com'``
 
 Can also be set via any `supported MicroProfile Config API source`_, e.g. the environment variable ``DATAVERSE_CORS_ORIGIN``.
 
+Behavior:
+
+* When a list of origins is configured, Dataverse echoes the single matching request ``Origin`` value in ``Access-Control-Allow-Origin`` and adds ``Vary: Origin`` to support correct proxy/CDN caching.
+* When ``*`` is configured, ``Access-Control-Allow-Origin: *`` is sent and ``Vary`` is not modified.
+* The legacy database setting ``:AllowCors`` is deprecated and no longer enables CORS automatically; you must configure ``dataverse.cors.origin``.
```
qqmyers: Same in this file - just drop everything related to :AllowCors

src/main/java/edu/harvard/iq/dataverse/dataaccess/GlobusOverlayAccessIO.java
```diff
@@ -403,9 +404,11 @@ private static String[] getAllowedEndpoints(String driverId) throws IOException
         } else {
             String rawEndpoints = getConfigParamForDriver(driverId, REFERENCE_ENDPOINTS_WITH_BASEPATHS);
             if (rawEndpoints != null) {
-                allowedEndpoints = getConfigParamForDriver(driverId, REFERENCE_ENDPOINTS_WITH_BASEPATHS).split("\\s*,\\s*");
+                allowedEndpoints = CsvUtil.split(
+                    getConfigParamForDriver(driverId, REFERENCE_ENDPOINTS_WITH_BASEPATHS)
+                ).toArray(new String[0]);
             }
-            if (rawEndpoints == null || allowedEndpoints.length == 0) {
+            if (rawEndpoints == null || allowedEndpoints == null || allowedEndpoints.length == 0) {
```
qqmyers: allowedEndpoints can't be null because CsvUtil.split() can't return that?

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+public final class CsvUtil {
+    /** Shared split regex allowing arbitrary surrounding whitespace. */
+    public static final String COMMA_BETWEEN_OPTIONAL_WHITE_SPACE = "\\s*,\\s*";
+    private static final Pattern COMMA_WS = Pattern.compile(COMMA_BETWEEN_OPTIONAL_WHITE_SPACE);
+
+    private CsvUtil() {
+    }
+
+    /** Strip outer quotes, remove remaining quotes, trim. */
+    public static String sanitize(String raw) {
```
qqmyers: Can this be private or, if it's one line (see comment below), just added to the split method? The two uses I see (in the CorsFilter) look like they can be replaced (via splitToSet?) pretty easily. Also - minor: sanitize is mostly used when some danger is being removed. If we need a separate method, can this be 'standardize' or just 'preprocess'?  (I guess it isn't 'normalize' or 'canonicalize' - that's more what the (unused) normalize method below does.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public final class CsvUtil {
+    /** Shared split regex allowing arbitrary surrounding whitespace. */
+    public static final String COMMA_BETWEEN_OPTIONAL_WHITE_SPACE = "\\s*,\\s*";
+    private static final Pattern COMMA_WS = Pattern.compile(COMMA_BETWEEN_OPTIONAL_WHITE_SPACE);
+
+    private CsvUtil() {
+    }
+
+    /** Strip outer quotes, remove remaining quotes, trim. */
```
pdurbin: @ErykKul I saw your note at https://github.com/IQSS/dataverse/pull/11745#pullrequestreview-3279070455 but I'm still wondering about the "remove remaining quotes" logic. (This is something @qqmyers brought up at standup this morning.) Is it necessary? You want to allow people to put quotes around values? Is there a part of the code that allows this already?

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+        return Arrays.stream(COMMA_WS.split(sanitized))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .map(CsvUtil::sanitize)
+                .collect(Collectors.toUnmodifiableList());
+    }
+
+    /** Split into a de-duplicated (in insertion order) set of sanitized tokens. */
+    public static Set<String> splitToSet(String rawCsv) {
+        List<String> list = split(rawCsv);
```
qqmyers: `return Set.copyOf(split(rawCsv));` ? or `return new HashSet<>(split(rawCsv));` ?

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+        if (list.isEmpty())
+            return Collections.emptySet();
```
qqmyers: If this is kept: I think we request to always use {} with an if, even for one line.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,74 @@
...
+        return list.stream().collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    /** Canonical normalized CSV string (tokens joined by ', '). */
+    public static String normalize(String rawCsv) {
```
qqmyers: Only tested - not used in Dataverse itself? Do we need it?

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: I may have been a bit too forward-thinking. My intention was to anticipate potential future use cases. Since we’ve been moving many settings to MicroProfile, I aimed for better compliance with RFC 4180. This would allow features like supporting line breaks and commas when using quotes.
That said, I completely agree with your point — we can always add this later if the need arises. For now, I’ll remove the commas and drop the RFC references.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: Simplified the code and quotes are not allowed anymore.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: No, but I simplified the code, as this was over engineered. The first trim removes the surrounding white spaces, the split removes the spaces adjacent to the comma.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: I have removed the method sanitize method since it is only trim() now. This method looks like this now:
```java
    /** Split list of trimmed tokens. */
    public static List<String> split(String rawCsv) {
        if (rawCsv == null) {
            return Collections.emptyList();
        }
        final String trimmedCsv = rawCsv.trim();
        if (trimmedCsv.isEmpty()) {
            return Collections.emptyList();
        }
        return Arrays.asList(SPLIT.split(trimmedCsv));
    }
```

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: Since it was only trim() now, the method is removed.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: Method removed. And yes, using {} is a good style I will pay attention to.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: Method removed.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: I kept only the splitToLowerCaseSet, since it is used, and simplified it:
```java
    /** Convenience: split into a lowercase set. */
    public static Set<String> splitToLowerCaseSet(String rawCsv) {
        if (rawCsv == null || rawCsv.trim().isEmpty()) {
            return Collections.emptySet();
        }
        return Set.copyOf(split(rawCsv.toLowerCase()));
    }
```

src/main/java/edu/harvard/iq/dataverse/dataaccess/GlobusOverlayAccessIO.java
ErykKul: I have refactored this method for better readability and removed the unnecessary null check:
```java
    private static String[] getAllowedEndpoints(String driverId) throws IOException {
        if (GlobusAccessibleStore.isDataverseManaged(driverId)) {
            final String value = getConfigParamForDriver(driverId, TRANSFER_ENDPOINT_WITH_BASEPATH);
            if (value == null) {
                throw new IOException("dataverse.files." + driverId + "." + TRANSFER_ENDPOINT_WITH_BASEPATH + " is required");
            }
            return new String[] { value };
        } else {
            final String raw = getConfigParamForDriver(driverId, REFERENCE_ENDPOINTS_WITH_BASEPATHS);
            if (raw == null) {
                throw new IOException("dataverse.files." + driverId + ".base-url is required");
            }
            // CsvUtil.split never returns null; may return empty list if raw is blank.
            final List<String> list = CsvUtil.split(raw);
            if (list.isEmpty()) {
                throw new IOException("dataverse.files." + driverId + ".base-url is required");
            }
            return list.toArray(new String[0]);
        }
    }
```

doc/release-notes/11744-cors-echo-origin-vary.md
ErykKul: I have updated the release note:  Allows readable spacing in CORS list settings (`dataverse.cors.methods`, `dataverse.cors.headers.allow`, `dataverse.cors.headers.expose`): spaces around commas are ignored; tokens are otherwise unchanged (no quote parsing).

doc/sphinx-guides/source/api/external-tools.rst
ErykKul: I have changed it to:
```md
.. note::
  Browser-based tools must have CORS explicitly enabled via :ref:`dataverse.cors.origin <dataverse.cors.origin>`. List every origin that will host your tool (or use ``*`` when a wildcard is acceptable). If an origin is not listed, the browser will block that tool's API requests even if the tool page itself loads.

```

ErykKul:
@qqmyers, I have processed your comments, I think it is better now.
@poikilotherm, Can you take a look at this PR? I think the most relevant part for you is the JvmSettings.java addition:
```java
    /** Lookup optional CSV value and return a List of tokens. */
    public java.util.Optional<java.util.List<String>> lookupCsvListOptional() {
        return lookupOptional().map(CsvUtil::split);
    }

    /** Lookup required CSV value and return a List of tokens (throws if missing). */
    public java.util.List<String> lookupCsvList() {
        return CsvUtil.split(lookup());
    }
```

doc/release-notes/11744-cors-echo-origin-vary.md
```diff
@@ -0,0 +1,33 @@
+# 11744: CORS handling improvements
+
+Modernizes CORS so browser integrations (previewers, external tools, JS clients) work correctly with multiple origins and proper caching.
+
+## Highlights
+- Echoes the request origin (`Access-Control-Allow-Origin`) when it matches `dataverse.cors.origin`.
+- Adds `Vary: Origin` for per-origin responses (not for wildcard).
+- Supports comma‑separated origin list; any `*` in the list = wildcard mode.
+- CORS now only enabled when `dataverse.cors.origin` is set (deprecated `:AllowCors` no longer enables it).
+- Allows readable spacing in CORS list settings (`dataverse.cors.methods`, `dataverse.cors.headers.allow`, `dataverse.cors.headers.expose`): spaces around commas are ignored; tokens are otherwise unchanged (no quote parsing).
```
qqmyers: This is true for all database and microprofile settings tat accept comma-separated values - probably should be it's own top-level entry rather than just part of the CORS entry.

doc/release-notes/11744-cors-echo-origin-vary.md
```diff
@@ -0,0 +1,33 @@
...
+dataverse.cors.origin=https://libis.github.io,https://gdcc.github.io
+dataverse.cors.origin=*
+```
+Optional (unquoted):
+```
+dataverse.cors.methods=GET, POST, OPTIONS, PUT, DELETE
+```
+
+## Compatibility
+- Must configure `dataverse.cors.origin`; `:AllowCors` no longer sufficient.
```
qqmyers: :AllowCors was deprecated and has now been removed.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -4943,19 +4947,17 @@ This can be helpful in situations where multiple organizations are sharing one D
 or
 ``curl -X PUT -d '*' http://localhost:8080/api/admin/settings/:InheritParentRoleAssignments``
 
-:AllowCors (Deprecated)
-+++++++++++++++++++++++
+:AllowCors (Deprecated – no longer used once dataverse.cors.* settings exist)
```
qqmyers: Can you just remove AllowCors from the SettingsServiceBean and drop this section? The only place I'd mention that it existed is in the release notes backward compatibility section.

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -175,6 +175,9 @@ File Previews
...
 
+.. note::
+   Some previewers run purely in the browser and make direct (JavaScript) requests back to the Dataverse API endpoints to retrieve file contents, metadata, or signed URLs. For these previewers to function when hosted on a different origin (e.g., a CDN or a separate previewer service), the Dataverse installation must have CORS enabled via :ref:`dataverse.cors.origin <dataverse.cors.origin>`. Administrators should configure the list of allowed origins to include the host serving the previewers. The deprecated ``:AllowCors`` database setting no longer enables CORS.
```
qqmyers: drop AllowCors here as well

qqmyers:
FWIW: My only change requests at this point are some doc changes and just removing the :AllowCors setting since it no longer does anything. Otherwise I think it's ready for QA.
@poikilotherm - if you want to review this I can leave it "In Review", otherwise I'll approve it after the next set of minor fixes and let it go to "Ready for QA" - let me know.

src/main/java/edu/harvard/iq/dataverse/util/ListSplitUtil.java
```diff
@@ -0,0 +1,45 @@
...
+ * <p>
+ * Behavior:
+ * - Leading/trailing whitespace of the whole input is ignored.
+ * - Whitespace immediately around commas is ignored ("GET, POST" == "GET,POST").
+ * - Tokens are otherwise preserved exactly as typed (no quote stripping, no escape processing).
+ * Not a full CSV parser: embedded commas, quoted fields with separators, and newlines inside tokens are NOT supported.
+ */
+public final class CsvUtil {
+    /** Split on commas, trimming any adjacent to comma whitespace. */
+    private static final Pattern SPLIT = Pattern.compile("\\s*,\\s*");
```
jo-pol, line 20: A less DIY approach

    import org.apache.commons.csv.CSVFormat;
    import org.apache.commons.csv.CSVParser;
    
    import java.io.IOException;
    import java.util.Arrays;
    import java.util.List;

    public static List<String> split(String rawCsv) {
        if( rawCsv == null ) {
            return List.of();
        }
        CSVFormat format = CSVFormat.RFC4180.builder().setTrim(true).get();
        try (CSVParser parser = CSVParser.parse(rawCsv, format)) {
            return Arrays.asList(parser.getRecords().get(0).values());
        } catch (IOException e) {
            // should not happen
            return Arrays.asList(rawCsv.split("\\s*,\\s*"));
        }
    }

Anything beyond newline would be ignored but quoted fields with comma's would be properly parsed and quotes and white space trimmed. Note that quoted values should not have surrounding white space

jo-pol: Correction: Anything beyond a newline _outside a quoted value_ ....

qqmyers: Thanks @jo-pol - this seems like a better way to get full CSV parsing. That said, we had a core team discussion yesterday where we agreed we didn't want to support double quotes/full RFC4180 in the settings, at least at this point, and @ErykKul took out the logic beyond stripping extra whitespace. (I think the main reason was to avoid having flexibility we don't need/use - the settings are usually simple words and all we've seen in practice  (including DANS issues) is problems where people add spaces for readability or possible cut/paste errors (spaces before commas or at the beginning/end). If we get more complex settings or start seeing double quotes, etc. we could revisit this.)

doc/sphinx-guides/source/installation/config.rst
poikilotherm: We have kept the old setting docs around for others before as well. See https://guides.dataverse.org/en/latest/installation/config.html#legacysystememail for an example.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Helpers for simple admin settings that accept comma-separated lists (origins, methods, headers, etc.).
+ * <p>
+ * Behavior:
+ * - Leading/trailing whitespace of the whole input is ignored.
+ * - Whitespace immediately around commas is ignored ("GET, POST" == "GET,POST").
+ * - Tokens are otherwise preserved exactly as typed (no quote stripping, no escape processing).
+ * Not a full CSV parser: embedded commas, quoted fields with separators, and newlines inside tokens are NOT supported.
+ */
+public final class CsvUtil {
```
poikilotherm: I very much agree that we should have such a class including tests etc instead of having individual split configurations in different places. Pro consistency!

What I don't like about this: this is not really an implementation of CSV. So why call it that, setting expectations we don't meet with it?

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -604,4 +606,13 @@ public String insert(String... arguments) {
         return String.format(this.getScopedKey(), (Object[]) arguments);
     }
     
+    /** Lookup optional CSV value and return a List of tokens. */
```
poikilotherm: Eryk I'm sorry but I really really really don't like this at all.

1. I don't understand why we would need this? 
    - It seems strange that any kind of CORS related setting would carry a "," as part of a value (so not just a separator).
    - Even so, why not just make people escape the value as "\," then? We are pointing people to using MicroProfile Config but at the same time try to extend to non-standard behavior. (See also [the spec](https://download.eclipse.org/microprofile/microprofile-config-3.0/microprofile-config-spec-3.0.html#_array_converters))
    - And why would we allow people to be able to give a list of something separated by comma plus space? What's wrong about just using a comma? Again, this is deliberately trying to extend what's in the spec of MPCONFIG...?
2. This is an extension of the interface which requires devs to actively use this. Thus, it will create inconsistencies, as we all know how things went creative with DB settings. Some options may allow "CSV" style, others may not. How do we communicate this properly? It's going to drive admins nuts -- let us not repeat the errors of early DB settings design. If we really think this is an absolute must and admins cannot escape a comma, let's at least make this a default converter for `String[]`. (This will however not help in cases were we want auto-splitting and conversion too, for example `Integer[]`.)
3. The "CSV" bit is so far only about allowing to have spaces trimmed from the value. So this is not even really looking into avoiding the split if the comma is part of the list element's value. Thus, I would not at all call this "CSV", because it is not going the whole way. Which is fine, but then don't call it that -- manage expectations, wording matters.
4. The behavior of lookupCsvList() is inconsistent. If it's not present, etc, we return an empty list instead of throwing a `NoSuchElementException`.
5. I do get the intention of enabling programmatic lookups for some `java.util.Collection<T>` return type (mind the generic) and reducing the boilerplate necessary to get there from a mere `T[]`. How about we change that instead?
    - This would actually be very handy in many situations and you could even use the automatic type conversion of MPCONFIG with it! (For example use an enum for CORS method names, enabling validation as a side effect.)
    - I suggest we provide sth. along the lines of `public <T extends Collection<E>, E> T lookup(Class<T> collectionType, Class<E> valueType)`. It should obviously support varargs as well. It should be enough to support `List<T>` and `Set<T>`. I am happy to give an implementation a whirl.
    - Once we have a simple way to retrieve a setting as a collection with converted types, it's fairly trivial to do the trimming in case some code wants that (see above how I feel about that) using stream processing.
    - If we want to generally allow spaced separation of list elements, it would be fairly simple to add this behavior of trimming spaces before converting as the consistent default one.

doc/sphinx-guides/source/installation/config.rst
qqmyers: Yes, but that is merely deprecated - it still works. :AllowCors no longer does anything.

poikilotherm: Is that what we want with backward compatibility? I could have dropped SystemEmail back in the day, but we didn't because we wanted to not break existing configurations. I'm puzzled - when did we move on from that established convention?

qqmyers:
> On a related note: In the docs, for some reason @qqmyers decided back in the day to branch of the list of CORS settings into their own sub-section. This breaks our usual pattern of having JVM options, Payara specials, Feature Flags and DB options. @pdurbin WDYT should we use this PR to make the pattern realign? (I completely agree with Jim that the whole config guide is nuts these days because of it's size, but IMHO we still should maintain consistency.)

FWIW: I didn't intend this as a separate section - all the settings are still at the end of the JVM settings list. I added the 'CORS Settings' header primarily because I added the .. _dataverse.cors: which I thought we'd want to link from other sections (and external material in the previewers repo). That said, it doesn't look like we've yet used that anchor, so if you want the header and/or anchor removed, I'm fine with it.

doc/sphinx-guides/source/installation/config.rst
qqmyers: I hope we can remove SystemEmail at some point and don't have a convention that deprecated things have to be supported forever. (Perhaps we need a policy that deprecated things can be removed without concern after n releases?) 

To split the issue: if the setting is really removed/does nothing, I think it should not be in the documentation, whereas deprecated but working things should remain. I don't think that's controversial?

As to whether :AllowCors should stay: it was already deprecated, although only in 6.7, so it would be relatively quick to remove it in 6.9 ~ six months later. That said, for performance reasons, it is only read at startup and so doesn't work like the other db settings which make live changes. (That's documented in the guide.) The change required to keep CORS working is just to add `dataverse.cors.origin=*` . So - my thought was, given that the change should be easy for users, and that @ErykKul had already removed it, that it made sense to go ahead and remove it now versus waiting a few more releases (and then forgetting it). That said, unless I'm missing something, it isn't technically hard to keep it, so if there is consensus to keep this/all deprecated things around longer than 2 releases, I don't think it's a big change to the PR (beyond documentation).

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
qqmyers: Do you have a suggestion for what to call things that are comma separated values? (I agree that CSV evokes the file standard, but I don't have any great ideas for an alternative. Something like SettingsUtil to get away from the specifics of CSV?)

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
qqmyers: I'm not sure I understand the comment about values with internal commas - I don't think anyone expects such values and I don't think the code tries to preserve them(?). 

FWIW: I think the intended change is just to standardize how settings work by allowing optional whitespace around commas. That's for two reasons: 1) we have admins adding settings with whitespace, especially after a comma that was breaking parsing/causing things to not work, and 2) we organically ended up with different settings having different rules as to whether spaces were allowed - after commas and sometimes before. (There are other issues/PRs in progress related to changing individual settings or partially standardizing (most/all going towards allowing whitespace I think)). 

My sense is that any solution (to standardizing how comma separated settings work) should avoid silent failures if people add whitespace (by allowing them or flagging the error) - that's the key issue. The most backward compatible option would be to allow whitespace around commas since that's already the case for some settings, but validating for no space around commas could be a solution if the break with backward compatibility is worth it. If how to proceed needs further discussion (? does it) or additional work is needed towards a consensus solution, I guess the two options are to hold this PR for changes or have @ErykKul drop this PR back to just CORS changes and make standardization a separate issue/PR.  The latter probably makes more sense if things will take long enough to put getting the CORS changes into 6.9 at risk, but I guess it's ultimately up to @ErykKul.

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
poikilotherm: Not sure. Maybe something very unfancy like "ListSplitUtil" or "ListSplitter"?

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
poikilotherm: I do understand the urge to use whitespace in options to make things more readable. It makes the life of an admin easier, so I'm all for it. This is exactly one of the reasons I wanted Payara so badly to add support for config file formats like JSON, YAML, or TOML. (And they added TOML in 6.2024.12 :pray: see also #10684) That way, you can express lists in so many different ways and with all the whitespace you want, verbose or inline style etc. Manually editing domain.xml should be retired for good.

So why not just stick with the MicroProfile Config Standard, save ourselves from maintaining a workaround to enable whitespace in lists and use a proper configuration format instead? Anybody who wants to stick to or mix in ways like environment variables, system properties and editing domain.xml can still do that. As we had people upgrade to a Payara 6.2025.2+ since v6.6, most installations should be able to use the TOML config file. (Obviously, any properties not yet migrated to MPCONFIG are not usable that way. But we could make the effort to change that for 6.9.) I'd be happy to provide a PR with docs, updated install scripts and base image to make for a smooth transition.

I'm also happy to prepare a small tech demo, but this might need to wait a bit until end of October.

It might still be worth creating a way to programmatically lookup a `Collection`type as suggested above (MPCONFIG has support for this via injection, but not for its programmatic API). I'm also happy to include that into a PR, either as part of the TOML effort or separatly.

By the way: using `\s*,\s*` is not just matching "` , `", but also allowing any amount of `[\t\n\r]` etc. Was that intended as a side effect?

doc/release-notes/11744-cors-echo-origin-vary.md
ErykKul: I made this commit: https://github.com/IQSS/dataverse/pull/11745/commits/8928d45e519ff9940603ab824cc85125eef83496

src/main/java/edu/harvard/iq/dataverse/util/CsvUtil.java
ErykKul: Renamed to ListSplitUtil.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
ErykKul: Proposed approach (open to alternatives): we only need a List<String> for now. Our MicroProfile settings are passed as JVM options and we intentionally allow spaces in values; several reworked settings already rely on this. Forbidding spaces now would create another breaking change. My proposal is to let MicroProfile return String[] per spec (keeping TOML and other config sources compatible) and then normalize tokens locally by trimming whitespace when building the list. This preserves current behavior while avoiding another settings migration.

```java
    /** Lookup optional CSV value and return a List of tokens. */
    public java.util.Optional<java.util.List<String>> lookupSplittedListOptional() {
        return lookupOptional(String[].class).map(splitted -> ListSplitUtil.split(String.join(",", splitted)));
    }

    /** Lookup required CSV value and return a List of tokens (throws if missing). */
    public java.util.List<String> lookupSplittedList() {
        return ListSplitUtil.split(String.join(",", lookup(String[].class)));
    }
```

ErykKul: After Copilot review I have this:
```java
    /** Lookup optional comma-separated value and return tokens as a List. */
    public Optional<List<String>> lookupSplittedListOptional() {
    return lookupOptional(String[].class)
        .map(values -> Arrays.stream(values).map(String::trim).toList());
    }

    /** Lookup required comma-separated value and return tokens as a List. */
    public List<String> lookupSplittedList() {
        return Arrays.stream(lookup(String[].class))
                .map(String::trim)
                .toList();
    }
```

qqmyers (approved):
The updates look good to me. As I'm not sure if @poikilotherm 's concerns have been addressed, I'm going to remove my assignment but leave this in In Review for him to request further changes or approve/move to Ready for QA.

src/main/java/edu/harvard/iq/dataverse/settings/JvmSettings.java
```diff
@@ -604,4 +605,17 @@ public String insert(String... arguments) {
         return String.format(this.getScopedKey(), (Object[]) arguments);
     }
     
+    /** Lookup optional comma-separated value and return tokens as a List. */
+    public Optional<List<String>> lookupSplittedListOptional() {
```
poikilotherm, line 624: I think this new approach makes much more sense now. It's a better balance between the different systems. We now rely on MPCONFIG to give us the data and apply some internal logic -- seems fine to me.


Some small suggestions:
1. There may be situations when you don't want the trimming to happen and preserve whitespaces. Should we add a boolean flag and make the method without it default to trimming? (Needs to be documented in the Javadocs, too)
2. We don't need to care about null values here (thx MPCONFIG), but what about empty elements after trimming? Do we want them? Should we filter for them? (`.filter(s -> !s.isEmpty())` after trimming) Reading the MPCONFIG spec, usually empty elements are not allowed and removed. (See the [",bar" examples](https://download.eclipse.org/microprofile/microprofile-config-3.1/microprofile-config-spec-3.1.html#_config_value_conversion_rules))
3. It would be great to have more extensive JavaDocs that also explain some of the background and limitations. The return values and potential throws should be explained.

pdurbin: @poikilotherm since you suggested some changes I assigned @ErykKul to this PR.

poikilotherm: @pdurbin @qqmyers I'd apprecciate input on the points above :wink:

pdurbin: I think they're good points but I'm ok with proceeding with QA.

@poikilotherm since @qqmyers has removed his assignment I think it's up to us to either move this PR to QA (by approving it) or asking @ErykKul to do more. What do you prefer?

ErykKul: I will update this pr, the remaining requested changes look small to me and I agree with them.

ErykKul: I have now this:
```java

    /**
     * Lookup an optional comma-separated value and return the tokens as an immutable list.
     * MicroProfile Config removes zero-length segments when it converts to {@code String[]}, but
     * it leaves any leading or trailing whitespace on the surviving tokens (including tokens that
     * contain only spaces). This convenience overload trims each token; after trimming, any token
     * that becomes empty (because it consisted solely of whitespace) is discarded so callers still
     * receive a list that is free of empty strings. Use the boolean overload with {@code false} if
     * you need the exact whitespace that MicroProfile provided.
     *
     * @return an {@link Optional} containing the list of tokens when the setting is present;
     *         an empty {@link Optional} if the setting is not configured
     */
    public Optional<List<String>> lookupSplittedListOptional() {
        return lookupSplittedListOptional(true);
    }

    /**
    * Lookup an optional comma-separated value and return the tokens as an immutable list.
    *
    * @param trimSpaces when {@code true}, individual elements are trimmed; tokens that become empty after
    *                   trimming (because they were all whitespace) are removed to preserve MicroProfile's
    *                   "no empty entries" guarantee; when {@code false}, the tokens are returned exactly as
    *                   produced by MicroProfile Config
     * @return an {@link Optional} containing the list of tokens when the setting is present;
     *         an empty {@link Optional} if the setting is not configured
     */
    public Optional<List<String>> lookupSplittedListOptional(boolean trimSpaces) {
        return lookupOptional(String[].class)
            .map(values -> Arrays.stream(values)
                .map(s -> trimSpaces ? s.trim() : s)
                .filter(s -> trimSpaces ? !s.isEmpty() : true)
                .toList());
    }

    /**
     * Lookup a required comma-separated value and return the tokens as an immutable list.
     * MicroProfile Config removes zero-length segments when it converts to {@code String[]}, but it
     * leaves any leading or trailing whitespace on the surviving tokens (including tokens that contain
     * only spaces). This convenience overload trims each token; after trimming, any token that becomes
     * empty (because it consisted solely of whitespace) is discarded so callers still receive a list that
     * is free of empty strings. Use the boolean overload with {@code false} if you need the exact whitespace
     * that MicroProfile provided.
     *
     * @return the list of tokens for the configured setting
     * @throws java.util.NoSuchElementException if the setting is missing or blank
     * @throws IllegalArgumentException if conversion to {@code String[]} fails
     */
    public List<String> lookupSplittedList() {
        return lookupSplittedList(true);
    }

    /**
    * Lookup a required comma-separated value and return the tokens as an immutable list.
    *
    * @param trimSpaces when {@code true}, individual elements are trimmed; tokens that become empty after
    *                   trimming (because they were all whitespace) are removed to preserve MicroProfile's
    *                   "no empty entries" guarantee; when {@code false}, the tokens are returned exactly as
    *                   produced by MicroProfile Config
     * @return the list of tokens for the configured setting
     * @throws java.util.NoSuchElementException if the setting is missing or blank
     * @throws IllegalArgumentException if conversion to {@code String[]} fails
     */
    public List<String> lookupSplittedList(boolean trimSpaces) {
        return Arrays.stream(lookup(String[].class))
            .map(s -> trimSpaces ? s.trim() : s)
            .filter(s -> trimSpaces ? !s.isEmpty() : true)
            .toList();
    }

```

pdurbin: Thanks, @ErykKul!

Issue #4284, opened 2017-11-14
Logging guidelines in the coding style document should be updated


michbarsinai:
Current type of logging is wasteful: the JVM concatenates the string, passes it to the logger, only to have the logger ignore it. Passing the message format and the argument array is much more efficient (It's an auto-fix in NetBeans, BTW).

current: `logger.fine("object " + o + " does not match " + m + ".");`
efficient: `logger.log(Level.FINE, "object  {0} does not match {m}", new Object[]{o,m});`

pdurbin:
Thanks for opening this issue, @michbarsinai

Isn't readability worth something? My eyes are bleeding a bit over `new Object[]{o,m})` and without tooling, would we even go on a quest to make these uglifications?

Has anyone measured performance to see if this will make much of a difference? This seems like classic premature optimization.

Dataverse is a slow app. I really doubt that changing the thousands of cases of these lines will make any difference. If we care about performance, let's measure and have some tooling tell us where Dataverse is slow. (Usually, it's database queries.) I'm fine with what's written at http://guides.dataverse.org/en/4.8.2/developers/coding-style.html#logging about using `logger.fine`. I probably wrote it. 😄 It looks nice and I haven't seen any data to indicate that it's especially slow.

What do others think? Do you like the ugly? Is it worth it? For more context, please see https://github.com/IQSS/dataverse/pull/4280#discussion_r150973811

michbarsinai:
Some points:
1. I disagree that's this is a premature optimization: we're not making some wrong assumptions that allow us to overly simplify an algorithm. There are no consequences except readability,  performance, and memory consumption. We construct strings at runtime, mostly for developer readability. They are hardly ever used in production.
2. Constructing a string is not cheap. It makes the JVM allocate a contiguous memory block the size of the string. So it's not just the construction (O(n)), it's the GC overhead (+fragmentations etc).
3. The constructed strings are arbitrarily long here (e.g. for objects that have large toString results, such as sets), this can, at times, become a real memory issue. proliferation of such logging messages, in conjunction with other memory-intensive actions, may cause sporadic out-of-memory errors, which will be hard to replicate.

So the suggested change is less readable, true, but is safer and more performant. The current logging style is problematic enough to warrant its own NetBeans validation and auto-fix, so clearly I'm not the only one thinking we should change the way we log.
[...]

PR #11753, opened 2025-08-19
add "review" dataset type

### LINKED ISSUE #11747: Add "review" scholarly research object
As part of the Trusted Data project...

- https://github.com/IQSS/dataverse-pm/issues/425

... we'd like to explore using dataset types to create a new type called "review".

pdurbin:

Adds "review" as dataset type and publishes `resourceTypeGeneral=Other` to DataCite for it.

TODO:

- [ ] add custom metadata blocks related to reviews

- Closes #11747

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DataCitation.java
```diff
@@ -740,8 +740,11 @@ public Map<String, String> getDataCiteMetadata() {
 
     public JsonObject getCSLJsonFormat() {
         CSLItemDataBuilder itemBuilder = new CSLItemDataBuilder();
+        // TODO consider making this a switch
         if (type.equals(DatasetType.DATASET_TYPE_SOFTWARE)) {
             itemBuilder.type(CSLType.SOFTWARE);
+        } else if (type.equals(DatasetType.DATASET_TYPE_REVIEW)) {
+            itemBuilder.type(CSLType.REVIEW);
```
pdurbin: @qqmyers I made this change without testing anything because I wasn't sure how to. Can you please advise? I'd like to add something under "how to test" about it.

qqmyers: I guess you can unit test w.r.t. making sure the CSLJson output has the review type if your datasettype is review. Beyond that, the CSL Json is used in front side JavaScript to generate any of thousands of citation formats, but any given format may or may not use CSLType in generating its output. I can't think of any easy way to test that (but that would really be testing the CSL Java and JavaScript libraries - if our code gets the CSLType into the CSL Json output here, things should work).

pdurbin: Ok, thanks. Do you know of any format that uses CSLType? Maybe we can test that type by manually inspecting its output?

qqmyers: I don't know. AI says APA for the social sciences and Chicago for history and arts, and MLA for humanities are popular for reviews - hopefully that means they check the type.

pdurbin: Ok, I'm still confused. Still not sure how to test this from JSF. I did look at APA and it says `[dataset]` (not `[review]` ) like this:

Screenshot 2025-08-21 at 11 28 18 AM

But I'm not sure if I'm looking in the right place.

If you think I should simply back out of this change, I'm ok with that. Less to mention in the release note. 😄

qqmyers: You can get the CSL format itself via api, e.g.  curl `https://demo.dataverse.org/api/datasets/:persistentId/versions/1.0/citation/CSL?persistentId=doi:10.70122/FK2/GJF7SF` - note the [native api guide](https://guides.dataverse.org/en/latest/api/native-api.html#get-citation-in-other-formats) appears to have an error and lists the format as "CSLJson" rather than "CSL".

pdurbin: Thanks, at https://dev1.dataverse.org/api/datasets/:persistentId/versions/1.0/citation/CSL?persistentId=doi%3A10.5072/FK2/RMEZNL I'm getting this:

```
{
    "id": "-GEN-2mqfdnmq0r",
    "type": "dataset",
    "categories": [
    ],
    "author": [
        {
            "family": "Simpson",
            "given": "Homer",
            "isInstitution": false
        }
    ],
    "issued": {
        "date-parts": [
            [
                2025
            ]
        ]
    },
    "DOI": "10.5072/FK2/RMEZNL",
    "publisher": "Root",
    "title": "Review of Darwin's Finches",
    "URL": "http://ec2-44-214-43-137.compute-1.amazonaws.com/citation?persistentId=doi:10.5072/FK2/RMEZNL",
    "version": "V1"
}
```

(I need to fix my siteUrl, obviously.) 😅 

I guess you're saying ideally "type" would be "review" or at least not "dataset" in that output. It makes me wonder what my change did, if anything. 🤔 

Also, thanks for the heads up about the typo in the guides.

pdurbin: I meant to say you've given me an API endpoint to dig into. Thanks!

jggautier:
Next Tuesday I'll add to this PR . properties files for the two metadata blocks

jggautier:
I put the two properties files for the metadata blocks into this branch's src/main/java/propertyFiles folder

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -824,11 +824,11 @@ Dataset Types
 
 .. note:: Development of the dataset types feature is ongoing. Please see https://github.com/IQSS/dataverse-pm/issues/307 for details.
 
-Out of the box, all datasets have a dataset type of "dataset". Superusers can add additional types such as "software" or "workflow" using the :ref:`api-add-dataset-type` API endpoint.
+Out of the box, all datasets have a dataset type of "dataset". Superusers can add additional types such as "software", "workflow", or "review" using the :ref:`api-add-dataset-type` API endpoint.
 
 Once more than one type appears in search results, a facet called "Dataset Type" will appear allowing you to filter down to a certain type.
 
-If your installation is configured to use DataCite as a persistent ID (PID) provider, the appropriate type ("Dataset", "Software", "Workflow") will be sent to DataCite when the dataset is published for those three types.
+If your installation is configured to use DataCite as a persistent ID (PID) provider, the appropriate type ("Dataset", "Software", "Workflow", "Review") will be sent to DataCite when the dataset is published for those types.
```
stevenwinship: Shouldn't this say ("Dataset", "Software", "Workflow", "Other")? Review is sent to Datacite as "Other"

pdurbin: Hmm, good point. Maybe I'll change this to "certain types" and link to the table I made elsewhere, or something. Thanks! 😅

pdurbin: Fixed in 90346e41a4.

src/main/java/edu/harvard/iq/dataverse/DataCitation.java
pdurbin: Fixed in 6994e22cfa. Thank again for the pointers!

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -4741,14 +4751,28 @@ static Response listDataverseInputLevels(String dataverseAlias, String apiToken)
...
-        Response response = given()
-                .get("/api/datasets/datasetTypes");
-        return response;
+        return getDatasetTypes(null);
+    }
+
+    public static Response getDatasetTypes(String acceptLanguage) {
+        RequestSpecification requestSpecification = given();
+        if (acceptLanguage != null) {
+            https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language
```
poikilotherm: This should be a comment, right?

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -5719,17 +5719,19 @@ public Response resetPidGenerator(@Context ContainerRequestContext crc, @PathPar
 
     @GET
     @Path("datasetTypes")
-    public Response getDatasetTypes() {
+    public Response getDatasetTypes(@HeaderParam(ACCEPT_LANGUAGE) String acceptLanguage) {
+        Locale locale = i18nUtil.parseAcceptLanguageHeader(acceptLanguage);
```
poikilotherm: IMHO any parsing of the locale should be done using JAX-RS mechanisms to follow DRY principle.

- Solution A) Keep `@HeaderParam`, but make it a `Locale`, moving the parsing to a `ParamConverter`. This is still a lot of repeated boilerplate code.

- Solution B) Have a `@Context HttpHeaders` parameter give you access via `headers.getAcceptableLanguages()` or a `@Context Request` give  you access via `request.getLanguage()` to the `Locale` without manual parsing. Still some boilerplate per method

- Solution C) Create a CDI `@Producer` method that is `@RequestScoped`, receiving the `Locale` as a class field `@Inject Locale`. This would be greatly enhanced by adding an annotation like `@ClientLocale` to be used as qualifier for both field and producer method. This is the least boilerplate code.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
+        return NullSafeJsonBuilder.jsonObjectBuilder()
                 .add("id", getId())
                 .add("name", getName())
+                .add("displayName", getDisplayName(locale))
+                .add("description", getDescription(locale))
                 .add("linkedMetadataBlocks", linkedMetadataBlocks)
                 .add("availableLicenses", availableLicenses);
     }
 
+    public String getDisplayName(Locale locale) {
```
poikilotherm, line 164: Is it really necessary to duplicate the code for displayName and description?

poikilotherm, line 164: All the log messages should use lambas to only execute the string concatenations when actually being logged.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
+            logger.fine("Locale couldn't be parsed, returning default display name: " + displayName);
+            return displayName;
+        }
+        if (locale.getLanguage().equals(Locale.ENGLISH.getLanguage())) {
+            // This is here to prevent looking up datasetTypes_en.properties, which doesn't exist.
+            // The English strings are in datasetTypes.properties (no _en).
+            logger.fine("Locale is English, returning default display name: " + displayName);
+            return displayName;
+        }
+        String propertiesFile = "datasetTypes_" + locale.toLanguageTag() + ".properties";
```
poikilotherm, line 180: This burns unnecessary CPU cycles every time it's looked up but is only used for a debugging log message.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
+        logger.fine("Getting display name for dataset type " + name + " and locale " + locale);
+        if (locale == null) {
+            logger.fine("Locale is null, returning default display name: " + displayName);
+            return displayName;
+        }
+        if (locale.getLanguage().isBlank()) {
+            logger.fine("Locale couldn't be parsed, returning default display name: " + displayName);
+            return displayName;
+        }
+        if (locale.getLanguage().equals(Locale.ENGLISH.getLanguage())) {
```
poikilotherm, line 174: IMHO this is the wrong place to do this. Shouldn't this be handled by the BundleUtil, being properly compatible with any other lookups that may need this? At least it is an implementation detail of BundleUtil, so doing this here is violating SOLID.

Also, the BundleUtil implementation will return null if the bundle cannot be found, thus this would be the same as putting a null locale. Why bother with these implementation details here?

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
                 .add("name", getName())
+                .add("displayName", getDisplayName(locale))
+                .add("description", getDescription(locale))
                 .add("linkedMetadataBlocks", linkedMetadataBlocks)
                 .add("availableLicenses", availableLicenses);
     }
 
+    public String getDisplayName(Locale locale) {
+        logger.fine("Getting display name for dataset type " + name + " and locale " + locale);
+        if (locale == null) {
```
poikilotherm, line 166: The lookup in BundleUtil is nullsafe - wouldn't it be OK to use that?

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
+            // The English strings are in datasetTypes.properties (no _en).
+            logger.fine("Locale is English, returning default display name: " + displayName);
+            return displayName;
+        }
+        String propertiesFile = "datasetTypes_" + locale.toLanguageTag() + ".properties";
+        try {
+            logger.fine("Looking up " + name + ".displayName in " + propertiesFile);
+            return BundleUtil.getStringFromPropertyFile(name + ".displayName", "datasetTypes", locale);
+        } catch (MissingResourceException e) {
+            logger.warning(name + ".displayName missing from " + propertiesFile + " (or file does not exist). Returning English version.");
```
poikilotherm: Please don't make this a warning. It will flood the server logs when everything is perfectly fine. Also, if the file does not exist, the resource bundle lookup will fail and return null. The missing exception is only thrown if the file exists, but has no entry for the key.

(Yes, it would be good to restructure the lookup functions or at least add explanatory Javadocs!!!)

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -105,11 +152,65 @@ public JsonObjectBuilder toJson() {
...
+            return displayName;
+        }
+    }
+
+    public String getDescription(Locale locale) {
```
poikilotherm, line 190: This is a mostly duplicated code bit. Can't the two be united?

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
```diff
@@ -36,19 +41,37 @@
...
+     * Machine readable name to use via API.
+     */
     // Any constraints? @Pattern regexp?
     @Column(nullable = false)
     private String name;
 
+    /**
+     * Human readable name to show in the UI.
+     */
+    @Column(nullable = false, columnDefinition = "VARCHAR(255) DEFAULT ''")
```
poikilotherm, line 66: On one hand the displayName is enforced on the API. Why isn't it enforced in the DB as well?

src/main/java/edu/harvard/iq/dataverse/search/SolrSearchServiceBean.java
```diff
@@ -759,12 +759,33 @@ public SolrQueryResponse search(
...
+                         * }
+                         * See also https://github.com/IQSS/dataverse-frontend/issues/809
+                         * and https://github.com/IQSS/dataverse/issues/11758 .
+                         *
+                         * We recognize that this will be a problem for internationalizing the SPA but
+                         * the SPA will likely have similar problems with facets like publicationStatus
+                         * where the labels are in English (e.g. {"Draft":42}). The Search API much use
+                         * the English string when faceting (e.g. "fq=publicationStatus:Draft").
+                         */
+                        localefriendlyName = facetFieldCount.getName();
```
poikilotherm, line 786: Indexing with the machine readable name is a very wise choice!
This is what happens, right?

src/main/java/edu/harvard/iq/dataverse/Dataverse.java
```diff
@@ -166,6 +166,12 @@ public String getIndexableCategoryName() {
     }
 
     private String affiliation;
+
+    /**
+     * If null, only the default dataset type (dataset) is allowed.
+     * See AbstractCreateDatasetCommand.
+     */
+    private String allowedDatasetTypes;
```
poikilotherm: Why is this a string???? "types" is plural, so this should be a `Set` or `List`, making it much easier to handle in other code.

If you want to avoid another FK+table, why not make JPA de/serialize it as comma separated or JSON? Using an `@Convert(converter = SetConverter.class)` is pretty easy!

Obviously not using a table has drawbacks if you need more complicated aggregations etc on the collection. But as you're using a simple `String` so far, maybe that's fine.

Another drawback to this way of a String-serialized list is that we can't just return the collection of types per collection for usage in the commands (which is why  there is list parsing in them...). We'D at least make this a bit simpler & cleaner by having a `Collection<String>` (set or list).

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -358,6 +358,10 @@ public static JsonObjectBuilder json(Dataverse dv, Boolean hideEmail, Boolean re
         if (childCount != null) {
             bld.add("childCount", childCount);
         }
+        String allowedDatasetTypes = dv.getAllowedDatasetTypes();
+        if (allowedDatasetTypes != null) {
+            bld.add("allowedDatasetTypes", allowedDatasetTypes);
```
poikilotherm: IMHO this should use a `JsonArray`, not a plain string.

src/main/java/propertyFiles/datasetTypes.properties
```diff
@@ -0,0 +1,9 @@
+# This file contains the strings for dataset types that can be
+# translated into other languages (French, Spanish, etc.).
+# Only the default dataset type (dataset) is included, as an example.
+# If you add additional dataset types (e.g. software), you don't
+# need to add them to this file if you are only running in English.
+# However, if you are running in additional languages, you should
+# add the additional dataset types to this file.
+dataset.displayName=Dataset
+dataset.description=A study, experiment, set of observations, or publication that is uploaded by a user. A dataset can comprise a single file or multiple files.
```
poikilotherm: This is missing entries for the other types we support (software, review, workflow).
Yes you say this is an example. Yet it is not complete and will require people to lookup the other keys from code. This makes localization unnecessary hard. We should make it easier for the community to just get this thing started and not have them figure out how to add the additional entries when syncing etc.

src/main/java/edu/harvard/iq/dataverse/i18n/i18nUtil.java
```diff
@@ -0,0 +1,25 @@
+package edu.harvard.iq.dataverse.i18n;
+
+import java.util.List;
+import java.util.Locale;
+
+public class i18nUtil {
```
poikilotherm: Classes in Java usually follow Pascal Case.

src/main/java/edu/harvard/iq/dataverse/i18n/I18nUtil.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Locale;
+
+public class i18nUtil {
+
+    /**
+     * @param acceptLanguageHeader The Accept-Language header value such as
+     *                             "Accept-Language: en-US,en;q=0.5"
+     * @return The first Locale or null.
+     */
+    public static Locale parseAcceptLanguageHeader(String acceptLanguageHeader) {
```
poikilotherm, line 34: Why not just use JAX-RS included `HttpHeaders` class and their implementation of the languages header parsing? I don't see any cross validation here to figure out supported languages of the installation and adapt accordingly.
If you plan on adding something like that, it may be a good idea to keep my comments on where to put such things in JAX-RS apps in mind.

src/main/resources/db/migration/V6.9.0.1.sql
```diff
@@ -0,0 +1,10 @@
+-- Add displayname column to datasettype table
```
poikilotherm, line 1: 1. Here's an idea: how about using a Java based migration and read these lines from the properties file instead of duplicating them?

2. If that seems to risky / complex, I'd appreciate putting the DDL changes into one transaction and then have another for the updates. This keeps things more readable.

3. In addition, as commented on the property file: if there already are other types we officially support (software, workflow, review) in the database, we should have a migration for these, too!

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -5853,7 +5867,7 @@ public Response deleteDatasetType(@Context ContainerRequestContext crc, @PathPar
         try {
             idToDelete = Long.parseLong(doomed);
         } catch (NumberFormatException e) {
-            throw new IllegalArgumentException("ID must be a number");
+            return error(BAD_REQUEST,"ID must be a number");
```
poikilotherm, line 5888: Instead of all the manual parsing and validating the string, just make it a plain and simple `@PathParam("id") @Min(1) Long doomed`. This habit of doing things with own code that the solid framework of bean validation already gives you adds unnecessary complexity and room for errors. (And it's not even complete, as it allows negative numbers, which makes no sense.)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
```diff
@@ -124,8 +129,36 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         DatasetType existingDatasetType = theDataset.getDatasetType();
         logger.fine("existing dataset type: " + existingDatasetType);
         if (existingDatasetType != null) {
-            // A dataset type can be specified via API, for example.
-            theDataset.setDatasetType(existingDatasetType);
+            List<DatasetType> allowedDatasetTypes = new ArrayList<>();
+            String allowedDatasetTypesString = theDataset.getOwner().getAllowedDatasetTypes();
```
poikilotherm: - A) This code must not parse the list. The model is  responsible for that. See my comment there on JPA converters.

- B) The deserialization/lookup from type names to actual POJO instances is incorrectly located here. It should be moved to either the context or a service layer. The command should not hold this contract, as it is likely to be reused in other places.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -5853,7 +5867,7 @@ public Response deleteDatasetType(@Context ContainerRequestContext crc, @PathPar
         try {
             idToDelete = Long.parseLong(doomed);
```
poikilotherm, line 5886: This method should also allow using a name or an ID, as the other methods. This is especially important now that we return a list of names for allowed types for a collection.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
```diff
@@ -116,4 +126,36 @@ private void setBooleanAttribute(CommandContext ctxt, boolean adminOnly) throws
...
+                    this, null, dataverse);
+        }
+        if (!(allowedDatasetTypesIn instanceof String stringValue)) {
+            throw new IllegalCommandException("'" + ATTRIBUTE_ALLOWED_DATASET_TYPES + "' requires a string value",
+                    this);
+        }
+
+        List<String> invalidDatasetTypes = new ArrayList<>();
+
+        String[] allowedDatasetTypeNames = stringValue.split(",");
```
poikilotherm, line 143: Technically, all of this class does things typically located in a service layer. But fine... At least, we should not use a simple comma separated list - a JSON list is much cleaner, as we are using a JSON API and other places like settings etc also support using of JSON Arrays. At some point we really should refactor the API, service and command layer to deal with all of this in a less chaotic way.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
```diff
@@ -116,4 +126,36 @@ private void setBooleanAttribute(CommandContext ctxt, boolean adminOnly) throws
...
+        for (String datasetTypeName : allowedDatasetTypeNames) {
+            DatasetType datasetType = ctxt.datasetTypes().getByName(datasetTypeName.trim());
+            if (datasetType == null) {
+                invalidDatasetTypes.add(datasetTypeName);
+            }
+        }
+
+        if (!invalidDatasetTypes.isEmpty()) {
```
poikilotherm, line 153: It is debatable if the validation of the names should be done in this class or in a separate service layer. As we currently don't have a service layer for DatasetTypes, maybe this one will do. (This is a design flaw of our REST API, which mixes being entry and service layer a lot. As stated above, this whole class mimics a service layer...)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
```diff
@@ -116,4 +126,36 @@ private void setBooleanAttribute(CommandContext ctxt, boolean adminOnly) throws
...
+            Map<String, String> fieldErrors = new HashMap<>();
+            fieldErrors.put(ATTRIBUTE_ALLOWED_DATASET_TYPES, "The following dataset types do not exist: "
+                    + String.join(", ", invalidDatasetTypes));
+            throw new InvalidFieldsCommandException("The collection could not be updated because "
+                    + ATTRIBUTE_ALLOWED_DATASET_TYPES + " is invalid.", this, fieldErrors);
+        }
+
+        dataverse.setAllowedDatasetTypes(stringValue);
```
poikilotherm: As described in the comments on the entity class, this should trade a collection, not a string. Leave the serialization to the model class - we only deal with user input here.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: We discussed this. I'll leave it alone for now but I'll put a comment in the code about these options I didn't know about.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: Sure, I can add delete by name. When I do, I'll also resolve https://github.com/IQSS/dataverse/pull/11753#discussion_r2788419843

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
pdurbin: Yeah, let's make this an array.

Also, let's update JsonParser to parse the array.

src/main/java/edu/harvard/iq/dataverse/Dataverse.java
pdurbin: We should probably do a List of type DatasetType.

pdurbin: Please see d2f12dfd9e - switch allowedDatasetTypes in entity from String to a ManyToMany List

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
pdurbin: Whoops, yes. Fixing!

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: Of course not. I was being lazy, thinking about this:

Screenshot 2026-02-12 at 12 10 31 PM

pdurbin: Where were you when @michbarsinai opened the following issue? 😅 

- #4284

Should we re-open it? 🤔

pdurbin: @poikilotherm as an IntelliJ user, you want me to click on one of these?

Screenshot 2026-02-12 at 12 20 31 PM

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: I'm happy to change this. What do you suggest?

src/main/java/propertyFiles/datasetTypes.properties
jggautier:
```suggestion
dataset.description=A study, experiment, set of observations, or publication. A dataset can comprise a single file or multiple files.
```

src/main/resources/db/migration/V6.9.0.1.sql
```diff
@@ -0,0 +1,8 @@
...
+ALTER TABLE datasettype ADD COLUMN IF NOT EXISTS displayname VARCHAR(255) NOT NULL DEFAULT '';
+-- Set displayname for dataset
+UPDATE datasettype SET displayname = 'Dataset' WHERE name = 'dataset';
+-- Add description column to datasettype table
+ALTER TABLE datasettype ADD COLUMN IF NOT EXISTS description VARCHAR(255);
+-- Set description for dataset
+UPDATE datasettype SET description = 'A study, experiment, set of observations, or publication that is uploaded by a user. A dataset can comprise a single file or multiple files.' WHERE name = 'dataset';
```
jggautier:
```suggestion
UPDATE datasettype SET description = 'A study, experiment, set of observations, or publication. A dataset can comprise a single file or multiple files.' WHERE name = 'dataset';
```

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: This is the same comment at https://github.com/IQSS/dataverse/pull/11753#discussion_r2788066853 so I'll resolve this one.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: I'm confused. I thought I was enforcing it in the DB with this:

`ALTER TABLE datasettype ADD COLUMN IF NOT EXISTS description VARCHAR(255);`

src/main/java/edu/harvard/iq/dataverse/search/SolrSearchServiceBean.java
pdurbin: Yep! Resolving. (Sorry, many comment to keep track of!)

This issue is related, which I'm fixing in this PR:

- #11758
- https://github.com/IQSS/dataverse-frontend/issues/809 (better writeup of the fix)

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
pdurbin: This has been changed to an array: https://github.com/IQSS/dataverse/commit/d2f12dfd9eec3d782e68493b6f9293fecaf60b9e

Woof. JsonParser. I looked at it briefly but I'm not sure I want to dive into this. 😬

src/main/java/propertyFiles/datasetTypes.properties
pdurbin: This is potentially a long conversation.

Should we move away from the current flexibility we have where any arbitrary dataset type can be created? Should we put a select few types in the database (such as "dataset", "software", "workflow", and "review") and have a mechanism for enabling them? And remove the ability to add arbitrary types?

This properties file is the way it is because only one type, dataset, ships with Dataverse. All other types have to be added manually via API, like we do with licenses. The pattern is the same. In https://github.com/IQSS/dataverse/blob/v6.9/src/main/java/propertyFiles/License.properties we only provide examples of licenses we ship with Dataverse. It's an exercise for the reader to add translations for licenses they add manually.

src/main/java/edu/harvard/iq/dataverse/i18n/i18nUtil.java
pdurbin: Bah, you're right.

src/main/java/edu/harvard/iq/dataverse/i18n/I18nUtil.java
pdurbin: This is pretty much the same comment as https://github.com/IQSS/dataverse/pull/11753#discussion_r2787986962 so I'm resolving this one (but I hear you!).

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: As discussed here, I plan to add delete by name: https://github.com/IQSS/dataverse/pull/11753#discussion_r2788495231

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
pdurbin: Fixed in this commit: https://github.com/IQSS/dataverse/commit/d2f12dfd9eec3d782e68493b6f9293fecaf60b9e

I'll resolve this but let's discuss over here: https://github.com/IQSS/dataverse/pull/11753#discussion_r2788307928

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
pdurbin: As I said on the call yesterday (thanks!), I'm a bit weirded out by the idea of passing JSON in a query parameter, but sure, let's keep talking about it.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
pdurbin: Maybe we're due for another rewrite! 😅

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
pdurbin: Fixed in this commit: https://github.com/IQSS/dataverse/commit/d2f12dfd9eec3d782e68493b6f9293fecaf60b9e

I'll resolve this but let's discuss over here: https://github.com/IQSS/dataverse/pull/11753#discussion_r2788307928

src/main/java/edu/harvard/iq/dataverse/i18n/i18nUtil.java
pdurbin: Fixed in f48b48e6cd. Resolving.

src/main/resources/db/migration/V6.9.0.1.sql
pdurbin: Let's keep talking about this.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
pdurbin: Fixed in 828094798. Resolving. Good catch. Thanks!

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
pdurbin: I added the javadoc in 8280947. Resolving.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: Fixed in 8280947. Resolving.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
```diff
@@ -121,11 +126,33 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
...
+            List<DatasetType> allowedDatasetTypesFinal = new ArrayList<>();
+            if (allowedByCollection.isEmpty()) {
+                // If allowedDatasetTypes is unspecified, assume
+                // only the default type (dataset) is allowed
+                allowedDatasetTypesFinal.add(defaultDatasetType);
+            } else {
+                allowedDatasetTypesFinal.addAll(allowedByCollection);
+            }
+            // Set type if allowed. Otherwise, return error.
+            if (allowedDatasetTypesFinal.contains(existingDatasetType)) {
```
poikilotherm: DatasetType does not implement `equals()` and `hashCode()`. This means `contains()` will rely on equality of the object reference, not its values. DatasetType should override and implement these methods.

(This is a subtle thing and might not even appear as a problem in usual tests, as in a single JVM with no JPA based lookups the equality byobject reference probably is given.)

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
poikilotherm: Both SQL in the migration and generated by the `@Column` here are a match (non null, default = ''). I'm saying: "why the default empty string"? On a database level we're saying "whatever", while in the API we say "must provide a value". I suggest we drop the "DEFAULT" bit and make sure to have a sane default value set for pre-existing types. (Putting my idea into the migration SQL: uppercase first-char of name attribute.)

src/main/resources/db/migration/V6.9.0.1.sql
poikilotherm: I may have an easier idea: how about taking any existing types' `name`, transform first char to uppercase and use as `displayName`? That should cover 95% of all use cases out there, as it is very valid English.

src/main/java/edu/harvard/iq/dataverse/i18n/I18nUtil.java
```diff
@@ -0,0 +1,46 @@
...
+        if (acceptLanguageHeader == null || acceptLanguageHeader.isEmpty()) {
+            return null;
+        }
+        List<Locale.LanguageRange> list = Locale.LanguageRange.parse(acceptLanguageHeader);
```
poikilotherm, line 38: Are we OK with any IllegalArgumentExceptions bubbling up?

> if a language range or a weight found in the given ranges is ill-formed

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -358,6 +358,18 @@ public static JsonObjectBuilder json(Dataverse dv, Boolean hideEmail, Boolean re
         if (childCount != null) {
             bld.add("childCount", childCount);
         }
+        List<DatasetType> allowedDatasetTypes = dv.getAllowedDatasetTypes();
+        if (allowedDatasetTypes != null && !allowedDatasetTypes.isEmpty()) {
+            JsonArrayBuilder jab = Json.createArrayBuilder();
+            for (DatasetType datasetType : allowedDatasetTypes) {
+                NullSafeJsonBuilder json = NullSafeJsonBuilder.jsonObjectBuilder()
+                    .add("name", datasetType.getName())
+                    .add("displayName", datasetType.getDisplayName())
```
poikilotherm, line 367: Are we fine with not returning localized strings here even if a Locale was requested via API Header? Wasn't the idea to spare the SPA a second trip and have the values localized to avoid fragment disruption? (I may have this wrong in my head. Feel free to ignore.)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
```diff
@@ -116,4 +126,39 @@ private void setBooleanAttribute(CommandContext ctxt, boolean adminOnly) throws
...
+        }
+        if (!(allowedDatasetTypesIn instanceof String stringValue)) {
+            throw new IllegalCommandException("'" + ATTRIBUTE_ALLOWED_DATASET_TYPES + "' requires a string value",
+                    this);
+        }
+
+        List<DatasetType> allowedDatasetTypes = new ArrayList<>();
+        List<String> invalidDatasetTypes = new ArrayList<>();
+
+        String[] allowedDatasetTypeNames = stringValue.split(",");
```
poikilotherm, line 143: The API endpoint doesn't do any input validation on the value. We should not allow a blank string here, as the resulting error would be confusing (empty spaces). Also, while at it, catch potential null values as a good practice.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
poikilotherm: I am suggesting transforming into a oneliner, using formatting and lambda expression to save CPU cycles.

```java
logger.fine(() -> String.format("Looking up %s.displayName in datasetTypes_%s.properties", name, locale.toLanguageTag()));
```

and 

```java
logger.fine(() -> String.format("%s.displayName missing from datasetTypes_%s.properties (or file does not exist). Returning English version.", name, locale.toLanguageTag()));
```

This code is still problematic, as it makes debugging statements about behaviour inside the BundleUtil that may or may not be correct once its internal logic changes. So it may send people on a false path down the road. It would be better to handle this inside the BundleUtil or, if we want to enable better debugging here, use the exception message if the exception has been thrown. Up to you.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
poikilotherm: I added a suggestion below with how to use a simple lambda expression and string formatting. This much more concise and readable when compared to java.util.logging with an Object[] for multiple positional parameters.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
pdurbin: Yes. Good point. I'll take a look at this.

pdurbin: Pushed: https://github.com/IQSS/dataverse/pull/11753/changes/c3b4de6410b4c2040041b677c087f668fb3fa8e2

src/test/java/edu/harvard/iq/dataverse/api/DatasetTypesIT.java
```diff
@@ -713,4 +792,313 @@ public void testCreateDatasetWithCustomType() {
...
+
+        Response setAllowedDatasetTypesFail = UtilIT.setCollectionAttribute(dataverseAlias, "allowedDatasetTypes",
+                nonExistentTypes, apiTokenSuperuser);
+        setAllowedDatasetTypesFail.prettyPrint();
+        setAllowedDatasetTypesFail.then().assertThat()
+                .statusCode(BAD_REQUEST.getStatusCode());
+
+        Response failEmptyString = UtilIT.setCollectionAttribute(dataverseAlias, "allowedDatasetTypes", " ",
+                apiTokenSuperuser);
+        failEmptyString.prettyPrint();
```
pdurbin, line 837: This is what the response looks like:

```
{
    "status": "ERROR",
    "message": "The collection could not be updated because allowedDatasetTypes is invalid.",
    "fieldErrors": {
        "allowedDatasetTypes": "The following dataset types do not exist:  "
    }
}
```

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/UpdateDataverseAttributeCommand.java
pdurbin: I added a test, at least, that demostrates that an empty string can't be passed. See https://github.com/IQSS/dataverse/pull/11753/changes#r2874654908

src/main/resources/db/migration/V6.9.0.1.sql
pdurbin: I just pushed https://github.com/IQSS/dataverse/pull/11753/changes/cc5b11d9b08f06cfd3ded9a258aedf8468e698a9 which will set displayName to Software for software, etc.

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: For the SQL side, please see https://github.com/IQSS/dataverse/pull/11753/changes#r2874761915

pdurbin:
I believe I've addresses all the code changes I talked about with @scolapasta and @jp-tosca yesterday.

I'll create a follow up issue about further improvements we can make after the code freeze for 6.10 has been lifted.

doc/sphinx-guides/source/admin/dataverses-datasets.rst
```diff
@@ -109,6 +109,78 @@ If the :AllowedCurationLabels setting has a value, one of the available choices
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Non-dataset types, such as the "review" type, are only available when a collection admin has enabled them, via API.
+
+Using the API :ref:`collection-attributes-api`, change the ``allowedDatasetTypes`` attribute so that it includes "review". If you only want to allow reviews, you can pass just ``review``. If you want to allow multiple dataset types, you can pass a comma-separated list, such as ``review,dataset``.
+
+Invite Users to Create Review Datasets
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+At this point, users should be able to create review datasets via API, you gave them permission on the collection. You can point them to :ref:`creating-a-review-dataset` for details.
```
scolapasta: this sentence doesn't quite make sense.

pdurbin: Good catch! I forgot an "if". Added in ce4a48f9ff.

src/main/resources/db/migration/V6.9.0.1.sql
poikilotherm: This is certainly the simplest and thus best solution for this. Great job! 😸

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
poikilotherm: Looks good to me now, let's run with this.

src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/XmlMetadataTemplate.java
```diff
@@ -836,12 +836,15 @@ private void writeResourceType(XMLStreamWriter xmlw, DvObject dvObject) throws X
...
+        String datasetTypeName = null;
         if (dvObject instanceof Dataset dataset) {
-            String datasetTypeName = dataset.getDatasetType().getName();
+            datasetTypeName = dataset.getDatasetType().getName();
             resourceType = switch (datasetTypeName) {
             case DatasetType.DATASET_TYPE_DATASET -> "Dataset";
             case DatasetType.DATASET_TYPE_SOFTWARE -> "Software";
             case DatasetType.DATASET_TYPE_WORKFLOW -> "Workflow";
+            // "Other" for now but we might ask DataCite to support "Review"
+            case DatasetType.DATASET_TYPE_REVIEW -> "Other";
```
poikilotherm, line 850: TIL about https://datacite-metadata-schema.readthedocs.io/en/4.7/appendices/appendix-1/resourceTypeGeneral/#peerreview Is this what you've been looking for?

pdurbin: We considered PeerReview but in the end went with Other. Here's the conversation from https://iqss.slack.com/archives/C099P44D1NF/p1755091921570099

Image

poikilotherm: Fair enough. I'm glad you already set the resource type as recommended by DataCite.

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -847,21 +847,113 @@ Dataset Types
 
 .. note:: Development of the dataset types feature is ongoing. Please see https://github.com/IQSS/dataverse-pm/issues/307 for details.
 
-Out of the box, all datasets have a dataset type of "dataset". Superusers can add additional types such as "software" or "workflow" using the :ref:`api-add-dataset-type` API endpoint.
+The vision for dataset types is to have variations on datasets. The best documented use case is :ref:`review-datasets-user`, as explained below, but other types of datasets are possible such as software datasets (see :ref:`api-add-dataset-type` for an example) or workflow datasets.
 
-Once more than one type appears in search results, a facet called "Dataset Type" will appear allowing you to filter down to a certain type.
+Out of the box, all datasets have a dataset type of "dataset", which is the traditional dataset in Dataverse. Superusers can add additional types using the :ref:`api-add-dataset-type` API endpoint. These additional dataset types cannot be used until a superuser has allowed them on a per-collection basis using the :ref:`collection-attributes-api` API endpoint (by passing ``allowedDatasetTypes``).
```
poikilotherm:
```suggestion
Out of the box, all datasets have a dataset type of "Dataset", which is the traditional research data related dataset in Dataverse. Superusers can add additional types using the :ref:`api-add-dataset-type` API endpoint. These additional dataset types cannot be used until a superuser has allowed them on a per-collection basis using the :ref:`collection-attributes-api` API endpoint (by passing ``allowedDatasetTypes``).
```

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -847,21 +847,113 @@ Dataset Types
...
 - :ref:`import-dataset-with-type`
 
-Dataset types can be listed, added, or deleted via API. See :ref:`api-dataset-types` in the API Guide for more.
+Once more than one type appears in dataset search results, a facet called "Dataset Type" will appear allowing you to filter down to a certain type.
+
+Dataset types can be linked with metadata blocks to make fields from those blocks available when datasets of that type are created or edited (via API). See :ref:`api-link-dataset-type` and :ref:`list-metadata-blocks-for-a-collection` for details.
+
+Dataset types can change the "type" in Citation Style Language (CSL) output. See :ref:`get-citation-in-other-formats` for details.
+
+If your installation is configured to use DataCite as a persistent ID (PID) provider, the dataset type may be sent to DataCite as ``resourceTypeGeneral``. See the table under :ref:`dataset-types-datacite` for details.
```
poikilotherm:
```suggestion
If your installation is configured to use DataCite as a persistent ID (PID) provider, the dataset type may be sent to DataCite as ``resourceTypeGeneral`` (see also `upstream schema <https://datacite-metadata-schema.readthedocs.io/en/latest/properties/resourcetype/>`_). See the table under :ref:`dataset-types-datacite` for details.
```

doc/sphinx-guides/source/user/dataset-management.rst
```diff
@@ -847,21 +847,113 @@ Dataset Types
...
+
+We recommend implementing a policy where there is only one review of a given resource per collection.
+
+Almost all functionality is the same between regular datasets and review datasets. Review datasets build upon existing dataset functionality such as custom metadata blocks, versioning, publishing workflows, permissions, and file handling.
+
+Review datasets build on the :ref:`dataset-types` feature, allowing users to choose a dataset type of "review", which leads to a couple differences from regular datasets.
+
+First, when multiple dataset types exist, a "Dataset Type" search facet appears that allows users to narrow results to the various kinds of dataset types that have been added, such as dataset, review, software, workflow, etc. (Under the "Collections, Datasets, Files" area, review datasets are considered datasets.)
+
+Second, when review datasets are published, different ``resourceType`` metadata is sent to DataCite. Review datasets send "Other" for the field resourceTypeGeneral ("Work Type" in the UI at https://commons.datacite.org). See the table under :ref:`dataset-types-datacite` for details.
```
poikilotherm:
```suggestion
Second, when review datasets are published, different ``resourceType`` metadata is sent to DataCite. Review datasets send "Other" for the field ``resourceTypeGeneral`` ("Work Type" in the UI at https://commons.datacite.org) and "Review" as type value. See the table under :ref:`dataset-types-datacite` for details and comparison. Please note that we are not using the official general type "PeerReview" because it is too narrow to fit all possible use cases. 
```

src/main/java/edu/harvard/iq/dataverse/dataset/DatasetType.java
pdurbin: Here's more of the story from https://iqss.slack.com/archives/C010LA04BCG/p1759948142160559

Image
Image

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/AbstractCreateDatasetCommand.java
```diff
@@ -121,11 +126,33 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         }
         
         DatasetType defaultDatasetType = ctxt.datasetTypes().getByName(DatasetType.DEFAULT_DATASET_TYPE);
+        // Why is this called existingDatasetType? Would incomingDatasetType make more sense?
         DatasetType existingDatasetType = theDataset.getDatasetType();
         logger.fine("existing dataset type: " + existingDatasetType);
         if (existingDatasetType != null) {
-            // A dataset type can be specified via API, for example.
-            theDataset.setDatasetType(existingDatasetType);
+            List<DatasetType> allowedByCollection = theDataset.getOwner().getAllowedDatasetTypes();
```
pdurbin: @scolapasta found a bug in this part of the code. As of 050439d, in JSF, one is still able to create a dataset in a collection that only allows review. I have a fix ready, which I'll push.

pdurbin: Fixed in 2b03a64ba5. Here's how the error looks in JSF:

Image

pdurbin:
Based on testing, I made the following changes:

2b03a64ba5 better handling of collections disallowing "dataset" datasetType #12115 #11747
38ec9a74b9 move "create collection" step down, just before "allow type" #11747
bebb87b62c add sample file dataset-create-review.json #11747

I also opened the following issue which can be observed in this pull request as well as the "develop" branch:

- #12196

PR #11755, opened 2025-08-20
External Tools research

g-saracca:
Draft PR only for Testing using the GHCR generated image

- Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -58,6 +58,9 @@ import edu.harvard.iq.dataverse.workflow.WorkflowServiceBean;
 import jakarta.ejb.EJB;
 import jakarta.ejb.EJBException;
 import jakarta.inject.Inject;
+import jakarta.json.Json;
+import jakarta.json.JsonArrayBuilder;
+import jakarta.json.JsonObjectBuilder;
 import jakarta.json.*;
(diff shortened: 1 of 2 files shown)


qqmyers:
I understand this is a draft PR for testing. FWIW: for production use, I think it would be better to split getting the listing of tools from getting the signed URLs, as we do now. A lot depends on whether the tools remain separate apps - if they aren't, they may not need signed URLs (but we would loose the ability to control what they can do if they can just use the SPA authentication.) Currently a tool is launched by giving it one short-lived signed URL as a base64-encoded token and it calls that to get all of the params and (potentially longer-lived) signed URLs its configured for, but all of this happens only when a tool is run, not when the initial listing of tools is shown.

g-saracca:
> I understand this is a draft PR for testing. FWIW: for production use, I think it would be better to split getting the listing of tools from getting the signed URLs, as we do now. A lot depends on whether the tools remain separate apps - if they aren't, they may not need signed URLs (but we would loose the ability to control what they can do if they can just use the SPA authentication.) Currently a tool is launched by giving it one short-lived signed URL as a base64-encoded token and it calls that to get all of the params and (potentially longer-lived) signed URLs its configured for, but all of this happens only when a tool is run, not when the initial listing of tools is shown.

Thanks Jim, I understand. I think I could use the already existing api endpoint to get all the available tools and then we could create a new endpoint to get the final url processed by the backend logic? But this last part only happening when the tool (iframe) is in view and not before?
Right now we need to support the File previewers.

PR #11759, opened 2025-08-21
Revert "Added a control to publishing a Dataset and cleaning dataversesubjects table"

pdurbin:
- Reverts IQSS/dataverse#11551

We're reverting this because we didn't notice until after it was merged that the Flyway script had the wrong number.

As of this writing it should be V6.7.1.2.sql

Our docs on number are here: https://guides.dataverse.org/en/6.7.1/developers/sql-upgrade-scripts.html

This one slipped through the cracks because we had an unexpected point release (6.7.1). At the time the PR was approved the number was correct.


ofahimIQSS:
this failed continuous integration (ansible terminated) - rerunning the job then merging after it passes

ofahimIQSS:
tests passed - merging

PR #11762, opened 2025-08-22
Fixing ror v2 api access for cvoc

ffritze:
The ROR-API has changed during the last weeks. In the future ROR API v1 will be deprecated.

The old search path for the termName in the old CVOC ROR script was "/value"
In version 2 of the ROR API this will no longer work. So the new search path in the cvocConf.json should be 
""termName": {
                "pattern": "{0}",
                "params": [
                    "/names/0/value"
                ]
            },"

Unfortunately the current implementation in DV 6.7.1 cannot handle the "0" index. It reads it as a String but it must be an Integer number regarding the JSON output from ROR V2.

This pull request provides a fix the specific implementation and it can handle now both String and Number as values for the ROR JSON output.

In cvocJson.conf after applying this pull request. Your can use the new ROR API endpoint: "https://api.ror.org/organizations/{0}"

- Closes #

**Additional documentation**:

This pull request is only necessary for the ROR API V2.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
@@ -794,7 +794,7 @@ public class DatasetFieldServiceBean implements java.io.Serializable {
                 }
 
             } else {
-                if (NumberUtils.isCreatable(pathParts[index])) {
+                if ((curPath instanceof JsonArray) && NumberUtils.isCreatable(pathParts[index])) {
                     try {
                         int indexNumber = Integer.parseInt(pathParts[index]);
                         curPath = ((JsonArray) curPath).get(indexNumber);
@@ -805,7 +805,7 @@ public class DatasetFieldServiceBean implements java.io.Serializable {
                     curPath = ((JsonObject) curPath).get(pathParts[index]);
                 }
                 // curPath = ((JsonObject) curPath).get(pathParts[index]);
-                logger.fine("Found next Path object " + curPath.toString());
+                logger.fine("Found next Path object " + ((curPath == null) ? "null" : curPath.toString()));
                 return processPathSegment(index + 1, pathParts, curPath, termUri);
             }
         } else {


src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
```diff
@@ -764,7 +765,17 @@ Object processPathSegment(int index, String[] pathParts, JsonValue curPath, Stri
                 }
 
             } else {
-                curPath = ((JsonObject) curPath).get(pathParts[index]);
+                if (NumberUtils.isCreatable(pathParts[index])) {
```
qqmyers: Is it worth handling the case where "0" is returned as a string too? This code assumes anything that can be a number is one.

qqmyers:
I think this is a useful update but I'm not sure that ROR v2 needs it: in https://github.com/gdcc/dataverse-external-vocab-support/pull/46 I'm suggesting that we just get the whole array with "params": ["/name"]

This is consistent with what we do with Skosmos, etc. where we get all the variants and names in different languages if they exist.

qqmyers:
Another ROR specific note - I'm not sure 0 is always the best to display )if you do want only one name) - in the PR the new ror.js script looks for the entry with the "ror_display" tag, which is what shows on their website.

ffritze:
I think the solution in ror.js is exactly what we need.
But the "params" configuration in cvocConf.json is a little bit different.
I would like to draw the attention to the case that if you specify "/names/0/value"  then you will get the following database entries (First attachment)
Bildschirmfoto vom 2025-08-22 13-49-11
But If you specify only "/names" you will get a whole JSON array in the database. Is that what you want (second attachment)?
Bildschirmfoto vom 2025-08-22 13-54-01

qqmyers:
Yes, having the array was my intent. That means in the exports like OAI_ORE, you'll see:```
"author": [
{
"citation:authorName": "Myers, Jim",
"citation:authorAffiliation": {
"scheme": "http://www.grid.ac/ontology/",
"termName": [
{
"lang": null,
"types": [
"acronym"
],
"value": "QDR"
},
{
"lang": "en",
"types": [
"ror_display",
"label"
],
"value": "Qualitative Data Repository"
}
],
"@type": "https://schema.org/Organization",
"@id": "https://ror.org/014trz974"
},
``` showing both the acronym and label (and other languages if they exist). That said, given that https://ror.org/014trz974 is really all that's needed to identify the affiliation org for machines, it's basically personal preference as to what else to include in the metadata exports. If having just "QDR" in the output is OK, the filtering could just get /names/0/value as you suggest.

stevenwinship:
Waiting on conflict resolution

src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
ffritze: I would say a number is a number. If you construct a JSON object then you shouldn't use a number as a string value for a key.

qqmyers: Let me rephrase - the code currently assumes that if the value is interpretable as a number, we must be parsing an array, whereas if the value is a json object, the keys would have to be strings. I don't know how often we'll encounter such things (objects with keys "1", "2", "3"), but it seems like it might be cleaner/safer to see if the value can be a number && the curPath is a json array, you're new code is invoked, otherwise it tries to deal with it as an object with a string key.

qqmyers:
FWIW: I'm reconsidering where using /names and retrieving the array is the best option here. While I think some export formats would benefit from being able to show all the options, our current DataCiteXML assumes that the value is a string, and, prior to #11782, getting an array would break things. 
While I think this PR is useful, for this case, I'm looking at making our parsing slightly more flexible such that we can look within the array for the object with the type "ror_display" - the value ROR shows in their UI, rather than picking the first array element (which would be supported by this PR). We have the rigth syntax for this already from our SkosMos examples, e.g. /names/types=ror_display/value but with that syntax, we assume the "types" key has a string value (that would match "ror_display") whereas in ROR, the types key is an array of strings, one of which has the value ror_display. I'll be putting in a PR for that - hopefully both PRs will have general utility.

qqmyers:
FYI: I created #11793 and added a generic release note that says there are enhancements to the ext. vocab mechanism, which should cover this PR as well (assuming both go into Dataverse 6.9 (which we want so that the additions are available before ROR's v1 API gets shut off).

src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
qqmyers: @ffritze - apologies for letting this sit. I'm suggesting the following change: https://github.com/IQSS/dataverse/commit/5cd4ecf3fcb7b2cf3b00b6b355308a6018e88fec . If that's acceptable to you, can you include it, and update your branch from IQSS/develop (as done in https://github.com/IQSS/dataverse/compare/develop...GlobalDataverseCommunityConsortium:dataverse:fix_ror-v2_api_cvoc) so our build system will be able to run it and we can plan on including it in Dataverse 6.9? (FWIW: I wasn't able to make a commit or PR against your branch)

ffritze: @qqmyers Thanks a lot. I will test it in our infrastructure. Currently we have a maintenance phase so I will postpone it until the beginning of next week.

qqmyers (approved):
OK - I'll approve this so it moves to Ready for QA. @ofahimIQSS - make sure @ffritze has completed his testing before you merge.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
ffritze: @qqmyers I can now confirm that your additional code in this branch is working as expected. This pull request is ready for production

pdurbin: @ffritze great! We were leaving you assigned to this PR until you confirmed this. I'll remove you. Have a good weekend!

ffritze: @pdurbin If you want, you can remove me.

landreev:
@qqmyers Could you please help me understand what I'm testing here? - There's no explicit "how to test", but the description says: 
> In cvocJson.conf after applying this pull request. Your can use the new ROR API endpoint: `"https://api.ror.org/organizations/{0}"`

However, on my own dev. instance I already had this CVOC configured with `"retrieval-uri": "https://api.ror.org/organizations/{0}"`. And it appears to be working with the develop branch as well. (I did double-check that the url above defaults to `https://api.ror.org/v2/organizations/{0}`). Is this because your #11793 has been merged already? 
Or do I need to change something else in the `:CVocConf`, or look more closely at what I end up with in the database and in metadata exports, and/or follow the "how to test" instruction from #11793? 

Also, does the release note you've added in the PR above cover or supersede everything in this PR as well? 

Sorry to be needy. I foolishly assumed this was going to be much more straightforward to merge.

qqmyers:
It depends on what's in your retrieval-filtering for the name. If you have https://github.com/gdcc/dataverse-external-vocab-support/blob/1dea77ee6e45388fcb4cb42ae6bf7bb4462f2736/examples/config/authorsOrcidAndRor.json#L29 , with the v2 ROR API you will fail to pick up a name (in the json stored in the externalvocabularyvalue table, and then used in the exports) - since there is only a names array and there's no longer a "name" entry. 

If you switch to `"params":["/names/0/value"]` , this PR will get the first (index 0) value from the names array (specifically the value of the "value" key in the first object in the array). 

#11793 is similar - it allows finding an array element by some value in an object in the array, so that `"params": ["/names/types=ror_display/value"]` will pick the name value in the object with "types" value (which is another array) containing "ror_display". ("ror_display" is the value ROR uses on it's page for the organization, so hopefully better than just getting the first value, but for other situations, getting the first or nth value may be needed, so this PR hopefully prepares us for those cases.)

Note the one trick in testing - if there's a value for the ROR in the externalvocabularyvalue table already, the code to get the value from ROR doesn't run and you won't see any change. You can delete the row or just truncate that table before a test.

landreev:
Tested, confirmed that with these changes, and the latest `:CVocConf` from gdcc/dataverse-external-vocab-support#46 the proper name values are saved, in `externalvocabularyvalue` and the exports. 
I did test with this branch synced up with develop. I take it there's no need to test _just_ the changes in this PR (i.e., without the #11793). (?)

landreev:
All right, merging.

PR #11763, opened 2025-08-25
Enhance Builtin users SPI security

### LINKED ISSUE #11689: Secure API_BEARER_AUTH_USE_BUILTIN_USER_ON_ID_MATCH with hardcoded 'builtin' OIDC attribute

This idea originated [from a discussion](https://github.com/IQSS/dataverse/pull/11622#discussion_r2216017175
) where it was suggested to use a custom attribute to ensure that a token is issued by Keycloak through the Builtin Users SPI. This approach helps prevent identity providers authorized in Dataverse—but not using the SPI—from sending claims that could mistakenly identify a builtin user, potentially leading to an impersonation issue, when API_BEARER_AUTH_USE_BUILTIN_USER_ON_ID_MATCH is enabled.

API User

https://github.com/IQSS/dataverse/pull/11622#discussion_r2216017175

Auth behavior when API_BEARER_AUTH_USE_BUILTIN_USER_ON_ID_MATCH is enabled

None

https://github.com/IQSS/dataverse/pull/11622

Yes

GPortas:

## Security improvements for `api-bearer-auth-use-builtin-user-on-id-match`

We’ve strengthened the security of the `api-bearer-auth-use-builtin-user-on-id-match` feature flag. It will now only work when the provided bearer token includes an `idp` claim that matches the Keycloak Service Provider identifier.

By enforcing this check, the risk of impersonation from other identity providers is significantly reduced, since they would need to be explicitly configured with this specific, non-standard identifier.

- Closes #11689

**Notes for the reviewer**:

I’ve updated the warning in the documentation related to impersonation risk. With this additional security measure in place, user impersonation is unlikely to occur unless an IdP is intentionally misconfigured.

1) Run the containerized dev environment under `conf/keycloak`

2) Obtain a bearer token by logging into Keycloak with built-in account credentials

`curl -X POST http://keycloak.mydomain.com:8090/realms/test/protocol/openid-connect/token   -H "Content-Type: application/x-www-form-urlencoded"   -d "client_id=test"   -d "client_secret=94XHrfNRwXsjqTqApRrwWmhDLDHpIYV8"   -d "grant_type=password"   -d "username=dataverseAdmin"   -d "password=admin1"   -d "scope=openid"`

2) Go to https://www.jwt.io/ and paste the access_token. You should see a new "idp" claim including the value "dv-builtin-users-authenticator" from the Keycloak SPI.

Screenshot 2025-08-27 at 09 48 32

3) Test an API endpoint requiring auth using the generated access token, and make sure it correctly authenticates the user.

No

Yes, attached

**Additional documentation**:

N/A

### DIFF
--- a/conf/keycloak/Dockerfile
+++ b/conf/keycloak/Dockerfile
@@ -29,7 +29,7 @@ COPY --from=builder /app/target/keycloak-dv-builtin-users-authenticator-1.0-SNAP
 
 # Copy additional configurations
 COPY ./builtin-users-spi/conf/quarkus.properties /opt/keycloak/conf/
-COPY ./test-realm.json /opt/keycloak/data/import/
+COPY ./test-realm-include-spi.json /opt/keycloak/data/import/
 
 # Set the Keycloak command
 ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
--- a/conf/keycloak/test-realm-include-spi.json
+++ b/conf/keycloak/test-realm-include-spi.json
@@ -1,0 +1,2383 @@
+{
+  "id": "80a7e04b-a2b5-4891-a2d1-5ad4e915f983",
+  "realm": "test",
+  "displayName": "",
+  "displayNameHtml": "",
+  "notBefore": 0,
+  "defaultSignatureAlgorithm": "RS256",
+  "revokeRefreshToken": false,
+  "refreshTokenMaxReuse": 0,
+  "accessTokenLifespan": 300,
+  "accessTokenLifespanForImplicitFlow": 900,
+  "ssoSessionIdleTimeout": 1800,
+  "ssoSessionMaxLifespan": 36000,
+  "ssoSessionIdleTimeoutRememberMe": 0,
+  "ssoSessionMaxLifespanRememberMe": 0,
+  "offlineSessionIdleTimeout": 2592000,
+  "offlineSessionMaxLifespanEnabled": false,
+  "offlineSessionMaxLifespan": 5184000,
+  "clientSessionIdleTimeout": 0,
+  "clientSessionMaxLifespan": 0,
+  "clientOfflineSessionIdleTimeout": 0,
+  "clientOfflineSessionMaxLifespan": 0,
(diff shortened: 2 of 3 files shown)


doc/release-notes/11689-builtin-users-api-bearer-auth-enhance.md
```diff
@@ -0,0 +1,9 @@
+## Security improvements for `api-bearer-auth-use-builtin-user-on-id-match`
+
+We’ve strengthened the security of the `api-bearer-auth-use-builtin-user-on-id-match` feature flag. It will now only work when the provided bearer token includes an `idp` claim that matches the Keycloak Service Provider identifier.
+
+By enforcing this check, the risk of impersonation from other identity providers is significantly reduced, since they would need to be explicitly configured with this specific, non-standard identifier.
```
qqmyers, line 5: Can this be part of the keycloak config, i.e. a static entry added there? Then we don't have to rely on the external provider. This is maybe less relevant for the internal users where you control the code, but it would be more under the site admin's control if there was a static value added to the token by keycloak (that was set for each login type).

GPortas: Yes. We could consider making the expected idp identifier configurable via a parameter in the SPI. Dataverse could then retrieve this value from a JVM option when verifying the token.

The main advantage I see in this approach is that it prevents the expected idp identifier claim value from being publicly exposed. This could indeed result in a more secure solution. That said, anyone who can successfully authenticate and obtain a token would still be able to decode it and see the idp claim value. Perhaps I’m missing other advantages of your approach.

qqmyers: I wasn't thinking about it being configurable - just using the Keycloak config to add a static value, e.g. 
image
Doing this, I don't think there would be any way for someone to configure a shib idp (or build a built-in account source like yours) that could allow impersonation.

ofahimIQSS:
Looks good to me - merging!

Issue #69, opened 2014-07-09
RESTful Search API


eaquigley:
---

Author Name: **Philip Durbin** (@pdurbin)
Original Redmine Issue: 3477, https://redmine.hmdc.harvard.edu/issues/3477
Original Date: 2014-02-03
Original Assignee: Philip Durbin

---

We have a beta RESTful search API in 4.0 but it should be finalized.

It returns JSON. Here's an example: http://dataverse-demo.iq.harvard.edu/api/search?q=tabulardata

It's also used in these integration tests: https://github.com/IQSS/dataverse/tree/master/scripts/search/tests

Requirements:

- use API keys: #3874
- enforce SSL (always?)
- provide a link for each item returned (file link will be for parent dataset)
- sorting
- start/offset (limit number returned?)

Questions:

- have showrelevance boolean?

Asked the community what they would like at https://groups.google.com/d/msg/dataverse-community/sXJmpoL84EU/gJmJhvRPm6YJ

---

Redmine related issue(s): [3430](https://redmine.hmdc.harvard.edu/issues/3430)

---

pdurbin:
As @scolapasta @eaquigley @rtreacy and I discussed, I'm taking this ticket over from @rtreacy .

pdurbin:
@kcondon I believe I've addressed almost all the concerns with the Search API from the meeting yesterday with you, @scolapasta and @eaquigley so I'm passing this to QA.

As we discussed, we need to QA the documentation as well, so please use http://guides.dataverse.org/en/latest/api/search.html for all the information you should need on using the Search API.

If you find any problems with how the code works, please let me know. I mentioned yesterday an issue I know about regarding datasets not showing descriptions. (This is why I said "almost all the concerns" above and you are welcome to open a ticket for this.)

If you don't like how the documentation reads, please give feedback to @eaquigley who has volunteered to improve the Search API docs in a separate ticket: #1328.

kcondon:
Thanks, I'll test it as I test all features and use the doc as functional requirements.


________________________________

kcondon:
Test all functions, public data only, payload contents for each type against cards, boolean operators.

Looks good, closing.

PR #11767, opened 2025-08-26
Extended io.gdcc.dataverse-spi (ExportDataProvider) interface

### LINKED ISSUE #11405: Refactor and optimize the Export metadata framework, especially for the data/variable-level metadata
Historically Dataverse has supported one format that encoded all the available DataVariable-level metadata from DataTable objects, the "full DDI". The reason our Exporters write the metadata into an OutputStream rather than return it as an object in memory was to accommodate potentially very large formats like the DDI. This export is still prohibitively expensive in terms of memory use since it receives the entire dataset-worth of variable-level metadata from `ExportDataProvider.getDatasetFileDetails()` as one big JsonArray. 

I would like to address this similarly to how we have added offset-length parameters to the /versions and /versions//files APIs for the SPA; and make it possible for `getDatasetFileDetails()` to page through the files/datatables in smaller batches. 

Once the DataExporter is refactored, the croissant exporter in the gdcc repo could be refactored as well. Just like the DDI it encodes the datavariable information. It's a little bit worse, because it doesn't stream its output either; it instead accumulates the entire json object in memory, then writes it all at once as `outputStream.write(job.build().toString().getBytes("UTF8"));`. But it should be very doable to make it stream instead. 

My secondary, less pressing concern is I want to have an option for the main `getDatasetJson()` method to skip the files info. (From what I can tell, `InternalExportDataProvier.getDatasetJson()` always calls `JsonPrinter.jsonWithCitation(DatasetVersion dsv, boolean includeFiles)` with `includeFiles=true`). If we are going to be exporting formats individually, this may be wasteful too, to pack thousands, or worse, of files into the json when exporting a cheap format like oai_dc that does not need them.

landreev:

I need additional mechanisms in the ExportDataProvide interface to implement the improvements in #11405 (primarily in the DDI exporter). This interface is maintained in the main source tree, but for the purposes of building Dataverse it is used as a standalone dependency package, io.gdcc.dataverse-spi. So, the changes to the interface need to be merged, and the new version of it published on maven central before I could make the final PR to close #11405.

- Closes 

I don't think there's a way to test this simple interface change by itself. This is an auxiliary PR that needs to be merged (and published) before I can make the main PR for #11405. Once that PR is made, the overall testing and QA will happen there. 

**Additional documentation**:

### DIFF
--- a/.github/workflows/spi_release.yml
+++ b/.github/workflows/spi_release.yml
@@ -42,8 +42,7 @@ jobs:
               with:
                   java-version: '17'
                   distribution: 'adopt'
-                  # TODO: change this from ossrh to central?
-                  server-id: ossrh
+                  server-id: central
                   server-username: MAVEN_USERNAME
                   server-password: MAVEN_PASSWORD
             - uses: actions/cache@v4
@@ -81,8 +80,7 @@ jobs:
                 with:
                     java-version: '17'
                     distribution: 'adopt'
-                    # TODO: change this from ossrh to central?
-                    server-id: ossrh
+                    server-id: central
                     server-username: MAVEN_USERNAME
                     server-password: MAVEN_PASSWORD
                     gpg-private-key: ${{ secrets.DATAVERSEBOT_GPG_KEY }}


pdurbin:
I made this pull request:

- #11774

It won't fix the `Dataverse SPI / Release Snapshot (pull_request)` check because I didn't even try editing the GitHub Action (update: I did leave some TODOs in f570aaaf41 we can experiment with). It does however explain how to publish a snapshot from a local env (which I did), so I hope it helps to keep this PR moving. I'm out for a few days so I'll put this back in "ready for review" so someone else can pick it up.

landreev:
@qqmyers @poikilotherm 
Jim and I were talking about this last week; but this may be a better place for the discussion. 
The motivation behind the PR is to add a few extra hooks to the ExportDataProvider interface in dataverse-spi. The extra goal is to implement it in a way that will not break any existing exporters, and will also allow us to keep adding more options going forward, similarly without breaking anything already implemented. 

The approach I am using currently: all the methods in ExportDataProvider have been made to accept an _optional_ ExportDataContext, as in 
```
JsonObject getDatasetJson(ExportDataContext... context);
```
In order to use any of the newly added options, an exporter will do something like 
```
datasetJson = exportDataProvider.getDatasetJson(ExportDataContext.context().withDatasetMetadataOnly()); 
or 
tabulardataJson = exportDataProvider.getTabularDataDetails(ExportDataContext.context().withOffset(offset).withLength(length));
```
but none of the already implemented exporters that are currently relying on ```getDatasetJson()``` etc. will need to be changed in any way. If/when we need more options in the future, they will be added to ExportDataContext, again, without breaking anything existing. 

I am very flexible on all this. I have working code in a separate branch that relies on this current implementation; but it should be trivial to change it in whatever way necessary if we end up implementing it differently. And I do want to invest into implementing it in a most optimal way.

I'm not super attached to the "context" concept above (my first implementation was via individual, multiple ExportDataOptions - ExportDataOption.java is still checked in for reference). 

Do we want the implementation to provide more information about the options currently supported? - This would potentially allow an exporter jar built using dataverse-spi version N to be able to both use some new options in that version, and to work with older Dataverses still using dataverse-spi prior to v. N; i.e., before requesting the `withPigLatin` option, it would be able to first ask whether it is supported by the provider. ... but is this something realistically useful/worth extra effort? 

Once again, I'm open to, and appreciate any feedback and suggestions.

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/ExportDataOption.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ *
+ * @author landreev
+ * Provides a mechanism for defining various data retrieval options for the 
+ * export subsystem in a way that should allow us adding support for more 
+ * options going forward with minimal or no changes to the existing code in 
+ * export plugins. 
+ */
+@Deprecated
+public class ExportDataOption {
```
stevenwinship, line 12: Looks like this is a new unused class. Instead of Deprecating why not just remove it?

stevenwinship: I just saw your comment (ExportDataOption.java is still checked in for reference). I don't think checking in unused code is good. Also, there are no tests included. Does this lower the code coverage? I'm not sure if code not under src/ is included in code coverage.

ofahimIQSS:
merging

ofahimIQSS:
https://github.com/IQSS/dataverse/actions/runs/18596172556 

Merge pull request #11767 from IQSS/11766-new-io.gdcc.dataverse-spi #69 failed

modules/dataverse-spi/src/main/java/io/gdcc/spi/export/ExportDataOption.java
landreev: @stevenwinship Thanks for reviewing this. Unfortunately, I missed your comments above yesterday. Yes, it would have been ideal to remove this before merging. I may make a quick followup pr removing it, before the interface jar is published on maven central. 

And yes, I only left it in place for the sake of showing the reviewers an alternative implementation I first tried.

landreev:
> https://github.com/IQSS/dataverse/actions/runs/18596172556
> 
> Merge pull request #11767 from IQSS/11766-new-io.gdcc.dataverse-spi #69 failed

OK, you didn't really mean #69, I see. I was royally confused by this at first. 
I'm assuming this is a known issue, that releases are still not being built automatically. (but we know how to make them manually). 
@pdurbin is this something that can be fixed easily by configuring a new secret for the action? Otherwise I may need your help with making the release. (I already got the credentials for making snapshot releases; but not "real" kind yet).

pdurbin:
@landreev sure, happy to help. Judging from https://repo1.maven.org/maven2/io/gdcc/dataverse-spi/ and this PR you want the next release to be 2.1.0.

I'm seeing this failure at https://github.com/IQSS/dataverse/actions/runs/18596172556/job/53023163304

`Error:  Failed to execute goal org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy (injected-nexus-deploy) on project dataverse-spi: Execution injected-nexus-deploy of goal org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy failed: Server credentials with ID "ossrh" not found! -> [Help 1]`

At minimum we should change these:

```
% ack TODO -A2 modules/dataverse-spi
modules/dataverse-spi/pom.xml
71:            <!--TODO: change this from ossrh to central?-->
72-            <id>ossrh</id>
73:            <!--TODO: change this url?-->
74-            <url>https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/</url>
75-        </repository>
--
115:                    <!--TODO: change this from ossrh to central?-->
116-                    <serverId>ossrh</serverId>
117:                    <!--TODO: change this URL?-->
118-                    <nexusUrl>https://s01.oss.sonatype.org</nexusUrl>
119-                    <autoReleaseAfterClose>true</autoReleaseAfterClose>```
```

I had to iterate quite a bit in the Croissant repo but we don't have many rules there. I force pushed to main several times. I can make a PR and we can merge it and then repeat until it's fixed, I guess. We might end up with a version higher than 2.1.0 though.

This module seems a bit different than Croissant where we have `-SNAPSHOT` in the version.

Anyway, let me make a PR with what I think we should do. Obviously, if @poikilotherm is around, he'll know better! 😅

pdurbin:
@landreev I made a PR:

- #11905

Please take a look! Happy to discuss!

PR #11779, opened 2025-08-29
IQSS/11776 indexing new files

### LINKED ISSUE #11776: New files in later versions not being indexed
Due to a mismatch in the return type from a query of file ids, the code added in #11374 silently fails to index files in new drafts added after v1 has been published.

* When does this issue occur?

* What happens?

develop

**Screenshots:**

*

PR in progress

qqmyers:

- Closes #11776

**Special notes for your reviewer**: This was tricky - the original code defined the query as returning Long file ids, but a) the database decided to send Integers anyway, and b) this then caused `changedFileMetadataIds.contains(fileMetadata.getId())` to fail even with a numeric match (Long(1) != Integer(1)).  That failure made it look like the file was the same in the released and draft versions, which we would not reindex (unless there was a variable-level metadata change).

The fix changes to return the file Ids as Integers - that should be a no-op, but may stop someone from assuming Long, and then checks the class returned from the query to assure Longs are getting into the changedFileMetadataIds array. I also added a warning if `changedFileMetadataIds.contains(fileMetadata.getId()) == false while the fileMap entry for the file is null (line 1531) which means the file wasn't in the released version (this shouldn't happen if the query and changedFileMetadataIds works as it should).

**Suggestions on how to test this**: There's an IT test - you should be able to add a dataset, publish, add a file, and verify you can't find the datafile via search or use the index/status api call and see that the file is in the db and not indexed. After the PR, this should be fixed (can recreate a new dataset or just reindex the existing one (by changing the dataset metadata for example).

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11776- index fix.md
+++ b/doc/release-notes/11776- index fix.md
@@ -1,0 +1,1 @@
+A bug, introduced in v6.7, that caused files in draft versions that were added after the initial dataset version was published, has been fixed.
--- a/src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
+++ b/src/main/java/edu/harvard/iq/dataverse/FileMetadata.java
@@ -94,9 +94,10 @@ import jakarta.validation.constraints.Pattern;
                 "        )",
                 resultSetMapping = "IdToLongMapping"
     )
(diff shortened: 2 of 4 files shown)


ofahimIQSS:
looks good from my end - merging

PR #11780, opened 2025-08-29
IQSS/11778 speedup custom file retrieval

### LINKED ISSUE #11778: Feature Request: Improve speed when retrieving custom files
The recent change in #11467 to dynamically decide the mimetypes of custom header, footer, analytics, logo, etc. files using Tika results in delays of ~80ms per file. At QDR, I observed a 1/4 second delay in page loading as we retrieved three files per page. As these files are controlled by admins and should have valid extensions, just using mime detection via our standard FileUtil method to get the mimetype from the extension should be sufficient. (It takes <<1 ms per call).

PR in progress

qqmyers:
**What this PR does / why we need it**: Stops using tika to determine the mime type for custom files - saving ~80ms per file access, which can occur for multiple files and for every page

- Closes #11778

**Special notes for your reviewer**: As noted in the issue, I'd expect using the extension is good enough when the files are supplied by an admin (as in this case). If we really want to use tika to inspect these files, we should cache the results.

**Suggestions on how to test this**: Not sure this is easy to test as there are other factors to page load speed. That said, adding a custom header, footer, analytics file, and logo might add enough delay (est 1/3 second) that you might see a difference in average load time. (At QDR, I added code to time the tika call and the FileUtil calls.)

**Is there a release notes update needed for this change?**: No - the PR that changed this was post 6.7 so the slower pages shouldn't be deployed in production anywhere.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/CustomizationFilesServlet.java
+++ b/src/main/java/edu/harvard/iq/dataverse/CustomizationFilesServlet.java
@@ -22,6 +22,7 @@ import jakarta.servlet.http.HttpServlet;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
+import edu.harvard.iq.dataverse.util.FileUtil;
 import jakarta.ejb.EJB;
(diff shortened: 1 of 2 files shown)


ofahimIQSS:
merging this, looks good from my end

PR #11781, opened 2025-08-29
IQSS/11777 improve MDC citation api scaling

### LINKED ISSUE #11777: Feature Request: Improve MDC Citation API scaling
The current /api/admin/makeDataCount/{id}/updateCitationsForDataset endpoint, which is often called in batch mode for all datasets in quick succession, does not handle queueing of requests, or throttling to match DataCite's rate limiting. It also reads the whole event report into memory before processing which is problematic for datasets with many files (as most of the report is the hasPart relationships we ignore in this call). For use with larger datasets, and use on larger instances, this should be improved.

QDR seeing site performance issues during the weekly cron job running the counter_weekly.sh script (which calls this api for all datasets).

PR in progress.

qqmyers:
**What this PR does / why we need it**: This PR makes the /api/admin/makeDataCount/{id}/updateCitationsForDataset call asynchronous, adds a queue for serially processing requests to it, adds an optional minimal delay between calls to the DataCite api triggered by these calls (to avoid hitting their rate limit), and improves memory use for datasets with many files. Together, these can help avoid performance issues and failures when periodically updating citations for many datasets (see the counter_weekly.sh script and MDC guides info).

- Closes #11777

**Special notes for your reviewer**: The PR adds an executor service configured with a queue size of 1000 which is probably sufficient for most installations. For Harvard, you'd definitely want this PR, but you'll also either need to increate the queue size (to the number of datasets you want to check) or modify the counter_weekly.sh script to make calls in smaller batches or watch for 503 responses and throttle sending new requests.

**Suggestions on how to test this**: This would be somewhat tricky as the result of running the api call is to update citations for datasets and test datasets probably won't have any. You can run the counter_weekly.sh script on an instance with many datasets (up to 1k), verify that you get OK responses in the counter_weekly.sh logging/no errors in the Dataverse log etc.

**Is there a release notes update needed for this change?**:added

**Additional documentation**: added, noted backward compatibility change as the call is now async.


stevenwinship:
@qqmyers Could you resolve the conflict. I'm not able to push the fix back.

ofahimIQSS:
looks good to me - merging

PR #11782, opened 2025-08-29
Improve DataCiteXML robustness w.r.t. ROR

qqmyers:
**What this PR does / why we need it**: In adapting to ROR v2 API, I hit a potential bug if the Ext. vocab configuration includes a whole array of names from ROR rather than a single one (as a JsonString) as with ROR v1. While this is something that can be handled by changing the configuration, it makes sense to have a check in the Dataverse code as well.

- Closes #

**Suggestions on how to test this**: Use the ORCID/ROR config in https://github.com/gdcc/dataverse-external-vocab-support/pull/46 and verify that you see a failure to export the DataCite XML when an author affiliation is used (e.g. "QDR"). With the fix, the DataCite xml should be generated and the ROR id included in it as the author affiliation (the org name will not be included).

**Is there a release notes update needed for this change?**: no? fixes an issue that would only be possible with the ROR v2 api

**Additional documentation**:


ofahimIQSS:
looks good from my end - merging

PR #11787, opened 2025-09-02
SignedUrl Fix when using http connector

qqmyers:
In proxy configurations where the SSL connection is terminated in the proxy and Dataverse is then contacted over http (e.g. using 
 `proxy_pass            http://localhost:8080/;`
in nginx, signed URLs currently fail because they are signed as https://... urls but Dataverse receives a call to the http://... form of the URL (hence the url strings being signed/verified differ by one 's' character.

This has also been seen when using Anubis and it's forwarding to Dataverse.

This PR adds support for Dataverse to use the value of the X-Forwarded-Proto HTTP header and, if that is 'https', to restore the original https form of the URL before trying to test the signature. The result is that signatures can work in such configurations.

- Closes #

Should be able to configure Anubis (as on demo) or adjust an nginx or apache proxy to forward to Dataverse over http - verify that signed URLs for external tools fail if the X-Forwarded-Proto Header is missing or (easier to test) is set incorrectly to http. Turning on the fine logging for the SignedUrlAuthMechanism class will show when the signed url is received starting with http and the X-Forwarded-Proto header is set to https allowing the signed url to be validated.

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -94,6 +94,16 @@ First of all, confirm that access is denied! If you are in fact able to access t
 
 Still feel like activating this option in your configuration? - Have fun and be safe!
 
+.. _signed-urls-forwarded-proto-header:
+
+Using X-Forwarded-Proto for Signed Urls
++++++++++++++++++++++++++++++++++++++++
+
+If you use an Apache or Nginx proxy, or have a firewall such as Anubis, and they are configured to forward traffic to Dataverse over http
+(i.e. your proxy receives user calls over https but forwards locally to Dataverse over http), signed urls, used by external tools and 
+upload apps (such as DVWebloader), are likely to fail unless you configure your proxy to send an X-ForwardedProto HTTP Header.
+This allows Dataverse to recognize that the communication from the user was over https and that validation of signed urls should assume 
+they started with https:// (rather than http:// as received from the proxy).
 
 .. _PrivacyConsiderations:
 
--- a/src/main/java/edu/harvard/iq/dataverse/api/auth/SignedUrlAuthMechanism.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/auth/SignedUrlAuthMechanism.java
@@ -16,6 +16,7 @@ import jakarta.ws.rs.core.UriInfo;
 
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.util.logging.Logger;
 
 import static edu.harvard.iq.dataverse.util.UrlSignerUtil.SIGNED_URL_TOKEN;
 import static edu.harvard.iq.dataverse.util.UrlSignerUtil.SIGNED_URL_USER;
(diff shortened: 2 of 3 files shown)


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -94,6 +94,16 @@ First of all, confirm that access is denied! If you are in fact able to access t
...
 Still feel like activating this option in your configuration? - Have fun and be safe!
 
+.. _signed-urls-forwarded-proto-header:
+
+Using X-Forwarded-Proto for Signed Urls
++++++++++++++++++++++++++++++++++++++++
+
+If you use an Apache or Nginx proxy, or have a firewall such as Anubis, and they are configured to forward traffic to Dataverse over http
+(i.e. your proxy receives user calls over https but forwards locally to Dataverse over http), signed urls, used by external tools and 
+upload apps (such as DVWebloader), are likely to fail unless you configure your proxy to send an X-ForwardedProto HTTP Header.
```
minorg: X-Forwarded-Proto

src/main/java/edu/harvard/iq/dataverse/api/auth/SignedUrlAuthMechanism.java
```diff
@@ -72,6 +75,18 @@ private User getAuthenticatedUserFromSignedUrl(ContainerRequestContext container
         }
         if (targetUser != null && userApiToken != null) {
             String signedUrl = URLDecoder.decode(uriInfo.getRequestUri().toString(), StandardCharsets.UTF_8);
```
minorg, line 77: IIRC the servlet container preserves the (client's) case in getRequestUri() so this might not startWith("http://") unless it's lowered.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -94,6 +94,16 @@ First of all, confirm that access is denied! If you are in fact able to access t
 
 Still feel like activating this option in your configuration? - Have fun and be safe!
 
+.. _signed-urls-forwarded-proto-header:
+
+Using X-Forwarded-Proto for Signed Urls
```
```suggestion
Using X-Forwarded-Proto for Signed URLs
```

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -94,6 +94,16 @@ First of all, confirm that access is denied! If you are in fact able to access t
...
++++++++++++++++++++++++++++++++++++++++
+
+If you use an Apache or Nginx proxy, or have a firewall such as Anubis, and they are configured to forward traffic to Dataverse over http
+(i.e. your proxy receives user calls over https but forwards locally to Dataverse over http), signed urls, used by external tools and 
+upload apps (such as DVWebloader), are likely to fail unless you configure your proxy to send an X-Forwarded-Proto HTTP Header.
+This allows Dataverse to recognize that the communication from the user was over https and that validation of signed urls should assume 
+they started with https:// (rather than http:// as received from the proxy).
```
```suggestion
If you use a proxy such as Apache or Nginx, or have a firewall such as Anubis, and they are configured to forward traffic to Dataverse over HTTP
(i.e. your proxy receives user calls over HTTPS but forwards locally to Dataverse over HTTP), signed URLs, used by external tools and 
upload apps (such as DVWebloader), are likely to fail unless you configure your proxy to send an X-Forwarded-Proto HTTP Header.
This allows Dataverse to recognize that the communication from the user was over HTTPS and that validation of signed URLs should assume 
they started with https:// (rather than http:// as received from the proxy).
```

landreev:
Tested on qa, with a restricted image file in an unpublished dataset: https://qa.dataverse.org/file.xhtml?datasetVersionId=605066&fileId=13257733. 
Earlier, we got previewers to work there by adding ProxyPassMatch rules sending /api/access/datafile and other APIs directly to Payara, bypassing Anubis. Confirmed that with stock 6.9 and these lines commented out in ssl.conf image previewer was no longer working, even when logged in as the superuser. "Bad signed url" error message. 
With this build, signed urls are working again, even when sent through Anubis.

Issue #46, opened 2014-07-09
Relevance of search results


eaquigley:
---

Author Name: **Elizabeth Quigley** (@eaquigley)
Original Redmine Issue: 3453, https://redmine.hmdc.harvard.edu/issues/3453

Original Assignee: Elda Sotiri

---
# Relevance

tl;dr: Should this return any results? http://dvn-build.hmdc.harvard.edu/search.xhtml?q=data+thisTermDoesNotAppear

This stems from previous discussions on what should be highlighted in the search results and what in the result was hit by the search term. During usability testing, users mentioned they were slightly confused as to why some results appeared based on their search. Am attaching Mike's search results card sketch from before.

During usability testing, a user attempted to narrow a search by adding an additional term. Adding the term did put the dataset she was thinking of at the top of her search results _but_ she received _more_ search results, not fewer. Internally, we've noticed this as well... if you add "thisTermDoesNotAppear" to the end of a search query that does match something (i.e. a search for "data"), you still get results. On a Trello card we were calling this the "something rather than nothing" feature: https://trello.com/c/MAr4d5Ow/17-prototype-search-browse

Stop words are a potential issue. Should they be stripped?

If a dataverse is called "Friday" (Friday Dataverse) and a search for "friday" is done, should it match everything in the "Friday" dataverse, including datasets and files? What about child dataverses? Should all datasets and files in the child dataverse match as well?

---
# Highlighing

As of this writing, highlighting only occurs on the "description" field. We probably want highlighting on more fields.

We've received feedback that the yellow highlighting is too yellow.

3/26/2014 note: This is still something that needs to be worked out. Which fields can a user search on that would pull a result? How do we show those fields in the search results card?

4/4/2014 note: The more I think about this, I think highlighting should only be appearing in the description field. If the search query hits the title or the citation, I don't think there is a need for highlighting. Additionally, since the citation is in a yellow box already, I don't think yellow highlighting on a yellow background would be that appealing to look at. This brings me to another thought, please see the section: Citation. (EQ)

4/14/2014 note: Highlighting has its own ticket at #3737 and has been moved to QA.

---
# Citation

If a user's search query hits a field within the citation, do we need a visual cue to show that is what caused the result to appear? If it's author that is the field hit, then wouldn't the author facet also be showing this? If so, then that would be three places (facet, search query in bar, and in the citation) so do we need an additional thing showing the user the relationship between the result and the query? Something to test. (EQ)

[...]

PR #11793, opened 2025-09-05
CVoc retrieval filtering enhancement

qqmyers:
Dataverse's external vocabulary mechanism includes an option for Dataverse to retrieve and cache metadata about a given vocabulary/PID indentifier by calling an external service (e.g. ORCID, ROR, SkosMos, OntoPortal) and filtering the returned JSON results to get the desired subset of information. As the range of examples has expanded, the syntax for specifying that filtering is being expanded. For example, #11762 allows retrieving the nth entry in a JSON array rather than the whole array. 

This PR extends a related functionality which allows retrieving a JSON object from an array based on a matching key:value pair, i.e. one can retrieve the object with the key:value "mykey":"myval" by specifying `/mykey=myval` as part of the path in the filter "params" entry.  Currently this functionality only works if the value of the key is a JSON String. However, with the ROR v2 API, we have a case where we'd like to retrieve the object which has the value "ror_display" for the "types" key where "ror_display" is a string within a JSON array of values. (The entry with the "ror_display" is the organization name that is used for display on the ROR page and is nominally the canonical name for the organization.

This PR modifies the parsing in Dataverse to support this case. As the ROR v2 updates are included in the gdcc/dataverse-external-vocab-support repo, examples and further description of the retrieval-filtering syntax will be added there.

- Closes #

**Special notes for your reviewer**: Nominally we want ROR v2 support to work well before v1 is retired in Dec. 2025.

**Suggestions on how to test this**: Configure a field to use the ROR v2 script (as in [#46](https://github.com/gdcc/dataverse-external-vocab-support/pull/46) and change the termName/params entry to be /names/types=ror_display/value (instead of /names) and verify that the OAI_ORE export includes the name of the organization (e.g. for author affiliation) and that the DataCite xml export also includes it.

**Is there a release notes update needed for this change?**: I'll add a generic one to say that this version extends the retrieval-filtering - to cover both this and #11762.

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/DatasetFieldServiceBeanTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/DatasetFieldServiceBeanTest.java
@@ -1,6 +1,9 @@
 package edu.harvard.iq.dataverse;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
@@ -17,7 +20,10 @@ import org.mockito.Mockito;
 
 import edu.harvard.iq.dataverse.settings.SettingsServiceBean;
 import jakarta.json.Json;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonArrayBuilder;
 import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
 
 public class DatasetFieldServiceBeanTest {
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetFieldServiceBean.java
```diff
@@ -764,13 +764,24 @@ Object processPathSegment(int index, String[] pathParts, JsonValue curPath, Stri
...
+                                        //Match a string value
                                         if (((JsonString) val).getString().equals(expected)) {
                                             logger.fine("Found: " + jo);
                                             curPath = jo;
                                             return processPathSegment(index + 1, pathParts, curPath, termUri);
                                         }
-                                    } else {
-                                        logger.warning("Expected a string value for " + keyVal[0] + " but found: " + val.getValueType());
+                                    } else if (val.getValueType() == JsonValue.ValueType.ARRAY) {
+                                        // Match one string in an array
```
stevenwinship, line 774: Could you add a test in DatasetFieldServiceBeanTest or an API test to exercise this code?

ofahimIQSS:
very nice, everything checks out - merging

PR #11794, opened 2025-09-05
show draft/unpublished in UI

### LINKED ISSUE #7618: File level permissions: restricted files in "draft format of a dataset" should not show up when assigning permissions to "files found in the published version of a dataset"
Tested here:
https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/E08MOR

When you have a published dataset with restricted files, and you create a draft of the dataset with additional new restricted files, and you are assigning file level permission to the published version, the files in the draft dataset are showing up on the list of restricted files in the permissions page. There is no indication to the admin that the "draft dataset files" have yet to be published. This can cause admins to accidentally give access to files not intended for use. Draft file should not even show up here. 

tested:
1. created a dataset with 2 restricted files in demo. published. 
2. created a draft version with one additional restricted file, left in draft format.
3. navigate to the published dataset, go to "permissions, files," and when select "grant access to users," there are 3 restricted files and not just the 2 expected from step 1. The additional file does not indicate it exists in a DRAFT dataset.

stevenwinship:
**What this PR does / why we need it**: When you have a published dataset with restricted files, and you create a draft of the dataset with additional new restricted files, and you are assigning file level permission to the published version, the files in the draft dataset are showing up on the list of restricted files in the permissions page. There is no indication to the admin that the "draft dataset files" have yet to be published. This can cause admins to accidentally give access to files not intended for use. Draft file should not even show up here.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/7618

- Closes #7618

**Suggestions on how to test this**: Follow the steps in the main issue

image

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DataFile.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataFile.java
@@ -303,7 +303,18 @@ public class DataFile extends DvObject implements Comparable {
     public void setDeleted(Boolean deleted) {
         this.deleted = deleted;
     }
-    
+
+    @Transient
+    private boolean unpublished;
+
+    public boolean getUnpublished() {
+        return unpublished;
+    }
+
+    public void setUnpublished(boolean unpublished) {
+        this.unpublished = unpublished;
+    }
+
     /*
     For use during file upload so that the user may delete 
     files that have already been uploaded to the current dataset version
@@ -596,7 +607,10 @@ public class DataFile extends DvObject implements Comparable {
         FileMetadata resultFileMetadata = null;
 
         if (fileMetadatas.size() == 1) {
+            setUnpublished(fileMetadatas.get(0).getDatasetVersion() == null || fileMetadatas.get(0).getDatasetVersion().getVersionState().equals(VersionState.DRAFT));
             return fileMetadatas.get(0);
+        } else {
+            setUnpublished(false); // Since only one can be in Draft assume there is a published version
         }
 
         for (FileMetadata fileMetadata : fileMetadatas) {
--- a/src/main/webapp/permissions-manage-files.xhtml
+++ b/src/main/webapp/permissions-manage-files.xhtml
@@ -335,6 +335,7 @@
                                         <h:outputText rendered="#{!empty file.directoryLabel}" value="#{file.directoryLabel}/" styleClass="text-muted"/>
                                         <h:outputText value="#{file.displayName}"/>
                                         <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.deleted']}) " rendered="#{file.deleted}"/>
+                                        <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.draftUnpublished']}) "  rendered="#{file.unpublished}"/>
                                     </p:column>
                                 </p:dataTable>
                             </div>
--- a/doc/release-notes/7618-file-level-permissions-restricted-draft.md
+++ b/doc/release-notes/7618-file-level-permissions-restricted-draft.md
@@ -1,0 +1,3 @@
+File level permissions: Restricted files in Draft will now show a "Draft/Unpublished" tag in the UI when granting file access
+
+See #7618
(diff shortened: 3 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/DataFile.java
```diff
@@ -596,7 +607,10 @@ public FileMetadata getLatestFileMetadata() {
         FileMetadata resultFileMetadata = null;
 
         if (fileMetadatas.size() == 1) {
+            setUnpublished(fileMetadatas.get(0).getDatasetVersion() == null || fileMetadatas.get(0).getDatasetVersion().getVersionState().equals(VersionState.DRAFT));
             return fileMetadatas.get(0);
+        } else {
+            setUnpublished(false); // Since only one can be in Draft assume there is a published version
```
qqmyers: Is deaccessioned an issue, or are those files already excluded from the list?

src/main/webapp/permissions-manage-files.xhtml
```diff
@@ -335,6 +335,7 @@
                                         <h:outputText rendered="#{!empty file.directoryLabel}" value="#{file.directoryLabel}/" styleClass="text-muted"/>
                                         <h:outputText value="#{file.displayName}"/>
                                         <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.deleted']}) " rendered="#{file.deleted}"/>
+                                        <h:outputText value=" (#{bundle['dataverse.permissionsFiles.files.draftUnpublished']}) "  rendered="#{file.unpublished}"/>
```
qqmyers: How about using file.released (the DvObject.isReleased() method). That is checking the publication date rather than looking for only a draft filemetadata, but I think it should work (and avoid the other changes).

src/main/java/edu/harvard/iq/dataverse/DataFile.java
```diff
@@ -303,7 +303,18 @@ public Boolean getDeleted() {
     public void setDeleted(Boolean deleted) {
         this.deleted = deleted;
     }
-    
+
+    @Transient
+    private boolean unpublished;
+
+    public boolean getUnpublished() {
```
qqmyers: If this is needed (see other comment), can this be a Boolean with the logic to calculate it added here and done only once (if unpublished == null)?

src/main/java/edu/harvard/iq/dataverse/DataFile.java
stevenwinship: They are excluded

src/main/webapp/permissions-manage-files.xhtml
stevenwinship: changed unpublished to return !isReleased

src/main/java/edu/harvard/iq/dataverse/DataFile.java
stevenwinship: changed unpublished to return !isReleased

src/main/webapp/permissions-manage-files.xhtml
qqmyers: Why not just !file.released here and no change in DataFile.java?

src/main/java/edu/harvard/iq/dataverse/DataFile.java
qqmyers: FWIW: I wasn't concerned about files in a deaccessioned version showing but what would happen for a file in the draft version when it was only previously in a deaccessioned version. It would have two filemetadatas and thus not show as draft. With the other code changes, I guess the equivalent question is whether deaccessioning the only version(s) a file was published in does/should set it's publication date to null. Probably a minor issue in any case and we should consider changing deaccession behavior here out of scope.

src/main/webapp/permissions-manage-files.xhtml
qqmyers: @stevenwinship - this is the only active suggestion - just further simplification.

stevenwinship: changed

qqmyers (approved):
Looks good.

PR #11795, opened 2025-09-08
Refer to configured datastore id when migrating from local datastore to S3

mjlassila:

This pull request is related to topic [#large-data > Filesystem -&gt; S3 migration documentation still valid?](https://dataverse.zulipchat.com/#narrow/channel/432390-large-data/topic/Filesystem.20-.3E.20S3.20migration.20documentation.20still.20valid.3F/with/537695359) in the chat.

Since Dataverse supports multiple datastores, including multiple S3-stores with user defined datastore ids, it is more clear if in example update queries do not use string `S3` as a placeholder for datastore id.


donsizemore:
@mjlassila looks good!

PR #11799, opened 2025-09-08
Expands InternalExportDataProvider per dataverse-spi 2.1.0 and refactors the DDI exporter to utilize the new functionality

### LINKED ISSUE #11405: Refactor and optimize the Export metadata framework, especially for the data/variable-level metadata

landreev:
**Un-drafting the PR now!** 

See issue #11405. Short version: the current export data framework is prohibitively expensive for datasets with very large amounts of tabular data (i.e., "ingested" DataFiles that have associated DataTable entities , with large numbers of child DataVariable objects). This PR improves it, via adding code that retrieves the data in manageable chunks. This in turn relies on the new version of the `ExportDataProvider` interface from https://github.com/gdcc/dataverse-spi. 

- Closes #11405

Note that the PR builds with a _snapshot_ release of [dataverse-spi](https://github.com/gdcc/dataverse-spi):
```
            <groupId>io.gdcc</groupId>
            <artifactId>dataverse-spi</artifactId>
            <version>2.1.0-SNAPSHOT</version>
```
I.e., the version 2.1.0 still needs to be properly released under gdcc. But I don't think that should be stopping putting the PR into review and merging it rather sooner than later. 

Note that this PR also removes `modules/dataverse-spi`, and the action that used to build snapshot releases of it (`.github/workflows/spi_release.yml`) from the main source tree; now that dataverse-spi lives under gdcc. 

Note specifically the test for the functionality, `TabularDataExportIT` added in this PR. ~~I have added it to the list of integration tests that are performed automatically by the workflow action (in `tests/integration-tests.txt`)~~ even though, strictly speaking, it is not a real integration test. [update: No, it is not executed as an "integration test". It is run by the GitHub action via [this change](https://github.com/IQSS/dataverse/pull/11799/changes#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R23) in the main pom.xml] I.e., it does not rely on calling a fully-functional Dataverse instance started remotely. It relies instead on the new test framework added in @poikilotherm's #12365 where a _local_ Docker instance of the database is spun up in which some dummy objects - Datasets, Versions, Datatables and Variables etc are created instantly, allowing testing of complex functionality that requires real database access. Without this new framework, the only way to test an export of a dataset with N large tabular files would be to actually ingest these N files via the API in the remote Dataverse instance used by the RestAssured tests. Which would add to the ongoing bloating of the test suite, and still would not allow to test the underlying functionality as thoroughly (for example, there is no easy way to count the actual database queries under such scenario).  

You can run the test in question in your own dev. environment via 
```mvn test -Dtest=TabularDataExportIT#exportTabularMetadata```
If you do not have Docker running, the test will be politely skipped. Otherwise it will create a throwaway postgres instance and proceed to test tabular data exports and associated data retrieval, comparing the results produced using the legacy methods vs. the new and optimized versions added in this PR and more. 

Please see the info under the "notes for the reviewer" and feel free to play and experiment with the new test described there. Please keep in mind that the test is still somewhat of a reference demonstrator of the new `JpaEntityManagerService`-based framework from #12365. It can be expanded and improved going forward.   

The ultimate test would be to test before vs. after on some real-life monster datasets. 
I would suggest these bad boys on qa: 
https://qa.dataverse.org/dataset.xhtml?persistentId=doi:10.7910/DVN/KZW8R2 (207 ingested tab. files; 186906 variables total)
https://qa.dataverse.org/dataset.xhtml?persistentId=doi:10.7910/DVN/FAEZIO (877 tab. files; 316983 variables)

"ddi" (aka the "full" ddi) is the export format that needs to be tested. 
Note that some extra steps will be required in order to test this on qa: 
- You will want to change the storage driver to local filesystem ("file") for the datasets in question - so that Dataverse can write the export files; 
Happy to assist with the above if needed. 

[Edit: additional info] The best way to test is by running an asynchronous export using the `reExportDataset` API. That tests 2 things at once: the new data export framework and its use by the DDI exporter; and the new behavior of the API that allows to re-export only the format(s) specified. For example: 
```curl "http://localhost:8080/api/admin/metadata/:persistentId/reExportDataset?persistentId=doi:10.7910/DVN/KZW8R2&formats=ddi";```
... as the export is performed in the background, its progress can be observed by checking the growing cache - for ex., `/usr/local/dvn/data/10.7910/DVN/KZW8R2/export_ddi.cached` and the server.log. (logging level is set to FINE for the ddi export util). 

The DDI export for the 2 datasets above IS working under stock 6.11; at least when it's not busy doing anything else. It takes a long time, and it uses an unhealthy amount of memory. It took 14 min. and 21 min. respectively. The total size of the DDI produced: 107MB and 190MB respectively. 
It is NOT going to be lightning fast under the new code either. The application performs roughly the same amount of work, retrieving the datavariable information from the database, etc. Some time savings can be observed because the heap stays under control. But the most important improvement is not the speed - we are ok with letting it run for however long it needs to in the background - but the fact that memory is only allocated in manageable chunks. 

Any other dataset can be tested of course. It needs to be reconfigured to use the local filesystem storage, by changing the `storageidentifier` to `file:10.7910/DVN/XXXXXX` in the corresponding `dvobject`. (this has been done for the datasets above). 

**Additional documentation**:

### CODE REVIEW

**Comment by landreev** [comment]
Note that the PR now builds under Jenkins (with a snapshot release of dataverse-spi that is now published).

File: src/test/java/edu/harvard/iq/dataverse/export/TabularDataExportIT.java [hunk 1]
```diff
@@ -0,0 +1,359 @@
...
+                    } else if (xmlr.getLocalName().equals("dataDscr")) {
+                        dataDscrComplete = true;
+                    }
+                }
+            }
+        } catch (XMLStreamException xse) {
+            assertTrue(false, "Unexpected XMLStreamException when attempting to parse the dataDscr section: " + xse.getMessage());
+        } finally {
+            if (xmlr != null) {
+                try {xmlr.close();} catch (XMLStreamException e) {}
```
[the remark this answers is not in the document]
landreev: Don't tell me what to do, reviewdog. That was formatted like that on purpose, to emphasize the fact that we don't care about that particular exception at that stage of the test.

landreev:
@poikilotherm Hi, I'm looking at the test results above, and seeing this failure to "Release Snapshot" for Dataverse SPI: 

Screenshot 2026-07-20 at 1 55 13 PM

Am I correct assuming this is an old action that simply needs to be removed, now that dataverse-spi lives in https://github.com/gdcc/dataverse-spi? 
Similarly, should I go ahead and purge `modules/dataverse-spi` from the main source tree? 
Please stop me if there is any reasons why I should not delete these. 
Thank you!

poikilotherm:
@landreev  Yes, the in-tree SPI stuff can be completely eradicated. ☢️ :smile:

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
```diff
@@ -1465,25 +1521,32 @@ private static void createOtherMats(XMLStreamWriter xmlw, List<FileDTO> fileDtos
...
     // -- L.A. 4.5 
-    
+    // [update:] Since the comment above was written, the method below was changed
+    // to operate on a JsonArray, as provided by the FileDetails method in the 
+    // ExportDataProvider. However, As of 6.8, this method is no longer used at 
+    // all. This is because the DTOs supplied by ExportDataProvider.getDatasetJson() 
+    // DO in fact contain enough information to generate the otherMat sections 
+    // properly, whether this is a short or a full version of the DDI. I am however leaving 
+    // this method here for reference. 
+
```
poikilotherm: As per your comment, IMO this method should be tagged @Deprecated to indicate its imminent removal.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
```diff
@@ -1608,44 +1673,142 @@ public static void createDataDscr(XMLStreamWriter xmlw, JsonArray fileDetails) t
...
+    /**
+     * 
+     * This public version of the method exists solely so that it can be called 
+     * with a custom batch size; primarily for tests.
+     * @param xmlw XML stream writer
+     * @varQuantityMap mapping of ordered datatables -> number of variables in each
+     * @exportDataProvider data provider instance
+     * @variablesBatchSize number of variables to use in forming processing batches
+    */
```
poikilotherm, line 1648: The integration test should be in the same package as the class (it already is in ...export, so it should move to ...export.ddi, the package this class is a part of). Then the method can be made package private and is accessible to the test right away without exposing it.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
```diff
@@ -1608,44 +1673,142 @@ public static void createDataDscr(XMLStreamWriter xmlw, JsonArray fileDetails) t
...
+                            .addFilePredicate(FileMetadataPredicates.ONLY_PUBLIC_FILES)
+                            .addFilePredicate(FileMetadataPredicates.ONLY_TABULAR_FILES)
+                            .addFilePredicate(FileMetadataPredicates.INCLUDE_TABULAR_DATA_VARIABLES)
+                            .build();
+                    PageRequest paginationRequest = PageRequest.of(dataTableStart, dataTablesThisBatch);
+                    Stream<JsonObject> tabularFileDetails = exportDataProvider.getDatasetFileDetails(exportQuery, paginationRequest);
+                    logger.fine("total number of variables in this batch: " + varQuantityThisBatch);
+
+                    int count = 0;
+                    //for (int i = 0; i < tabularFileDetails.size(); i++) {
```
poikilotherm: Is this a comment to explain the iterator pattern? If so, please use human language and remove the code block ;-)

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -319,16 +319,72 @@ public void exportAllFormats(Dataset dataset) throws ExportException {
         }
 
     }
+    
+    public void exportFormats(Dataset dataset, List<String> formatNames) throws ExportException {
```
poikilotherm, line 299: This method is missing Javadocs explaining scope, reasoning and behavior.

Also:

1. I encourage you to use an empty list instead of a null value, which IMO is much more Javaish.
2. There is no null check on the dataset parameter.
3. Why not use DatasetVersion and split the logic in two methods? The getExport() method uses a DatasetVersion, thus IMO using a DatasetVersion + list of formats is more aligned. There can be a second method with Dataset + list of formats for convenience, that handles the business logic of which version may be exported (calling the one with version + formats internally).

For Johannes I will be adding a method with cardinality in the other direction: one format, multiple versions. Eventually, we will refactor the service to support exporting arbitrary versions, where it also makes sense to have version + formats.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+     * Only one context object is supported
+     * @param DatasetExportQuery
+     * @return
+     */
+    private boolean isOnlyDatasetLevelMetadataRequested(DatasetExportQuery query) {
+
+        Set<DatasetMetadataPredicates> predicates = query.getDatasetPredicates();
+
+        for (DatasetMetadataPredicates p : predicates) {
+            // @todo This is pending on adding a dedicated DATASET_LEVEL_ONLY predicate
```
poikilotherm: No. The DatasetExportQuery has predicates only for the dataset's own metadata:

> Predicates for controlling which dataset metadata is included in an export and at what level of detail.

The dataset export query may have a distinct `FileExportQuery` containing detailed information as its own predicates about the file metadata. At the moment, there is only one case where this method may return `false`: 

1. There is a file export query within the dataset export query and it has a predicate "SKIP_FILES". (Otherwise it's either all files, public and/or tabular only).

We might need to tighten the definition in the export query spec, here's what it says now:

> File metadata shaping is optional: if no FileExportQuery is provided, methods that include file metadata will apply their own defaults. Methods that do not return file metadata will ignore any nested FileExportQuery.

At time of writing, this seemed fine. But maybe it isn't? Should we do an explicit default? Do I remember correctly that in the past `getDatasetJson()` did *not* return file level metadata? (Then we should keep this default for backward compatibility, changing it would require a breaking change - and we shouldn't do those at this point.)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+    }
+
+    /**
+     * Service method for checking a FileExportQuery for a specific predicate specified.
+     *
+     * @param query
+     * @param predicate
+     * @return
+     */
+    private boolean checkForPredicate(FileExportQuery query, FileMetadataPredicates predicate) {
```
poikilotherm: 1) Why a for loop? This is a set, you can just ask for `.contains()`?
2) Would it help if the FileExportQuery (and dataset query) has a helper function for this builtin? This way you can ask the query directly and maybe don't need all these small util functions?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+        }
+
+        if (isOnlyTabularMetadataRequested(query) && isDataVariableMetadataRequested(query)) {
+
+            fileMetadatas = datasetVersionFilesService.getTabularDataFileMetadatas(dv,
+                    pageRequest.getLimit(),
+                    pageRequest.getOffset(),
+                    isOnlyPublicMetadataRequested(query));
+
+            for (FileMetadata fileMetadata : fileMetadatas) {
```
poikilotherm: No. Please don't. This is not the way the whole stream business was intended to go.

Instead of doing a roundtrip with a JsonArray, we can stream from the retrieved `List` directly!

Something like this:

```java
return fileMetadatas.stream()
    .map(fileMetadata -> JsonPrinter.jsonDatafileWithDatatableForExport(fileMetadata.getDataFile(), fileMetadata))
    .map(JsonObjectBuilder::build);
```

This will defer the creation of the JsonObjects until the exporter actually consumes them one by one from the stream.

Also, this opens up the possibility to (at a later point) use a streaming result from JPA. We do explicit pagination here (fine), but for the method without explicit pagination it will make things more efficient if JPA handles paged retrieval as streams.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+     * getDatasetFileDetails();
+     *
+     */
+    public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query, PageRequest pageRequest) {
         JsonArrayBuilder jab = Json.createArrayBuilder();
-        for (FileMetadata fileMetadata : dv.getFileMetadatas()) {
-            DataFile dataFile = fileMetadata.getDataFile();
-            jab.add(JsonPrinter.json(dataFile, fileMetadata, true));
+
+        List<FileMetadata> fileMetadatas;
```
poikilotherm: Please don't use C-style variable definition for a list that is used in exactly one place. It could even be inlined completely. Thx!

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -11,45 +12,78 @@
...
 /**
  * Provides all data necessary to create an export
- * 
+ *
  */
 public class InternalExportDataProvider implements ExportDataProvider {
 
-    private DatasetVersion dv;
+    private final DatasetVersion dv;
     private JsonObject jsonRepresentation = null;
```
poikilotherm, line 46: None of these lazy-loaded, cached entities are thread safe. This may be a problem if the same provider is handed to multiple exporters in different threads. IMHO it's a huge design smell.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -11,45 +12,78 @@
...
+    private JsonObject jsonRepresentationNoFiles = null;
     private JsonObject schemaDotOrgRepresentation = null;
     private JsonObject oreRepresentation = null;
+    private JsonArray fileAndDataDetails = null;
```
poikilotherm, line 50: fileAndDataDetails is shared between two semantically different methods. The non-paginated getDatasetFileDetails() and the streaming getDatasetFileDetails(FileExportQuery query) both read/write the same cache field. 

But they conceptually might return different views (one may be filtered by the query). Caching them together means the first call wins and the query in the second call is silently ignored if the cache is already populated.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+        }
+
+        return false;
+    }
+
+    /**
+     * This method is exclusively for use in IT tests
+     * @param versionFilesService
+     */
+    public void injectVersionFilesService(DatasetVersionFilesServiceBean versionFilesService) {
```
poikilotherm: Any such service should always be injected by explicit constructor parameter, not a method. Hand-rolled dependency injection is prone to creating all sorts of fancy problems.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+                    pageRequest.getLimit(),
+                    pageRequest.getOffset(),
+                    isOnlyPublicMetadataRequested(query));
+
+            for (FileMetadata fileMetadata : fileMetadatas) {
+                DataFile dataFile = fileMetadata.getDataFile();
+                jab.add(JsonPrinter.jsonDatafileWithDatatableForExport(dataFile, fileMetadata));
+            }
+
+            return jab.build().stream().map(jsonValue -> jsonValue.asJsonObject());
```
poikilotherm: This lambda at least should be a method reference. (But it'll be gone anyway when streaming the result list directly.)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+     *
+     */
+    public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query, PageRequest pageRequest) {
         JsonArrayBuilder jab = Json.createArrayBuilder();
-        for (FileMetadata fileMetadata : dv.getFileMetadatas()) {
-            DataFile dataFile = fileMetadata.getDataFile();
-            jab.add(JsonPrinter.json(dataFile, fileMetadata, true));
+
+        List<FileMetadata> fileMetadatas;
+        if (datasetVersionFilesService == null) {
```
poikilotherm, line 189: This whole block is a huge design smell. If the hand rolled injection failed, we're now using CDI from inside a POJO being moved around and not at all being instantiated in a container context to lookup an EJB. Please don't. The proven best practice to handle direct dependency injection into POJOs is via constructor argument!

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+            try {
+                datasetVersionFilesService = CDI.current().select(DatasetVersionFilesServiceBean.class).get();
+            } catch (java.lang.IllegalArgumentException | IllegalStateException ie) {
+                throw new ExportException("EJB DatasetVersionFilesService is not available; " + ie.getMessage());
+            }
+        }
+
+        if (datasetVersionFilesService == null) {
```
poikilotherm, line 197: Using a second null check as an additional safeguard is superfluous and unnecessary once best practices are applied.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+            throw new ExportException("EJB DatasetVersionFilesService is not available");
+        }
+
+        if (isOnlyTabularMetadataRequested(query) && isDataVariableMetadataRequested(query)) {
```
poikilotherm: Why this artifical limitation to these requests? Would it be a huge endevaour to enable use of `DatasetVersionFileServiceBean.getFileMetadatas()`, which also allows for pagination to support the other cases?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+            return builder.parse(new InputSource(new StringReader(dataciteXmlString)));
+        } catch (ParserConfigurationException | SAXException | IOException px) {
+            return null;
+        }
+
+    }
+
     @Override
     public JsonArray getDatasetFileDetails() {
+        if (fileAndDataDetails == null) {
```
poikilotherm, line 147: Shouldn't this be a simple call to `getaDatasetFileDetails()` with a query that says "all files"? This would avoid the duplication. (And also avoid a fight over the cached `fileAndDataDetails`)

The resulting stream would be consumed in this method to generate a JsonArray on the fly using a `jakarta.json.stream.JsonCollectors.toJsonArray()`.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+                jab.add(JsonPrinter.json(dataFile, fileMetadata, true, false, true));
+            }
+            fileAndDataDetails = jab.build();
+        }
+        return fileAndDataDetails;
+    }
+
+    @Override
+    public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query) {
+        if (fileAndDataDetails == null) {
```
poikilotherm: 1. I get the intention that caching this might be beneficial if someone wants to query the *same* files multiple times. But does this sound like a realistic use case? Would someone actually try to retrieve the data *multiple* times? If a community plugin actually does that, would that be *our* problem to address? We can save the memory and gain thread safety by avoiding this for such niche edge cases.

2. If we really think we should optimize for caching here, let's do a thread-safe map with the result in it and the file export query as a key. This way, in case someone ask for different data the next time around, they'll get a consistent answer. The current implementation would simply ignore changed predicates.

3. The streaming idea would be somewhat contradicted by this caching. Should there be a marker on the export query that something should be cached for repeated queries?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+            JsonArrayBuilder jab = Json.createArrayBuilder();
+            for (FileMetadata fileMetadata : dv.getFileMetadatas()) {
```
poikilotherm: I get the intent. But there is no need for these loops. Instead, the result from `dv.getFileMetadatas()` can be streamed directly! This way, we have a lazy lookup, only executed once each item is consumed from the stream.

```java
dv.getFileMetadatas().stream()
    .map(fileMetadata -> JsonPrinter.json(fileMetadata.getDataFile(), fileMetadata, true, false, true))
    .map(JsonObjectBuilder::build)
```

In the current state, there is no filtering, ignoring the FileExportQuery parameter. This may be fine for this first iteration, but should be clearly documented and an issue should be opened to keep a record as a remaining task.

At a later point, we may introduce streamed results from the database via the `DatasetVersionFileServiceBean`, using either JPA's keyset pagination (which is potentially faster than any manual paging as the DBMS can cache the query result and skip over that instead of re-doing the query with a skip-first-n appraoch!) or a scrollable, streamed result set.

See also https://wiki.eclipse.org/EclipseLink/Examples/JPA/Pagination

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+                DataFile dataFile = fileMetadata.getDataFile();
+                jab.add(JsonPrinter.json(dataFile, fileMetadata, true, false, true));
+            }
+            fileAndDataDetails = jab.build();
+        }
+        return fileAndDataDetails;
+    }
+
+    @Override
+    public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query) {
```
poikilotherm: This method is missing Javadocs. In addition, the query parameter is not checked for being null. (The contract says query=null results in an NPE.)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
                 dv.getDataset().getGlobalId().asString(), new DataCitation(dv).getDataCiteMetadata(), dv.getDataset());
     }
-    
+
+    @Override
+    public Document getDataCiteXml(DatasetExportQuery query) {
+        // Note that the query parameter is ignored, for now
+        String dataciteXmlString = getDataCiteXml();
+
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
```
poikilotherm: This is a security issue, as the factory is not instantiated with secure options. Didn't Jim create some XmlUtil to make it easier to init these things correct?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +101,187 @@ public JsonObject getDatasetORE() {
...
+
+        try {
+            DocumentBuilder builder = factory.newDocumentBuilder();
+
+            return builder.parse(new InputSource(new StringReader(dataciteXmlString)));
+        } catch (ParserConfigurationException | SAXException | IOException px) {
+            return null;
```
poikilotherm: Why would any exception here be hidden??? Shouldn't it result in an ExportException? At least it should be logged, right?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -11,45 +12,78 @@
...
-    
+
     InternalExportDataProvider(DatasetVersion dv, InputStream is) {
         this.dv = dv;
         this.is=is;
     }
 
     @Override
-    public JsonObject getDatasetJson() {
+    public JsonObject getDatasetJson(DatasetExportQuery query) {
```
poikilotherm: I see the former method without the parameter returned the DV with files included. The SPI provides a default override for this now using a default query which is empty.

Should we make the default override in the SPI use a query with an ALL_FILES predicate to keep consistency or would you even make the default dataset query carry this predicate? (My gut says: let's move on to only provide dataset metadata here, only have the legacy endpoint return it all as before.)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -11,45 +12,78 @@
...
+        if (isOnlyDatasetLevelMetadataRequested(query)) {
+            // If we already have the "full" Json representation (with files)
+            // generated, should we return it (potentially moving MUCH more json
+            // than the client needs, or spend extra cycles generating the short
+            // form from scratch? - I'm choosing to go with latter.
+            if (jsonRepresentationNoFiles == null) {
```
poikilotherm: My gut tells me: if we really want to cache here, let's use a thread-safe map and the query as the key. In the edge case that someone really wants to retrieve the metadata again, they get a consistent result, as the query might be different the next time around!

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -319,16 +319,72 @@ public void exportAllFormats(Dataset dataset) throws ExportException {
...
+        for (Exporter e : exporterMap.values()) {
+            String formatName = e.getFormatName();
+            formatNames.add(formatName);
+            clearCachedExport(dataset, formatName);
+        }
+        clearCachedFormats(dataset, formatNames);
+        dataset.setLastExportTime(null);
+    }
+    
+    public void clearCachedFormats(Dataset dataset, List<String> formatNames) throws IOException {
```
poikilotherm, line 359: The behavior here is not aligned with `exportFormats`. With that method using a null list will clear all, here you don't even check for null, risking an NPE.

I'd argue the same way as exportFormats: use a List, either empty or filled, disallow null. If empty, clear all.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
```diff
@@ -319,16 +319,72 @@ public void exportAllFormats(Dataset dataset) throws ExportException {
...
+            throw new ExportException("Service configuration error during export. " + serviceError.getMessage());
+        } catch (RuntimeException e) {
+            logger.log(Level.FINE, e.getMessage(), e);
+            throw new ExportException(
+                    "Unknown runtime exception exporting metadata. " + (e.getMessage() == null ? "" : e.getMessage()));
+        }
+    }
 
-            dataset.setLastExportTime(null);
+    public void clearAllCachedFormats(Dataset dataset) throws IOException {
```
poikilotherm, line 354: This is a redundant method. People can just call the other defined methods. If you really want this for convenience, don't add a second implementation.

tests/integration-tests.txt
```diff
@@ -1 +1 @@
...
+DataversesIT,DatasetsIT,SwordIT,AdminIT,BuiltinUsersIT,UsersIT,UtilIT,ConfirmEmailIT,FileMetadataIT,FilesIT,SearchIT,InReviewWorkflowIT,HarvestingServerIT,HarvestingClientsIT,MoveIT,MakeDataCountApiIT,FileTypeDetectionIT,EditDDIIT,ExternalToolsIT,AccessIT,DuplicateFilesIT,DownloadFilesIT,LinkIT,DeleteUsersIT,DeactivateUsersIT,AuxiliaryFilesIT,InvalidCharactersIT,LicensesIT,NotificationsIT,BagIT,MetadataBlocksIT,NetcdfIT,SignpostingIT,FitsIT,LogoutIT,DataRetrieverApiIT,ProvIT,S3AccessIT,OpenApiIT,InfoIT,DatasetFieldsIT,SavedSearchIT,DatasetTypesIT,DataverseFeaturedItemsIT,SendFeedbackApiIT,CustomizationIT,JsonLDExportIT,WorkflowsIT,LDNInboxIT,LocalContextsIT,CorsIT,GuestbooksIT,LocallyFairIT,TabularDataExportIT
```
poikilotherm: Please do not add this test here. It's an integration test, not an API test. For historic reasons, this file has been around and collects these API tests. It should be torched with fire.

Your performance test will be automatically picked up for execution during `mvn verify -Dit.groups=performance`

If you feel we should make performance tests a part of the regular Github Action flow, let's add it there! (currently only integration and migration are picked up, see `pom.xml` property `it.groups`)

src/main/java/edu/harvard/iq/dataverse/api/Metadata.java
```diff
@@ -86,10 +89,15 @@ public Response reExportAll(
...
             @Parameter(description = "Dataset id or persistent identifier to re-export.", required = true)
-            @PathParam("id") String id) {
+            @PathParam("id") String id, 
+            @QueryParam("formats") String formats) {
         try {
             Dataset dataset = findDatasetOrDie(id);
-            datasetService.reExportDatasetAsync(dataset);
+            List<String> formatNames = null;
+            if (formats != null) {
+                formatNames = new ArrayList<>(Arrays.asList(formats.split(",")));
```
poikilotherm: As user input, the list of formats shall be verified on this code level, before entering any services.

Side note: The try/catch thing around everything is an anti-pattern and always has been. Let's not fix it here, but I wanted to repeat this, as it is an anti pattern all over our API code.

poikilotherm (dismissed):

poikilotherm:
Sorry, I submitted the review before adding a comment! My bad!

Overall, I think the SPI refactoring was a good choice. The code here is much clearer to read now, as responsibilities are clearer.
There are still a few kinks, which should be addressed. I'd be happy to help if you want to do a coding session via VC. Some of my remarks might be best solved by talking about it and improving wording and code on the SPI end. Let me know if you'd prefer me just sending patches... :see_no_evil:

tests/integration-tests.txt
landreev: I just wanted to confirm the test worked under the new github action. I have a sentence in the PR description acknowledging that it's not really an integration test. 😄 
Thank you for the pointer on `<it.groups>`, will do.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
landreev: Fair. But I am leaning towards just nuking it, the method and the comments.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
landreev: I cannot do this easily. I.e., I can move the class into `...export.ddi` and make the method above package private. BUT, then package private constructors in `export.InternalExportDataProvider` become a problem.  
Objectively, most of the code in this class is dedicated to testing the functionality in InternalExportDataProvider, more than the DDI export utility. So I'll pass on this suggestion if that's ok.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: I assumed this was a temporary hack. 
Please note that I requested to have the paginated version of `getDatasetFileDetails` return a JsonArray instead of a Stream in the final 2.1.0 of the SPI. We talked about this at some point.

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
landreev: Nah, just dead code - removed.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: Please see my reply in response to your review of lines 191-194, https://github.com/IQSS/dataverse/pull/11799#discussion_r3629001023

poikilotherm: @landreev the comment you linked was about nuking unused code. I'm not sure I get what you're saying...? 🙈

landreev: Is it? 🤔 It should point to this exchange: 
(my request to change the return type of the paginated method in the final 2.1.0 being the key; Stream does not seem to be as useful in this case). 
Screenshot 2026-07-23 at 9 02 47 AM

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: Good to know. This is why I appreciate your reviews/feedback. 
Done.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: > I get the intention that caching this might be beneficial if someone wants to query the same files multiple times. But does this sound like a realistic use case?

This is a realistic use case in exactly 1 real life scenario: when a full export of a dataset is happening, and then every metadata format is generated, sequentially, using the same instance of InternalExportDataProvider. 
There are exactly 2 formats on the list (as of now!) that actually utilize this "details" data (i.e., the metadata describing ingested tabular data - DataTables/DataVariables/... hierarchy): ddi and croissant. Without caching (as implemented in 6.11/develop) Dataverse retrieves and serializes the entire mess of this data twice. 

Once this pr is merged, the ddi export will start using the paginated approach for datasets with large numbers of variables. I am hoping to refactor the croissant export to do the same shortly thereafter. Both exports will continue retrieving the "file details" in full for datasets/versions with manageable numbers of variables. It can be argued that caching is less important in such a case - but it still feels a bit wrong to generate something from scratch that you just spent some cycles building - ? 
In any event I am open to suggestions on how to handle this. And as I said elsewhere, will gladly accept your help with all the thread safety business in this pr.

landreev: I mentioned this before, but will do it here as well: I keep thinking that a potentially better solution for the above is to use the "file details" from the ExportDataProvider ONLY in the ddi export. In croissant (and any other format that needs to encode the data/variable-level metadata) we could instead rely on what the existing `HtmlCodeBookExporter` (HTML version of the ddi) is doing: 
```
    @Override
    public  Optional<String> getPrerequisiteFormatName() {
        //This exporter relies on being able to get the output of the ddi exporter
        return Optional.of("ddi");
    }
```
i.e., declare ddi as the dependency and use that as the data source. (ddi is essentially guaranteed to encode everything we know about the datavariable-level objects). The expensive task of generating the ddi export is done once. Then the xml is cached in storage and can be streamed and parsed on the fly for cheap. 

If we were to stick with this approach, the issue of caching on the ExportDataProvider level becomes moot.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
poikilotherm: Alright, I did invest a new round of thinking about this.

When I wrote the new contracts in the SPI, I was trying to address the problem of the potentially heavily skewed JPA entity graph blowing up RAM usage by moving to a "lazyness" approach. By avoiding a big chunk of serialization into one `JsonArray` and instead picking the individual objects off one by one, the JPA lazyness would transfer down into the exporter.

What I did not keep on my radar is the most used access pattern at play here: exports happen in a sequential bulk of operations. Any time a dataset version is released, all the formats are touched. The "make a single export" pattern is only an occasional thing, for example for retrieving exports of draft dataset versions, which are not cached.

Your train of thought laid out in another comment about caching the serialized stream is true. Re-reading the serialized data can save us a lot of cycles, given our most often used access pattern to the data. Personally, I don't like your idea of reusing the DDI export for accessing file metadata from other exporters. It will require parsing the XML (cycles), it will increase complexity (know how to read DDI or provide a library) and creates a huge dependency between exports. Our current approach to ordering the exports is potentially too naive for this.

Things to keep in mind here:
1. For now, we run sequentially. However, the next optimization would be to make use of multiple CPUs and run exports in parallel. Thus, we need to keep an eye on thread safety around a) the JPA entities we use and b) the cache we provide.
2. We can still employ something like keyset pagination or a scrollable resultset for those methods which shall return potentially all the file metadata in bulk. But - we should a circuit breaker!
    - If the serialized data gets out of hand (and we cannot determine this a-priori), throw a dedicated exception (living in the SPI, same as the `ExportException`) so an exporter can switch to manual pagination. This way we don't blow up the server memory anymore and an exporter can move on to use the paginated approach. I think it's important to keep in mind that a community exporter plugin developer may not be aware of the consequences and we should be nice to them.
    - The circuit breaker needs to evaluate memory consumption by either monitoring real memory usage or estimating by model. Crucial: we can only rely on measurements *during the build* of our entities, as we cannot possibly foresee how skewed the JPA entity graph will be due to files' metadata.

As for building the cached entities to be reused: let's use a `List<JsonObject>` instead of `Stream<JsonObject>` on the SPI contract. Please note: we shall *not* cache the unserialized JPA results! These are lazy loaded by design and can never be made thread-safe. The only way to get around this is by moving them into an immutable data structure (ideally a DTO, but `JsonObject` will suffice for now), forcing the DB retrieval!

As for thread safety of the actual caches: we must build and retrieve the lists in `synchronized` blocks as immutable objects (`JsonObject` is immutable right away, the `List` just needs a `.copyOf()`) and save them in a simple `HashMap`. There is no need for a more sophisticated map implementation, as long as we have all `put()` and `get()` happen in a synchronized block.

I'm aiming for a `List` here, as for the long term, I really would like us to move on from `JsonObject` to actual DTOs, which will be much easier to handle in exporters than parsing the JSON. Using a `List` is a nice tradeoff between both worlds, also allowing to access it as a `.stream()`.

For the map keys, we shall use the `FileExportQuery` handed to us. If different exporters inquiry about different details, they must receive correct data. Yes, this might mean rebuilding, but there is no good way around this except trying to cut  away bits after the serialization. That's still CPU cycles and is way more complex.

For the paginated method, caching is much harder. The whole point of the paginated approach was to avoid having all the elements in memory, so caching all of them defeats the purpose. However, if we ever go for *parallelized* exports, we may return to caching by using a well designed LRU cache (drift will be an issue!), so some parts *may* be reused.

landreev: Re-reading the sentence about `List<JsonObject>`, I'm realizing this may be more complicated than I thought (referring to what I said on slack)

> As for building the cached entities to be reused: let's use a `List<JsonObject>` instead of `Stream<JsonObject>` on the SPI contract.

Were you referring specifically to my request, to change the return type of the _paginated_ version of `getDatasetFileDetails()` (only), from `Stream<JsonObject>` to `JsonArray` in 2.1.0? - That's how I first read it. I.e., read it as "how about `List<JsonObject>` instead of `JsonArray`?" 
But I am having a nagging suspicion now that you were talking about
```
public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query) {
```
i.e., the non-paginated version; could you please confirm before I ask any followup questions?

poikilotherm: It was about changing *both*! Let me expand a bit.

For both method the streamed approach would make sense if the data really is only ever read *once*. The deferring of serialization would save a lot of resources, as we would apply a streamed approached from the database down to the actual export and onto the stream of the HTTP response or file writing. This would be nice and lean and wouldn't require using some intermittent `JsonArray` at all..

But, as you outlined, our access pattern most often is "bursting": we export a single version with quite a bunch of exports at once. This makes it valuable to have the *serialized* data cached, even allowing for parallel processing instead of consuming a new stream every time.

One might argue for the method with the manual pagination the stream should stick around, as we won't ever be able to hold all the peaces of metadata in memory for huge file metadata cases. There the lazyness would actually come in handy. If we at some point aim for caching some of the pages when using them *in parallel*, this would be a different story. Thus, let's unify on `List` and not change gears again later.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: fair.

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: fair. 
There's also something weird going on in that method (it ends up trying to clear the caches twice). 
... it kinda looks like I may have started refactoring it, then forgot about half way through.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: Before we dive into this further, I also skipped the beginning of that sentence on the first reading:

> As for building the cached entities to be reused: let's use a `List<JsonObject>` instead of `Stream<JsonObject>` on the SPI contract.

i.e., the "cached entities to be reused" part. To be clear, I had no plans to cache and reuse the _paginated_ chunks. I was ok with the idea that the 2 currently implemented datavariable-level exporters, ddi and croissant, will be going through all these multiple pages-worth of data twice during a full reexport of large datasets; there will be wasted cpu cycles and db queries, but in my experience cycles are not what usually crashes servers. Running out of memory does though. The very problem with monster datasets, for which pagination will be used, is that they have too much tabular metadata. Which is a good reason to NOT want to cache it in memory. 

My thinking was along the same lines with the version of the method that was going to return a Stream. That it will be streaming _instead_ of caching. For that to make practical sense, we would need to re-implement how the results are retrieved from the database underneath (at the moment it is still one big query). No exporters in this pr are actually calling the streaming method as of yet. Paginated methods are used for monster datasets. For dataset with reasonable numbers of variables the legacy, deprecated `public JsonArray getDatasetFileDetails()` is called and in that case the JsonArray does get cached. So, during a full reexport the croissant exporter will get it for free, once it's generated for the ddi. 

Does any of this make sense?

landreev: (I do agree that the current implementation of `public Stream<JsonObject> getDatasetFileDetails(FileExportQuery query)` is problematic; and that "nobody uses it _yet_" is not a good excuse)

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: done

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: done

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: Items 1. and 2. done. 
I am a little bit less enthusiastic about touching the dataset vs version part. 
This is a very ancient legacy, that Dataverse only exports the latest published version as part of the standard export framework. Some exceptions have been made to accommodate exporting (but not caching?) drafts in some cases (hence `getExport(DatasetVersion ...)`, I believe). But I don't want to be the one to open the can of worms of making arbitrary versions readily exportable. (the caching framework would need to be reconsidered, if nothing else)

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: Agree. This is what I meant when I acknowledged that the implementation of the streaming `getDatasetFileDetails(FileExportQuery query)` was "problematic" in another thread. 
I will change is so that only the legacy, deprecated `getDatasetFileDetails()` uses this caching mechanism.

landreev: OK, done (for now). The streaming `getDatasetFileDetails(FileExportQuery query)` method is not utilizing _any_ caching at this point. 
How we refactor the method going forward and whether any caching is going to be implemented, and how, will depend on whether we keep `Stream<JsonObject>` as the return type, or change it to `List<JsonObject>` as is being discussed elsewhere, or ...

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: This sounds ok, I think... but this may be one thing that could benefit from actually talking about in some synchronous way.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: done

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: done

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: The answers are 
1. Yes - the for loop was utterly unnecessary (done)
2. Yes - we agreed last week to have a dedicated helper function added (I will wait for that part).

src/main/java/edu/harvard/iq/dataverse/export/ddi/DdiExportUtil.java
landreev: Can this be resolved? - Otherwise please advise on how to refactor this.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: As we agreed last week, this was refactored as suggested in the top comment in the thread.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: We have agreed to keep `Stream<JsonObject>` as the return types of the 2 new methods. And NOT try to cache anything, except in the deprecated legacy methods. OK to resolve?

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: Streaming has been refactored per suggestions in another thread, as 
```
dv.getFileMetadatas().stream()
    .map(fileMetadata -> JsonPrinter.jsonDatafileWithDatatableForExport(fileMetadata.getDataFile(), fileMetadata))
    .map(JsonObjectBuilder::build);
```

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: done, or mostly done, I think.

landreev:
(will resolve conflicts shortly)

landreev:
Merge conflicts resolved; will keep an eye on the tests.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
```diff
@@ -67,26 +118,159 @@ public JsonObject getDatasetORE() {
...
+    }
+
     @Override
     public JsonArray getDatasetFileDetails() {
-        JsonArrayBuilder jab = JsonUtil.createArrayBuilder();
-        for (FileMetadata fileMetadata : dv.getFileMetadatas()) {
-            DataFile dataFile = fileMetadata.getDataFile();
-            jab.add(JsonPrinter.json(dataFile, fileMetadata, true));
+        if (fileAndDataDetails == null) {
+            JsonArrayBuilder jab = Json.createArrayBuilder();
```
pdurbin: When I run this...

```
mvn test -Dtest=TabularDataExportIT#exportTabularMetadata
```

... I get these errors:

```
[INFO] Reading API signatures: /Users/pdurbin/github/iqss/dataverse/src/maven/forbiddenapis.txt
[INFO] Loading classes to check...
[INFO] Scanning classes for violations...
[ERROR] Forbidden method invocation: jakarta.json.Json#createArrayBuilder() [Use edu.harvard.iq.dataverse.util.json.JsonUtil.createArrayBuilder(...) instead]
[ERROR]   in edu.harvard.iq.dataverse.export.InternalExportDataProvider (InternalExportDataProvider.java:148)
[ERROR] Forbidden method invocation: jakarta.json.Json#createArrayBuilder() [Use edu.harvard.iq.dataverse.util.json.JsonUtil.createArrayBuilder(...) instead]
[ERROR]   in edu.harvard.iq.dataverse.export.InternalExportDataProvider (InternalExportDataProvider.java:179)
[ERROR] Scanned 1225 class file(s) for forbidden API invocations (in 0.77s), 2 error(s).
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

pdurbin (commented):
A quick review.

src/main/java/edu/harvard/iq/dataverse/api/Metadata.java
```diff
@@ -76,20 +83,30 @@ public Response reExportAll(
                 return error(Response.Status.BAD_REQUEST, "Invalid date format for olderThan parameter. Expected format: YYYY-MM-DD");
             }
         }
-        datasetService.reExportAllAsync(reExportDate);
+        
+        // validate the format names argument supplied (or obain an empty list
```
pdurbin:
```suggestion
        // validate the format names argument supplied (or obtain an empty list 
```

src/main/java/edu/harvard/iq/dataverse/api/Metadata.java
```diff
@@ -76,20 +83,30 @@ public Response reExportAll(
...
+    public Response exportDatasetByPersistentId(
             @Parameter(description = "Dataset id or persistent identifier to re-export.", required = true)
-            @PathParam("id") String id) {
+            @PathParam("id") String id,
+            @QueryParam("formats") String formats) {
         try {
             Dataset dataset = findDatasetOrDie(id);
-            datasetService.reExportDatasetAsync(dataset);
+            // validate the format names argument supplied (or obain an empty list 
+            // that the method will produce if null is supplied):
```
pdurbin:
```suggestion
            // validate the format names argument supplied (or obtain an empty list 
            // that the method will produce if null is supplied):
```

src/main/java/edu/harvard/iq/dataverse/api/Metadata.java
```diff
@@ -143,4 +159,28 @@ public Response exportOaiSet(
...
+        Set<String> supportedFormatNames = new HashSet<>();
+        for (String[] providerLabels : ExportService.getInstance().getExportersLabels()) {
+            supportedFormatNames.add(providerLabels[1]);
+        }
+
+        //for (String formatName : formatNames) {
+        //    if (!supportedFormatNames.contains(formatName)) {
+        //        throw new BadRequestException(formatName + " is not a supported format");
+        //    }
+        //}
```
pdurbin: Delete this? 🤷

src/test/java/edu/harvard/iq/dataverse/export/TabularDataExportIT.java
pdurbin, line 1: TabularDataExportIT.java is new. Should it be added to tests/integration-tests.txt?

pdurbin: Nevermind. @landreev pointed me to the description. It runs because of this: 

```
-        <it.groups>integration,migration</it.groups>
+        <it.groups>integration,migration,performance</it.groups>
```

src/main/java/edu/harvard/iq/dataverse/export/ExportService.java
landreev: I believe the consensus was that the last part of the above would be addressed in stage 2 of refactoring, in another pr.

src/main/java/edu/harvard/iq/dataverse/export/InternalExportDataProvider.java
landreev: The final decision was to leave this legacy/deprecated method's implementation as is, for the sake of ensuring the legacy behavior was unchanged.

doc/release-notes/11405-export-improvements.md
```diff
@@ -0,0 +1,2 @@
+Dataverse metadata export code has been re-factored to use the new and improved Data Export framework (note also that the relevant interface and the accompanying code have been moved out of the main Dataverse repository and into the dedicated [GDCC project](https://github.com/gdcc/dataverse-spi)). In this release this will result in a measurable improvement in exporting the DDI format. But it will be possible to take advantage of this refactoring in the exports of other data-rich formats in future releases.
+An optional query parameter ``formats`` has been added to the ``reExportAll`` and ``reExportDataset``, allowing an administrator to force re-export of only the formats specified.
```
pdurbin: We can wordsmith this at release time but a few thoughts.

I'd probably pitch this as two related features:

```
## Highlights

### Export improvements

- Faster DDI export
- Ability to export only selected formats 
```

Then, further down:

```
## API updates

### Ability to export only selected formats

curl example, link to API Guide.
```

Still further down:

```
## Updates for developers

We've refactored, blah blah blah.
```

doc/sphinx-guides/source/api/changelog.rst
```diff
@@ -26,6 +26,7 @@ v6.12
   - **/api/admin/index/perms/{id}**
 
 - Dataset creation API calls may now behave differently when neither a license nor terms are provided, depending on the new :ref:`dataverse.feature.do-not-assume-default-license` feature flag.
+- The metadata export APIs ``/api/admin/metadata/reExportAll`` and ``/api/admin/metadata/<id>/reExportDataset`` now support an optional parameter ``formats=`` allowing an administrator to force a reexport of the specified formats only.
```
pdurbin: Meh. Can we remove this? It's not a backward-incompatible change. It's just a change.

doc/release-notes/11405-export-improvements.md
landreev: Sure, makes sense. Will do.

landreev:
The "how to test" section was updated with some extra info.

pdurbin:
I reexported /usr/local/dvn/data/10.7910/DVN/KZW8R2/export_ddi.cached and it worked fine. I didn't get an exact time but let's say it look 11 minutes or so.

Merging!

pdurbin (approved):

PR #11805, opened 2025-09-09
New Dataset Move Notification

### LINKED ISSUE #11670: Feature Request: Notifications - Collection stakeholders should receive a notification if a dataset is "moved" from their collection
When a dataset is moved from a collection, notifications should be sent to the dataset depositor, collection curator, and collection admin.

(Example users' roles: collection administrator, collection curator, dataset depositor)
Dataset depositor, collection curator, and collection administrator.

Accidental move of datasets from one journal into another, which caused repository managers and support team to figure out what happened. 

Notifications. Whenever content permission is changed, notifications should be sent.

stevenwinship:
**What this PR does / why we need it**: When a dataset is moved from a collection, notifications should be sent to the dataset depositor, collection curator, and collection admin.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11670

- Closes #11670

**Suggestions on how to test this**: Set SendNotificationOnDatasetMove setting. See MoveIT test. Test API and Dashboard in JSF UI

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
@@ -218,7 +218,7 @@ public class MoveIT {
         notificationListResponse.then().assertThat()
                 .statusCode(OK.getStatusCode());
         boolean found = false;
-        List<Map<String, String>> notifications = notificationListResponse.body().jsonPath().getList("data.notifications");
+        List<Map<String, String>> notifications = notificationListResponse.body().jsonPath().getList("data");
 
         for (Map<String, String> notification : notifications) {
             if ("DATASETMOVED".equalsIgnoreCase(notification.get("type"))) {


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
```diff
@@ -54,6 +56,7 @@ public MoveDatasetCommand(DataverseRequest aRequest, Dataset moved, Dataverse de
         this.moved = moved;
         this.destination = destination;
         this.force= force;
+        this.allowSelfNotification = allowSelfNotification;
```
qqmyers, line 60: Does this blow up (NPE) for a null input?

src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
```diff
@@ -145,6 +156,85 @@ public void testMoveDataset() {
...
+        moveDataset.then().assertThat()
+                .statusCode(OK.getStatusCode())
+                .body("data.message", equalTo(BundleUtil.getStringFromBundle("datasets.api.moveDataset.success")));
+
+        // verify that a notification was sent to user1
+        Response getNotifications = UtilIT.getNotifications(user1ApiToken);
+        getNotifications.prettyPrint();
+        getNotifications.then().assertThat()
+                .statusCode(OK.getStatusCode())
+                .body("data.notifications[0].type", equalTo("DATASETMOVED"))
```
qqmyers: Given that this test can (currently) be run in parallel with other tests, other notifications can be created and notifications[0] may be something else (so random test fail possibilities). Maybe another @Isolated annotation? Or loop through notifications in case there's more than one?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
```diff
@@ -145,12 +148,55 @@ public void executeImpl(CommandContext ctxt) throws CommandException {
...
+        // 3. Get all users with publish permission on the dataset's original owner (dataverse) and notify them.
+        Map<String, AuthenticatedUser> recipients = ctxt.permissions().getDistinctUsersWithPermissionOn(Permission.PublishDataset, originalOwner);
+        // make sure the requestor is in the recipient list in case they don't match the permission
+        if (requestor != null) {
+            recipients.put(requestor.getIdentifier(), requestor);
+        }
+
+        recipients.values()
+                .stream()
+                .filter(recipient -> allowSelfNotification || !recipient.equals(requestor))
```
qqmyers: Can you just delete the requestor from the Map up front if allowSelfNotification is false?

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
stevenwinship: fixed NPE and moved sendNotification to onSuccess

src/test/java/edu/harvard/iq/dataverse/api/MoveIT.java
stevenwinship: refactored the test to find the notification in the list

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/MoveDatasetCommand.java
stevenwinship: removed requestor if not self notify

PR #11819, opened 2025-09-15
Prevent ExcludeEmailFromExport to hide the user email when an authorized user is obtaining metadata for editing a dataset

### LINKED ISSUE #11714: Prevent ExcludeEmailFromExport to hide the user email when an authorized user is obtaining metadata for editing a dataset

Currently, when the `ExcludeEmailFromExport` setting is enabled, the getVersion `/{id}/versions/{versionId}` API prevents the author email field from being returned. This behavior makes sense for dataset display in view mode, but it causes issues because the dataset editing form also relies on this same endpoint to pre-populate fields, including the author email.

At present, within the SPA we use the` {id}/versions/{versionId}` endpoint (with metadata blocks) for two different purposes:

- Displaying datasets in view mode.
- Populating the dataset edit form.

For view mode, it is acceptable for author emails to be hidden or displayed based on the configuration. However, for the edit form, we always require access to the author email field when the user has edit permissions.

**Requirement**

Enhance the API so that when an authorized user (with dataset edit permissions) is editing a dataset, the API always returns the author email, regardless of the ExcludeEmailFromExport setting.

**Potential Approaches**

Introduce an additional option/flag in the getVersion endpoint to indicate that the request is being made for editing purposes (with appropriate permission checks).

Alternatively, create a separate endpoint dedicated to retrieving dataset version data for editing purposes.

API User

SPA Release

ExcludeEmailFromExport setting behavior

Do not exclude author email when dataset metadata is obtained for editing purposes and user has edit permission.

None

Yes

stevenwinship:
**What this PR does / why we need it**: SPA needs to get the contact email to fill the edit screen even when the ExcludeEmailFromExport setting is set. Only for users calling the get dataset version API and the user has EditDataset permissions.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11714

- Closes #11714

**Suggestions on how to test this**: See DatasetIT test 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
@@ -495,17 +495,17 @@ public class JsonPrinter {
     }
 
     public static JsonObjectBuilder json(DatasetVersion dsv, boolean includeFiles) {
-        return json(dsv, null, includeFiles, false,true);
+        return json(dsv, null, includeFiles, false, true, false);
     }
     public static JsonObjectBuilder json(DatasetVersion dsv, boolean includeFiles, boolean includeMetadataBlocks) {
-        return json(dsv, null, includeFiles, false, includeMetadataBlocks);
+        return json(dsv, null, includeFiles, false, includeMetadataBlocks, false);
     }
     public static JsonObjectBuilder json(DatasetVersion dsv, List<String> anonymizedFieldTypeNamesList,
                                          boolean includeFiles, boolean returnOwners) {
-        return  json( dsv,  anonymizedFieldTypeNamesList, includeFiles,  returnOwners,true);
+        return  json(dsv, anonymizedFieldTypeNamesList, includeFiles, returnOwners, true, false);
     }
     public static JsonObjectBuilder json(DatasetVersion dsv, List<String> anonymizedFieldTypeNamesList,
-        boolean includeFiles, boolean returnOwners, boolean includeMetadataBlocks) {
+        boolean includeFiles, boolean returnOwners, boolean includeMetadataBlocks, boolean ignoreSettingExcludeEmailFromExport) {
         Dataset dataset = dsv.getDataset();
         JsonObjectBuilder bld = jsonObjectBuilder()
                 .add("id", dsv.getId()).add("datasetId", dataset.getId())
@@ -554,10 +554,8 @@ public class JsonPrinter {
                 .add("studyCompletion", dsv.getTermsOfUseAndAccess().getStudyCompletion())
                 .add("fileAccessRequest", dsv.getTermsOfUseAndAccess().isFileAccessRequest());
         if(includeMetadataBlocks) {
-            bld.add("metadataBlocks", (anonymizedFieldTypeNamesList != null) ?
-                    jsonByBlocks(dsv.getDatasetFields(), anonymizedFieldTypeNamesList)
-                    : jsonByBlocks(dsv.getDatasetFields())
-            );
+            bld.add("metadataBlocks",
+                    jsonByBlocks(dsv.getDatasetFields(), anonymizedFieldTypeNamesList, ignoreSettingExcludeEmailFromExport));
         }
         if(returnOwners){
             bld.add("isPartOf", getOwnersFromDvObject(dataset));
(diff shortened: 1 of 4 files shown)


doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
```diff
@@ -0,0 +1,7 @@
+New query parameter (ignoreSettingExcludeEmailFromExport) for API /api/datasets/:persistentId/versions/{versionId}
+
```
sekmiller, line 2: Do we need to add this info to the native api doc

doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
sekmiller, line 1: Should we put this in the api doc or are we just expecting it to be used by the SPA?

doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
pdurbin: API doc please.

doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
pdurbin: Same. API doc, please.

doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
stevenwinship: Done

doc/release-notes/11714-prevent-exclude-email-from-export-for-permitted-user-get-dataset-version.md
stevenwinship: Done

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
```diff
@@ -101,6 +98,10 @@ public static void setUpClass() {
          */
     }
 
+    @AfterEach
```
qqmyers, line 97: This works, but it's called for every test in the class when most tests don't use it, and the testExcludeEmail is still doing its own call to delete the setting (but, as you noted the other day/in another PR, if an assert fails, the line to cleanup won't get called), and it's deleted again in the @AfterAll method.

Does it make sense to remove the redundant deletes and keep the @AfterEach? Or just use try/finally in the two tests that change the setting to assure it's deleted even if they fail?

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
```diff
@@ -668,14 +666,22 @@ public static JsonObjectBuilder json(MetadataBlock block, List<DatasetField> fie
...
         JsonObjectBuilder blockBld = jsonObjectBuilder();
 
         blockBld.add("displayName", block.getDisplayName());
         blockBld.add("name", block.getName());
 
         final JsonArrayBuilder fieldsArray = Json.createArrayBuilder();
         Map<Long, JsonObject> cvocMap = (datasetFieldService==null) ? new HashMap<Long, JsonObject>() :datasetFieldService.getCVocConf(true);
-        DatasetFieldWalker.walk(fields, settingsService, cvocMap, new DatasetFieldsToJson(fieldsArray, anonymizedFieldTypeNamesList));
+        if (ignoreSettingExcludeEmailFromExport) {
+            DatasetFieldWalker.walk(fields, null, cvocMap, new DatasetFieldsToJson(fieldsArray, anonymizedFieldTypeNamesList));
```
qqmyers: While it's true today that the SettingsService is only used in the DatasetFieldWalker for the ExcludeEmail setting today, setting it to null based on the ignoreSetting... flag is ~kludgy. 

Can you either add comments here and in the DatasetFieldWalker to note that the SettingsService will be null in this case (and any other use of the SettingsService in the walker would break), or, perhaps cleaner, refactor to just pass in the boolean to ignore the email field or not (which would be ignoreSettingExcludeEmailFromExport && the setting in this case). (Not sure there will ever be another setting used in the walker so maybe comments are good enough?)

qqmyers (changes requested):
This looks good to me. I made a couple suggestions for cleanup.

I also noticed that there is a Dataverses api call that also uses this setting - probably another case that will eventually need this override.

src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java
stevenwinship: Cleaned up

src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java
stevenwinship: Refactored

qqmyers (approved):
Looks good - nice improvement to the DatasetFieldWalker!

PR #11825, opened 2025-09-17
11824 long metadata values break the table layout

### LINKED ISSUE #11824: Very long metadata values break the table layout
When a metadata field contains a very long value (e.g., a URL with many parameters or a string without spaces), the HTML table layout breaks because the value is treated as a single unbreakable word.

[Example available here](https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/9XVBIO)
Some metadata cells display long values that do not wrap properly and break the layout.

**Expected behavior**

Long values should display correctly without breaking the layout.
They should wrap across multiple lines within the cell width.

**Suggested fix**

Apply a generic CSS rule to metadata cells, for example:

```
table.metadata {
  word-break: normal;
  overflow-wrap: anywhere;
}
```
This ensures that long values (URLs, space-free strings, etc.) wrap properly and do not break the layout.

stevenferey:

- Closes #11824

### DIFF
--- a/src/main/webapp/resources/css/structure.css
+++ b/src/main/webapp/resources/css/structure.css
@@ -600,6 +600,7 @@ div[id$='roleDisplay'] span.label, div[id$='roleDetails'] span.label {display:in
 #title {font-size:36px !important;line-height:1.1;font-weight:bold;margin-top:0px;margin-bottom:0px;}
 #author, #contact {display:block; margin-bottom:.5em;}
 
+/* #11824 */
 table.metadata {
   word-break: normal;
   overflow-wrap: anywhere;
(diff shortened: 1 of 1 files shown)


ofahimIQSS:
I'm still seeing the issue 
internal:
image
image

local:
image
image

qqmyers:
The issue was only about the layout in the metadata table, which I think is fixed in the example dataset in the images. 

That said, does it make sense to move the css statements in the PR up to https://github.com/IQSS/dataverse/blob/56bf62880a863281874565c1cffa41490115e6df/src/main/webapp/resources/css/structure.css#L7 so they affect the default for all fields. From a quick test, I think resolves all of these issues at once/ may be a better approach than just adding them to specific elements as we've been doing. @stevenferey - do you want to adjust your PR to do that?

ofahimIQSS:
I like this approach

ofahimIQSS:
thanks, looks good from my end - merging

PR #11832, opened 2025-09-19
QDR-DataCite Scaling

qqmyers:
**What this PR does / why we need it**: This PR improves scaling of requests to DataCite in two ways:

- adds retries with a delay if/when DataCite responds with error codes that indicate requests are being throttled (429) or their server is temporarily not responding (503, 504)
- optionally checks with DataCite to see if updates are needed before sending updates

The former is fairly straight forward - rather than failing immediately, Dataverse will wait/temporarily slow requests to see if DataCite recovers/Dataverse can drop below the rate limit. If things recover, Dataverse's operations will succeed. If not, there could be a delay of ~ 1minute before a final error occurs and the operation fails.

The latter is perhaps more controversial (there was discussion several years ago about whether this is useful): instead of always sending an update, causing DataCite to write info, this optional change causes Dataverse to first query DataCite (a read) and only send an update if the local info is different than what DataCite has. In cases such as file DOIs where changes are infrequent, this results in many reads and few writes instead of many writes and DataCite (and growing records as they track all writes of new metadata) which appears to be faster. It may be generally useful, but installations not using file DOIs may not want to try it.

- Closes #

**Special notes for your reviewer**: QDR had trouble publishing a dataset with >10K files before this change and succeeded after.

**Suggestions on how to test this**: Minimally regression test (w/ and w/o flag). 

Could also attempt to create/publish a dataset with file DOIs and many files using the DataCite test server and see if the changes increase the success rate/largest size that succeeds and/or improves performance (i.e. with the flag on.) I'm not sure this is worth it given the testing/deployment at QDR.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/datacite/DOIDataCiteRegisterService.java
+++ b/src/main/java/edu/harvard/iq/dataverse/pidproviders/doi/datacite/DOIDataCiteRegisterService.java
@@ -95,7 +95,14 @@ public class DOIDataCiteRegisterService {
             }
             retString = "metadata:\\r" + client.postMetadata(xmlMetadata) + "\\r";
         }
-        if (!target.equals(client.getUrl(numericIdentifier))) {
+        String currentUrl = null;
+        try {
+            //May get a 204 if the DOI is still draft
+            currentUrl = client.getUrl(numericIdentifier);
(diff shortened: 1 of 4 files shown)


landreev:
Looks great, merging. 
I did not drive into actually triggering Datacite rate limit/429s. But the code looks bulletproof.

PR #11844, opened 2025-09-25
clean up use of session in Access APIs for SPA

### LINKED ISSUE #11740: Access API file downloads do not work with Bearer Token authentication on draft datasets.

This issue can be easily reproduced from the SPA, which uses the bearer token authentication mechanism. Files in a dataset’s draft version are subject to permission checks to ensure that the user attempting to download them has the necessary access rights. The Access API currently relies on the Dataverse session (via the JSESSIONID cookie) to verify user permissions.

We need to decouple the Access API from this responsibility, so that it validates permissions for the authenticated user provided by the API authentication filter—regardless of the authentication mechanism used. In the SPA’s case, this is a bearer token.

At present, there are multiple references to DataverseSession within the Access API. When implementing these changes, it’s important to maintain backward compatibility, since JSF uses these endpoints and relies on the session cookie authentication mechanism.

This is the error we are obtaining from the SPA when attempting to download a file from a draft dataset version:

Image

* When does this issue occur?

Download files from a draft version using the access API

SPA Dataset page

* What happens?

We encounter an API error and files are not downloaded.

SPA and API users

Files to be correctly downloaded from draft versions.

Latest

stevenwinship:
**What this PR does / why we need it**: We need to decouple the Access API from this responsibility, so that it validates permissions for the authenticated user provided by the API authentication filter—regardless of the authentication mechanism used. In the SPA’s case, this is a bearer token.

Must still be backward compatible with JSF so session is still needed but localized 

The original issue also has a bug in that the api called is not being sent as bearer token.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11740

- Closes #11740
This fix doesn't close this issue until the frontend adds the bearer token to the api call

**Special notes for your reviewer**: Removed a lot of the code referencing the session. Some code was not being utilized as downstream code doesn't behave as it did when this older code was written. I tried to consolidate the code accessing the session to a single method.

**Suggestions on how to test this**: Once the frontend is fixed the UI can be used to test. Until then a curl to the endpoint with the Bearer token can be used. I tested by adding a log statement to display the bearer token (BearerTokenAuthMechanism) from the previous call to pass to this api call. 

**Additional documentation**:


stevenwinship:
I removed the code that compared the authUser and sessionUser to null since these calls no longer returned null, they returned guestUser if the user was not found

src/main/java/edu/harvard/iq/dataverse/GuestbookResponseServiceBean.java
```diff
@@ -762,30 +762,13 @@ private void initCustomQuestions(GuestbookResponse guestbookResponse, Dataset da
...
-            guestbookResponse.setAuthenticatedUser(getAuthenticatedUser(user));
-        } else {
-            guestbookResponse.setEmail("");
-            guestbookResponse.setName("");
-            guestbookResponse.setInstitution("");
-            guestbookResponse.setPosition("");
-            guestbookResponse.setAuthenticatedUser(null);
-        }
-        guestbookResponse.setSessionId(session.toString());
+    private void setUserDefaultResponses(GuestbookResponse guestbookResponse, DataverseSession session, User user) {
```
sekmiller, line 773: Wondering if we need two different methods for setUserDefaultResponses(see 774 below). Is the session user ever null - seems like it would always at least return a guest user.

stevenwinship: Session user is never null. It defaults to Guest User

stevenwinship: This method is passed the user from the API call so it can be either an Auth user or the session user
The other method is used by JSF where it gets the user from the session only. I will consolidate the code but keep 2 methods

stevenwinship: We should create a new issue to clean up the guestbook. It would seem that it needs work beyond the scope of this issue

stevenwinship: Created Issue https://github.com/IQSS/dataverse/issues/11964

landreev:
@stevenwinship 
Please advise if I'm doing something wrong, but I'm not seeing the desired behavior in the UI so far. Getting a 403 when attempting to download a file from the SPA even though I should be authorized to do so: 
```
{"status":"ERROR","code":403,"message":"Not authorized to access this object via this API endpoint. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.","requestUrl":"http://qa.dataverse.org/api/v1/access/datafile/13378680?format=original","requestMethod":"GET"}
```

A quick guess (but see below for more details): any chance this is because when a JSESSION is present, the API gives it a priority over the bearer token? - Note that when using the browser, there will always be a JSESSION cookie; but it will probably be a guest user jsession, regardless of what the user is logged in under the SPA, unless you log in under JSF as well. 

Testing on qa, under the scenario Ellen suggested: the above is a restricted file in a published dataset (https://qa.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/DVN/H3KBKK), where access was explicitly granted to a non-root user. When logged in as that user in the SPA, the UI is showing the file as unlocked and gives me an option to download it. But then I get a 403 from the access API. 

Tested the above in Chrome where I went directly to /spa and logged in via HarvardKey. 
Earlier I had a misleading experience in Firefox where I WAS able to download that file in the SPA. However, I realized I was only able to download it because I was still logged in as the superuser under the JSF UI in a different window (i.e., my browser was sending along a JSESSION of the superuser, which the API still recognizes). Once I logged out in the JSF window, I could no longer download the file under the SPA, with the same 403.

landreev:
P.S. the access log entry: (note the non-null jsession) 
```
"::1" "20/Jan/2026:15:30:38 +0000" "GET /api/access/datafile/13378680?format=original HTTP/1.1" 403 301 "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:140.0) Gecko/20100101 Firefox/140.0" "https://qa.dataverse.org/spa/datasets?persistentId=doi:10.70122/DVN/H3KBKK&version=1.0" "c06f1fc15019a40473e2b50d1c26" 
```

landreev:
That said, I'm not seeing the bearer token in the headers when the SPA issues that /api/access/datafile call (?) - 
from the javascript console in Chrome: 
Screenshot 2026-01-20 at 12 47 43 PM

landreev:
@stevenwinship @ekraffmiller 
If I extract the bearer token from the headers of the (successful) `/api/v1/access/datafile/13378680/userPermissions` call made by the SPA (screenshot below), I can download the restricted file on the command line with curl, via 
```
curl -L -H "Authorization: Bearer ... snipped for brevity ..." https://qa.dataverse.org/api/v1/access/datafile/13378680
```
It's working properly if I send along an unauthorized jsession cookie as well  (via `--cookie "JSESSIONID=..."`), so please disregard my earlier guess/theory. 
 
So, I am inclined to go ahead and merge the PR, since it appears to be working as advertised. Whatever problems that remain appear to be on the SPA side and will hopefully be resolved soon. 

Screenshot 2026-01-20 at 1 49 03 PM

PR #11847, opened 2025-09-25
QDR: Make clearExportTimestamps invalidate existing cached entries

qqmyers:


qqmyers:
I was hoping to support incremental generation of exports in this PR, and fix a couple related issues (a null lastExportTime breaks the calculation to update the DDI export as embargoes expire, a failure of any export leaves the lastExportTime null, the call to update the lastExportTime when an individual format is created during the Datasets /export api call isn't actually updating the db), but I realize now that with the export timer, an exportAll is going to get fired off within a day anyway. I think without the timer, this might be useful, but it does mean flipping to a design where not all export files will exist (perhaps reasonable) and lastExportTime means ~oldestExportGenerationTime, but this is getting bigger than I want to tackle now.

PR #11851, opened 2025-09-29
Notifications API fixes for In-App

### LINKED ISSUE #11804: Update Notifications API response values
For notifications about files, the notification message needs a dataset link, rather than a file link. So these notification types should be updated to return `datasetDisplayName` and `datasetPersistentId`:
REQUESTFILEACCESS
REQUESTEDFILEACCESS
GRANTFILEACCESS
REJECTFILEACCESS

Also, for notification type DATASETMENTIONED, `additionalInfo` should return a JSON object.  The properties in the object are:
`@id` - a url/uri to the resource that mentions the dataset
`@type` - what type of resource this is 
`name` - name of the resource
`relationship` - how the resource is related to the dataset

GPortas:

**Endpoint:** `notifications/all`

**Enhancements:**

* When the query parameter `inAppNotificationFormat=true` is set:

    * Notifications of types:

        * `REQUESTFILEACCESS`
        * `REQUESTEDFILEACCESS`
        * `GRANTFILEACCESS`
        * `REJECTFILEACCESS`

  now return both the **dataset display name** and **dataset persistent identifier**.

Notifications of type `DATASETMENTIONED` now return a **formatted JSON** in the `additionalInfo` field when this field contains a valid persisted JSON string, instead of a raw JSON string.
 
- Closes #11804

- Visual inspection of the new added properties, and related unit test cases
- Test generating the updated notification types and test the endpoint for InApp notifications by calling:

`curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/notifications/all?inAppNotificationFormat=true"
`
No
Yes, attached


ekraffmiller:
Hi @GPortas , I tested this with the frontend, and I found some issues in the response data.

For GRANTFILEACCESS and REJECTFILEACCESS notifications, it's returning objectDeleted: true, rather than the datasetDisplayName and datasetPersistentId. For example for GRANTFILEACCESS, the notification coming from the API:

```
 {
 displayAsRead: true 
 id: 9 
 objectDeleted: true 
 sentTimestamp: "2025-10-03T00:23:32Z" 
 type: "GRANTFILEACCESS"
 }
```


 For REQUESTEDFILEACCESS, it's not returning the full datasetPersistentId.
For example it is returning:
`datasetPersistentIdentifier: "FK2/1LPTK4"`
when it should be:
`datasetPersistentIdentifier: ”doi:10.5072/FK2/1LPTK4"   `

GPortas:
@ekraffmiller It should be fixed now, thanks.

PR #11857, opened 2025-10-01
Metadata language api call

### LINKED ISSUE #11856: Feature Request: Add a new API endpoint for querying the metadata language of a Dataverse collection
A new API call for querying the metadata language of a Dataverse collection like this:
`GET /dataverses/{alias}/metadataLanguage`
If there are no metadata languages configured on the server, the response is an empty array.
If there are metadata languages configured on the server, but the collection doesn't have a mandatory metadata language, then the response is the array of available metadata languages as introduced [here](https://guides.dataverse.org/en/latest/installation/config.html#allowing-the-language-used-for-dataset-metadata-to-be-specified), otherwise the response is the mandatory metadata language.

API User

I found a bug in the RSpace Dataverse integration (reported [here](https://github.com/rspace-os/rspace-dataverse-adapter/issues/2)). This new API call would help fixing this issue.

Only in Rspace:
https://github.com/rspace-os/rspace-web/issues/478
https://github.com/rspace-os/rspace-dataverse-adapter/issues/2

Yes, I will open a pull request soon.

akos326:
Adds a new API endpoint for querying the metadata language of a Dataverse collection.

- Closes #11856

Start a Dataverse instance and add some metadata languages as described [here](https://guides.dataverse.org/en/latest/installation/config.html#allowing-the-language-used-for-dataset-metadata-to-be-specified).
No, only the API.
Yes, the documentation of this new API endpoint.


src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -5001,4 +5001,12 @@ public static Response callCallbackUrl(String callbackUrl) {
...
     }
+
+    public static Response getDataverseMetadataLanguage(String alias, String apiToken) {
+        return given()
+                .header(API_TOKEN_HTTP_HEADER, apiToken)
+                .get("/api/dataverses/"
+                        + alias
+                        + "/metadataLanguage");
+    }
 }
```
pdurbin, line 5170: Strange. A number of tests are failing.

Screenshot 2025-10-01 at 09-27-58 IQSS-Dataverse-Develop-PR_PR-11857 #1 edu harvard iq dataverse api Jenkins

Screenshot 2025-10-01 at 09-28-07 IQSS-Dataverse-Develop-PR_PR-11857 #1 test - testBagItExport Jenkins

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -2457,4 +2457,35 @@ private String getSuperuserToken() {
         UtilIT.makeSuperUser(username);
         return adminApiToken;
     }
+    
+    @Test
+    public void testGetDataverseMetadataLanguage() {
+        Response createUser = UtilIT.createRandomUser();
+        createUser.prettyPrint();
```
pdurbin, line 2766: I would suggest checking for OK or CREATED (I forget which) here.

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -2457,4 +2457,35 @@ private String getSuperuserToken() {
...
+        String username = UtilIT.getUsernameFromResponse(createUser);
+        String apiToken = UtilIT.getApiTokenFromResponse(createUser);
+        UtilIT.deleteSetting(":MetadataLanguages");
```
pdurbin: This should also go in the `@BeforeAll` setupClass and the `@AfterAll` afterClass.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1969,4 +1969,15 @@ public Response createTemplate(@Context ContainerRequestContext crc, String body
...
+    @GET
+    @AuthRequired
+    @Path("{identifier}/metadataLanguage")
+    public Response getMetadataLanguage(@Context ContainerRequestContext crc, @PathParam("identifier") String dvIdtf) {
+        return response(req -> {
+            Dataverse dataverse = findDataverseOrDie(dvIdtf);
+            return ok(jsonLanguage(execCommand(
+                    new GetDataverseMetadataLanguageCommand(req, dataverse))));
+        }, getRequestUser(crc));
+    }
```
pdurbin, line 2014: This API call should be documented in doc/sphinx-guides/source/api/native-api.rst

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1969,4 +1969,15 @@ public Response createTemplate(@Context ContainerRequestContext crc, String body
             return e.getResponse();
```
pdurbin, line 2001: Can you please add a release note snippet? Please see https://guides.dataverse.org/en/6.8/developers/version-control.html#writing-a-release-note-snippet

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -1969,4 +1969,15 @@ public Response createTemplate(@Context ContainerRequestContext crc, String body
...
         }
     }
+
+    @GET
+    @AuthRequired
+    @Path("{identifier}/metadataLanguage")
```
qqmyers: Perhaps {identifier}/allowedMetadataLanguages ? or allowedDatasetMetadataLanguages? - this may be more than one language (or none) and it doesn't represent the metadataLanguage of the collection itself - it's the set of metadataLanguages allowed for datasets in the collection.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseMetadataLanguageCommand.java
```diff
@@ -0,0 +1,41 @@
...
+
+import edu.harvard.iq.dataverse.Dataverse;
+import edu.harvard.iq.dataverse.DvObjectContainer;
+import edu.harvard.iq.dataverse.authorization.Permission;
+import edu.harvard.iq.dataverse.engine.command.AbstractCommand;
+import edu.harvard.iq.dataverse.engine.command.CommandContext;
+import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
+import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
+
+public class GetDataverseMetadataLanguageCommand extends AbstractCommand<Map<String, String>> {
```
pdurbin, line 15: I'm refreshing myself on how this feature even works by reading https://guides.dataverse.org/en/6.8/installation/config.html#allowing-the-language-used-for-dataset-metadata-to-be-specified

In this command we're doing a get. Is there a command or at least an API endpoint for doing a set? As of #7958 the docs above say that the metadata language can be configured per collection but is only possible via the UI? 🤔

akos326: You're right, there's no API endpoint currently to set a metadata language in a collection. It's possible via the UI and Dataverse has a setter method

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Done

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Calling it allowed is misleading, because if the user sets the metadata language for the collection, then it will be mandatory.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseMetadataLanguageCommand.java
pdurbin: Ok, thanks for confirming. Are you interested in expanding the scope of this pull request to include the setter method?

Also, it could be expanded, if you want, to include a getter for the :MetadataLanguages setting. Please see https://guides.dataverse.org/en/6.8/developers/api-design.html#exposing-settings

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
pdurbin: Thanks!

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
pdurbin: If it helps: https://guides.dataverse.org/en/6.8/contributor/documentation.html

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
qqmyers: Then it would be the only one allowed. I'm mostly concerned with this being singular when it can be several languages and that it sounds like it applies to the collection when it only applies to new datasets in the collection. Any name that addresses those would work.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseMetadataLanguageCommand.java
akos326: Currently the GET method is sufficient for my purposes, but adding a POST doesn't seem to be much work. I'll do it tomorrow.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Okay, I will document it, once we settled on a name for the endpoint.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Okay, I think I'm gonna go with allowedMetadataLanguages.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
qqmyers: I haven't investigated, but when :MetadataLanguages is set, I think datasets created via the API need to specify the metadataLanguage they use. So perhaps every test run after it's set is failing? If that's true, in addition to deleting the setting at the end of the test, the test may need to be annotated to not run in parallel with anything else (`@ResourceLock(GLOBAL_TEST_LOCK)` ?).

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
akos326: This part is pretty common in other test cases too. It would make sense to check for CREATED in all of them, but that's beyond the scope of this pull request, I think.

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Done!

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseMetadataLanguageCommand.java
akos326: Done, we have a PUT method now!

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
akos326: Why isn't it enough to delete it in the beggining and end of the test case?

stevenwinship: If the test case fails, the cleanup past the assert will not run. `@AfterAll` will run after the failure assuring the cleanup will happen

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
```diff
@@ -72,6 +72,7 @@
 import org.glassfish.jersey.media.multipart.FormDataBodyPart;
 import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
 import org.glassfish.jersey.media.multipart.FormDataParam;
+import org.springframework.security.access.method.P;
```
stevenwinship: This is unused and the only change to this class. Please remove the line

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
stevenwinship: To prevent tests from running in parallel:
`@Execution(SAME_THREAD)`
public class DataversesIT {

stevenwinship: Move setting clean up to `BeforeAll` and `AfterEach`

`@BeforeAll`
    public static void setUpClass() {
       ...
       UtilIT.deleteSetting(":MetadataLanguages");
}

'@AfterEach'
    public void afterEach() {
        UtilIT.deleteSetting(":MetadataLanguages");
    }


Remove `@ResourceLock(value = "")`

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -2731,4 +2734,40 @@ private String getSuperuserToken() {
         UtilIT.makeSuperUser(username);
         return adminApiToken;
     }
+    
+    @Test
+    @ResourceLock(value = "")
```
stevenwinship: Remove `@ResourceLock`(value = "")

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
qqmyers: > @Execution(SAME_THREAD)

FWIW - my concern is that any test (in any test class) that tries to create a dataset, can fail if a metadatalanguage is required and one is not present in the json. It looks like this annotation just assures tests in this class run serially, but it wouldn't stop other test classes from running at the same time.

stevenwinship: @qqmyers I was looking into that as we speak. My problem is the `@ResourceLock' is causing all the tests to fail for me.

stevenwinship: Ah.  error is "key must not be blank"

stevenwinship: This worked `@ResourceLock(value = "MetadataLanguages")`

qqmyers: What about @ResourceLock(GLOBAL_TEST_LOCK)  ? Using MetadataLanguages only stops it from running in parallel with other tests that lock on the same resource.

stevenwinship: What I'm testing now:

`@Execution(ExecutionMode.CONCURRENT)`
`@ResourceLock(value = "MetadataLanguages", mode = ResourceAccessMode.READ)`
public class DataversesIT {


`@Test`
`@ResourceLock(value = "MetadataLanguages", mode = ResourceAccessMode.READ_WRITE)`
public void testDataverseMetadataLanguage() {

stevenwinship: It might be easier to change the code so leaving the lang null or empty defaults it to the first one in the list from the setting. Or Mocking the setting (which isn't good since you are testing the setting)

stevenwinship: `@ResourceLock(value = "MetadataLanguages", mode = ResourceAccessMode.READ_WRITE)`
`@Isolated`
public class DataversesIT {

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
akos326: Done

src/main/java/edu/harvard/iq/dataverse/api/Dataverses.java
akos326: Done

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
akos326: Done

src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
```diff
@@ -48,18 +53,27 @@
...
     }
 
     @AfterAll
     public static void afterClass() {
         Response removeExcludeEmail = UtilIT.deleteSetting(SettingsServiceBean.Key.ExcludeEmailFromExport);
+        UtilIT.deleteSetting(SettingsServiceBean.Key.MetadataLanguages);
+    }
+
+    @AfterEach
+    public static void afterEach() {
```
stevenwinship: @akos326  Please remove "static" from afterEach

akos326: Done

ofahimIQSS:
Hello! Conflicts need to be resolved - please update. Thanks!

akos326:
Thank you @ofahimIQSS! I resolved the merge conflicts

PR #11859, opened 2025-10-01
Version summaries pagination

GPortas:

- Closes #

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -3138,8 +3138,15 @@ public class Datasets extends AbstractApiBean {
     @GET
     @AuthRequired
     @Path("{id}/versions/compareSummary")
-    public Response getCompareVersionsSummary(@Context ContainerRequestContext crc, @PathParam("id") String id,
-                                      @Context UriInfo uriInfo, @Context HttpHeaders headers) {
+    public Response getCompareVersionsSummary(@Context ContainerRequestContext crc, @PathParam("id") String id) {
+        return response(req -> {
+            try {
+                return ok(json(execCommand(new GetDatasetVersionSummariesCommand(req, findDatasetOrDie(id)))));
+            } catch (WrappedResponse wr) {
+                return wr.getResponse();
+            }
+        }, getRequestUser(crc));
+        /*
         try {
             Dataset dataset = findDatasetOrDie(id);
             User user = getRequestUser(crc);
@@ -3188,7 +3195,7 @@ public class Datasets extends AbstractApiBean {
             return ok(differenceSummaries);
         } catch (WrappedResponse wr) {
             return wr.getResponse();
-        }
+        }*/
     }
 
     private JsonObject getDeaccessionJson(DatasetVersion dv) {
--- a/src/main/java/edu/harvard/iq/dataverse/datasetversionsummaries/DatasetVersionSummary.java
+++ b/src/main/java/edu/harvard/iq/dataverse/datasetversionsummaries/DatasetVersionSummary.java
@@ -1,0 +1,72 @@
+package edu.harvard.iq.dataverse.datasetversionsummaries;
+
+import edu.harvard.iq.dataverse.DatasetVersion;
+import edu.harvard.iq.dataverse.DatasetVersion.VersionState;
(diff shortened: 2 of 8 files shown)


ChengShi-1:
Hi Guillermo, thanks a lot for these fixes! 💯 

I tested it locally and found something that may need attention.

For dataset:
1. The result is missing **contributors** field for dataset, the payload was like
```
 {
id: number
versionNumber: string
summary?: DatasetVersionSummary | DatasetVersionSummaryStringValues
contributors: string
publishedOn?: string
}
```
Now, it is
```
{

"id": 4,
"versionNumber": "3.1",
"publishedOn": "2025-10-20",
"summary": { }
}
```

2. It seems like it counts `"changedFileMetaData" : 2` forever. However, it happens to JSF as well, which always count `"changedFileMetaData" : 1`, no matter how many changes the file has.

For file summaries, `fileDifferenceSummary` field shows File: added, deleted.... but it is not showing File Categories and Tabular Tags, File Access and File Metadata there.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -3138,76 +3138,21 @@ public Response getCompareVersions(@Context ContainerRequestContext crc, @PathPa
     @GET
     @AuthRequired
     @Path("{id}/versions/compareSummary")
```
sekmiller, line 3140: Should we look into adding a command for compare versions above?

GPortas: Yes, and this applies to many other endpoints across the API. It’s not something we need to change separately, but whenever we modify an endpoint, we should take the opportunity to refactor it to use commands and follow the defined layered architecture.

GPortas:
Everything should be working now as expected, except for the pre-existing issue mentioned in #11921. Thanks, @ChengShi-1!

src/main/java/edu/harvard/iq/dataverse/FileMetadataVersionsHelper.java
```diff
@@ -191,7 +93,7 @@ private FileMetadata getPreviousFileMetadata(FileMetadata fileMetadata, FileMeta
     }
 
     //TODO: this could use some refactoring to cut down on the number of for loops!
-    private FileMetadata getPreviousFileMetadata(FileMetadata fileMetadata, DatasetVersion currentversion) {
+    public FileMetadata getPreviousFileMetadata(FileMetadata fileMetadata, DatasetVersion currentversion) {
```
sekmiller: Not sure why this was made public. the only usage is inside the helper

GPortas: Thanks — I forgot to make it private again. I had to set it to public during the multi-step refactor while extracting logic and responsibilities from this class.

FWIW - The API no longer uses this method; instead, it uses a more efficient JPA Criteria–based DataFileServiceBean method.

GPortas: I just made it private again

ChengShi-1:
> Hi Guillermo, thanks a lot for these fixes! 💯
> 
> I tested it locally and found something that may need attention.
> 
> For dataset:
> 
> 1. The result is missing **contributors** field for dataset, the payload was like
> 
> ```
>  {
> id: number
> versionNumber: string
> summary?: DatasetVersionSummary | DatasetVersionSummaryStringValues
> contributors: string
> publishedOn?: string
> }
> ```
> 
> Now, it is
> 
> ```
> {
> 
> "id": 4,
> "versionNumber": "3.1",
> "publishedOn": "2025-10-20",
> "summary": { }
> }
> ```
> 
> 2. It seems like it counts `"changedFileMetaData" : 2` forever. However, it happens to JSF as well, which always count `"changedFileMetaData" : 1`, no matter how many changes the file has.
> 
> For file summaries, `fileDifferenceSummary` field shows File: added, deleted.... but it is not showing File Categories and Tabular Tags, File Access and File Metadata there.

I tested dataset summaries and file summaries again, and they work well now. Thanks for resolving them!

ofahimIQSS:
looks good - going to perform regression in QA. Merging after CI completes

ofahimIQSS:
everythings looking good here- merging

Issue #9998, opened 2023-10-11
Migrate storage configuration to MPCONFIG (milestone D)


poikilotherm:
**Overview of the Feature Request**
As with our other JVM options, let's make it possible to configure storage using MicroProfile Config.

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
- Making it simpler to test the storage layer
- Make it aligned with the rest of the JVM options already enabled
- Make it easier to supply configuration in a container context

**What existing behavior do you want changed?**
Do not only accept system options, but enable using all config sources MPCONFIG provides for storage configuration.

**Any brand new behavior do you want to add to Dataverse?**
None

**Any open or closed issues related to this feature request?**
- #7000
- #9662 
- #6656

PR #11862, opened 2025-10-02
update docs to suggest using Docker in production

pdurbin:

Now that we have images tagged with version numbers, I think we're ready to suggest that people can try using Docker in production

None.

**Additional documentation**:


pdurbin:
During today's Containerization working group meeting I talked through this PR with @poikilotherm @beepsoft and @akio-sone. The recording is here: https://harvard.zoom.us/rec/share/qw5BTHsX132xEYBue7rKrPx6n7_3oL1J5AoOz40y9STlp_z_cuHM2qbTcOtWuaDG.34EEw9prY2HxDp9c

In short, I'm removing the scary language about never using these containers in production. @johannes-darms is using them, and others.

@poikilotherm rightly pointed out that not everything can be configured via MPConfig, such as storage, but we have an issue tracking this:

- #9998

There will always be something to improve. We have tagged versions now, at least, so people can pin their installations to a specific version. I think it's time.

I'm moving this to "ready for review". I'm happy to make changes, of course!

PR #11865, opened 2025-10-06
restored suppressing dataverse field

jo-pol:
See https://github.com/IQSS/dataverse/pull/11301
and https://github.com/IQSS/dataverse/pull/11700

logged the duration on a VM with almost 100K dataverses

findall: PT15.632S
getDataverseCount: PT-0.026S

Looks a good enough improment to me.

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -105,6 +105,8 @@ import jakarta.inject.Named;
 import jakarta.persistence.OptimisticLockException;
 
 import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
 import org.primefaces.event.FileUploadEvent;
 import org.primefaces.model.file.UploadedFile;
 
@@ -332,6 +334,7 @@ public class DatasetPage implements java.io.Serializable {
     private List<SelectItem> linkingDVSelectItems;
     private Dataverse linkingDataverse;
     private Dataverse selectedHostDataverse;
+    private boolean hasDataversesToChoose;
 
     public Dataverse getSelectedHostDataverse() {
         return selectedHostDataverse;
@@ -1779,6 +1782,11 @@ public class DatasetPage implements java.io.Serializable {
 
     public void setDataverseTemplates(List<Template> dataverseTemplates) {
         this.dataverseTemplates = dataverseTemplates;
+    }
+
+    public boolean isHasDataversesToChoose() {
+        this.hasDataversesToChoose = dataverseService.getDataverseCount() > 1;
+        return this.hasDataversesToChoose;
     }
 
     public Template getDefaultTemplate() {
--- a/src/main/webapp/dataset.xhtml
+++ b/src/main/webapp/dataset.xhtml
@@ -777,7 +777,7 @@
                     <!-- Create/Edit editMode -->
                     <ui:fragment rendered="#{DatasetPage.editMode == 'METADATA' or DatasetPage.editMode == 'CREATE'}">
                         <p:focus context="datasetForm"/>
-                        <div class="form-group">
+                        <div class="form-group" jsf:rendered="#{DatasetPage.hasDataversesToChoose and DatasetPage.editMode == 'CREATE'}">
                             <label jsf:for="#{DatasetPage.editMode == 'CREATE' ? 'selectHostDataverse_input' : 'select_HostDataverse_Static'}" class="col-md-3 control-label">
                                 #{bundle.hostDataverse}
                                 <span class="glyphicon glyphicon-question-sign tooltip-icon"
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseConverter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseConverter.java
@@ -13,19 +13,34 @@ import jakarta.faces.context.FacesContext;
 import jakarta.faces.convert.Converter;
 import jakarta.faces.convert.FacesConverter;
 
+import java.util.logging.Logger;
+
 /**
  *
  * @author skraffmiller
  */
 @FacesConverter("dataverseConverter")
 public class DataverseConverter implements Converter {
+    private static final Logger logger = Logger.getLogger(DatasetPage.class.getCanonicalName());
+
     
     //@EJB
     DataverseServiceBean dataverseService = CDI.current().select(DataverseServiceBean.class).get();
 
     @Override
     public Object getAsObject(FacesContext facesContext, UIComponent component, String submittedValue) {
-        return dataverseService.find(new Long(submittedValue));
+        if (submittedValue == null || !submittedValue.matches("[0-9]+")) {
+            logger.fine("Submitted value is not a host dataverse number but: " + submittedValue);
+            return CDI.current().select(DatasetPage.class).get().getSelectedHostDataverse();
+        }
+        else {
+            try {
+                return dataverseService.find(Long.parseLong(submittedValue));
+            } catch (NumberFormatException e) {
+                logger.warning("Submitted value is out of range for a Long: " + submittedValue);
+                return CDI.current().select(DatasetPage.class).get().getSelectedHostDataverse();
+            }
+        }
         //return dataverseService.findByAlias(submittedValue);
     }
 
--- a/src/main/java/edu/harvard/iq/dataverse/DataverseConverter.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DataverseConverter.java
@@ -21,7 +21,7 @@ import java.util.logging.Logger;
  */
 @FacesConverter("dataverseConverter")
 public class DataverseConverter implements Converter {
-    private static final Logger logger = Logger.getLogger(DatasetPage.class.getCanonicalName());
+    private static final Logger logger = Logger.getLogger(DataverseConverter.class.getCanonicalName());
 
     
     //@EJB


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -105,6 +105,8 @@
 import jakarta.persistence.OptimisticLockException;
 
 import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
```
qqmyers: These aren't used any more.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1781,6 +1784,11 @@ public void setDataverseTemplates(List<Template> dataverseTemplates) {
         this.dataverseTemplates = dataverseTemplates;
     }
 
+    public boolean isHasDataversesToChoose() {
+        this.hasDataversesToChoose = dataverseService.getDataverseCount() > 1;
```
qqmyers: FWIW (and so we don't lose it):[ the original PR has a suggestion ](https://github.com/IQSS/dataverse/pull/11301/files#r2399004092) that one could query ~ "SELECT 1 FROM Dataverse d limit 2" which would be more efficient that getting a count and still address the issue (is there more than one dataverse so the user can actually make a choice). I'll also note that this form might be easier to extend to look for dataverses where a user has permission to add a dataset in the future. That said, just getting the count is orders of magnitude faster than the problematic query in the original PR, so I think this is fine unless @jo-pol wants to go further.

src/main/webapp/dataset.xhtml
```diff
@@ -777,7 +777,7 @@
                     <!-- Create/Edit editMode -->
                     <ui:fragment rendered="#{DatasetPage.editMode == 'METADATA' or DatasetPage.editMode == 'CREATE'}">
                         <p:focus context="datasetForm"/>
-                        <div class="form-group">
+                        <div class="form-group" jsf:rendered="#{DatasetPage.hasDataversesToChoose and DatasetPage.editMode == 'CREATE'}">
```
qqmyers, line 780: FWIW: Some additional small performance tweaks that could be made:

- I think it's true that JSF will evaluate left to right, so if DatasetPage.editMode == 'CREATE' were first and was false, the hasDataverseToChoose query would not get made. 
- JSF can call methods multiple times (up to 7!) so caching the boolean answer to hasMultipleDataversesToChoose in the DatasetPage would also improve performance (probably more important if/when there's also a permission check to see if the user has permissions to add datasets in more than one dataverse.).

qqmyers (changes requested):
This PR updates #11301 and removes the significant performance issue there. I think that, to move forward, this just needs to have a release note, and to get rid of the unused imports. I've made a couple notes about other possible minor performance improvements - I think unless @jo-pol wants to make those or someone on the team thinks they're needed, they probably aren't needed (I'll poll people at standup/triage tomorrow).

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
jo-pol: > to look for dataverses where a user has permission to add a dataset in the future

That thought crossed my mind too. We now still might show the field while there is nothing to choose.

src/main/webapp/dataset.xhtml
jo-pol: Line 778 takes care of the first bullet. A break point confirmed `hasMultipleDataversesToChoose` is called only when editing a dataset, not when showing. It is indeed called some seven times for a new dataset.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1781,6 +1782,14 @@ public void setDataverseTemplates(List<Template> dataverseTemplates) {
         this.dataverseTemplates = dataverseTemplates;
     }
 
+    public boolean isHasDataversesToChoose() {
+        // TODO we actually need to look for dataverses where a user has permission to add a dataset
```
jo-pol: Tried to create a new dataset with a user that had no permission to any of the 100K numbered test dataverses. It took a few seconds when searching with 2 digits. 
So a variant of `DataverseServiceBean.filterDataversesForHosting`  (on [github](https://github.com/DANS-KNAW/dataverse/blob/34338774bd4331d98a3538bdd82d6841e703dc0b/src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java#L550-L576)) without a pattern for the query and exiting after one (or two?) successfull authourisation checks would take way too long for users with permission for too few dataverses.


A more efficient query might be something like
```
SELECT dv
FROM roleassignment ra
JOIN dataverse      dv
  on dv.id = ra.definitionpoint_id
where ra.assigneeidentifier in ('@user001', ':authenticated-users');
```
However, `RoleAssignment.findAssigneesWithPermissionOnDvObject` (on [github](https://github.com/DANS-KNAW/dataverse/blob/7a3964b1aff229a9c73cb56e0cc1d4a254023eae/src/main/java/edu/harvard/iq/dataverse/RoleAssignment.java#L61-L79)) is dazzling me.

ofahimIQSS:
discussed during triage - Jim will follow up

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
qqmyers: Yeah - that query is more complex because it has to look for filedownloads configured on the dataset owning a file, etc. I haven't fully thought it through, but this custom code from QDR might be a reasonable model: https://github.com/QualitativeDataRepository/dataverse/blob/b9c0bdc374518cc499eba564e8ad04ec059bc1ff/src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java#L1014-L1051 - it checks for superuser first, and if, the user isn't one, gets the list of roles that include the relevant permission (AddDataset in this case) and looks for a case where the user has that role on something. I think you could further constrain by looking for the roleassignment's dvobject having dtype='Dataverse'. Setting the limit to 2 would then let you see if that user has a choice somewhere. I think this is reasonably performant, but I haven't tested on a very large database (as with the case in this PR, it was orders of magnitude faster than the code I replaced).

Do you want to try that? And maybe cache the result so it isn't called 7 times and call this PR ready to move forward?

src/main/webapp/dataset.xhtml
qqmyers: 778 limits it to showing only when the mode is CREATE (or METADATA) so I guess changing 780 would only help in the METADATA case. Definitely minor.

jo-pol: Do you mean to allways suppress the host dataverse on editing a dataset?

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
jo-pol: Looking at the dazzling query PermissionServiceBean.LIST_ALL_DATAVERSES_USER_HAS_PERMISSION I might have to do something with groups too. That might be more work than @janvanmansum baragained for.

Anyway, even on a simplified solution I get:

```
jakarta.ejb.EJBException: An exception occurred while creating a query in EntityManager: 
Exception Description: Syntax error parsing [SELECT dv FROM roleassignment ra JOIN dataverse      dv   ON dv.id = ra.definitionpoint_id WHERE ra.assigneeidentifier = '@user001' OR ra.assigneeidentifier = ':authenticated-users' limit 2]. 
[90, 182] The expression is not a valid conditional expression.
```

The query works in psql.

src/main/webapp/dataset.xhtml
jo-pol: Ran a test but neither the template nor the host datavese fields are shown when editing metadata while the same user has these fields when creating a new dataset.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -1781,6 +1782,14 @@ public void setDataverseTemplates(List<Template> dataverseTemplates) {
...
+        if (this.hasDataversesToChoose == null) {
+            this.hasDataversesToChoose = dataverseService.getDataverseCount() > 1;
+        }
```
jo-pol, line 1800: This way the method may still be called 7 times, but the query just once per (new) dataset.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
jo-pol: The simplified attempt: https://github.com/DANS-KNAW-jp/dataverse/commit/4fcf8e1fdf3b32c98d8f3ff034f60a36c6ebac74

src/main/webapp/dataset.xhtml
qqmyers: My suggestion wasn't to change any behavior, just the efficiency. When you get to line 780, which means you're in CREATE or METADATA mode because of line 778, the current code will first call the hasDataversesToChoose method and, if it's true, will then check if you're in CREATE mode. Changing the order (`{DatasetPage.editMode == 'CREATE' and DatasetPage.hasDataversesToChoose}`) would mean it would first check that you're in CREATE mode (which is cheap), and only if it's true would the more expensive hasDataversesToChoose method get called. 

As you pointed out, since line 778 already limits this to two modes, the only savings would be when the user is in METADATA mode - hasDataversesToChoose wouldn't get called in that case. So - it's more of a minor issue than if 780 could get called for EDIT and other modes.

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
jo-pol: Hacked PermissionServiceBean.findPermittedCollections with "limit 2" for when it should have returned 50K instances. That reacted quickly. So I need 

    permissionService.findPermittedCollections(req, user, Set.of(Permission.CreateDataset))

with a variant of the query in the simplified attempt.

qqmyers: @jo-pol  - Not sure where my comment went - so ~retyping. 
The main issue with the 'simplified attempt' is that it's a native sql query and not JPA. That could be addressed by using createNativeQuery() or using JPA. I suggest the following code (have not tested). It handles all the groups the user might be in and limits which roles are checked, along with keeping your check that the definitionpoints are Dataverses.

I'm not sure if it is more efficient that the findPermittedCollections - certainly looks simpler. The code below does break finding the roles and groups out into separate queries (less efficient?), but it may be more efficient to be checking the role ids (which hopefully are indexed) rather than doing the bit-level or on the permissions. I'm not sure how well postgres can optimize in either case.

```
        Set<RoleAssignee> ras = new HashSet<>(groupService.groupsFor(req));
        ras.add(user);

        List<String> raIds = ras.stream().map(roas -> roas.getIdentifier()).collect(Collectors.toList());

        List<Long> roleIds = new ArrayList<>();
        
        for (DataverseRole role : roleService.findAll()) {
            if (role.permissions().contains(Permission.AddDataset)) {
                roleIds.add(role.getId());
            }
        }
        // Just check for more than one matching record
        // More efficient than counting all records
        try {
            return em.createQuery(
                "SELECT ra.id FROM RoleAssignment ra " +
                    "JOIN Dataverse dv ON dv.id = ra.definitionPoint.id " +
                    "WHERE ra.assigneeIdentifier IN :assigneeIdentifiers " +
                    "AND ra.role.id IN :roleIds",
            Long.class)
            .setParameter("assigneeIdentifiers", raIds)
            .setParameter("roleIds", roleIds)
            .setMaxResults(2)
            .getResultList().size() > 1;

        } catch (NoResultException e) {
            return false;
        }
```

jo-pol: @qqmyers 
> That could be addressed by using createNativeQuery()

That is how the permissionService executes the query anyway and how I reused it with `limit 2` added. Reusing existing logic reduces maintenance risks IMHO. Updated the cover message with how I tested.

doc/release-notes/11865-restored-suppress-dataverse-field
```diff
@@ -0,0 +1,9 @@
+### Restored suppression of the host dataverse field
+
+When creating/editing a dataset, the _host dataverse_ field is not shown when there is no other dataverse than the root dataverse.
```
qqmyers: when the user can only add datasets to one collection.

src/main/java/edu/harvard/iq/dataverse/DataverseServiceBean.java
```diff
@@ -1276,6 +1276,16 @@ public long getDataverseCount() {
         return em.createNamedQuery("Dataverse.countAll", Long.class).getSingleResult();
     }
 
+    public boolean getHasRolesOnMultipleDataverses(String assignee) {
```
qqmyers: No longer used

doc/release-notes/11865-restored-suppress-dataverse-field
```diff
@@ -0,0 +1,9 @@
+### Restored suppression of the host dataverse field
```
qqmyers: Since the release note is what users see, I'd suggest changing the title to "Suppression of the Host Dataverse field" and removing the second (for developers) and third (no longer true) paragraphs.

src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java
```diff
@@ -953,96 +953,90 @@ public List<Dataverse> findPermittedCollections(DataverseRequest request, Authen
...
-                } else {
-                    sqlCode = LIST_ALL_DATAVERSES_SUPERUSER_HAS_PERMISSION.concat(WHERE).concat(SEARCH_PARAMS);
+            else {
+                Query query = em.createNativeQuery(sqlCode.concat(AND).concat(SEARCH_PARAMS), Dataverse.class);
+                setSearchParamValues(searchTerm, query);
+                return query.getResultList();
+            }
+        }
+        return null;
+    }
```
jo-pol, line 971: I'm puzzled how to assemble the URL for testing linkingDataverses with the API.

jo-pol: Weird, Intellij does not blame me for the code inside getIpRange and setSearchParamValues but github does [blame](https://github.com/IQSS/dataverse/blame/737659031a78dde963405404eebaf5f576ad624e/src/main/java/edu/harvard/iq/dataverse/PermissionServiceBean.java) me.

jo-pol: The following command looks more like the intellij result

     git blame -w -M -C --minimal -L 954,1040 PermissionServiceBean.java

qqmyers:
@jo-pol - are you done with changes? If so, we can remove your assignment and let it go into QA.

jo-pol:
> @jo-pol - are you done with changes? If so, we can remove your assignment and let it go into QA.

I am done with the changes and unassigned myself. I'd recommend to test linkingDataverses with the API. Don't have a clue to do it myself.

ofahimIQSS:
Hey @jo-pol, I wasn't able to reproduce the issue on demo where we have trouble editing the host DV - could you please try and reproduce the issue in demo (running v6.8 at the moment) -

Thanks!

jo-pol:
> could you please try and reproduce the issue in demo (running v6.8 at the moment) -

The following scenario still blocks the deposit page. The top bar and footer are accessible, the deposit form becomes faint and no new draft appears when selecting "_My Data_".

- In the user interface click Add Data → New Dataset
- The cursor is placed in the “Host Dataverse” field.
- Quickly type: space, back-space, tab
- Fill in the required fields and click Save.

ofahimIQSS:
Thank you! 
I was able to reproduce the issue - going to try the fix now, will take it from here.

pdurbin:
This seems to work. First, I ran `mvn test -Dtest=InReviewWorkflowIT` to create an author user and a curator user. A collection that both can create datasets in was also created.

Then I made it so no one can create anything in the root collection without explicit permission:

Screenshot 2025-11-06 at 3 40 31 PM

The author doesn't see the "host dataverse" field since they can only create datasets in the one collection:

Screenshot 2025-11-06 at 3 39 03 PM

For the curator, I created an extra collection and gave them access to create datasets so they see the "host dataverse" field:

Screenshot 2025-11-06 at 3 40 09 PM

Screenshot 2025-11-06 at 3 41 44 PM

Merging.

PR #11895, opened 2025-10-13
DD-2023 Missing body entity with Pre-Workflow POST of AuthorizedExter…

aliassheikh:
DD-2023 Missing body entity with Pre-Workflow POST of AuthorizedExternalStep

We are not able to publish a dataset created on WEB-UI on  the DANS DataStations using IQSS release 6.7.1.
When a Publish action is started, the dataset is become locked with a Workflow lock.

DANS microservices should get an http message from IQSS `AuthorizedExternalStepwith` with dataset’s information but the message is empty and caused process failure.

https://drivenbydata.atlassian.net/browse/DD-2023 

- Closes #

The added entity-body in source file `AuthorizedExternalStep.java` is comparable with `HttpSendReceiveClientStep.java` in the same package.

Add a unittest 
`src/test/java/edu/harvard/iq/dataverse/workflow/internalspi/AuthorizedExternalStepTest.java`

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/workflow/internalspi/AuthorizedExternalStep.java
```diff
@@ -1,6 +1,7 @@
 package edu.harvard.iq.dataverse.workflow.internalspi;
```
pdurbin, line 1: I'm just leaving this comment at the top.

@aliassheikh do you think this pull request deserves a [release note snippet](https://guides.dataverse.org/en/6.8/developers/version-control.html#writing-release-note-snippets)?

Also, the current docs on Workflows are at https://guides.dataverse.org/en/6.8/developers/workflows.html . Should this page be updated as well?

src/main/java/edu/harvard/iq/dataverse/workflow/internalspi/AuthorizedExternalStep.java
```diff
@@ -1,6 +1,7 @@
...
 
 import edu.harvard.iq.dataverse.util.BundleUtil;
```
pdurbin, line 3: Another comment near the top.

@aliassheikh it's awesome that you added a unit test but can you please also update the "how to test" part of the description for how this should be tested when in QA? Thanks!

src/main/java/edu/harvard/iq/dataverse/workflow/internalspi/AuthorizedExternalStep.java
aliassheikh: @pdurbin, 
No, I don't think it's necessary.
It worked for version 6.7, and the problem was caused by code refactoring.
The bug/code is fixed by this fix.

src/main/java/edu/harvard/iq/dataverse/workflow/internalspi/AuthorizedExternalStep.java
aliassheikh: @pdurbin,
I'm thinking about an answer.
In short: it is about a https communication intends as a pre-publish Workflow [dataverse-workflow-help](https://guides.dataverse.org/en/6.7/developers/workflows.html)
Perhaps QA has test-scenarios like this already.

pdurbin (approved):
@aliassheikh I'm approving this but please do communicate with QA about how to test. Thanks!

PR #11898, opened 2025-10-14
adding indexes and other sql for performance improvements with deleting Datasets and DatasetVersions

### LINKED ISSUE #10765: dataverse 6.3 fail to create table
Run python install.py in a fresh environment.

* When does this issue occur?
Whenever you access dataverse.

The home page shows 404 error.

* What happens?
install.py fails with the following error

```bash
Command start-domain executed successfully.
/tmp/dvinstall
Payara setup complete
Fri Aug  9 14:34:57 UTC 2024

Installing additional configuration files (Jhove)...
done.
Deploying the application (dataverse.war)

PER01003: Deployment encountered SQL Exceptions:
        PER01000: Got SQLException executing statement "ALTER TABLE fileaccessrequests ADD CONSTRAINT FK_fileaccessrequests_DATAFILE_ID FOREIGN KEY (DATAFILE_ID) REFERENCES DVOBJECT (ID)": org.postgresql.util.PSQLException: ERROR: constraint "fk_fileaccessrequests_datafile_id" for relation "fileaccessrequests" already exists
```

        "version": "6.3",
        "build": "1607-8c99a74"

### LINKED ISSUE #11828: Performance degradation in reharvesting
Investigating this harvested performance issue in IQSS prod.: 
Harvesting a new dataset is generally fast (some fraction of a sec./dataset); 
Re-harvesting that same dataset the performance is _much_ (10s of times) worse. 
This must somehow be a function of the overall database size; since it's not observable w/ small test databases. 
It is reproducible on the perf. system w/ a clone of the prod. database (even slower there). 
It does not appear to be a function of the number of datasets in the collection (observable in a collection with very few harvested datasets).

I made a [PR](https://github.com/IQSS/dataverse/pull/10836) a year ago specifically to make re-harvesting cheaper. It is absolutely possible that something I did there backfired and made things worse. 
One extra confusing detail though is that it also takes forever to delete harvested datasets when deleting a client (appears to take exactly as long to delete as it does to re-harvest/update these datasets). Harvested datasets are removed via a cascade in that scenario; I did not touch that part in 10836.  

I don't know yet for sure if the issue is in the database update or delete,  or in solr reindexing. But, FWIW, when deleting a client, the dataset cards in the collection disappear almost instantly, and then it takes forever for the dataset objects to disappear from the db. 

I'm looking into this since this is an effective blocker for a few large remote collections that need to be re-harvested. 

(to be clear, I have no solid evidence yet of this being unique to harvested objects; it's just that with harvesting you often have real life cases where you need to update large numbers - hundreds or even thousands - of datasets all at once)

stevenwinship:
**What this PR does / why we need it**: Slow performance deleting Datasets needs to be addressed

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11828

- Closes #11828
- Closes #10765

**Suggestions on how to test this**: test with new db and upgrading existing db
Edit from L.A.: The performance aspect can be easily seen with the prod. db clone on qa.dataverse.org. 
The indexes that the flyway script creates were already put in place there, but I have removed them, so the db can be tested in the "before" state. Note that unless the PR goes through QA before the next db sync, on Oct. 26, the indexes will need to be removed again (edit: the db restore script now does this automatically; we'll just need to remember to remove the lines after this PR is merged). 
The easiest way to see the gains from the indexes is to search on the following tables, before and after this build is deployed: `guestbookresponse` by either `datasetversion_id` or `authenticateduser_id` (the improvements are really dramatic on this table); and `datasetmetrics` by `dataset_id`. 
To test the performance of actually deleting datasets, I suggest running the destroy API on any of the following harvested datasets, some before and some after: 
```
doi:10.5683/SP3/BD9AMB
doi:10.5683/SP3/NIQKF5
doi:10.5683/SP3/Z4Y0LK
doi:10.5683/SP3/3N6JVZ
doi:10.5683/SP3/PTHRS7
doi:10.5683/SP3/REQLIL
```
(i.e., something like ```time curl -H "X-Dataverse-key: ..." -X DELETE "http://localhost:8080/api/datasets/:persistentId/destroy/?persistentId=doi:10.5683/SP3/Z4Y0LK```)

Each of the above has exactly 100 files in it. I suggest using harvested datasets, since there are no physical files to delete, so the cost observed is purely for deleting the datasets and datafiles from the database and solr. You should be able to see some very noticeable improvement. (the datasets above have been selected in the current prod. db snapshot; I will need to double-check they are still there and still have the same numbers of files if the PR is being tested after Oct. 26 - edit: confirmed, as of Nov. 17)

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11850-big-data-admin-guide.md
+++ b/doc/release-notes/11850-big-data-admin-guide.md
@@ -1,0 +1,3 @@
+### Big Data Admin Section
+
+- A new section - Scaling Dataverse with Data Size - has been added to the Admin Guide. It is intended to help administrators configure Dataverse appropriately to handle larger amounts of data.
--- a/doc/sphinx-guides/source/admin/big-data-administration.rst
+++ b/doc/sphinx-guides/source/admin/big-data-administration.rst
@@ -1,0 +1,321 @@
+Scaling Dataverse with Data Size
+================================
+
+This section is intended to help administrators configure Dataverse appropriately to handle larger amounts of data.
+
+Scaling is a complex subject: there are many options available in Dataverse that can improve performance with larger scale data, some of which
+work differently than Dataverse's default configuration, potentially requiring user education, and some which can require additional expertise to manage.
+
+In general, there are three dimensions in which Dataverse can scale:
+  
+1. **Storage size of individual files and aggregate storage size**
+2. **Number of files per dataset**
+3. **Number of datasets**
+
+
+.. contents:: |toctitle|
+        :local:
+
+.. _choose-store:
+
+Storage: Choosing the Right Store
+---------------------------------
+
+The main issues for handling larger files and larger aggregate data size relate to the performance of the storage used and how involved the Dataverse server is in the data transfer. 
+With appropriate configuration, Dataverse can support file sizes and aggregate dataset sizes into the terabyte scale and beyond.
+
+The primary choice in Dataverse related to storage is which types of "store" (also called "storage driver") to use:
+
+.. _file-stores:
+
+File Stores
+~~~~~~~~~~~
+
+The default storage option in Dataverse uses the local file system. When files are transferred to Dataverse, they are first stored in a
+temporary location on the Dataverse server. Any zip files uploaded are unzipped to create multiple individual file entries. Once an upload is completed, 
+Dataverse copies the files to permanent storage. Dataverse also takes advantage of the file being local to inspect its bytes to determine its 
+MIME type, and, for tabular data, to ":doc:`ingest </user/tabulardataingest/index>`" it - extracting metadata about the variables used in the file and creating a tab-separated values (TSV)
+version of the file.
+
+Benefits: 
+
+- This option requires no external services and can potentially handle files into the gigabyte (GB) size range. For smaller institutions,
+  and in disciplines where datasets do not have more than a few hundred files and files are not too large, this can be the simplest option.
+- Unzipping of zip archives can be simpler for users than having to upload many individual files and was, at one time, the only way to 
+  preserve file path names when uploading. In addition, some of the unzipped files might be in a format that can be previewed or otherwise acted upon - see :ref:`file-handling` in the User Guide.
+
+Challenges: In general, file storage is not a good option for larger data sizes - both in terms of file size and number of files. Contributing factors include:
+ 
+- Because temporary storage is used, transfers will temporarily use several times as much space as the final transfer. Unzipping also increases the final storage size of a dataset.
+- Because all uploads use the same temporary storage, temporary storage must be large enough to handle multiple users uploading data.
+- Each file is uploaded as a single HTTP request, which can cause long transfer times which, in turn, can trigger timeout errors in Dataverse or any proxy or load balancer in front of Dataverse.
+- Uploading many files at once can trigger any rate limiter in front of the Dataverse server (i.e. used to throttle AI bots) resulting in failures.
+- Because transfers (both uploads and downloads) are handled by the Dataverse server, they add to server processing load which can affect overall performance.
+- Cost: local file storage must be provisioned in advance based on anticipated demand. It can involve up-front costs (for a local disk), or, when procured from a 
+  cloud provider, is likely to be more expensive than object storage from that provider (see below).
+
+.. _s3-stores:
+
+S3 Stores: Object Storage via S3
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A more scalable option for storage is to use an object store with transfers managed using the Simple Storage Service (S3) protocol. S3-compatible storage can 
+easily be bought (rented) from major cloud providers, but may also be available from institutional clouds. It is also possible to run open source software to provide
+S3 storage over a local file system (making it possible to enjoy the advantages discussed below while still leveraging local file storage). 
+
+While S3 Stores can be configured to handle uploads and downloads as with file storage (with zip files being unzipped, but having many of the same challenges in terms of temporary storage and server load as discussed above) 
+they can also be configured to use "direct" upload and download. In this configuration, the actual transfer of file bytes is between the user's local machine and
+the S3 store. In this configuration, files are never stored on the Dataverse server. Dataverse does not attempt to unzip zip files, and they are stored as a single file in the dataset.
+
+Benefits: S3 offers several advantages over file storage:
+ 
+- Scalability: S3 is designed to handle large amounts of data. It can handle individual files up to several TB in size. 
+  Because S3 supports breaking files into pieces, Dataverse can transfer a file in pieces (several in parallel, potentially thousands of pieces per file) making transfers faster
+  and more robust (a failure requires only resending the failed piece). It may also be the case that users will have a faster network connection to the S3 store 
+  (e.g. in a commercial cloud or High Performance Computing center) than they do to the Dataverse server, reducing transfer time.
+- High Availability: S3 provides redundancy beyond what is available with a local file system (valuable for preservation, potentially reducing the need to perform data integrity checks).
+
+Challenges:
+
+- One additional step that is required to enable direct uploads via a Dataverse installation and for direct download to work with previewers and direct upload to work with DVWebloader (:ref:`folder-upload`) is to allow cross site (CORS) requests on your S3 store.
+- Cost: S3 offers a pricing model that allows you to pay for the storage and transfer of data based on current usage (versus long term demand) but commercial 
+  providers charge more per TB than the equivalent cost of a local disk (though commercial S3 storage is cheaper than commercial file storage).
+  There can also be egress and other charges. Overall, S3 storage is generally more expensive than local file storage but cheaper than cloud file storage.
+  Running a local S3 storage or leveraging an institutional service can further reduce costs.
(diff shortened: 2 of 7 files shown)


landreev:
I'm leaning towards spending more time on investigating ways to improve the performance of dataset deletes. Which can still be fairly described as unacceptable. 
While it is true that an empty draft dataset can be deleted in 2-4 sec. (which is not ideal, but somewhat acceptable), it gets _much_ worse with datasets with many files. Now that the performance overhead on deleting a DatasetVersion has been eliminated, the time it takes to delete a dataset appears to be a **perfectly linear function** of the number of files - something like 1+ sec. per file (!). It appears (though needs to be confirmed) that this 1+ sec. comes simply from the time it takes to delete a single dvobject row.

landreev:
In the context of harvesting specifically (since that was the original reason for opening the issue):
Some harvested datasets need to be deleted as part of regular harvesting - the datasets that for whatever reason are no longer available at the source repository. With the above-mentioned math, 1+ sec. per file, clearly observed. 
With the performance issue on datasetversion addressed, I was able to re-harvest our largest harvested collections. With the largest one, Borealis, where ~20,000 datasets needed to be updated, the whole operation took 10.5 hrs. However, a disproportionate part of that time - 2.5 hrs - was spent deleting the 407 datasets that had to be dropped (close to 9,000 files between them). 

It is likely that the number of datasets that need to be deleted during routine harvesting will always be relatively low. This math however makes a harvested collection like Borealis extremely time-consuming to delete in full (with close to ~500,000 files in it, it would take about a week at this rate). Which defies the idea that harvested content was supposed to be disposable, easy to erase and repopulate from scratch.

qqmyers:
The DestroyDatasetCommand does do a loop through all files and runs a DeleteDataFileCommand for each - that can't be efficient.

landreev:
@qqmyers That may be true. However, if you look at the harvester code, it does not in fact rely on calling the DeleteDataFileCommand for harvested files - it just deletes them with `em.remove()` (_"unceremonioiusly"_, according to my 6 year old comment there). And it still takes 1+ sec. per file. It. really does appear that this is what it takes to delete a single row from the dvobject table in the prod. db, outside of Dataverse. This is observable, both in the prod. db and on qa, by inserting empty unreferenced rows with just the minimum of non-nullable fields, and deleting them, via psql.
This does take a 1+ sec. on dvobject; deleting a similar row from datafile (~ same size table) takes a minor fraction of a sec. 
I haven't been able to find anything as stupid (and therefore easy to fix) as the missing important index that was affecting deletes on datasetversion. But I want to explore some other leads (including what I posted in dv-tech the other day).

landreev:
> ... [harvester] does not in fact rely on calling the DeleteDataFileCommand for harvested files

I am struggling however trying to remember the reason it still bothers calling DestroyDatasetCommand on the remaining file-less dataset; rather than just removing the whole thing with `em.remove()`... Because it just doesn't look like anything DestroyDataset does applies to harvested datasets...

landreev:
P.S. I am wondering if deleting harvested datasets with a single cascaded remove would indeed speed it up.

stevenwinship:
Another way to speed up the delete (or make it look like it's faster) is to run the file deletes in a separate thread.

landreev:
Deleting the files in separate threads, i.e. asynchronously would be something to consider. 
Simply spawning off background threads for this may be a bit dangerous. (In the context of deleting a large number of datasets with a larger number of files in them, all these threads may just overwhelm and crash the server - ?). But it should be solvable. For example, a dataset delete may simply unlink the datafiles and filemetadatas from the dataset and datasetversions respectively. And then we could have an asynchronous process that will be deleting all such unlinked files sequentially, taking as long as needed and only whenever Payara has spare cycles. 

... But I'm still hoping that this can be sped up for real. It just doesn't feel right, for a delete of a single row to cost more than a second.

landreev:
OK, I rearranged things in the database such that it now costs 0.01 sec. to delete an empty row from dvobject, and 0.03 sec. to delete one from datafile. 
This was achieved at the expense of making a delete from the dataset table cost a hair under 1 sec. 
This would still seriously speed up deleting real life datasets, since there are generally many more files than datasets. 

... but all this makes me think that there is still something un-indexed that's responsible for most of this premium on just 1 of the 4 tables involved. 
Will add more details next week.

landreev:
Ouch. It was expensive to delete rows from the table simply because we didn't have an index on dataset_id in DATASETMETRICS. I don't know how it took me this long to see it. ☹️ 

But redefining a few foreign key constraints can speed it up just a little bit more.

landreev:
I said earlier

> ... redefining a few foreign key constraints can speed it up just a little bit more.

To explain what I was on about:
 
We have 39 foreign key constraints defined on the 3 sub-tables of DvObject - dataverse_id (21), dataset_id (9) and datafile_id (9). (I am not counting the instances where an entity actually has to reference dvobject_id, i.e., _any_ DvObject).  Every one of them is defined on the common DvObject however, like this: 
```
grep -i fk_metric_dataverse_id ~/domain1/generated/ejb/dataverse-6.8/dataverse-6.8_VDCNet-ejbPU_createDDL.jdbc
ALTER TABLE METRIC ADD CONSTRAINT FK_METRIC_dataverse_id FOREIGN KEY (dataverse_id) REFERENCES DVOBJECT (ID)
```
(and not `... DATAVERSE (ID)`, that is)
From what I can gather, this appears to be a fairly arbitrary implementation choice in the jpa. 
As a result, every time you delete a row from the DvObject table, postgres has to check these 39 places where the id may be referenced. Even with the indexes now in place in all the larger tables, this results in measurable overhead. On the dataverse side, this performance premium has to be paid when deleting _any_ DvObject. Note that, in our case, the fact that we forgot to define an index on dataset_id in DatasetMetrics was causing DataFile deletes to be expensive as well. 

Redefining these constraints to point directly to the specific sub-tables spreads this expense across the 3 of them, so that you "only pay for what you need" when deleting a specific sub-entity. I have tried it on qa.dataverse.org (see the script attached) and it did not cause any immediate issues in Dataverse. 

[constraints_redefined.sql](https://github.com/user-attachments/files/23002160/constraints_redefined.sql)

While I would consider trying this in our prod. (especially if we have to delete a large collection with 100Ks of files in it), I'm a bit hesitant to include the above in a released flyway script. The savings, even for us with our giant database, are on a much smaller scales than what's already achieved with the 2 extra indexes. Plus there are very few real life instances where this performance degradation would even be an issue in the first place. There is however a very real possibility that such a flyway script will fail on some instances, resulting in deployment failures. This is because of another side effect of how these constraints have been defined so far (on top of the performance premium): since there are no foreign key constraints on the sub-tables, it is possible to delete one of them, leaving the parent dvobject behind. For example, in our prod. database there are currently a few dvobjects with the `dtype=Dataverse`, without any corresponding rows in the Dataverse table. Which in turn made the following constraint from the script above to fail, requiring manual cleanup:
```
ALTER TABLE DATAVERSECONTACT ADD CONSTRAINT FK_DATAVERSECONTACT_dataverse_id FOREIGN KEY (dataverse_id) REFERENCES DATAVERSE (ID);
```

I don't feel like subjecting other instances to having to resolve such potential issues, especially if they don't have much to gain as a result.

landreev:
All in all, I am really happy with the performance gains achieved just by adding the 2 indexes in the flyway script here. The speedup is an order of magnitude or better, when deleting either a DatasetVersion, or a Dataset and/or DataFile. 
The massive re-harvesting of Borealis that I mentioned [earlier](https://github.com/IQSS/dataverse/pull/11898#issuecomment-3411559523) was done overnight, Sun.-Mon. last week, Oct. 12. Meaning a large portion of the collection was still in the un-updated state in the prod. db clone on qa up until last night. So I re-ran a partial harvest of it on qa over the weekend, and was able to observe that the deletes were now as fast as updates. 

Further improvements can likely be made, following the leads discussed earlier. For ex., there doesn't seem to be any need to run a real DestroyDatasetCommand when deleting a harvested dataset. However, I prefer to experiment with such things in a harvesting-specific PR that I'll be making soon. 

So, I will be moving this back into ready for review shortly.

src/main/resources/db/migration/V6.8.0.1.sql
```diff
@@ -0,0 +1,12 @@
+CREATE INDEX IF NOT EXISTS  INDEX_GUESTBOOKRESPONSE_datasetversion_id ON GUESTBOOKRESPONSE (datasetversion_id);
+
+CREATE INDEX IF NOT EXISTS INDEX_DATASETMETRICS_dataset_id ON DATASETMETRICS (dataset_id);
+
+DO $$
+BEGIN
+    IF EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name = 'datasetversion' AND column_name = 'versionstate') THEN
```
landreev: This is not a big deal by any means. But I'm fairly positive that these "IF EXISTS .." lines are not needed for a simple update like this. 
Granted, this is not causing any problems either. But I like to keep these scripts as clean and simple as possible. I will update it.

landreev:
Added details to the "how to test" section. Should be ready for review now.

landreev:
Merge conflict resolved, flyway script renamed to .3.

pdurbin:
API test are passing.

I tested a fresh db in Docker. Worked fine.

Then I blew away my database, switched to develop, spun up Dataverse, and created a dataset to give myself a bit of an existing db.

Then I switched back to this pull request and deployed it (keeping the db). It seems to work fine. Merging.

PR #11904, opened 2025-10-17
Adapt to LC's change to use doi array

qqmyers:
**What this PR does / why we need it**: LocalContexts has updated their software and is replacing the publication_doi field with a doi array of values. (They do have some backward compatibility for using publication_doi, but this field is not populated for new projects on their site.) This PR updates Dataverse's queries to address that change, looking in the doi array for the dataset's PID and searching the doi array. It also makes a change to just search for the raw PID which, since LC does a 'contains' search internally, will match regardless of whether the user enters the doi:* form, or the https://doi.org/* form. A similar change checks for one of the two forms when retrieving specific project info.

To Do: 
- Outside the PR - assure that documentation in the external vocab repo, and any Harvard specific documentation, is up to date

- Closes #

**Suggestions on how to test this**: On a machine configured for LocalContexts, assure that one can link/unlink an LC project using the LocalContexts metadata block.

**Additional documentation**:


qqmyers:
FYI w.r.t. QA - the build is currently failing due to maven.dataone.org being unresponsive. The prior build passed (https://jenkins.dataverse.org/blue/organizations/jenkins/IQSS-Dataverse-Develop-PR/detail/PR-11904/5/pipeline/) and the only change since was just to fix a conflict in the list of IT tests being run.

PR #11908, opened 2025-10-18
A quick fix for the missing org attribute in ShibServiceBean

### LINKED ISSUE #11907: Missing attribute name in the mdq implementation
Opening the issue so that I can make a quick PR with a trivial fix. 
Somehow (via a botched copy-and-paste maybe?) the name of one of the saml attributes in ShibServiceBean got blanked in the final version of the mdq pr.

landreev:

This is a trivial fix for a bug that was introduced right before the shib/mdq PR got merged; possibly as a botched copy-and-paste. The issue was only manifesting itself w/ brand new accounts; and was relatively minor and easy to miss. 

Kudos to @donsizemore for spotting the issue. 

- Closes #11907

The only 2 instances where the bug and the fix can be tested are demo.dataverse.org and dataverse6.rdmc.unc.edu. (the fix is already deployed on demo). 
The actual test is to set the `affiliation` in the `authenticateduser` table to null for your harvardKey or a similar institutional account, then log in. With the fix in place it will be properly populated with "Harvard University" or whatever institutional equivalent after a successful login. 

**Additional documentation**:


donsizemore:
dataverse6.rdmc was also fixed yesterday afternoon. we verified that new accounts got the correct affiliation, and existing accounts with manually munged affiliations had their affiliation corrected at the next sign-in. thank you @landreev !

ofahimIQSS:
tested in demo, looks good from my end - ready to merge

PR #11935, opened 2025-10-28
cpwebapp.sh macos fix

beepsoft:

When using the file watcher to automatically copy changed .xhtml files to the dev docker  container using cpwebapp.sh this script fails because macOS's grep doesn't support -P. This PR fixes it by using awk instead, which should work on all unix platforms.


docker/util/intellij/cpwebapp.sh
```diff
@@ -9,15 +9,26 @@ PROJECT_DIR="$1"
...
+  if [[ -z "${VERSION:-}" ]]; then
+    echo "Error: Could not extract <revision> from $POM" >&2
+    exit 1
+  fi
+
+  CONTAINER="dev_dataverse"
+
+  # Build target path
+  RELATIVE_PATH_WITHOUT_WEBAPP="${RELATIVE_PATH#src/main/webapp/}"
+  TARGET_PATH="/opt/payara/appserver/glassfish/domains/domain1/applications/dataverse-$VERSION/${RELATIVE_PATH_WITHOUT_WEBAPP}"
```
stevenwinship, line 28: Error response from daemon: Could not find the file /opt/payara/appserver/glassfish/domains/domain1/applications/dataverse-6.8 in container dev_dataverse

I don't see this directory in my dev_dataverse container

$ ls -al /opt/payara/appserver/glassfish/domains/domain1/applications/
total 20
drwxr-xr-x 4 payara payara 4096 Nov  7 20:57 .
drwxrwxr-x 1 payara root   4096 Nov  7 20:57 ..
drwxr-xr-x 4 payara payara 4096 Nov  7 20:57 ejb-timer-service-app
drwxr-xr-x 3 payara payara 4096 Nov  7 20:57 __internal

stevenwinship: I do see a read-only version here: /opt/payara/deployments/dataverse/

stevenwinship: Ok. I got it working :)

stevenwinship (approved):

stevenwinship:
Since I already tested this and it requires other setup that others might not want to do, I decided to also perform the QA task.

PR #11937, opened 2025-10-28
Feature: force content type

ffritze:
It features a force contentType implementation if one is not satisfied with the redetection behaviour / result.

- Closes #
It is not related to an issue. It's a new feature.

Tests are implemented and are passing.

It extends the Files API but does not change any old or existing API implementation.

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
@@ -1370,7 +1370,7 @@ public class UtilIT {
     static Response forceFileType(String fileId, boolean dryRun, String apiToken, String contentType) {
         return given()
                 .header(API_TOKEN_HTTP_HEADER, apiToken)
-                .post("/api/files/" + fileId + "/forcecontenttype?contentType="+contentType+"dryRun=" + dryRun);
+                .post("/api/files/" + fileId + "/forcecontenttype?contentType="+contentType+"&dryRun=" + dryRun);
     }
 
     static Response getSwordAtomEntry(String persistentId, String apiToken) {


qqmyers:
FWIW: Given that mimetype is part of the file metadata, it might be simpler to allow the existing POST /api/files/{id}/metadata call to handle it. That may be as simple as just adding mimetype to https://github.com/IQSS/dataverse/blob/53bff4ce6c96b1373c23a2ab694bd44fcbaa8d66/src/main/java/edu/harvard/iq/dataverse/datasetutility/OptionalFileParams.java#L516 . 

W.r.t. allowing any mimetype - there is logic in Dataverse to override the mimetype sent in in specific cases when new files are added (https://github.com/IQSS/dataverse/blob/53bff4ce6c96b1373c23a2ab694bd44fcbaa8d66/src/main/java/edu/harvard/iq/dataverse/util/FileUtil.java#L855 ).  I don't know if any of this is useful/needed as a check when updating the mimetype. It looks like the code in this PR is refusing to update if there has been a successful ingest - seems like at least that is needed to avoid breaking ingested files - do the other rules also make sense? (Is the use case to override Dataverse's assignment of a mimetype when the internal algorithms have figured one out, or just to be able to update one coming from a browser/cases where Dataverse doesn't recognize the mimetype? )

pdurbin:
"Please note that it’s possible to “trick” a Dataverse installation into giving a file a content type (MIME type) of your choosing. For example, you can make a text file be treated like a video file with -F 'file=@README.txt;type=video/mpeg4', for example." -- https://guides.dataverse.org/en/6.8/api/native-api.html#add-a-file-to-a-dataset

stevenwinship:
@ffritze It doesn't look like this PR is ready for review. Will you be addressing the comments made by @qqmyers and @pdurbin ?
Adding a waiting tag until a response to the comments is made

pdurbin:
@stevenwinship these PRs are moved to "ready for review" during Triage Tuesday. Probably a core developer should assign themselves as well as the contributor, put the PR in "in review", and work with the contributor. We can see at standup if there are any volunteers. 😄

qqmyers:
@ffritze - after some core team discussion, we recognize that having a way to update the mimetype of a file would be useful, but we're planning to reject this PR _as is_ (with a new endpoint). If you're willing/able to update the existing /api/files/{id}/metadata to include being able to change the mimetype, we'd be happy to review/QA/accept a PR with that approach.

ffritze:
@qqmyers OK, thanks for the update. I will close then this PR so far and see if I have time for your requested feature.

PR #11943, opened 2025-10-30
Add check for findByAltGlobalId when parsing handles

stevenwinship:
**What this PR does / why we need it**: Requests for Handle redirect containing /citation was not redirecting to dataset.xhtml causing a Not Found response. 

**Which issue(s) this PR closes**:Does not full close https://github.com/IQSS/dataverse.harvard.edu/issues/315

- Closes #

**Suggestions on how to test this**: Tested manually by opening https://qa.dataverse.org/citation?persistentId=hdl:1902.1/00021 in the browser and seeing that it redirects to https://qa.dataverse.org/dataset.xhtml?persistentId=hdl:1902.1/00021. Testing with the actual link (https://hdl.handle.net/1902.1/00021) will redirect to Production (https://dataverse.harvard.edu/citation?persistentId=hdl:1902.1/00021) causing the Page Not Found. 

**Additional documentation**:


landreev:
Until this /citation fix is deployed, should we put an Apache rewrite rule in place for this? 
The citation servlet redirects to both dataset.xhtml and file.xhtml; but our handles are all for datasets, so we can safely rewrite to dataset.xhtml in all cases.

landreev:
Re-tested to be safe. Merging.

PR #11947, opened 2025-10-31
A workaround for the annoying "draft not found" after a successful completion of async. publishing

### LINKED ISSUE #7653: Dataset Publish - Missing success msg, and unfortunate "DRAFT was not found" on page reload
Two issues, the green dataset publish success msg has gone missing, and when I reload the dataset pg in the browser after said "silent" success, there is a big, yellow, unfortunate "DRAFT is missing" warning msg, because the URL reads "version=DRAFT" and not "version=1.0".

Saw this on localhost and reproduced on demo (5.3)

1. create a draft dataset... save
2. publish... see yellow "publish in progress" lock warning msg
3. see page spinner check the system if dataset lock goes away...
4. page sees dataset lock is removed.. lock warning is removed... blue "DRAFT" label under title changes to "Version 1.0"...

And that's it.

No big green SUCCESS msg let you know you're dataset is published for the world to see. You have to piece together that the removal of the lock msg, and the changing of the status label (if you notice it) is "visual indication" enough.

The bundle has a dataset publish success message, `dataset.message.publishSuccess=This dataset has been published.`

In DatasetPage.java (line 2532) there appears to be a call for it as well...
```
    private String releaseDataset(boolean minor) {
        if (session.getUser() instanceof AuthenticatedUser) {
            try {
                final PublishDatasetResult result = commandEngine.submit(
                    new PublishDatasetCommand(dataset, dvRequestService.getDataverseRequest(), minor)
                );
                dataset = result.getDataset();
                // Sucessfully executing PublishDatasetCommand does not guarantee that the dataset 
                // has been published. If a publishing workflow is configured, this may have sent the 
                // dataset into a workflow limbo, potentially waiting for a third party system to complete 
                // the process. So it may be premature to show the "success" message at this point. 
                if ( result.isCompleted() ) {
                    JsfHelper.addSuccessMessage(BundleUtil.getStringFromBundle("dataset.message.publishSuccess"));
                } 
                ...
```

This change appears to have been introduced to solve _Unexplained publishing failures #7303_, which required a 4.20 patch from @landreev, when he introduce the publish lock to the system. He has also informed me that the loss of the big green success msg was discussed and determined to be acceptable.

With that said, I would also like to point a SECOND bug related to this. After the dataset is published, if you reload the page in your browser, you receive a big yellow "DRAFT was not found" warning msg, which is because the URL you reloaded in your browser is still "dataset.xhtml?version=DRAFT" after you have published, and not "dataset.xhtml?version=1.0".

Screen Shot 2021-03-04 at 4 42 11 PM

### LINKED ISSUE #8136: When publishing a version, there is a confusing message on page reload
After publishing a new draft of a dataset from the dataset page, if you reload the page, a yellow message says, 

"Info - The "DRAFT" version was not found. This is version [new version]." 

The URL shows "DRAFT" even though the dataset is published. 
The message is confusing, because the meaning and reason/importance is unclear. 

This happens to all users who publish. 

I expect the URL to reflect that I am on the new/latest version, and that the message not appear. 

Reproduced on Demo v 5.6

Screen Shot 2021-10-08 at 4 47 21 PM

### LINKED ISSUE #8548: Dataset page stays on the url with "version=DRAFT" after the dataset is published
... this results in "DRAFT version not found" warning message. Refreshing the page does not fix it, because "version=DRAFT" is in the url. 
This is a sub-issue of #7653. Opening a "small chunk" issue for this specific problem, since it's the most annoying to the users; and should be the easiest to fix.

### LINKED ISSUE #11866: Publishing and dataset landing page: Once a dataset is published, the draft URL and draft message should disappear
on Harvard Dataverse production

A dataset is published, and the draft to publish messages shows up. Publishing happens, and the draft messages and the draft URL do not refresh. I refresh the page, and the draft message and URL still do not refresh to the published (links)

Image

landreev:

This has been annoying users for years now (#8136 #7653 #8548... there are possibly more that I couldn't find right away). This 3-line workaround makes it less annoying.

It may be debatable whether it actually "closes" the issue below. I am selling it as a "workaround" only. But it definitely improves the user experience; and we likely don't want to invest any more effort than this into a jsf page at this point. So may as well close it. 

- Closes #8136 
- Closes #7653
- Closes #8548
- Closes #11866

Prior to this build: publish a dataset; when the async. publishing process completes in the background and the lock disappears, the url in the browser location bar will still have "version=DRAFT" in it. You may or may not see the yellow "draft not found, verison such and such shown instead" warning at this point. Try to reload the page - now you will be presented with the annoying yellow message. (Apparently that's what real life users keep doing: instead of manually removing the "version=" parameter from the url, they keep reloading the page, hoping that it will go away - and keep seeing the warning message again...)

With this build deployed: you should not be seeing the yellow warning message at any point during the publishing process. Reloading the page will quietly send you to the now published version. (Assuming publishing has actually worked; if it has failed for whatever reason, you will stay on your draft version, which is what we want)

**Additional documentation**:


landreev:
(this is already in the HDV 6.8 prod. patch deployed)

stevenwinship:
Tested:
Original code did show DRAFT in url and yellow Draft not found message when reloading
New code brought up published version when reloading the url with version=DRAFT

stevenwinship:
Tested reloading while the publish was taking place. Draft shows until publish finishes. Then published version shows. Nice!

PR #11950, opened 2025-11-04
BugFix: controlled vocab values validation

### LINKED ISSUE #11900: Required controlledVocabulary metadata marked as valid while empty
### Bug description  
In Dataverse (v6.8), when a previously optional metadata field becomes required for a dataset, or when a dataset is created via the API without all required metadata (`dataverse.api.allow-incomplete-metadata=true`), the system displays an *Incomplete metadata* label and normally prevents DRAFT form validation or dataset publication.  

However, if the user uploads a file, Dataverse inserts a record for the missing required `controlledVocabulary` metadata into the `datasetfieldvalue` table.  

**Consequence:** required `controlledVocabulary` metadata fields (e.g. `subject`) are treated as valid in the dataset metadata form even though their value is empty.  

### Steps to reproduce  
1. Create a dataset via API with `dataverse.api.allow-incomplete-metadata=true` without including all required metadata.  
2. Upload a file to the dataset.  
3. Open the dataset metadata edit form.  
4. Notice that required `controlledVocabulary` fields appear valid although no value is set.  

### Expected behavior  
Required `controlledVocabulary` fields should remain invalid until a valid value is provided.  

### Affected version  
Dataverse v6.8  

### Are you thinking about creating a pull request for this issue?
No fix is currently planned by the team.

ErykKul:
Bug fix.

- Closes #11900

- see also the test; it illustrates how the bug can happen

- hard to reproduce the bug, the easiest way was to write a test for it and ensure no regression can happen.

- no

- probably not? Maybe one liner: "bugfix: controlled vocab values validation when empty"?

**Additional documentation**:
-


src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
```diff
@@ -1788,11 +1788,18 @@ private void removeEmptyValues() {
...
             final Iterator<DatasetFieldValue> i = dsf.getDatasetFieldValues().iterator();
             while (i.hasNext()) {
-                final String v = i.next().getValue();
+                final DatasetFieldValue fieldValue = i.next();
+                final String v = fieldValue.getValue();
                 if (StringUtils.isBlank(v) || DatasetField.NA_VALUE.equals(v)) {
                     i.remove();
+                    continue;
+                }
+                if (isControlledVocabulary && dsf.getDatasetFieldType().getControlledVocabularyValue(v) == null) {
```
qqmyers: Do you know if these changes are needed (the ones in this class. I can see why the change in DatasetPage is important)? At a minimum, I think if isControlledVocabulary is true, there shouldn't ever be a regular value unless it's temporary, so you could always remove it in that case (possibly just adding` || isControlledVocabulary` to line 1795 instead of a second if statement. 

Beyond that though - we are hopefully never adding a temporary value that is not blank or NA_VALUE, so I'd think that the original code should catch and remove the temp value without changes in this class (with your change to DatasetPage). (If not - if there are temp values getting added that are other strings, maybe we should find/fix those? Or catch them for normal fields too and not just controlled vocab?)

qqmyers:
FWIW: The previous build had a failure in the Harvest from DataCite test, but just rerunning the job fixed it, so presumably random/transient and not related to this PR.

src/main/java/edu/harvard/iq/dataverse/DatasetVersion.java
qqmyers: @ErykKul - just a ping - this is waiting on you for a response

ErykKul: I was mixing two things here, the code here was for removing no-longer-existing vocab values, but I am not sure if that was a good idea, and definitely should not be mixed in this PR with the bug solution for "valid" empty values (I first though it was the reason for them being empty and valid: an old controlled vocabulary value that was removed from the metadata-block shows empty but it is not empty and it is not NA value -> however, that was not the reason for the bug, I could reproduce it in a way as the issue describes).

ErykKul: I reverted the file.

ErykKul: I also removed the unit test; it was testing the not-in-vocabulary value in CV field:
```java
        DatasetField subjectField = new DatasetField();
        subjectField.setDatasetFieldType(subjectType);
        subjectField.setDatasetFieldValues(new ArrayList<>());
        subjectField.setControlledVocabularyValues(new ArrayList<>());
        DatasetFieldValue orphanValue = new DatasetFieldValue(subjectField);
        orphanValue.setValue("__placeholder__");
        subjectField.getDatasetFieldValues().add(orphanValue);
```
This is not the cause of the bug, so it was testing the wrong thing (and started failing after cleanup of the file)

ErykKul: I also re-added `workingVersion.validate();` in DatasetPage, it has side effects we still want to have (set validation messages on dataset fields, that are then showed in the UI).

qqmyers (approved):
Looks good. isValid() is checking that the version is valid after removal of temporary N/A values which makes sense. Not sure the underlying code is very efficient (cloning the version) but we use it elsewhere.

stevenwinship:
@stevenferey @ErykKul 
Unable to reproduce this bug. More info in the main issue

stevenwinship (approved):

landreev:
@ErykKul @sekmiller 
Sorry, we had to revert this PR in a hurry earlier this week. Please see the revert PR #12342, but in short, it broke a few  useful things.

stevenwinship:
@ErykKul I had an issue restoring you branch and merging develop into it. You can either try to get your branch back or use the one I created that also has some code in  [ImportServiceBean.java‎](https://github.com/IQSS/dataverse/pull/11950#diff-13cefe6e15545de5ec4f01baec5658ae20ae1fb071d776eba20d1f27825e141f)
11900-improved-cvoc-value-validation2

PR #11953, opened 2025-11-04
AI Guide: add another MCP server for Dataverse

pdurbin:

@beepsoft told me about https://github.com/hunrenai4impact/dataverse-project-mcp

I haven't tried it myself, but this PR adds it to the AI Guide.


4tikhonov:
Can you ask him about maintenance plan since it's not clear how it will be supported?

beepsoft:
I don't think we should add it to the guide, because this is just a trial, not something that will be maintained in the long run.

And with the new developments by Jan in pyDataverse adding support for easy creation of MCP-s I would even consider this obsolete.

pdurbin:
Ah, ok. I'll close this, then.

I was simply going through my bookmarks and trying to figure out where to put all the most interesting links!

See also [#community > MCP (Model Context Protocol) @ 💬](https://dataverse.zulipchat.com/#narrow/channel/375707-community/topic/MCP.20.28Model.20Context.20Protocol.29/near/553700255)

PR #11971, opened 2025-11-12
Globus patch - no need to lock the dataset while an upload is in progress, when tasks are monitored asynchronously.

landreev:

Globus uploads have been handled the same way as ingests - a dedicated lock is placed on the dataset for the duration preventing any further transfers or edits. This does not appear to be necessary when the asynchronous, database queue-based task monitoring mode is enabled. Since the whole point of Globus support, at HDV at least, is for handling larger/TB-sized data these transfers can take a long time (days potentially) and keeping the dataset locked further complicates an already cumbersome workflow. 

There's no corresponding issue as of now, this started as a production patch. 

- Closes #

I still have no idea how to go about creating meaningful tests for any Globus-related functionality. Any feedback is welcome. 

This can be tested on one of the instances where Globus storage is configured: demo and dataverse-internal. 
In a collection with a Globus storage volume assigned, starting a long-ish running Globus upload (will need to be something in at least 10s of MBs; and this is definitely a PR that will be easier to test from home, since transfers are obscenely fast between NESE and Harvard local networks). With this build, it should be possible to start another Globus transfer; the "add files" and all the other buttons except for "Publish Dataset" should stay enabled for the duration. 
A finer test will be to stack multiple simultaneous transfers, and confirm that a) the above is still true b) that the Publish button will stay disable for as long as at least one transfer is still active, but becomes enabled again once the last one finishes. Similarly, the message about publishing being disabled should disappear at the end. 
I suggest not to actually try and publish the dataset; simply because then you will be able to delete the draft, and have the files stored at NESE permanently erased in the process. Even though it's a tape volume dedicated to testing that demo and internal are configured to use, it's still prudent not to leave junk on it unnecessarily. 

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -2391,7 +2392,8 @@ public boolean isValidOrCanReviewIncomplete() {
 
     private void displayLockInfo(Dataset dataset) {
         // Various info messages, when the dataset is locked (for various reasons):
-        if (dataset.isLocked() && canUpdateDataset()) {
+        boolean globusUploadInProgress = globusService.isUploadTaskInProgressForDataset(dataset.getId());
```
stevenwinship: Is this call time consuming? It might be quicker to make the call only if FeatureFlags.GLOBUS_USE_EXPERIMENTAL_ASYNC_FRAMEWORK is enabled.

boolean isGlobalAsyncEnabled = FeatureFlags.GLOBUS_USE_EXPERIMENTAL_ASYNC_FRAMEWORK.enabled();

boolean globusUploadInProgress = !isGlobalAsyncEnabled ? false : globusService.isUploadTaskInProgressForDataset(dataset.getId());

String globusBundleKey = isGlobalAsyncEnabled ? "file.globusUpload.inProgressMessageAsync.details" : "file.globusUpload.inProgressMessage.details";

src/main/webapp/dataset.xhtml
```diff
@@ -360,16 +360,23 @@
...
+                                                            <f:passThroughAttribute name="class" value="btn btn-default btn-access btn-publish dropdown-toggle #{!DatasetPage.hasValidTermsOfAccess  or !valid ? 'disabled' : ''}"/>
                                                             <f:passThroughAttribute name="data-toggle" value="dropdown"/>
                                                             <f:passThroughAttribute name="aria-haspopup" value="true"/>
                                                             <f:passThroughAttribute name="aria-expanded" value="false"/>
                                                         </c:if>
+                                                        <!--c:if test="#{(showSubmitForReviewLink or showReturnToAuthorLink) and showPublishLink and DatasetPage.lockedFromPublishing}" -->
+                                                            <!-- f:passThroughAttribute name="class" value="btn btn-default btn-access btn-publish dropdown-toggle disabled"/ -->
+                                                        <!-- /c:if -->
```
stevenwinship: Is there a reason for adding this commented out code?
I'll approve it.

stevenwinship (approved):

src/main/webapp/dataset.xhtml
landreev: Removed, thanks.

landreev:
Added "how to test" info and size 10.

landreev:
I noticed that this PR was still sitting in "ready for QA" (unfortunately).  But I've used the opportunity to add another fix/improvement to it. Similarly to everything else in this PR it's already deployed in prod. at HDV.

landreev:
As I mentioned earlier, putting this on hold, pending dataverse-internal coming back to life.

cmbz:
2026-05-20: Moved to 6.12 because it is not urgent to do before the 6.11 release.

cmbz:
2026-08-12: @landreev do you want this for 6.12? Please let me know.

landreev:
> 2026-08-12: @landreev do you want this for 6.12? Please let me know.

@cmbz No strong preference

cmbz:
2026-08-12: Moved to 6.13 after conversation with @landreev

PR #11982, opened 2025-11-17
safe proof sql calls

stevenwinship:
**What this PR does / why we need it**: Safe proof SQL calls in search where solr not used

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse-security/issues/111

- Closes [#](https://github.com/IQSS/dataverse-security/issues/111)

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -40,6 +40,7 @@
 import jakarta.inject.Named;
 import jakarta.persistence.*;
 import jakarta.persistence.criteria.*;
+import org.jsoup.internal.StringUtil;
```
qqmyers: I think we usually use org.apache.commons.lang3.StringUtils for isBlank() (org.jsoup.internal.StringUtil isn't used anywhere currently)

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
```diff
@@ -281,60 +282,22 @@ public List<FileMetadata> findFileMetadataByDatasetVersionId(Long datasetVersion
...
-        String searchClause = "";
-        if(searchTerm != null && !searchTerm.isEmpty()){
-            searchClause = " and  (lower(o.label) like '%" + searchTerm.toLowerCase() + "%' or lower(o.description) like '%" + searchTerm.toLowerCase() + "%')";
+        String searchTerm = !StringUtil.isBlank(userSuppliedSearchTerm) ? "%"+userSuppliedSearchTerm.trim().toLowerCase()+"%" : null;
+
+        String selectClause = "select o.datafile_id from FileMetadata o where o.datasetversion_id = " + datasetVersionId;
+        String searchClause = searchTerm != null ? " and (lower(o.label) like ? or lower(o.description) like ?)" : "";
+        String orderByClause = " order by o." + sortFieldAndOrder.getSortField() + " " + sortFieldAndOrder.getSortOrder();
+
+        // NOTE: since datafile_id and datasetversion_id are indexes we can't use a TypedQuery
```
qqmyers: This isn't true - see https://github.com/IQSS/dataverse/blob/b78060aac62d8d40cb897bbf8a35b1c97a6e5ef7/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java#L706 for example - you just have to reference the object directly rather than the *_id column. I don't think there's any reason a typed query has to be used though, so just asking for the comment to be deleted.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
stevenwinship: fixed

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
stevenwinship: removed comment

landreev:
Huh, I wrote some of that insecure code 6 years ago. Although it really looks like I copy-and-pasted most of it from an even older fragment, that's been there for a decade, from the very beginning of Dataverse. An unpleasant blast from the past. 
Looking good so far, but going to test a bit more before merging.

src/main/java/edu/harvard/iq/dataverse/DataFileServiceBean.java
landreev: Everything is working properly as far as I can see. But, what is the actual reason it is still using a NativeQuery, instead of the "normal" kind? - j/c.

stevenwinship: I couldn't get it to work with a dynamic column name.  "o.mycolumnName" I.e "o.title"
I didn't try to get the OBJECT(o) to work

I'll give it a try

landreev: To be clear, I didn't think this was super important. If it is non-trivial to switch to a non-native query, we may as well just merge it.

landreev: @stevenwinship Should I just go ahead and merge it?

landreev: Ok, merging.

PR #11991, opened 2025-11-18
IQSS/11990 Controlled Vocabulary Field editing error

### LINKED ISSUE #11990: Editing a Controlled Vocabulary field can update published version

When editing a published version, if a single-choice controlled vocabulary field is edited, the change happens in both the new draft version and the original published version.

This does not appear to happen with a multiple choice field such as Language.

* When does this issue occur?
Only if a draft does not already exist.

* What happens?

6.8+ - looks like all versions are affected.

**Screenshots:**

*

PR to follow.

qqmyers:
**What this PR does / why we need it**: This PR fixes the issue - where editing a single-value CVV field starting from a published version would also update the field in the published version (causing the version difference view to not show the change).

- Closes #11990

**Special notes for your reviewer**: Took a while to find, but relatively straight forward. When an edit is made from a published version, we make a copy of the version and all fields. For DatasetFields using ControlledVocabularyValues, the code just set the new field in the new version to share the array of values from the original version. When an edit was made, the change was to the shared array and thus got saved for both versions. A fix is to copy the values in the array rather than sharing the array.

I confirmed the issue on demo, and confirmed the fix on a test machine. 

I would have expected the error to occur with fields like language, but it appears that when multiple values are used, the JSF is sending a new array rather than editing the existing one, so the fact that the original copy was shared doesn't matter.
I did not test the API, but I would guess the problem could occur there as well since it was related to the DatasetField.copy method.

W.r.t. release notes - the copy code is 11+ years old, so I'm assuming the bug is that old and affects basically all versions of Dataverse currently in use.

Not sure if this has been seen in any of the version differencing related issues.

**Suggestions on how to test this**: Per the issue, create and publish a dataset. Change the value in a CVV single value field (publication relationType or contributor type) and save. Check the version difference details or look at the published version and, before the PR, you'll see the new value in the published version.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetField.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetField.java
@@ -172,7 +172,6 @@ public class DatasetField implements Serializable {
     }
 
     @OneToMany(mappedBy = "parentDatasetField", orphanRemoval = true, cascade = {CascadeType.REMOVE, CascadeType.MERGE, CascadeType.PERSIST})
-    @OrderBy("displayOrder ASC")
     private List<DatasetFieldCompoundValue> datasetFieldCompoundValues = new ArrayList<>();
 
     public List<DatasetFieldCompoundValue> getDatasetFieldCompoundValues() {


src/main/java/edu/harvard/iq/dataverse/DatasetField.java
```diff
@@ -604,14 +604,15 @@ private DatasetField copy(Object versionOrTemplate, DatasetFieldCompoundValue pa
...
+                dsf.setDatasetVersion((DatasetVersion) versionOrTemplate);
             } else {
                 dsf.setTemplate((Template) versionOrTemplate);
             }
         }
         
         dsf.setParentDatasetFieldCompoundValue(parent);
-        dsf.setControlledVocabularyValues(controlledVocabularyValues);
+        
+        dsf.getControlledVocabularyValues().addAll(controlledVocabularyValues);
```
stevenwinship, line 615: True, it's not the same List but since the List.addAll is not a deep copy. Wouldn't the items in the new list still refer to the same objects in the old list?

qqmyers: Yes, and that's OK - they are CVV values that are used system-wide. It's only the List that shouldn't get reused.

stevenwinship (approved):

Issue #11992, opened 2025-11-20
Setting a field to 'hidden' is not working


LauraHuisintveld:
**What steps does it take to reproduce the issue?**

1. Create a new dataverse
2. Set the fields 'Series' and 'related material' (for example) to hidden by unchecking the checkboxes.
3. Click done.
4. Click 'save changes'.
5. Add a dataset. I expected these fields not to be visible when creating a new one, but I still see them. Also, if you revisit the 'General info' page, the boxes are checked again, indicating the changes have not been saved.

 **When does this issue occur?**
When editing the metadata settings and creating an new dataset.

 **Which page(s) does it occurs on?**
Dataverse - General information page
Dataset - Create page

 **What happens?**
If you set a metadata field to 'hidden' this setting will be ignored.
The condition that decides if a field should be displayed on the Edit dataset pages doesn't check the 'include' setting anymore.

 **To whom does it occur (all users, curators, superusers)?**
Dataverse admins, curators, dataset creators

**Which version of Dataverse are you using?**
6.7.1. I could also reproduce it on demo.datasverse.org, which is on 6.8

**Are you thinking about creating a pull request for this issue?**
DANS is working on a fix, so probably a pull request will follow.

PR #11995, opened 2025-11-20
IQSS/8690 Make producer optional

### LINKED ISSUE #8690: Dataverse collection managers can't hide the Producer field, and possibly any field with a child field that's conditionally required

* When does this issue occur?
  When trying to customize a Dataverse collection so that the Producer field is hidden

  The Dataverse collection "General Information" page

* What happens?
  It's not possible to uncheck the checkbox that would hide the Producer fields. The check box is greyed out or inactive.

  Curators, superusers

  The checkbox can be unchecked so that it's possible to hide the Producer field when people edit datasets in that collection

  v5.10.1, but I see this bug on Dataverse installations running v5.8.3 and I suspect that any installation with the "conditionally required" feature (https://github.com/IQSS/dataverse/issues/7606), added in v5.4, has this bug. I also spun up [an instance on AWS](http://ec2-3-94-8-112.compute-1.amazonaws.com) where the conditional requirement of the Producer's Name child field was removed in the citation.tsv file and in that AWS instance I was able to hide the Producer field.

https://github.com/IQSS/dataverse/issues/7606

**Screenshots:**

Producer field's checkbox can't be unchecked, so field can't be hidden

<kbd>Screen Shot 2022-05-11 at 9 42 26 AM</kbd>

In the AWS instance I can uncheck the Producer field's checkbox to hide the field (and Producer Name field is not conditionally required)

<kbd>Screen Shot 2022-05-11 at 10 05 35 AM</kbd>

qqmyers:
**What this PR does / why we need it**: This PR makes the change suggested in #8690 to not require the producerName so that the producer field becomes optional.

- Closes #8690 

**Special notes for your reviewer**: There's discussion in the issue. If we want to set producerName to not be required at the block level, this PR will do it. (Probably should include a release note w.r.t. how to set it to required at the root Dataverse level if you want for backward compatibility, and with the usual instructions for updating the block. Happy to add that when I get back.)

**Suggestions on how to test this**: Testing requires a fix for #11992 - e.g. PR #12017 .Once that is available/merged, you should be able to check in the UI that it is now possible to uncheck the Producer field on a given Dataverse and verify that the field no longer shows for datasets created within it.

**Additional documentation**:


cmbz:
2026-02-25: Moving to On Hold until we've decided the best approach for solving the problem.

Issue #11529, opened 2025-05-27
Extend Storage Quotas to individual datasets and user accounts


cmbz:
(edited by @landreev 05-27-2025; question marks indicate issues that may need additional discussion before we have a clear idea of what to implement and how)

## Overview of the Feature Request
-  Last year we have added storage quotas that can be enabled on collections. However, that cannot be used to track and control storage use in the datasets that installations like IQSS allow users to create in the root-level collection. 
- We have a local curation request (see https://github.com/IQSS/dataverse.harvard.edu/issues/365) to address this by adding quotas that can be configured on per user account-basis. 
- While it should be easy to force an account to become read-only once a certain amount of data is deposited with it, it seems like it would be very easy to bypass, by giving write access to a dataset to other accounts (?). 
- It may however be possible to address this by being able to set quotas both on a per-user AND per-dataset basis (?)

Extending the currently implemented collection quotas to be configurable on individual datasets should be trivial. The same code that keeps track of collection storage use already does that for datasets as well. The actual quota check methods already work on either collections or datasets. That internal functionality just needs to be extended to the APIs that set the quotas and report storage use. 

It should not be difficult to add keeping track of storage use by user accounts either. Provided we simply count the sizes of all the files created by the given `authenticateduser`. (?)

## Any open or closed issues related to this feature request?
- https://github.com/IQSS/dataverse.harvard.edu/issues/364
- https://github.com/IQSS/dataverse.harvard.edu/issues/365
- https://github.com/IQSS/dataverse/issues/7829
- https://github.com/IQSS/dataverse/issues/11275

landreev:
Now that we have enabled quotas on all top-level collections at HDV, we'll need to prioritize this issue, so that we can set up quotas for the datasets created directly in the top-level, root collection, and the users who own them. 
This will take some design and dev. effort. In the meantime though, I am going to open a sub-issue for one intermediate, partial solution that can be produced very easily.

cmbz:
2026-02-25: Next steps
- Talk to @sbarbosadataverse and @scolapasta and define what a user quota is/means. Which takes precedence: collection or user quota (e.g., whichever is lower?)

cmbz:
2026-03-30: Proposal
(from @sbarbosadataverse, @scolapasta, and @pdurbin)
- By default, individual users have quotas
- By default, collections do not have quotas
- _Only_ collections that have paid (e.g., via an invoice) will have quotas
[...]

PR #11940, opened 2025-10-29
Fixes for #11716

qqmyers:
**What this PR does / why we need it**:  The original PR did not use a valid method for determining whether the store supports directUpload. This PR addresses that, along with making several other minor improvements:

- using a more direct method to get the directDownload flag to avoid sending the dataset as a parameter and instantiating a StorageIO object
- moves the changelog entry to 6.9 as the original #11716 did not make 6.8
- clarifies some documentation
- adds an uploadOutOfBand boolean entry (stores with this = true allow creating file entries by just sending metadata and a storageIdentifier - like S3 direct upload except that the files have to have been put in place by some out-of-band mechanism (e.g. manual copying into a file store)
- adds a unit test to assure that the store properties are read correctly (when stored as System properties)

**Which issue(s) this PR closes**:

- Closes #

**Special notes for your reviewer**:

**Suggestions on how to test this**: Call these apis for datasets/dataverses and verify that the correct store is returned and that the values match what is set (e.g. as jvm-options).

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #11997, opened 2025-11-20
Adds configurable storage quotas on individual datasets

### LINKED ISSUE #11987: Add storage quotas on individual datasets
This is a child issue of #11529. I'm opening it because this will be very easy to achieve - the code that's behind the collection-level quotas is already implemented to cover all the `dvobjectcontainers`, i.e. both Collections and Datasets. 
Producing this patch will allow us to at least quickly slap quotas on all the datasets sitting directly in the top-level collection at HDV. This will not address the issue of the same user creating multiple datasets in root - which needs to be fully addressed in the parent issue, but that will take more of an appreciable design and dev. effort.

landreev:

This is a low-hanging fruit/easily-achievable part of the parent issue #11529 that will make it possible to configure quotas on individual datasets. The primary use case is datasets in root and/or another collection where defining a collection-level quota may be impractical. The functionality was already there - the implementation underneath operates on DvObjectContainers; it was in fact already possible to create a dataset-level quota by directly inserting an entry into the `storagequota` table. This PR exposes the functionality via the new APIs under `/api/datasets/`.

The part of the parent issue that will require more design and dev., configurable quotas for users, remains to be prioritized separately

- Closes #11987

Note that as of right now (11 am Nov. 21) the checklist below says that the continuous integration test has failed ("The build of this commit was aborted"). But this is because I killed a redundant extra Jenkins run that was triggered by a typo fix in the release note. The last Jenkins run did actually pass - see https://jenkins.dataverse.org/job/IQSS-Dataverse-Develop-PR/view/change-requests/job/PR-11997/. 

Straightforward. Enable a quota on a dataset (see the section of the API guide and/or the FilesIT test that are being added here), make sure uploads are blocked once it is reached. 

**Additional documentation**:

### DIFF
--- a/doc/sphinx-guides/source/api/native-api.rst
+++ b/doc/sphinx-guides/source/api/native-api.rst
@@ -1316,9 +1316,11 @@ The following convenience API shows the dynamic values of the *remaining* storag
    curl  -H "X-Dataverse-key: $API_TOKEN" "http://localhost:8080/api/datasets/$dataset-id/uploadlimits" 
    {
      "status": "OK",
-     "uploadLimits": {
-       "numberOfFilesRemaining": 20,
-       "storageQuotaRemaining": 1048576
+     "data": {
+       "uploadLimits": {
+         "numberOfFilesRemaining": 20,
+         "storageQuotaRemaining": 1048576
+       }
      } 
    }
 
@@ -1328,7 +1330,9 @@ Or, when neither limit is present:
 
    {
      "status": "OK",
-     "uploadLimits": {} 
+     "data": {
+       "uploadLimits": {}
+     }
    }
 
 This API requires the Edit permission on the dataset. 


doc/sphinx-guides/source/api/native-api.rst
```diff
@@ -1270,7 +1275,40 @@ To delete a storage quota configured for a collection:
...
 
-Use the ``/settings`` API to enable or disable the enforcement of storage quotas that are defined across the instance via the following setting. For example,
+Storage Quotas on Individual Datasets
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: 
+
+  curl -H "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/datasets/$ID/storage/quota"
+
+Will output the storage quota allocated (in bytes), or a message indicating that the quota is not defined for the specific collection. The user identified by the API token must have the ``Edit`` permission on the dataset.
```
stevenwinship: copy-paste?  specific collection should say specific dataset

landreev: Just noticed and fixed it a few minutes ago, independently.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/DeleteDatasetQuotaCommand.java
```diff
@@ -0,0 +1,53 @@
...
+        if (targetDataset == null) {
+            throw new IllegalCommandException("", this);
+        }
+        
+        StorageQuota storageQuota = targetDataset.getStorageQuota();
+        
+        if (storageQuota != null && storageQuota.getAllocation() != null) {
+            ctxt.dataverses().disableStorageQuota(storageQuota);
+        } 
+        // ... and if no quota was enabled on the collection - nothing to do = success
```
stevenwinship: 'collection' should be 'dataset'
maybe a note that says ctxt.dataverses() handles quotas for collections and datasets

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetQuotaCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author landreev
+ * The sole purpose of this command is to check the permissions
+ * when it's called by the /api/datasets/.../storage/quota api.
+ */
+// No @RequiredPermissions defined, dynamic
```
stevenwinship, line 21: the api doc says this requires edit dataset permission. I don't see that limitation here

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -1118,5 +1119,34 @@ public int getDataFileCountByOwner(long id) {
...
+    /**
+     * 
+     * @todo: consider moving the quota methods, from here and the DataverseServiceBean,
+     * to DvObjectServiceBean. 
+     */
+    public void saveStorageQuota(Dataset target, Long allocation) {
+        StorageQuota storageQuota = target.getStorageQuota();
+        
+        if (storageQuota != null) {
+            storageQuota.setAllocation(allocation);
```
stevenwinship, line 1132: Is there any check to prevent negative numbers? I guess 0 or negative will just block the ability to store more data.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDatasetQuotaCommand.java
landreev: Oh yeah, wrong API - it's the `/api/data*/storage/use` APIs that require Edit-level permissions.

stevenwinship: I see that both of the endpoints (data*/$ID/storage/quota) are documented to require "Edit"
Neither GET dataverse or dataset actually require the permission to get the quota.

landreev: That's what I said, wrong APIs in the guide. Fixed.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
landreev: Yes, that is the answer - it is possible to set it to a negative number. But it will be equivalent to zero in practice. 
Tbh, I am generally less concerned about preventing invalid or meaningless values from being entered when it comes to superusers-only APIs. Under the assumption that they should know what they are doing, and/or can be expected to own the consequences of their actions.

doc/sphinx-guides/source/admin/dataverses-datasets.rst
```diff
@@ -269,6 +269,27 @@ The current driver can be seen using::
...
+   curl  -H "X-Dataverse-key: $API_TOKEN" "http://localhost:8080/api/datasets/$dataset-id/storageDriver?showRemainingQuotas=true" 
+   {
+     "status": "OK",
+     "data": {
+       "label": "S3",
+       "type": "s3",
+       "name": "s3",
+       "directDownload": true,
+       "directUpload": false,
+       "numberOfFilesRemaining": 20,
```
qqmyers: FWIW: These two aren't really store related and they're very dynamic. Seems like just calling the get quota api for them would be cleaner.

landreev: Sorry, missed this last week. 
I don't have any reason to insist on adding this info to the /storageDriver API specifically. 
My only rationale was that the SPA is already calling it for the purposes of rendering the Upload Files page, so may as well pack all the info it will need for that in one place. (This was a last minute idea, I ran it by Steven on slack and I incorporated his suggestions). 

> Seems like just calling the get quota api for them would be cleaner.

I'm happy to remove these fields from /storageDriver. However, I would like to be able to pass it to the SPA in one call. It can otherwise be obtained using the already existing APIs, but it would take multiple (a potentially annoying number of) calls, to obtain the quotas, then the current use, and then calculate the remaining allocations (for the storage size in particular, it may take more than 2 calls with the current API - depending on where the quota is inherited from)
So, I'd rather add something like `/api/datasets/{id}/storage/limits` or `/api/datasets/{id}/storage/remaining`.

landreev: Perhaps `/api/datasets/{id}/uploadLimits`.

landreev:
> The JSF UI disables the file upload button/drag-n-drop area when it isn't possible to upload a file due to either file or size limits, but the 'Upload-a-folder', 'Dropbox', and probably Globus uploads are not disabled

I have this addressed in another patch PR - #11988 (deployed at HDV already)

landreev:
> ... This appears to be caused by delete just setting the local quota allocation to null rather than removing the StorageQuota

I did notice that myself the other day. I may have even done that on purpose... but the behavior is inconsistent/counter-intuitive, I agree.

landreev:
> The /storageDriver?showRemainingQuotas api call (already being changed) shows no new keys if quota not set - with the replacement api should there be a positive indication in the json that there are no quotas in that case?

That was also part of my rationale behind shoving this limits info into an existing API (optionally, when present); vs. figuring out how to communicate that there are no limits, in an API dedicated to showing the limits. 
My current suggestion (https://github.com/IQSS/dataverse/pull/11997/commits/1fbbff0bfc721cb5188a2636641d2504b1e970b1) is to have a dedicated API: 
```
/api/datasets/NNNNN/uploadlimits
{
  "status": "OK",
  "data": {
    "uploadLimits": {
      "numberOfFilesRemaining": 8,
      "storageQuotaRemaining": 1046483
    }
  }
}
```
or, in the absence of any limits
```
{
  "status": "OK",
  "data": {
    "uploadLimits": {}
  }
}
```
Would this be positive enough?

landreev:
> Unlike size quotas, file count limits can't be set on collections/inherited. Is that by design or something that could/should be added at some point?

I didn't implement those... but I am 99.9% positive that they are in fact inherited, from the nearest container with a limit defined. ... will double-check.

landreev:
> File limits have no setting, size quotas do: :UseStorageQuotas - is a setting needed? Should it cover both file/size limits

Yeah, this seems somewhat arbitrary; in practice it is simply a result of these two limits having been implemented by different developers. Although I should have thought of making them more consistent when reviewing the file counts PR. 

I personally feel like having a global setting is useful - for the sake of being able to flip the enforcement on and off, without losing the configuration. Would probably vote for having 2 separate settings for enforcing these.

landreev:
Overall, I would prefer to have a separate PR for following up on many of the suggestions above, rather than trying to address them here on a short notice (again, this was supposed to be a "low effort" PR). 
With some exceptions - I will go ahead and address this: 

> The pattern of using a POST with a numeric path variable, used in this PR and earlier ones, seems to be ~non-restful - 

I don't know what I was thinking there really... Did I copy-and-paste that solution from some other place in the API ?? - idk. 
And if we want it to be changed to a more reasonable `-X PUT -d '...'`, I feel like now's the good, or even last chance - up until now no one has been using these APIs in real life (afaik), but we are actively instituting quotas in prod. here, and I am in the process of passing the command lines for the APIs to the support/curation team.

Somewhat on the fence about the `-X DELETE /api/data*/{id}/storage/quota` behavior. I agree that it's not ideal as implemented; but it's not as urgent for us, since our quotas will be defined only on collections and individual datasets immediately in the root collection. I.e., no dataset should have a quota defined on more than one ancestor. ... but may as well. 

This I find very troubling: 

> For a non-superuser, adding more files than allowed by quota results in a subset of files being added (with no warning that all files won't be added)
For a non-superuser, adding a file over the limit results in the file simply disappearing in the upload pane (again w/o any explanation)

I'm quite positive that both mine, and Steven's implementations did have such warning messages in the JSF UI. Did they get muted by later changes to the page? - idk. I want this fixed, but in a separate pr.

landreev:
The changes made per QA feedback:
- /storageDriver changes rolled back (branch synced w/ #11940 now in develop);
- dedicated convenience API /uploadlimits added instead;
- the collection/dataset set quota APIs changed to use PUT; 
- the release note, the guides and the RestAssured tests changed accordingly.

I made an attempt to make -X DELETE remove the quota, instead of setting allocation to null; for some unclear reason that same code was bombing for datasets (??) - so I left it unchanged in both cases, for now.

qqmyers:
FWIW: I tested the size and file quota add/delete apis and verified that the new /uploadlimits api correctly reports when a file or size limit, or both, exist, and returns {"status":"OK","data":{"uploadLimits":{}}} when they don't.

I see a sematic API IT test failing in the latest build - I've seen that in other PRs - looks like an intermittent issue with the order of two description fields that I'll investigate, but I'll go ahead and merge this as it shouldn't be related.

qqmyers (approved):
Latest changes look good

PR #12006, opened 2025-11-26
Add security warning to production container guide

ghost:
This is a follow-up to PR #11862 and the email on the production container documentation.

- Added a `.. warning::` block at the top of `production.rst` to state that the demo tutorial is not secure by default.
- Updated the text to recommend using the demo as a "structural template" instead of saying that it teaches security configuration.

Thank you!


pdurbin:
@tsuretettee thanks! We can preview your changes here: https://dataverse-guide--12006.org.readthedocs.build/en/12006/container/running/production.html

PR #12031, opened 2025-12-09
add cascade for dataverse delete with metrics

### LINKED ISSUE #12030: Unable to delete Dataverse when a row exists in the metrics table
Due to a foreign key in DB table 'metric' a Dataverse can not be deleted via the UI when a metric exists.
The delete functionality must also delete any rows from the metric table.

* To whom does it occur (all users, curators, superusers)? All users

stevenwinship:
**What this PR does / why we need it**: dataverse delete fails if there are cached metrics due to a foreign key constraint

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/12030

- Closes #12030

**Suggestions on how to test this**: create/publish dataverse. request metrics to cause a cached row in metric table. Delete the dataverse

**Additional documentation**:

### DIFF
--- a/conf/keycloak/builtin-users-spi/pom.xml
+++ b/conf/keycloak/builtin-users-spi/pom.xml
@@ -100,7 +100,7 @@
     </build>
 
     <properties>
-        <keycloak.version>26.4.7</keycloak.version>
+        <keycloak.version>26.5.0</keycloak.version>
         <java.version>17</java.version>
         <jakarta.persistence.version>3.2.0</jakarta.persistence.version>
         <mindrot.jbcrypt.version>0.4</mindrot.jbcrypt.version>
--- a/doc/release-notes/11914-set-template-default-api.md
+++ b/doc/release-notes/11914-set-template-default-api.md
@@ -1,0 +1,15 @@
+## New Endpoint: POST `/dataverses/{id}/template/default/{templateId}`
+
+A new endpoint has been implemented to set the default template to a given dataverse collection.
+
+### Functionality
+- Sets the default template of the given dataverse collection.
+- You must have edit dataverse permission in the collection in order to use this endpoint.
+
+## New Endpoint: DELETE `/dataverses/{id}/template/default`
+
+A new endpoint has been implemented to remove the default template to a given dataverse collection.
+
+### Functionality
+- Removes the default template of the given dataverse collection.
+- You must have edit dataverse permission in the collection in order to use this endpoint.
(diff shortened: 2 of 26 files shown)


src/main/resources/db/migration/V6.8.0.4.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE metric DROP CONSTRAINT fk_metric_dataverse_id;
+ALTER TABLE metric ADD CONSTRAINT fk_metric_dataverse_id FOREIGN KEY (dataverse_id) REFERENCES dataverse(id) ON DELETE CASCADE;
```
qqmyers: I don't think this would usually be generated (with the ON DELETE CASCADE part) - the Cascade remove annotation works at the ORM layer. At least I haven't seen it before looking at the table descriptions for other places we have cascade remove annotations. I guess I'd suggest seeing if the remove works without changing the db constraint if you haven't already.

stevenwinship: I re-tested everything and the annotation on Metric didn't do anything. The only change that works is the constraint change in this sql

stevenwinship: I tested both with upgraded db/dataverse and new db

qqmyers: Hmm - that seems like a problem - what annotations did you try? Orphan?

qqmyers: Are we missing the equivalent of https://github.com/IQSS/dataverse/blob/70b6e260d364aadb3cb1c5374b38cfd7b1bae5ed/src/main/java/edu/harvard/iq/dataverse/Dataset.java#L144-L145 for Dataverse/Metric ?

stevenwinship: yes. Dataverse doesn't have that. There is no DataverseMetrics class

stevenwinship: added 
`@OneToMany(mappedBy = "dataverse", orphanRemoval = true, cascade = {CascadeType.REMOVE, CascadeType.MERGE, CascadeType.PERSIST})
    private List<Metric> dataverseMetrics = new ArrayList<>();`

sekmiller:
It does what it purports to do. - if there are records on the metric table I can still delete the dataverse and the metric record is also deleted. 
 I did find a couple of anomalies with the get Metrics API. If you call it with a parentAlias that exists but isn't published it will return a value (in my case the total number of dataverses in my local db). It doesn't create a record in the metric table. If you call it with a parentAlias that doesn't exist you get an "Api endpoint doesn't exist error" not a dataverse not found error. Out of scope for this, but possibly worth fixing?

PR #12053, opened 2025-12-17
Manage Guestbooks page performance

landreev:

The page is a 🐖 that penalizes collections for their popularity. 

- Closes #

**Additional documentation**:


landreev:
As discussed on slack: this has already been deployed on qa. 
To get a full picture of how awful it was, develop build would need to be deployed, and the new `index_guestbookresponse_dataset_id_guestbook_id` would need to be dropped. 
But, per everybody on slack, https://qa.dataverse.org/manage-guestbooks.xhtml?dataverseId=99 used to take ~2 min. to load without these fixes.

PR #12063, opened 2025-12-19
OAI_ORE and Archival Bag updates

qqmyers:
**What this PR does / why we need it**: This PR contains multiple updates to the OAI-ORE metadata export and archival Bag output:

OAI-ORE
- now uses URIs for checksum algorithms, as required to create valid JSON-LD
- a bug causing failures with deaccessioned versions when the deaccession note ("Deaccession Reason" in the UI) was null (which has been allowed via the API). 
- the "https://schema.org/additionalType" is updated to "Dataverse OREMap Format v1.0.2" to indicate that the out has changed 

Archival Bag 
- for dataset versions with no files, the (empty) manifest-<alg>.txt file created will now use the default algorithm defined by the "FileFixityChecksumAlgorithm" setting rather than always defaulting to "md5"
- a bug causing the bag-info.txt to not have information on contacts when the dataset version has more than one contact has been fixed
- values used in the bag-info.txt file that may be multi-line (with embedded CR or LF characters) are now properly indented/formatted per the BagIt specification (i.e. Internal-Sender-Identifier, External-Description, Source-Organization, Organization-Address).
- the name of the dataset is no longer used as a subdirectory under the data directory (dataset names can be long enough to cause failures when unzipping)
- a new key, "Dataverse-Bag-Version" has been added to bag-info.txt with a value "1.0", allowing tracking of changes to Dataverse's acrhival bag generation
- improvements to handling file retrieval errors and throttling
- Downloads to create bag are no longer counted as file downloads for the dataset

(I assume these don't work to automate closing, but they're here to let DANS see the correlations)

- Closes #DD-2109
- Closes #DD-2082
- Closes #DD1508
- Closes #DD2112

**Special notes for your reviewer**: FWIW: Some of the simpler fixes are in a single commit if you want to look at them separately.

**Suggestions on how to test this**: There are new tests for the multiline wrapping functionality and overall generation of the bag-info-txt file. Testing overall bag generation means configuring an archiver per the guides (probably the local file one) and verifying the output. The main changes to check are in the OAI-ORE file, bag-info file, and that there is no data/* dir using the dataset title. One could also verify that the manifest file name with no files matches the configured algorithm.
Given that DANS is interested in these, I assume they'll also be testing - could coordinate.

**Is there a release notes update needed for this change?**: note added

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11085-suppress-breadcrumbs
+++ b/doc/release-notes/11085-suppress-breadcrumbs
@@ -1,0 +1,1 @@
+This release fixes a bug which was allowing the viewing of host collections' names when using anonymized preview URLS. 
--- a/src/main/webapp/dataset.xhtml
+++ b/src/main/webapp/dataset.xhtml
@@ -21,6 +21,7 @@
             <ui:param name="version" value="#{DatasetPage.workingVersion}"/>
             <ui:param name="deaccessioned" value="#{version.deaccessioned}"/>
             <ui:param name="locked" value="#{DatasetPage.locked}"/>
+            <ui:param name="showDataverseHeader" value="#{!DatasetPage.anonymizedAccess}"/>
             <ui:param name="anonymizedAccess" value="#{DatasetPage.anonymizedAccess}"/>
             <ui:param name="anonymizedAccessEnabled" value="#{DatasetPage.anonymizedAccessEnabled}"/>
             <ui:param name="canUpdateDataset" value="#{DatasetPage.canUpdateDataset()}"/>


qqmyers:
A bug leading to a DatasetsIT test failure seen on this PR (run 9) is hopefully fixed in #12128. I have not updated the code here but did start a rerun.

src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -864,9 +873,13 @@ private String generateInfoFile() {
         if (aggregation.has(JsonLDTerm.schemaOrg("includedInDataCatalog").getLabel())) {
             catalog = aggregation.get(JsonLDTerm.schemaOrg("includedInDataCatalog").getLabel()).getAsString();
         }
-        info.append(catalog + ":" + aggregation.get(JsonLDTerm.schemaOrg("name").getLabel()).getAsString());
+        catalog=catalog.trim().replaceAll("[\\r\\n:]","_");
+        info.append(catalog + ":" + multilineWrap(aggregation.get(JsonLDTerm.schemaOrg("name").getLabel()).getAsString()));
         info.append(CRLF);
 
+        //Add a version number for our bag type - should be updated with any change to the bag content/structure
+        info.append("Dataverse-Bag-Version: 1.0");
```
stevenwinship, line 881: I'd like to see this as a final static string defined at the top of the class

stevenwinship: I'll approve without this being changed

qqmyers: This was fixed in #12133 - I won't change it here, but it's in the final PR.

PR #12067, opened 2026-01-03
Feature: Optionally make embargo reason required

### LINKED ISSUE #8692: Set embargo date API results in internal server error
I used this curl command in the command line:

curl -H "X-Dataverse-key: $API_TOKEN" -H "Content-Type:application/json" "$SERVER_URL/api/datasets/:persistentId/files/actions/:set-embargo?persistentId=$PERSISTENT_IDENTIFIER" -d "$JSON" -v

* What happens?
I get an internal server error back:

* upload completely sent off: 78 out of 78 bytes
< HTTP/1.1 500 Internal Server Error
< Date: Wed, 11 May 2022 14:27:47 GMT

I used an API-token of a superuser. 

version 5.10.1 (on our DataverseNL test server.)

### LINKED ISSUE #11956: Feature Request: Modifications to embargo function
We suggest some improvements to the embargo function:
1. When a file is made unavailable with the "restrict" feature, in the file overview this is clearly visible with a grey tag "Not available until XXXX-XX-XX". This is not the case when a file is made unavailable using the embargo function however (it is indicated, but much more discretely). Could the same or similar tag be added automatically to embargoed files?
2. When placing an embargo on a file, the depositor is asked to specify why "Enter a short reason why this embargo exists". This however is currently not a mandatory field. Could it be made so? This would be more in line with the principle of "as open as possible, as closed as necessary", i.e. the deponent must justify why it is necessary to kept the data closed.
3. When an image file (e.g. .jpg) is embargoed, the image still appears as a thumbnail in the file overview. Even worse, when you click to open the file preview, a larger version of the image previews. So the embargo function does not really work as intended for image files.

These improvements to the embargo function will be relevant for several types of users, including depositors, curators, and future users of the dataset.

At DataverseNO, which is a repository for open research data, we would like to switch from using the "restricted files" options, which enables files essentially to be made unavailable indefinitely, to embargo, which would allow files to be temporarily unavailable for a maximum period of time. This has led us to check out this feature in more depth and come with some suggestions for improvement.

Not that I'm aware of.

We would like to sponsor this issue, but of course would like to hear any feedback first.

qqmyers:
**What this PR does / why we need it**: This PR adds a flag that, when set, makes it required to set the reason when creating an embargo, via the UI or API. In addition it adds a new check to assure that a supplied reason, whether required or not, is not blank.

- Closes #11956
- Closes #8692

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -1532,9 +1532,9 @@ public class Datasets extends AbstractApiBean {
             reason = json.getString("reason");
         }
         if(reason == null && FeatureFlags.REQUIRE_EMBARGO_REASON.enabled()) {
-            return error(Status.BAD_REQUEST, "Reason is required for embargoes");
+            return error(Status.BAD_REQUEST, BundleUtil.getStringFromBundle("embargo.reason.required"));
         } else if(reason != null && reason.isBlank()) {
-            return error(Status.BAD_REQUEST, "Reason cannot be blank (whitespace only)");
+            return error(Status.BAD_REQUEST, BundleUtil.getStringFromBundle("embargo.reason.blank"));
         }
         embargo.setReason(reason);
 


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1520,8 +1528,16 @@ public Response createFileEmbargo(@Context ContainerRequestContext crc, @PathPar
                 return error(Status.BAD_REQUEST, "Date available can not exceed MaxEmbargoDurationInMonths: "+maxEmbargoDurationInMonths);
             }
         }
-
-        embargo.setReason(json.getString("reason"));
+        String reason = null;
+        if(json.containsKey("reason")) {
+            reason = json.getString("reason");
+        }
+        if(reason == null && FeatureFlags.REQUIRE_EMBARGO_REASON.enabled()) {
```
sekmiller, line 1534: Minor point, but would we want to check for blank ((null or blank) and flag) here when the reason is required to return the reason is required message?

qqmyers: The next check is for blank - just a slightly different error message.

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -1520,8 +1528,16 @@ public Response createFileEmbargo(@Context ContainerRequestContext crc, @PathPar
...
+            return error(Status.BAD_REQUEST, "Reason is required for embargoes");
+        } else if(reason != null && reason.isBlank()) {
+            return error(Status.BAD_REQUEST, "Reason cannot be blank (whitespace only)");
+        }
+        embargo.setReason(reason);
```
sekmiller, line 1539: I guess it's out of scope for this, but did you consider using a command? (To do permission checks and even apply the FileUtil validation method.)

src/main/java/propertyFiles/Bundle.properties
```diff
@@ -41,6 +41,8 @@ embargoed.wasthrough=Was embargoed until
 embargoed.willbeuntil=Draft: will be embargoed until
 embargo.date.invalid=Date is outside the allowed range: ({0} to {1})
 embargo.date.required=An embargo date is required
+embargo.reason.required=An embargo reason is required
+embargo.reason.blank=An embargo reason cannot be blank
```
sekmiller, line 45: These could be used in the Datasets method

src/test/java/edu/harvard/iq/dataverse/api/DatasetsEmbargoAPITest.java
```diff
@@ -0,0 +1,246 @@
...
+
+import static jakarta.ws.rs.core.Response.Status.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+@LocalJvmSettings
+public class DatasetsEmbargoAPITest {
```
sekmiller, line 44: I've been writing integration tests for my api endpoints and unit tests for commands. Beyond not having to deal with the database is there a reason for testing like this?

qqmyers: My understanding is you can't easily test FeatureFlags (or their non-default value) in IT tests since we don't have a way to change the flag remotely from the test. This style of test allows changing the flag value.

sekmiller: Right. I've run into issues with setting feature flags for integration tests. Thanks!

pdurbin: Related:

- #10128

src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
qqmyers: For this PR I was just following the existing code, but yeah, that could have been done in a command. I'm not sure I understand about the FileUtil method - that seems pretty tied to JSF.

sekmiller (approved):

PR #12081, opened 2026-01-14
reExportAll Improvements

qqmyers:
**What this PR does / why we need it**:  This PR makes several improvements to the admin/metadata/reExportAll API call:
- Fatal run time problems, such as a problem with the database connection (seen at DANS) get logged to the export log before terminating, making it easier to see when there has been a problem
- a new optional ?olderThan=`<YYYY-MM-DD>` parameter has been introduced to allow incremental reExportAll, i.e. allowing reExportAll to be done in smaller chunks, or to allow reExporting datasets last updated before some exporter change occurred, or to restart after a failure (use olderThan=<the date or the original failed run>)
- Explicit marking of the overall exportAllDatasets method as not supporting a transaction to avoid pulling all datasets into one transaction (the exportAllFormatsInNewTransaction(dataset) adds datasets to individual transactions as they are processed)
- Unused fallback code that would write exportLogger info to the main server.log file has been removed

- Closes #

**Special notes for your reviewer**: As most of the changes are to error handling or, with the olderThan date param, would require testing over multiple days, I don't see any new tests to add. The existing tests should assure regression testing.

**Suggestions on how to test this**: Rely on the DatasetsIT tests for regression testing, test performance on large databases (hopefully not having a single overall transaction will noticeably improve memory use/performance). The new olderThan param can be tested on an existing db with a range of lastExport dates. W.r.t. error handling, the only thing I can think to try (if desired) would be to kill the database during a reExportAll run and verify that the export log has logging of that problem. (Would have to restart the db and payara after that I assume.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
@@ -738,11 +738,11 @@ public class DatasetServiceBean implements java.io.Serializable {
         try {
             fileHandler = new FileHandler(logFileName);
             exportLogger.setUseParentHandlers(false);
+            exportLogger.addHandler(fileHandler);
         } catch (IOException | SecurityException ex) {
             Logger.getLogger(DatasetServiceBean.class.getName()).log(Level.SEVERE, null, ex);
             return;
         }
-
         exportLogger.info("Starting an export all job");
 
         for (Long datasetId : findAllLocalDatasetIds()) {


src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -715,30 +722,27 @@ public void exportAll() {
...
         Integer countAll = 0;
         Integer countSuccess = 0;
         Integer countError = 0;
         String logTimestamp = logFormatter.format(new Date());
         Logger exportLogger = Logger.getLogger("edu.harvard.iq.dataverse.harvest.client.DatasetServiceBean." + "ExportAll" + logTimestamp);
         String logFileName = System.getProperty("com.sun.aas.instanceRoot") + File.separator + "logs" + File.separator + "export_" + logTimestamp + ".log";
         FileHandler fileHandler;
-        boolean fileHandlerSuceeded;
         try {
             fileHandler = new FileHandler(logFileName);
```
stevenwinship: I see the fileHandler being created and later closed. I don't see it being used anywhere in this code.

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -778,10 +790,7 @@ public void exportAllDatasets(boolean forceReExport) {
         exportLogger.info("Datasets failures: " + countError.toString());
         exportLogger.info("Finished export-all job.");
 
-        if (fileHandlerSuceeded) {
-            fileHandler.close();
-        }
-
+        fileHandler.close();
```
stevenwinship, line 800: remove

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
```diff
@@ -768,6 +773,13 @@ public void exportAllDatasets(boolean forceReExport) {
                         } catch (Exception ex) {
                             exportLogger.log(Level.INFO, "Error exporting dataset: " + dataset.getDisplayName() + " " + dataset.getGlobalId().asString() + "; " + ex.getMessage(), ex);
                             countError++;
+                        } catch (Throwable t) {
+                            exportLogger.log(Level.SEVERE, "Fatal error exporting dataset: " + dataset.getDisplayName() + " " + dataset.getGlobalId().asString() + "; " + t.getClass().getName() + ": " + t.getMessage(), t);
+                            exportLogger.info("Datasets processed before fatal error: " + countAll.toString());
+                            exportLogger.info("Datasets exported successfully: " + countSuccess.toString());
+                            exportLogger.info("Datasets failures: " + countError.toString());
+                            fileHandler.close();
```
stevenwinship, line 788: remove fileHandler.close();

src/main/java/edu/harvard/iq/dataverse/DatasetServiceBean.java
qqmyers: good catch - line 748 was still needed.

stevenwinship (approved):

PR #12082, opened 2026-01-14
Improve Permission Indexing Performance

qqmyers:
**What this PR does / why we need it**: #11822 in v6.9 made changes that significantly sped permission reindexing across many datasets (e.g. when someone is granted new roles on the root dataverse). However, in doing so, it introduced an additional find(dataset_id) that slowed cases where content and permission indexing were done on one dataset. This PR, from discussion with @landreev (who identified the issue), refactors to avoid requiring a find that essentially reloads the dataset/versions/filemetadatas unnecessarily. 

It does this by passing a list of datasetversions rather than a dataset id to indexDatasetFilesInNewTransaction. Since that method is in a new transaction, and we want to avoid merging the versions into the new transaction context for performance reasons (and its not really needed since we are not writing any changes to them), there is also new code that assures the fileMetadatas are loaded before the method is called in the case where they will be used (there is a JvmSettings.MIN_FILES_TO_USE_PROXY setting to avoid pulling fileMetadatas from the db when there are many files)

- Closes #

**Suggestions on how to test this**: Regression testing - permission indexing should work as before in all cases, calling /admin/index/status shouldn't show any (new?) issues related to permission docs. For performance, there are two main cases:
- adding a role for someone on the root collection, triggering (only) permission indexing across many dvobjects
- editing a dataset with many files

In both cases, the PR should improve performance over 6.9 - not sure yet how big the change in the first case is. The latter should restore the pre-6.8 performance or improve on it.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
@@ -382,7 +382,7 @@ public class SolrIndexServiceBean {
             for (DatasetVersion version : versionsToReIndexPermissionsFor(dataset)) {
                 if (desiredCards.get(version.getVersionState())) {
                     int fileCount = dataFileService.findCountByDatasetVersionId(version.getId()).intValue();
-                    if (fileCount >= fileQueryMin) {
+                    if (fileCount <= fileQueryMin) {
                         // IMPORTANT: This triggers the loading of fileMetadatas within the current transaction
                         version.getFileMetadatas().size();
                     }


landreev:
> there is a JvmSettings.MIN_FILES_TO_USE_PROXY setting to avoid pulling fileMetadatas from the db when there are many files

Outside the scope of the PR, but this reminded me that I never got around to experimenting with this setting. What would be a sensible number to try - 1K maybe?

landreev:
> In both cases, the PR should improve performance over 6.9 - not sure yet how big the change in the first case is. 

In the first case, i.e. - a permission reindexing of all the datasets in root after a role assignment - it took a bit longer than the result I recorded back in Sept. when testing #11822. I assumed it was on account of the database growth, but didn't confirm that for sure. (the logs had been erased since; I could easily find how many datasets/files had been created since then; but would have been a bit less trivial to find out how many had been deleted). 

If you are really curious to know the answer, I can run it under 6.9 vs this PR on the same db.

landreev:
Since I have already invested heavily into testing this branch, I will petition/volunteer myself to QA and merge the PR. 
I am curious to re-test perm reindex with https://github.com/IQSS/dataverse/pull/12082/commits/a403cb0f27971cae90ccdc10e2614869b77cc18f added since, with/without some MIN_FILES_TO_USE_PROXY configured. 
Also debating if I should run another full reindex (= an extra ~day of cpu time on qa), out of curiosity and/or abundance of caution.

src/main/java/edu/harvard/iq/dataverse/search/SolrIndexServiceBean.java
```diff
@@ -365,14 +368,24 @@ public IndexResponse indexPermissionsOnSelfAndChildren(DvObject definitionPoint)
...
+             * 
+             **/
             Map<DatasetVersion.VersionState, Boolean> desiredCards = searchPermissionsService.getDesiredCards(dataset);
             List<DatasetVersion> versionsToIndex = new ArrayList<>();
             for (DatasetVersion version : versionsToReIndexPermissionsFor(dataset)) {
                 if (desiredCards.get(version.getVersionState())) {
-                    // IMPORTANT: This triggers the loading of fileMetadatas within the current transaction
-                    version.getFileMetadatas().size(); 
+                    int fileCount = dataFileService.findCountByDatasetVersionId(version.getId()).intValue();
+                    if (fileCount >= fileQueryMin) {
```
landreev: Sorry, is this correct?

qqmyers: Absolutely not! :-) Good catch. I recall going back to add the = sign, but mistyped/missed the > was wrong to start with.

qqmyers:
> > In both cases, the PR should improve performance over 6.9 - not sure yet how big the change in the first case is.
> 
> In the first case, i.e. - a permission reindexing of all the datasets in root after a role assignment - it took a bit longer than the result I recorded back in Sept. when testing #11822. I assumed it was on account of the database growth, but didn't confirm that for sure. (the logs had been erased since; I could easily find how many datasets/files had been created since then; but would have been a bit less trivial to find out how many had been deleted).
> 
> If you are really curious to know the answer, I can run it under 6.9 vs this PR on the same db.

The bug you found would probably have loaded fmds unnecessarily for that case, so maybe that increased the overall time. 
I don't think the exact answer is important as long as it's clear that there's no huge decrease.

qqmyers:
> > there is a JvmSettings.MIN_FILES_TO_USE_PROXY setting to avoid pulling fileMetadatas from the db when there are many files
> 
> Outside the scope of the PR, but this reminded me that I never got around to experimenting with this setting. What would be a sensible number to try - 1K maybe?

I think I suggested 200 somewhere. I don't think the exact number is critical.

landreev:
> The bug you found would probably have loaded fmds unnecessarily for that case, so maybe that increased the overall time.

To clarify, I was referring to my testing last week, before a403cb0f27971cae90ccdc10e2614869b77cc18f was made. 
That's why I want to re-test, maybe w/ different batch sizes configured.

landreev:
> To clarify, I was referring to my testing last week, before [a403cb0](https://github.com/IQSS/dataverse/commit/a403cb0f27971cae90ccdc10e2614869b77cc18f) was made.
> That's why I want to re-test, maybe w/ different batch sizes [file proxy limits rather] configured.

I just re-ran the reindexing-after-role-assignment test, _without_ any MIN_FILES_TO_USE_PROXY configured. 
Were there any reasons to expect the changes in a403cb0 to speed things up **3X+**, even without taking advantage of file file proxy? - that is what I'm seeing now.  

This doesn't make too much sense. I may have screwed something up in my testing. But the logging shows the right number of files/datasets processed.

landreev:
Ok, n/m, I did in fact have the JVM setting all along, set to 200 per your recommendations.

landreev:
Testing permission reindex of the prod. root collection after a role assignment;  1824420 files, 32217 datasets.
A huge improvement from the latest tweak: 

1. The branch in its current form, with min-files-to-use-proxy=200: 32 min. 
2. Stock 6.9, same jvm setting: 106 min. 
3. Current branch, without the jvm setting configured: 115.5 min. 

4. Plus, same test performed last week (with 1815504 files, 32124 datasets):
The branch prior to a403cb0, w/ min-files-to-use-proxy set (meaning, it was using the file proxy within, but was always initializing all the filemetadatas outside of the inner transaction - ?): 114 min.  

The end result of the PR, (1.) makes perfect sense in retrospect, as .getFileMetadatas() on large datasets is a known performance bottleneck. 

The fact that 2. was faster than 4. ... makes sense too, I think?

> this reminded me that I never got around to experimenting with this setting

I was wrong obviously. I did "get around to experimenting" with it, and the setting had been in place since I tested the PR where it was introduced.

landreev:
A summary of the end result of merging the PR: 

Reindexing of permissions only on a large collection, after a role assignment: much (3+X) better than 6.9, infinitely better than 6.8 for HDV root, or a similar collection. 
Full reindexing of datasets: much better than 6.9; although still a hair slower than 6.8. (apparently only for datasets w/ min-files-to-use-proxy or fewer files, set to 200 for the test). The impact is not significant enough to justify more work at this point, IMO but should be worth documenting as extra data for future dev. 

The impact in the context of a full reindex of HDV: 

| version  |       <=200 files  |   >200 files  |    Total        |
|---------|-----------------|-------------|------------|
| 6.8         |    21h 15m          |  7h 5m         |   28h 20m  |
| PR 12082 |    22h 40m         |  7h 5m         |   29h 45m  |

12393 collections, 254008 datasets, 6256315 files.

250138 datasets w/ <= 200 files, 1726520 files total
3870 datasets w/ > 200 files, 4529795 files total

landreev:
Since I was already geeking out/being OCD about it, extracted a bit more info from the logs re: slowdown observed 6.8 vs. the pr when reindexing datasets w/ 200 or fewer files: 

|          | 0 files | 1-10 files | 11-100 files | 101-200 files |
|----------|---------|------------|--------------|---------------|
| 6.8      | 2h 49m  | 13h 34m    | 4h 0m        | 50m           |
| PR 12082 | 2h 56m  | 14h 37m    | 4h 13m       | 51m           |
| slowdown | 4%	     | 8%     	  | 5%	      	 | 2% 	      	 |


The breakdown per database objects:

|          | 0 files | 1-10 files | 11-100 files | 101-200 files |
|----------|---------|------------|--------------|---------------|
| datasets | 51K     | 166K       | 31K          | 3K            |
| files    | -       | 449K       | 883K         | 396K          |

One low-effort thing I may try at some point - see what happens if we only call `self.indexDatasetBatchInNewTransaction()` w/ min-files-to-use-proxy+ files, and otherwise call it w/out the "self." part.

PR #12093, opened 2026-01-16
Indexing Simplifications

qqmyers:
**What this PR does / why we need it**: This PR provides a simpler mechanism to get the versions to index relative to the logic in getDesiredCards(). It also drops a ~duplicate method for getting the desired versions (one was used for content indexing, the other for permissions) and adds a unit test to verify that new code retrieves the expected versions in various cases.

**Which issue(s) this PR closes**:

- Closes #

**Special notes for your reviewer**: This is built on #12082 which will hopefully be merged soon. It may be easier to look at the two commits if you review before that PR merges.

**Suggestions on how to test this**: Regression-only - should not change any functionality.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

PR #12094, opened 2026-01-16
Permission indexing creates unused, orphan perm docs for unchanged files in draft versions

qqmyers:
**What this PR does / why we need it**: At some point, content indexing was updated such that files in a released version, whose metadata has not changed in a draft version, are no longer indexed in the draft version (so you could still find the file from the released version, but would not know it existed in a newer draft as well from the search results alone - a minor inconvenience to avoid doubling the cost of indexing files when new drafts are created.)

However, permission indexing was not updated to match and thus, when permissions are indexed, files in this situation get perm docs for both the released version (where a matching content doc exists) and the draft version (where no content doc has been created). This PR fixes that and uses the same approach in permission indexing to not perm reindex files if their metadata does not change.

- Closes #

**Special notes for your reviewer**: The PR builds on #12082 and #12093 which can hopefully be merged first. The final fix here is relatively straight forward - just ~reusing the existing logic from the content indexing - I did change the query to return file ids rather than filemetadata ids since we don't have easy access to the filemetadata ids on the permission side due to earlier optimizations.

**Suggestions on how to test this**: Primarily regression testing - visible indexing and search shouldn't change. The one effect that can be seen is that there will no longer be perm docs for a files in a post-release draft version that don't have metadata changes. That can be seen via the Solr console/direct query to Solr.

**Additional documentation**:


landreev:
I'm ready to approve this, but will wait until the other 2 PRs are merged, to avoid any confusion.

PR #12101, opened 2026-01-21
Call Archiving commands asynchronously from UI to avoid timeouts

qqmyers:
**What this PR does / why we need it**: When archiving takes too long (e.g. a 10K file dataset), launching it from the UI can cause a timeout/rollback. This PR updates the code to archive a version from the versions table to run asynchronously and to handle any OptimisticLockException if the dataset changes before the archiving in the background completes.

- Closes #

Note: - another PR is coming that will use the asyncCommand call when doing an 'Update-Current-Version' publish action.

**Suggestions on how to test this**: Regression - confirm that calling an archiving workflow from the versions table still works. Not sure it's practical to try archiving something big enough to see the timeout issue with the current code.

**Is there a release notes update needed for this change?**: Will create an aggregate one.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/EjbDataverseEngine.java
```diff
@@ -348,6 +352,28 @@ public <R> R submit(Command<R> aCommand) throws CommandException {
...
         }
     }
+
+    /**
+     * Submits a command for asynchronous execution.
+     * The command will be executed in a separate thread and won't block the caller.
+     * 
+     * @param <R> The return type of the command
+     * @param aCommand The command to execute
+     * @param user The user executing the command
```
stevenwinship, line 362: Can be removed - no user in signature

qqmyers: fixed in #12167

PR #12103, opened 2026-01-21
Allow archiving when using 'Update-Current-Version'

qqmyers:
**What this PR does / why we need it**: Using 'Update Current Version' allows admins to replace an existing version (versus creating a new version). The logic for that command is capable of re-archiving the version if the old archiving files have been deleted from the archival repository, but would fail of the files from the initial publication existed.

This PR adds a new method to indicate if archivers are capable of deleting the old copy and, if so, to allow re-archiving to succeed in the update-current-version case (by triggering the delete of the old files as part of archiving.
Since re-archiving a large dataset automatically when fixing a small typo may not be worthwhile, the PR includes a flag to optionally stop Update Current Version from re-archiving and, instead, just marking the archival copy as existing but 'obsolete'.
The PR also updates the version table to show this info and to allow manually triggering re-archiving in the 'obsolete' case (along with the cases when archiving has not been done or has failed.)

- Closes #

This builds on/includes the changes from #12101 (the first 4 commits)

**Suggestions on how to test this**: Configure an archiving workflow e.g. using the Local /file based archiver. Publish a version and then make a small update and use Update Current Version w/ and w/o the flag. When the workflow runs, it should succeed (overwriting the earlier datacite.xml and bag for that version) and when it doesn't the status shown on the dataset page version table (for superusers) should show an obsolete status and allow you to submit manually if you want to update the bag.

**Is there a release notes update needed for this change?**: Will add aggregate one (in #12122)

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -2067,6 +2067,7 @@ public Response submitDatasetVersionToArchive(@Context ContainerRequestContext c
             if(dv==null) {
                 return error(Status.BAD_REQUEST, "Requested version not found.");
             }
+            //ToDo - allow forcing with a non-success status for archivers that supportsDelete()
```
qqmyers, line 2070: FWIW: Since you can set the archival status to null via API, it would just be a convenience to be able to force re-archiving in one call - hence a note in the code but not planning to do it for this PR.

stevenwinship (approved):

PR #12118, opened 2026-01-27
Reduce Schema.org jsonld for datasets with many files

qqmyers:
**What this PR does / why we need it**: At QDR, we noticed a dataset with ~10K files was not getting indexed by Google. Investigating, we say that Google was failing to read the schema.org information in the page header, apparently due to it's length.

After finding some guidance on an alternate way to provide file download URLs - https://github.com/ESIPFed/science-on-schema.org/blob/main/guides/Dataset.md#accessing-data-through-a-service-endpoint - I went ahead and implemented that approach. It essentially replaces writing a json object per file with one object and a list of fileIds which is significantly shorter.

Subsequent testing at QDR shows that this was effective in allowing Google to index that dataset again.

- Closes #

**Special notes for your reviewer**: posting as a draft in case people want to go forward with it while I'm out. Probably needs a release note about the new jvmoption and doc updates for it. It should probably go in the new https://guides.dataverse.org/en/latest/admin/big-data-administration.html as well.

**Suggestions on how to test this**: Should be able to set the flag to a low number to trigger the new format with just a few files and could then use Google's tools to verify it can parse the output (I forget which tool/the URL for it).

**Additional documentation**:


pdurbin:
Related:

- https://github.com/mlcommons/croissant/issues/646
- #12123
- #12191

rtreacy (commented):
Claude
Overview

Addresses a real production issue (QDR: Google failing to index a ~10K-file dataset due to oversized Schema.org JSON-LD). Instead of emitting one verbose DataDownload object per file in distribution, datasets exceeding a new configurable threshold (dataverse.exports.schema-dot-org.max-files-for-download-entries, default unlimited) instead emit a single potentialAction/SearchAction with a urlTemplate and a regex valuePattern enumerating file IDs. Small, targeted diff: DatasetVersion.java, a new JvmSettings entry, docs, and a release note. No test changes.

Issues found

1. Inconsistent handling of HIDE_SCHEMA_DOT_ORG_DOWNLOAD_URLS (correctness) — DatasetVersion.java:2131
The existing distribution path respects hideFilesBoolean (suppresses contentUrl when the admin sets dataverse.files.hide-schema-dot-org-download-urls=true). The new potentialAction branch ignores this setting entirely — it always emits the /api/access/datafile/{fileId} URL template plus every file ID, regardless of the flag. An installation that disabled download URLs for smaller datasets gets them back for larger ones. This should check the same setting (and probably fall back to a plain distribution-less object, or omit potentialAction, when hiding is requested).

2. Wrong list used for the file-ID pattern (correctness) — DatasetVersion.java:2140
The size check uses fileMetadatasSorted, but the loop that builds valuePattern iterates the raw fileMetadatas instance field instead of fileMetadatasSorted. Looks like a copy/paste slip — it happens to work today, but it's inconsistent with the guarded variable and fragile if the two lists ever diverge (sorting, filtering, or future refactors of getFileMetadatasSorted()). Should be fileMetadatasSorted for consistency with the surrounding code.

3. No test coverage for the new path (test-coverage) — DatasetVersionTest.java / SchemaDotOrgExporterTest.java
Neither test file was touched. Existing getJsonLd() tests only cover datasets with no files at all, so nothing exercises the new potentialAction branch, the threshold comparison boundary, or interaction with HIDE_SCHEMA_DOT_ORG_DOWNLOAD_URLS. Given SchemaDotOrgExporterTest.java is explicitly called out in a comment as "the place for more extensive tests," this is a real gap — worth at least one test with file count > threshold verifying potentialAction structure and one at/under threshold verifying distribution is still used.

4. Stray indentation left over from re-wrapping (style) — DatasetVersion.java:2117
The fileObject.add("contentUrl", ...) line wasn't re-indented when the surrounding block gained an extra nesting level (the new if (fileMetadatasSorted.size() <= maxFilesForDownloadEntries) wrapper). Minor, but mvn spotless:apply should catch this before merge. Also note trailing-whitespace-only blank lines added in JvmSettings.java around the new EXPORTS block.

5. Still O(n) in file count (efficiency, minor) — DatasetVersion.java:2143
The fallback trades a verbose per-file JSON object for a compact ID, which is a big win, but valuePattern still grows linearly with file count (a huge |-joined regex for e.g. 100K files). Not a blocker — the QDR field test at ~10K files confirms this works well in practice — but worth a one-line callout in the PR/docs that this is a mitigation, not a fully constant-size solution, in case someone hits an even larger dataset later.

Other notes

- Endpoint path /api/access/datafile/{fileId} used in the urlTemplate is correct per Access.java's @Path("datafile/{fileId:.+}").
- Default (Integer.MAX_VALUE via orElse) preserves current behavior for anyone not setting the new option — good backward compatibility.
- Docs (config.rst, native-api.rst) and release note are well-written and correctly cross-referenced; nice that the PR author's own draft-PR TODO ("probably needs docs") was actually done in this diff.
- Consider also documenting the new setting in big-data-administration.html guide as the author suggested in the PR description — that wasn't done here.

Overall: solid, well-motivated fix for a real scalability bug. The two correctness issues (missing hide-URLs check, wrong list variable) should be fixed before merge; test coverage should be added given this is user-facing SEO-affecting output.

rtreacy:
@qqmyers , can you take a look at the issues found section of the comments from Claude?

qqmyers:
@rtreacy - I've addressed relevant AI points (1-4) - I did not add a warning about O(n) efficiency (point 5) - overkill as O(n) behavior with files is true widely in the codebase, but did add an entry in the big-data-administration guide (3rd Other Note).

I also merged and fixed newly added Json.* calls which are forbidden (so it should now pass that check).

rtreacy (approved):
previous review comments have been addressed

PR #12121, opened 2026-01-28
Improve PID update performance

qqmyers:
**What this PR does / why we need it**: The original code (from 8 yrs ago) was doing an em.merge and em.flush after every file PID update. This PR uses the merged dataset (which should assure the files are merged) and relies on the automatic flush at the end of the transaction instead, avoiding the extra db traffic.

- Closes #

**Special notes for your reviewer**: FWIW: I have not tried to perf test this to know if it is significant but it seems like useful cleanup.

**Suggestions on how to test this**: ~regression test - file pids and the globalidcreatetime updated on them when the /api/datasets/<id>/modifyRegistrationMetadata is called or when the update-current-version option is used for publication as before, recognizing that the code now checks to make sure the DataCite xml is different before updating it so some change to the list of dataset creators, or file name, mimetype is needed. Adding an extra author/creator is probably the easiest test.

**Is there a release notes update needed for this change?**: seems pretty small to be listed in the notes

**Additional documentation**:


rtreacy (approved):
Looks good. Removes final from target field. Replaces merge(); flush(); with all or nothing merge

qqmyers:
@sekmiller - tests are passing now - just needed to merge

PR #12128, opened 2026-01-29
Fix for ConcurrentMod Exception seen in DatasetsIT test

qqmyers:
**What this PR does / why we need it**: I saw a500 error in at `edu.harvard.iq.dataverse.api.DatasetsIT.testSummaryDatasetVersionsDifferencesAPI(DatasetsIT.java:6629) `on #12063 and think I've seen it before. Looking in the server log I saw
```
 [#|2026-01-28T21:41:52.572+0000|SEVERE|Payara 6.2025.10|edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler|_ThreadID=70;_ThreadName=http-thread-pool::http-listener-1(3);_TimeMillis=1769636512572;_LevelValue=1000;|
  _status="ERROR";_code=500;_message="Internal server error. More details available at the server logs.";_incidentId="2c8887ad-d870-4183-ac5d-7aae8663e609";_internalError="ConcurrentModificationException";_requestUrl="http://localhost:8080/api/v1/datasets/253/add";_requestMethod="POST"||#]

[#|2026-01-28T21:41:52.573+0000|SEVERE|Payara 6.2025.10|edu.harvard.iq.dataverse.api.errorhandlers.ThrowableHandler|_ThreadID=70;_ThreadName=http-thread-pool::http-listener-1(3);_TimeMillis=1769636512573;_LevelValue=1000;|
  java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:967)
	at edu.harvard.iq.dataverse.FileMetadata.getCategoriesByName(FileMetadata.java:291)

```
and investigating with some AI help, it appears that the ~root problem is that the code was modifying a (managed) array that may be used in multiple threads.

The fix here should prevent that via synchronization and avoiding using the same list in new copies of the FileMetadata while still addressing the issue of seeing an IndirectList that led to the original code.

- Closes #

**Special notes for your reviewer**: Aside from adding synchronization, this PR makes some changes to code that was copying from one filemetadata to another - reusing/extending the createCopy instead of having redundant code. (createCopyInVersion now makes sure the calls to set the version in the fmd and to add the fmd to the version's list are done rather than making those separate calls). That fixes a couple cases where there was a fmd2.setCategories(fmd1.getCategories()) (sharing the same managed array across fmds) which should be OK. I also realized the createCopy wasn't copying any prov entry which I assume it should have been doing?

**Suggestions on how to test this**: Regression-only -we shouldn't see any new errors. The IT tests should get this error any more, but it was pretty rare already.

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/DatasetVersionDifferenceTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/DatasetVersionDifferenceTest.java
@@ -76,7 +76,8 @@ public class DatasetVersionDifferenceTest {
         datasetVersion2.setVersionState(DatasetVersion.VersionState.DRAFT);
         datasetVersion2.setTermsOfUseAndAccess(new TermsOfUseAndAccess());
         datasetVersion2.getTermsOfUseAndAccess().setLicense(license);
-
+        datasetVersion.setFileMetadatas(new ArrayList<>());
+        
         // Published version's two files
         DataFile dataFile = new DataFile();
         dataFile.setId(1L);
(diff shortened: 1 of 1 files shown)


stevenwinship:
I tested both the old code and new code with simultaneous tests (DatasetsIT.testSummaryDatasetVersionsDifferencesAPI) running in multiple windows. I didn't see any errors in either version. No regression but worth monitoring and noting future issues.

PR #12131, opened 2026-01-30
Support "holey" archival bags

qqmyers:
**What this PR does / why we need it**: Previously, Dataverse always archived a dataset as a single zipped bag file whose size was limited only by the (compressed) size of the metadata and data. This PR adds support for the "holey" bag mechanism in which some/all datafiles are not sent in the bag but are instead listed in a fetch.txt file. To complete the bag, the receiver must read the fetch file, retrieve the listed files from their URLs and place them at the specified location (path/filename).

Whether files are included is determined by two new jvm options:
```
dataverse.bagit.holey.max-file-size
dataverse.bagit.holey.max-data-size
```
which take the largest allowed (uncompressed) data file size and the max aggregate (uncompressed) size that should be zipped.
Files are now processed in order of increasing size which means that the zip will include the largest number of files possible if the max-data-size limit is used.

- Closes #
DANS DD-2157

This builds on #12063. 
The internal BagGenerator was recently updated to include the gbrecs parameter to suppress download counts when the downloads are for archival purposes. This PR also adds that parameter to the URLs in the fetch.txt file to assure that the third-party receiver doesn't accidentally trigger download counts.

In it's first iteration, this PR assumes that the receiver will retrieve the missing files directly from Dataverse, which means it may need an API key/other credential to get those files. The next step will be to add an option for those files to be transferred separately to the same place the bag is sent and to adjust the fetch file accordingly. This will assure that, with holey bags, completion of the archiving by Dataverse means that all data is in the receiving service. 

**Suggestions on how to test this**: Setup archiving and set either/both of the settings above and verify the split between files in the zip and listed in fetch.txt is correct.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/bagit/OREMap.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/bagit/OREMap.java
@@ -505,11 +505,16 @@ public class OREMap {
                 for (String prefix : context.keySet()) {
                     localContext.putIfAbsent(prefix, context.getString(prefix));
                 }
-                JsonObjectBuilder job = Json.createObjectBuilder(datasetFieldService.getExternalVocabularyValue(val));
-                job.add("@id", val);
-                JsonObject extVal = job.build();
-                logger.fine("Adding: " + extVal);
-                vals.add(extVal);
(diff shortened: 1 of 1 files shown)


qqmyers:
wrong base branch

PR #12144, opened 2026-02-04
Support partially unzipped archival bags

qqmyers:
**What this PR does / why we need it**: Extending #12133, this PR leverages the logic there which allowed setting per file and per dataset size limits above which, datafiles would just be listed in the bag's fetch.txt file (creating a "holey" bag). This PR adds an option to go ahead and send the oversized and place them such that, when the zipped bag is unzipped you will have a complete bag (perhaps an "un-holey one?). This allows admins to avoid arbitrarily large zipped bags (and temp space to create them) while avoiding the need for some active process on the archival platform to retrieve additional (possibly restricted etc.) files before the bag is complete.

- Closes #

**Special notes for your reviewer**: Since this is intended to go out with 6.10, I've gone ahead and changed the .holey.* config options from #12133 (since they don't just apply to holey bags now) and will just document the final names.

**Suggestions on how to test this**: As this nominally affects/works with any/all archivers, one could test all the variants we have. I'd suggest Harvard do a basic test of the local file archiver and probably the DRS one (it has no changes but it uses the S3 archiver which has been updated) that was created specifically for Harvard use. (Nominally could just regression test the DRS one if there's no interest in using the bag size limits or holey bag concept.)

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
@@ -98,7 +98,11 @@ import jakarta.enterprise.inject.spi.CDI;
 public class BagGenerator {
 
     private static final Logger logger = Logger.getLogger(BagGenerator.class.getCanonicalName());
-
+    
+    static final String CRLF = "\r\n";
+    
+    protected static final int MAX_RETRIES = 5;
+    
     private ParallelScatterZipCreator scatterZipCreator = null;
     private ScatterZipOutputStream dirs = null;
 
@@ -319,7 +323,7 @@ public class BagGenerator {
         boolean first = true;
         for (Entry<String, String> pidEntry : pidMap.entrySet()) {
             if (!first) {
-                pidStringBuffer.append("\r\n");
+                pidStringBuffer.append(CRLF);
             } else {
                 first = false;
             }
@@ -334,7 +338,7 @@ public class BagGenerator {
         first = true;
         for (Entry<String, String> sha1Entry : checksumMap.entrySet()) {
             if (!first) {
-                sha1StringBuffer.append("\r\n");
+                sha1StringBuffer.append(CRLF);
             } else {
                 first = false;
             }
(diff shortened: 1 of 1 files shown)


doc/release-notes/12144-un-holey-bags.md
```diff
@@ -0,0 +1,21 @@
+This release contains multiple updates to the OAI-ORE metadata export and archival Bag output:
```
stevenwinship, line 1: This release note contains all the same information as the 12063 release note. Are you planning on consolidating or should you only include the new info in this doc?

src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -1009,75 +1285,114 @@ private HttpGet createNewGetRequest(URI url, String returnType) {
...
+     */
+    public InputStreamSupplier getInputStreamSupplier(final String uriString) {
 
         return new InputStreamSupplier() {
             public InputStream get() {
                 try {
                     URI uri = new URI(uriString);
-
                     int tries = 0;
                     while (tries < 5) {
```
stevenwinship: this "5" is used so often (I counted 7 in this method) that it should be a constant 'final int MAX_TRIES = 5;'

src/main/java/edu/harvard/iq/dataverse/util/bagit/BagGenerator.java
```diff
@@ -1009,75 +1285,114 @@ private HttpGet createNewGetRequest(URI url, String returnType) {
...
+                            logger.log(Level.SEVERE, "InterruptedIOException when retrieving file: " + uriString + " - Operation was interrupted", e);
+                            Thread.currentThread().interrupt(); // Restore interrupt status
                         } catch (IOException e) {
-                            // Retry if this is a potentially temporary error such
-                            // as a timeout
+                            // Retry if this is a potentially temporary error such as a timeout
                             tries++;
-                            logger.log(Level.WARNING, "Attempt# " + tries + " : Unable to retrieve file: " + uriString,
-                                    e);
+                            logger.log(Level.WARNING, "IOException when retrieving file: " + uriString + " (attempt " + tries + " of 5)", e);
```
stevenwinship: even this "5" should use constant

doc/release-notes/12144-un-holey-bags.md
qqmyers: #12167 adds a combined one and deletes all the individual ones.

PR #12152, opened 2026-02-06
fix: fix bug in notifications API related to SUBMITTEDDS notifications

vera:

While using the notifications API with `inAppNotificationFormat=true`, I saw a lot of notifications about supposedly deleted datasets being submitted for review:

image

However, I knew these datasets weren't actually deleted. After looking at the relevant Dataverse code, I believe I found a bug in `InAppNotificationsJsonPrinter`.

It tries to use the notification's `objectId` to look up the relevant dataset by ID, however, the notification's `objectId` is not a dataset ID. It's a dataset version ID:

https://github.com/IQSS/dataverse/blob/8c23bf2030152373b8c357207456b91b45be710e/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/SubmitDatasetForReviewCommand.java#L66

I've changed the code to correctly use the `objectId` to look up a dataset version by ID.

/

I believe this bug might not always show up, for example in simple test cases which only create a single dataset with a single version, because in that case, it's possible that the dataset version ID and dataset ID are in sync. I've therefore slightly extended the `InReviewWorkflowIT` test case to create a second dataset version to try and force the bug to show up.

In my tests, it's worked. Without the fix, the test failed:

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   InReviewWorkflowIT.testCuratorSendsCommentsToAuthor:459 1 expectation failed.
JSON path data[0].datasetPersistentIdentifier doesn't match.    
```

With the fix, the test succeeded.

/

I'm not sure this is a highly relevant bug, so maybe not.

**Additional documentation**:

/

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/util/json/InAppNotificationsJsonPrinterTest.java
+++ b/src/test/java/edu/harvard/iq/dataverse/util/json/InAppNotificationsJsonPrinterTest.java
@@ -402,13 +402,16 @@ public class InAppNotificationsJsonPrinterTest {
         userNotification.setObjectId(1L);
         userNotification.setRequestor(requestor);
 
+        DatasetVersion datasetVersion = mock(DatasetVersion.class);
         Dataset dataset = mock(Dataset.class);
         Dataverse owner = mock(Dataverse.class);
+
+        when(datasetVersion.getDataset()).thenReturn(dataset);
 
         when(dataset.getGlobalId()).thenReturn(testGlobalId);
         when(dataset.getDisplayName()).thenReturn("Submitted Dataset");
         when(dataset.getOwner()).thenReturn(owner);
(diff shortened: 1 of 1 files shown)


stevenwinship:
@vera 
Could you take a look at the failing test (InAppNotificationsJsonPrinterTest.testAddFieldsByType_submittedDs)?
I merged develop into this branch because of a small conflict (nothing to do with this failure)
The failing test works if I undo your change to the "addDatasetVersionFields(notificationJson, userNotification);" line.

vera:
> @vera Could you take a look at the failing test (InAppNotificationsJsonPrinterTest.testAddFieldsByType_submittedDs)? I merged develop into this branch because of a small conflict (nothing to do with this failure) The failing test works if I undo your change to the "addDatasetVersionFields(notificationJson, userNotification);" line.

Should be fixed now!

PR #9239, opened 2022-12-20
extract metadata (NcML XML) from NetCDF/HDF5 files, new "requirements" option for external tools

pdurbin:
**What this PR does / why we need it**:

From talking to researchers who use NetCDF and HDF5 files, we believe there is value in extracting metadata from those files to give some insight about their contents.

I plan to also make a pull request in the https://github.com/gdcc/dataverse-previewers repo to add a preview tool for the extracted file, which is in NcML format.

**Which issue(s) this PR closes**:

- Closes #9153
- Relates to #9242

**Special notes for your reviewer**:

Additional scope we could consider:

- an API endpoint to extract NcML from existing NetCDF and HDF5 files
- have the "requirements" option apply in more places than file level preview tools
- making the extraction feature modular

**Suggestions on how to test this**:

- try various NetCDF and HDF5 files and see if NcML aux files are created
- try the new "requirements" option for external tools
- use the proposed NcML previewer in this PR: https://github.com/gdcc/dataverse-previewers/pull/18

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

No but here's an example of two HDF5 files where one doesn't show the eyeball because it doesn't meet the requirement of having a aux file:

Screen Shot 2022-12-20 at 10 05 08 AM

**Is there a release notes update needed for this change?**:

Yes, included.

**Additional documentation**:

Included.

Issue #9153, opened 2022-11-09
Extract metadata from NetCDF and HDF5 files as XML in NcML format


pdurbin:
Assuming PR #9152 is merged we'll have a library in place to start extracting XML from NetCDF and HDF5 files.

The supported XML format is called NcML and is described here: https://docs.unidata.ucar.edu/netcdf-java/current/userguide/ncml_overview.html

Yesterday there was general agreement among devs that it would be fine to save the XML as a derivative or aux file.

This will open the door for previewing the file as raw XML to start.

Additionally, we could work on created a dedicated previewer that shows the data in a nicer way than raw XML.

The code we write will look something like this:

```
String ncml = netcdfFile.toNcml(file.getName());
```

Here's the output for an HDF5 file at `src/test/resources/hdf/hdf5/vlen_string_dset` (from the PR above):

```
<?xml version="1.0" encoding="UTF-8"?>
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" location="file:vlen_string_dset">
  <variable name="DS1" shape="4" type="String" />
</netcdf>
```

Here's part of the output for a NetCDF file at `src/test/resources/netcdf/madis-raob.nc` (also from the PR above):

```
<?xml version="1.0" encoding="UTF-8"?>
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" location="file:madis-raob.nc">
  <dimension name="recNum" length="1" isUnlimited="true" />
  <dimension name="manLevel" length="22" />
  <dimension name="sigTLevel" length="150" />
  <dimension name="sigWLevel" length="76" />
  <dimension name="mWndNum" length="4" />
  <dimension name="mTropNum" length="4" />
  <dimension name="staNameLen" length="50" />
  <dimension name="QCcheckNum" length="10" />
  <dimension name="QCcheckNameLen" length="60" />
  <dimension name="maxStaticIds" length="1000" />
  <dimension name="totalIdLen" length="50" />
  <dimension name="nInventoryBins" length="32" />
  <variable name="nStaticIds" shape="" type="int">
    <attribute name="_FillValue" type="int" value="0" />
  </variable>
  <variable name="staticIds" shape="maxStaticIds totalIdLen" type="char">
    <attribute name="_FillValue" value="" />
  </variable>
  <variable name="lastRecord" shape="maxStaticIds" type="int">
    <attribute name="_FillValue" type="int" value="-1" />
[...]

PR #12190, opened 2026-03-03
QDR: Public Auxfile Previewers

qqmyers:

This PR displays previewers that require auxiliary file(s) that are public even in cases where the user cannot download the datafile itself (i.e. it's restricted).

Dataverse allows upload of "auxiliary files" that are associated with a datafile, and it allows making auxiliary files public when the datafile itself is restricted/embargoed/beyond it's retention period. We also allow previewers and externalTools in general to 'require' auxiliary files and to display their content. However, the code cannot currently recognize that, when an auxiliary file is public, a tool requiring it can probably be used when the user does not have the ability to download the restricted datafile itself.

QDR has developed a sensitive data example with the QDAS previewers where this is relevant. QDAS codebooks can contain some entries that are more sensitive than others. The QDAS tools from Dataverse now have the ability to create 'redacted' versions of the codebook that remove such entries and to write the redacted version as a public auxiliary file. The intent is to allow people who do not have access to the full file to see only the less sensitive content.

This PR refactors the previewer code to allow this use case. It changes the overall logic of identifying which tools to display for a given datafile by making/caching the list of tools a given user is allowed to use, including tools that require a auxiliary file that exists and is public even if the user cannot access the file. (The original code made/cached a list of all the tools available for a given file and then checked the user's permission to access the file dynamically/multiple times.)

- Closes #

Possible design changes:

- This may be a feature that should be optional
- It may be useful to go further and specify which permissions a tool needs rather than assuming that requiring an auxiliary file means the tool can be used by anyone when the auxiliary file is public
- It may be that the rules for explore/config/query tools should be updated for similar use cases

**Additional documentation**:


cmbz:
2026-05-28: Added to _Ready for Review_ after discussing with @qqmyers (since the PR includes the `DataverseNO` label) - would be good to get this into 6.11

src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
```diff
@@ -5998,10 +5997,14 @@ public List<DatasetField> getDatasetSummaryFields() {
...
-        return previewTools.size() > 0;
-    }
+    public boolean isShowPreviewButton(FileMetadata fmd) {
+        DataFile dataFile = fmd.getDataFile();
+        List<ExternalTool> previewTools = getPreviewToolsForDataFile(dataFile, fileDownloadHelper.canDownloadFile(fmd));
+        if (previewTools.isEmpty()) {
+            return false;
+        }
+        return true;
+                }
```
stevenwinship: fix indentation

qqmyers: fixed

stevenwinship (approved):

src/main/java/edu/harvard/iq/dataverse/api/Files.java
```diff
@@ -22,6 +22,7 @@
 import io.gdcc.spi.export.ExportException;
```
pdurbin, line 22: This PR should have a release note, shouldn't it?

pdurbin:
I tested this PR and I'll merge it shortly.

It does what it says it does. It shows the "preview" eyeball button for restricted files that have public auxiliary (aux) files.

It reminded me that when we implemented metadata extraction from NetCDF and HDF5 files into aux files in NcML format, those aux files are public, even when the file is restricted. Or rather, I don't know if we explictly thought much about how we are automatically making public aux files from restricted files. I checked the PR and issue below and don't see any mention of "restricted":

- #9239
- #9153

If anyone reading this is concerned and wants this behavior to change, please open an issue! In 06352fb I at least added a link to the release note and to the guides about how public NcML aux files are created even for restricted NetCDF and HDF5 files. Again, this is existing behavior, going back to Dataverse 5.13 when the PR above was merged.

Going back to the main use case this current PR fixes, I'll share some "before" and "after" screenshots.

## Before: No preview of public aux files for restricted files (no eyeball)

before

## After: Public aux files are preview-able even for restricted files (eyeball is present)

after

Along the way I found what I'd consider to be a bug in JSF but since we're rewriting the frontend in React, for now I only mentioned it [in Slack](https://iqss.slack.com/archives/C04MJ76A34Z/p1788361008786619). Here's the bug: when you set isPublic=false when uploading an aux file, I don't think we should be offering it for download like this...

JSF bug

... expecially since when you click it (as a guest, for example), you get an error: `{"status":"ERROR","code":403,"message":"Not authorized to access this object via this API endpoint...`

Finally, going back to how we auto-generate NcML files for even restricted NetCDF and HDF5 files, here's how the preview looks:

restricted

PR #12200, opened 2026-03-09
Improve the Efficiency of the /api/index/perms API call

qqmyers:
**What this PR does / why we need it**: The (undocumented?) API call /api/index/perms iterates through all dvobjects in the database in one synchronous transaction, making it `unusable. This PR replaces that logic with an asynchronous iteration over the datasets and dataverses in the root dataverse and use of the index self and children logic used elsewhere. That should make it much faster/less memory intensive.

- Closes #

**Special notes for your reviewer**: FWIW - the old code calls findAll dvobjects which, despite the allExceptFiles variable name, includes files. The code then puts all the files in a map and, as far as I can see, never used them.

**Suggestions on how to test this**: Run the api call before and after, confirm speed/memory improvement, check the db to make sure all last per index times are updated after the new call.

**Additional documentation**:


rtreacy (commented):
via Claude --
  PR #12200: Improve the Efficiency of the /api/index/perms API call (by qqmyers)
  
  The PR replaces a synchronous, memory-heavy indexAllPermissions() with an @Asynchronous method that iterates dataverses individually, delegating to the existing
  indexPermissionsOnSelfAndChildren() logic.

  Key recommendations:

  1. Add @TransactionAttribute(TransactionAttributeType.NOT_SUPPORTED) — Without it, the async method holds a single transaction open for the entire run (potentially hours on large installations),
  risking timeouts.
  2. Guard against concurrent invocations — The fire-and-forget async nature means nothing prevents multiple overlapping runs. An AtomicBoolean or @Singleton/@Lock(WRITE) pattern would prevent
  this.
  3. No authentication on the endpoint (pre-existing) — Any anonymous user can trigger resource-intensive reindexing via a simple GET request.
  4. Confirm old method is fully deleted — The old indexAllPermissions() becomes dead code after this change.

  The approach itself is sound — reusing indexPermissionsOnSelfAndChildren with its batched sub-transactions and em.clear() every 10 dataverses is correct for memory management.

qqmyers:
Changes made per review, plus - changed to POST as best practice to avoid cross site issues, also updated call to permission reindex a single dataset to use POST/require superuser and added documentation (calls were previously undocumented) in release note, solr guide and change log.

rtreacy (commented):
Claude:
. Overall the design is solid: it replaces a single-transaction full-DB scan with an async, batched reindex reusing the existing indexPermissionsOnSelfAndChildren machinery, and adds proper superuser auth + a 409-conflict guard against concurrent runs. Good documentation additions too.

Two issues mentioned by Claude - 
possible race condition, which to me seems improbable
Dead code left over from the refactor 

I'm inclined to approve - will discuss at standup

rtreacy (approved):
Approved after discussion at standup of above issues

qqmyers:
@rtreacy - my changes were meant to address all the review issues. Are the issues raised by Claude:  

> possible race condition, which to me seems improbable, Dead code left over from the refactor

obsolete?

rtreacy:
@qqmyers Here are the specifics of the issues Claude raised. Do you think these need to be addressed before merge?

Two issues found:

1. Race condition in the concurrency guard (SolrIndexServiceBean.java:212) — indexingInProgress is only set to true inside the @Asynchronous method body, which runs on a separate thread pool after being dispatched. Two admin requests arriving close together can both pass the isIndexingPermissionsInProgress() check in Index.java before either async invocation actually sets the flag, letting two full reindex passes run concurrently — exactly what the 409 guard is meant to prevent.
2. Dead code left over from the refactor (SolrIndexServiceBean.java:143) — constructDatafileSolrDocsFromDataset has no remaining callers (its only caller, determineSolrDocsForFilesFromDataset, was deleted), and its indentation was also broken in the diff, deviating from the file's 4-space convention.

No test coverage exists for the old endpoint behavior, so nothing breaks there, but consider adding an IT test for the new POST/superuser/409 behavior.


I thought that two admin requests being that close together was an improbable occurrence, but I brought it up for discussion in standup before approving. If you think that's wrong, we should send it back to work on. That was the potentially most critical issue. Also wasn't sure if cleaning up dead code was a prerequisite for approval, although looking back that was the original recommendation of the previous review

qqmyers:
@rtreacy - here's my AI's response to yours. The deadcode was removed and the test for whether a run has already started has been slightly refactored to avoid the possibility of a quick repeat sneaking through (which I agree is probably very rare). There's a new test - it relies on the reindexing being slow enough that we can make a second API call before the reindexing completes. Hopefully that's always true but we should watch for the IT tests to complete successfully.

rtreacy (approved):
Claude
Latest change (commit 7f319fb3b4, "respond to review comments") fixes both issues from my earlier review:

1. Race condition — fixed. asyncIndexAllPermissions() now synchronously does the compareAndSet and returns a boolean; the actual work moved to a new @Asynchronous doAsyncIndexAllPermissions() invoked via self. only after the flag is successfully claimed. Index.java now checks the return value directly instead of a separate check-then-act. This closes the TOCTOU window I flagged.
2. Dead code — fixed. The unused constructDatafileSolrDocsFromDataset method was deleted entirely.

PR #12212, opened 2026-03-12
DANS/QDR ORE Bag PR with fix for collections

qqmyers:
**What this PR does / why we need it**: This PR includes a fix for the problems seen with collection creation/publication/deletion that appeared with #12167. It also has a minor merge fix for determining the hashtype to use when files are missing hash values in the database and removes an ignorehashes variable that was legacy from before Dataverse which doesn't ever set it to true.

- Closes #

**Special notes for your reviewer**: This branch has all the #12167 changes. It is therefore not up to date with develop which has reverted all of those. I'm not sure how that should be handled. 

FWIW: The fix for collections is basically to revert the update of permission time to run in the command transactions for collections while leaving datasets to work with it in a new transaction (~as appropriate for them since indexing is done asynchronously, in parallel with workflows, and adding old information from a shared context is a problem). Collection commands appear to rely on the call to update permission time to store required info for the collection (hard to track, but switching the permission time update to a separate transaction leads to errors in the log about not finding the alias of the collection). Changing that in the future seems like a useful thing to do.

**Suggestions on how to test this**: DANS has tested the minor fix for hash value type. To test the collections functionality, trying to create/publish/delete collections from the UI should work. 

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DvObjectServiceBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DvObjectServiceBean.java
@@ -223,6 +223,10 @@ public class DvObjectServiceBean implements java.io.Serializable {
      * updateContentIndexTime method.
      */
     @TransactionAttribute(REQUIRES_NEW)
+    public DvObject updatePermissionIndexTimeInNewTransaction(DvObject dvObject) {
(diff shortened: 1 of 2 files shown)


sekmiller:
Closed as included in 12213

PR #12220, opened 2026-03-16
fix jsf guestbook response cancel button

### LINKED ISSUE #12205: Guestbook Response JSF popup Cancel button not working
**What steps does it take to reproduce the issue?** Create a Dataverse/Dataset and upload files. Add a Guestbook with required fields to the Dataset. Publish

* When does this issue occur? Try to download a file and you will see the Guestbook Response dialog. Press cancel and nothing happens. 

* Which page(s) does it occurs on? Dataset page when trying to download an image

* What happens?

* To whom does it occur (all users, curators, superusers)? All

* What did you expect to happen? Dialog should close. The workaround is to hit the browser's back button.

**Which version of Dataverse are you using?** 6.9

**Screenshots:**

*

stevenwinship:
**What this PR does / why we need it**: Pressing 'Cancel' button on JSF Guestbook Response screen (preview, download, or request access) does not cancel the dialog

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12205

- Closes #12205

**Special notes for your reviewer**: lines prior to 291 are just reformatted because the spacing was off. No code changes were made. The new code comes after line 291

**Suggestions on how to test this**: Make a dataset require a guestbook response. Try to preview the file. Try to download the file. Also with dataverse.files.guestbook-at-request=true try to request access to a restricted file. Old version - cancel does nothing. This version - Cancel closes the tab/dialog and returns you to the files tab.

**Additional documentation**:


sekmiller (approved):

src/main/webapp/guestbook-terms-popup-fragment.xhtml
```diff
@@ -25,123 +25,123 @@
...
+                    <h:outputText id="fdDisclaimer" class="col-sm-6"
+                        value="#{MarkupChecker:sanitizeBasicHTML(workingVersion.termsOfUseAndAccess.disclaimer)}"
+                        escape="false" />
+                </div>
+            </ui:fragment>
             <div class="form-group"
-                jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.disclaimer}">
-                <label jsf:for="fdDisclaimer" class="col-sm-3 control-label">
-                    #{bundle['file.dataFilesTab.terms.list.termsOfUse.addInfo.disclaimer']}
+                jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.termsOfAccess and (hasRestrictedFile or !empty fileDownloadHelper.filesForRequestAccess)}">
```
qqmyers: These don't look related?

stevenwinship: removed

src/main/webapp/guestbook-terms-popup-fragment.xhtml
```diff
@@ -25,123 +25,123 @@
...
+                        escape="false" />
+                </div>
+            </ui:fragment>
             <div class="form-group"
-                jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.disclaimer}">
-                <label jsf:for="fdDisclaimer" class="col-sm-3 control-label">
-                    #{bundle['file.dataFilesTab.terms.list.termsOfUse.addInfo.disclaimer']}
+                 jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.disclaimer}">
+                <label jsf:for="fdTermsOfAccess" class="col-sm-3 control-label">
+                    #{bundle['file.dataFilesTab.terms.list.termsOfUse.addInfo.termsOfsAccess']}
```
youseihuayu-wonderful: These lines differ from `develop` in two ways that look unrelated to the Cancel button change. Was this intended?

**Condition** — `develop` renders on `termsOfAccess`, this branch renders on `disclaimer`:

```diff
- jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.termsOfAccess and (hasRestrictedFile or !empty fileDownloadHelper.filesForRequestAccess)}"
+ jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.disclaimer}"
```

A dataset with Terms of Access but no disclaimer no longer renders the section at all, so a user accepts the guestbook and downloads restricted files without seeing the access terms. A dataset with a disclaimer but no Terms of Access renders an empty labelled row. The `hasRestrictedFile` guard is dropped as well.

**Bundle key** — `file.dataFilesTab.terms.list.termsOfUse.addInfo.termsOfsAccess` is not in `Bundle.properties`; only `file.dataFilesTab.terms.list.termsOfAccess.termsOfsAccess` exists (line 2102). `bundle` is a JSF `<resource-bundle>`, so the label renders as `???file.dataFilesTab.terms.list.termsOfUse.addInfo.termsOfsAccess???`.

**Origin** — both arrived in 05bc2a1c (Jul 6), on the same line as @qqmyers' earlier "These don't look related?". The earlier commits match `develop` here. The approval predates 05bc2a1c by about ten weeks, and no CI job exercises this fragment's rendering.

The suggestion restores both values and keeps this branch's indentation:

```suggestion
                 jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.termsOfAccess and (hasRestrictedFile or !empty fileDownloadHelper.filesForRequestAccess)}">
                <label jsf:for="fdTermsOfAccess" class="col-sm-3 control-label">
                    #{bundle['file.dataFilesTab.terms.list.termsOfAccess.termsOfsAccess']}
```

The Cancel button behaviour is being verified separately.

youseihuayu-wonderful (commented):
Reviewed the diff against `develop`. One finding, left inline on the Terms of Access block — the `rendered` condition and the bundle key both changed in 05bc2a1c and look unrelated to the Cancel fix.

The Cancel button change itself reads correctly. `required` on the guestbook inputs is gated on the `DO_GB_VALIDATION_<context>` parameter, which the Cancel buttons deliberately do not pass, so switching to `p:commandButton` does not pull validation into the cancel path.

One minor observation, not blocking: `styleClass` moved from `btn btn-link` to `btn btn-default`, which differs from the other cancel buttons in the codebase and makes Cancel look the same as the Accept button beside it.

Runtime verification of the three cancel paths (preview tab, download dialog, request access) is still pending a preview image for this branch.

src/main/webapp/guestbook-terms-popup-fragment.xhtml
youseihuayu-wonderful: The undo in 257febcb restored the Terms of Access regression. Same two lines as before:

- `rendered` keys off `disclaimer` instead of `termsOfAccess` (and drops the restricted-file / request-access guard)
- bundle key `file.dataFilesTab.terms.list.termsOfUse.addInfo.termsOfsAccess` is not in `Bundle.properties`; only `file.dataFilesTab.terms.list.termsOfAccess.termsOfsAccess` exists

Please re-apply. This is independent of the JSF `13-dataset-actions` Select All failure — that spec does not touch this fragment, and it passed locally against this preview.

```suggestion
                 jsf:rendered="#{!empty workingVersion.termsOfUseAndAccess.termsOfAccess and (hasRestrictedFile or !empty fileDownloadHelper.filesForRequestAccess)}">
                <label jsf:for="fdTermsOfAccess" class="col-sm-3 control-label">
                    #{bundle['file.dataFilesTab.terms.list.termsOfAccess.termsOfsAccess']}
```

youseihuayu-wonderful (commented):
Re-posted the Terms of Access suggestion on current HEAD. 257febcb undid e2fe762e and put the regression back; the JSF Select All failure is a separate issue.

youseihuayu-wonderful:
Update: Steven applied the two-line restoration directly in `bf019c4` and then merged the latest `develop` into the branch (`ad315c1`).

The current official checks pass:

- JSF: **43/43 passed**, including `13-dataset-actions` on Chromium, Firefox, and WebKit: https://github.com/IQSS/dataverse/actions/runs/31705546363
- Integration tests: passed: https://github.com/IQSS/dataverse/actions/runs/31705546389
- CodeQL: passed

The earlier trace showed no `filesTable` `toggleSelect` request and no JSF/EL/bundle error. That fixture also does not render this Terms of Access block. With the restoration now present on #12220 and its current CI passing, #12608 is superseded.

pdurbin:
@stevenwinship can you please decide if you'd like to merge this PR by @youseihuayu-wonderful ...

- #12608

... into this one (#12220)?

src/main/webapp/guestbook-terms-popup-fragment.xhtml
stevenwinship: This suggestion failed to build

stevenwinship: Ok. I see it's changed

stevenwinship:
merged the latest suggested code

src/main/webapp/guestbook-terms-popup-fragment.xhtml
```diff
@@ -309,9 +309,17 @@
...
                 <f:param name="DO_GB_VALIDATION_#{popupContext}" value="true"/>
             </p:commandButton>
-            <button class="btn btn-link" onclick="PF('guestbookAndTermsPopup').hide();PF('blockDatasetForm').hide();" type="button">
-                #{bundle.cancel}
-            </button>
+            <!--PREVIEW TAB CANCEL BUTTON-->
+            <p:commandButton styleClass="btn btn-default" value="#{bundle.cancel}"
+                             rendered="#{popupContext == 'previewTab'}"
+                             onclick="history.back();"
+                             update="fileForm:tabView">
```
qqmyers, line 316: Given the onclick method, this doesn't appear to have any effect.

src/main/webapp/guestbook-terms-popup-fragment.xhtml
```diff
@@ -309,9 +309,17 @@
                              update="guestbookUIFragment">
                 <f:param name="DO_GB_VALIDATION_#{popupContext}" value="true"/>
             </p:commandButton>
-            <button class="btn btn-link" onclick="PF('guestbookAndTermsPopup').hide();PF('blockDatasetForm').hide();" type="button">
-                #{bundle.cancel}
-            </button>
+            <!--PREVIEW TAB CANCEL BUTTON-->
+            <p:commandButton styleClass="btn btn-default" value="#{bundle.cancel}"
+                             rendered="#{popupContext == 'previewTab'}"
+                             onclick="history.back();"
```
qqmyers, line 315: If I navigate to the datafile page, and then click cancel, I end up back on the dataset page, which seems counter-intuitive as I wouldn't expect to get pushed back to the dataset page (I have to accept or just ignore the dialog and switch to another pane to stay on the file page). 

Same concern if I start from a search page - clicking cancel pushes me back to the search result page.

Going back to the dataset page makes some sense if I just click the preview button, since I started on the dataset page in that case, but I think it might be better to just stay on the datafile page and switch to the metadata tab. This onclick would do that: `onclick="var metadataTab = $('a[href$=&quot;metadataMapTab&quot;]'); if (metadataTab.length) { metadataTab.click(); } else { history.back(); } "` I suspect the check that the tab exists and the option to still use history.back() if not isn't really needed (if you have a previewTab so this button is rendered, the metadataTab has to be there too.).

The one minor weirdness in doing this is that the large thumbnail on the metadata pane is also labelled "Preview" so it appears that when you to cancel while trying to see the preview, you see it anyway (you don't -  it's a 400px thumbnail for types that have one). I'd suggest just changing that label on the metadata tab to "Thumbnail" to address this. It's the `file.metadata.preview` entry in the `Bundle.properties`

src/main/webapp/guestbook-terms-popup-fragment.xhtml
```diff
@@ -309,9 +309,17 @@
                              update="guestbookUIFragment">
                 <f:param name="DO_GB_VALIDATION_#{popupContext}" value="true"/>
             </p:commandButton>
-            <button class="btn btn-link" onclick="PF('guestbookAndTermsPopup').hide();PF('blockDatasetForm').hide();" type="button">
-                #{bundle.cancel}
-            </button>
+            <!--PREVIEW TAB CANCEL BUTTON-->
+            <p:commandButton styleClass="btn btn-default" value="#{bundle.cancel}"
```
qqmyers, line 313: For all three p:CommandButtons added in this PR: they need a type="button" attribute. Without that, they make an extra ajax call to the server (as they act as a submit button by default). I confirmed this in QA (and also noted that one can achieve ~the same thing by having the `onclick` javascript have a `return false;` at the end, but using the type attribute is cleaner.)

doc/release-notes/12205-guestbook-response-jsf-popup-cancel-button.md
```diff
@@ -0,0 +1 @@
+This release fixes a bug which, in the JSF UI, the 'Cancel' button on the Guestbook Response dialog/preview tab does not close the dialog/tab. With this fix the dialog will close or, when in the preview tab, the tab will go back to the 'Files' tab.
```
qqmyers, line 1: This would change with the suggested updates - `... when on the file page, clicking cancel on the preview tab will shift the view to the metadata tab.`

src/main/webapp/guestbook-terms-popup-fragment.xhtml
stevenwinship: Without this code there is no cancel button. Only an "accept" button. To cancel you must click the browser's back button.

src/main/webapp/guestbook-terms-popup-fragment.xhtml
stevenwinship: The purpose of this cancel button was to send you back to the place you started (Dataset Page Files Tab). Since that's not what you want I will remove the button so the user can choose to go back with the back button or select the metadata tab.

stevenwinship: Since this PR is to fix the "cancel" buttons that don't work and this "cancel" button was never there it will be restored to the original functionality

src/main/webapp/guestbook-terms-popup-fragment.xhtml
stevenwinship: added type="button"

doc/release-notes/12205-guestbook-response-jsf-popup-cancel-button.md
stevenwinship: removed the button

src/main/webapp/guestbook-terms-popup-fragment.xhtml
stevenwinship: Removed the cancel button

qqmyers: FWIW: The comment was on line 316 where the code is trying to make a call to he server to update the tab view while your onclick was sending people to another page.

src/main/webapp/guestbook-terms-popup-fragment.xhtml
qqmyers: The point in my comment is that people are not always starting from the same place - could even be an external site. It's worse if I go to the file page manually and then click the cancel (history.back) button, I'm sent away from the page I navigated to, possibly back to some other site, rather than being able to see the rest of the data file info on the page. My suggestion was to redirect people to the metadata tab instead, which avoids leaving them on a non-working preview tab.

stevenwinship: Understood. My point is that clicking a button that does nothing is a bug. If the button never existed there is no need to fix it

Issue #12183, opened 2026-02-26
Wording of 404static.xhtml is not generic


JeyaSam:
Hi!
    Creating this as a follow-up from https://dataverse.zulipchat.com/#narrow/channel/378866-troubleshooting/topic/404static.2Exhtml.20file . 404static.xhtml file introduced in v6.0 is not generic and contains harvard specific link and text:

<img width="1215" height="459" alt="Image" src="https://github.com/user-attachments/assets/e7f5ef82-0863-480d-b2ac-19263dee09ea" />

So, we suggest following changes to 404static.xhtml to make it more generic:
1. Remove following code snippet as it is Harvard specific: 
                    <div class="col-sm-8 small">
                        <p>Copyright &#169; 2023, The President &#38; Fellows of Harvard College | <a href="http://best-practices.dataverse.org/harvard-policies/harvard-privacy-policy.html" target="_blank">Privacy Policy</a>
                        </p>
                    </div>

2. In the file, #{bundle['footer.dataverseProject']} is used for Header. Instead can use #{bundle['dataverse']}:
    **Reference to code:**         
             <div class="navbar-header">
               <h:outputFormat class="navbar-brand custom-logo" value="#{bundle['footer.dataverseProject']}"/>
             </div>
             
3. The logo image at the bottom could be specified as "./resources/images/dataverseproject_logo.png" instead of "https://cdn.rawgit.com/IQSS/dataverse/develop/src/main/webapp/resources/images/dataverseproject_logo.png" .			
   Some installations like us may face CSP restriction to have external link references as shown below:

<img width="1765" height="672" alt="Image" src="https://github.com/user-attachments/assets/149b42f6-6b6a-4763-904c-34ff70411665" />

The missing alert-danger spec for error message  is also due to this external reference URL of bootstrap.min.css and bootstrap-theme.min.css in 404static.xhtml file.

PR #12225, opened 2026-03-18
add a GitHub Workflow to add bugs to the project board

pdurbin:

We'd like all new issues that are labeled as bugs to be automatically added to project 34. They will show up (among other places) in the "Open Bugs" view at https://github.com/orgs/IQSS/projects/34/views/46

Here are the docs: https://github.com/actions/add-to-project

I added a secret called ADD_TO_PROJECT_PAT for @dataversebot to the "dataverse" repo. See "Creating a PAT and adding it to your repository" in the docs above. I made a classic token with "project" scope, as required. PAT stands for Personal Access Token ~~and it's currently tied to my account (@pdurbin)~~.

I also gave @dataversebot write access to project 34 here: https://github.com/orgs/IQSS/projects/34/settings/access

Merge this. Create a new issue with `label:"Type: Bug"`. Check if it was added to the project. Find an existing issue that doesn't have `label:"Type: Bug"`. Add that label. See if it was added to the project. Try creating a issue without that label. It shouldn't be added.

Also, you can test over at https://github.com/IQSS/dataverse-installations because I set it up there first. (I added the same ADD_TO_PROJECT_PAT there too.)

Here's how it looks:

Screenshot 2026-03-19 at 12 26 29 PM

### DIFF
--- a/.github/workflows/add_bugs_to_project.yml
+++ b/.github/workflows/add_bugs_to_project.yml
@@ -1,0 +1,18 @@
+name: Add bugs to project board
+
+on:
+  issues:
+    types:
+      - opened
+      - labeled
+
+jobs:
+  add-to-project:
+    name: Add bug to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v1.0.2
+        with:
+          project-url: https://github.com/orgs/IQSS/projects/34
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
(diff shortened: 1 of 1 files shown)


pdurbin:
I merged this PR and tested it on the following issue by adding the "Type: Bug" label:

- https://github.com/IQSS/dataverse/issues/12183

Here's a screenshot:

Screenshot 2026-03-24 at 4 04 11 PM

PR #12078, opened 2026-01-08
Add css to Mime type properties.

bjosel:
**What this PR does / why we need it**:
The PR avoids flooding the payara logs with [INFO] messages of the form: 

_css is a filename/extension Dataverse doesn't know about. Consider adding it to the MimeTypeDetectionByFileExtension.properties file.|#]_

**Which issue(s) this PR closes**:

- Closes #12074

**Special notes for your reviewer**:
Can't tell if the payara messages are due to specifics of our installation but they occur on all systems we deployed Dataverse on (development, staging, production).

**Suggestions on how to test this**:
Check payara logs for above message before and after applying the patch

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
No

**Is there a release notes update needed for this change?**:
No

**Additional documentation**:

PR #12231, opened 2026-03-19
Add missing mimetype mapping

janvanmansum:

Add css=text/css to `MimeTypeDetectionByFileExtension`. This prevents a lot of logging lines like this:
```
css is a filename/extension Dataverse doesn't know about. Consider adding it to the MimeTypeDetectionByFileExtension.properties file.|#]
```

N/a

N/a

Requesting any page with customization would trigger above logging.

No

No

**Additional documentation**:

N/a


qqmyers:
Duplicates #12078 which will be in 6.10

Issue #11556, opened 2025-06-06
Extend and document recommended database connection settings


poikilotherm:
**Overview of the Feature Request**

We have some (limited) docs about what kind of options you *can* set, but there is no docs about what would be a good idea for a production setup.

I suppose https://guides.dataverse.org/en/latest/installation/config.html#additional-recommendations would be a good place for such a thing. Or put it at https://guides.dataverse.org/en/latest/installation/config.html#database-persistence

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
My own troubles when I had to restart the Postgres DB. Talking with fellow ops.

**What existing behavior do you want changed?**
- Add hints about connection validation (probably recommend auto-commit as a first step).
- Add `connectTimeout=30&socketTimeout=60` via DATAVERSE_DB_PARAMETERS

**Any brand new behavior do you want to add to Dataverse?**
None.

**Any open or closed issues related to this feature request?**
#9504
#8959
#7980

**Are you thinking about creating a pull request for this feature?**  
Probably will do.

PR #12241, opened 2026-03-20
Improved db connection

qqmyers:
**What this PR does / why we need it**: At QDR, we've recently see Dataverse stop responding without any out of memory or high load issues. 

The one thing I've seen in the log that is potentially related are errors related to the postgres connection:
- Exception DTX5007:Exception StackTrace javax.transaction.xa.XAException: jakarta.resource.ResourceException: This Managed Connection is not valid as the physical connection is not usable
-  RAR5031:System Exception
- org.eclipse.persistence.exceptions.DatabaseException\nInternal Exception: org.postgre       sql.util.PSQLException: Connection has been closed automatically because a new connection was opened for the same PooledConnection or the PooledConnection has been closed.\nError Code: 0\nCal       l: SELECT ID, CONTENT, LANG, NAME FROM SETTING WHERE ((NAME = ?) AND (LANG = ?))\n\tbind => [2 parameters bound]\nQuery: ReadAllQuery(name=\"Setting.findByName\" referenceClass=Setting sql=\"       SELECT ID, CONTENT, LANG, NAME FROM SETTING WHERE ((NAME = ?) AND (LANG = ?))\")"

Showing AI how the DataSource was configured, it pointed to an issue where we are specifying a pooling connector while Payara is also setting up a pool, causing duplication and potential issues under high load.

This PR switches to the recommended "Simple" connector and adds a new "fish.payara.fail-all-connections=${MPCONFIG=dataverse.db.fail-all-connections:true}" option that seems useful for recovering more quickly when connections are going bad.

While the relevant Dataverse code hasn't changed lately, it's possible that the update to Payara 7/new Eclipselink has changed something that is making the problem more visible.

Even if this is not related to the problem we're seeing at QDR, it seems like a useful update.

- Closes #

AI also recommended several settings that we don't have by default, which ~match Payara's documentation on advanced db settings. I don't know if changing the defaults in our code, or adding these to the recommended settings when upgrading/the docker default config would be useful:
-Ddataverse.db.is-connection-validation-required=true
-Ddataverse.db.validate-atmost-once-period-in-seconds=30
-Ddataverse.db.connection-validation-method=custom-validation
-Ddataverse.db.validation-classname=org.glassfish.api.jdbc.validation.PostgresConnectionValidation
-Ddataverse.db.fail-all-connections=true
-Ddataverse.db.connection-leak-timeout-in-seconds=300

Nominally make sure Dataverse works after the change, probably perf test and assure performance doesn't drop and performance under load is as/more stable than before.

FWIW: All of these changes are on QDR test machines - we'll be monitoring performance and hopefully pushing them to production where we've seen outages.

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
@@ -46,6 +46,7 @@ import javax.sql.DataSource;
             "fish.payara.validation-table-name=${MPCONFIG=dataverse.db.validation-table-name:}",
             "fish.payara.validation-classname=${MPCONFIG=dataverse.db.validation-classname:}",
             "fish.payara.validate-atmost-once-period-in-seconds=${MPCONFIG=dataverse.db.validate-atmost-once-period-in-seconds:0}",
+            "fish.payara.fail-all-connections=${MPCONFIG=dataverse.db.fail-all-connections:true}",
             // LEAK DETECTION
             "fish.payara.connection-leak-timeout-in-seconds=${MPCONFIG=dataverse.db.connection-leak-timeout-in-seconds:0}",
             "fish.payara.connection-leak-reclaim=${MPCONFIG=dataverse.db.connection-leak-reclaim:false}",
@@ -54,9 +55,8 @@ import javax.sql.DataSource;
             // LOGGING, SLOWNESS, PERFORMANCE
             "fish.payara.statement-timeout-in-seconds=${MPCONFIG=dataverse.db.statement-timeout-in-seconds:-1}",
             "fish.payara.slow-query-threshold-in-seconds=${MPCONFIG=dataverse.db.slow-query-threshold-in-seconds:-1}",
-            "fish.payara.log-jdbc-calls=${MPCONFIG=dataverse.db.log-jdbc-calls:false}",
-            // OTHER OPTIONS
-            "fish.payara.fail-all-connections=${MPCONFIG=dataverse.db.fail-all-connections:true}"
+            "fish.payara.log-jdbc-calls=${MPCONFIG=dataverse.db.log-jdbc-calls:false}"
+
         })
 public class DataSourceProducer {
 
--- a/src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
+++ b/src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
@@ -12,10 +12,8 @@ import javax.sql.DataSource;
         name = "java:app/jdbc/dataverse",
         // The app server (Payara) deploys a managed pool for this data source for us.
         // We don't need to deal with this on our own.
-        //
-        // HINT: PGSimpleDataSource would work too, but as we use a connection pool, go with a javax.sql.ConnectionPoolDataSource
-        // HINT: PGXADataSource is unnecessary (no distributed transactions used) and breaks ingest.
-        className = "org.postgresql.ds.PGConnectionPoolDataSource",
+
+        className = "org.postgresql.ds.PGSimpleDataSource",
         
         // BEWARE: as this resource is created before defaults are read from META-INF/microprofile-config.properties,
         // defaults must be provided in this Payara-proprietary manner.
@@ -56,7 +54,9 @@ import javax.sql.DataSource;
             // LOGGING, SLOWNESS, PERFORMANCE
             "fish.payara.statement-timeout-in-seconds=${MPCONFIG=dataverse.db.statement-timeout-in-seconds:-1}",
             "fish.payara.slow-query-threshold-in-seconds=${MPCONFIG=dataverse.db.slow-query-threshold-in-seconds:-1}",
-            "fish.payara.log-jdbc-calls=${MPCONFIG=dataverse.db.log-jdbc-calls:false}"
+            "fish.payara.log-jdbc-calls=${MPCONFIG=dataverse.db.log-jdbc-calls:false}",
+            // OTHER OPTIONS
+            "fish.payara.fail-all-connections=${MPCONFIG=dataverse.db.fail-all-connections:true}"
         })
 public class DataSourceProducer {
 
--- a/doc/release-notes/12241-simple db connector.md
+++ b/doc/release-notes/12241-simple db connector.md
@@ -1,0 +1,1 @@
+This version of Dataverse changes the class used for connecting with Postgres. This should help avoid problems seen under high load and potentially improve performance. See #12241 for details.
--- a/doc/sphinx-guides/source/installation/config.rst
+++ b/doc/sphinx-guides/source/installation/config.rst
@@ -949,6 +949,9 @@ Connection Validation
    * - dataverse.db.validate-atmost-once-period-in-seconds
      - Specifies the time interval in seconds between successive requests to validate a connection at most once.
      - ``0`` (disabled)
+   * - dataverse.db.fail-all-connections
+     - When true, refreshes all connections when one fails, speeding recover vs. detecting and refreshing connections individually.
+     - ``true`` (enabled)
 
 Connection & Statement Leaks
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^


jp-tosca:
Moved after triage, @landreev has interest on reviewing this PR.

jp-tosca:
@poikilotherm Just to manage the board, are you reviewing this?

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
```diff
@@ -12,10 +12,8 @@
         name = "java:app/jdbc/dataverse",
         // The app server (Payara) deploys a managed pool for this data source for us.
         // We don't need to deal with this on our own.
-        //
-        // HINT: PGSimpleDataSource would work too, but as we use a connection pool, go with a javax.sql.ConnectionPoolDataSource
-        // HINT: PGXADataSource is unnecessary (no distributed transactions used) and breaks ingest.
-        className = "org.postgresql.ds.PGConnectionPoolDataSource",
+
+        className = "org.postgresql.ds.PGSimpleDataSource",
```
poikilotherm, line 16: Sorry Jim, but I have to push back hard on this one. This is gross AI slop.

I don't know which model you used and how you prompted, but obviously the result did not take any of https://jdbc.postgresql.org/documentation/datasource docs into account.

You're right in that Payara manages the connection pool for us, which is why we do not use any of the application managed connection classes (`PGSimpleDataSource` or `PGPoolingDataSource`). As documented by the PG JDBC developers, we are using the `PGConnectionPoolDataSource`.

It seems a *very* bad idea to go away from that. Please revert this. It would be good to leave an additional note and reference to the JDBC driver docs, so any AI model reading this in the future will be made aware and is less likely to halucinate.

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
```diff
@@ -48,6 +46,7 @@
             "fish.payara.validation-table-name=${MPCONFIG=dataverse.db.validation-table-name:}",
             "fish.payara.validation-classname=${MPCONFIG=dataverse.db.validation-classname:}",
             "fish.payara.validate-atmost-once-period-in-seconds=${MPCONFIG=dataverse.db.validate-atmost-once-period-in-seconds:0}",
+            "fish.payara.fail-all-connections=${MPCONFIG=dataverse.db.fail-all-connections:true}",
```
poikilotherm, line 49: I agree to adding the option for configuration.

However, please do not change the default from Payara's default `false` to an implicit Dataverse `true`. IMO unaligned options between Payara and Dataverse makes things more complex and brittle and may create more brainfog for admins trying to do the right thing. In addition, to anyone not aware of any of this it may have unintended side effects. Thanks!

poikilotherm (changes requested):
> AI also recommended several settings that we don't have by default, which ~match Payara's documentation on advanced db settings. I don't know if changing the defaults in our code, or adding these to the recommended settings when upgrading/the docker default config would be useful:
-Ddataverse.db.is-connection-validation-required=true
-Ddataverse.db.validate-atmost-once-period-in-seconds=30
-Ddataverse.db.connection-validation-method=custom-validation
-Ddataverse.db.validation-classname=org.glassfish.api.jdbc.validation.PostgresConnectionValidation
-Ddataverse.db.fail-all-connections=true
-Ddataverse.db.connection-leak-timeout-in-seconds=300

Which of those did you have enabled at the time the failures occurred?
At least the connection validation is an absolute must for any production instance and if you did not have these, this would be the first thing to address before any code changes. (I've left some remarks on the code changes, too.)
Any connection in the pool not being validated will be happily handed over to the application when requested even if the underlying TCP connection was killed, as it is not being renewed or removed from the pool.

There have been reports about the exact error message you posted in https://github.com/payara/Payara/issues/769. There even were reports where people said connection validation doesn't help. But all of this was a long time ago. Thus, it would be very interesting to know which validation options were applied at time of occurrence, as it may point to bugs hidden in Payara or pgJDBC.

In general I'd be more cautious about the suggestions AI makes here, as it already botched up the data source class. Let's go through them:

>-Ddataverse.db.is-connection-validation-required=true
-Ddataverse.db.validate-atmost-once-period-in-seconds=30
-Ddataverse.db.connection-validation-method=custom-validation
-Ddataverse.db.validation-classname=org.glassfish.api.jdbc.validation.PostgresConnectionValidation

I absolutely agree that any production instance should enable connection validation. It's why I created #11556.

The connection validation class is correct. In my issue I documented to recommend "auto-commit" mode for connection validation first ([implementation](https://github.com/payara/Payara/blob/ae955810e0a5955a8fe66f4e919a5377ff71a229/appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/ManagedConnectionFactoryImpl.java#L386)), as it is easier to setup with less options that need to be set (now edited). If y'all think using the [`SELECT 1` statement the custom class](https://github.com/payara/Payara/blob/ae955810e0a5955a8fe66f4e919a5377ff71a229/appserver/connectors/connectors-internal-api/src/main/java/org/glassfish/api/jdbc/validation/PostgresConnectionValidation.java#L61) does is the better option, that's fine by me, too.

Personally, I'd favor a much shorter once-per-period to detect failures earlier, e.g. in case of a database restart or failover. As the validation is only executed on *idle* connections, it should not have much of an impact on a busy installation. Monitoring is probably still a good idea.

> -Ddataverse.db.fail-all-connections=true
-Ddataverse.db.connection-leak-timeout-in-seconds=300

I do agree it is good advice to enable logging of connection leaks. For production systems, [Payara recommends *enforcing* the reclaim, too](https://docs.payara.fish/community/docs/Technical%20Documentation/Payara%20Server%20Documentation/General%20Administration/Administering%20Database%20Connectivity.html#statement-leak-detection-and-leaked-statement-reclamation). 

I'm not sure if we should use a five minute timeout here, though. Even though we have long running operations (like exports), these suffer from N+1 query expansion and not long running queries. Maybe we have a chance in a larger production system like QDR or Harvard to enable logging with 60s first and see if we get any hits?

At the same time, we should to enable and monitor statement leak detection, too (`dataverse.db.statement-leak-timeout-in-seconds`), e.g. using 30 secs at first. This is very much associated with `dataverse.db.statement-timeout-in-seconds`, as the [documentation clearly says one needs to make sure `statement timeout < statement leak timeout < connection leak timeout`](https://docs.payara.fish/community/docs/Technical%20Documentation/Payara%20Server%20Documentation/General%20Administration/Administering%20Database%20Connectivity.html#statement-leak-detection-and-leaked-statement-reclamation). The default statement timeout is "indefinetly" `-1`, which for a logging / monitoring situation is fine, but it may clash when switching to enforcing. We should probably extend our documentation on all of this or point upstream.

On a related note: just in case - please don't mixup the leak detection with the maxIdleTimeout=300s. The leak detection is about connections not returning to the pool, the idle timeout is about removing connections from the pool that are no longer active.

On another related note: do y'all think we should ship a production profile for MPCONFIG that enables certain proven good values? (`-Dmp.config.profile=prod` or [env var `MP_CONFIG_PROFILE`](https://github.com/IQSS/dataverse/blob/7904e996940de80ce46a645cbef5d97e8104c978/src/main/docker/Dockerfile#L30)) Unfortunately, we can't ship these via a profiled microprofile-config.properties, as that file will not be read early enough for the DataSourceDefinition options. (Also, we'd need an additional profile "ct-prod" as Payara only supports one profile value and we already use the "ct" one.)

src/main/java/edu/harvard/iq/dataverse/util/DataSourceProducer.java
qqmyers: Thanks. I agree. I think AI (and people) get confused between the `PGPoolingDataSource `(which, if I uderstand, tried to make it's own pool and is now deprecated) and the `PGConnectionPoolDataSource` which works with a container managed pool. It sounds like Payara also works with `PGSimpleDataSource` (hence no failures in my testing) but the ConnectionPool datasource is the right one. 

Without this, just adding the fail-all-connections probably is not a priority, so I'll close the PR.

pdurbin: > so I'll close the PR.

I moved it to "merged" so we can talk about it at the next standup. Afterwards I'll either move it to "done" or take it off the board.

PR #12256, opened 2026-03-25
MyData API: Return empty result set instead or error

### LINKED ISSUE #11447: MyData API Endpoint - don't return error for empty result set
Call the MyData endpoint for a user that has no associated data.  It should return successfully, with an empty list of items.
Instead, it returns an error response: 
`{"success":false,"error_message":"Sorry, no results were found."}`
This makes it hard to distinguish between zero results, which is not an error, and other unexpected errors.

Return a normal response, with an empty list of items.
For example, 
```
{
    "success": true,
    "data": {
        "pagination": {
            "numResults": 0,
           . . . etc
        },
        "items": [],
        "total_count": 0,
        . . . etc
}
```

latest develop branch

stevenwinship:
**What this PR does / why we need it**: SPA needs empty result set instead of error when no results are found

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11447

- Closes #11447

**Additional documentation**: changelog as this is a breaking change


sekmiller:
@stevenwinship I'm getting an error in the DataRetrieverApiIT. On line 103 it's finding 0 instead of 1. Also the branch is out of date with dev.

stevenwinship:
I'm not seeing the error. There is a 4 second sleep that might be too short?
I can't imagine that 4 seconds would be too short though

I updated the branch and will run it again

sekmiller (approved):
Thanks for the update

ChengShi-1:
I just tested it locally, when there is an empty result, it returns successfully. 
`{"success":true,"message":"Sorry, nothing was found for these roles: Admin, Contributor","data":{"pagination":{"isNecessary":false, ...`

When there is not empty result, no message is shown.
`{"success":true,"data":{"pagination":{"isNecessary":false,"numResult...`

PR #12262, opened 2026-03-26
update making release docs post 6.10.1

pdurbin:

- Switch to building the war file with GitHub Actions instead of Jenkins
- Build the artifacts earlier and upload them to the draft release
- Replace old 5.x examples with 6.x.

Preview at https://dataverse-guide--12262.org.readthedocs.build/en/12262/developers/making-releases.html


landreev (approved):
Perfect. 
Once my action allows to put something arbitrary in the version label we can decide on what it should be going forward.

Issue #10981, opened 2024-07-19
Publishing: Published datasets should contain files


sbarbosadataverse:
The existing publishing model in HDV allows the publishing of datasets without files. This lends to the curation team contacting depositors about uploading files via support and often to deleting "empty" datasets.

To decrease the support contact and dataset deletions, change the publishing model to require all deposits contain at least one file before publishing can happen.

## Related
- Pending

jggautier:
I've been working with a depositor who published datasets without files a while back and we've been working through some complications related to this. After seeing that @stevenwinship is working on this, this week I mentioned to the depositor that Harvard Dataverse are planning to require files before publishing datasets and I asked them for their thoughts on this. [They wrote](https://help.hmdc.harvard.edu/Ticket/Display.html?id=350022#txn-8503563) that they supported the idea 👍

DS-INRAE:
@stevenwinship will this be configurable ?
We allow depositors to deposit dataset without files if they have a link to the data provided as Alternative URL, so the ability to deposit datasets without file is essential to our installation.

DS-INRAE:
> @stevenwinship will this be configurable ? We allow depositors to deposit dataset without files if they have a link to the data provided as Alternative URL, so the ability to deposit datasets without file is essential to our installation.

Should have read the PR first (but it would be better if it was also written in the issue anyway) it is said clearly that it is configured at the collection level

cmbz:
2025-04-28
- Feature enhancement: A message to user on failed publication should indicate that datasets require data files to be published

pdurbin:
@cmbz I thought this was implemented already. Please see the screenshot at https://github.com/IQSS/dataverse/pull/10994#discussion_r1852314343

pdurbin:
Great. New issues are best for feature enhancements, of course. 😄

PR #12266, opened 2026-03-30
Fix Publish and Submit for Review when no files are present in Dataset

### LINKED ISSUE #12258: Publishing: Published dataset version should contain files
**What steps does it take to reproduce the issue?** Per https://github.com/IQSS/dataverse/issues/10981 a dataset should not be publishable if it contains no files. This issue should check the number of files in the Dataset version and not the total number of files in the dataset. 
This is only an issue when the attribute 'requireFilesToPublishDataset' is set

* When does this issue occur? When a dataset has a file but the new version has removed the file (leaving 0 files)

Definition of Done:
*  1. The dataset version should not be publishable without at least 1 file.
*  2. The user should not be able to submit the dataset for review without files.

**Any related open or closed issues to this bug report?**https://github.com/IQSS/dataverse/issues/10981

stevenwinship:
Per https://github.com/IQSS/dataverse/issues/10981 a dataset should not be publishable if it contains no files. This issue should check the number of files in the Dataset version and not the total number of files in the dataset.
This is only an issue when the attribute 'requireFilesToPublishDataset' is set

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12258

- Closes #12258

Submit for review button on JSF UI Dataset Page will be hidden if no files are contained in the Dataset Version

requireFilesToPublishDataset set to true
image

requireFilesToPublishDataset set to false
image

**Additional documentation**: changelog

### DIFF
--- a/doc/release-notes/12258-publish-submit-contains-files.md
+++ b/doc/release-notes/12258-publish-submit-contains-files.md
@@ -1,0 +1,2 @@
+## Bug: Publish and Submit for review must contain files
+When `requireFilesToPublishDataset` is set on a Dataverse a Dataset must contain files to be published or submitted for review. This fix make sure the dataset version contains files, and not just the dataset. It also adds this check to the `Submit for Review` functionality.
--- a/doc/sphinx-guides/source/api/changelog.rst
+++ b/doc/sphinx-guides/source/api/changelog.rst
@@ -7,6 +7,13 @@ This API changelog is experimental and we would love feedback on its usefulness.
     :local:
     :depth: 1
 
+v6.11
+-----
+
+- The following API will now return ``403`` if the ``requireFilesToPublishDataset`` flag is set and the dataset version contains 0 files.
+
+  - **/api/datasets/{Id}/submitForReview**
+
 v6.10
 -----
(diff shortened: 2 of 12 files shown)


landreev (approved):
Haven't tested, but the code looks good.

sekmiller:
Getting a weird result here. If I'm logged in as a superuser and the dataset has no files, but is required to at the collection level there's a banner for "You must submit for Review", but there's no submit button (and there shouldn't be a button because the super user should be able to publish?). Nor is there a Publish button.

This also happens to non-superusers with publish dataset privileges. If the collection requires files and they are not present the "Submit for Review" banner is displayed - but there's no Submit button.

sekmiller:
Maybe out of scope, but can we add a banner to the dataset page when files are required but not present to explain why the publish or submit button are not available?

sekmiller:
If you are a Contributor without the right to publish datasets then if files are required and they are not present then the display shows a "Submit for Review" banner, but there is no Submit button. Once you add a file then you get both the banner and the button - again I think it would be appropriate to have a new banner stating that files are required before a dataset may be submitted for review.

PR #12268, opened 2026-03-31
Added Dataset.getGlobalIdForFileStorage for direct upload URLs that support legacy lowercase DOIs

janvanmansum:
Adds `Dataset.getGlobalIdForFileStorage()` and uses it in the calls to `s3io.generateTemporaryS3UploadUrls()` instead of `Dataset.getGlobalId().toString()`. When Dataverse contains datasets with DOIs which were added before Dataverse started enforcing uppercase DOIs, the upload URLs didn't match the actually keys in the object store for those datasets, causing direct upload to fail at the "completion" step (with multi-part uploads, not sure if single part uploads failed). This changed fixes the problem.

N/a

I have tested this with new datasets (all uppercase DOIs) and with legacy DOIs from our system, which contain lowercase letters. Both work. I have only been able to test through the API, although the UI-code has also been updated.

(Jim:) Minimally, one could edit the db to set a dataset identifier to lower/mixed case. (Not sure if using a lower case shoulder still results in a lowercase entry in the db.) One could also find/configure a dataset with an alternate identifier and configure it for use in file storage. In either case, direct uploads via the UI should work. Same for the API but in that case one could also inspect the results of the /api/datasets/{id}/uploadurls call and verify that the case used in the abort and complete URLs match the URLs for parts w.r.t. case.

No

See: https://github.com/IQSS/dataverse/pull/12268/changes#diff-3187604c28e828f652f43814550e9271409b81a62db28527c71fdd3c08cf9344

**Additional documentation**:
N/a

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/Dataset.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataset.java
@@ -662,7 +662,7 @@ public class Dataset extends DvObjectContainer {
         return retVal;         
     }
 
-    public String getGlobalIdForFileStorage() {
+    public String getGlobalIdForFileStorageAsString() {
         return getProtocolForFileStorage() + ":" + getAuthorityForFileStorage() + "/" + getIdentifierForFileStorage();
     }
 
--- a/src/main/java/edu/harvard/iq/dataverse/Dataset.java
+++ b/src/main/java/edu/harvard/iq/dataverse/Dataset.java
@@ -662,6 +662,10 @@ public class Dataset extends DvObjectContainer {
         return retVal;         
     }
 
+    public String getGlobalIdForFileStorage() {
+        return getProtocolForFileStorage() + ":" + getAuthorityForFileStorage() + "/" + getIdentifierForFileStorage();
+    }
+
     public String getNextMajorVersionString() {
         // Never need to get the next major version for harvested studies.
         if (isHarvested()) {
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -2798,7 +2798,7 @@ public class Datasets extends AbstractApiBean {
             String storageIdentifier = null;
             try {
                 storageIdentifier = FileUtil.getStorageIdentifierFromLocation(s3io.getStorageLocation());
-                response = s3io.generateTemporaryS3UploadUrls(dataset.getGlobalIdForFileStorage(), storageIdentifier, fileSize);
+                response = s3io.generateTemporaryS3UploadUrls(dataset.getGlobalIdForFileStorageAsString(), storageIdentifier, fileSize);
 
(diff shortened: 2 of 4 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -2798,7 +2798,7 @@ public Response getMPUploadUrls(@Context ContainerRequestContext crc, @PathParam
             String storageIdentifier = null;
             try {
                 storageIdentifier = FileUtil.getStorageIdentifierFromLocation(s3io.getStorageLocation());
-                response = s3io.generateTemporaryS3UploadUrls(dataset.getGlobalId().asString(), storageIdentifier, fileSize);
+                response = s3io.generateTemporaryS3UploadUrls(dataset.getGlobalIdForFileStorage(), storageIdentifier, fileSize);
```
qqmyers: FWIW: I think this would fix an issue for datasets using an alternative id for file storage as well.

src/main/java/edu/harvard/iq/dataverse/Dataset.java
```diff
@@ -662,6 +662,10 @@ public String getIdentifierForFileStorage(){
         return retVal;         
     }
 
+    public String getGlobalIdForFileStorage() {
```
qqmyers: Minor - should this be getGlobalId _String_ ForFileStorage ?

qqmyers (approved):
Looks good. Note that DANS tested this on DOIs using lowercase from their early migrations.

src/main/java/edu/harvard/iq/dataverse/Dataset.java
janvanmansum: Yes, that would be more consistent, or it could return a GlobalId-object (unless that automatically uppercases the identifier).

qqmyers: Good point - I don't see anything in GlobalId that changes case, so I think if you initialize it from the *ForFileStorage entries it should be fine.

janvanmansum: OK, I'll make the change and check the upper/lower-casing in the debugger while testing.

janvanmansum: @qqmyers : I have opted for renaming the method, after all. I __could__ create a `GlobalId` object, but I would either to have set `urlPrefix` and `providerName` to `null` or extend `PidUtils` with a `parseAsGlobalID` that leaves casing unchanged. The first option would problably result in an exception if someone down the road decided to call `asURL()` on it. The second option would be harder to test. I say we keep the change as minimal as possible.

qqmyers (approved):
reapproving after change

stevenwinship:
tested develop and this branch. get upload urls and"abort" upload. Was able to see the errors in develop branch and verify that this fix worked. I wasn't able to test "complete"

PR #12269, opened 2026-03-31
Add usage and response counts to get guestbooks list

### LINKED ISSUE #12260: API: Enhance GET Guestbooks with usage count and response count fields
Currently, the GET Guestbook API returns metadata about guestbooks (e.g., name, required fields, custom questions), but it does not include usage and responses, needed for Manage Guestbooks table
Image

Get a list of Guestbooks linked to a Dataverse Collection: GET `/api/guestbooks/{dataverseIdentifier}/list` also include usage count and response count fields

https://github.com/IQSS/dataverse/pull/12110

stevenwinship:
**What this PR does / why we need it**:Currently, the GET Guestbook API returns metadata about guestbooks (e.g., name, required fields, custom questions), but it does not include usage and responses, needed for Manage Guestbooks table

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12260

- Closes #12260

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Guestbooks.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Guestbooks.java
@@ -60,7 +60,7 @@ public class Guestbooks extends AbstractApiBean {
     @GET
     @AuthRequired
     @Path("{identifier}/list")
-    public Response getGuestbooks(@Context ContainerRequestContext crc, @PathParam("identifier") String identifier, @QueryParam("ignoreStats") boolean ignoreStats) {
+    public Response getGuestbooks(@Context ContainerRequestContext crc, @PathParam("identifier") String identifier, @QueryParam("includeStats") boolean includeStats) {
         return response( req -> {
             Dataverse dataverse = findDataverseOrDie(identifier);
             final Long dataverseId = dataverse.getId();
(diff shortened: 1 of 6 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Guestbooks.java
```diff
@@ -57,16 +60,22 @@ public Response getGuestbook(@Context ContainerRequestContext crc, @PathParam("i
...
             Dataverse dataverse = findDataverseOrDie(identifier);
+            final Long dataverseId = dataverse.getId();
             if (!permissionSvc.request(req).on(dataverse).has(Permission.EditDataverse)) {
                 return error(Response.Status.FORBIDDEN, "Not authorized");
             }
             List<Guestbook> guestbooks = guestbookService.findGuestbooksForGivenDataverse(dataverse);
             JsonArrayBuilder guestbookArray = Json.createArrayBuilder();
             JsonPrinter jsonPrinter = new JsonPrinter();
             for (Guestbook gb : guestbooks) {
+                // default is to include the stats. Ignore the stats for a faster reply if they are not needed
```
scolapasta: Is there a reason that the default is to include the stats? generally, faster defaults are "better" (for some definition of that word) and then you add the parameter if you need the slower.

If we do change the parameter should naturally be "includeStats". (I would even think we could consider the name change if we don't, partially in case we someday decide to change the default, the name works better for both cases, imo)

scolapasta (changes requested):
Left one comment about the default for discussion. 

Also branch is out of date.

src/main/java/edu/harvard/iq/dataverse/api/Guestbooks.java
stevenwinship: Since JSF includes the stats by default I wanted the API to match. Most of the time the stats are desired. I just added a way to exclude them if someone didn't want them or just wanted a faster listing

stevenwinship: I could just leave it out and always include the stats so we don't break backward compatibility. SPA is already using this API and expects the stats

stevenwinship: Changed the parameters to "includeStats" defaulted to false

scolapasta (approved):

ChengShi-1 (approved):
looks good with api tests

PR #12320, opened 2026-04-10
 'API inconsistency for release time between JsonParser and JsonPrinter

### LINKED ISSUE #11594: API inconsistency for release time between JsonParser/JsonPrinter

1. Download JSON of dataset / dataset version.
2. Try to create a new dataset from that.
3. Watch it fail because
    - [the printer has "releaseTime" as attribute for dataset versions with a timestamp,](https://github.com/IQSS/dataverse/blob/976d10cf4e8ed4a9f82bd723050dfb63af975807/src/main/java/edu/harvard/iq/dataverse/util/json/JsonPrinter.java#L490)
    - [but the parser requires a "releaseDate" with a date.](https://github.com/IQSS/dataverse/blob/976d10cf4e8ed4a9f82bd723050dfb63af975807/src/main/java/edu/harvard/iq/dataverse/util/json/JsonParser.java#L444)

It was especially obvious during testing. Used a JSON from an instance to create a unit test and had to debug why it wasn't accepting the data.

API

**What happens?**
See above.

Everybody when trying to create/import a dataset with an existing PID.

That I can import the data I dumped somewhere else.

1. Not sure how many people may have used "releaseDate" in the parser in the past, but as the issue has been around forever, maybe next to no one. So maybe it's fine to rename the parameter and just add a note to the API docs about the change.
2. Also, up until now the parser accepted only a *YYYY-mm-dd* formatted date, not a timestamp (`java.util.Date`). I suppose we need a backward compatible fix? Or not?

I didn't see the fields being used in pyDataverse or easyDataverse.

`develop` (but the discrepancy has been around since 2014 according to the Git log)

I didn't see any.

Side note: #6810 - such discrepancies happen when doing these things manually and with no schemas around.

Yes.

stevenwinship:
**What this PR does / why we need it**: The output of JsonPrinter includes "releaseTime" where the input to JsonParser only accepted "releaseDate"

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11594

- Closes #11594

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11594-release-time-inconsistency-api.md
+++ b/doc/release-notes/11594-release-time-inconsistency-api.md
@@ -1,2 +1,2 @@
 ## Bug
-API inconsistency for release time between JsonParser/JsonPrinter has been addressed. For backward compatibility new "releaseTime" and original "releaseDate" will be allowed in the Json giving preferance to the new "releaseTime" field. Also, "releaseTime" field will populate the db with date/time. The original "releaseDate" field will continue to truncated to the date only.
+API inconsistency for release time between JsonParser/JsonPrinter has been addressed. For backward compatibility new "releaseTime" and original "releaseDate" will be allowed in the Json giving preference to the new "releaseTime" field. Also, "releaseTime" field will populate the db with date/time. The original "releaseDate" field will continue to be truncated to the date only.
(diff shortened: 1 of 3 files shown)


rtreacy (approved):
Claude
Overall this is a small, well-targeted, backward-compatible fix with adequate test coverage.
Two minor findings (neither blocking):
1. The new test shadows the class field sut (the actual JsonParser under test) with an unrelated local String sut = "foobar", and redundantly re-instantiates a parser/mocks instead of reusing the existing sut field from @BeforeEach — just a readability/duplication nit.
2. The release note has two typos ("preferance" → "preference", "continue to truncated" → "continue to be truncated").

PR #12331, opened 2026-04-15
External Controlled Vocabularies: Support vocab specific retrieval filtering

qqmyers:
**What this PR does / why we need it**: To support use of multiple services on one field, i.e. to allow selecting a Person (ORCID) or organization (ROR), it is important for Dataverse to be able to retrieve and cache information about either PID type (i.e. so that names, etc. can be used in exports). To do that, the External Vocabulary mechanism (specifically the CVocConf schema) is being extended to support including retrieval-filtering and other elements per vocabulary (versus one for the whole field). That is in https://github.com/gdcc/dataverse-external-vocab-support/pull/58 . This PR makes the associated changes to Dataverse's code to use those per-vocab entries.

- Closes #

**Suggestions on how to test this**: Using the person-or-org script in the external vocab repo is probably the easiest way to see it in action.

**Additional documentation**:


cmbz:
2026-05-28: Added to _Ready for Review_ after discussing with @qqmyers (since the PR includes the `DataverseNO` label) - would be good to get this into 6.11

stevenwinship (approved):

qqmyers:
JSF fail appears to be https://github.com/gdcc/dataverse-jsf-tests/issues/1

PR #12332, opened 2026-04-15
Inconsistent behaviour of dataset delete API

### LINKED ISSUE #11253: Inconsistent behaviour of dataset delete API
While testing the dataset delete API, I just noticed the following behavior, which was unexpected to me:

When logged in as a superuser, and then calling the [delete API](https://guides.dataverse.org/en/latest/api/native-api.html#delete-unpublished-dataset) on a dataset with exactly one published version and no draft currently existing, the request seems to be automatically "upgraded" to a destroy action. The request succeeds and the dataset is deleted.

I expected the request to fail, since the delete API is not meant for deleting published datasets (at least according to the docs where the API is titled ["Delete Unpublished Dataset"](https://guides.dataverse.org/en/latest/api/native-api.html#delete-unpublished-dataset) and the destroy API exists, which is in contrast meant for deleting published datasets as a superuser).

What makes the behavior even more confusing to me, as soon as more than one published version exists, the delete API always fails even when I'm a superuser (message: `This is a published dataset with multiple versions. This API can only delete the latest version if it is a DRAFT`).

When not logged in as a superuser, everything works as I expected (delete API succeeds for drafts but fails for published versions).

Here's the related code: https://github.com/IQSS/dataverse/blob/f4380eafe647cc43b2f507f74123c7d9112c32c8/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L278-L287

And my question: Is this intended behavior? If yes, I think the special case (superuser + exactly one published version allows delete API to succeed) should be documented in the API docs.

stevenwinship:
**What this PR does / why we need it**: API call of delete Dataset does not match what has been documented. Superuser must use /destroy to delete a released Dataset. Delete Dataset API should only work for "Draft" datasets.

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/11253

- Closes #11253

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/PermissionsWrapper.java
```diff
@@ -249,7 +249,7 @@ public boolean canIssueUpdateDatasetCommand(DvObject dvo){
 
     // DELETE DATASET
     public boolean canIssueDeleteDatasetCommand(DvObject dvo){
```
sekmiller, line 257: minor, but do we want to say canIssueDeleteDatasetVersionCommand since that's what you're returning

sekmiller (dismissed):
Minor change suggestion - change the name of the permissions wrapper method to reflect the command tested. (it's only called in a couple of places)

src/main/java/edu/harvard/iq/dataverse/PermissionsWrapper.java
stevenwinship: I was just matching the other methods. I.e.
// UPDATE DATASET
    public boolean canIssueUpdateDatasetCommand(DvObject dvo){
        return canIssueCommand(dvo, UpdateDatasetVersionCommand.class);
    }

stevenwinship: If I change one then I would need to change them all

sekmiller: Ok. withdrawn.

sekmiller (approved):

qqmyers (changes requested):
From QA: This appears to now work as discussed in the issue, which I think continues to be inconsistent with the documentation. When published version(s) and a draft exist, the [Delete Unpublished Dataset API](https://guides.dataverse.org/en/latest/api/native-api.html#delete-unpublished-dataset) just deletes the draft and doesn't fail because the dataset has been published.

I'd suggest one of two changes:

1. make this only work if there is only a draft, or
2. adjust the docs to make it "Delete Unpublished Dataset Changes" and explain it removes a draft version (and metadata changes and deletes any files added in the draft) which results in deletion of the dataset if there are no published versions.

If it always deletes a draft (if one exists) - i.e. option 2, it's more of a duplicate of the DELETE /api/datasets/versions/{versionId} call (which, despite the signature, only works on drafts), so limiting this call to truly deleting unpublished datasets only might be more useful, but it's a bigger backward incompatible change. Option 2 also results in redundant code if there's no clean-up.

I'm not sure what the process should be to make a choice - perhaps just standup discussion?

Also - the PR doesn't update the change log to note the backward incompatibility (for superusers only at present).

Otherwise - the api does remove the draft, delete draft only files and not files in published versions, does redirect superusers to the destroy api as intended.

stevenwinship:
@qqmyers @pdurbin I added to the native api and changelog docs. Please take a look

pdurbin:
We discussed this yesterday and I suggested to @qqmyers that he could make a fresh pull request show how he thinks the APIs should change. He said he would but no rush! As discussed, this is not a must have.

qqmyers:
#12639 extends/replaces this PR.

pdurbin:
Closing in favor of this one:

- #12639

PR #12341, opened 2026-04-20
Fix timing issue when posting guestbook response and requesting the d…

### LINKED ISSUE #12340: BUG: Timing constrain issue between posting guestbook response and downloading file
Bad check for Guestbook Response post time and signed url call to download.
Only giving a few seconds instead of a minute to start the download

stevenwinship:
**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12340

- Closes #12340

**Suggestions on how to test this**: post guestbook response. loop calling api with signed url to make sure you can call download within the minute

[edit:] The old/incorrect behavior, currently in the develop branch: 
0-10 sec.: the signed url is working as it should;
11-60 sec.: error message "you may not download this file without the required Guestbook response ..." (??)
60+sec: signed url expires. 
[/edit - L.A.]

**Additional documentation**:

### DIFF
--- a/doc/release-notes/12340-timing-issue-guestbook-response.md
+++ b/doc/release-notes/12340-timing-issue-guestbook-response.md
@@ -1,0 +1,2 @@
+## BUG 
+Timing issue fixed where user only had a few seconds instead of a minute to call the File download API after POSTing the guestbook response
--- a/src/main/java/edu/harvard/iq/dataverse/api/Access.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Access.java
@@ -131,6 +131,7 @@ public class Access extends AbstractApiBean {
     DataverseFeaturedItemServiceBean dataverseFeaturedItemServiceBean;
     
     private static final String DEFAULT_BUNDLE_NAME = "dataverse_files.zip";
+    private static final int GUESTBOOK_RESPONSE_SIGNEDURL_TIMEOUT_MINUTES = 1;
     //@EJB
     
     // TODO: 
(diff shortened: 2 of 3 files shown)


landreev (commented):
Makes sense overall. Should we make it a jvm setting, and configurable, in seconds? - Although I suspect that 1 min. will just work for everybody.

landreev:
> Makes sense overall. Should we make it a jvm setting, and configurable, in seconds? - Although I suspect that 1 min. will just work for everybody.

Your call really. Happy to approve as soon as I hear back either way.

landreev (approved):
... or I can just approve. If/when someone has a practical need to reduce, or increase this lifespan, we'll deal with it.

landreev:
The Jenkins failure is an aws fluke - ?

landreev:
> The Jenkins failure is an aws fluke - ?

Not an "aws fluke", but 
```
TASK [dataverse : download custom maven version] *******************************
fatal: [localhost]: FAILED! => {"changed": false, "dest": "/tmp/apache-maven-3.9.14-bin.tar.gz", "elapsed": 0, "msg": "Request failed", "response": "HTTP Error 404: Not Found", "status_code": 404, "url": "http://mirrors.ibiblio.org/apache/maven/maven-3/3.9.14/binaries/apache-maven-3.9.14-bin.tar.gz"}
```
still, for reasons unrelated to/before it gets to the tests.

landreev:
OK, Jenkins tests passed. (thanks @donsizemore!)

PR #12345, opened 2026-04-21
Required controlledVocabulary metadata marked as valid while empty

### LINKED ISSUE #11900: Required controlledVocabulary metadata marked as valid while empty

stevenwinship:
**What this PR does / why we need it**: Previous PR caused post merge test failures in Jenkins. This PR includes the code from the previous PR plus the fixes needed to pass the tests

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11900

- Closes #11900

**Special notes for your reviewer**: Original PR https://github.com/IQSS/dataverse/pull/11950

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValidator.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldValidator.java
@@ -59,10 +59,26 @@ public class DatasetFieldValidator implements ConstraintValidator<ValidateDatase
         }
 
         // if value is not primitive or not empty
-        if (!dsfType.isPrimitive() || !StringUtils.isBlank(value.getValue())) {
+        // For controlled vocabulary fields, check that actual CV values are selected,
+        // not just that datasetFieldValues contains something (which might be an invalid N/A placeholder)
+        // See https://github.com/IQSS/dataverse/issues/11900
+        if (!dsfType.isPrimitive()) {
             return true;
         }
-       
+
+        if (dsfType.isControlledVocabulary()) {
+            // For CV fields, check if there are actual controlled vocabulary values selected
+            if (value.getControlledVocabularyValues() != null && !value.getControlledVocabularyValues().isEmpty()) {
+                return true;
+            }
+            // If no CV values, fall through to required field check below
+        } else {
+            // For non-CV primitive fields, check if value is not blank
+            if (!StringUtils.isBlank(value.getValue())) {
+                return true;
+            }
+        }
+
         if (value.isRequired()) { 
             String errorMessage = null;
             DatasetFieldCompoundValue parent = value.getParentDatasetFieldCompoundValue();
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java
@@ -111,7 +111,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.primefaces.event.FileUploadEvent;
 import org.primefaces.model.file.UploadedFile;
 
-import jakarta.validation.ConstraintViolation;
 import java.util.Arrays;
 import java.util.HashSet;
 import jakarta.faces.model.SelectItem;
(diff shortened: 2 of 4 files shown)


cmbz:
2026-04-22: @ErykKul we're returning this PR back to you for further fixes.

ErykKul:
Root cause: allowHarvestingMissingCVV was gating the sanitizer, but it should only affect the parser.

The parser uses the flag to decide whether to throw on an unmappable controlled vocabulary string. The sanitizer handles a different case: required controlled vocabulary fields that come up empty after initDatasetFields() because the source format (e.g. oai_dc) does not carry them. Those fields must always be filled with a proper ControlledVocabularyValue wrapping "N/A", not a plain "N/A" string, otherwise the second validation pass still fails.

The fix removes the gate and routes by field type instead. Controlled vocabulary fields get a ControlledVocabularyValue("N/A"), other fields get setSingleValue("N/A"). With that, allowHarvestingMissingCVV=false correctly imports 7 of 8 records: the parser rejects the 1 record with a bad controlled vocabulary string, and the sanitizer rescues the 7 records with empty controlled vocabulary fields.

rtreacy (approved):
Looks good generally. Passes tests. On coding style, is it really necessary/desirable to use import edu.harvard.iq.dataverse.*;  ?

stevenwinship:
QAing this pr I found the following issue:

Setup: JVM setting -Ddataverse.api.allow-incomplete-metadata=true (same result if not set)
Run HarvestingClientsIT

Create a new dataset in UI and observe the "Subject" drop down. All the 'N/A' values that were assigned in the harvesting (ImportServiceBean sanitize) were also added to the drop down list.

image

jp-tosca:
Let us know when you can look at this @ErykKul :) thank you!

ErykKul:
The previous fix created new `ControlledVocabularyValue(null, "N/A", f.getDatasetFieldType())`, a transient JPA entity pointing to the actual field type (e.g. Subject). Because `DatasetField.controlledVocabularyValues` has `@ManyToMany(cascade = {CascadeType.MERGE})`, merging the `DatasetField` on save cascaded to this new entity and inserted it as a permanent row in controlledvocabularyvalue with the Subject's datasetfieldtype_id. After running HarvestingClientsIT, every CV field that was sanitized with N/A added a new "N/A" entry to its dropdown. Running the test repeatedly multiplied the rows.

The fix uses `datasetfieldService.findNAControlledVocabularyValue()` instead, which returns the existing managed singleton (the same entity SWORD uses) that has `datasetFieldType = null`. Merging a managed entity only updates the join table — no new row is ever inserted into controlledvocabularyvalue, and the Subject dropdown stays clean. The `if (naValue != null)` guard handles the rare case where the singleton was never loaded via `loadNAControlledVocabularyValue`; in that case the field is left empty and the second-pass validation will throw `ImportException`, which is correct. If the N/A placeholder doesn't exist, the record cannot be sanitized and should be skipped.

PR #12348, opened 2026-04-21
Add dataset type ID to JSON output in JsonPrinter

jp-tosca:

The Id on the type is required for mapping on JS Dataverse


landreev (approved):
looks good

PR #12350, opened 2026-04-21
Allow non UTC lastUpdateTimestamp

### LINKED ISSUE #12349: Fix update time stamp check to not assume UTC
**What steps does it take to reproduce the issue?** Run on OS that uses local time zone instead of UTC
(running on MAC, not in docker, can be used)

* When does this issue occur? Calling an update metadata api for dataset or datafile with none UTC lastUpdateTimeStamp

* What happens? Fails as the times don't match because of the assumption that the time is in UTC

* To whom does it occur (all users, curators, superusers)? API user

stevenwinship:

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12349

- Closes #12349

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
@@ -61,7 +61,11 @@ import jakarta.ws.rs.core.Response.Status;
 
 import java.io.InputStream;
 import java.net.URI;
-import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoUnit;
 import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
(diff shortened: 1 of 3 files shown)


landreev:
I may be missing something, but I'm not sure if this is the right way to address the issue. 
We do NOT want to allow "non UTC lastUpdate timestamps". The method should never see any timestamps other than in UTC (with the "Z"). That's how Dataverse reports them, regardless of where the system lives. And from what I understand, in the workflow for which the API is designed, this timestamp will always come from the output of our APIs. 
For example, in the context of the test in FilesIT.java, it will be the output of `/api/files/<id>`, with something like
```
        "dataFile": {
             ...
            "publicationDate": "2026-04-20",
            "lastUpdateTime": "2026-04-20T20:25:05Z",
            "fileAccessRequest": true
         }
```
So, the way the format was hardcoded in `validateInternalTimestampIsNotOutdated()`, as 
```
DateUtil.parseDate(sourceLastUpdateTime, "yyyy-MM-dd'T'HH:mm:ss'Z'")
```
was in fact correct - I think it's fair to reject any other formats there. What was not correct was that it was parsed ignoring the time zone completely, so it defaulted to local. But it just needs to force the UTC, aka "+00:00" - and then everything will just work. I think?

src/main/java/edu/harvard/iq/dataverse/api/AbstractApiBean.java
```diff
@@ -488,21 +492,32 @@ public Command<DatasetVersion> handleLatestPublished() {
     }
 
     protected void validateInternalTimestampIsNotOutdated(DvObject dvObject, String sourceLastUpdateTime) throws WrappedResponse {
-        Date date = sourceLastUpdateTime != null ? DateUtil.parseDate(sourceLastUpdateTime, "yyyy-MM-dd'T'HH:mm:ss'Z'") : null;
-        if (date == null) {
+        LocalDateTime dtIN = null;
+        try {
+            dtIN = sourceLastUpdateTime != null ? LocalDateTime.parse(sourceLastUpdateTime, DateTimeFormatter.ISO_OFFSET_DATE_TIME) : null;
```
landreev, line 497: DateTimeFormatter.ISO_INSTANT is what we want here, I believe. 
That will force the "2011-12-03T10:15:30Z"-like format; and the resulting date will be parsed as UTC.

stevenwinship:
Closing

PR #12358, opened 2026-04-24
Enable S3 multipart upload (in non-direct upload)

janvanmansum:
This PR re-enables multipart upload for non-direct uploads for S3-compatible storage. At some point between 6.7.1 and 6.9 the default changed from enabled to disabled. This PR explicitly enables it when creating the `S3AsyncClient`. It will only be used when necessary anyway.

Without this change, an upload of a file larger than 1G  using a storage driver that does not support direct upload, the upload fails silently when clicking Save on the confirmation screen. So the file is uploaded to the Dataverse server, but then uploading it from there to the object store fails, without any error message to the user.

N/a

N/a

Perform an upload of a file larger than 1G, which will require the client to use multipart upload.

No

See: doc/release-notes/12358-enable-s3-multipart-upload.md

**Additional documentation**:
No

### DIFF
--- a/doc/release-notes/12358-enable-s3-multipart-upload.md
+++ b/doc/release-notes/12358-enable-s3-multipart-upload.md
@@ -1,0 +1,1 @@
+Fixed a defect that caused upload of files larger than 1G to fail silently for S3 storage, unless direct upload was used. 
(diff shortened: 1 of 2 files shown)


qqmyers (approved):
makes sense - thanks for the PR

landreev:
Wow, I was completely unaware of this! 
I was seeing evidence of large uploads failing in our prod. use, but I assumed it was a result of a recent migration to a new S3 bucket. 
Thank you for the PR, we will try to merge it asap.

bikramj:
Heads-up for anyone on S3-compatible storage (we use OpenStack Swift's S3 API):

  `multipartEnabled(true)` turns on multipart for **downloads** too, not just uploads.
  With it on, `getObject` first sends `GET <key>?partNumber=1` to fetch part 1. Swift's
  S3 API only supports that since **Swift 2.34.0 (OpenStack Dalmatian)** — older Swift
  returns **405 Method Not Allowed**, so downloads start failing (we saw it on thumbnails,
  logos, and cached exports on Swift 2.29 / Yoga). The upload fix itself worked.

  Probably no impact for AWS S3 users, but if anyone runs Swift older than 2.34, you may
  want to scope multipart to the upload path only (e.g. a separate multipart-enabled client
  just for the TransferManager, leaving downloads on the plain client). Just flagging so
  it's on the radar.

PR #12363, opened 2026-04-27
Fixed Anonymous guestbook file download via persistentId

### LINKED ISSUE #12245: Anonymous guestbook download via :persistentId returns a signed URL that immediately fails with 401 Bad signed URL, numeric dataset id works
Summary:
For a public dataset with a guestbook assigned, anonymous guestbook download through the access API behaves differently depending on the dataset identifier used. The :persistentId guestbook-signing path returns a signed URL that immediately fails with 401 Bad signed URL, while the numeric dataset id path returns a signed URL that works. The logged-in :persistentId path also works.

Create or use a published public dataset with a guestbook assigned.
As an anonymous user, submit a guestbook response to the access API using the :persistentId endpoint with signed=true.
Take the exact signedUrl returned by the API response.
Fetch that exact returned URL.
Repeat the same test using the numeric dataset id endpoint instead of :persistentId.
Repeat the :persistentId test as a logged-in user with a bearer token.

* When does this issue occur?
When an anonymous user downloads a public dataset that requires a guestbook response and the signed URL is generated through the :persistentId guestbook access endpoint.

It is visible from the SPA dataset page download flow for public datasets with guestbooks assigned. It is also reproducible directly at the API level via the access endpoints.

* What happens?
The anonymous POST /api/v1/access/dataset/:persistentId?...&signed=true request returns status: OK and returns a signedUrl, but fetching that exact returned signed URL immediately fails with:

{"status":"ERROR","message":"Bad signed URL"}
and HTTP status 401 Unauthorized.

By contrast:

anonymous POST /api/v1/access/dataset/2?signed=true succeeds and the returned signed URL works
logged-in POST /api/v1/access/dataset/:persistentId?...&signed=true succeeds and the returned signed URL works

It occurs for anonymous users (guest) downloading a public dataset with a guestbook through the :persistentId guestbook-signing path.

It does not reproduce for:

logged-in users using the :persistentId guestbook-signing path
anonymous users using the numeric dataset id guestbook-signing path

The exact signed URL returned by the anonymous :persistentId guestbook access POST should be valid and should download the dataset successfully, just like:

the numeric dataset id version for anonymous users
the :persistentId version for logged-in users

Latest develop branch version

**Screenshots:**

API-level reproduction commands and results:

Anonymous :persistentId path, fails:

PID='doi:10.5072/FK2/I0NJBF'

RESP=$(curl -sS -X POST "http://localhost:8000/api/v1/access/dataset/:persistentId?persistentId=${PID}&signed=true" \
  -H 'Content-Type: application/json' \
  -d '{
    "guestbookResponse": {
      "name": "Guest User",
      "email": "guest@example.com",
      "institution": "Test Institution",
      "position": "Researcher",
      "answers": []
    }
  }')

echo "$RESP" | jq .

SIGNED_URL=$(echo "$RESP" | jq -r '.data.signedUrl // empty')

if [ -n "$SIGNED_URL" ]; then
  curl -i "$SIGNED_URL"
fi
Observed:

POST returns status: OK
GET on exact returned signed URL returns 401 Unauthorized
body: {"status":"ERROR","message":"Bad signed URL"}
Anonymous numeric dataset id path, succeeds:

RESP=$(curl -sS -X POST "http://localhost:8000/api/v1/access/dataset/2?signed=true" \
  -H 'Content-Type: application/json' \
  -d '{
    "guestbookResponse": {
      "name": "Guest User",
      "email": "guest@example.com",
      "institution": "Test Institution",
      "position": "Researcher",
      "answers": []
    }
  }')

echo "$RESP" | jq .

SIGNED_URL=$(echo "$RESP" | jq -r '.data.signedUrl // empty')

if [ -n "$SIGNED_URL" ]; then
  curl -i "$SIGNED_URL"
fi
Observed:

POST returns status: OK
GET on exact returned signed URL returns 200 OK
Logged-in :persistentId path, succeeds:

TOKEN='<bearer-token>'
PID='doi:10.5072/FK2/I0NJBF'

RESP=$(curl -sS -X POST "http://localhost:8000/api/v1/access/dataset/:persistentId?persistentId=${PID}&signed=true" \
  -H "Authorization: Bearer ${TOKEN}" \
  -H 'Content-Type: application/json' \
  -d '{
    "guestbookResponse": {
      "name": "Dataverse Admin",
      "email": "dataverse@mailinator.com",
      "institution": "IQSS",
      "position": "Admin",
      "answers": []
    }
  }')

echo "$RESP" | jq .

SIGNED_URL=$(echo "$RESP" | jq -r '.data.signedUrl // empty')

if [ -n "$SIGNED_URL" ]; then
  curl -i "$SIGNED_URL"
fi
Observed:

POST returns status: OK
GET on exact returned signed URL returns 200 OK

stevenwinship:
**What this PR does / why we need it**:Guest downloads using persistentId were not working When guestbook responses were given.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12245

- Closes #12245

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
@@ -63,6 +63,11 @@ public class FilesIT {
         removePublicInstall.then().assertThat().statusCode(200);
     }
 
+    @AfterEach
+    public void resetClass() {
+        UtilIT.deleteSetting(SettingsServiceBean.Key.FilePIDsEnabled);
+    }
+
     @AfterAll
     public static void tearDownClass() {
         UtilIT.deleteSetting(SettingsServiceBean.Key.PublicInstall);
@@ -3869,6 +3874,7 @@ public class FilesIT {
     @Test
     public void testDownloadFileWithGuestbookResponse() throws IOException, JsonParseException {
         msgt("testDownloadFileWithGuestbookResponse");
+        UtilIT.enableSetting(SettingsServiceBean.Key.FilePIDsEnabled);
         // Create superuser
         Response createUserResponse = UtilIT.createRandomUser();
         assertEquals(200, createUserResponse.getStatusCode());
@@ -3929,7 +3935,9 @@ public class FilesIT {
         JsonObjectBuilder json4 = Json.createObjectBuilder().add("description", "my description4").add("directoryLabel", "data/subdir1").add("categories", Json.createArrayBuilder().add("Data"));
         uploadResponse = UtilIT.uploadFileViaNative(datasetId.toString(), "src/main/webapp/resources/images/Robot-Icon_2.png", json1.build(), ownerApiToken);
         uploadResponse.then().assertThat().statusCode(OK.getStatusCode());
+        uploadResponse.prettyPrint();
         Integer fileId4 = JsonPath.from(uploadResponse.body().asString()).getInt("data.files[0].dataFile.id");
+        String filePersistentId = JsonPath.from(uploadResponse.body().asString()).getString("data.files[0].dataFile.persistentId");
 
         // Restrict files
         Response restrictResponse = UtilIT.restrictFile(fileId1.toString(), true, ownerApiToken);
@@ -4044,13 +4052,18 @@ public class FilesIT {
         Response guestbookResponses = UtilIT.getGuestbookResponses(dataverseAlias, guestbook.getId(), ownerApiToken);
         assertTrue(guestbookResponses.prettyPrint().contains("My Name," + user2Email + ",My Institution,My Position"));
 
-        // Get Signed Download Url with guestbook response using persistentId
-        // POST /api/access/dataset/:persistentId?persistentId=doi:10.xxxx/FK2/ABC
-        downloadResponse = UtilIT.downloadFilesUrlWithGuestbookResponse(persistentId, apiToken, guestbookResponse);
+        // Get Signed Download Url for guest with guestbook response using file's persistentId
+        // POST /api/access/dataset/:persistentId?persistentId=
+        downloadResponse = UtilIT.downloadFilesUrlWithGuestbookResponse(filePersistentId, null, guestbookResponseForGuest);
         downloadResponse.prettyPrint();
         downloadResponse.then().assertThat()
                 .statusCode(OK.getStatusCode());
         signedUrl = UtilIT.getSignedUrlFromResponse(downloadResponse);
+        // verify that the fileId is correct
+        assertTrue(signedUrl.contains("/access/datafile/" + fileId4 + "?"));
+        // verify that the persistentId is no longer in the url
+        assertFalse(signedUrl.contains("persistentId"));
+        // verify that the signed url is good
         signedUrlResponse = get(signedUrl);
         assertEquals(OK.getStatusCode(), signedUrlResponse.getStatusCode());
     }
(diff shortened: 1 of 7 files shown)


landreev (commented):
One quick request: 
Could you please fix all the misspelled `signature-secret` in config.rst as well in the same pr.

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
```diff
@@ -1302,7 +1302,7 @@ static Response downloadFilesUrlWithGuestbookResponse(String persistentId, Strin
         if (body != null) {
             requestSpecification.body(body);
         }
-        String getString = "/api/access/dataset/:persistentId?persistentId=" + persistentId;
+        String getString = "/api/access/datafile/:persistentId?persistentId=" + persistentId;
```
landreev: From a quick glance, hard-coding "datafile" here may not be ideal. 
This UtilIT method is also called from `FilesIT.testDownloadFileWithGuestbookResponseUsingBearerToken()` method, where the pid appears to be of a dataset and in the context of  `/api/access/dataset`. 
I see that the test is `@Disabled`, but I'm assuming you want it to work on demand.

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
```diff
@@ -4044,13 +4056,18 @@ public void testDownloadFileWithGuestbookResponse() throws IOException, JsonPars
         Response guestbookResponses = UtilIT.getGuestbookResponses(dataverseAlias, guestbook.getId(), ownerApiToken);
         assertTrue(guestbookResponses.prettyPrint().contains("My Name," + user2Email + ",My Institution,My Position"));
 
-        // Get Signed Download Url with guestbook response using persistentId
-        // POST /api/access/dataset/:persistentId?persistentId=doi:10.xxxx/FK2/ABC
-        downloadResponse = UtilIT.downloadFilesUrlWithGuestbookResponse(persistentId, apiToken, guestbookResponse);
+        // Get Signed Download Url for guest with guestbook response using file's persistentId
+        // POST /api/access/dataset/:persistentId?persistentId=
+        downloadResponse = UtilIT.downloadFilesUrlWithGuestbookResponse(filePersistentId, null, guestbookResponseForGuest);
```
landreev, line 4069: Also, this looks like this test of `/api/access/datafile/:persistentId` is being added at the expense of dropping testing of `/api/access/dataset/:persistentId` that used to be here.
We probably want to test both, no?

stevenwinship: I created this test for specifically for this. It's only used in 1 place. It was incorrect. It should only be used for file persistentId

src/test/java/edu/harvard/iq/dataverse/api/UtilIT.java
stevenwinship: I fixed this test. It should be using the file persisted id

src/test/java/edu/harvard/iq/dataverse/api/FilesIT.java
stevenwinship: fixed the comment

landreev: @stevenwinship 
> I created this test for specifically for this. It's only used in 1 place. It was incorrect. It should only be used for file persistentId

My point was, this test, FilesIT.testDownloadFileWithGuestbookResponse() is working in the develop branch now, in its current form. 
I.e., it appears to be successfully testing `/api/access/dataset/:persistentId` - a download of an entire dataset-worth of files, with a guestbook and a signed url; even if that was not the intent of the test - would it not be useful, to have tests for *both* `/api/access/dataset/:persistentId` and `/api/access/datafile/:persistentId` going forward? 

... assuming I'm reading the code correctly, etc.

stevenwinship: added the test and made a fix to Access class to know which ids to use (dataset or datafile(s))

landreev (approved):
I meant to wait for Jenkins run to finish and move it into "ready"... but forgot about it/got distracted.

PR #12368, opened 2026-04-29
12367 GitHub actions workflow for jenkins redundancy

srmanda-cs:
**What this PR does / why we need it**:
Add a GitHub Actions Workflow that replicates the exact same tests that Jenkins run with the same configurations, but with containers instead

**Which issue(s) this PR closes**:
- Closes #12367 
- Closes #9916 
- Closes #9790

**Special notes for your reviewer**:
All your suggestions and criticism are deeply welcome, and since it is a major workflow I'm willing to work on it again and again until it meets Dataverse's needs accurately and addresses any concerns the team has. There will probably be a lot, since this is a prototype workflow that I quickly hacked together in a few weeks.

**Suggestions on how to test this**:
Clicking run workflow should do it.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:
N/A

**Is there a release notes update needed for this change?**:
N/A

**Additional documentation**:
Depends on where and how you would want me to document it.

PR #12372, opened 2026-04-30
Fix DataRetrieverApiIT.testRetrieveMyDataCollections() for github mig…

### LINKED ISSUE #12371: Fix DataRetrieverApiIT.testRetrieveMyDataCollections() on Github actions
DataRetrieverApiIT.testRetrieveMyDataCollections() is failing on Github Actions but passes in Jenkins.

This relates to Issue: https://github.com/IQSS/dataverse/pull/12368

stevenwinship:
**What this PR does / why we need it**: DataRetrieverApiIT.testRetrieveMyDataCollections() fails in Github

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/12371

- Closes #12371

**Additional documentation**:

### DIFF
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
@@ -115,7 +115,6 @@ public class DataRetrieverApiIT {
     @Disabled("Temporarily disabled because this integration test is not reliable in CI; re-enable once stabilized. All assertions return one extra dataset than expected.")
     @Test
     public void testRetrieveMyDataCollections() throws InterruptedException {
-        int rootCount = 1; // everyone has access to this dataverse
         List<Map<String, String>> items;
         Response createDataverseResponse;
         Response retrieveMyCollectionListResponse;
@@ -142,7 +141,7 @@ public class DataRetrieverApiIT {
         // Get the base number of collections since it's not always 1 for root.
         // There may be others left from another test that everyone can access
         retrieveMyCollectionListResponse = UtilIT.retrieveMyCollectionList(User1ApiToken, null);
-        rootCount = retrieveMyCollectionListResponse.getBody().jsonPath().getList("data.items").size();
+        int rootCount = retrieveMyCollectionListResponse.getBody().jsonPath().getList("data.items").size();
 
         // User1 creates 15 Dataverses and adds a role to each allowing User2 access
         List<String> dataverses = new ArrayList<>();
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
@@ -112,7 +112,6 @@ public class DataRetrieverApiIT {
     }
 
     // Test getting a list of collections that the user can add datasets to
-    @Disabled("Temporarily disabled because this integration test is not reliable in CI; re-enable once stabilized. All assertions return one extra dataset than expected.")
     @Test
     public void testRetrieveMyDataCollections() throws InterruptedException {
         List<Map<String, String>> items;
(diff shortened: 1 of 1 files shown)


stevenwinship:
related to issue https://github.com/IQSS/dataverse/pull/12368

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
```diff
@@ -137,6 +137,11 @@ public void testRetrieveMyDataCollections() throws InterruptedException {
         String User3Username = UtilIT.getUsernameFromResponse(createUserResponse);
         String User3ApiToken = UtilIT.getApiTokenFromResponse(createUserResponse);
 
+        // Get the base number of collections since it's not always 1 for root.
+        // There may be others left from another test that everyone can access
+        retrieveMyCollectionListResponse = UtilIT.retrieveMyCollectionList(User1ApiToken, null);
+        rootCount = retrieveMyCollectionListResponse.getBody().jsonPath().getList("data.items").size();
```
pdurbin: Should we delete `int rootCount = 1; // everyone has access to this dataverse` above and put an `int` in front of rootCount here?

pdurbin (commented):
A quick look.

src/test/java/edu/harvard/iq/dataverse/api/DataRetrieverApiIT.java
stevenwinship: sure

ashmanda-cs:
After a quick check a few days back when I changed User 1 to greater than or equal to, the assertions seemed to be failing for User 2 and User 3 as well. There was always one extra dataset than expected. It was a different ID from user1's extra collection but it remained consistent from test to test. This is the primary reason why I changed it to disabled instead.

stevenwinship:
@srmanda-cs Did the new test show that the root count changed from 1 to 2? If not then there is another issue. The root count should be added to each user's count and therefore work. I can't get it to fail locally so it's hard to test

ashmanda-cs:
> @srmanda-cs Did the new test show that the root count changed from 1 to 2? If not then there is another issue. The root count should be added to each user's count and therefore work. I can't get it to fail locally so it's hard to test

I realize that it is incredibly difficult to test, because it is so environment specific :(

I have little to no idea why that specific test is failing. All I realize is that when I changed it to greater than or equal to for user1, user2 returned 18 count instead of the expected 17. The root dataverse count remains constant, it is the number of other dataverses that the users have access to that have increased. Its very similar to how the Mail test was the only one failing along with this when I just ran a full Maven verify.

And as I have mentioned to Phil previously, I really don't want to create more work for y'all than you already have. I can provide you with more testing, more logs, and more specific details but it will take me a lot of time to dig into this as its a novel codebase to me. Which is why I sought your help.

pdurbin:
I'm thinking we should merge the new GitHub Action first:

- #12368

Then we should rebase this PR (#12372) after the merge and see if it's passing or not.

stevenwinship:
@srmanda-cs It's not making more work for us. It's helping us understand the issues that can arise from the migration. I have gone through this migration before between Jenkins and GitLab. Asynchronous tests and Test suites that have different settings (JVM, etc.) can greatly speed up testing but do make testing a bit more challenging.
 If you can comment here with the logs I would appreciate it.

ashmanda-cs:
This is one of the runs that outlines the specific test failure: https://github.com/IQSS-Dataverse-Development/dataverse/actions/runs/25069771575

The specific workflow file, the code changes, the full docker logs, and the full execution run can be obtained from these single links. If you click on the workflow button, then go to the gear icon, and click download raw logs you should see multiple megabytes worth of raw logs.

This is what I've noticed happens when I change it to greater than or equal to for user1, user 2 starts failing: https://github.com/IQSS-Dataverse-Development/dataverse/actions/runs/25071885059

I believe these logs to be very comprehensive and thorough, but also immensely dense and hard to read. I genuinely appreciate your help! My initial thoughts were that they're a permissions issue letting :authenticated-users see more dataverses than they're allowed to. But jenkins truncates all test output, so I couldn't directly verify it. Please let me know wherever I can help, I am ready to jump in.

stevenwinship:
I think the better solution is to not look at the size of the list but to verify the contents include the desired collections

ashmanda-cs:
> I think the better solution is to not look at the size of the list but to verify the contents include the desired collections

If it fits the purpose of why the test exists, then it sounds like an amazing solution to me!

pdurbin:
@stevenwinship I just merged this PR:

- #12368

Can you please update this branch and get the test working again? Thanks!

pdurbin (approved):
Both Jenkins and GitHub Actions are showing that the test is fixed so I'll merge this.

Image

Image

Issue #9969, opened 2023-10-03
OIDC: enable mapping `groups` claim to user groups


poikilotherm:
**Overview of the Feature Request**

Currently, we cannot make the OIDC provider put remote users into certain groups based on the `groups` claim that can be received from an identity provider. This is crucial for SPA functionality and other use cases.

**What kind of user is the feature intended for?**
Sysadmin

**What inspired the request?**
The upcoming SPA and working on #9268

**What existing behavior do you want changed?**
Map users to groups using admin specified configuration.

**Any brand new behavior do you want to add to Dataverse?**
Nope

**Any open or closed issues related to this feature request?**
#9268 #8349 #6679 #5974

cmbz:
2024/01/08: @scolapasta will follow up with Oliver to discuss.

johannes-darms:
Shall we create a document to draft ideas how this can be resolved/implemented?

pdurbin:
@johannes-darms sure! Here's a doc! https://docs.google.com/document/d/1eSmPoPSiBVO_KMJTorVR0Jaomi8BnbjjDI8ZpFtDI48/edit?usp=sharing

Issue #6679, opened 2020-02-25
OIDC: make use of attribute "email_verified"


poikilotherm:
When we introduced OpenID Connect support back in #5974 and PR #6433, we left out a few things. One of those was support for the email verified status of new user accounts.

Currently, only the Shibboleth provider offers support for this:

https://github.com/IQSS/dataverse/blob/9ce4aa00603d9fd9233a5c4d99b527ed230af1ce/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationServiceBean.java#L609-L618

Within OpenID Connect, we have a defined scope `email` attribute `email_verified`, which can be set by the provider. (This is one of the points where OIDC offers more that OAuth2 only...).
We should start to support this, as it makes the process more easy, when the IDM/IAM/OIDC provider already did the verification for us.

I'm not sure if UI/UX team is involved here, as there is no UI change necessary.

djbrooke:
@poikilotherm, sorry if this is a naive question, but is there a security risk in letting the provider set the email_verified status? Some providers ostensibly would be more trustworthy than others. Would we want the installation admin to set this instead?

pdurbin:
Related:

- Shibboleth users who predate Shibboleth are assumed to be email-verified but aren't #5663 
- Email confirmation: user account restriction #3300

poikilotherm:
Hi @djbrooke,
obviously you will need to trust your provider :wink: 

To provide an example: https://login.helmholtz-data-federation.de which we are using for Jülich DATA, is an IDM based on [Unity](http://www.unity-idm.eu). It has rules configured that people signing in via SAML, having the IdP send an email address and originating from inside DFN AAI (German Research Network), have set the attribute email_verified to `true`. Anyone else is untrusted.
That's fine for us, but obviously your mileage may vary.

I am more than happy to make this configurable, so the admin can decide whether the provider shouldn't be trusted at all. There already is an attribute in `AuthenticationProvider`, which looks like a perfect match to me:
https://github.com/IQSS/dataverse/blob/9ce4aa00603d9fd9233a5c4d99b527ed230af1ce/src/main/java/edu/harvard/iq/dataverse/authorization/AuthenticationProvider.java#L49-L53

So when setting the attribute, it would be a simple `&&` comparison with this attribute. We should leave this as `false` for a sane default (better safe than sorry).

I'll create another issue at this point about refactoring the configuration, as this should be done before this then.

pdurbin:
[...]

PR #12378, opened 2026-05-05
feat(auth): sync OIDC user properties

### LINKED ISSUE #10266: user data is not updated when changed in OIDC IdP.
**What steps are required to reproduce the problem?**
1) Configure the Dataverse with OIDC authentication.
2) Create a new user via OIDC
3) Log out via OIDC
3) Change email, first name or surname in the OIDC IdP provider
4) Login via OIDC, the updated values are not available in Dataverse. 
**When does this problem occur?**

**What happens?**
Changes to properties within the IdP are not reflected in dataverse. 

**For whom does it occur (all users, curators, super users)?**
All OIDC users.

The dataverse user data is updated as soon as an OIDC user is authenticated. Even better would be to use the IdP directly to get the information, but updating the information at each login would at least get round the problem. 

5.13

**Are there any open or closed issues related to this bug report?**

AnneGerlach:
### What this PR does / why we need it:

This PR adds synchronization of user properties from the OIDC identity provider to the Dataverse `AuthenticatedUser`.

Currently, Dataverse authenticates users via bearer tokens but does not persist updated user properties from the OIDC provider. This can lead to outdated user data (e.g. incorrect email addresses), which becomes problematic for features such as notifications or user-facing metadata.

This implementation follows a **comparison-based approach**:

- On each authenticated request, user properties from the `OAuth2UserRecord` (OIDC / Keycloak) are compared with the stored `AuthenticatedUser`
- Updates are only applied if values differ
- This avoids unnecessary database writes and keeps the solution efficient

The following fields are synchronized:

- `firstName`
- `lastName`
- `email`
- email verification state

In addition, this PR synchronizes the OIDC `email_verified` claim with Dataverse’s internal representation of email verification (`emailConfirmed` timestamp).

Mapping:

- `email_verified = true` → set `emailConfirmed` timestamp  
- `email_verified = false` → set `emailConfirmed` to `null`  
- claim missing → keep existing value unchanged  

This PR also addresses a small UI inconsistency:
(See screenshots below)

Dataverse keeps the `AuthenticatedUser` in the session, which can lead to stale values being displayed in the Account Information page even after the database has been updated.

To resolve this:

- The user is reloaded from the database using a cache-bypassing query
- Session user properties are updated **in-place**
- This avoids using `session.setUser(...)`, which would otherwise invalidate the session/view state

---
**:black_flag: The feature is controlled via a feature flag:**

`dataverse.feature.oidc-user-property-sync`

- Disabled by default
- Allows safe rollout and testing

---
### Which issue(s) this PR closes:

- Closes #10266  
- Related:
  - #9969
  - #6679

---

### Special notes for your reviewer:

- The `email_verified` claim is treated as optional to ensure compatibility with providers that do not expose this claim
- Email synchronization prefers `displayInfo.email` and falls back to available email addresses if needed
- Updates are only persisted when values actually change

Alternative approaches were considered:

- **Timestamp-based sync (`updated_at` claim)**  
  Not feasible, as this claim is not provided by Keycloak by default and would require schema changes

- **Push-based updates from the IdP**  
  Not suitable, as the backend should not depend on external triggers or actively query the identity provider

👉 Therefore, a comparison-based approach was chosen to keep the implementation simple and compatible with any OIDC provider

---

### Suggestions on how to test this:

Run all tests:

```bash
mvn test
```
or only the relevant ones in:

```bash
mvn -Dtest=AuthenticationServiceBeanTest test
```
Optional manual verification:
Before testing, ensure that an OIDC authentication provider is configured (in my case Keycloak).  

- Create/login a user via OIDC in the local setup (docker-compose-dev.yml)
- Update user properties in Keycloak (e.g. firstname, email)
- Generate a new token and call:

```bash
curl -i \
  -H "Authorization: Bearer $TOKEN" \
  http://localhost:8080/api/users/:me
```

Expected:

- The response reflects the updated values  
- The changes are persisted in the Dataverse database  
- No update is triggered if values remain unchanged
- the Account Information page reflects the updated first name, last name, and email without requiring logout/login

Example response:

```json
{
  "status": "OK",
  "data": {
    "id": 2,
    "identifier": "@user",
    "displayName": "NewFirst NewLast",
    "firstName": "NewFirst",
    "lastName": "NewLast",
    "email": "newemail-user@mailinator.com",
    ...
  }
}
```

:handshake: If helpful, I can provide a more detailed step-by-step manual test guide.

---

### Does this PR introduce a user interface change? If mockups are available, please link/include them here:

No new UI is introduced.

However, this PR fixes an issue where the Account Information page could display stale user data after OIDC updates.

Before:
image

After:
image

### Is there a release notes update needed for this change?:

Yes, a release note has been added:  doc/release-notes/10266-oidc-property-sync.md

### Additional documentation:

Documentation for the new feature flag has been added to the configuration guide.

Thanks for reviewing :sunflower:


johannes-darms:
@pdurbin @poikilotherm another feature from us ;)

pdurbin:
@vera can you please review this and if you're happy with it, resolve merge conflicts?

vera:
@pdurbin I've resolved the merge conflicts

PR #12381, opened 2026-05-05
Featured Items API can now retrieve images for creator

### LINKED ISSUE #11537: Feature Items API Unable to Retrieve Images from Unpublished Collections

1. Create a collection (unpublished)
2. Create a feature item with images and text 
3. Retrieve the feature items information you just created, there is no image

* When does this issue occur?
Feature items 

* What happens?
When the collection is created by me, but it is not published, then I create a feature item with image, the image wouldn't be shown. 

spa users

If the collection is created by me but not published, the images should be shown.

stevenwinship:
**What this PR does / why we need it**: Creator of a featured item of a non published dataverse could not see the images.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/11537

- Closes #11537

**Suggestions on how to test this**: See DataversesIT testListFeaturedItems. 

**Additional documentation**:

### DIFF
--- a/doc/release-notes/11537-featured-items-retrieve-image-bug.md
+++ b/doc/release-notes/11537-featured-items-retrieve-image-bug.md
@@ -1,0 +1,3 @@
+## BUG
+
+Featured Item creator can now view/download images when a dataverse is not published. ViewUnpublishedDataverse is not required for the featured item creator.
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseFeaturedItemCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/GetDataverseFeaturedItemCommand.java
@@ -30,8 +30,11 @@ public class GetDataverseFeaturedItemCommand extends AbstractCommand<DataverseFe
 
     @Override
     public Map<String, Set<Permission>> getRequiredPermissions() {
-        return Collections.singletonMap("",
-                dataverseFeaturedItem.getDataverse().isReleased() ? Collections.emptySet()
-                        : Collections.singleton(Permission.ViewUnpublishedDataverse));
+        // If the dataverse is not released only a user with ViewUnpublishedDataverse permissions or the creator can access the featured item and its images
+        if (!dataverseFeaturedItem.getDataverse().isReleased() && !getRequest().getUser().equals(dataverseFeaturedItem.getDataverse().getCreator())) {
+            return Collections.singletonMap("", Collections.singleton(Permission.ViewUnpublishedDataverse));
+        } else {
+            return Collections.singletonMap("",Collections.emptySet());
+        }
     }
 }
--- a/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
+++ b/src/test/java/edu/harvard/iq/dataverse/api/DataversesIT.java
@@ -2184,6 +2184,17 @@ public class DataversesIT {
                 .body("data[2].type", equalTo("custom"))
                 .statusCode(OK.getStatusCode());
 
+        // Verify that the unpublished image can be downloaded by its creator and not by a user without ViewUnpublishedDataverse permissions
+        JsonPath path = JsonPath.from(listDataverseFeaturedItemsResponse.body().asString());
+        String imageUrl = path.getString("data[2].imageFileUrl");
+        Response downloadResponse = given().get(imageUrl + "?key=" + apiToken);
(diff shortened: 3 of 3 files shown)


rtreacy (approved):
Claude
PR #12381 Review: Featured Items API can now retrieve images for creator

What it does

Fixes #11537: the creator of an unpublished dataverse couldn't view/download their own featured-item images, because GetDataverseFeaturedItemCommand.getRequiredPermissions() unconditionally required Permission.ViewUnpublishedDataverse whenever the dataverse wasn't released — with no allowance for the dataverse's own creator. The fix adds an explicit carve-out: if the requesting user equals dataverseFeaturedItem.getDataverse().getCreator(), no permission is required, same as the published case.

Correctness

- getRequest().getUser().equals(...) is safe against the two null-ish edge cases that matter: an anonymous/guest user (GuestUser.equals falls back to reference equality against an AuthenticatedUser → false, correctly not treated as creator) and a dataverse with no recorded creator (getCreator() returns null; AuthenticatedUser.equals(null) returns false via the instanceof guard, so the code correctly falls back to requiring ViewUnpublishedDataverse rather than NPEing or accidentally granting access).
- AuthenticatedUser.equals() compares by database ID (DvObject.java/AuthenticatedUser.java:439), not reference identity, so comparing the request's live user object against the JPA-loaded creator relationship works correctly across different object instances.
- The fix is narrowly scoped — it changes behavior only for the literal creator of the dataverse; every other user (including dataverse admins/curators without the creator flag) still goes through the unchanged ViewUnpublishedDataverse check, so there's no broadened attack surface beyond the intended case.
- @AuthRequired on the Access.getDataverseFeatureItemImage endpoint guarantees an authenticated user reaches this command, so there's no unauthenticated-guest path to worry about here.

Test coverage

- The new assertions in testListFeaturedItems exercise the actual bug: the dataverse in that test is never published (unlike the sibling testUpdateFeaturedItems, which explicitly publishes), so isReleased() is false and the fixed branch is genuinely exercised — this isn't a test that would've passed before the fix, which is exactly what you want for a regression test.
- Confirms both sides: creator's token → 200, unrelated random user's token → 204 (matching this endpoint's existing convention of returning null/204 rather than 403 on permission failure — pre-existing behavior in Access.java, unaffected by this PR).
- No test for a non-creator user who does have ViewUnpublishedDataverse (e.g., a curator role-assignee) still succeeding — that path is unchanged by this diff and was presumably already covered elsewhere, so it's a minor gap rather than a real risk.

Style (minor, non-blocking)

- Collections.singletonMap("",Collections.emptySet()) is missing a space after the comma — trivial whitespace nit.
- The double-negative condition (!isReleased() && !user.equals(creator)) reads a little awkwardly; something like a single boolean (isCreator) combined with isReleased() || isCreator in a ternary would be marginally more readable, but this is a matter of taste, not a defect.

Verdict

Small, well-targeted bug fix with a test that actually proves the fix works and would have failed before it. No correctness, security, or regression concerns. Good to merge as-is; the style nits aren't worth blocking on.

youseihuayu-wonderful:
Thanks Philip @pdurbin — I followed your guidance and ran QA locally with Docker.

Environment: mvn -Pct clean package docker:run
Branch: 11537-featured-items-retrieve-image-bug
Commit: d45be966476c1e4023de93ea977211f8f87f9ccf

I focused on the API path for #11537 as you suggested (SPA frontend not in scope for this round):

Created an unpublished collection as User A and added a featured item with an image
As the creator, downloading the featured-item image returned HTTP 200
As an unrelated authenticated user, downloading the same image returned HTTP 204
Also re-ran DataversesIT#testListFeaturedItems locally — passed
From an API QA perspective, I think #12381 is ready to merge.

youseihuayu-wonderful (approved):
API QA completed locally with Docker.

**Environment**
- Command: `mvn -Pct clean package docker:run`
- Branch: `11537-featured-items-retrieve-image-bug`
- Commit: https://github.com/IQSS/dataverse/commit/d45be966476c1e4023de93ea977211f8f87f9ccf

**Scope**
API path for https://github.com/IQSS/dataverse/issues/11537 (SPA frontend out of scope for this round).

**Results**
- Created an unpublished collection as User A and added a featured item with an image
- As the creator, downloading the featured-item image returned HTTP 200
- As an unrelated authenticated user, downloading the same image returned HTTP 204
- `DataversesIT#testListFeaturedItems` passed locally

**Conclusion**
From an API QA perspective, https://github.com/IQSS/dataverse/pull/12381 is ready to merge.

Approving.

PR #12397, opened 2026-05-18
Fixes to counter-weekly and updateCitationsForDataset API

### LINKED ISSUE #12384: Fixes for APIs related to Make Data Count when trying to update datasetexternalcitations
The purpose of this issue is to get counter_weekly.sh running in Harvard Dataverse.

API http://localhost:8080/api/dataverses/$1/contents encounters a null pointer exception in Harvard Dataverse preventing the counter_weekly script from running. The 2 values (id and name) are both defined as not null in the database but are giving null when the Json is being created.

API internal error 4f4fde25-388e-4c1f-b294-32fc629c84f3: Value in JsonObjects name/value pair cannot be null
        at org.eclipse.parsson.JsonObjectBuilderImpl.validateValue(JsonObjectBuilderImpl.java:213)
        at org.eclipse.parsson.JsonObjectBuilderImpl.add(JsonObjectBuilderImpl.java:80)
        at edu.harvard.iq.dataverse.api.Dataverses$1.visit(Dataverses.java:1210)
        at edu.harvard.iq.dataverse.api.Dataverses$1.visit(Dataverses.java:1205)
        at edu.harvard.iq.dataverse.Dataverse.accept(Dataverse.java:871)

It may also be necessary to add paging to the api as the number of dataverses being returned could cause a timeout.

Related Issue: https://github.com/IQSS/dv-api-metrics/issues/14

stevenwinship:
**What this PR does / why we need it**: Logging was only displaying :persistentId for the dataset id instead of the actual db id

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12384

- Closes #12384

**Special notes for your reviewer**: counter-weekly.sh exists in 2 places. I copied the script to the doc area instead of just removing one and modifying the doc.

**Additional documentation**:

### DIFF
--- a/conf/mdc/counter_weekly.sh
+++ b/conf/mdc/counter_weekly.sh
@@ -31,7 +31,7 @@ case $HTTP_STATUS in
         # Successfully queued
         # Extract status from the nested data object
         STATUS=$(echo "$RESPONSE_BODY" | jq -r '.data.status')
-        
+
         # Extract message from the nested data object
         if echo "$RESPONSE_BODY" | jq -e '.data.message' > /dev/null 2>&1 && [ "$(echo "$RESPONSE_BODY" | jq -r '.data.message')" != "null" ]; then
             MESSAGE=$(echo "$RESPONSE_BODY" | jq -r '.data.message')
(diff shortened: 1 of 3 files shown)


rtreacy (approved):
update works and logging is improved

PR #12403, opened 2026-05-20
Auto Create thumbnail so Search API will return image_url when publis…

### LINKED ISSUE #11586: Feature Request: Add imageUrl to Get Dataset API JSON response
For the View Dataset page, the SPA needs a thumbnail image in the Dataset Citation block, if one is available.  To support this, we need to add an optional imageUrl property to the json response.

API User

 https://github.com/IQSS/dataverse-frontend/issues/203

### LINKED ISSUE #11588: Search API: doesn't return image_url for newly created dataset
After creating a dataset with an image file in the SPA, the search API does not return an image_url for the dataset.
It seems that an extra step is needed first, which happens when the logo is requested.  This issue occurs even for a published unrestricted dataset, so it isn't related to permissions.

1. Create a dataset with an image file in the SPA. 
2. Publish the dataset
3. Go to the Collections page.  The search API call for the Collections page returns the datasetPreview without a image_url property.
4. Go to the JSF and view the Collections page.  This triggers a call to datasets/{datasetId}/logo
5. Go back to the Collections page in the SPA - now the image_url is returned in the datasetPreview for the page

* When does this issue occur?
With a Dataset that is created in the SPA

Collections Page (search API)

* What happens?
image_url is not returned

all users

image_url should be returned for a thumbnail of the image file in the dataset

**Screenshots:**

SPA - search results with missing image_url
Image

View the Page in JSF
Image

Go back to SPA -  now image_url appears in search results
Image

stevenwinship:
**What this PR does / why we need it**: SPA not getting image_url through Search API. JSF does get the image_url.

**Which issue(s) this PR closes**: https://github.com/IQSS/dataverse/issues/11588

- Closes #11588
- Closes #11586

With this PR, the behavior should be the same JSF vs the SPA: a thumbnail for the dataset will be auto-selected when the dataset is published; unless a specific thumbnail has already been picked by the author while it was still in draft, OR auto-selection has been disabled on the instance (such as on qa.dataverse.org).  

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
@@ -1,6 +1,5 @@
 package edu.harvard.iq.dataverse.engine.command.impl;
 
-import com.rometools.utils.Lists;
 import edu.harvard.iq.dataverse.ControlledVocabularyValue;
 import edu.harvard.iq.dataverse.CurationStatus;
 import edu.harvard.iq.dataverse.DataFile;
@@ -18,8 +17,6 @@ import edu.harvard.iq.dataverse.Embargo;
 import edu.harvard.iq.dataverse.UserNotification;
 import edu.harvard.iq.dataverse.authorization.Permission;
 import edu.harvard.iq.dataverse.authorization.users.AuthenticatedUser;
-import edu.harvard.iq.dataverse.dataaccess.ImageThumbConverter;
-import edu.harvard.iq.dataverse.dataset.DatasetThumbnail;
 import edu.harvard.iq.dataverse.dataset.DatasetUtil;
 import edu.harvard.iq.dataverse.engine.command.CommandContext;
 import edu.harvard.iq.dataverse.engine.command.DataverseRequest;
@@ -27,7 +24,6 @@ import edu.harvard.iq.dataverse.engine.command.RequiredPermissions;
 import edu.harvard.iq.dataverse.engine.command.exception.CommandException;
 import edu.harvard.iq.dataverse.pidproviders.PidProvider;
 import edu.harvard.iq.dataverse.privateurl.PrivateUrl;
-import edu.harvard.iq.dataverse.settings.FeatureFlags;
 import edu.harvard.iq.dataverse.util.BundleUtil;
 import edu.harvard.iq.dataverse.workflow.WorkflowContext;
 import edu.harvard.iq.dataverse.workflow.WorkflowContext.TriggerType;
@@ -178,7 +174,9 @@ public class FinalizeDatasetPublicationCommand extends AbstractPublishDatasetCom
 
         // Populate thumbnail if needed and allowed
         if (theDataset.getThumbnailFile() == null && !theDataset.isUseGenericThumbnail()) {
-            ctxt.datasetVersion().getThumbnailByVersionId(version.getId());
+            Long thumbnailFileId = ctxt.datasetVersion().getThumbnailByVersionId(version.getId());
+            theDataset.setThumbnailFile(ctxt.datasetVersion().getDataFileById(thumbnailFileId));
+            logger.info("Setting default thumbnail " + theDataset.getThumbnailUrl());
         }
 
         // 
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
@@ -175,8 +175,10 @@ public class FinalizeDatasetPublicationCommand extends AbstractPublishDatasetCom
         // Populate thumbnail if needed and allowed
         if (theDataset.getThumbnailFile() == null && !theDataset.isUseGenericThumbnail()) {
             Long thumbnailFileId = ctxt.datasetVersion().getThumbnailByVersionId(version.getId());
-            theDataset.setThumbnailFile(ctxt.datasetVersion().getDataFileById(thumbnailFileId));
-            logger.info("Setting default thumbnail " + theDataset.getThumbnailUrl());
+            if (thumbnailFileId != null) {
+                theDataset.setThumbnailFile(ctxt.datasetVersion().getDataFileById(thumbnailFileId));
+                logger.info("Setting default thumbnail " + theDataset.getThumbnailUrl());
+            }
         }
 
         // 
--- a/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
+++ b/src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
@@ -1,5 +1,6 @@
 package edu.harvard.iq.dataverse.engine.command.impl;
 
+import com.rometools.utils.Lists;
 import edu.harvard.iq.dataverse.ControlledVocabularyValue;
 import edu.harvard.iq.dataverse.CurationStatus;
 import edu.harvard.iq.dataverse.DataFile;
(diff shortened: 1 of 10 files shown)


src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
```diff
@@ -165,6 +170,15 @@ public PublishDatasetResult execute(CommandContext ctxt) throws CommandException
                 lock.setInfo(info);
                 ctxt.datasets().addDatasetLock(theDataset, lock);
             }
+
```
landreev: I very much like the idea of performing the auto-selection during the publication of the dataset. 

I will say upfront that that is the only thing I like about the PR, unfortunately. Everything else will have to be rearranged a bit, sorry. 

For starters, let's move this code to the FinalizeDatasetPublicationCommand. I.e., the second part of the publication process that is performed asynchronously, in the background. Any tasks that have any potential for taking an appreciable amount of time should always go there. 

I will add separate comments on everything else.

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
```diff
@@ -165,6 +170,15 @@ public PublishDatasetResult execute(CommandContext ctxt) throws CommandException
...
+            // Populate thumbnail if needed (Only look for DataFiles)
+            if (theDataset.getThumbnailFile() == null && !FeatureFlags.DISABLE_DATASET_THUMBNAIL_AUTOSELECT.enabled()) {
```
landreev: This logic also needs to include 
``` && ! theDataset.isUseGenericThumbnail() ```

File: src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java [hunk 3]
```diff
@@ -165,6 +170,15 @@ public PublishDatasetResult execute(CommandContext ctxt) throws CommandException
...
+                List<DatasetThumbnail> candidatesList = DatasetUtil.getThumbnailCandidates(theDataset, false, ImageThumbConverter.DEFAULT_CARDIMAGE_SIZE);
```
landreev: Most importantly, we definitely DO NOT WANT to use ```DatasetUtil.getThumbnailCandidates()``` for this!
If you look at the method, it does attempt to get a thumbnail _for every image in the dataset_. (i.e., it does not stop after finding the first valid thumbnail). We don't want to get stuck processing tens of thousand of TIFF files here. 

The method was not on my radar, I must have forgotten about it. But it appears that it was written specifically for the thumbnail selection widget, to give the author a choice of every thumbnail in the dataset. It is also called from the API method that serves the SPA equivalent. I am not entirely happy about its use in that scenario either; plus I am seeing some other bad things in it. But it is outside of the scope of this issue so I will bring that up separately. 

But, for the purposes of selecting just one random thumbnail for the dataset we need to do something else here. 
One quick and easy way would be to call `DatasetVersionServiceBean.getThumbnailByVersionId()`, as in 
```
ctxt.datasetVersion().getThumbnailByVersionId(theDataset.getLatestVersion().getId())
```

The method above is used when the auto-selection logic is triggered from jsf. It is mostly my old code that you also moved around in 2024. It is super optimized and relies on db queries to, hopefully, quickly find one image with ```.isPreviewImageAvailable()``` that is already present; it appears to cut some corners when it cannot, but it is guaranteed to be fast. The method was created specifically for the UI, where we never want to have the user waiting in real time for too long. If this is an asynchronous publication task, we could afford to take longer and be more thorough in trying to find a workable image. Your call I would say. 

Among other things, the method above relies on a NativeQuery to assign the found thumbnail to the dataset as well; which in this case may conflict with the  command trying to update the same dataset; so something may need to be moved around.

landreev: (OTOH, if we were to rely on the `.getThumbnailByVersionId()` method above, the logic could be left in `PublishDatasetCommand`; since it is guaranteed to be fast)

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
stevenwinship: moved code to FinalizeDatasetPublicationCommand

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
stevenwinship: done

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/PublishDatasetCommand.java
stevenwinship: Switched to use getThumbnailByVersionId()

src/main/java/edu/harvard/iq/dataverse/engine/command/impl/FinalizeDatasetPublicationCommand.java
```diff
@@ -171,7 +175,12 @@ public Dataset execute(CommandContext ctxt) throws CommandException {
         
         //Use dataset pub date (which may not be the current date for migrated datasets)
         updateFiles(new Timestamp(version.getReleaseTime().getTime()), ctxt);
-        
+
+        // Populate thumbnail if needed and allowed
+        if (theDataset.getThumbnailFile() == null && !theDataset.isUseGenericThumbnail()) {
```
landreev, line 176: One last thing: we still want to have ```&& !FeatureFlags.DISABLE_DATASET_THUMBNAIL_AUTOSELECT.enabled()``` here as well. 
Disabling auto-selection instance-wide with the feature flag above is not going to set the useGeneric boolean on every existing dataset.

stevenwinship: That flag is already checked in ctxt.datasetVersion().getThumbnailByVersionId(version.getId());

public Long getThumbnailByVersionId(Long versionId) {
        if (versionId == null) {
            return null;
        }

        if (!FeatureFlags.DISABLE_DATASET_THUMBNAIL_AUTOSELECT.enabled()) {
            Long thumbnailFileId;

landreev (approved):

landreev:
Hopefully Ellen can test and merge this quickly when she is back.

ekraffmiller:
@landreev @stevenwinship 
I tried the scenario again, getting the same behavior: 
1 create a dataset with image file, view the search results on Modern ui - thumbnail is missing.
2. go to JSF search results - thumbnail is populated.
3. go back to Modern UI, thumbnail is displayed.

https://github.com/user-attachments/assets/811fa0ff-51b0-4485-81f6-7229d951f08c

stevenwinship:
I wonder if it's being cached? Can you view it from another browser to see?

ekraffmiller:
> I wonder if it's being cached? Can you view it from another browser to see?

When I look at the API call from within the browser, the API response doesn't contain the image_url, so I don't think it's a caching issue.

landreev:
Is there anything noteworthy in the server log?

landreev:
... during publishing specifically.

stevenwinship:
I see this after publishing (in Chrome)

image

stevenwinship:
This was on Firefox

image

stevenwinship:
I see the difference now.

I added the file and immediately pressed publish and I see the image
If I add the file then go back to the search list and select and publish I don't see the image

image


Is it possible that this is a UI issue? The api doesn't know the sequence of events.

landreev:
Very interesting! Does the dataset end up with useGenericThumbnail=true in the process?

stevenwinship:
The issue with the images not showing up following Ellen's workflow was due to a race condition where the thumbnail was being updated via a sql call and the Dataset was being merged via java. If the Dataset merge happened first the sql update would be fine. If the sql update happened first the merge would put back an empty thumbnail. The fix was to add the thumbnail to the dataset so the merge would have it

landreev:
In my defense, I mentioned in an earlier comment that that was a possibility - that the NativeQuery hack could conflict with the rest of the publish command. 🙂

ekraffmiller:
It is working as expected now from the frontend, but not sure why the integration test is failing.

stevenwinship:
> It is working as expected now from the frontend, but not sure why the integration test is failing.

I'm testing it now locally. Since they are search tests it could be an index timing issue

bjosel:
> I see the difference now.
> 
> I added the file and immediately pressed publish and I see the image If I add the file then go back to the search list and select and publish I don't see the image
> image
> 
> Is it possible that this is a UI issue? The api doesn't know the sequence of events.

Hi all,

we still see the described behaviour on our instance (disapeearing thumbnail in dataset overview page). We are running 6.11 already. I tested on demo.dataverse.org and there the thumbnail doesn't disappear.

PR #12405, opened 2026-05-22
Allow Dataset Creation API to take Template ID

### LINKED ISSUE #12132: Support applying template id in API of dataset creation and persist dataset–template associations
Extend the native API (POST /api/dataverses/{identifier}/datasets) so callers can specify a template, have that template applied to the initial DatasetVersion. If a dataset is created with a template, then the id of the template should be included in the object sent to the createDataset endpoint.

api users

In template management page, delete template action is unclear to users

Let POST /api/dataverses/{identifier}/datasets accept a template (e.g., templateId query param or JSON field)

API changes to allow the closer connection between template management and create dataset

https://github.com/IQSS/dataverse/issues/11789

stevenwinship:
**What this PR does / why we need it**: Users creating Datasets need to be able to set the Template via the Create API

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12132

- Closes #12132

**Additional documentation**:


ekraffmiller:
@stevenwinship to support the SPA, we also need the Get Dataset API to return the template_id, can you add that to this PR?  Or would it be better to have a separate issue?

ekraffmiller:
> @stevenwinship to support the SPA, we also need the Get Dataset API to return the template_id, can you add that to this PR? Or would it be better to have a separate issue?

Update from Steve, this is already in the current PR. thanks!

ekraffmiller:
@stevenwinship I'm testing the SPA and I realized that with this update, we also need to make sure that if a template_id is passed to the create dataset API, the usageCount for that template needs to be incremented.

ekraffmiller:
@stevenwinship this API also needs to apply terms of use and license values from the template, if they exist.  

Edit:  I think we could send the terms of use and license from the template in the API payload, rather than having that logic happen in the backend.  I'm going to look at that more.

stevenwinship:
> @stevenwinship I'm testing the SPA and I realized that with this update, we also need to make sure that if a template_id is passed to the create dataset API, the usageCount for that template needs to be incremented.

Fixed

ekraffmiller:
> @stevenwinship this API also needs to apply terms of use and license values from the template, if they exist.
> 
> Edit: I think we could send the terms of use and license from the template in the API payload, rather than having that logic happen in the backend. I'm going to look at that more.

Confirmed that we can handle this on the client side, no need to update the API

rtreacy (commented):
Claude
PR #12405 Review: Allow Dataset Creation API to take Template ID

What it does

Adds support for "templateId": $templateId in the POST /dataverses/{identifier}/datasets JSON payload. JsonParser.parseDataset() now looks up the template and calls dataset.setTemplate(template); JsonPrinter echoes templateId back in dataset JSON; and CreateNewDatasetCommand's existing (Dataset, DataverseRequest, Template, boolean) constructor is tweaked to fall back to theDataset.getTemplate() when no Template object is passed explicitly — which is how the value set by the JSON parser reaches the command that bumps the template's usage count. JsonParser's constructor gained a TemplateServiceBean dependency, which required updating every call site (production and test) — hence the size of the diff.

Correctness issues

1. Missing dataverse scoping on the template lookup (authorization gap). templateService.find(templateId) (JsonParser.java:292) looks up any template in the system by ID, with no check that it belongs to the target dataverse (owner) or one of its ancestors. Contrast with the UI flow (DatasetPage.java:2247), which only offers templates from dataverseService.find(ownerId).getTemplates(). Via this API, a user with AddDataset permission on dataverse A can supply the numeric ID of a template that lives in an unrelated, possibly private/restricted dataverse B, and:
- dataset.setTemplate(template) persists that cross-dataverse association and is echoed back as templateId in the dataset JSON.
- CreateNewDatasetCommand.postPersist calls ctxt.templates().incrementUsageCount(template.getId()) unconditionally on that foreign template.

   The practical impact is limited — in this API flow getVersionToPersist() still calls the no-arg theDataset.getOrCreateEditVersion(), so the template's field defaults are never applied to the new dataset (unlike the UI path); the effect is essentially an ID-based association plus a usage-count bump on a template the caller may have no visibility into. Still, it's a real scoping gap versus the UI's behavior and worth validating (template.getDataverse() is an ancestor of owner, or reject with 400) before merging.

2. Non-existent templateId fails silently. templateService.find() returns null for an unknown ID (thin wrapper over em.find), and the code does nothing to handle that — dataset.setTemplate(null) is a no-op. Every other invalid-value case in this same method throws a JsonParseException (see the Invalid dataset type: ... branch immediately above this new code, JsonParser.java:288). A typo'd or stale templateId should produce a clear 400 error ("Invalid templateId: " + templateId), not silently create the dataset without the requested template.

3. obj.getInt("templateId", 0) can throw an uncaught ClassCastException. If a caller sends "templateId": "5" (string instead of number), JsonObject.getInt throws ClassCastException, not JsonParseException. The only wrapper at the call site (Dataverses.parseDataset, line ~779) catches JsonParsingException | JsonParseException — a ClassCastException would propagate as an unhandled 500 instead of a clean 400. Worth wrapping the read (or catching/rethrowing as JsonParseException) for consistency with how the rest of this method reports bad input.

Test coverage

- The new DataversesIT test covers the happy path well (create with a valid templateId, verify it round-trips via GET, clean up).
- No test for an invalid/non-existent templateId (would currently reveal issue #2 — silently succeeds instead of erroring).
- No test asserting that a templateId from an unrelated dataverse is rejected (would currently reveal issue #1 — silently succeeds).
- Since neither is currently rejected, existing tests can't catch these gaps; adding negative-path tests here would both document intended behavior and force the scoping/error-handling to be decided explicitly.

Style / minor

- Several files unrelated in substance to the feature (ImportGenericServiceBean.java, ImportServiceBean.java, SchemaDotOrgExporterTest.java, FeedbackUtilTest.java, DataversesIT.java) show large import-block reshuffles/wildcards (import edu.harvard.iq.dataverse.*;, jakarta.ejb.*;, java.util.*;) alongside the one-line change they actually needed (adding a TemplateServiceBean mock/field). This looks like an IDE "organize imports" pass swept up in the commit — harmless (the project's style guide explicitly says wildcard imports are "neither encouraged nor discouraged," and checkstyle's AvoidStarImport is disabled), but it's pure churn that bloats the diff and increases merge-conflict risk for unrelated in-flight PRs touching these same files. Worth trimming to just the needed changes if easy to do, though not blocking.
- Long.valueOf(templateId) where templateId is declared int templateId = obj.getInt(...) is a bit roundabout (auto-boxing an int into a Long via valueOf rather than just widening) — cosmetic only.

Verdict

The core wiring (constructor threading, JSON round-trip) is correct and the happy-path test passes. Before merging, I'd want: (1) a decision on whether templateId should be scoped to the target dataverse's own/inherited templates, since that's the existing UI behavior and its absence is the main risk here, and (2) explicit error handling for an unknown templateId instead of silent success, matching the pattern used for every other invalid field in this method.

rtreacy:
@stevenwinship can you take a look at  the Correctness issues, Test coverage and Style / motor sections of the Claude review above?

stevenwinship:
Made fixes and added tests based on Claude

stevenwinship:
@rtreacy I made changes based on the Claude review. Please re-review

rtreacy (approved):
Previous review comments from Claude have been addressed

ChengShi-1:
Hi @stevenwinship , could you resolve the conflicts?

qqmyers:
@ChengShi-1 - I resolved the merge conflicts.

doc/release-notes/12132-dataset-creation-with-template.md
```diff
@@ -0,0 +1,3 @@
+## Feature ##
+
+POST /dataverses/{identifier}/datasets payload JSON now allows for "templateId":$templateId to be included to set the Template on the new Dataset.
```
pdurbin, line 3: @stevenwinship this should have been documented in the API Guide. I just created an issue for this:

- #12677

Can you please make a PR to add those docs?

PR #12109, opened 2026-01-22
S3 Store Cleanup/ Leak close

qqmyers:
**What this PR does / why we need it**: While working on archiving, I noticed occasional warnings related to the netty library (see #12108). In looking into that w/AI, I saw that a) we were closing the s3Presigner after use while also caching it for reuse, b) that we were not closing the presigners and clients at shutdown, and c) were not using the latest aws sdk. 

This PR fixes all of these but, as far as I can tell, does not resolve #12108 itself. Nevertheless, it seems like these are useful updates to have. - perhaps #12593 resolves it as that uses an even newer version of netty.

**Which issue(s) this PR closes**:

- Closes #

**Special notes for your reviewer**:

**Suggestions on how to test this**: regression/auto-tests only. Check for the "Closing all S3 clients and transfer managers." log when shutting Dataverse/Payara down.

**Does this PR introduce a user interface change? If mockups are available, please link/include them here**:

**Is there a release notes update needed for this change?**:

**Additional documentation**:

Issue #12036, opened 2025-12-11
Remove/Discourage using MinIO as a S3 Storage Option


poikilotherm:
With Minio move to no longer support an Open Source model, we should encourage people to a) not use it for any new deployments (if they don't want to pay for a license) and b) move away from it as they no longer will provide maintenance (if they cannot afford a paid license).

See also https://bizety.com/2025/12/06/minio-in-maintenance-mode-open-source-alternatives for details on what changed.

Necessary changes:

- Remove from docs/guides (maybe keep a note that it used to work as an S3 target)
- Highlight current tools usable to build an S3 compatible storage (maybe include paid and free options?)
- Remove and/or replace the MinIO container from/in docker compose
- Update dataverse-ansible to not use MinIO by default
  - https://github.com/gdcc/dataverse-ansible/issues/412

pdurbin:
As discussed in [Slack](https://iqss.slack.com/archives/C010LA04BCG/p1765450337247799), I agree with @poikilotherm that for dev stuff we should probably try [Garage](https://github.com/deuxfleurs-org/garage) or [RustFS](https://github.com/rustfs/rustfs).

As a side note, in [Jenkins](https://github.com/gdcc/dataverse-ansible/blob/b3c31b1c780544bdf2cf314980af48a6137eab55/tests/group_vars/jenkins.yml#L311) and [locally](https://github.com/IQSS/dataverse/blob/v6.9/docker-compose-dev.yml#L36) we also use LocalStack. My original plan was to only use MinIO but I never got direct upload working so I [added](https://github.com/IQSS/dataverse/blob/v6.9/src/test/java/edu/harvard/iq/dataverse/api/S3AccessIT.java#L242) LocalStack to test that.

srmanda-cs:
If it's the direction that y'all want to go in, I can work on porting over the current MinIO tests over to Localstack. Once that's done, we can re-evaluate and move to Garage or something as needed. Would that be a desired outcome?

pdurbin:
@srmanda-cs thanks for offering! I looked at [old notes](https://dataverse.zulipchat.com/#narrow/channel/375812-containers/topic/testing.20S3/near/398340496) and I don't have a strong preference for what we use in dev/testing. If you want to try removing MinIO from the tests, please feel free to try! And yes, perhaps in the future we can consider other moves.

@poikilotherm do you have time to help with the docs part you outlined above? Once we have a branch started, of course.

srmanda-cs:
[...]

PR #12413, opened 2026-05-29
Disable s3 multipart download option

### LINKED ISSUE #12660: multipartEnabled breaks downloading/publishing large datafiles uploaded from webui to s3 bucket
We are new to Dataverse. And some related configurations:

- using OSN (open storage network) s3 bucket as our backend storage
- direct upload is not supported currently as the CORS policy was not implemented on the bucket from OSN

Recently, we upgraded to 6.11 from 6.10.1.
And noticed two issues:

1. One thing we noticed was that, when trying to publishing a dataset which includes a file ~160MB, we got the error message: 
Image

Once the file is removed from the dataset, publishing was fine.

2. Another issue was with direct-download disabled, when downloading this file (from the draft since it failed to publish ), the downloaded file is truncated at ~8MB. However, with direct-download enabled, the downloaded file is complete.

On our test dataverse instance, with the same S3 backend storage (Open Storage Network) and upload-redirect=false:
1. Upload a ~160 MB file to a draft dataset through the web UI. Dataverse writes it to S3 as a multipart object — ETag ends in -20, PartsCount: 20, part 1 is ~ 8MB

2.  Publish the dataset. Publishing fails, the dataset is locked with FileValidationFailed and and /api/admin/validate/dataset/files/{id} returns:
```
Checksum Validation Unexpected Exception:
  software.amazon.awssdk.services.s3.model.S3Exception:  (Service: S3, Status Code: 412,
  Request ID: tx00000dc10857e1119c449-…) (SDK Attempt Count: 1)
```
3. Outside Dataverse, download that file and re-upload the identical bytes to the same key with a single PUT (aws s3api put-object), so the object is no longer multipart. Nothing about the file changes — the MD5 is identical before and after, and it still matches datafile.checksumvalue.
4. Publish again. It succeeds.

The only variable between steps 2 and 4 is whether the file carries multipart part boundaries. And it looks like the reason it fails is that the ETag returned for a partNumber request is the part's ETag, not the object's. The SDK carries the first part's ETag forward as If-Match on the next part, and it never matches. Reproduced by hand against the storage, with no Dataverse involved:

```
$ aws s3api head-object --bucket B --key K --query ETag --output text
"44c9c29827fb957f4e20c100604564dc-20"                     # object

$ aws s3api get-object --bucket B --key K --part-number 1 /dev/null \
    --query '[ContentLength,PartsCount,ETag]' --output text
8388608   20   "21253d656a4ae7fcd84024696eefb724"         # part 1

$ aws s3api head-object --bucket B --key K --part-number 2 \
    --if-match '"21253d656a4ae7fcd84024696eefb724"'
An error occurred (412) when calling the HeadObject operation: Precondition Failed
```

Additionally, datasets published before we upgraded to 6.11 contain files of similar size that are single-part objects. For example, a 116 MB file uploaded in April returns:
```
$ aws s3api head-object --bucket B --key K --query '[ETag,ContentLength]' --output text
"dd57f7377b99dfaed6926f42b9f80e74"   116061051

$ aws s3api get-object --bucket B --key K --part-number 1 /dev/null \
    --query '[ContentLength,PartsCount]' --output text
116061051   None
```

Possible cause:

And one thing we noticed changed from the v6.10 to v6.11 was the addition of the following in the file S3AccessIO.java:
```
// Always enable multipart upload. It is only used when necessary
            s3CB.multipartEnabled(true);
```

And the [doc](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/s3/S3AsyncClientBuilder.html#multipartEnabled(java.lang.Boolean)) mentioned that
> Enables automatic conversion of GET, PUT and COPY methods to their equivalent multipart operation.
So enabling it appears to have enabled multipart download as well, which matches the truncated file we experienced.

We are in the process of getting CORS enabled to do direct uploading, it may be able to prevent this error.

* When does this issue occur?
Publishing datasets
Downloading datasets

Dataset draft publishing page

* What happens?
This seems to have something to do with the multipart upload/download

curators, or all users if downloading

v6.11

**Screenshots:**

* 

*

janvanmansum:
### Cover Message for Pull Request

This PR addresses a `412 Precondition Failed` error encountered when downloading certain large objects from S3-compatible storage (specifically Ceph) that were originally uploaded using multipart upload (e.g., via the `boto3` library with default 8MiB part sizes).

The issue arises because the AWS SDK for Java, when multipart support is enabled, automatically attempts a parallel multipart download if it detects an ETag with a part-count suffix (e.g., `-5`). During this process, the client sends an `If-Match` header containing the ETag of an individual part to ensure consistency. However, some servers like Ceph expect the ETag for the entire object, leading to the `412` error.

To resolve this, the `S3AccessIO` class has been refactored to use two distinct S3 clients:
1.  **A Write Client**: Has multipart enabled to support efficient uploads (fixing issues with large file uploads).
2.  **A Read Client**: Has multipart disabled by default. This forces the server to handle the reassembly of parts and prevents the client from sending part-specific `If-Match` headers that trigger the 412 error.

Additionally, a new configuration option `disable-multipart-download-for-indirect-download` has been introduced to control this behavior.

- Closes #12660

The fix involves spliting the S3 client into `s3ReadClient` and `s3WriteClient`. The `s3ReadClient` is configured with `.multipartEnabled(false)` if the new configuration setting is set to `true`. By default `s3ReadClient` and `s3WriteClient` refer to the same client, which has multipart enabled.

1.  **Reproduction**: Attempt to download a file from a Ceph-backed S3 store that was uploaded in multiple parts (verify the ETag has a suffix like `-5`). Confirm it fails with `412 Precondition Failed` without this PR. (JM) Set `dataverse.files.<id>.download-redirect=false` so that Dataverse is involved in the download.
2.  **Verification**: Apply the PR and verify the download succeeds.
3.  **Regression**: Verify that regular file uploads and downloads (non-multipart) still work as expected.
4.  **Configuration**: Test setting `dataverse.files.<id>.disable-multipart-download-for-indirect-download` to `false` and verify the `412` error returns (if testing against a problematic server).

No.

https://github.com/IQSS/dataverse/pull/12413/changes#diff-827dccc047f09653dade26fe2dce6be6d20ae6a739448e2b9f2890849b64d2fd

**Additional documentation**:

Preview docs at https://dataverse-guide--12413.org.readthedocs.build/en/12413/installation/config.html#reported-working-s3-compatible-storage


doc/release-notes/12413-disable-multipart-download-for-indirect-download.md
```diff
@@ -0,0 +1,5 @@
+A new configuration setting has been introduced for S3 compatible storage drivers that addresses an incompatibility between the AWS S3 library used in Dataverse and certain S3 implementations such as the Ceph Object Gateway:
```
qqmyers, line 1: @janvanmansum - I made some edits - feel free to revert. I wanted to make it clear this new setting is mainly to address a problem, not something you'd want to set true otherwise. I also guessed that you're talking about the Ceph Object Gateway.

Beyond that, I might suggest adding Ceph to the list at https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/installation/config.rst#reported-working-s3-compatible-storage with a note about this setting.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -104,14 +105,15 @@ public class S3AccessIO<T extends DvObject> extends StorageIO<T> {
...
-    boolean s3pathStyleAccess = false;
 
-    private static HashMap<String, S3AsyncClient> driverClientMap = new HashMap<String, S3AsyncClient>();
-    private static HashMap<String, S3Presigner> driverPresignerMap = new HashMap<String, S3Presigner>();
-    private static HashMap<String, AwsCredentialsProvider> driverCredentialsProviderMap = new HashMap<String, AwsCredentialsProvider>();
-    private static HashMap<String, S3TransferManager> driverTMMap = new HashMap<String, S3TransferManager>();
+    private static final ConcurrentHashMap<String, S3AsyncClient> driverDownloadClientMap = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, S3AsyncClient> driverUploadClientMap = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, S3Presigner> driverPresignerMap = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, AwsCredentialsProvider> driverCredentialsProviderMap = new ConcurrentHashMap<>();
```
qqmyers, line 116: Just noting that this PR has some general improvements ~unrelated to the issue, such as the use of ConcurrentHashMaps.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -305,18 +312,19 @@ public InputStream getInputStream() throws IOException {
...
                 setInputStream(responseInputStream);
-            } catch (InterruptedException | ExecutionException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
             }
-        }
-        if (super.getInputStream() == null) {
-            throw new IOException("Cannot get InputStream for S3 Object" + key);
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
qqmyers, line 321: Resetting the interrupt is a recommended practice, not related to the issue.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1014,8 +1079,6 @@ public String generateTemporaryDownloadUrl(String auxiliaryTag, String auxiliary
             } catch (S3Exception e) {
                 logger.warning("Exception generating temporary S3 url for " + key + " (" + e.getMessage() + ")");
                 return null;
-            } finally {
-                s3Presigner.close();
             }
```
qqmyers, line 1081: Another general fix. FWIW: Not closing the presigners is one of the fixes in #12109 which has been sitting. We probably want the other fixes there too.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -1189,38 +1248,49 @@ private static long getMinPartSize(String driverId) {
     }
 
     private static S3TransferManager getTransferManager(String driverId) {
-        if (driverTMMap.containsKey(driverId)) {
-            return driverTMMap.get(driverId);
-        } else {
+        return driverTMMap.computeIfAbsent(driverId, id -> {
```
qqmyers, line 1250: Another general improvement

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -725,17 +768,21 @@ public List<String> listAuxObjects() throws IOException {
...
                 } else {
                     nextContinuationToken = null;
                 }
             }
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IOException("S3AccessIO: Failed to get aux objects for listing.", e);
+        }
+        catch (ExecutionException e) {
```
qqmyers, line 784: Minor - this PR seems to be inconsistent in putting the catch on the line with the closing } or the next line and changes whether else is on a new line or not. It would be good to have things consistent (I think we generally have catch and else on the same line as the } )

qqmyers (changes requested):
The basic change here - to provide the option to disable use of multipart transfer for download - is necessary for DANS to use the code with Ceph, (with download-redirect false). The implementation is straight forward - the code is refactored to have separate read/write client variables, but these still point to one client unless the new setting is true, in which case download uses a separate aws s3 client with multipart disabled.

The other general changes (noted in other comments) all see reasonable as well, though I think #12109 should make the same release.

The two requested changes are just to add Ceph in the list of known compatible instances and (minor) to clean up some of the formatting/reformatting that happened.

For QA, I'll note that DANS is using this already so I think the key thing is to regression test - not sure the core team needs to configure Ceph to test the actual fix.

janvanmansum:
@qqmyers 
> The two requested changes are just to add Ceph in the list of known compatible instances and (minor) to clean up some of the formatting/reformatting that happened.

* I guess you addressed adding Ceph to the list of known compatible instances yourself - or do you mean something else than mentioning it in the release note?
* About the formatting, I actually spent a lot of time trying to get it right. Is there a formatting policy that you can point me to so I can look for an automated tool to do this? Restoring the code formatting to its (frankly already mostly messy) former state is turning out to be the major headache with every pull request, when it should actually be something that is easily done automatically. (I even tried to make my AI do it, but it spent so many coins figuring it out, that I aborted the attempt.)

qqmyers:
@janvanmansum - the formatting guidance is at https://guides.dataverse.org/en/latest/developers/coding-style.html#java (as usual feel free to improve, add config for your ide, etc.). In general formatting varies and no one is too strict, but the if/else and try/catch using new lines caught my eye here as not being common. I agree it's frustrating trying to match the non-standard formatting for whatever file you edit.

As for documentation, I was suggesting adding Ceph to the config.rst guide at the link in the comment - https://github.com/IQSS/dataverse/blob/develop/doc/sphinx-guides/source/installation/config.rst#reported-working-s3-compatible-storage. I did change the release note, but did not add it in the config guide (are there any other s3 options needed to make Ceph work that should be listed there?).

qqmyers (approved):
Looks good.

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1461,6 +1462,10 @@ You may provide the values for these via any `supported MicroProfile Config API
 Reported Working S3-Compatible Storage
 ######################################
 
+`Ceph Object Gateway <https://docs.ceph.com/en/reef/radosgw/#ceph-object-gateway>`_ (added July 2026/Dataverse v6.12)
+Set ``dataverse.files.<id>.disable-multipart-download-for-indirect-download=true`` if not using direct download. 
+(This forces the S3 server to handle part reassembly and avoid incompatible headers that cause `412` errors from the Ceph Gateway.)
+
```
pdurbin, line 1486: Just a heads up that there was a merge conflict here that I resolved in 9668f11f25beb876d945db43dc3af36c51a3f42d

The conflict was because of this PR we merged:

- #12036

doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1417,28 +1417,29 @@ List of S3 Storage Options
 .. table::
     :align: left
```
pdurbin, line 1419: I'm just reporting this here, randomly, so I can resolve it later.

On my laptop I configured opened the dropdown for storage...

Image

... and picked LocalStack:

Image

When I tried to upload a file I saw this in the logs:

`dev_localstack> 2026-08-28T18:59:51.195  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.HeadObject => 404 (NoSuchKey)`

And in the UI the file I was uploading flashed and then disappeared:

Image

In CI, API test are passing with LocalStack so I'm not sure what's going on. I must admit that locally, on my laptop, I never change it from the default! 😅

pdurbin: I just tried again on develop (82e84a1080) and got the same error. So maybe something's wrong with LocalStack in our dev env. 🤔

pdurbin: On develop I switched to "LocalStackNoDirect" and I was able to upload the file.

Screenshot 2026-08-28 at 3 34 42 PM

It was stored here:

`"storageIdentifier": "localstack_noredirect://mybucket-noredirect:1a049dd7a79-b009fc090014",`

From our dev compose file, here's the config:

        -Ddataverse.files.localstack_noredirect.label=LocalStackNoRedirect
        -Ddataverse.files.localstack_noredirect.custom-endpoint-url=http://localstack:4566
        -Ddataverse.files.localstack_noredirect.custom-endpoint-region=us-east-2
        -Ddataverse.files.localstack_noredirect.bucket-name=mybucket-noredirect
        -Ddataverse.files.localstack_noredirect.path-style-access=true
        -Ddataverse.files.localstack_noredirect.upload-redirect=false
        -Ddataverse.files.localstack_noredirect.download-redirect=false
        -Ddataverse.files.localstack_noredirect.access-key=default
        -Ddataverse.files.localstack_noredirect.secret-key=default

qqmyers: This PR should only affect download, so hopefully unrelated. The one thing I can think of that may be relevant - we do a HEAD to get the size of files being direct uploaded and I believe we saw issues before with 'eventual consistency' - the file had uploaded successfully but HEAD returned 404. We may have added a retry or delay for that - I don't recall without digging, but if LocalStack can't immediately find the file (waiting for some indexing etc.) it might look like what you're seeing.

pdurbin: Ok, thanks. I threw this back in "ready for QA" and unassigned myself.

Per [Slack](https://iqss.slack.com/archives/C010LA04BCG/p1787944407979189), it sounds like we should probably update internal or the QA server to the newest Payara and test S3 stuff there.

Additional S3 PRs we want to get in 6.12:

- Must have: https://github.com/IQSS/dataverse/pull/12109
- Could have: https://github.com/IQSS/dataverse/pull/12454

pdurbin:
@janvanmansum (or anyone who knows) does this PR address either of the issues below?

- #12649
- #12660

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -18,6 +18,7 @@
 import software.amazon.awssdk.services.s3.S3AsyncClientBuilder;
```
pdurbin, line 18: @janvanmansum or @qqmyers can you please resolve the merge conflicts?

PR #12418, opened 2026-05-29
Removing deprecated Meta class

### LINKED ISSUE #12417: Remove Meta Class
Meta class is deprecated and no longer needed. It should be removed

stevenwinship:
**What this PR does / why we need it**: Meta class is not needed

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12417

- Closes #12417

**Suggestions on how to test this**: call GET /api/meta/datafile/13962692

**Additional documentation**:


landreev (commented):
Looks good. Could you please also remove this method with the path hard-coded:
https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L4884-L4889
Aside from that, will approve once Jenkins finishes; probably first thing next week.

landreev (approved):

PR #12420, opened 2026-06-01
Suppress hidden datasetContactEmail field when using CVoc

qqmyers:
**What this PR does / why we need it**: #10503 added the raw values of CVoc managed fields as hidden DOM elements to allow CVoc scripts to use those values in constructing the display. This is needed for some use cases (see #10503 and related issue), harmless in most, but a problem for one use case - the datasetContactEmail field. Because it (and only it right now as far as I can see) is give the `#EMAIL` format, the display value is set to `''` (actually to the value of the `dataset.email.hiddenMessage` string which is `''`) when the field is not managed by CVoc. When it is managed (i.e. to allow an ORCID or ROR entry for the contact), #10503 added the real value (i.e. the private email) as a hidden field, which means it is accessible via the page source. This PR suppresses adding a hidden field if the format is `#EMAIL`.

- Closes #

**Special notes for your reviewer**: One could argue that the :ExcludeEmailFromExport setting should govern the display for this field instead of the one-off #EMAIL format, but that's a bigger change.

**Suggestions on how to test this**: Configure CVoc on the contact field with the email as a managed field and check the page source to see if the email is a hidden field or not.

**Is there a release notes update needed for this change?**: probably not?

**Additional documentation**:


sekmiller (approved):
As stated this approach will work, and the larger change to referring to the flag can be avoided for the JSF UI.

pdurbin:
For the JSF version of Harvard Dataverse, we don't plan to enable CVoc for the contact field.

jggautier:
@pdurbin I updated [your comment](https://github.com/IQSS/dataverse/pull/12420#issuecomment-5357835680) to clarify that we don't plan to enable CVoc for the contact field for the JSF version of Harvard Dataverse

pdurbin:
@qqmyers helped me understand the bug. On the develop branch, I can see my email address in the DOM:

Screenshot 2026-09-03 at 10 55 13 AM

pdurbin:
The PR seems like a good fix to me. I can't find the email in the DOM anymore:

Screenshot 2026-09-03 at 12 01 40 PM

PR #12421, opened 2026-06-02
Retention period edge case

jo-pol:
**Why we need this PR**:

When you select a retention period the current date + one year is offered as an option but not allowed. The error message however says it is allowed.
image

**What this PR does**:

- java validation subtracts one day from the minimal date to turn `isBefore(...)` into `<= ...`
- end date: no longer checked nor reported

- Closes #

WEB-UI:
* Add a file to a draft dataset, save the changes but do not publish.
* Select _retention period_ from _edit files_.
* Pick edge cases from the calendar, type invalid edge case manually.

API: 
* try edge cases with https://guides.dataverse.org/en/latest/api/native-api.html#set-a-retention-period-on-files-in-a-dataset

Messages will now look like:
image

With a date over a thousand years in the future, you will no longer get the web-ui message:
```datasetForm:fileRetentionDate: Validation Error: Value must be between 2027-06-01 and 3026-06-01.```

**Additional documentation**:


stevenwinship (approved):
Tested with and without this PR. Looks good and fixed the problem

qqmyers:
@jo-pol - In doing QA on this PR, I noticed a couple ~related issues: 
- Due to https://github.com/IQSS/dataverse/blob/0b87f4b631a829f325581221bc47afa8cb3d6747/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L2307 and https://github.com/IQSS/dataverse/blob/0b87f4b631a829f325581221bc47afa8cb3d6747/src/main/java/edu/harvard/iq/dataverse/FilePage.java#L296, the initial choice for retention period is the day after the minimum - seems like plusDays(1) isn't needed if the minimum now has a minusDays(1).

- If I cancel the retention dialog and reopen it, the default date shown is 1000 years in the future. I think that's due to creation of a new Retention at https://github.com/IQSS/dataverse/blob/0b87f4b631a829f325581221bc47afa8cb3d6747/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L6849 (which defaults to 1000 years in the constructor), without code like this https://github.com/IQSS/dataverse/blob/0b87f4b631a829f325581221bc47afa8cb3d6747/src/main/java/edu/harvard/iq/dataverse/DatasetPage.java#L2305-L2308 which is done at page init but not in the clearRetentionPopup method. (Similar issue with the FilePage.)

As these are all retention date related and relatively simple, would you like to fix them in this PR, or do you want to treat them as a separate issue/possible PR for the future?

jo-pol:
@qqmyers 

> treat them as a separate issue/possible PR for the future

This PR is already waiting too long

PR #12434, opened 2026-06-04
NPE for preview of draft file with retention period

### LINKED ISSUE #12433: Preview of draft file with retention period causes internal server error

Add a file to a dataset, set a retention period (with or without a reason), do not publish, try to preview the file with a retention period.

* When does this issue occur?

When trying to preview a draft file with a retention period

files

* What happens?

Internal server error:` java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "str" is null`
Top of the stack trace:

    at java.lang.String.concat(String.java:2771)
    at jdk.internal.reflect.GeneratedMethodAccessor374.invoke(Unknown Source:-1)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:569)
    at jakarta.el.ELUtil.invokeMethod(ELUtil.java:215)
    at jakarta.el.BeanELResolver.invoke(BeanELResolver.java:487)
    at jakarta.el.CompositeELResolver.invoke(CompositeELResolver.java:198)
    at org.glassfish.expressly.parser.AstValue.getValue(AstValue.java:298)
    at org.glassfish.expressly.parser.AstValue.getValue(AstValue.java:144)
    at org.glassfish.expressly.parser.AstChoice.getValue(AstChoice.java:44)
    at org.glassfish.expressly.ValueExpressionImpl.getValue(ValueExpressionImpl.java:138)
        this: "ValueExpression[#{(empty FilePage.file.embargo) and !(empty FilePage.file.released) and !(empty FilePage.file.retention) ? ''.concat(FilePage.file.publicationDateFormattedYYYYMMDD).concat('; ') : ''}]"
    at org.jboss.weld.module.web.el.WeldValueExpression.getValue(WeldValueExpression.java:50)

https://github.com/IQSS/dataverse/blob/ac8bdc17020bd87193b3d4fe0431a6804fbf2e6d/src/main/webapp/file.xhtml#L569-L576

Curators and dataset owners 

v6.9

#12432  applies to embargo and retention. This issue applies only to the retention period.
 

**Screenshots:**

*

jo-pol:

- Closes #12433

The change was suggested by copilot, it fixes the reported problem but I'm pretty clueless.

See issue

**Additional documentation**:


src/main/webapp/file.xhtml
```diff
@@ -566,7 +566,8 @@
                                                         <th scope="row">
                                                             #{bundle['file.metadataTab.fileMetadata.publicationDate.label']}
                                                         </th>
-                                                        <td>#{(empty FilePage.file.embargo) and !(empty FilePage.file.released) and !(empty FilePage.file.retention) ? ''.concat(FilePage.file.publicationDateFormattedYYYYMMDD).concat('; ') : ''}
+                                                        <td>#{(empty FilePage.file.embargo) and FilePage.file.released and !(empty FilePage.file.retention) ? FilePage.file.publicationDateFormattedYYYYMMDD : ''}
```
qqmyers: I think the root cause is not addressed by this fix, although I can see why it would work. Since FilePage.file.released is a boolean `!(empty FilePage.file.released)` is always true (false is not empty). So the first branch with the concats runs when the publication date is null. I think if you replace that with `FilePage.file.released` it should work without needing to split adding the ; into a separate step. (Unless we have a bug where the released is true and the pub date is still null. If so, testing if the pub date is null might be a better choice.)

jo-pol: Also noticed repeating the tag 'draft' when both an embargo and retention is set, on the dataset page as well as the file page. I suppose the second should become a ';'. At the moment I have no clue about the origin of this tag and it might not be a use case.
image
image

qqmyers: I'm sure the Draft: is being added to the beginning of separate messages for the embargo and retain logic. If this is common and confusing enough, code could probably be added to suppress Draft: in the retained message if the file is also embargoed. Probably a separate issue/PR though. I'll go ahead and approve this.

qqmyers (approved):
@jo-pol - I think this is good to go, except it should probably have a release note that says the bug is fixed. Can you add that? I'll approve so it can get into the ready for QA queue.

Issue #12001, opened 2025-11-24
Feature Request: API  to support Download Terms of Use


ekraffmiller:
**Overview of the Feature Request**
To support production file download in the SPA, we need to add a form for the user to accept terms of use & sign the guestbook, if required (replicating JSF UI).

API requirements:

1. In Get Dataset endpoint, update the response to also return the Guestbook for the dataset, if it exists.
2. Update download API endpoint: Save Guestbook Response - saves the guestbook data entered by the user at time of download.

Download Terms of Use example:

<img width="3468" height="2842" alt="Image" src="https://github.com/user-attachments/assets/5eb5f2ba-d43a-4659-86c1-4d24f9088a9d" />

Form for creating a Dataset Guestbook:

<img width="1093" height="581" alt="Image" src="https://github.com/user-attachments/assets/c1531e89-c0bd-44e0-888d-ae9b5165318c" />

qqmyers:
FWIW: It might be better and easier to extend the request access and download APIs to accept guestbook responses (versus a new save guestbook response endpoint) . On the back-end the response is linked to the access of download request, so sending the guestbook response separately means additional tracking.

pdurbin:
We should merge this first:

- #11844

pdurbin:
Be aware of the guestbook-at-request setting.

PR #12435, opened 2026-06-04
Fix URL signing api for URLs containing special characters

ErykKul:
Bug fix; URL signing for URLs containing special characters broken in 6.10, this PR fixes the introduced bug.

Issue not created, bug fixed directly here.

Added tests to prevent regression. Also, API requires now the signing secret to be set, otherwise it will not work.

No special testing needed, the unit tests should be sufficient.

No

Yes, included.

**Additional documentation**:
No


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -3349,9 +3349,10 @@ are time limited and only allow the action of the API call in the URL. See :ref:
 :ref:`api-native-signed-url` for more details. 
 
 The key used to sign a URL is created from the API token of the creating user plus a signing-secret provided by an administrator.
-**Using a signing-secret is highly recommended.** This setting defaults to an empty string. Using a non-empty 
-signing-secret makes it impossible for someone who knows an API token from forging signed URLs and provides extra security by 
-making the overall signing key longer.
+**A non-empty signing-secret is required to request signed URLs through the API.** If it is not configured, the
+``/api/admin/requestSignedUrl`` endpoint (see :ref:`api-native-signed-url`) returns an error instead of issuing a
+weakly-signed URL. (The setting otherwise defaults to an empty string.) A non-empty signing-secret makes it impossible for
+someone who only knows an API token to forge signed URLs, and provides extra security by making the overall signing key longer.
```
stevenwinship: You may want to add to the changelog.rst that pre-existing signed urls that haven't expired will no longer work once the signing-secret is added.

ErykKul: Added this to the release note (doc/release-notes/fix-url-signing-special-characters.md). The upgrade note now calls out that setting (or changing) the signing secret invalidates any previously issued, not-yet-expired signed URLs, so clients/integrations have to request new ones.

pdurbin:
@ErykKul sorry, we bumped this to 6.12 given our capacity and how you said you're probably the only person using this API. 😄

src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -2453,7 +2453,13 @@ public Response getSignedUrl(@Context ContainerRequestContext crc, JsonObject ur
...
             return error(Response.Status.FORBIDDEN, "Requesting signed URLs is restricted to superusers.");
         }
-        
+
+        // Require a signing secret: without it the key is only the user's API token, which is too weak.
+        if (JvmSettings.API_SIGNING_SECRET.lookupOptional().orElse("").isEmpty()) {
+            return error(Response.Status.INTERNAL_SERVER_ERROR,
+                    "Requesting signed URLs requires a signing secret to be configured. Please set the dataverse.api.signing-secret JVM option.");
+        }
+
```
stevenwinship, line 2745: My concern with this being here is that it doesn't cover the URL signing when requesting a file download with a guestbook response. Those APIs still work without the secret. (See Access.java) These APIs all call UrlSignerUtil.signUrl, which should have this code to return an error if no signing-secret exists.

{
    "status": "OK",
    "data": {
        "signedUrl": "http://localhost:8080/api/v1/access/dataset/4?gbrecs=true&gbrids=12&until=2026-06-15T20:36:32.302&user=usere7c863fd&method=GET&token=c99c3f9393be5ddedbd72829b6a8b29de3310b06f6692b77286351ba079c82af177c0e8c8df237afe7d6611b7c044e79cc7e402042bf07aa1f9cbbb9cf855298"
    }
}

ErykKul: Good catch. Two commits: first I added the same guard to the guestbook download (`Access.returnSignedUrl`), then I centralized it in `UrlSignerUtil` so every API-token-based signer requires the secret.

I kept it out of `signUrl()` itself because the remote/Globus overlay stores sign with their own per-store key and must keep working without the global one.

Heads up: centralizing means *all* such signing now needs the secret: external tool/Globus callbacks and permission-history links too, not just the two endpoints. Every install using them must set `dataverse.api.signing-secret`.

That may be too much; happy to scope it back to just `requestSignedUrl` + the guestbook download if you prefer.

stevenwinship (approved):

pdurbin:
@ErykKul I'm putting this on hold for now since I heard you're on vacation until Aug 4. Please resolve merge conflicts when you get back. Thanks!

ErykKul:
@pdurbin, I did the merge. This one was easy. If there is something bigger, then I will pick it up after Aug 4.

src/main/java/edu/harvard/iq/dataverse/util/UrlSignerUtil.java
```diff
@@ -45,19 +44,12 @@ public class UrlSignerUtil {
...
-            } catch (URISyntaxException e) {
-                logger.severe("Invalid URL for signing: " + baseUrl + " " + e.getMessage());
-            }
-        }
+        // Strip reserved signing params that may already be in the base URL, using exact-string
+        // surgery rather than URIBuilder. The URL must be signed exactly as provided (the pre-6.10
+        // behavior): validation reconstructs the signing string from the URL-decoded request, so
+        // re-encoding here (e.g. percent-encoding ':' and '/' in DOIs) would change the signed bytes
+        // and the signature would no longer match.
+        baseUrl = stripReservedParameters(baseUrl);
```
qqmyers: Thanks for working with the existing code. FWIW: I think it would be cleaner overall for this general utility to not try fixing a caller's mistakes - perhaps it should throw an exception if one of the four params needed by the algorithm are sent (and never touch the url on the good path and thus not needing to strip anything). Similarly flagging/removing  signed and key here because the caller doesn't strip them seems backwards.

qqmyers:
@ErykKul - are you around to be able to fix the minor merge conflict? (If not, I can fix it and go forward with QA/merge). Also - do I understand correctly there is no visible effect in 6.10 - the URLs it creates internally don't have characters that are affected by the bug, so the only thing to QA is regression testing and trying the create signed url api with something like a doi (with : and / ).

src/main/java/edu/harvard/iq/dataverse/util/UrlSignerUtil.java
ErykKul: This wasn't introduced in this PR. It came in with the 6.10 guestbook changes (#12001), where the `?signed=true` flow signs the incoming request URI, which can carry `key=<apitoken>` and `signed=true`. But agreed it's worth picking up here: I'll make signUrl throw if any of the four algorithm params are present and never touch the URL otherwise, and move the (byte-exact) stripping helper out of signUrl to the callers that build URLs from incoming requests.

ErykKul: Done in 2e0749fcc4: `signUrl` now throws an `IllegalArgumentException` if the base URL already contains one of the four signing params and never touches the URL otherwise. The byte-exact stripping helper moved out of `signUrl` to the `?signed=true` flow in Access (the caller that actually receives those params in the request URI), and `/api/admin/requestSignedUrl` now returns a 400 naming the offending parameter instead of silently rewriting the supplied URL. Release note, docs and tests updated accordingly.

ErykKul:
The conflict should be gone already, I merged develop this morning before pushing (GitHub shows the branch as mergeable now).

And yes, that's my understanding too: the URLs Dataverse creates internally only contain numeric ids, so the only visible breakage in 6.10 is for `requestSignedUrl` with special characters in the url, like a doi (that's how rdm-integration ran into it). Besides that and regression testing, two things from this PR worth a quick try in QA: `requestSignedUrl` now returns 400 when the url already contains a reserved param, and the signing endpoints now error without `dataverse.api.signing-secret` instead of falling back to a weak key.

ErykKul:
@qqmyers Heads up before QA: a few more commits landed since my comments above. While testing against a real client I also made validation verbatim-first — the signature is checked against the raw request URI first, with the URL-decoded check kept as fallback — so a URL signed exactly as it will be used now works as received, and everything that validated before still validates. A final review pass added a couple of fixes on top: crafted `user=!...` values get a proper 401 instead of a 500, and tool-manifest urlTemplates get reserved params stripped before signing again like 6.10 did (otherwise `key={apiToken}` in a template would sign the user's real API token into the URL). All with unit tests; docs and release note updated. That's everything from my side.

qqmyers:
@ErykKul - can we coordinate in slack and discuss this PR?

pdurbin:
@ErykKul just a reminder that we'd like your input and code freeze for 6.12 is next Thursday! 🥶

pdurbin:
We think only @ErykKul is actively using this but we'd like to get it into 6.12.

ErykKul:
@pdurbin  I am on it. Trying to wrap it up. Not ready yet, I will try to have it ready by tomorrow evening.

ErykKul:
@qqmyers Ready for another look. The signing/validation fix itself is unchanged since your last pass; the rework around it, per our discussion:

Signing secret is now a DB setting (:ApiSigningSecret), auto-generated on first use. I dropped the JVM option entirely: there's no good way to generate a secret and keep it synced across the deployment shapes people actually run (docker compose, Kubernetes, hosted Payara). Admins can still store their own; a minimum length of 32 is enforced on every write path (single PUT, localized, bulk). DELETE rotates it. Upgrading invalidates outstanding signed URLs (release note covers it). With the secret always present, all the "no secret configured" fallback paths are gone.

Reserved query parameters are now rejected, never silently stripped — requestSignedUrl and tool-manifest allowedApiCalls templates return a 400 naming the parameter (note: a pre-existing manifest that misused one worked silently in 6.10 and will 400 at launch after upgrade).

One behavior change to ratify: requestSignedUrl with an explicit but unknown user now returns 400 instead of silently signing with the calling superuser's token (a typo used to mint a full-privilege URL). Omitting user still defaults to the caller, as already documented: I wondered whether that default should be more explicit (or required), but that felt out of scope for this PR.

Also while in there: constant-time signature comparison, and the secret is redacted from the action log.

PR #12436, opened 2026-06-04
Adding filtering and pagination to collectionList APIs

### LINKED ISSUE #12423: Update collectionList API: add pagination and filtering
Update mydata/retrieve/collectionList to take parameters for pagination and filtering by collection name string.  Needed for the Host Dataverse dropdown in Create Dataset Page

stevenwinship:
**What this PR does / why we need it**: Result lists for collections can be quite large so a way to filter and use pagination is needed.

**Which issue(s) this PR closes**:https://github.com/IQSS/dataverse/issues/12423

- Closes #12423

**Special notes for your reviewer**: Switched to family "A" pagination. I did not change the output from "data.items" to "data" since that would be a breaking api change and data also contains count and totalCount.

**Suggestions on how to test this**: See DataRetrieverApi tests. create many dataverses owned or given access to a user. Test getting the list with and without pagination. Get the list as a superuser. Filter the list with a sort term.

**Additional documentation**:

https://dataverse-guide--12436.org.readthedocs.build/en/12436/api/native-api.html#list-dataverse-collections-a-user-can-act-on-based-on-their-permissions


qqmyers (dismissed):
I've looked through the code and it looks like it addresses the issue. 

However, it seems to be a one off approach - no other paginated apis use 'nextOffset' and 'previousOffset', this api asks for offset (in # of items, not pages) and pageSize, while /api/datasets/{id}/versions/{versionId}/files uses offset and limit, this api returns a 'count' while the /files api and three others return 'totalCount', written by an ok() method while this one sends a 'count' (in all cases, but I think this is the number returned when paging, not all that exist) along with a duplicate 'pageSize' when paging is used and these are all done in the jsonArray(LIst<Dataverses>, Pager) call (rather than somewhere reusable).

I don't know how much to push into any give PR, but, since pagination is a common theme in the API changes for the modern interface, can we standardize an approach (has this been discussed already?)? Are offsets in terms of pages or items? Do API calls return prev and next? Are the items or page numbers? Etc. Can the apis use the same keys? In the same json structure?

I've marked this as requesting changes, but I think we made need some discussion to decide what (if anything) should actually be changed before this goes forward.

pdurbin:
At standup we talked about how we have various patterns for pagination, which I wrote about here: https://github.com/IQSS/dataverse/commit/cfc5301be5419cb9e3cc72b4c721d46d2bbde718

| Family | Endpoint | Request parameters | Response keys |
| --- | --- | --- | --- |
| A | `GET /api/datasets/{id}/versions/{versionId}/files` | `offset`, `limit` | `totalCount`, `data` |
| A | `GET /api/datasets/{id}/versions/compareSummary` | `offset`, `limit` | `totalCount`, `data` |
| A | `GET /api/notifications/all` | `offset`, `limit` | `totalCount`, `data` |
| A | `GET /api/files/{id}/versionDifferences` | `offset`, `limit` | `totalCount`, `data` |
| B | `GET /api/search` | `start`, `per_page` | `total_count`, `start`, `count_in_response`, `items` |
| C | `GET /api/admin/list-users` | `selectedPage`, `itemsPerPage` | `userCount`, `selectedPage`, `pagination{...}`, `users` |
| C | `GET /api/access/datafile/{id}/listRequests` | `start`, `per_page` | `data`, `pagination{...}` |
| D | `GET /api/users/{identifier}/allowedCollections/{permission}` | `offset`, `pageSize` | `count`, `items`, `pageSize`, `nextOffset`, `prevOffset` |
| D | `GET /api/mydata/retrieve/collectionList` | `offset`, `pageSize` | `count`, `items`, `pageSize`, `nextOffset`, `prevOffset` |

@stevenwinship will try to switch to the "A" family that the frontend expects.

Also, there's a Pager.java class at https://github.com/IQSS/dataverse/blob/v6.11/src/main/java/edu/harvard/iq/dataverse/mydata/Pager.java . We might want to create a new Pager class to rule them all. 💍

PR #12449, opened 2026-06-09
Fix license description to use DB shortDescription as fallback instead of name

### LINKED ISSUE #9529: License Name used as Description when License.properties info for the license doesn't exist
Dataverse displays the license description as the title attribute for the license icon and as a pop-up when you hover over the license. When the description for the license has not been added to the License.properties file, the current code defaults to returning the license name instead of the database value for the license description. This means the license description is not available in the UI in those cases, and it causes a minor accessibility warning in Deque Axe as the title for the image duplicates the text next to the icon.

* When does this issue occur?
Whenever License.properties does not have the description for a given license

Dataset page

* What happens?
image
instead of 
image
(images use two different QDR licenses but in the first the popup is just the name of the license instead of the description ("QDR Controlled Access Conditions"))

the default should be the database value of the description.

5.13/dev

Root cause: https://github.com/IQSS/dataverse/blob/61da081f1fb2520370b21a40fb940481ccea3ca5/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java#L584-L601

won-seoop:
## Summary

Fixes #9529

When a license's description key is missing from `License.properties`, `getLocalizedLicenseDetails()` was falling back to the **license name** instead of the stored `shortDescription` value in the database. This caused:

- The license icon's `title` attribute to show the license name instead of the description
- A minor accessibility warning in Deque Axe (image title duplicates adjacent text)

## Changes

**`DatasetUtil.getLocalizedLicenseDetails`**:
- When `DESCRIPTION` lookup fails (key absent in `License.properties` or exception), now falls back to `license.getShortDescription()` (the DB-stored value) instead of `license.getName()`
- Falls back further to `licenseName` only if `shortDescription` is also `null`
- `NAME` lookups retain the existing fallback to `licenseName` (no behavior change)

**`DatasetUtilTest`**: Added three regression tests covering:
- DESCRIPTION fallback to `shortDescription` when key is absent from properties
- DESCRIPTION fallback to `licenseName` when `shortDescription` is also null
- NAME fallback to `licenseName` (ensuring existing behavior is preserved)

## Test plan

- [ ] Run `./mvnw test -pl src/test -Dtest=DatasetUtilTest` to verify new unit tests pass
- [ ] Manually add a custom license via the API without a matching entry in `License.properties` and confirm the hover tooltip shows the database description rather than the license name

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
+++ b/src/main/java/edu/harvard/iq/dataverse/dataset/DatasetUtil.java
@@ -709,11 +709,15 @@ public class DatasetUtil {
             }
         }
         catch (Exception e) {
-            localizedLicenseValue = licenseName;
+            localizedLicenseValue = null;
         }
 
         if (localizedLicenseValue == null) {
-            localizedLicenseValue = licenseName ;
+            if (LicenseOption.DESCRIPTION.name().equals(keyPart) && license.getShortDescription() != null) {
+                localizedLicenseValue = license.getShortDescription();
+            } else {
+                localizedLicenseValue = licenseName;
+            }
         }
         return localizedLicenseValue;
     }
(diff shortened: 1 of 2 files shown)


stevenwinship (approved):

qqmyers:
Looks good - thanks for closing a 2023 bug!

Issue #12376, opened 2026-05-05
Feature Request: Don't create a dataset draft when uploading unchanged metadata


vera:
**Overview of the Feature Request**

Currently, when uploading metadata via the API (e.g., `PUT /api/datasets/:persistentId/versions/:draft`), Dataverse creates a new draft version even if the uploaded metadata is identical to the existing version. This leads to the creation of "empty" drafts.

These empty drafts are visible in the UI's dataset page "Versions" tab with an empty change summary.

<img width="827" height="132" alt="Image" src="https://github.com/user-attachments/assets/47144048-6f8a-4a1a-9684-517b16dd34ab" />

.

<img width="897" height="166" alt="Image" src="https://github.com/user-attachments/assets/eae52d2a-fb72-4a2e-9413-7f68d665568e" />

I propose that Dataverse should automatically detect if the incoming metadata matches the current version and, if so, skip the creation of a new draft, and return a successful response without modifying the dataset state.

**What kind of user is the feature intended for?**
(Example users roles: API User, Curator, Depositor, Guest, Superuser, Sysadmin)

API User, Sysadmin

**What inspired the request?**

Automated scripts that synchronize external metadata with Dataverse. Currently, these scripts must implement client-side logic (retrieving existing metadata, normalizing it, and comparing it with the new data) to decide whether an update is necessary.

**What existing behavior do you want changed?**

The API currently accepts metadata and creates/updates a draft version regardless of whether the content has actually changed.

**Any brand new behavior do you want to add to Dataverse?**

Dataverse should perform a comparison between the incoming metadata and the latest version. If they are semantically equivalent, the API should return a success status (e.g., `200 OK`) without creating a new draft or modifying the existing one.

**Any open or closed issues related to this feature request?**

Not aware of any.

**Are you thinking about creating a pull request for this feature?**  
Help is always welcome, is this feature something you or your organization plan to implement?

Yes

cc @annegerlach

PR #12452, opened 2026-06-11
feat: prevent draft creation for unchanged dataset metadata

AnneGerlach:
## What this PR does / why we need it:

🎫 **Closes:** #12376 

This PR prevents Dataverse from creating or updating a draft version when metadata uploaded via `PUT /api/datasets/{id}/versions/:draft` is semantically identical to the existing dataset version.

Previously, repeated uploads of unchanged metadata created unnecessary draft versions. This was particularly problematic for automated synchronization workflows, which had to implement their own comparison logic before submitting updates.

The implementation reuses the existing `DatasetVersionDifference` logic to detect unchanged metadata and terms updates and skips draft creation or update when no effective changes are detected.

## Changes

* Added a no-op check before draft creation and draft update in `PUT /api/datasets/{id}/versions/:draft`
* Reused the existing `DatasetVersionDifference` implementation to detect unchanged metadata and terms updates
* Skipped draft creation/update and returned the existing dataset version when no effective changes are detected
* Added integration tests covering both scenarios:

  * no draft exists
  * draft already exists

Run tests:

```bash
mvn test -Dtest=DatasetsTest,DatasetsIT
```

or tests manually:

### Test no-op metadata update

1. Login as `dataverseAdmin`
2. Create and publish a dataset via the UI
3. Generate an API token and copy the dataset PID (export those: API_TOKEN and PID)

```bash
# Download latest version metadata
curl -H "X-Dataverse-key: $API_TOKEN" \
"http://localhost:8080/api/datasets/:persistentId?persistentId=$PID" \
| jq '.data.latestVersion' > latestVersion.json
```

```bash
# Re-upload unchanged metadata
curl -X PUT \
-H "X-Dataverse-key: $API_TOKEN" \
-H "Content-Type: application/json" \
--data @latestVersion.json \
"http://localhost:8080/api/datasets/:persistentId/versions/:draft?persistentId=$PID"
```

```bash
# Verify no draft was created
curl -H "X-Dataverse-key: $API_TOKEN" \
"http://localhost:8080/api/datasets/:persistentId/versions?persistentId=$PID" \
| jq
```

### Test changed metadata

```bash
# Modify metadata in the file e.g.
nano latestVersion.json
```

```bash
# Upload changed metadata
curl -X PUT \
-H "X-Dataverse-key: $API_TOKEN" \
-H "Content-Type: application/json" \
--data @latestVersion.json \
"http://localhost:8080/api/datasets/:persistentId/versions/:draft?persistentId=$PID"
```

```bash
# Verify draft now exists
curl -H "X-Dataverse-key: $API_TOKEN" \
"http://localhost:8080/api/datasets/:persistentId/versions?persistentId=$PID" \
| jq
```

## Special notes for your reviewer:

Thanks for reviewing 🌻 (and special thanks to @vera for reviewing beforehand :tulip: )

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -849,7 +849,12 @@ public class Datasets extends AbstractApiBean {
                 return error( Response.Status.BAD_REQUEST, "You may not add files via this api.");
             }
 
-            boolean updateDraft = ds.getLatestVersion().isDraft();
+            DatasetVersion latestVersion = ds.getLatestVersion();
+            if (isDatasetVersionNoOp(incomingVersion, latestVersion)) {
+                return ok(json(latestVersion, true));
+            }
+
+            boolean updateDraft = latestVersion.isDraft();
 
             DatasetVersion managedVersion;
             if (updateDraft) {
@@ -880,6 +885,12 @@ public class Datasets extends AbstractApiBean {
             return ex.getResponse();
 
         }
+    }
+
+    // Helper extracted to make no-op detection testable.
+    static boolean isDatasetVersionNoOp(DatasetVersion incomingVersion, DatasetVersion latestVersion) {
+        DatasetVersionDifference diff = new DatasetVersionDifference(incomingVersion, latestVersion);
+        return diff.getDetailDataByBlock().isEmpty() && diff.getChangedTermsAccess().isEmpty();
     }
 
     @GET
--- a/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
+++ b/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
@@ -850,9 +850,10 @@ public class Datasets extends AbstractApiBean {
             }
 
             DatasetVersion latestVersion = ds.getLatestVersion();
-            if (isDatasetVersionNoOp(incomingVersion, latestVersion)) {
-                return ok(json(latestVersion, true));
-            }
+            // TODO Implement no-op detection
+//            if (isDatasetVersionNoOp(incomingVersion, latestVersion)) {
+//                return ok(json(latestVersion, true));
+//            }
 
             boolean updateDraft = latestVersion.isDraft();
 
--- a/doc/release-notes/12376-skip-empty-drafts-for-unchanged-metadata.md
+++ b/doc/release-notes/12376-skip-empty-drafts-for-unchanged-metadata.md
@@ -1,0 +1,1 @@
+When metadata uploaded via `PUT /api/datasets/{id}/versions/:draft` is identical to the existing dataset version, Dataverse no longer creates or updates a draft version. This prevents repeated uploads of unchanged metadata from creating unnecessary drafts or rewriting an existing draft.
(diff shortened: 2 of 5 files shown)


src/main/java/edu/harvard/iq/dataverse/api/Datasets.java
```diff
@@ -849,7 +849,12 @@ public Response updateDraftVersion(@Context ContainerRequestContext crc, String
                 return error( Response.Status.BAD_REQUEST, "You may not add files via this api.");
             }
 
-            boolean updateDraft = ds.getLatestVersion().isDraft();
+            DatasetVersion latestVersion = ds.getLatestVersion();
```
stevenwinship, line 852: Could you add a release note explaining the change

stevenwinship:
@AnneGerlach  When I tried this, changing the dataset contact email, it was treated as a no-op

Before:
"datasetContactEmail": {
                "typeName": "datasetContactEmail",
                "multiple": false,
                "typeClass": "primitive",
                "value": "spruce@mailinator.com"
              }

Here is what I sent:
"datasetContactEmail": {
                "typeName": "datasetContactEmail",
                "multiple": false,
                "typeClass": "primitive",
                "value": "spruce1@mailinator.com"
              }

Nothing changed when I get the draft:
"datasetContactEmail": {
                                    "typeName": "datasetContactEmail",
                                    "multiple": false,
                                    "typeClass": "primitive",
                                    "value": "spruce@mailinator.com"
                                }

Without this PR the email value gets updated
"value": "spruce1@mailinator.com"

vera:
> @AnneGerlach When I tried this, changing the dataset contact email, it was treated as a no-op

You're right. It seems this approach of simply re-using the `DatasetVersionDifference` implementation to detect no-op updates doesn't work.

Since `DatasetVersionDifference` uses display values and not the raw values for comparison, this fails to detect changes for a couple of specific field types:

- Compound child fields whose display format hides their actual value, such as #EMAIL (e.g. `datasetContactEmail`)
- Primitive fields with fieldType = "email", which are completely skipped during the comparison
- Primitive fields when only the letter case is changed, because primitive comparisons are case-insensitive
- Fields with display sanitization or escaping, where values become identical after sanitization/escaping

I've added examples for these to the `testUpdateDatasetMetadataNoOp`, so the IT tests are going to fail now.

To fix this, the change detection needs to be reworked to be based on raw values, not display values.

vera:
As a next step, I temporarily disabled no-op detection and added a comprehensive DatasetsIT test documenting the current behavior of the update-draft API (PUT `/api/datasets/:persistentId/versions/:draft`). The test covers the fields accepted by the API and is intended to catch any regressions introduced by no-op detection.

Note that this may have unearthed some unintended behaviour, or at least I'm not sure if it's intended, which is that it's possible for users to update the following DatasetVersion fields when the API is used to create a new draft:

* deaccessionLink
* deaccessionNote
* releaseDate
* archiveTime
* UNF

See here: https://github.com/IQSS/dataverse/pull/12452/changes/641b418702f044a4461bb558eeddccc5e110e288#diff-f7f8c09756ce44634bf7d9a243e628166a608079fffe7cd71f77f56f949114f6R7674

Confirmed that the new test passes with no-op detection disabled.

vera:
Change detection is now reworked. All relevant tests pass locally. Ready for re-review

stevenwinship:
@vera Could you update this branch and make sure all the tests are passing before I re-review it? Thanks

vera:
> @vera Could you update this branch and make sure all the tests are passing before I re-review it? Thanks

Done!

stevenwinship:
Please include a Unit Test for DatasetVersionUpdateDifference. 

Edit: I see that the DatasetsTest unit test covers this class.

stevenwinship (approved):
The new changes look good and have fixed the issue I was seeing.

PR #12454, opened 2026-06-11
S3 fixes for non AWS instances

qqmyers:
**What this PR does / why we need it**: QDR has used [storJ](https://www.storj.io/) and is now switching to use [BackBlaze B2](https://www.backblaze.com/cloud-storage). 

This PR fixes a critical issue we found using B2 which may be relevant for other stores that do not support tagging (which we use to tag files as temporary until the upload completes). In short, while the .disable-tagging=true setting stops Dataverse from writing a tag, it still tried to delete the tag even with that flag (resulting in B2 deleting the file :-)). The change here only tries to delete the tag if we set it/disable-tagging is false/not set. (This appears to be the only change needed for B2 - I've added it to the list of compatible S3 implementations.)

For storJ, which works fine, we noticed that the /cleanStorage APi call to delete unnecessary files (i.e. upload attempts where the file was sent to S3 and the call to addFiles to Dataverse didn't happen/failed) fails when there are many files in the dataset (probably > 1000). This PR updates the code to send a maxKeys(1000) when any list of objects is done - apparently without this storJ doesn't send a continuation token.

With the fix for storJ and /cleanStorage, the PR refactors the code to list objects in S3 to reduce duplicate/unnecessary code.

- Closes #

**Suggestions on how to test this**: The changes have been tested at QDR. If more than regression testing is desired, one can get a free B2 account (up to 10GB) for testing and set up a B2 store - should work in creating datasets, etc. as any other S3 store. For storJ, the problem can only be seen with more than 1K files and only with the /cleanStorage api call - happy to try and provide details if someone wants to recreate the issue. (I did have a local config issue causing trouble as well so I can't guarantee that setting maxkeys was required, but I have verified the new code works with storj and B2.

(FWIW - the /cleanStorage API only looks in the store/bucket set for the dataset, so if files have been added to a different bucket (the store was changed at some point before/after uploads), it misses bad files and doesn't show any error.)

**Additional documentation**:

Preview docs at https://dataverse-guide--12454.org.readthedocs.build/en/12454/installation/config.html#reported-working-s3-compatible-storage


doc/sphinx-guides/source/installation/config.rst
```diff
@@ -1479,6 +1479,12 @@ You may provide the values for these via any `supported MicroProfile Config API
 Reported Working S3-Compatible Storage
 ######################################
 
+
+`BackBlaze B2 <https://www.backblaze.com/cloud-storage>`_
+  (as of 6/11/2026)
```
```suggestion
  (As of 2026-06-11)
```

Image

https://xkcd.com/1179/

src/main/java/edu/harvard/iq/dataverse/S3PackageImporter.java
```diff
@@ -77,23 +77,17 @@ public void copyFromS3(Dataset dataset, String s3ImportPath) throws IOException
...
-
-        List<S3Object> storedDcmDatasetFilesSummary = new ArrayList<>(listRes.contents());
-
-        while (listRes.isTruncated()) {
-            logger.fine("S3 listAuxObjects: going to next page of list");
-            listReq = listReq.toBuilder().continuationToken(listRes.nextContinuationToken()).build();
-            listRes = s3.listObjectsV2(listReq);
-            storedDcmDatasetFilesSummary.addAll(listRes.contents());
+            throw new IOException("S3 listObjects: failed to get a listing for " + dcmDatasetKey);
         }
```
pdurbin, line 92: I'm just putting this note at the bottom.

There are three failing checks:

Image

@qqmyers can you please take a look?

pdurbin (commented):
A quick review.

src/main/java/edu/harvard/iq/dataverse/S3PackageImporter.java
pdurbin: Tests are passing now.

pdurbin (approved):
Tests are passing and docs are looking good now. Approved.

We should decide how much testing we want to do. I haven't done any myself.

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -697,56 +697,53 @@ private File createTempFile(Path path, InputStream inputStream) throws IOExcepti
         return targetFile;
     }
```
pdurbin, line 737: @qqmyers there are merge conflicts now that #12413 has been merged. Can you please resolve them?

src/main/java/edu/harvard/iq/dataverse/dataaccess/S3AccessIO.java
```diff
@@ -735,63 +735,56 @@ private File createTempFile(Path path, InputStream inputStream) throws IOExcepti
...
```
pdurbin, line 737: From https://github.com/IQSS/dataverse/actions/runs/33790054932/job/100764204354?pr=12454

```
2026-09-03T19:08:29.2397160Z [ERROR] Failures: 
2026-09-03T19:08:29.2399163Z [ERROR]   DatasetsIT.testDeleteDatasetWhileFileIngesting:2461 1 expectation failed.
2026-09-03T19:08:29.2400385Z JSON path message doesn't match.
2026-09-03T19:08:29.2401083Z Expected: Dataset cannot be edited due to dataset lock.
2026-09-03T19:08:29.2401791Z   Actual: null
```

However, that was from a2a298beb26697a3468b16f8fe10f737c22c51ec and there's a newer commit so let's see what happens. Tests are still running.

pdurbin: Tests are passing now. Resolving.

PR #12455, opened 2026-06-11
Fix missing file sizes

qqmyers:
**What this PR does / why we need it**: QDR has noticed a production case where ~2000 files did not have file sizes in the database. It is believed that this can/could have occurred during S3 direct upload when the code tries to retrieve the size from S3. If the S3 store is only 'eventually consistent' this call can fail, resulting in a null file size in the database. (Whether this is possible now or was only an issue in past releases is not clear.).

Overall, this state is non-fatal, but it does affect the reported download size when downloading a dataset (i.e. the download may be bigger than shown due to the files with no size not adding to the total).

This PR adds an /api/admin/datafiles/integrity/fixmissingfilesizes (somewhat analogous to the one to fixmissingoriginalfilesizes) that will, for stores indicating that Dataverse can access the files (e.g. not a remote store where the URL may be blocked or point to a remote trusted store landing page whose size is not that of the real file), try to retrieve the size and update the database.

It may be worth seeing whether there are instances for which `select count(*) from datafile where filesize is null;` is not zero to decide whether this is a useful addition for the community (feel free to comment on this PR if you see this in your installation).

- Closes #

**Suggestions on how to test this**: It's not clear how to trigger this condition - easiest to just delete some filesizes from a test database and confirm that the api call restores them.

**Additional documentation**:


src/main/java/edu/harvard/iq/dataverse/api/Admin.java
```diff
@@ -1723,6 +1723,40 @@ public Response fixMissingOriginalSizes(@Parameter(description = "Maximum number
...
+    @Path("datafiles/integrity/fixmissingfilesizes")
+    @AuthRequired
+    @POST
+    public Response fixMissingFileSizes(@Context final ContainerRequestContext crc, @QueryParam("limit") Integer limit) {
+
+        User u = getRequestUser(crc);
+        if (!u.isSuperuser()) {
+            return error(Status.FORBIDDEN, BundleUtil.getStringFromBundle("admin.api.auth.mustBeSuperUser"));
+        }
+        JsonObjectBuilder info = Json.createObjectBuilder();
```
stevenwinship: Change to JsonUtil.create
You may need to rebase first

stevenwinship (approved):

PR #12469, opened 2026-06-16
#12464 update code to work with both versions of display value string

### LINKED ISSUE #12464: Trailing comma when selecting a single country (e.g. Afghanistan)

I'm on 1e03c0960b on the develop branch (pre 6.11 release). I can't reproduce this on demo running 6.10.1 ~~so I believe it's a regression~~.

Note that #11507 has not been merged yet so it's unrelated to this ~~regression~~ bug.

How to reproduce:

- Enable the geospatial block
- Select Afganistan as the country
- Observe that there's a trailing comma in the "view mode) (screenshots below)

Image
Image

Here's how it looks on demo ( https://demo.dataverse.org/dataset.xhtml?persistentId=doi:10.70122/FK2/6TJX7C ), with no trailing comma:

Image

sekmiller:
**What this PR does / why we need it**: Fixes the removal of the trailing comma in a Compound Value field regardless of whether the Display Value string is "VALUE," or "VALUE, ".

- Closes #12464 

**Special notes for your reviewer**: Probably a more elegant way to do this, but I wanted to make sure it would work for the two display value strings that we know about.

**Suggestions on how to test this**: Remember for the geospatial it will remove the trailing comma for each given group of values (Country, State, City). There are no commas separating the groups - they are put on a new line. So if all your geospatial items are countries then there are no commas at all.

**Is there a release notes update needed for this change?**: no. fixing something introduced by another PR

**Additional documentation**:

### DIFF
--- a/src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
+++ b/src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
@@ -177,12 +177,16 @@ public class DatasetFieldCompoundValue implements Serializable {
                 if (StringUtils.isBlank(format)) {
                     format = "#VALUE";
                 }
+                /*
+                if (format.equals("#VALUE,")) {
+                    format = "#VALUE, ";
+                }*/
                 String sanitizedValue = childDatasetField.getDatasetFieldType().isSanitizeHtml() ? MarkupChecker.sanitizeBasicHTML(childDatasetField.getValue()) :  childDatasetField.getValue();
                 if (!childDatasetField.getDatasetFieldType().isSanitizeHtml() && childDatasetField.getDatasetFieldType().isEscapeOutputText()){
                     sanitizedValue = MarkupChecker.stripAllTags(sanitizedValue);
                 }
                 //if a series of child values is comma delimited we want to strip off the final entry's comma
-                if (format.equals("#VALUE, ")) fixTrailingComma = true;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
```diff
@@ -177,12 +177,16 @@ public Map<DatasetField, String> getDisplayValueMap() {
                 if (StringUtils.isBlank(format)) {
                     format = "#VALUE";
                 }
+                /*
+                if (format.equals("#VALUE,")) {
+                    format = "#VALUE, ";
+                }*/
```
pdurbin: I'd suggest deleting this.

pdurbin (commented):
Just a tiny comment. I haven't tested the code but I'm ready to approve, once the tests pass.

src/main/java/edu/harvard/iq/dataverse/DatasetFieldCompoundValue.java
sekmiller: Thanks.

sekmiller: taken care of and checked in

pdurbin (approved):
I haven't tested this, but it looks like a good fix to me and API tests are passing. Approved.

stevenwinship:
image

